PR #491, opened 2022-06-01
feat: Fetcher field selection and optimistic filter evaluation

### LINKED ISSUE #490: Fetcher field selection and filter optimization
Need to add field selection to the filter so it actually fetch fields of documents we care about only. Additionally, we can support filtering during fetching, so we can optimistically ignore a document if it doesn't match.

jsimnz:
## RELEVANT ISSUE(S)

Resolves #490 

## DESCRIPTION

Refactors the db document fetcher to handle
1) Field selection
2) Filter selection

Side-effect is to optimize the fetcher to use the BadgerBD efficient key only iteration, which required additional changes in the go-datastore interfaces, BadgerDB, and BadgerDB datastore interface.

This has significant performance increases depending on the type of query (2-4x)

This is a pretty notable change, not in its design, but the side effect in the other packages. We can track the required changes for the dependant repos:
BadgerDB - https://github.com/sourcenetwork/badger/pull/1
go-datastore - https://github.com/sourcenetwork/go-datastore/pull/1

This PR is an *extreme* WIP, still has lingering artifacts from the several previous attempts at this goal, and debug print statements. 

Additionally, many additional benchmarks have been added throughout this effort and need to be cleaned up.

### HOW HAS THIS BEEN TESTED?

At the moment it uses the existing integration testing suite. I focused on the query tests, there may be others this breaks.

### CHECKLIST:

- [x] I have commented the code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the repo-held documentation.
- [x] I have made sure that the PR title adheres to the conventional commit style (subset of the ones we use can be found under: [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)

### ENVIRONMENT / OS THIS WAS TESTED ON?

Please specify which of the following was this tested on (remove or add your own):
- [ ] Arch Linux
- [x] Debian Linux
- [ ] MacOS
- [ ] Windows


jsimnz:
Closing as this is too old and there is a new PR with a diff approach #1500

PR #773, opened 2022-09-03
test: Restructure schema defaults under a separate package

shahzadlone:
Resolves #TBD (once feedback is received)

## Description
This PR is inspired by a discussion I had with @AndrewSisley. Attempts to move the "defaults" of schema testing logic into a separate folder / package as it can grow massive over time.

The scope of this PR isn't and shouldn't be to resolve all limitations of the schema testing package, but only to move the defaults we currently have into the default package, and split it into logical source files.

### DIFF
--- a/tests/integration/schema/defaults/args.go
+++ b/tests/integration/schema/defaults/args.go
@@ -10,14 +10,6 @@
 
 package defaults
 
-var allDefaultGroupArgs = []Field{
-	FilterArg,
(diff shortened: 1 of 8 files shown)


shahzadlone:
Closing as consensus is to prefer raw structures.

Moved the branch to: https://github.com/shahzadlone/defradb/tree/lone/test/refactor-schema-testing-defaults

PR #1346, opened 2023-04-13
tools: Add script to automate updating of cids

### LINKED ISSUE #1345: Tool to automate updating cids as much as possible
Throw together a quick bash script to help automate the tedious and manual process of updating the cids.

It currently takes dev's about 30-45 minutes to update these manually.

shahzadlone:
## Relevant issue(s)
Resolves #1345

## Description
- Automate the process of updating old cids to new cids.
- The script takes about 2-3 minutes on my WSL machine.
- Use `make fix:cids-unsafe` to run the script even if you have changes (uncommitted).
- Use `make fix:cids` to run in safe mode (only runs if you are at the root directory with no uncommitted changes).

### Looking for feedback:
- I don't know if this will work on MacOS or not. Would be nice if someone can test that.
- Would be nice if someone can double check it does what it needs to on linux as well.

## How has this been tested?
Tried to make a change to the composite which resulted in a cid change, after running this script it updated all the expected cids to actual cids.

- Linux (WSL2 - Manjaro)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,6 +186,14 @@ test\:bench-short:
 test\:scripts:
 	@$(MAKE) -C ./tools/scripts/ test
 
+.PHONY: fix\:cids
+fix\:cids:
+	./tools/scripts/fix_failing_cids.sh "safe"
+
+.PHONY: fix\:cids-unsafe
+fix\:cids-unsafe:
+	./tools/scripts/fix_failing_cids.sh "unsafe"
+
 # Using go-acc to ensure integration tests are included.
 # Usage: `make test:coverage` or `make test:coverage path="{pathToPackage}"`
 # Example: `make test:coverage path="./api/..."`
--- a/tools/scripts/fix_failing_cids.sh
+++ b/tools/scripts/fix_failing_cids.sh
@@ -1,0 +1,460 @@
+#!/bin/bash
+
+# Copyright 2023 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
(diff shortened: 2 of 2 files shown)


shahzadlone:
> I thought/assumed the only stuff in our tools folder was actually used in the CI (or in the future by users).
IMO this is a personal script, including it in this repo suggests we have a collective responsibility to maintain it (which I dont want, partly because there are much more maintainable ways to make cid changes easier). If you want a easy command to run it, you could just add an alias - can even alias it to make fix:cids 😁

I think I looked at the `tools/scripts` folder differently, I looked at them as "casual dump of helpful scripts" that may or may not be used in CI, with no requirement on them to be maintained with high priority. i.e. our `benchstat` script is broken and we haven't fixed it (not relevant until benchmark stuff is sorted) and is being used in CI as well.

question: what is meant by more maintainable ways? like a program in Go or something?

> That is a cool tool Shahzad. I would host it somewhere other than this repo though. Maybe something like github.com/sourcenetwork/cid-updater?

I am not sure if having a separate repo for every tool we might write is useful, also might cause friction in using it (have to clone every repo, make an alias to every cloned path). I would rather have a universal/common path containing my scripts / tools. For me that was either defradb repo under `tools/scripts` or my personal dotfiles.

So, I have posted the script to my personal dotfiles repo that has my other scripts and stuff: https://github.com/shahzadlone/lonefiles/blob/distro-arch-linux/.local/bin/fix_failing_cids.sh

Issue #1349, opened 2023-04-13
Tests sometime log datastore close error but still pass


AndrewSisley:
It looks like something is still occasionally trying to access the datastore after the database has been closed.  The test still passed in this case, but there is a chance that it may be occasionally responsible for test flakiness.  Rate of occurrence seems quite low (<~= 1:100), but I have seen it a few times now.

```
gotestsum --format testname -- ./tests/integration/subscription -race -shuffle=on
PASS tests/integration/subscription.TestSubscriptionWithUpdateAllMutations (0.21s)
PASS tests/integration/subscription.TestSubscriptionWithFilterAndOneCreateMutation (0.16s)
PASS tests/integration/subscription.TestSubscriptionWithFilterAndCreateMutations (0.12s)
PASS tests/integration/subscription.TestSubscriptionWithCreateMutations (0.13s)
PASS tests/integration/subscription.TestSubscriptionWithUpdateMutations (0.12s)
PASS tests/integration/subscription.TestSubscriptionWithFilterAndOneCreateMutationOutsideFilter (0.11s)
2023-04-13T16:36:42.440Z, ERROR, defra.db, datastore closed
PASS tests/integration/subscription
```

My best guess so far is that this is being logged in db/subscriptions.go `handleSubscription`, with an event being received after close:
```
for evt := range pub.Event() {
		txn, err := db.NewTxn(ctx, false)
		if err != nil {
			log.Error(ctx, err.Error())
			continue
		}
```

`*simpleChannel[T].Close()` is asynchronous, and does leave room for it to not have fully closed before the root datastore has closed.  It would however be odd that an 'extra' event could be received after the datastore has been closed - this would suggest that the test is exiting prematurely, and that extra, unexpected, events are being received.

AndrewSisley:
Related to: https://github.com/sourcenetwork/defradb/issues/1351

Issue #1338, opened 2023-04-12
Flaky - TestP2PWithMultipleDocumentsWithUpdateAndDeleteBeforeConnectSingleDeleteWithShowDeleted


shahzadlone:
**Describe the problem**
Flakey test.

**To Reproduce**
`make test`

**Expected behavior**
Not be flaky

**Platform**

latest commit on develop:
`bbe8408a (HEAD -> develop, origin/develop) 2 hours ago fix: API address parameter validation (#1311)
`

**Additional context**
```
=== Failed
=== FAIL: tests/integration/net/state/simple/peer TestP2PWithMultipleDocumentsWithUpdateAndDeleteBeforeConnectSingleDeleteWithShowDeleted (11.35s)
2023-04-12T11:22:09.996-0400    ERROR   dht     go-libp2p-kad-dht@v0.21.1/dht.go:655    failed to validate if peer is a DHT peer        {"peer": "12D3KooWQ8jR6HdUsPYZ3wPKs9SpaYT2giX6Hx2dsDfjm4DsVHAq", "error": "datastore closed", "errorVerbose": "datastore closed. Stack: /home/shahzad/~~/defradb/datastore/memory/errors.go:20 (0x1c62407)\n\tinit: ErrClosed      = errors.New(\"datastore closed\")\n/home/shahzad/.local/bin/go1.19.5/src/runtime/proc.go:6329 (0x482566)\n\tdoInit: f()\n/home/shahzad/.local/bin/go1.19.5/src/runtime/proc.go:6306 (0x4824b1)\n\tdoInit: doInit(t2)\n/home/shahzad/.local/bin/go1.19.5/src/runtime/proc.go:6306 (0x4824b1)\n\tdoInit: doInit(t2)\n/home/shahzad/.local/bin/go1.19.5/src/runtime/proc.go:6306 (0x4824b1)\n\tdoInit: doInit(t2)\n/home/shahzad/.local/bin/go1.19.5/src/runtime/proc.go:233 (0x474db3)\n\tmain: doInit(&main_inittask)\n/home/shahzad/.local/bin/go1.19.5/src/runtime/asm_amd64.s:1594 (0x4a90a1)\n\tgoexit: BYTE\t$0x90\t// NOP\n"}
2023-04-12T11:22:09.996-0400, INFO, defra.tests.integration, , {"Database": "badger-in-memory"}
[...]

PR #1352, opened 2023-04-13
fix: Guarantee event processing order

### LINKED ISSUE #1351: simpleChannel does not guarantee order of execution
For example if:

1) Create Doc
2) Subscribe to events

(2) may be processed before (1), resulting in the subscription erroneously receiving the event for (1).

Likely responsible for a significant portion of recent test flakiness.

AndrewSisley:
## Relevant issue(s)

Resolves #1351 #1349
May also resolve much of the current test flakiness, but there may be other underlying issues so I'll leave those open for a bit. It does not resolve all of the flakiness and I have seen one P2P local failure on this branch (logged in https://github.com/sourcenetwork/defradb/issues/1338 )

## Description

Modifies the simpleChannel to guarantee the processing order of stuff. 

The previous solution did not do this, and allowed new subscriptions to pick up database events that had previously occurred but had not yet been processed by `handleChannel`.

A common scenario for how this can result in a test failure is documented in #1351

Also includes a fix for https://github.com/sourcenetwork/defradb/issues/1349 as I believe it occurs in the same part of the codebase, for similar reasons, and there is no point putting up a second, dependant review for it.

### DIFF
--- a/events/simple.go
+++ b/events/simple.go
@@ -22,13 +22,9 @@ type simpleChannel[T any] struct {
 	isClosed        bool
 }
 
-type subscribeCommand[T any] struct {
-	subscriptionChannel Subscription[T]
-}
+type subscribeCommand[T any] Subscription[T]
 
-type unsubscribeCommand[T any] struct {
-	subscriptionChannel Subscription[T]
-}
+type unsubscribeCommand[T any] Subscription[T]
 
 type publishCommand[T any] struct {
 	item T
@@ -60,7 +56,7 @@ func (c *simpleChannel[T]) Subscribe() (Subscription[T], error) {
 	// It is important to set this buffer size too, else we may end up blocked in the handleChannel func
 	ch := make(chan T, c.eventBufferSize)
 
-	c.commandChannel <- subscribeCommand[T]{ch}
+	c.commandChannel <- subscribeCommand[T](ch)
 	return ch, nil
 }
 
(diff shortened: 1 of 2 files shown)


events/simple.go
```diff
@@ -11,26 +11,39 @@
...
+	// WARNING: This does mean that non-event commands can block the database if the buffer
+	// size is breached (e.g. if many subscribe commands occupy the buffer).
+	commandChannel  chan any
+	eventBufferSize int
+	hasClosedChan   chan struct{}
+	isClosed        bool
 }
 
-// NewSimpleChannel creates a new simpleChannel with the given subscriberBufferSize and
+type subscribeCommand[T any] Subscription[T]
```
AndrewSisley, line 28: Note: Fred suggested dropping the `command` suffix for these, or shortening it to `cmd`.  I have a minor preference for the current names (I like it being clear that they are commands, and not an e.g. `subscribe` type, I also prefer the more verbose name).

Happy to change this though if others want, or if Fred's preference is/becomes stronger than I think it is :)

islamaliev: I would be against dropping the suffix because at the moment it clearly shows that these types belong to a single concept. Shortening would be fine though especially because `cmd` is well-known. But my personal preference is to keep `command`

islamaliev (approved):
LGTM

events/simple.go
shahzadlone: Prefer to keep it the way it is.

shahzadlone (approved):
LGTM! Tested this quite a few times again, and can confirm it is significantly less flakey! Even had a pass on my machine that would never pass before.

events/simple.go
AndrewSisley: Will keep as-is then :)

AndrewSisley:
> LGTM! Tested this quite a few times again, and can confirm it is significantly less flakey! Even had a pass on my machine that would never pass before.

Really glad to hear - thanks again for your help with this one :)

PR #1362, opened 2023-04-17
fix: Improve NonNull error message

### LINKED ISSUE #1191: Error when adding non-null developer-defined type
Schema:
```
type A {
  s: String
}

type B {
  as: [A!]
}
```

Error on adding schema:
```
  {
    "message": "no type found for given name. Type: A",
    "extensions": {
      "status": 500,
      "httpError": "Internal Server Error"
    }
  }
```

AndrewSisley:
## Relevant issue(s)

Resolves #1191

## Description

Improves the NonNull error message when attempting to declare an inline array for an unsupported type.

### DIFF
--- a/tests/integration/schema/simple_test.go
+++ b/tests/integration/schema/simple_test.go
@@ -243,7 +243,7 @@ func TestSchemaSimpleErrorsGivenNonNullManyRelationField(t *testing.T) {
 						Dogs: [Dogs!]
 					}
 				`,
-				ExpectedError: "no type found for given name. Type: Dogs",
+				ExpectedError: "NonNull variants for type are not supported. Type: Dogs",
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


tests/integration/schema/simple_test.go
```diff
@@ -227,5 +227,26 @@ func TestSchemaSimpleErrorsGivenNonNullField(t *testing.T) {
 		},
 	}
 
-	testUtils.ExecuteTestCase(t, []string{"users"}, test)
+	testUtils.ExecuteTestCase(t, []string{"Users"}, test)
```
shahzadlone, line 230: thought: Would be nice to have an error raised upon incorrect / unknown Collection name.

shahzadlone (approved):
LGTM! Thanks

tests/integration/schema/simple_test.go
AndrewSisley: I think a better fix would to not require this param at all :)

PR #1366, opened 2023-04-17
test: Retry test doc mutation on transaction conflict 

### LINKED ISSUE #1365: Test utils CreateDoc action does not retry on txn conflict
Caused a test failure locally today.

AndrewSisley:
## Relevant issue(s)

Resolves #1365

## Description

Adds Retry mechanics for test doc create and delete on transaction conflict.

The logic existed for update, and has now been expanded to cover Create and Update.

It could probably be extended to cover requests, but I dont think any P2P tests use mutation requests at the moment.

### DIFF
--- a/tests/integration/utils2.go
+++ b/tests/integration/utils2.go
@@ -824,22 +824,40 @@ func updateDoc(
 	var expectedErrorRaised bool
 	actionNodes := getNodes(action.NodeID, nodes)
 	for nodeID, collections := range getNodeCollections(action.NodeID, nodeCollections) {
-		// If a P2P-sync commit for the given document is already in progress this
-		// Save call can fail as the transaction will conflict. We dont want to worry
-		// about this in our tests so we just retry a few times until it works (or the
-		// retry limit is breached - important incase this is a different error)
-		for i := 0; i < actionNodes[nodeID].MaxTxnRetries(); i++ {
-			err = collections[action.CollectionID].Save(ctx, doc)
-			if err != nil && errors.Is(err, badgerds.ErrTxnConflict) {
-				time.Sleep(100 * time.Millisecond)
-				continue
-			}
-			break
-		}
+		err := withRetry(
+			ctx,
+			actionNodes,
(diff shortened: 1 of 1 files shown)


tests/integration/utils2.go
```diff
@@ -824,24 +841,42 @@ func updateDoc(
...
 
+// withRetry attempts to perform the given action, retrying up to a DB-defined
+// maximum attempt count if a transaction conflict error is returned.
+//
+// If a P2P-sync commit for the given document is already in progress this
+// Save call can fail as the transaction will conflict. We dont want to worry
+// about this in our tests so we just retry a few times until it works (or the
+// retry limit is breached - important incase this is a different error)
+func withRetry(
+	ctx context.Context,
```
fredcarle, line 864: suggestion: `ctx` isn't used so you can probably remove it from the parameters.

tests/integration/utils2.go
```diff
@@ -824,24 +841,42 @@ func updateDoc(
...
+	nodes []*node.Node,
+	nodeID int,
```
fredcarle, line 866: suggestion: You could use `node *node.Node` instead and avoid having to pass 2 params.

fredcarle (approved):
Approving but I have a couple suggestions before you merge :)

tests/integration/utils2.go
AndrewSisley: Ops cheers - it was used, and then I switched to the func param.  Will remove

- [ ] rm ctx

tests/integration/utils2.go
AndrewSisley: I was 50-50 about that one, in my head at least a lot of the test utils code does stuff like that - preferring the refactoring out of finding the correct thing instead of saving on param count.  Will have a look in a bit and then see what I feel like in the moment unless you have a strong preference.

- [ ] maybe just pass in node

AndrewSisley: Leaving as-is

PR #1369, opened 2023-04-18
fix: Unlock replicator lock before returning error

### LINKED ISSUE #1367: error in peer.setReplicator can cause indefinite-lock
If a replicator already exists `p.mu` will lock, but never unlock, causing the next `Lock` call to block indefinitely.

I doubt this is causing flaky tests, but it should be fixed and is quick to do so.

AndrewSisley:
## Relevant issue(s)

Resolves #1367

## Description

Unlocks a lock that is held before returning an error.

Unlikely to be tripping up any of our tests, but is undesirable and should be fixed.


fredcarle (approved):
Nice catch 👍

PR #1405, opened 2023-04-25
fix: Limit the size of http request bodies that we handle

### LINKED ISSUE #1322: Use of `io.ReadAll` in http handler can result in denial of service attack
Multiple handlers use `io.Readall`.
https://github.com/sourcenetwork/defradb/blob/01bfd49410505d27a1719939ebbcd0c7dea96992/api/http/handlerfuncs.go#L127

AndrewSisley:
## Relevant issue(s)

Resolves #1322

## Description

Limits the size of http request bodies that we handle to make things slightly harder for anyone to DoS a defra node.

Timeout time preferred by me over a byte limit as I find it much more descriptive, and will scale nicely with the system/hardware-context (more powerful machines can handle bigger req. bodies). Is undoubtably a runtime cost involved though compared to just using an int.

I set all the limits to 1 second, as that seems like plenty for now.  Should be added to the config at somepoint IMO though (not now).

### DIFF
--- a/api/http/handlerfuncs.go
+++ b/api/http/handlerfuncs.go
@@ -11,11 +11,13 @@
 package http
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"mime"
 	"net/http"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	dshelp "github.com/ipfs/boxo/datastore/dshelp"
@@ -124,7 +126,7 @@ func execGQLHandler(rw http.ResponseWriter, req *http.Request) {
 				handleErr(req.Context(), rw, ErrBodyEmpty, http.StatusBadRequest)
 				return
 			}
-			body, err := io.ReadAll(req.Body)
+			body, err := readWithTimeout(req.Context(), req.Body, time.Second)
 			if err != nil {
 				handleErr(req.Context(), rw, errors.WithStack(err), http.StatusInternalServerError)
 				return
@@ -155,7 +157,7 @@ func execGQLHandler(rw http.ResponseWriter, req *http.Request) {
 }
 
 func loadSchemaHandler(rw http.ResponseWriter, req *http.Request) {
-	sdl, err := io.ReadAll(req.Body)
+	sdl, err := readWithTimeout(req.Context(), req.Body, time.Second)
 	if err != nil {
 		handleErr(req.Context(), rw, err, http.StatusInternalServerError)
 		return
@@ -182,7 +184,7 @@ func loadSchemaHandler(rw http.ResponseWriter, req *http.Request) {
 }
 
 func patchSchemaHandler(rw http.ResponseWriter, req *http.Request) {
-	patch, err := io.ReadAll(req.Body)
+	patch, err := readWithTimeout(req.Context(), req.Body, time.Second)
 	if err != nil {
 		handleErr(req.Context(), rw, err, http.StatusInternalServerError)
 		return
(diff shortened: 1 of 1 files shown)


api/http/handlerfuncs.go
```diff
@@ -124,7 +126,7 @@ func execGQLHandler(rw http.ResponseWriter, req *http.Request) {
 				handleErr(req.Context(), rw, ErrBodyEmpty, http.StatusBadRequest)
 				return
 			}
-			body, err := io.ReadAll(req.Body)
+			body, err := readWithTimeout(req.Context(), req.Body, time.Second)
```
fredcarle, line 129: suggestion: I don't think we should have a timeout function like this. Timeouts for http requests should be handled at the server level. We have a limited implementation of it at the moment but when we update Go to 1.20, we'll have some added functionalities to help us with this.

api/http/handlerfuncs.go
```diff
@@ -322,3 +324,32 @@ func subscriptionHandler(pub *events.Publisher[events.Update], rw http.ResponseW
 		}
 	}
 }
+
+// readWithTimeout reads from the reader until either EoF or the given maxDuration has been reached.
+func readWithTimeout(ctx context.Context, reader io.Reader, maxDuration time.Duration) ([]byte, error) {
```
fredcarle, line 329: suggestion: Taking my above suggestion into account, this could be changed to a `readWithLimit` function where the limit is the maximum payload size we will allow.

fredcarle (changes requested):
I think the structure of this change is good but the limiting parameter should be size and not time.

AndrewSisley:
> Im very much against a time based limit for requests. Requests can absolutely take more than a second if there is a large dataset and a complex/inefficient query.

Of course they can, but if we want to handle that then we can use a bigger timeout?

> The original problem was the fact that our original implementation would arbitrarily read the entire incoming request payload, which according to the HTTP spec for POST is technically unlimited, meaning you could send a 1TB request and crash the server.

You disagree that the proposed solution solves this?  And you don't think it scales better to the host hardware/context?  A fixed size limit for all would mean that a defra server running on highly constrained device would still be exposed to this, whereas a large device would be needlessly limited.

jsimnz:
> You disagree that the proposed solution solves this? And you don't think it scales better to the host hardware/context? A fixed size limit for all would mean that a defra server running on highly constrained device would still be exposed to this, whereas a large device would be needlessly limited.

Theres very few (none) instances where we need to support incoming request payloads larger than 4GB for example which is the common limit POST data. Now thats to accomodate binary uploads, I would say we don't need anything larger than a couple of MB. 

This can still be exposed from a configuration perspective. So large device/small device is somewhat irrelevant to this discussion.

Its more about what is a better metric to protect against this DOS attack, and the primary concern is unbounded read buffers, ie byte size limit.

Theres no benefit to time limit in context of the DOS attack, and I believe it is a poor metric to limit on as there is just too many different ways someone might unknowingly/accidentally hit that limit without understanding why their requests are failling.

AndrewSisley:
> ... Theres no benefit to time limit in context of the DOS attack, and I believe it is a poor metric to limit on as there is just too many different ways someone might unknowingly/accidentally hit that limit without understanding why their requests are failling.

Time limits the amount of bytes that can be read, in a way that allows the limit to roughly scale with the host capabilities, reducing the likelihood that users will need to (re)configure the value.

A couple of MB seems quite small if you thought 1 second was too small, do you have a specific default value in mind?

Note: If capping by byte count the loop will be removed, reading in chunks would serve no purpose in that case.

jsimnz:
> Time limits the amount of bytes that can be read

Yes, but not limited to the "bytes you can read" which is where my strong objection comes from. Limiting requests by time not only limits bytes you can read, it also encompasses time it takes to parse request, compile plan graph, execute query, serialize response, and write bytes back to client.

That is far more than just the amount of bytes. If the concern is the DOS from unbounded incoming byte buffers, then we **need** to limit the size of those buffers, and limit the max size of the incoming request payload.

> A couple of MB seems quite small if you thought 1 second was too small, do you have a specific default size-value in mind?

As the incoming request is only the GQL data in most cases, or the schema SDL, the "couple of MBs" was to anchor the conversation in the magnitude of data we're dealing with. Ultimately it doesn't matter what the exact limit is, its more that we get the magnitude correct (couple of MBs, 100s of MBs, GBs, etc).

The time based approach is simple non-sencicle🙃

api/http/handlerfuncs.go
jsimnz, line 129: in terms of concrete suggestion for my comments, this whole PR can be turned into a single line change, which also avoids context, additional goroutine for timeouts, select call etc.

```go
r := io.LimitReader(req.Body,  1 << (10 * 2)) // here  1 << (10 * 2) is a Megabyte, can be expressed however
body, err := io.ReadAll(r)
```

jsimnz (changes requested):

jsimnz (approved):
LGTM

fredcarle (approved):
Yass! I like this version a lot :)

api/http/handlerfuncs.go
AndrewSisley: `io.ReadAll` is potential very wasteful memory-wise as it relies on `append` for resizing the result slice. It could theoretically mean that the capacity (allocated mem cost) of the result is actually (2n - 1).

I also really dont want any more magic in the magic number, and doing that calc inline like that is not very readable, with or without a comment.

PR #1408, opened 2023-04-26
feat: Add allowed origins config

### LINKED ISSUE #1355: Add configuration for allowed origins
Currently, if users start developing web apps that access an instance of DefraDB directly, they will be confronted with CORS errors from the browser because the app and DefraDB will operate on different ports locally and possibly different domains publicly. 

We should allow configuration of the HTTP API with a list of allowed origins.

*Note*: `WithAllowedOrigins` already exists on the API but we need to add the configuration through the CLI and config files.

fredcarle:
## Relevant issue(s)

Resolves #1355 

## Description

This PR adds the option to set allowed origins for the HTTP API using the config file or a CLI flag.

## Tasks

## How has this been tested?

Unit tests and curl with allowedOrigins set to http://test.com
```shell
# this will result in an unsuccessful Preflight request
curl -i http://localhost:9181/api/v1/ping -H "Origin: http://no.com" -H "Access-Control-Request-Method: GET" -X OPTIONS

# this will result in an successful Preflight request
curl -i http://localhost:9181/api/v1/ping -H "Origin: http://test.com" -H "Access-Control-Request-Method: GET" -X OPTIONS
```

- MacOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,19 @@ Note: `sudo` is needed above for the redirection server (to bind port 80).
 
 A valid email address is necessary for the creation of the certificate, and is important to get notifications from the Certificate Authority - in case the certificate is about to expire, etc.
 
+## Supporting CORS
+
+When accessing DefraDB through a frontend interface, you may be confronted with a CORS error. That is because, by default, DefraDB will not have any allowed origins set. To specify which origins should be allowed to access your DefraDB endpoint, you can specify them when starting the database:
+```shell
+defradb start --allowedorigins=https://yourdomain.com
+```
+
+If running a frontend app locally on localhost, allowed origins must be set with the port of the app:
+```shell
+defradb start --allowedorigins=http://localhost:3000
+```
+
+The catch-all `*` is also a valid origin. 
 
 ## Community
 
--- a/README.md
+++ b/README.md
@@ -379,12 +379,12 @@ A valid email address is necessary for the creation of the certificate, and is i
 
 When accessing DefraDB through a frontend interface, you may be confronted with a CORS error. That is because, by default, DefraDB will not have any allowed origins set. To specify which origins should be allowed to access your DefraDB endpoint, you can specify them when starting the database:
 ```shell
-defradb start --allowedorigins=https://yourdomain.com
+defradb start --allowe-dorigins=https://yourdomain.com
 ```
 
 If running a frontend app locally on localhost, allowed origins must be set with the port of the app:
 ```shell
-defradb start --allowedorigins=http://localhost:3000
+defradb start --allowed-origins=http://localhost:3000
 ```
 
 The catch-all `*` is also a valid origin. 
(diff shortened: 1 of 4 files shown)


AndrewSisley:
> Note: The default has been set to * which will allow all origins to send requests to the API. I feel like that's a good default for development but not for prod for obvious reasons. Do we want to be restrictive by default or open by default?

My vote would be for the opposite, and have it locked down by default. Forgetting (or not knowing about this) could be an expensive mistake. I also think that locked down by default is the norm (I remember having to mess with stuff like this a few times to enable CORS, but never to block it).

AndrewSisley (approved):
LGTM.  Approving assuming that the default value conversation is resolved before merged :)

fredcarle:
> My vote would be for the opposite, and have it locked down by default. Forgetting (or not knowing about this) could be an expensive mistake. I also think that locked down by default is the norm (I remember having to mess with stuff like this a few times to enable CORS, but never to block it).

I'm happy with that 👍

README.md
```diff
@@ -375,6 +375,19 @@ Note: `sudo` is needed above for the redirection server (to bind port 80).
 
 A valid email address is necessary for the creation of the certificate, and is important to get notifications from the Certificate Authority - in case the certificate is about to expire, etc.
 
+## Supporting CORS
+
+When accessing DefraDB through a frontend interface, you may be confronted with a CORS error. That is because, by default, DefraDB will not have any allowed origins set. To specify which origins should be allowed to access your DefraDB endpoint, you can specify them when starting the database:
+```shell
+defradb start --allowedorigins=https://yourdomain.com
```
shahzadlone, line 382: suggestion:

How about just `origins`. Mostly because we don't have a `blockedorigins` or any other option like that. Also `allowedorigins` is less readable, perhaps `allowedOrigins` if you want to stick with the "allowed" word.

```suggestion
defradb start --allowedorigins=https://yourdomain.com
```

fredcarle: The thing with `--origins` is that it's a little vague and the http header is called `Allowed-Origins`. I feel like it makes more sense to have the full representation here as well. `allowedOrigins` could be nicer.

AndrewSisley: nitpick: I have a very minor preference for `allowed-origins`, but only if it we dont already camel case any other existing multi-word params/flags

shahzadlone: I do like `allowed-origins`, would prefer that over `allowedOrigins`

shahzadlone (approved):
LGTM

README.md
orpheuslummis: personally i'd like `--api.corsdomain` or `http.corsdomain` depending on which top-level key we want to represent the http system

fredcarle: I'd rather we stay closer to what it represents. `allowed-origins` is quite good.

orpheuslummis: so the fully qualified key from the config's perspective would be `api.allowed-origins`, and as an env. variable it would be `DEFRA_API_ALLOWED_ORIGINS` perhaps. i'm not sure yet about if it's necessary to quote such a key with a `-` in Yaml  - it might have to be

```yaml
[api]
"allowed-origins": ...
```

fredcarle: I don't think it is according to what I see here https://www.cloudbees.com/blog/yaml-tutorial-everything-you-need-get-started

PR #1410, opened 2023-04-27
chore: Enshrine defra logger names

### LINKED ISSUE #1148: Logger names are prepended with `defra.` make them less usable
At the moment,
- we prepend all logger names with `defra.`, e.g `defra.http`, `defra.node`, etc.
- using the full `defra.module` name is necessary from the config perspective, reducing convenience. e.g. `--loglevel fatal,defra.http=debug`

We would like to be able to use only defra `module` name instead in config, e.g. `--loglevel fatal,http=debug`

orpheuslummis:
## Relevant issue(s)

Resolves #1148

## Description

Simple approach: just rename the loggers.
Suggestion: non-defra loggers could have a prefix like e.g. `x.` 

## Tasks

## How has this been tested?

CI

### DIFF
--- a/api/http/handler_test.go
+++ b/api/http/handler_test.go
@@ -86,7 +86,7 @@ func TestNewHandlerWithLogger(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	assert.Equal(t, "defra.http", kv["logger"])
+	assert.Equal(t, "http", kv["logger"])
 }
 
 func TestGetJSON(t *testing.T) {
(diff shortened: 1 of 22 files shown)


AndrewSisley (approved):
I'm approving the code changes, but am unsure as to whether this is a good feature change long-term.

I can see the `defra` prefix as being quite handy when embedding defra. I do think explicilty hardcoding the namespaces for every logger instance is quite bad anyway though, and we should probably switch the constructor to some kind of parent-child like mechanic where the namespace is constructed without redefining the strings on every instance - is a problem for another day though, and the code changes here facilitate that somewhat.

orpheuslummis:
That said, it is a good idea to figure out how Defra logging can play nice in the embedding case

PR #1414, opened 2023-04-27
ci: Reconfigure CodeCov action to ensure stability

### LINKED ISSUE #1413: Fix code covs flakey fails and don't run on fork pushes
Fix code covs flakey behavior and don't run on fork pushes

shahzadlone:
## Relevant issue(s)
Resolves #1413

## Description
Note: The first 3 commits are house keeping stuff.

- Ensures Codecov doesn't run on every push (code coverage reports will only generate for every PR, or pushes on `master` and `develop`). This also means contributors don't need to have codecov to have a build run successfully on their fork pushes.
- Partial hits are considered hits, we tried to find a way to do this a while back but this feature didn't exist.
- If for whatever reason the code cov token doesn't exist, still run the action (but might be flakey - is a non-issue unless infuture we enable codecov on pushes).

## How has this been tested?
Fork pushes and this PR pushes.


AndrewSisley:
> Partial hits are considered hits, we tried to find a way to do this a while back but this feature didn't exist.

This sounds like us fudging the numbers, prioritising marketing over engineering - what are your reasons for wanting to do this?

.github/codecov.yml
```diff
@@ -16,7 +16,7 @@ coverage:
     project:
       default:
         target: 50%
-        threshold: 0.05 # allow this much decrease on project
+        threshold: 5% # allow this much decrease on project
```
AndrewSisley, line 19: question: why?

AndrewSisley (changes requested):
I have two questions, one major, one minor that need resolving before merge please :)

.github/codecov.yml
shahzadlone: I noticed that I was getting a lint warning, and that the code cov documentation used percentages instead of decimals as well.

AndrewSisley:
> I disagree that this is fudging the numbers, I would argue that our test coverage right now as it stands is numbers fudged against us. I can share more thoughts in a discord thread as this seems like a good discussion point. It's probably also best to do this in a separate PR.

Good call, can you please remove this from this PR? Even if we all agreed this is what we want I think it would be very good for it to have its own commit, and documentation on why we want to do this (if we do). A discord thread would also be good to figure out where the team stands on this.

shahzadlone:
> > I disagree that this is fudging the numbers, I would argue that our test coverage right now as it stands is numbers fudged against us. I can share more thoughts in a discord thread as this seems like a good discussion point. It's probably also best to do this in a separate PR.
> 
> Good call, can you please remove this from this PR? Even if we all agreed this is what we want I think it would be very good for it to have its own commit, and documentation on why we want to do this (if we do). A discord thread would also be good to figure out where the team stands on this.

Yup totally aligned with that

shahzadlone:
Introducing a retry mechanic to get by for now, seems like codcov team is aware of this bug that happens without a token. Until they push it we can rely on this retry mechanic.

In future should probably rework using a 2 step workflow where first work is not privileged and generates the report, then the report is passed to a privileged `workflow run` action that has the secrets to upload using the token.

.github/workflows/code-test-coverage.yml
```diff
@@ -37,9 +42,29 @@ jobs:
       - name: Generate full test coverage report using go-acc
         run: make test:coverage
 
-      - name: Upload coverage to Codecov
+      - name: Upload coverage to Codecov without token, retry on failure
+        env: 
+          codecov_secret: ${{ secrets.CODECOV_TOKEN }}
+        if: env.codecov_secret == ''
+        uses: Wandalen/wretry.action@v1.0.36
```
AndrewSisley, line 49: todo: Can we please at least use a more popular dependency *if* we are going to introduce one?  The one currently on this line only has 35 stars, whereas this one has 245 and was built at/for a well known company whereas `wandalen`s appears to be purely a personal project.

AndrewSisley (dismissed):

.github/workflows/code-test-coverage.yml
shahzadlone: Which one are you recommending? I just saw many people use this one to fix the code cov flakey stuff. Happy to look into another one that might be more reputable.

AndrewSisley: lol, I forgot to link it. I had it open and ready for you sorry. This is the one I meant to suggest: https://github.com/marketplace/actions/retry-step

shahzadlone: It seems to be intended to replace the run step commands not action uses. LMK if you see an example for retries upon an action use.

shahzadlone: More info:
https://stackoverflow.com/a/71583518

AndrewSisley: You should be able to run the code cov action as a command no?

I'm just very uncomfortable introducing such an unused personal project into our CI system, and I think if I had to choose between using that action, or having a flaky code cov action, I would probably go for the flaky code cov action.

shahzadlone: I would rather use the official codecov github action than the shell command as it takes into account some built-in github metadata when it uploads the data to codcov.

I disagree that it is `an unused personal project` it has thousands of uses already: https://github.com/search?q=Wandalen%2Fwretry.action+path%3A.github%2Fworkflows%2F+language%3AYAML&type=Code&s=indexed&o=desc

Some of the users are like shopify, open-metadata, Isar Database, Airbyte, Immich

We shouldn't not solve the problem just because we have to use a 3rd party action. We already use a lot of 3rd party actions, even when we weren't open-sourced.

shahzadlone: Also keep in mind, this will mostly be a fallback case when a token isn't available (ASFAIK it will always be there). So majority of times this won't even be hit or used once the rework of this action is done.

fredcarle: I think it's pretty safe to use the one Shahzad has at the moment. Specially that it's set on a version so future changes won't affect us unless we update without checking what the changes are.

fredcarle (approved):
LGTM!

Issue #953, opened 2022-11-25
Convert the explain tests to use the new explain setup


shahzadlone:
# Description

Currently we have 3 types of testing frameworks:
1) Old integration testing setup
2) New explain setup, this was the setup we needed to convert all explain tests to from (1). (#949)
3) New integration testing setup based on actions. (#1089)
4) After (3) was introduced, it is much more flexible, so the goal is to integrate (2) into (3) so we can use it for explain tests as well. Inorder to do (4) we need to do this ticket (#953) first (a) from below. The issue ticket for (4) is #1243.

## Situation
We are currently in the middle of the migrations between:
a. (1) -> (2) ... **This is what this ticket is for** (#953)
b. (1) -> (3)
c. After (4) is done do: (2) -> (4) 

# Default Explain Tests to convert:
### NOTE: Must be done as part of (a) before we can do (4).
- [x] basic_test.go
- [x] invalid_type_arg_test.go
- [x] dagscan_test.go
- [x] create_test.go
- [x] update_test.go
- [x] delete_test.go
- [x] with_filter_key_test.go
- [x] with_filter_test.go
- [x] with_limit_test.go
- [x] with_order_test.go
- [x] top_with_average_test.go
- [x] top_with_count_test.go
- [x] top_with_sum_test.go
- [x] type_join_test.go
- [x] with_average_test.go
- [x] with_count_test.go
- [x] with_sum_test.go
- [x] group_test.go
- [x] group_with_average_test.go
- [x] group_with_dockey_test.go
- [x] group_with_filter_test.go
- [x] group_with_limit_test.go
- [x] group_with_order_test.go

# Simple Explain tests to convert:
- [x] ### NOTE: Must be done as part of (a) before we can do (4).

# Explain Explain tests to convert:

PR #1416, opened 2023-04-27
test: Convert mutation explain tests to new explain setup

### LINKED ISSUE #1379: Convert basic default explain tests including filter tests.
Convert basic default explain tests including filter tests.

More info in parent ticket: https://github.com/sourcenetwork/defradb/issues/953

### LINKED ISSUE #1417: Convert default mutation explain tests
Convert default mutation explain tests.

More info in parent ticket: https://github.com/sourcenetwork/defradb/issues/953

shahzadlone:
## Relevant issue(s)
- Part of #953
- Resolves #1417

## Description
Continue converting explain tests to the new explain setup before we can integrate the entire setup to the new action based setup. #953 Has a lot more detail on the entire plan.

This PR converts all the default mutation explain tests to the new explain setup.

- Splits delete tests into two files one with error and without.
- Splits multi tests from single function into separate test functions (easier to maintain and debug). 
- Removes some tests that provide no value wrt to non-execution explain tests.


tests/integration/explain/default/create_test.go
```diff
@@ -80,32 +79,19 @@ func TestExplainMutationCreateSimpleDoesNotCreateDocGivenDuplicate(t *testing.T)
 			},
 		},
 
-		Results: []dataMap{
+		ExpectedTargets: []explainUtils.PlanNodeTargetCase{
```
fredcarle, line 82: question: It seams as if a lot of the changes to the tests, including this one, are testing less of the output. Am I missing something or is this the intent?

shahzadlone: It's now testing the pattern (ordering of the nodes), and only the content of the nodes that we care about for each test separately. For this specific test for example we only care about the `createNode`'s attributes (point 2 from the below link).

More info can be found here:
https://github.com/sourcenetwork/defradb/pull/949#issuecomment-1327719247

AndrewSisley (approved):
LGTM

tests/integration/explain/default/create_test.go
fredcarle: Thanks for referencing that comment! It makes much more sense to me now :)

fredcarle (approved):
LGTM :)

PR #1432, opened 2023-05-01
chore: Parallelize unit tests

### LINKED ISSUE #1431: Unit tests are not parallelized
Unit tests are not parallelized, leading to slower test execution times.

orpheuslummis:
## Relevant issue(s)

Partially resolves #1431

## Description

Increase parallelization of unit tests

Skipped packages because they had issues when parallelized (e.g. data race)
- config
- events
- api/http
- node

Note that "nearly all" of the test execution time is not in the unit tests, but rather in the integration test suite.

## Tasks

## How has this been tested?

CI

### DIFF
--- a/client/errors_test.go
+++ b/client/errors_test.go
@@ -17,7 +17,6 @@ import (
 )
 
 func TestNewUnexpectedType(t *testing.T) {
-	t.Parallel()
 	someString := "defradb"
 	someLocation := "foo"
 	err := NewErrUnexpectedType[int](someLocation, someString)
--- a/client/errors_test.go
+++ b/client/errors_test.go
@@ -17,6 +17,7 @@ import (
 )
 
 func TestNewUnexpectedType(t *testing.T) {
+	t.Parallel()
 	someString := "defradb"
 	someLocation := "foo"
 	err := NewErrUnexpectedType[int](someLocation, someString)
--- a/merkle/crdt/merklecrdt_test.go
+++ b/merkle/crdt/merklecrdt_test.go
@@ -40,6 +40,7 @@ func newTestBaseMerkleCRDT() (*baseMerkleCRDT, datastore.DSReaderWriter) {
 }
 
 func TestMerkleCRDTPublish(t *testing.T) {
+	t.Parallel()
 	ctx := context.Background()
 	bCRDT, store := newTestBaseMerkleCRDT()
 	delta := &corecrdt.LWWRegDelta{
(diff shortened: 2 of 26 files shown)


client/errors_test.go
```diff
@@ -17,6 +17,7 @@ import (
 )
 
 func TestNewUnexpectedType(t *testing.T) {
+	t.Parallel()
```
AndrewSisley, line 20: nitpick: This line very much appears to be useless, as it is the only such call within the package

merkle/crdt/merklecrdt_test.go
```diff
@@ -40,6 +40,7 @@ func newTestBaseMerkleCRDT() (*baseMerkleCRDT, datastore.DSReaderWriter) {
 }
 
 func TestMerkleCRDTPublish(t *testing.T) {
+	t.Parallel()
```
AndrewSisley, line 43: nitpick: This line very much appears to be useless, as it is the only such call within the package

AndrewSisley (approved):
Approving because it has a tangible positive impact.  

I really really wish that this didnt require a call per test function though, it is really ugly IMO and I am annoyed at the Go dev team for not coming up with something nicer.

Might be worth waiting on another approval before merge, as this does IMO have a negative impact on the maintainability of the code (is ugly, and requires conscious thought, especially if a test stops working because of it (do you remove parallel, or change the prod code)).

client/errors_test.go
orpheuslummis: what about leaving it as a default norm to for the tests to be defined?

AndrewSisley: I really dont think we should suggest or enforce that, it is essentially dead, untested, code.  It suggests that the test runs in parallel, but it does not, and in many cases the ability to run that test in parallel has/will not be tested. The introduction of a second Parallel call may cause the first/original test to fail.

orpheuslummis: resolved

merkle/crdt/merklecrdt_test.go
orpheuslummis: resolved

client/errors_test.go
orpheuslummis: *trivia*: in Rust tests run in parallel by default, but can be made sequential with a synchronization mechanism or `#[serial]` attribute from crate https://docs.rs/serial_test/

AndrewSisley: Yes, I'm a bigger fan of this way round.  Opting out of running them in parallel is more like accepting (and warning/flagging) that the production code is broken and needs fixing.

jsimnz:
Considering the majority of our test time is in the integration, what is the decrease in test time with this change?

orpheuslummis:
> Considering the majority of our test time is in the integration, what is the decrease in test time with this change?

probably quite negligible 

most usefulness comes from identifying potential brittle areas in the system

PR #1434, opened 2023-05-01
ci: Add a separate workflow to run the linter

### LINKED ISSUE #1435: Separate linter so we can disable benchmark flow until it's fixed
Previously linter was part of the benchmark flow...
But benchmarks are flakey...
So until https://github.com/sourcenetwork/defradb/issues/1062 gets resolved
We should turn off benchmark workflow and add a lint only workflow.

shahzadlone:
## Relevant issue(s)
Resolves #1435

## Description
Add a Lint only workflow.

## How has this been tested?
CI on branch of a fork and this PR

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,70 @@
+# Copyright 2023 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+name: Lint Workflow
+
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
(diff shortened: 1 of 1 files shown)


AndrewSisley:
question: Why does this not remove the lint stuff from the lint-then-benchmark workflow?

shahzadlone:
> question: Why does this not remove the lint stuff from the lint-then-benchmark workflow?

That entire workflow file will be disabled after this is merged. I don't want to make changes to that workflow as it will be dead until we fix it (would rather preserve it in the state it used to work in rather than in a changed further broken state).

AndrewSisley:
> > question: Why does this not remove the lint stuff from the lint-then-benchmark workflow?
> 
> That entire workflow file will be disabled after this is merged. I don't want to make changes to that workflow as it will be dead until we fix it (would rather preserve it in the state it used to work rather than in a changed further broken state).

Those are good reasons, cheers for the explanation :)

AndrewSisley (approved):
LGTM :)

PR #1444, opened 2023-05-03
PR: Configure go partials as hits.

### LINKED ISSUE #1443: Code coverage behaving negatively against us
Code coverage IMO is acting negatively against us at the moment.

### Consider:
```go
+:1:  next, err := n.plan.Next()
*:2:  if err != nil {
-:3:      return false, err
-:4:  }
```

### Where:
`+` = covered: line hit.
`*` = partial: **line hit**, but didn't enter it's sub-block.
`-` = not-covered: line was missed (not-hit).

### Coverage Situation:
(1) Code coverage is currently calculated by dividing total `+` lines by total lines (which includes all `-`,`+`,`*`).
From our example above this means code coverage would be 1/4 = `25%`.
(2) From a purely execution perspective, `line 1` and `line 2` are both executed and hit. Which in theory would give a coverage of 2/4 = `50%`
(3) If we consider only lines that aren't covered that actually matter (i.e. not count lines with closing/opening brace). Only missed line in that case would be `line 3` and `line 4` not being counted in total number of lines. Coverage would then be 2/3 = `66.66%`.

It's even worse to stick to (1) if you have this for example:
```go
*:1:  if err := n.plan.NextErr(); err != nil {
-:2:      return false, err
-:3:  }
```
(1) - 0% coverage
(2) - 33.33% coverage
(3) - 50% coverage

### Problem:
We currently have (1), which really works negatively for us specially coding in GO, as these error check blocks can really build up, and for each block we have an extra 2 lines that work against us, one the partial hit line, and one the closing brace. We have discussed this previously here: 
https://discord.com/channels/427944769851752448/887814166331662396/966693461430513725

### Solution:
Include partials as hits (there is an option with codecov now). This would mean that at bare minimum it will help us reach (2) by offsetting against the extra brace.

Our coverage goes up approximately by `+5.66%` to a total of `77.87%`.

shahzadlone:
## Relevant issue(s)
Resolves #1443 

## Description
Partials considered as hits, more info in the related issue.

## How has this been tested?
CodeCov CI Report


shahzadlone:
https://github.com/sourcenetwork/defradb/issues/1443#issuecomment-1568338851

PR #1450, opened 2023-05-04
feat: Non-unique secondary index

### LINKED ISSUE #297: Sec. Indexes: Implement non-unique indexes (querying excluded)

Use posting list

islamaliev:
## Relevant issue(s)

Resolves #297 

## Description

This PR introduces first functionality for secondary indexes which includes:
- specifying index for a collection as part of the collection schema and parsing it
- mocks for interfaces the index functionality depends on
- db methods to create, delete and retrieve indexes
- storing index data in the system storage
- storing indexed fields' values to data storage
- change to the testing framework to allow testing of indexes

This PR doesn't include using indexes for fetching data for queries to keep the PR smaller.

## Tasks

## How has this been tested?

Unit tests

- MacOS


client/collection.go
```diff
@@ -136,6 +136,18 @@ type Collection interface {
...
 	GetAllDocKeys(ctx context.Context) (<-chan DocKeysResult, error)
+
+	// CreateIndex creates a new index on the collection.
+	CreateIndex(ctx context.Context, desc IndexDescription) (IndexDescription, error)
+
+	// DropIndex drops an index from the collection.
+	DropIndex(ctx context.Context, indexName string) error
+
+	// DropAllIndexes drops all indexes from the collection.
+	DropAllIndexes(ctx context.Context) error
```
AndrewSisley, line 147: thought: This looks like a fun function, I would suggest we do not offer this level of destruction to our users, I think dropping of indexes should be deliberate.

core/key.go
```diff
@@ -210,6 +218,22 @@ func NewCollectionSchemaVersionKey(schemaVersionId string) CollectionSchemaVersi
...
 }
 
+func NewCollectionIndexKey(colID, name string) CollectionIndexKey {
+	return CollectionIndexKey{CollectionID: colID, IndexName: name}
+}
+
+func NewCollectionIndexKeyFromString(key string) (CollectionIndexKey, error) {
+	keyArr := strings.Split(key, "/")
+	if len(keyArr) < 4 || len(keyArr) > 5 || keyArr[1] != "collection" || keyArr[2] != "index" {
+		return CollectionIndexKey{}, errors.WithStack(ErrInvalidKey, errors.NewKV("Key", key))
```
AndrewSisley, line 228: todo: Calling `WithStack` should not be required here.  All Defra errors already do this, the function is there to add stack stuff to non-defra errors :)

core/key.go
```diff
@@ -106,6 +107,13 @@ type CollectionSchemaVersionKey struct {
 
 var _ Key = (*CollectionSchemaVersionKey)(nil)
 
+type CollectionIndexKey struct {
+	CollectionID string
+	IndexName    string
```
AndrewSisley, line 112: suggestion: I think we should use the Ids here, and those are uint32s not strings.

client/collection.go
islamaliev: I though this would be a convenient method for collection deletion so that we don't need to go through additional hassle of retrieving the list of current indexes and delete them one by one.

fredcarle: That's a good idea although we could probably achieve this using a method on collection without it being part of the interface.

core/key.go
islamaliev: just copied from another place. Will change

client/collection.go
islamaliev: made it private

core/key.go
islamaliev: kinda followed the lead of `CollectionKey`:
```go
type CollectionKey struct {
	CollectionName string
}
```

AndrewSisley: I think that is the only place we persist collection name like this - it is an index to the collection CID using name to easily identify clashes.  All the other persisted keys use the collectionID I think.

core/key.go
```diff
@@ -67,6 +68,15 @@ type DataStoreKey struct {
 
 var _ Key = (*DataStoreKey)(nil)
 
+// IndexDataStoreKey is a type that represents a key of an indexed document in the database.
+type IndexDataStoreKey struct {
+	CollectionID string
```
AndrewSisley, line 73: todo: It is non-obvious to me as to which collection id this is.  Can you please document whether it is the integer (as a string), the cid, or the name?

suggestion: `IndexID` is less ambiguous, but I think it is worth being really explicit and add some documentation for the prop.

core/key.go
```diff
@@ -318,6 +351,101 @@ func (k DataStoreKey) ToPrimaryDataStoreKey() PrimaryDataStoreKey {
...
+// splitting the input using '/' as a field deliminator.  It assumes
+// that the input string is in the following format:
+//
+// /[CollectionID]/[IndexID]/[FieldID](/[FieldID]...)
+//
+// Any properties before the above (assuming a '/' deliminator) are ignored
+func NewIndexDataStoreKey(key string) (IndexDataStoreKey, error) {
+	indexKey := IndexDataStoreKey{}
+	if key == "" {
+		return indexKey, ErrEmptyKey
```
AndrewSisley, line 364: nitpick: I don't like returning `indexKey` here and on similar error lines below, its technical-correctness is somewhat dependent on other lines (including future modifications) not mutating the `indexKey` var, which is actually done a bunch of times later in the func.  

Technical-correctness aside, I see it as conceptually incorrect, as we dont actually want to return `indexKey`, it is just that it happens to currently have the same (empty) value that we do want to return.

Doing this will also allow you to move the `indexKey` declaration lower into the function, closer to where it is actually used, and reducing the scope of its existence.

core/key.go
```diff
@@ -401,6 +529,27 @@ func (k CollectionSchemaVersionKey) ToDS() ds.Key {
 	return ds.NewKey(k.ToString())
 }
 
+func (k CollectionIndexKey) ToString() string {
+	result := COLLECTION_INDEX
+
+	if k.CollectionName != "" {
+		result = result + "/" + k.CollectionName
+		if k.IndexName != "" {
```
AndrewSisley, line 537: todo: It is non-obvious as to why we only want to append `IndexName` if `CollectionName` is not empty - could you please either change this so that `IndexName` is not dependent on `CollectionName`, or document why it is?

AndrewSisley (changes requested):
Submitting a partial review for now, as I need a breather from it - it looks good in general, but I have have few low-scope todos so far

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+func NewCollectionIndex(
+	collection client.Collection,
+	desc client.IndexDescription,
+) CollectionIndex {
+	index := &collectionSimpleIndex{collection: collection, desc: desc}
+	schema := collection.Description().Schema
+	fieldID := schema.GetFieldKey(desc.Fields[0].Name)
+	field := schema.Fields[fieldID]
```
AndrewSisley, line 102: todo: FieldID is not necessarily the slice index. Please do not rely on this. There is a function on collection description `GetFieldByID` that you should use.  John might be moving this onto schema in his fetcher PR.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+type collectionSimpleIndex struct {
+	collection  client.Collection
+	desc        client.IndexDescription
+	convertFunc func(any) ([]byte, error)
+}
+
+var _ CollectionIndex = (*collectionSimpleIndex)(nil)
+
+func (i *collectionSimpleIndex) getDocKey(doc *client.Document) (core.IndexDataStoreKey, error) {
+	indexedFieldName := i.desc.Fields[0].Name
```
AndrewSisley, line 116: todo: Please explain with a code-comment as to why `Fields[0]` is okay :)

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	prefixKey.IndexID = strconv.Itoa(int(i.desc.ID))
+	q, err := txn.Datastore().Query(ctx, query.Query{
+		Prefix: prefixKey.ToString(),
+	})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := q.Close(); err != nil {
+			log.ErrorE(ctx, "Failed to close collection query", err)
```
AndrewSisley, line 188: todo: Similar to earlier comment, I cant see why this is logged and discarded, please return it, or at least document why with a code-comment.

Also for `deserializePrefix`

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+			desc.Fields[i].Direction = client.Ascending
+		}
+	}
+	return nil
+}
+
+func generateIndexName(col client.Collection, fields []client.IndexedFieldDescription, inc int) string {
+	sb := strings.Builder{}
+	direction := "ASC"
+	//if fields[0].Direction == client.Descending {
```
AndrewSisley, line 237: todo: Please remove the commented out code

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+		if desc.Fields[i].Direction == "" {
+			desc.Fields[i].Direction = client.Ascending
+		}
+	}
+	return nil
+}
+
+func generateIndexName(col client.Collection, fields []client.IndexedFieldDescription, inc int) string {
+	sb := strings.Builder{}
+	direction := "ASC"
```
AndrewSisley, line 236: todo: It is not clear from the PR description, or the public functions that only `ASC` is implemented. Can you please document this :)

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	//if fields[0].Direction == client.Descending {
+	//direction = "DESC"
+	//}
+	sb.WriteString(col.Name())
+	sb.WriteByte('_')
+	sb.WriteString(fields[0].Name)
```
AndrewSisley, line 242: todo: Please explain with a code comment as to why `fields[0]` is okay

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+		if err := q.Close(); err != nil {
+			log.ErrorE(ctx, "Failed to close collection query", err)
+		}
+	}()
+
+	for res := range q.Next() {
+		if res.Error != nil {
+			return res.Error
+		}
+		err = txn.Systemstore().Delete(ctx, ds.NewKey(res.Key))
```
AndrewSisley, line 317: question: How sure are you that this works?  It is unusual for mutation of an enumerable that you are currently iterating through to be safe/correct.

request/graphql/schema/collection.go
```diff
@@ -159,15 +132,211 @@ func fromAstDefinition(
...
 		Name: def.Name.Value,
 		Schema: client.SchemaDescription{
 			Name:   def.Name.Value,
 			Fields: fieldDescriptions,
 		},
+		Indexes: indexDescriptions,
 	}, nil
 }
 
+func isValidIndexName(name string) bool {
```
AndrewSisley, line 155: todo: I think this is worth documenting, both for us internal developers, and for external users of the client package.

AndrewSisley (changes requested):
Taking another break - added some more comments :)

db/index.go
fredcarle, line 1: todo: All exported types should have documentation.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+func validateIndexDescription(desc client.IndexDescription) error {
+	if desc.ID != 0 {
+		return NewErrNonZeroIndexIDProvided(desc.ID)
+	}
+	if len(desc.Fields) == 0 {
+		return ErrIndexMissingFields
+	}
+	if len(desc.Fields) == 1 && desc.Fields[0].Direction == client.Descending {
```
fredcarle, line 220: question: We can't have a descending order if we only have one field for the index?

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+	return nil
+}
+
+func generateIndexName(col client.Collection, fields []client.IndexedFieldDescription, inc int) string {
+	sb := strings.Builder{}
+	direction := "ASC"
+	//if fields[0].Direction == client.Descending {
+	//direction = "DESC"
+	//}
```
fredcarle, line 239: todo: Remove commented code.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+		err = json.Unmarshal(res.Value, &element)
+		if err != nil {
+			return nil, NewErrInvalidStoredIndex(err)
+		}
+		elements = append(elements, element)
+	}
+	return elements, nil
+}
+
+func (c *collection) getIndexes(ctx context.Context, txn datastore.Txn) ([]CollectionIndex, error) {
```
fredcarle, line 352: thought: It's a bit odd that this function name suggests that the indexes will be returned (as it does) but that it also assigns the indexes to a collection struct field. At a mininmum this should be documented on the function since it's weird to see a dropped return value when calling this function.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+			}
+		}
+		if !found {
+			return NewErrNonExistingFieldForIndex(field.Name)
+		}
+	}
+	return nil
+}
+
+func (c *collection) processIndexName(
```
fredcarle, line 460: suggestion: Can you please rename of document this function as the name doesn't make it clear as to what is happening bellow.

request/graphql/parser.go
```diff
@@ -103,7 +103,10 @@ func (p *parser) Parse(ast *ast.Document) (*request.Request, []error) {
 	return query, nil
 }
 
-func (p *parser) ParseSDL(ctx context.Context, schemaString string) ([]client.CollectionDescription, error) {
+func (p *parser) ParseSDL(ctx context.Context, schemaString string) (
+	[]client.CollectionDescription,
+	error,
+) {
```
fredcarle, line 109: question: Why this change?

db/collection.go
```diff
@@ -56,6 +57,10 @@ type collection struct {
 	schemaID string
 
 	desc client.CollectionDescription
+
+	isIndexCached  bool
```
AndrewSisley, line 61: todo: This bool represents a race condition, where it can be set to true, and then meanwhile, whilst collection is in use, an index is mutated/added/etc.

I am not sure we can be sure that this will only ever cause performance issues, especially regarding unique indexes and backfill.

I would suggest a rethink here.  Please bear in mind that the collection may currently outlive any given transaction that may be in scope at any given point in time (I dislike this anyway, but it is how the current system behaves, and we either need to change that, or handle it for stuff like this).

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	indexKey, err := c.processIndexName(ctx, &desc)
+	if err != nil {
+		return nil, err
+	}
+
+	txn, err := c.getTxn(ctx, false)
```
AndrewSisley, line 411: todo: It does appear that you are failing to commit implicit transactions here, meaning that unless users explicitly provide a transaction to the collection instance, changes made in this function will be discarded.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	desc client.IndexDescription,
+) (client.IndexDescription, error) {
+	index, err := c.createIndex(ctx, desc)
+	if err != nil {
+		return client.IndexDescription{}, err
+	}
+	if c.isIndexCached {
+		c.indexes = append(c.indexes, index)
+	}
+	return index.Description(), nil
```
AndrewSisley, line 263: todo: I think it is worth documenting that the returned description is scoped to the collection transaction if one is provided, and may not be available elsewhere until the collection transaction is committed.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+func (c *collection) DropIndex(ctx context.Context, indexName string) error {
+	key := core.NewCollectionIndexKey(c.Name(), indexName)
+
+	txn, err := c.getTxn(ctx, false)
```
AndrewSisley, line 269: todo: This appears to have the same problem as `createIndex` - it doesn't look like you are commiting implicit transactions?

request/graphql/schema/collection.go
```diff
@@ -23,8 +23,19 @@ import (
 	"github.com/graphql-go/graphql/language/source"
 )
 
+const (
+	indexDirectiveLabel          = "index"
```
AndrewSisley, line 27: todo: I cannot spot where the index directive is declared, and made available to the GQL schema - are you sure this works via a gql client?

todo: Related to the above, I think it is well worth adding schema tests to make sure this works, and continues to do so.

todo: I think this is the wrong place to declare these consts, I would move them to `request/graphql/schema/types.types.go`, alongside the explain directives (a good example of how to declare and make the directive available to the GQL type system).

request/graphql/schema/collection.go
```diff
@@ -159,15 +132,211 @@ func fromAstDefinition(
...
+	if directions != nil {
+		if len(directions.Values) != len(desc.Fields) {
+			return client.IndexDescription{}, ErrIndexWithInvalidArg
+		}
+		for i := range desc.Fields {
+			dirVal, ok := directions.Values[i].(*ast.EnumValue)
+			if !ok {
+				return client.IndexDescription{}, ErrIndexWithInvalidArg
+			}
+			if dirVal.Value == "ASC" {
```
AndrewSisley, line 260: todo: `"ASC"` and `"DESC"` should really be using constants declared in the `client/requests` package I think, instead of duplicating the underlying strings.  It makes it much easier to identify the related components if nothing else.

AndrewSisley (changes requested):
Made it all the way through - looks good, but I think you really need to add some integration tests, and there are a few localised issues that need addressing.

db/fetcher/encoded_doc.go
```diff
@@ -20,6 +20,18 @@ import (
 	"github.com/sourcenetwork/defradb/core"
 )
 
+type EncodedDocument interface {
```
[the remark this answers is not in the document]
islamaliev: yes

client/index.go
```diff
@@ -0,0 +1,47 @@
...
+	// Fields contains the fields that are being indexed.
+	Fields []IndexedFieldDescription
+	// Unique indicates whether the index is unique.
+	Unique bool
+}
+
+// CollectionIndexDescription describes an index on a collection.
+// It's useful for retrieving a list of indexes without having to
+// retrieve the entire collection description.
+type CollectionIndexDescription struct {
```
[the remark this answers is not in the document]
jsimnz: Agreed

core/key.go
```diff
@@ -106,6 +116,13 @@ type CollectionSchemaVersionKey struct {
 
 var _ Key = (*CollectionSchemaVersionKey)(nil)
 
+type CollectionIndexKey struct {
```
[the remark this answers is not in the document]
jsimnz: I actually don't think it should be string names here as far as I know, thoughts @islamaliev ?

db/fetcher/encoded_doc.go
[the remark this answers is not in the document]
AndrewSisley: That's a pain - you sure there is no way to have this private?  I cant see why it would need to be public in Go.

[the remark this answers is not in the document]
AndrewSisley: I think we might be misunderstanding each other.

If the interface is private, and located where it currently is, there appears to be nothing stopping the mock from implementing it in another package (where and how it currently is).

Only the code that touches both the mock and the interface need access to both types. And this way we dont need to change the production code exposure for the sake of the tests.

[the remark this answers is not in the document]
AndrewSisley: Ahhh - I see the public `Fetcher` interface is previously returning a private type - that is quite bad, and reason enough to make the return type (interface or otherwise) public. Regardless of the desire to mock.

I'm happy with the current state :+1:

.github/codecov.yml
```diff
@@ -62,5 +62,6 @@ comment:
 
 ignore:
   - "tests"
+  - "**/mocks/*"
```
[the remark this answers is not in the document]
islamaliev: I addd a readme file in `tests` folder that explains it

db/fetcher/encoded_doc.go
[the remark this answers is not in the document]
AndrewSisley: https://github.com/sourcenetwork/defradb/issues/1552

client/collection.go
```diff
@@ -136,6 +136,15 @@ type Collection interface {
 
 	// GetAllDocKeys returns all the document keys that exist in the collection.
 	GetAllDocKeys(ctx context.Context) (<-chan DocKeysResult, error)
+
+	// CreateIndex creates a new index on the collection.
+	CreateIndex(ctx context.Context, desc IndexDescription) (IndexDescription, error)
```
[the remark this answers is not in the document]
islamaliev: tests are supposed to assert on the observable behaviour but without querying functionality it's hard to observe the impact of indexes beside from checking what's stored in the db, which is done already by unit tests. 
Creation and deletion of indexes will be part of integration tests for querying.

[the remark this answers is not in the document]
AndrewSisley: No, but what you can test is that CreateIndex results in stuff being returned from GetIndexes.  And you can also verfiy that creating indexes does not break queries on those object.

client/descriptions.go
```diff
@@ -29,6 +29,9 @@ type CollectionDescription struct {
 
 	// Schema contains the data type information that this Collection uses.
 	Schema SchemaDescription
+
+	// Indexes contains the indexes that this Collection has.
+	Indexes []IndexDescription
```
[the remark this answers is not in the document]
islamaliev: originally `[]IndexDescription` was separated from `CollectionDescription` but @jsimnz insisted to include it here.
What do you mean by "disable"?
I'm not even sure what would be the best way to test it because the observable behaviour is not quite observable.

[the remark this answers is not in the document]
AndrewSisley: The location here is correct IMO.  By disable, I mean make it return an error if such a modification is attempted - this is already done for some stuff, for example `TestSchemaUpdatesRemoveFieldErrors` and `TestSchemaUpdatesAddSimpleErrorsAddingSchema`.

You can use these tests as examples as to how to test it.

db/fetcher/encoded_doc.go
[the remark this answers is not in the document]
jsimnz: Trying to remember if there was a (bad) reason for this, or just accidentally.

I suspect the method returning the private type was itself private at some point, and either intentionally or accidentally exposed.

core/key.go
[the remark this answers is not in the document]
islamaliev: what is explained in SIP is related to keys for docs' indexed field values. And indeed there we use IDs:
```go
type IndexDatastoreKey struct {
	CollectionID string
	IndexID	     string
	FieldValues  []string
}
```
It's important there to keep the index as short as possible for better fetching performance.
The one in question in turn represent a key that points for an index description which is not supposed to be used as often and doesn't impose any requirements on traits like soft-ability.
On the other hand it will save us few fetches when, for example, a user requests to list all indexes. In this case we would need just to through the keys.
What do you think?

core/key.go
```diff
@@ -318,6 +351,101 @@ func (k DataStoreKey) ToPrimaryDataStoreKey() PrimaryDataStoreKey {
...
+
+func (k *IndexDataStoreKey) ToDS() ds.Key {
+	return ds.NewKey(k.ToString())
+}
+
+func (k *IndexDataStoreKey) ToString() string {
+	sb := strings.Builder{}
+
+	if k.CollectionID == "" {
+		return ""
```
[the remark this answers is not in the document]
AndrewSisley: If you are happy with the behaviour here I am (I might have a slight preference for `/1//3`, but not enough to override author preference).

I dont want to have to scan through unit tests to find out the reasons for this though, it looks potentially accidental and can cost time.  Can you please state the reasons for doing this as a code comment.

core/key.go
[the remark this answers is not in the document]
AndrewSisley: > On the other hand it will save us few fetches when, for example, a user requests to list all indexes. In this case we would need just to through the keys.

IMO that is far less importance performance-wise than whatever work we need to do on a normal fetch/query using an index.

> It's important there to keep the index as short as possible for better fetching performance.

This to me suggests that an int (or smaller) would be much better here, for both collectionID and indexID no? Particularly for `IndexDatastoreKey`.

core/key.go
AndrewSisley: > But then when asked to decompose from a string it does a lot of assumptions about the elements of the string

It does this because it is safe to do so, and if it doesnt follow the expected format something else is very broken.

> this comment is somewhat related to ...

Whatever gets decided there (I'm happy as-is) should also be documented here. It will likely save developer time in the future.

db/collection_index.go
```diff
@@ -0,0 +1,520 @@
...
+		c.indexes = append(c.indexes, index)
+	}
+	err = c.indexExistingDocs(ctx, txn, index)
+	if err != nil {
+		return client.IndexDescription{}, err
+	}
+	return index.Description(), nil
+}
+
+func (c *collection) newFetcher() fetcher.Fetcher {
```
AndrewSisley, line 224: question: I was just looking to pull this into my branch when I noticed - this feels like an odd location for such a function, particularly given that Lens was also looking to use it (created a lensed fetcher in `collection_index.go` feels extra odd).

At the moment it doesnt really do anything as far as I can tell so it is hard to really comment on it.  What is your plan for this function?

core/key.go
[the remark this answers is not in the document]
islamaliev: you are right, it's not the important. I guess I chose this approach because it was a tiny bit easier to implement.
But I still didn't hear any strong objections against `CollectionName` vs `CollectionID`.
Regarding `IndexDatastoreKey`, it's actually supposed to be given int, but converted to string. I saw other examples (other keys) that do it this way. Wasn't sure why myself.

db/collection_index.go
```diff
@@ -0,0 +1,183 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *collection) collectIndexedFields() []*client.FieldDescription {
+	fieldsMap := make(map[string]*client.FieldDescription)
+	for _, index := range c.indexes {
```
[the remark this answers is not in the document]
islamaliev: well, you don't need all occurrences of the field. Here we just want to collect all indexed fields to send them to the fetcher to limit it's scope.
I will add a comment

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/core"
+	"github.com/sourcenetwork/defradb/datastore"
+	"github.com/sourcenetwork/defradb/errors"
+)
+
+const (
+	indexFieldValuePrefix = "v"
+	indexFieldNilValue    = "n"
```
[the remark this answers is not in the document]
islamaliev: you mean `indexFieldValuePrefix` and `indexFieldNilValuePrefix`?
But one is a prefix and the other is not.

db/errors.go
```diff
@@ -52,36 +66,41 @@ var (
...
+	ErrDocumentAlreadyExists          = errors.New("a document with the given dockey already exists")
+	ErrDocumentDeleted                = errors.New("a document with the given dockey has been deleted")
+	ErrUnknownCRDTArgument            = errors.New("invalid CRDT arguments")
+	ErrUnknownCRDT                    = errors.New("unknown crdt")
+	ErrSchemaFirstFieldDocKey         = errors.New("collection schema first field must be a DocKey")
+	ErrCollectionAlreadyExists        = errors.New("collection already exists")
+	ErrCollectionNameEmpty            = errors.New("collection name can't be empty")
+	ErrSchemaIdEmpty                  = errors.New("schema ID can't be empty")
+	ErrSchemaVersionIdEmpty           = errors.New("schema version ID can't be empty")
+	ErrKeyEmpty                       = errors.New("key cannot be empty")
```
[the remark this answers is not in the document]
islamaliev: that's not my code (just formatted) but I can adjust it

db/index.go
islamaliev: that's correct. Different direction are only for composite indexes

core/key.go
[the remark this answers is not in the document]
AndrewSisley: > it's actually supposed to be given int, but converted to string. I saw other examples

Are you thinking about `FieldID`?  `FieldID` is like that as it is/can-be a string, as sometimes it is assigned `C` for composites. I cant think of any other props that do that off the top of my head.

db/collection_get.go
```diff
@@ -48,13 +48,19 @@ func (c *collection) get(
 	ctx context.Context,
 	txn datastore.Txn,
 	key core.PrimaryDataStoreKey,
+	fields []*client.FieldDescription,
```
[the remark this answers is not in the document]
islamaliev: `Fetcher.Init` method takes as one of the parameters `[]*client.FieldDescription`. In this file `get` method used to pass `nil` to it, which mean all properties AFAIK. Now I let the caller of `get` decide which properties it want to fetch.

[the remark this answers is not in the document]
AndrewSisley: Why not remove it if it is always the same value?

@jsimnz do you have a plan for this prop?

tests/integration/schema/updates/index/simple_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+package index
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestPatching_ForCollectionWithIndex_StillWorks(t *testing.T) {
```
AndrewSisley, line 19: praise: Nice test, I totally forgot about this when we were chatting earlier :)

tests/integration/index/create_test.go
```diff
@@ -0,0 +1,135 @@
...
+					query  {
+						Users {
+							_key
+							Name
+							Age
+						}
+					}`,
+				Results: []map[string]any{
+					{
+						"_key": "bae-52b9170d-b77a-5887-b877-cbdbb99b009f",
```
AndrewSisley, line 52: suggestion: I don't see much value added here by including the key in the query, and it will mean an extra location to change if/when the dockey changes.  Suggest removal of the `_key` field.

Same for `TestIndexCreate_ShouldNotHinderQuerying` and `TestIndexDrop_ShouldNotHinderQuerying`

tests/integration/index/drop_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+package index
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestIndexDrop_ShouldNotHinderQuerying(t *testing.T) {
```
AndrewSisley, line 19: todo: I think it well worth also adding a test for dropping an index that does not exist.

tests/integration/index/create_test.go
```diff
@@ -0,0 +1,135 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package index
```
AndrewSisley, line 11: todo: Sorry if I'm jumping the gun a bit and you are already working/planning on this, but I think we also need tests to assert that `GetIndexes` actually returns the indexes created.

tests/integration/index/drop_test.go
AndrewSisley, line 1: suggestion: `create_drop_test.go` might be a better name for this file, and then you could have another file `drop_test.go` that includes the test for dropping an index that doesnt exist. Quite minor though atm given that there are only a few tests within this package, but I think it would be slightly more consistent with other test file names (fairly debatable though, and quite heavily caveated)

tests/integration/index/create_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"testing"
+
+	"github.com/sourcenetwork/defradb/request/graphql/schema"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestIndexCreateWithCollection_ShouldNotHinderQuerying(t *testing.T) {
```
AndrewSisley, line 20: suggestion: I think someone on the team (maybe John) hates underscores in test names.  I like them, and use them in how I would otherwise name my tests, but I think you should check in with the rest of the team RE this on discord/standup in case anyone with strong feelings misses it at review time.

AndrewSisley (changes requested):
Gave the integration tests a look over, will go over the prod code again once you re-quest review.  Tests look great, thanks a bunch for adding them now, just left a couple of small comments.

request/graphql/schema/index_test.go
```diff
@@ -0,0 +1,303 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1:
```suggestion
// Copyright 2023 Democratized Data Foundation
```

db/index_test.go
```diff
@@ -0,0 +1,1265 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1:
```suggestion
// Copyright 2023 Democratized Data Foundation
```

db/indexed_docs_test.go
```diff
@@ -0,0 +1,1023 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1:
```suggestion
// Copyright 2023 Democratized Data Foundation
```

shahzadlone (dismissed):

core/key.go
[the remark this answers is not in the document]
AndrewSisley: Ah, I see `CollectionKey` is really just an index of the collection by collection name, that is it's only usage as far as I can see.

db/collection_index.go
```diff
@@ -0,0 +1,183 @@
...
+	prefix := core.NewCollectionIndexKey("", "")
+	q, err := txn.Systemstore().Query(ctx, query.Query{
+		Prefix: prefix.ToString(),
+	})
+	if err != nil {
+		return nil, NewErrFailedToCreateCollectionQuery(err)
+	}
+	defer func() {
+		if err := q.Close(); err != nil {
+			log.ErrorE(ctx, "Failed to close collection query", err)
```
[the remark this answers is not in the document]
AndrewSisley: Doesn't make it wonderful :)  If we can return the error, we should, and if we cant we need to document very clearly as to why not.  Copy-paste is not a valid reason IMO, `getAllCollections` could be a historical accidental that we do not with to propagate.

db/index.go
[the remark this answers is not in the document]
fredcarle: You're right. No need to change then.

tests/integration/index/create_test.go
fredcarle: I'm fine with it as long as there is only one and the use is to separate the function being tested from extra details.

AndrewSisley: Lol - why only one?  The test name format I am most familiar with uses two: `Action_Result_Input`, e.g. `Add_ReturnsFive_GivenThreeAndTwo`.  Used that all the way through my career prior to Source, and oddly enough, I think it was taken from a Google coding standards doc (I guess not one used by the Go team :)).

db/collection_get.go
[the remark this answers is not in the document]
islamaliev: it used to pass the same value, but now it's used for indexes to fetch a document for which an index needs to be updated. If the doc has bunch of props, this can play significant role (given the fetcher takes advantage of it)

db/collection_index.go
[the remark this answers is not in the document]
islamaliev: I disapprove copy&paste myself but this PR touches a lot of parts of the code. With some of them I wasn't quite familiar with and looked for guidance in form of existing code. Also I knew you would points our spot like in the code review ;)

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+
+type CollectionIndex interface {
+	Save(context.Context, datastore.Txn, *client.Document) error
+	Update(context.Context, datastore.Txn, *client.Document, *client.Document) error
+	RemoveAll(context.Context, datastore.Txn) error
+	Name() string
+	Description() client.IndexDescription
+}
+
+func getFieldValConverter(kind client.FieldKind) func(any) ([]byte, error) {
```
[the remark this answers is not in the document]
islamaliev: so far I didn't see it in the code. Please show me

db/collection_get.go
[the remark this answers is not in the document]
jsimnz: > Why not remove it if it is always the same value?
> @jsimnz do you have a plan for this prop?

Current dev branch doesn't do anything with the `[]FieldDescription`, regardless of the value, it will always return all fields for the collection. The pending Fetcher PR *does* make use of this field, and will respect the defined set of fields by only returning those. If `nil` is passed, it will return *all* fields. 

Since this will be merged pretty soon, its prob best to keep this here on this PR.

core/key.go
```diff
@@ -318,6 +351,101 @@ func (k DataStoreKey) ToPrimaryDataStoreKey() PrimaryDataStoreKey {
...
+	}
+	indexKey.IndexID = elements[1]
+
+	for i := 2; i < len(elements); i++ {
+		_, err = strconv.Atoi(elements[i])
+		if err != nil {
+			return IndexDataStoreKey{}, ErrInvalidKey
+		}
+		indexKey.FieldValues = append(indexKey.FieldValues, elements[i])
+	}
```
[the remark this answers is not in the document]
islamaliev: that's a good question. I don't know the answer myself :D 
I think it's a leftover since a played around with different types. This method `NewIndexDataStoreKey` is actually never used (only in test), that's why I've spotted any issues.

db/index.go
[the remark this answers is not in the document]
AndrewSisley: These a few, the inverse is done in encoded doc, and theres some very similar logic somewhere in collection on write, and some more in the gql and filter stuff I think.

db/collection_get.go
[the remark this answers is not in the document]
AndrewSisley: cheers

db/index.go
AndrewSisley: > But now thinking about this I realized that Badger is not the only db that can be used with defra. So it's better to separate those actions to be on the safe side.

That's a good point, happy with as is if you can find similar documentation on the ipfs interface declarations though.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+	indexDescriptions := make([]client.IndexDescription, 0, len(indexes))
+	for _, index := range indexes {
+		indexDescriptions = append(indexDescriptions, index.Description())
+	}
+
+	return indexDescriptions, nil
+}
+
+func (c *collection) createIndex(
```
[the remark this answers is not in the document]
fredcarle: I know you moved this to `collection_index.go` but the same comment applies. It would be nice to have `createIndex(...)` under the `CreateIndex(...)` function.

tests/integration/index/create_test.go
fredcarle: My point is that I don't want to start seeing stuff like `Test_Index_Create_With_Collection...`.

`Action_Input_Result` could be fine too.

core/key.go
```diff
@@ -318,6 +390,112 @@ func (k DataStoreKey) ToPrimaryDataStoreKey() PrimaryDataStoreKey {
...
+
+	indexKey := IndexDataStoreKey{CollectionID: uint32(colID)}
+
+	indID, err := strconv.Atoi(elements[1])
+	if err != nil {
+		return IndexDataStoreKey{}, ErrInvalidKey
+	}
+	indexKey.IndexID = uint32(indID)
+
+	for i := 2; i < len(elements); i++ {
```
shahzadlone, line 428: todo: A comment here would be nice to quickly highlight why we are starting from `i := 2`. Is easy to see why now, but if someone removes the above elements [0] and [1] or adds new ones to the logic, the comment will be helpful to insure they update the index.

db/index.go
shahzadlone: +1

db/collection_get.go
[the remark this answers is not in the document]
islamaliev: I didn't realise `fetcher` didn't use the `fields` because I didn't dig it the much. But it might change in the future and I'd say it should. Anyway I think that API for `Fetcher` needs to change first.

db/index.go
```diff
@@ -0,0 +1,495 @@
...
+
+type collectionSimpleIndex struct {
+	collection  client.Collection
+	desc        client.IndexDescription
+	convertFunc func(any) ([]byte, error)
+}
+
+var _ CollectionIndex = (*collectionSimpleIndex)(nil)
+
+func (i *collectionSimpleIndex) getDocKey(doc *client.Document) (core.IndexDataStoreKey, error) {
```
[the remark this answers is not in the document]
islamaliev: well, it doesn't imply dockey but some document's key which makes sense I would say.
But you are right, it's a bit confusing.

request/graphql/parser.go
islamaliev: I changed this method's signature at some point but then changed it back. The method was too long to fit in one line. I can revert this change

db/collection_index.go
islamaliev: It was just the simplest way i could impose a custom fetcher on a collection for testing purposes. It serves it's purpose but it's not very sophisticated. Didn't feel like doing more changes in this PR. I will probably create a task to address this.

AndrewSisley: Okay no worries - can you please move it to another location though?  Maybe collection.go?

datastore/mocks/DAGStore.go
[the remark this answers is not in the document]
islamaliev: yeah, it's generated

AndrewSisley (approved):
Really great work Islam, looks good to me :)

db/collection_index.go
```diff
@@ -0,0 +1,533 @@
...
+			return err
+		}
+		colIndexes = append(colIndexes, index)
+	}
+	c.desc.Indexes = indexDescriptions
+	c.indexes = colIndexes
+	return nil
+}
+
+func (c *collection) GetIndexes(ctx context.Context) ([]client.IndexDescription, error) {
```
fredcarle, line 393: todo: Please add documentation to exported methods :)

fredcarle (changes requested):
Just a few more todos before I approve. I noted a few methods that should have documentation as they are exported. I think there are a few others in your PR that are exported without documentation. I'm flagging them as this is good for internal devs when looking at the code but also because it adds useful information with intelisense in IDEs.

fredcarle (approved):
LGTM! Great work Islam!

PR #1456, opened 2023-05-05
test: Add tests to assert schema id order independence

### LINKED ISSUE #863: Use order independant CID generation for SchemaID
At the moment, I *think* its based on a JSON input to the multihash function, which is *order dependant*. We need to change so that it matches the dockey version CID generation, so that it is order *independant*. IE: Use CBOR.

```[tasklist]
### Prerequisites
- [ ] https://github.com/sourcenetwork/defradb/issues/851
- [ ] #852
```

AndrewSisley:
## Relevant issue(s)

Resolves #863

## Description

Adds tests to assert schema id order independence.  Ticket suggested the go json lib is order dependent, this appears to be untrue.


fredcarle (approved):
Thanks for confirming this Andy. I think if cbor is something we want regardless, we can come back to it in a future cycle.

PR #1460, opened 2023-05-05
fix: Rename commit field input arg to fieldId

### LINKED ISSUE #1452: commits input arg is call `field`, but the corresponding field is call `fieldId`
e.g.:
```
commits (field: "1") {
	fieldId
}
```
should be:
```
commits (fieldId: "1") {
	fieldId
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #1452 

## Description

Renames commit `field` input arg to `fieldId`, to match the field `fieldId`. Includes the explain output.

### DIFF
--- a/client/request/commit.go
+++ b/client/request/commit.go
@@ -19,10 +19,10 @@ var (
 type CommitSelect struct {
 	Field
 
-	DocKey    immutable.Option[string]
-	FieldName immutable.Option[string]
-	Cid       immutable.Option[string]
-	Depth     immutable.Option[uint64]
+	DocKey      immutable.Option[string]
+	FieldIDName immutable.Option[string]
+	Cid         immutable.Option[string]
+	Depth       immutable.Option[uint64]
 
 	Limit   immutable.Option[uint64]
 	Offset  immutable.Option[uint64]
--- a/client/request/commit.go
+++ b/client/request/commit.go
@@ -19,10 +19,10 @@ var (
 type CommitSelect struct {
 	Field
 
-	DocKey      immutable.Option[string]
-	FieldIDName immutable.Option[string]
-	Cid         immutable.Option[string]
-	Depth       immutable.Option[uint64]
+	DocKey  immutable.Option[string]
+	FieldID immutable.Option[string]
+	Cid     immutable.Option[string]
+	Depth   immutable.Option[uint64]
 
 	Limit   immutable.Option[uint64]
 	Offset  immutable.Option[uint64]
--- a/client/request/consts.go
+++ b/client/request/consts.go
@@ -20,6 +20,7 @@ const (
 	DocKey      = "dockey"
 	DocKeys     = "dockeys"
 	FieldName   = "field"
+	FieldIDName = "fieldId"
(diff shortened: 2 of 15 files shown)


client/request/consts.go
```diff
@@ -20,6 +20,7 @@ const (
 	DocKey      = "dockey"
 	DocKeys     = "dockeys"
 	FieldName   = "field"
+	FieldIDName = "fieldId"
```
AndrewSisley, line 23: Note: new property required, as some references to `FieldName` were actually for `field` in stuff like aggregates.

client/request/commit.go
```diff
@@ -19,10 +19,10 @@ var (
 type CommitSelect struct {
 	Field
 
-	DocKey    immutable.Option[string]
-	FieldName immutable.Option[string]
-	Cid       immutable.Option[string]
-	Depth     immutable.Option[uint64]
+	DocKey      immutable.Option[string]
+	FieldIDName immutable.Option[string]
```
fredcarle, line 23: thought: I'm not sure `FieldIDName` makes sense here. `FiedID` seemed more appropriate.

planner/mapper/commitSelect.go
```diff
@@ -23,7 +23,7 @@ type CommitSelect struct {
 	DocKey immutable.Option[string]
 
 	// The field for which commits have been requested.
-	FieldName immutable.Option[string]
+	FieldIDName immutable.Option[string]
```
fredcarle, line 26: thought: Same as above.

client/request/consts.go
fredcarle: There was already `FieldIDFieldName` bellow.

request/graphql/parser/commit.go
```diff
@@ -37,9 +37,9 @@ func parseCommitSelect(schema gql.Schema, parent *gql.Object, field *ast.Field)
 		} else if prop == request.Cid {
 			raw := argument.Value.(*ast.StringValue)
 			commit.Cid = immutable.Some(raw.Value)
-		} else if prop == request.FieldName {
+		} else if prop == request.FieldIDName {
```
fredcarle, line 40: thought(out-of-scope): I'm noticing that the `prop` is being compared to field names and we have the proper constants to make it a more visually descriptive conditional.

```Go
LinksFieldName           = "links"
HeightFieldName          = "height"
CidFieldName             = "cid"
DockeyFieldName          = "dockey"
CollectionIDFieldName    = "collectionID"
SchemaVersionIDFieldName = "schemaVersionId"
FieldNameFieldName       = "fieldName"
FieldIDFieldName         = "fieldId"
DeltaFieldName           = "delta"
```

fredcarle (changes requested):
I have a few thoughts I'd like to discuss before approving.

client/request/consts.go
AndrewSisley: `FieldIDFieldName` is the field name, not the input arg name.

fredcarle (approved):
Thanks for the change Andy!

PR #1461, opened 2023-05-06
refactor: Improve rollback on peer P2P collection error

### LINKED ISSUE #1389: Improve rollback of AddP2PCollection and RemoveP2PCollection
At the moment, if the transaction commit fails, the functions will return errors but the pubsub topics won't be rolled back, leaving the database and the topics in an incorrect state.

fredcarle:
## Relevant issue(s)

Resolves #1389 

## Description

This PR improves the rollback of P2P collection manipulation errors. It ensures that a rollback is applied if the transaction commit fails.

## Tasks

## How has this been tested?

Unit test coming with `net` package code coverage improvements.

### DIFF
--- a/net/peer.go
+++ b/net/peer.go
@@ -893,6 +893,7 @@ func (p *Peer) RemoveP2PCollections(collections []string) error {
 		if err != nil {
 			return p.rollbackRemovePubSubTopics(removedTopics, err)
 		}
+		removedTopics = append(removedTopics, col)
 	}
 
 	if err = txn.Commit(p.ctx); err != nil {
--- a/net/peer.go
+++ b/net/peer.go
@@ -763,6 +763,26 @@ type EvtPubSub struct {
 	Peer peer.ID
 }
 
+// rollbackAddPubSubTopics removes the given topics from the pubsub system.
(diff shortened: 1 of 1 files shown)


net/peer.go
```diff
@@ -873,24 +891,19 @@ func (p *Peer) RemoveP2PCollections(collections []string) error {
...
 
 	// If removing the collection topics succeeds, we add back the collections' documents
 	// to the pubsub topics.
-	for _, col := range collections {
-		c, err := store.GetCollectionBySchemaID(p.ctx, col)
-		if err != nil {
-			return err
-		}
-		keyChan, err := c.GetAllDocKeys(p.ctx)
+	for _, col := range storeCollections {
```
AndrewSisley, line 905: suggestion: This block can still fail - suggest adding logic for rollback of the final addPubSubTopic calls, and/or a big comment warning about this, particularly as it may fail after the txn has been committed.

AddP2PCollections has the same issue, but I dont think it would result in any visible end-result data differences (unlike this, where docs will no longer sync).

AndrewSisley (approved):
Looks better - cheer Fred, I think there are still a couple of holes here, but they are harder to plug and can perhaps be done later :)

PR #1462, opened 2023-05-06
refactor: Add prefix to pubsub topics

### LINKED ISSUE #1283: Peer.GetAllP2PCollections returns database state, not server state
Peer.GetAllP2PCollections returns database state, not server state.

I'm totally happy if this is the desired behaviour we want long term, but the functions around it `AddP2PCollections` and `RemoveP2PCollections` deal with server state as well as the persisted db state, and there is always going to be some risk that the two may differ (we will never be perfect coders).  

The function documentation also states that it `gets all the collectionIDs from the pubsup topics`, which to me (an internal dev) reads like it gets the state of the server.

If we want to keep this function's behaviour as-is, I'd suggest a tweaking of the func documentation, and perhaps a new function that returns the actual state of the server.  

Given that `db.GetAllP2PCollections` also exists (and is pretty much all the Peer func does), I would suggest that we change `Peer.GetAllP2PCollections` as the client/db function covers the behaviour that this function currently has and there seems little reason to duplicate it.

Note: Unsure whether to flag this as a bug, documentation or code quality, so I put it as bug for now.  Probably quite low priority right now.

fredcarle:
## Relevant issue(s)

Resolves #1283

## Description

This PR adds prefixes to pubsub topics which allows filtering by topic type. It enables returning the list of P2P collections based on the topics the server is registered to instead of those that are stored in the system store. 

Note 1: This PR builds on #1461.

Note 2: Maybe this PR should just be a draft as it's intended purpose is to generate a discussion around the approach.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/net/peer.go
+++ b/net/peer.go
@@ -893,6 +893,7 @@ func (p *Peer) RemoveP2PCollections(collections []string) error {
 		if err != nil {
 			return p.rollbackRemovePubSubTopics(removedTopics, err)
 		}
+		removedTopics = append(removedTopics, col)
 	}
 
 	if err = txn.Commit(p.ctx); err != nil {
--- a/net/peer.go
+++ b/net/peer.go
@@ -15,6 +15,7 @@ package net
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -685,10 +686,10 @@ func (p *Peer) handleDocUpdateLog(evt events.Update) error {
 func (p *Peer) pushLogToReplicators(ctx context.Context, lg events.Update) {
 	// push to each peer (replicator)
 	peers := make(map[string]struct{})
-	for _, peer := range p.ps.ListPeers(lg.DocKey) {
+	for _, peer := range p.ps.ListPeers(newTopic(dockeyPrefix, lg.DocKey)) {
 		peers[peer.String()] = struct{}{}
 	}
-	for _, peer := range p.ps.ListPeers(lg.SchemaID) {
+	for _, peer := range p.ps.ListPeers(newTopic(collectionPrefix, lg.SchemaID)) {
 		peers[peer.String()] = struct{}{}
 	}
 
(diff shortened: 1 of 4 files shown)


jsimnz:
Need to be explicit in the description if a PR is building on top of another open PR, so reviewers know how to most effectively review things :)

fredcarle:
> Need to be explicit in the description if a PR is building on top of another open PR, so reviewers know how to most effectively review things :)

100% I opened this and it was a little late for me 😅

net/server.go
```diff
@@ -499,6 +499,15 @@ func peerIDFromContext(ctx context.Context) (libpeer.ID, error) {
 	return pid, nil
 }
 
+const (
+	dockeyPrefix     = "d/"
+	collectionPrefix = "c/"
+)
+
+func newTopic(prefix, topic string) string {
```
AndrewSisley, line 507: thought: This func is a bit strange to me, it is basically an alias of the `+` operator, and I dont see much benefit in it.

Might be nicer to replace it with two functions, `newDockeyTopic` and `newCollectionTopic`.

fredcarle: I rearranged a little bit to improve the rollback. Let me know what you think.

net/peer.go
```diff
@@ -907,28 +927,59 @@ func (p *Peer) RemoveP2PCollections(collections []string) error {
...
+			ID:   colID,
+			Name: col.Name(),
+		})
+	}
+
+	return p2pCols, txn.Commit(p.ctx)
+}
+
+// GetAllP2PCollectionsFromServer gets all the collectionIDs that have been added to the
+// pubsub topics from the system store.
```
islamaliev, line 964: todo: update comment

2 methods `GetAllP2PCollectionsFromStore` and `GetAllP2PCollectionsFromServer` have the same comment

net/server.go
islamaliev: I don't know if it's go's convention or something, but it seems like usually function with prefix `new` should allocate/create a new object. Maybe `makeTopic`, `toTopic` or something similar would fit better.
On the other hand I support delegating the logical blocks to separate function (even if they just concatenate strings) because if we change the way topic are composed we will change it only in one place

islamaliev (commented):
I didn't see any tests. Is it because P2P aspect is hard to test?

fredcarle:
Dropping proposal.

PR #1483, opened 2023-05-10
refactor: [Proof of concept] Add strict ordering of serialized fieldIDs

jsimnz:
*note this is potentially necessary for #490, as I need to implement a `Seek` method. Its not 100% as I can order keys on the fetcher using the same (broken) ordering, and the seek would work correctly, but since I was already in this area, and the question came up recently regarding strict ordering of keys, I thought I would explore this*

Uses an alternative approach then what was previously discussed regarding binary encoding of fieldIDs when serializing data store keys.

Implements a "string based" strict ordering, instead of a "binary" based approach. By applying a padding of zeros, and limiting the stringified fieldID to 3 characters will gurantee ordering.

Basically, "001" < "002" < "011", where as before
"1" < "11" < "2".

This has the effect that we can only have 999 fields per document.

### DIFF
--- a/core/key.go
+++ b/core/key.go
@@ -48,6 +48,15 @@ const (
 	PRIMARY_KEY               = "/pk"
 	REPLICATOR                = "/replicator/id"
 	P2P_COLLECTION            = "/p2p/collection"
+
+	// fieldID string size to calculate padding
+	// ie: 	12   => 0012
+	//		3    => 0003
+	//		512  => 0512
+	fieldIDStringSize = 3
+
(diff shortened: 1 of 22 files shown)


core/key.go
```diff
@@ -48,6 +48,15 @@ const (
 	PRIMARY_KEY               = "/pk"
 	REPLICATOR                = "/replicator/id"
 	P2P_COLLECTION            = "/p2p/collection"
+
+	// fieldID string size to calculate padding
+	// ie: 	12   => 0012
+	//		3    => 0003
+	//		512  => 0512
```
islamaliev, line 55: question: does this indentation have a meaning or it's an accident?

PR #1488, opened 2023-05-10
test: Convert explain count tests to new explain setup

### LINKED ISSUE #1486: Convert default count explain tests to new explain setup
Convert default count explain tests to new explain setup.

More info in parent ticket: https://github.com/sourcenetwork/defradb/issues/953

shahzadlone:
## Relevant issue(s)
- Part of #953
- Resolves #1486 

## Description
Continue converting explain tests to the new explain setup before we can integrate the entire setup to the new action-based setup. #953 Has a lot more detail on the entire plan.

- This PR converts all the default explain count tests.
- Splits count join tests in a separate file.
- Adds some new tests that were missed before.

### DIFF
--- a/tests/integration/explain/default/with_count_test.go
+++ b/tests/integration/explain/default/with_count_test.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Democratized Data Foundation
+// Copyright 2023 Democratized Data Foundation
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt.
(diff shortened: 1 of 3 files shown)


tests/integration/explain/default/with_count_test.go
```diff
@@ -76,68 +76,76 @@ func TestExplainQueryOneToManyWithACount(t *testing.T) {
...
+				ExpectedAttributes: dataMap{
+					"joinType":    "typeJoinMany",
+					"rootName":    "author",
+					"subTypeName": "books",
+				},
+			},
+			{
+				TargetNodeName:    "scanNode", // inside of root
+				OccurancesToSkip:  0,
+				IncludeChildNodes: true, // should be leaf of it's branch, so will have no child nodes.
```
fredcarle, line 118: question: I saw this on the other PR I approved but now I'm starting to scratch my head when looking at the comment. Is it normal that the field says true to include child nodes but the comment says "will have no child node"?

fredcarle (approved):
LGTM. Just one question :)

tests/integration/explain/default/with_count_test.go
shahzadlone: `scanNode` is a special case as it should be the last node (leaf) in all these cases. The `IncludeChildNodes: true` is there to ensure that there are no nodes inside / under the `scanNode`. If some child nodes were ever introduced in the future, it would ensure to fail the test (as assumptions have changed) and should deliberately handle/change the test then based on the new change.

fredcarle: Ok thanks for the explanation. I do feel like the name is a little misleading though. Out of scope here of course but maybe something we can revisit later?

shahzadlone: `IncludeChildNodes` includes child nodes if they are available. I thought it covered the purpose well. Open to suggestions if you have any?

fredcarle: Oh! I get it now. It’s the comment on hat threw me off. I thought the comment was saying that we set it to true because there will be no child nodes. Not “although” it’s set to true, there will be no child nodes.

PR #1491, opened 2023-05-10
fix: Make nested joins work correctly from primary direction

### LINKED ISSUE #1490: Nested joins do not work correctly from primary directions
Broken out from https://github.com/sourcenetwork/defradb/issues/1469

AndrewSisley:
## Relevant issue(s)

Resolves #1490

## Description

Issue found by Digital Gaia. Nesting joins would result in the parent span being propagated into the child/subType scanNode, this caused the scan node to scan the wrong prefix and return no inner child records.


fredcarle (approved):
🎆🤘

Issue #935, opened 2022-11-12
Document serialization needs strong typing


jsimnz:
Broken out from https://github.com/sourcenetwork/defradb/pull/931/files#r1017152428

At the moment, there isn't enforced type validation/safety when creating documents, only on updating docs.

This causes various issues, but it came to light while working on #931 because DateTime fields are getting serialized as strings, and there's no entry point to conviently handle this atm.

AndrewSisley:
Suggest doing this after https://github.com/sourcenetwork/defradb/issues/1830

AndrewSisley:
Also blocked by https://github.com/sourcenetwork/defradb/issues/1847

PR #1500, opened 2023-05-12
refactor: Fetcher filter and field optimization

### LINKED ISSUE #490: Fetcher field selection and filter optimization

### LINKED ISSUE #1582: groupBy item in result set even when not requested
When we query with a `groupBy`, the `groupBy` item is returned in the result set even though it wasn't explicitly requested. This goes agains the graphql specs that states that the shape of the response must match that of the request. https://spec.graphql.org/October2021/#sec-Language.Fields, https://spec.graphql.org/October2021/#sec-Overview
```graphql
query {
  Users(groupBy: [Name]) {
    _group {
      Age
    }
  }
}
```
returns
```json
{
  "Name": "Alice",
  "_group": [
    {"Age": 19}
  ]
}
```
but should return
```json
{
  "_group": [
    {"Age": 19}
  ]
}
```

jsimnz:
## Relevant issue(s)

Resolves #490 
Resolves #1582 (indirectly)

## Description

This is a reduced version of #491. It takes a very different approach, and tries to keep as much of the existing Fetcher structure as possible.

Basically, this will try to eagerly ignore documents that don't pass the given filter at the fetcher level. This means we can apply various optimizations then if the filter was applied at the scanNode level like before. 

There will be two followup tasks in future PRs.
A) Proper lexigraphical encoding
B) Proper datastore iterator instead of query.

~~**There are still some outstanding bugs and failing tests, but I wanted to get the reviews coming in on the core implementation while I continue to squash bugs to parallelize the time spent**~~

~~Also..ignore some of the lingering Print statements for debugging, they will be removed before merging.~~

## Tasks

- [ ] Remove debug print statements.

## How has this been tested?

Manually, unit + integration tests, CI

- Ubuntu (WSL2)


planner/scan.go
```diff
@@ -103,23 +115,32 @@ func (n *scanNode) Next() (bool, error) {
...
+
+	// Keeping for refence as there is an outstanding
+	// question.
+	//
+	// Don't need to run filter here anymore since the
+	// fetcher will run the filter for us
+	//
+	// Question: How do we want to handle `info.filterMatches`
+	// now? @shahzadlone
+	/*
```
jsimnz, line 143: @shahzadlone

db/fetcher/fetcher.go
```diff
@@ -48,14 +59,21 @@ type DocumentFetcher struct {
 	order        []dsq.Order
 	curSpanIndex int
 
-	schemaFields map[uint32]client.FieldDescription
-	fields       []*client.FieldDescription
+	filter  *mapper.Filter
+	mapping *core.DocumentMapping
```
AndrewSisley, line 63: todo: This does not appear to be used and should be removed

db/fetcher/encoded_doc.go
```diff
@@ -178,14 +183,32 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
 
 // @todo: Implement Encoded Document type
 type encodedDocument struct {
+	mapping *core.DocumentMapping
```
AndrewSisley, line 186: suggestion:  This is one change that I am really not a fan of.  I was under the impression that fetcher was something that 'transcended' the planner system, yet here we are making a planner/mapper type a property on a core fetcher type (instead of a param to a couple of planner-centric funcs).

IIRC the only reason `DocumentMapping` is in the `core` package is because it had to be referenced by `fetcher.FetchNextDoc` and `encodedDocument.DecodeToDoc`.  It is very coupled to the planner system and if you really want it as a prop on a fetcher packages type, I would instead suggest that we instead move fetcher out of `db` and into `planner` - I don't like that very much either, but here we are increasing the level of coupling between the two.

Might be worth having a call about this one.

db/fetcher/encoded_doc.go
```diff
@@ -211,15 +234,35 @@ func (encdoc *encodedDocument) Decode() (*client.Document, error) {
...
+func (encdoc *encodedDocument) DecodeToDoc() (core.Doc, error) {
+	return encdoc.decodeToDoc(false)
+}
+
+func (encdoc *encodedDocument) decodeToDocForFilter() (core.Doc, error) {
+	return encdoc.decodeToDoc(true)
+}
+
+func (encdoc *encodedDocument) decodeToDoc(filter bool) (core.Doc, error) {
+	if encdoc.mapping == nil {
```
AndrewSisley, line 246: related to my comment on the `mapping` prop, I see the fact that we have to check this here (in the hot/usage path) and potentially error as a symptom of a bad constructor/type.

AndrewSisley (commented):
Leaving some early comments as I know you are actively working on this and looking for feedback asap. As mentioned over discord, I really like the core approach.

db/fetcher/encoded_doc.go
```diff
@@ -178,14 +183,32 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
...
-	Properties map[client.FieldDescription]*encProperty
+	Properties []*encProperty
+
+	// tracking bitsets
+	// A value of 1 indicates a required field
+	// 0 means we we ignore the field
+	// we update the bitsets as we collect values
+	// by clearing the bit for the FieldID
+	filterSet *bitset.BitSet // filter fields
+	selectSet *bitset.BitSet // select fields
```
AndrewSisley, line 198: todo: These appear unused and should be removed

db/fetcher/fetcher.go
```diff
@@ -48,14 +59,21 @@ type DocumentFetcher struct {
...
+
+	filterFields map[uint32]client.FieldDescription
+	selectFields map[uint32]client.FieldDescription
+	// allfields       map[uint32]*client.FieldDescription
+
+	// static bitsets
+	filterSet *bitset.BitSet
```
AndrewSisley, line 70: todo: The usage of this property is not self explanatory and needs code-comments, I am very tired atm, but it should still be easier to understand than it currently is for me :)

db/fetcher/encoded_doc.go
```diff
@@ -178,14 +183,32 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
...
 	Key        []byte
-	Properties map[client.FieldDescription]*encProperty
+	Properties []*encProperty
+
+	// tracking bitsets
+	// A value of 1 indicates a required field
+	// 0 means we we ignore the field
+	// we update the bitsets as we collect values
+	// by clearing the bit for the FieldID
+	filterSet *bitset.BitSet // filter fields
```
AndrewSisley, line 197: suggestion: I'm not a fan of the fact that this property is only ever used from outside of this type - to me it strongly suggests that it should be moved off the `encodedDocument` type and held somewhere else, but perhaps I will warm to it as my brain-fog fades and the documentation increases :)

nitpick: Also feels a bit odd that this (and `doc`, and the `encProperty`s in `Properties`) is a pointer.

db/fetcher/fetcher.go
```diff
@@ -218,91 +264,166 @@ func (df *DocumentFetcher) startNextSpan(ctx context.Context) (bool, error) {
...
+		Key:   dsKey,
+		Value: res.Value,
+	}
+	return false, kv, nil
+}
+
+// seekKV will seek through results/iterator until it reaches
+// the target key, or if the target key doesn't exist, the
+// next smallest key that is greater than the target.
+func (df *DocumentFetcher) seekKV(key string) (bool, *KeyValue, error) {
```
AndrewSisley, line 369: todo: It doesn't look like `key` needs to be a string, and it feels very strange that it is. Is this just because there is no easy to use std Golang []byte comparator and you want to save immediate dev time? If so I think it is worth documenting that (and very-maybe an issue).

db/fetcher/fetcher.go
```diff
@@ -370,12 +516,59 @@ func (df *DocumentFetcher) FetchNext(ctx context.Context) (*encodedDocument, err
...
+		if df.filter != nil {
+			fmt.Println("checking for seek")
+			if df.filterSet.Equal(df.doc.filterSet) {
+				filterDoc, err := df.doc.decodeToDocForFilter()
+				if err != nil {
+					return nil, err
+				}
+
+				// run filter
+				passed, err := mapper.RunFilter(filterDoc, df.filter)
```
AndrewSisley, line 530: suggestion: This at first glance looked like a horribly inefficient way of doing this, before I remembered that multiple fields might be involved. I think it is worth documenting this.

(also if for whatever reason you strongly disagree with that suggestion, please then remove the `// run filter` comment as it doesnt add any value - I assume you added it as you were planning out the implementation)

db/fetcher/fetcher.go
```diff
@@ -370,12 +516,59 @@ func (df *DocumentFetcher) FetchNext(ctx context.Context) (*encodedDocument, err
...
-			return df.doc, nil
+		if docDone {
+			fmt.Println("doc done")
+			// run filter
+			if df.filter != nil {
+				decodedDoc, err := df.doc.DecodeToDoc()
+				if err != nil {
+					return nil, err
+				}
+				passed, err := mapper.RunFilter(decodedDoc, df.filter)
```
AndrewSisley, line 554: todo: You appear to be running the filter twice when `docDone` is true, please remove that before merge :)

db/fetcher/encoded_doc.go
jsimnz: Its a hard call, I'd say youre both right and wrong here.

Its needed here as you note because of `fetcher.FetchNextDoc` which returns a `core.Doc` type, which is used by the planner. In reality, the `DocumentMapping` and `core.Doc` could really be used anywhere, as its conceptually only tied to the `CollectionDescription`. 

I was almost going to hoist it into the description just to see what that would look like, as that would mean a `core.Doc` could be used anywhere. Theres nothing specific `core.Doc` has to the planner. The goal of it is to have an effecient array based representation for a document, instead of relying on a `map[string]any` everywhere. Its primarily used in the planner, but as there are other places that document handling needs to be refactored (#935) it could be the case the `core.Doc` is used everywhere.

But thats a bigger conversation that we don't need to get into.
--

Bottom line, the fetcher now needs to be able to run a filter evaluation, it can use the old style of filter evaluation that was used before the mapper existed, but then we would be doing double work to decode the document. Once to a `map[string]any` to run with the old version of `parser.RunFilter`, and once to return for `FetchNextDoc`.

Additionally, when `mapper` was implemented, the `connor` package was migrated to use `map[FilterKey]any` instead of `map[string]any` which is generated in `mapper/targetable.go`.

So it seems its somewhat bound to use the mapper system due to the filter requirements from within the fetcher.

Open to suggestions.
--

One reason to avoid putting `fetcher` in `planner` is that it can be run entirely on its own, as you can see in `db/fetcher_test.go`, and from within the Collection API, which uses the fetcher.

Happy to have a call 📞

db/fetcher/encoded_doc.go
jsimnz: the cost of checking this is miniscule compare to the cost of decoding and running the filter.

Same answer to my thoughts above. Fetcher now needs to run filters, filters need a `core.Doc`, which you can only get by using a `core.DocumentMapping`.

If we use `map[string]any` to run the filter on the fetcher (like we did pre-mapper days) then we can remove the dependency on the mapping stuff, but then we double the decoding work.

db/fetcher/encoded_doc.go
jsimnz: Most of the properties of `encodedDocument` are directly accessed outside the type. Its really just a container type for holding state, and then its reponsible for combining state.

the doc `filterSet` could be on the `fetcher` directly, which would then have two sets, a "static" and a "dynamic" set, but as this is specific to the other fields within the `encodedDocument` type, it felt appropriate to keep here. Also means if I need to pass around the doc for any reason, the bitset goes with it.

re nitpick: Theres no good reason for the `encProperty` to be a pointer, but `BitSet` *must* be a pointer type as that's how its defined in its package. `New` method returns pointer, all methods are defined on pointer reciever, etc.

db/fetcher/encoded_doc.go
AndrewSisley: I'm much less worried about performance here - I flagged this as it is an error path that we do not/should not need in the 'doing' part of the code

db/fetcher/fetcher.go
jsimnz: This was mostly just a conceptual similarity choice as atm the KV level keys are considered strings, but digging further into it theres benefits to being a string.

Firstly, there is an equivalent byte comparator as string comparator.

But, at the moment, `core.DataStoreKey.Bytes()` just calls `[]byte(datastorekey.ToString())` anyway. Which will result in an extra allocation anyway ("safe" conversion between strings and bytes allocates+copies).

At the moment, theres just a lot of conversion back and forth between strings, bytes, back to strings, and bytes, and prob a few more. So until we get organized around the proper binary serialization, it doesn't really matter which to use here. But, the one fact we know is that `datastorekey.ToString()` is cheaper than `datastorekey.Bytes()`. So its a toss.

db/fetcher/fetcher.go
jsimnz: there was a comment above the `if docDone` line that said `// if docDone` xD. Just my process, thanks for flagging the comment.

What seemed ineffecient at first glance? It gets run anyway, previously on scanner, now on fetcher.

db/fetcher/errors.go
```diff
@@ -25,6 +25,7 @@ const (
 	errVFetcherFailedToDecodeNode   string = "(version fetcher) failed to decode protobuf"
 	errVFetcherFailedToGetDagLink   string = "(version fetcher) failed to get node link from DAG"
 	errFailedToGetDagNode           string = "failed to get DAG Node"
+	errMissingMap                   string = "missing document mapper"
```
shahzadlone, line 28: nitpick: 
```suggestion
	errMissingMapper                   string = "missing document mapper"
```

planner/scan.go
```diff
@@ -103,23 +115,32 @@ func (n *scanNode) Next() (bool, error) {
...
-			return false, err
-		}
-		n.execInfo.docFetches++
+	var err error
+	n.docKey, n.currentValue, err = n.fetcher.FetchNextDoc(n.p.ctx, n.documentMapping)
+	if err != nil {
+		return false, err
+	}
+	// @question: Should this be after the length check?
+	n.execInfo.docFetches++
```
shahzadlone, line 124: answer: Represents the attempts at fetching docs. Since the Fetch call is before and we do this after ensuring it was not a failure, it's fine where it is at the moment.

planner/scan.go
shahzadlone: Technically if `len(n.currentValue.Fields) != 0` shouldn't the state still be the same (i.e. doc match was found)? It's not explicitly a filter match, but would be fine to keep it anywhere after the if and before the `return true, nil`.
Also, open to changing the name if you have any better ideas to represent a "successful document found while scanning / fetching, which matched / passed the filters"

shahzadlone (commented):
Skimmed high-level, I mostly really like the changes, much smaller and simpler than the previous fetcher stuff.

Shared my thoughts on the questions you asked.

Will give a more deeper second look soon.

planner/scan.go
jsimnz: Commenting here for both this answer and your other comment.

Technically it means that `docFetches` and `filterMatches` will basically be the same value (or differ by at most 1).

db/fetcher/encoded_doc.go
fredcarle: If we need to use a pointer's value within the function that doesn't define it, then we need to check that it's not nil. I feel like it makes sense to do that here.

db/fetcher/fetcher.go
fredcarle: I think we can revisit this when we change the key encoding.

db/fetcher/fetcher.go
fredcarle: I struggle to see this. Can you point me to where the second run of the filter happens?

db/fetcher/fetcher.go
```diff
@@ -118,14 +149,29 @@ func (df *DocumentFetcher) init(
 	}
 	df.kvIter = nil
 
-	df.schemaFields = make(map[uint32]client.FieldDescription)
-	for _, field := range col.Schema.Fields {
-		df.schemaFields[uint32(field.ID)] = field
+	df.selectFields = make(map[uint32]client.FieldDescription, len(fields))
+	// @todo: We can probably make this size more accurate, this is an upper bound
```
fredcarle, line 153: question: How would you make this more accurate? You seem to be using all the allocated KVs in the for loop.

db/fetcher/fetcher.go
```diff
@@ -13,27 +13,38 @@ package fetcher
...
+	Init(col *client.CollectionDescription, fields []client.FieldDescription, filter *mapper.Filter, docmapper *core.DocumentMapping, reverse bool, showDeleted bool) error
 	Start(ctx context.Context, txn datastore.Txn, spans core.Spans) error
 	FetchNext(ctx context.Context) (*encodedDocument, error)
 	FetchNextDecoded(ctx context.Context) (*client.Document, error)
 	FetchNextDoc(ctx context.Context, mapping *core.DocumentMapping) ([]byte, core.Doc, error)
 	Close() error
 }
 
+// KeyValue is a KV store response containing the resulting core.Key and byte array value.
+type KeyValue struct {
```
fredcarle, line 43: question: Does this need to be public? Asking because it doesn't seem to be use outside of this file.

fredcarle: Nevermind. `DocumentFetcher.KV()` returns this type.

db/fetcher/fetcher.go
```diff
@@ -370,12 +516,59 @@ func (df *DocumentFetcher) FetchNext(ctx context.Context) (*encodedDocument, err
...
+				return df.doc, nil
+			}
+
+			if !spansDone {
+				fmt.Println("span not done yet...")
+				continue
+			}
+
+			return nil, nil
 		}
```
fredcarle, line 572: suggestion: This could probably be changed to
```go
if docDone {
    return df.doc, nil
}
if spanDOne {
    return nil, nil
}
```

planner/scan.go
shahzadlone: > Technically it means that `docFetches` and `filterMatches` will basically be the same value (or differ by at most 1).

I am going to talk from the perspective of what used to happen... The majority of the time, `docFetches` and `iterations` would be similar (because for every iteration we would attempt to fetch. But it's **not strictly** true. Also `docFetches` and `filterMatches` will mostly **NOT be same**, their difference is **not strictly** most by 1. It all depended on the number of docs and how many match the query or not. Consider these examples from our current execution explain tests.

## Consider scan node execute results...

### `TestExecuteExplainRequestWithOrderFieldOnParent`
```
"iterations":    uint64(3),
"docFetches":    uint64(3),
"filterMatches": uint64(2),
```

### `TestExecuteExplainMutationRequestWithCreate`
```
"iterations":    uint64(1),
"docFetches":    uint64(1),
"filterMatches": uint64(1),
```

### `TestExecuteExplainMutationRequestWithDeleteUsingFilter`
```
"iterations":    uint64(2),
"docFetches":    uint64(3),
"filterMatches": uint64(1),
```

### `TestExecuteExplainMutationRequestWithUpdateUsingFilter`
```
"iterations":    uint64(4),
"docFetches":    uint64(6),
"filterMatches": uint64(2),
```

### `TestExecuteExplainRequestWithDocumentsButNoMatches`
```
"iterations":    uint64(1),
"docFetches":    uint64(3),
"filterMatches": uint64(0),
```

For example in the last example it had two documents, it fetched 2 times and none matched the filter. The third time it fetched there was no more docs so total attempts are 3, with 0 matches, and we entered scan node once.

## Fetcher update in this PR
I have to read the implementation in more detail to see if it is still possible to differentiate between the `attempts fetches took place till match found` vs `total maches found`. Might even just need to return a variable with result from the fetcher call if we want to keep track similar to before of all the fetch attempts till match.

db/fetcher/encoded_doc.go
jsimnz: > I'm much less worried about performance here - I flagged this as it is an error path that we do not/should not need in the 'doing' part of the code

This doesn't address my point that we *must* do this here, I see no way around this. Moreover, this is based on how you wrote the mapper/filter/doc.

Tangentially, your noted objection to these `x != nil` come up all the time in reviews, and we keep reminding you that its as command and idiomatic as it gets for Go code.

planner/scan.go
jsimnz: My whole point is that because of the changes in this PR, the behavior is now different. Previously, the scanner "fetched" a document from the fetcher, then the scanner was responsible for filtering. 

One of the core changes in this PR is that we are moving the filtering *from* the scanner *to* the fetcher.

So keeping track of both `filterFetches` and `filterMatches` in the scanner is meaningless now.

db/fetcher/fetcher.go
jsimnz: the `@todo` here is old, the line it refers to has been removed since opening the PR xD

db/fetcher/fetcher.go
jsimnz: It doesn't need to be public. The `NextKV`, `NextKey`, and `ProcessKV` public methods all don't need to exist, theyre old methods mostly for debugging, they can be removed, so KeyValue can be made private.

planner/scan.go
shahzadlone: Yup makes sense, I did review the implementation in more depth and seems like in order to get that info now we would have to have the fetcher logic store that metric and then communicate back to the scanner. IMO let's keep the metrics we collect scoped to planner nodes only so the ones we need to bend over backward for in a more complicated manner "out-side" the plan nodes, I would skip those metrics for now.
tldr: can just remove/skip the metric that isn't as relevant now

db/fetcher/encoded_doc.go
jsimnz: Even if it's a function parameter it can still be nil, it doesn't change anything.

In fact, it is a function parameter, just at a different function / call site to initialize the fetcher.

It should still be checked.

--

Regarding the idiomatic. Ultimately it doesn't matter what other languages do. This *Is* Go, and Go convention wins. The language authors and core maintainers have been very explicit that the goals as a language are not to be fancy, and will often ignore more modern PL techniques and research, in an attempt at "simplicity" and clarity.

The pointer has to be checked regardless. Putting it in something like a homegrown results or option container doesn't change that, nor does making it a function argument.

My comment was directed at the historical nature of this thread, as it comes up often, and the response is often the same on both ends.

PS. Word for word my previous response was 50% directed at you lol. The first half of the message was at your suggestion, as I pointed out there's almost no way around having the doc mapper in the fetcher unless either refactor the filter stuff, or double the decoding work. Neither seem ideal compared to leaving as is.

jsimnz: There's a reply here in the other comment thread you raised the issue on too, with my same points, from Friday.

https://github.com/sourcenetwork/defradb/pull/1500#discussion_r1192614605

db/fetcher/encoded_doc.go
```diff
@@ -27,6 +28,10 @@ type encProperty struct {
 	Desc client.FieldDescription
 	Raw  []byte
 
+	// Filter flag to determine if this flag
+	// is needed for eager filter evaluation
+	Filter bool
```
islamaliev, line 33: question : don't we want to stick to a "normal" convention of naming `bool` vars with prefixes like "Is", "Has"... This one could be names `ShouldFilter`
Current name is surely shorted, but reading `if prop.Filter {` feels a bit ambiguous (at least to my view)
Maybe it's still my C++ vision where you can if basically any type :)

db/fetcher/encoded_doc.go
```diff
@@ -211,15 +234,35 @@ func (encdoc *encodedDocument) Decode() (*client.Document, error) {
...
+	if encdoc.doc == nil {
+		doc := encdoc.mapping.NewDoc()
+		encdoc.doc = &doc
+	}
+	encdoc.doc.SetKey(string(encdoc.Key))
+	for _, prop := range encdoc.Properties {
+		if encdoc.doc.Fields[prop.Desc.ID] != nil { // used cached decoded fields
+			continue
+		}
+		if filter && prop.Filter != filter { // only get filter fields if filter=true
```
islamaliev, line 258: nitpick: could write `if filter && !prop.Filter {`

islamaliev (commented):
I assume our fetcher is already somewhat covered with tests, but I see here quite some new or updated logic (like filtering, mapping, seeking...). Isn't it supposed to be covered as well?

db/fetcher/encoded_doc.go
AndrewSisley: We spoke about this a while back, and I don't think we'll see eye-to-eye on this any time soon :)  Author view to trump reviewer (ignoring CTO status ;)), and mark this convo as resolved.

I have a plan for the fetcher that I want to try out in the near future anyway, which will pretty much bypass this conflict entirely if it works out.

db/fetcher/encoded_doc.go
AndrewSisley: Resolving this convo, as it is going nowhere, and is of very low value lol.  Leave as-is if prefered.

db/fetcher/encoded_doc.go
AndrewSisley: > Most of the properties of encodedDocument are directly accessed outside the type. Its really just a container type for holding state, and then its reponsible for combining state.

Didnt realize you saw it that way - in that case I do think that this should have no funcs defined on it then as that means this type is used both as an object, and as a data-struct, which I really dislike. Change that is well out of scope regardless of what anyone thinks of it here though.  Convo resolved :)

db/fetcher/fetcher.go
AndrewSisley: > At the moment, theres just a lot of conversion back and forth between strings, bytes, back to strings, and bytes, and prob a few more. So until we get organized around the proper binary serialization, it doesn't really matter which to use here. But, the one fact we know is that datastorekey.ToString() is cheaper than datastorekey.Bytes(). So its a toss.

Ah fair enough, happy as-is then.

> I think we can revisit this when we change the key encoding.

I do see that as quite tangential/separate, and wouldnt recommend doing those two tasks together.

db/fetcher/fetcher.go
AndrewSisley: > What seemed ineffecient at first glance?

I honestly cant remember, will reply again if I think I remember and if this code is still relevant (just going through the old comments before reviewing the new code atm)

db/fetcher/encoded_doc.go
```diff
@@ -178,14 +183,32 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
...
+	selectSet *bitset.BitSet // select fields
+
 }
 
 // Reset re-initializes the EncodedDocument object.
 func (encdoc *encodedDocument) Reset() {
-	encdoc.Properties = make(map[client.FieldDescription]*encProperty)
+	encdoc.Properties = make([]encProperty, 0)
 	encdoc.Key = nil
+	if encdoc.mapping != nil {
```
AndrewSisley, line 206: suggestion: This is a bit odd without a `New` function.  Here `mapping` should never be nil no?  Yet at the same time if it ever is, the `doc` property will never be cleared by reset leaving this instance in an invalid state.

I'd suggest adding a `New(...) EncodedDocument` func that ensures `mapping` is never nil, or if for whatever reason you really dont want that, add an `else` clause that just sets `doc` to `nil`.

db/fetcher/fetcher.go
```diff
@@ -13,27 +13,39 @@ package fetcher
...
 	"github.com/sourcenetwork/defradb/db/base"
+	"github.com/sourcenetwork/defradb/planner/mapper"
+	"github.com/sourcenetwork/defradb/request/graphql/parser"
 )
 
 // Fetcher is the interface for collecting documents from the underlying data store.
 // It handles all the key/value scanning, aggregation, and document encoding.
 type Fetcher interface {
-	Init(col *client.CollectionDescription, fields []*client.FieldDescription, reverse bool, showDeleted bool) error
+	Init(col *client.CollectionDescription, fields []client.FieldDescription,
```
AndrewSisley, line 33: todo: Breaking up the props onto multiple lines, but still with multiple props per line is really quite hard to read and I'd much prefer that each prop would have their own line. and not share with the func name and brackets:

```
foo(
    a int,
    b int,
    ...
)
```

db/fetcher/fetcher.go
```diff
@@ -370,12 +523,58 @@ func (df *DocumentFetcher) FetchNext(ctx context.Context) (*encodedDocument, err
 			return nil, err
 		}
 
-		end, err := df.nextKey(ctx)
+		if df.filter != nil {
+			// only run filter if we've collected all the fields
+			// required for filtering. This is tracked by the bitsets.
+			if df.filterSet.Equal(df.doc.filterSet) {
+				df.ranFilter = true
```
AndrewSisley, line 530: todo: Minor, this looks like it is set in the wrong location - if errors are produced `ranFilter` will be true, yet the filter may not have been run.  I think this should be set after the error check for `mapper.RunFilter` (currently between ln 539 and 540)

planner/datasource.go
```diff
@@ -43,7 +43,7 @@ func (p *Planner) getCollectionScanPlan(parsed *mapper.Select) (planSource, erro
 	}
 
 	scan := p.Scan(parsed)
-	err = scan.initCollection(colDesc)
+	err = scan.initCollection(colDesc) // @todo: Should `initCollection` be merged into `p.Scan(...)`
```
AndrewSisley, line 46: todo: I have an very large preference for this todo to be done.  I guess it is technically out of scope, but `initCollection` is amusingly horrible (presumably from a time where unsure whether it would be shared/re-inited) and it would be quick todo.

Failing that, please make sure that either the todo is removed, or is bound to a ticket, before you merge this PR

planner/mapper/mapper.go
```diff
@@ -119,36 +120,91 @@ func resolveOrderDependencies(
...
 	// If there is orderby, and any one of the condition fields that are join fields and have not been
 	// requested, we need to map them here.
+outer:
 	for _, condition := range source.Value().Conditions {
-		if len(condition.Fields) <= 1 {
-			continue
-		}
-
-		joinField := condition.Fields[0]
+		fields := condition.Fields[:] // copy slice
```
AndrewSisley, line 125: todo: this copy appears to be pointless, as the only other write is a full overwrite of the variable. You should be able to just do `fields := condition.Fields` here (as it was before)

planner/mapper/mapper.go
```diff
@@ -119,36 +120,91 @@ func resolveOrderDependencies(
...
+		for {
+			numFields := len(fields)
+			if numFields == 2 {
```
AndrewSisley, line 128: todo: Please document as to why toggling on `numFields == 2`, ` > 2` and `<= 1`, as it is not immediately clear to me

planner/mapper/mapper.go
```diff
@@ -119,36 +120,91 @@ func resolveOrderDependencies(
...
-			mapping.Add(index, joinField)
+				// make sure the actual target field inside the join field
+				// is included in the select
+				targetFieldName := condition.Fields[1]
+				targetField := &Field{
+					Index: innserSelect.FirstIndexOfName(targetFieldName),
+					Name:  targetFieldName,
+				}
+				innserSelect.Fields = append(innserSelect.Fields, targetField)
+				continue outer
```
AndrewSisley, line 145: todo: The two `continue outer` statements appear to be redundant, as there is nothing else within the loop that would otherwise be executed.  Please remove them along with the declaration of `outer`.

planner/mapper/mapper.go
AndrewSisley: This actually smells a bit like a bug - do you mean to use `fields` for something besides a counter?  If so can you please add a failing test, then fix the bug (within this PR)?  Or otherwise quite heavily document why the counter is as such, yet you are not really using it for anything besides the `if`s

planner/mapper/mapper.go
```diff
@@ -119,36 +120,91 @@ func resolveOrderDependencies(
...
+func resolveChildOrder(
+	descriptionsRepo *DescriptionsRepo,
+	descName string,
+	orderChildField string,
+	mapping *core.DocumentMapping,
+	existingFields *[]Requestable,
+) (*Select, error) {
+	childFieldIndexes := mapping.IndexesByName[orderChildField]
+	// Check if the join field is already mapped, if not then map it.
+	if isOrderJoinFieldMapped := len(childFieldIndexes) != 0; !isOrderJoinFieldMapped {
```
AndrewSisley, line 177: suggestion: I'm not sure `isOrderJoinFieldMapped` adds anything positive here, suggest removal and just do `if len(childFieldIndexes) != 0 {`

planner/mapper/mapper.go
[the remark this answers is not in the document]
AndrewSisley: If you did this you could then move the contents of `resolveChildOrder` into this function and remove the pointer based mutation stuff that I was going to otherwise comment about (param `existingFields`), as `resolveChildOrder` will only be called once.

planner/mapper/mapper.go
```diff
@@ -251,13 +306,15 @@ func resolveAggregates(
 					return nil, err
 				}
 
+				// ensure the field were aggregating is included in the join select
```
AndrewSisley, line 309: question: This comment appears misplaced?  Can it be removed?

planner/mapper/mapper.go
```diff
@@ -743,7 +806,14 @@ func resolveInnerFilterDependencies(
 
 		if keyIndex >= len(mapping.ChildMappings) {
 			// If the key index is outside the bounds of the child mapping array, then
-			// this is not a relation/join and we can continue (no child props to process)
+			// this is not a relation/join and we can add it to the fields and
+			// continue (no child props to process)
+
+			newFields = append(existingFields, &Field{
```
AndrewSisley, line 812: question: I cannot tell why this has been added - can you please explain?

planner/mapper/targetable.go
```diff
@@ -86,6 +86,51 @@ func NewFilter() *Filter {
...
+			outkey, _ := mapping.TryToFindNameFromIndex(keyType.Index)
+			if childMapping, ok := tryGetChildMapping(mapping, keyType.Index); !ok {
+				outmap[outkey] = filterObjectToMap(mapping, subObj)
+			} else {
+				outmap[outkey] = filterObjectToMap(childMapping, subObj)
+			}
+
+		case *Operator:
+			switch keyType.Operation {
+			case "_and", "_or":
```
AndrewSisley, line 111: suggestion: `"_and"` and `"_or"` should really be consts in the `client.request` package.  Marked as suggestion as you could happily argue that this is out of scope and you are just moving them.

planner/mapper/targetable.go
```diff
@@ -86,6 +86,51 @@ func NewFilter() *Filter {
 	}
 }
 
+func (f *Filter) ToMap(mapping *core.DocumentMapping) map[string]any {
```
AndrewSisley, line 89: todo: I think it is worth adding a quick comment to this function, as it is bit involved and the `any` part of the return type is not very descriptive.

planner/mapper/targetable.go
```diff
@@ -86,6 +86,51 @@ func NewFilter() *Filter {
...
+	outmap := make(map[string]any)
+	if obj == nil {
+		return nil
+	}
+	for k, v := range obj {
+		switch keyType := k.(type) {
+		case *PropertyIndex:
+			subObj := v.(map[connor.FilterKey]any)
+			outkey, _ := mapping.TryToFindNameFromIndex(keyType.Index)
+			if childMapping, ok := tryGetChildMapping(mapping, keyType.Index); !ok {
```
AndrewSisley, line 103: suggestion: The `!ok` is a bit weird here as `childMapping` is consumed in the `else` not the `if`.  I was going to suggest just inverting, but you can probably go one better and remove the duplicate `filterObjectToMap` func call and `outmap` set:

```
var propMapping *core.DocumentMapping
if childMapping, ok := tryGetChildMapping(mapping, keyType.Index); ok {
    propMapping = childMapping
} else {
    propMapping = mapping
}
outmap[outkey] = filterObjectToMap(propMapping, subObj)
```

planner/scan.go
```diff
@@ -61,17 +63,84 @@ func (n *scanNode) Kind() string {
...
-	if err := n.fetcher.Init(&n.desc, n.fields, n.reverse, n.showDeleted); err != nil {
+	if err := n.fetcher.Init(&n.desc, n.fields, n.filter, &n.slct.DocumentMapping, n.reverse, n.showDeleted); err != nil {
 		return err
 	}
 	return n.initScan()
 }
 
 func (n *scanNode) initCollection(desc client.CollectionDescription) error {
 	n.desc = desc
+	n.initFields(n.slct.Fields)
```
AndrewSisley, line 74: question: I dont understand why you are doing all this here still? If you have sorted everything out in the mapper, you should not have to do anything like this in the planner.

AndrewSisley (changes requested):
I went through again, happy to have a call sometime - I don't understand why you are still doing a bunch of mapper work within the planner (scanNode)

planner/sum.go
```diff
@@ -43,6 +43,7 @@ func (p *Planner) Sum(
 	field *mapper.Aggregate,
 	parent *mapper.Select,
 ) (*sumNode, error) {
+	// panic("sum")
```
AndrewSisley, line 46: todo: Please remove :)

db/fetcher/encoded_doc.go
jsimnz: Wanted a `Reset` instead of a `New` so it keeps using the same allocated/defined struct on the fetcher.

As for the `else`, the doc will already be `nil` if the `mapping` is also `nil`, and the mapping never gets updated once its set from the beginning, so its either `nil` or not for the whole thing. I can add if you you prefer, but its not a codepath that will add any meaning/safety imo.

db/fetcher/fetcher.go
jsimnz: Yea I didn't like it when I was doing it, but I also hate one line per for interface definitions. But I guess Ill have to, since the both alternatives suck more 🙃

db/fetcher/fetcher.go
jsimnz: Ahh, yea it should be. I don't think it makes a diff since it will error out anyway, but for sanity 👍

planner/mapper/mapper.go
jsimnz: Its copied because of this line further down in the `else if numFields > 2`:
```go
fields = fields[1:] // chop off the front item, and loop again on inner
```

Without the copy this would mutate the original slice.

planner/mapper/mapper.go
jsimnz: Added comment

planner/mapper/mapper.go
```diff
@@ -119,36 +120,91 @@ func resolveOrderDependencies(
...
-
-		joinField := condition.Fields[0]
+		fields := condition.Fields[:] // copy slice
+		for {
+			numFields := len(fields)
+			if numFields == 2 {
+				joinField := condition.Fields[0]
+
+				// ensure the child select is resolved for this order join
+				innserSelect, err := resolveChildOrder(descriptionsRepo, descName, joinField, mapping, existingFields)
```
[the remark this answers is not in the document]
jsimnz: good catch

planner/mapper/mapper.go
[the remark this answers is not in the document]
jsimnz: I wanted each `if/elseif/else` to be explicit, as they need to handle different cases, so I didnt want to get too smart about shared logic between them.

I'd also prefer to keep `resolveChildOrder` to not make this function more complex then it needs to be as its already a double `for` loop and a muli-level `else if`. Strong preference to keep it out.

As for the pointer based mutation stuff on the `existingFields` that was like this before this PR, so im just keeping the existing semantics.

lmk

planner/mapper/mapper.go
jsimnz: The main theme of most (all?) of the `mapper` changes is that it originally didn't explicitly state a lot of the dependent fields that are indirectly required through filters, orders, aggregates, groups, etc. Since before, the `fetcher` just always *all* fields, so we didn't need to be picky.

This particular change will add a related type field defined on a parent filter to the child selection set.
ie:
```graphql
query {
  User(filter: {friend: {age: _eq: 10}} {
    name
  }
}
```

This query would generate a virtual sub select as we dont include the `friend` sub type in the selection set, however, before it basically would be an empty selections set on the `friend` type, because again, the fetcher would automatically return all fields. 

With the new changes to the fetcher, we need to include the fields on the virtual sub select that we care about.

planner/mapper/targetable.go
jsimnz: Theres a number of spots this could be applied, so Ill mark as out of scope, and open an issue 👍

planner/mapper/targetable.go
jsimnz: just going to invert the `if/else` - again I try not to be too clever to reduce lines by pulling out common stuff in `if` blocks, unless it really adds clarity/semantics somehow. Given the one-liner of each, I like the clarity it current has.

planner/scan.go
jsimnz: Mentioned elsewhere briefly, but to summarize. All the changes in the `mapper` are to surface the fields we actually need/care about.

These fields are surfaced in the mapper, so they can be consumed here, by explicitly telling the fetcher (via the scanNode) which fields we care about.

jsimnz:
> I went through again, happy to have a call sometime - I don't understand why you are still doing a bunch of mapper work within the planner (scanNode)

Mentioned in some of the comments, this PR effectively does 2 things. 
1) The actual fetcher optimization for field selection. 
2) The mapper changes to surface all the various ways where dependent fields are actually needed.

The code in the planner (scan node specifically) is effectively the bridge between these two changes. Once the required fields have been identified and indexed in the mapper (via filters, aggregates, groups, orders, sub types, etc) the fetcher is told which fields we need.

Happy to go thru on a call, lmk.

planner/mapper/targetable.go
jsimnz: The return is a `map[string]any` which is the most common map type used in the entire code base for representing query related stuff. But I will add a docstring for the function

db/fetcher/encoded_doc.go
AndrewSisley: > Wanted a Reset instead of a New so it keeps using the same allocated/defined struct on the fetcher.

I think you misunderstand why I mentioned `New`, by declaring a constructor you are declaring a restricted initial state.  There you can enforce `mapping` is not nil and reset loses a dead code branch (the implicit else), as well as possibility of the invalid `doc` state that I mentioned. `New` can even call `Reset` internally for a stronger initial state guarantee/description if you want.

planner/mapper/mapper.go
AndrewSisley: This is part of my point. You are copying the slice multiple times but only ever using its length.

planner/mapper/mapper.go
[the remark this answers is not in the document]
AndrewSisley: >  I didnt want to get too smart about shared logic between them.

I dont think what I'm suggesting is particularly fancy, it is the same code, just with reduced duplication and if-else scope.  I dont see why you wouldnt want that.

> Strong preference to keep it out.

Fair enough, I will review it then.

planner/mapper/mapper.go
AndrewSisley: Ah of course - and the 'new' leaf fields are added to the  `ChildMappings` - thanks for the explanation. I wonder if it is worth rewording the code-comment a little more to make this a bit more clear?

planner/mapper/targetable.go
AndrewSisley: Fair enough, happy with just the inversion - I think which one I'd go with myself would be dependent on time-of-day/position-of-Jupiter/etc lol

planner/mapper/targetable.go
AndrewSisley: > The return is a map[string]any which is the most common map type used in the entire code base for representing query related stuff.

I see that as making it almost worse though, as this type in 99% of other cases is a document, not a gql filter with a fairly limited shape.

planner/scan.go
```diff
@@ -60,17 +63,84 @@ func (n *scanNode) Kind() string {
...
+					n.tryAddField(target.ChildTarget.Name)
+				} else {
+					n.tryAddField(target.Field.Name)
+				}
+			}
+		}
+	}
+}
+
+func (n *scanNode) tryAddField(fieldName string) bool {
```
AndrewSisley, line 132: todo: As far as I can see the returned boolean is never used, can you please remove it (or use it), and maybe rename this function to `addIfNotExists` or similar

planner/scan.go
```diff
@@ -60,17 +63,84 @@ func (n *scanNode) Kind() string {
...
+		// for, including those that are needed by higher level aggregates
+		// or grouping alls, which them selves might have further dependants
+		switch requestable := r.(type) {
+		// field is simple as its just a base level field
+		case *mapper.Field:
+			n.tryAddField(requestable.GetName())
+		// select might have its own select fields and filters fields
+		// @todo: Might need to check that these are sub selects
+		// are _groups
+		case *mapper.Select:
```
AndrewSisley, line 98: question: This case looks like it should be impossible within the scanNode - are you really sure it ever gets run?  And if so when?

planner/scan.go
```diff
@@ -60,17 +63,84 @@ func (n *scanNode) Kind() string {
...
+					n.desc.Schema,
+				)
+				for _, fd := range fieldDescs {
+					n.tryAddField(fd.Name)
+				}
+			}
+			n.tryAddField(requestable.Field.Name + "_id") // foreign key for type joins
+			n.initFields(requestable.Fields)
+		// aggregate might have its own target fields and filter fields
+		case *mapper.Aggregate:
```
AndrewSisley, line 111: question: Similar to the `*mapper.Select` case - are you really sure this is possible here? And if so, how?

planner/scan.go
```diff
@@ -60,17 +63,84 @@ func (n *scanNode) Kind() string {
...
+	if err := n.fetcher.Init(&n.desc, n.fields, n.filter, n.slct.DocumentMapping, n.reverse, n.showDeleted); err != nil {
 		return err
 	}
 	return n.initScan()
 }
 
 func (n *scanNode) initCollection(desc client.CollectionDescription) error {
 	n.desc = desc
+	n.initFields(n.slct.Fields)
+	if n.slct.OrderBy != nil {
```
AndrewSisley, line 75: question: I think this might have been what threw me at first, and I'm still not sure why you are inspecting order conditions here. Why is this necessary, and are you really sure it is?  It does look like you could delete this block.

planner/mapper/mapper.go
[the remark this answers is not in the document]
AndrewSisley: > Fair enough, I will review it then.

Looked it over - I hate the input pointer mutation stuff, but that is inherited from this function itself (I missed that yesterday), and whilst if that variable was not a pointer in `resolveOrderDependencies` it would be easy enough to write `resolveChildOrder` in such a way to avoid the pointer stuff, it is not, and changing that is much more of a pain and out of scope.

I do also dislike that `resolveChildOrder` is really two functions pretending to be one, on account of the if-else, but that is fairly minor and easy to change in the future if anyone feels the need to.

planner/scan.go
jsimnz: yupp

planner/scan.go
jsimnz: This one may be changed since it was defined. I'll double check if this one is duplicating efforts, as a lot of the order by changes came after this bit. Will confirm

planner/mapper/mapper.go
jsimnz: sorry, its *mutated* in place after its copied

jsimnz: Apologies, talking with fred, I realized there is a bug (not sure why the tests arent tracking), I think this was potentially the result of a bad rebase, not sure. Currently its `joinField := condition.Fields[0]`, but it should be `joinField := fields[0]` which is why the `fields` slice keeps getting popped from the front.

It *can* still be replaced with length and index tracking, but the idea that the `fields` *should* be used to track the "recursion" depth thru the slice, and we just pop from the front. Will update and we can decide if it should be changed.

jsimnz: Updated @AndrewSisley, see latest commit [here](https://github.com/sourcenetwork/defradb/pull/1500/commits/1c0d009b2d564dabafb575d627ec4efdaf23dffc). 

Hopefully makes more sense, and correcting the bug also affected another comment you had about the order stuff in the scanNode.

planner/scan.go
AndrewSisley: Same as other comment, please expand

planner/mapper/mapper.go
AndrewSisley: Ah okay, makes sense now - thanks.  Any chance you can please add a test that would have caught that bug?  If it caught you out it can easily catch someone else out modifying the code in the future.

planner/mapper/targetable.go
jsimnz: The same structured is used for parsing filter and order types, its used for rendering documents to their map, etc. Its usually pretty clear that its just a generic dictionary style map with string keys, and w.e values. This isn't any diff. +1 for the comment on the function, but the return type should be fairly clear as far as Go idiom is concerned.

planner/mapper/mapper.go
jsimnz: Ill see if I can add one, but I don't think so beyond what we already have. The reason being, despite there being a bug here, it was accidentally being fixed in the scanNode `initCollection` method when it did its own `order` processing. When that was removed and this was fixed, they cancelled eachother out.

All 3 cases of this `if` block are covered by the existing tests.

planner/mapper/mapper.go
jsimnz: > Ah of course - and the 'new' leaf fields are added to the ChildMappings - thanks for the explanation. I wonder if it is worth rewording the code-comment a little more to make this a bit more clear?

Can update the comment. Small note, the new leaf fields are added to the `mapper.Select` type, not the ChildMapping.

planner/mapper/mapper.go
fredcarle: This makes a lot more sense now. Thanks for fixing!

planner/type_join.go
```diff
@@ -545,12 +555,23 @@ func (n *typeJoinMany) Next() (bool, error) {
...
 		// @todo: handle index for one-to-many setup
 	} else {
+		// filterRaw := request.Filter{
+		// 	Conditions: map[string]any{
+		// 		n.rootName + "_id": map[string]any{
+		// 			"_eq": n.currentValue.GetKey(),
+		// 		},
+		// 	},
+		// }
+		// filter := mapper.ToFilter(filterRaw, n.subType.DocumentMap())
```
fredcarle, line 565: todo: please remove commented code :)

planner/type_join.go
```diff
@@ -545,12 +555,23 @@ func (n *typeJoinMany) Next() (bool, error) {
...
 		fkIndex := &mapper.PropertyIndex{
 			Index: n.subSelect.FirstIndexOfName(n.rootName + "_id"),
 		}
 		filter := map[connor.FilterKey]any{
-			fkIndex: n.currentValue.GetKey(), // user_id: "bae-ALICE" |  user_id: "bae-CHARLIE"
+			fkIndex: map[connor.FilterKey]any{
+				mapper.FilterEqOp: n.currentValue.GetKey(), // user_id: "bae-ALICE" |  user_id: "bae-CHARLIE"
```
fredcarle, line 571: question: Is the comment still relevant? I can't see how it helps understanding what's happening.

planner/scan.go
jsimnz: Two instances right now. For regular type joins, it may need the coresponding <type>_id field. For `_groups` it needs to get the fields within the group selectionSet. Both of these are defined as `mapper.Select`s within a scan context.

request/graphql/parser/filter.go
```diff
@@ -190,6 +190,55 @@ func parseVal(val ast.Value, recurseFn parseFn) (any, error) {
...
+
+func parseFilterFieldsForDescriptionSlice(conditions []any, schema client.SchemaDescription) []client.FieldDescription {
+	fields := make([]client.FieldDescription, 0)
+	for _, v := range conditions {
+		switch cond := v.(type) {
+		case map[string]any:
+			parsedFields := parseFilterFieldsForDescriptionMap(cond, schema)
+			fields = append(fields, parsedFields...)
+		default:
+			// error
```
fredcarle, line 236: todo: This doesn't add anything useful since there are no errors created and this function doesn't return an error either. I would suggest removing the default case.

planner/scan.go
jsimnz: I did however remove the `filter` loop, as that was moved to the mapper after this change, so was able to trim some stuff.

planner/scan.go
jsimnz: note: Cant (at least at the moment) remove the filter loop like we did with the select case, as its a little more complicated with the potential group stuff. Couldnt potentially update some of the mapper to handle things, but it gets into the weeds even more than we already have xD. Let me know, happy to dive into it and make further changes so we can remove the filter loop here.

client/descriptions.go
```diff
@@ -115,6 +103,18 @@ func (sd SchemaDescription) GetFieldKey(fieldName string) uint32 {
 	return uint32(0)
 }
 
+// GetField returns the field of the given name.
+func (sd SchemaDescription) GetField(name string) (FieldDescription, bool) {
+	if !sd.IsEmpty() {
```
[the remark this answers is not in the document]
jsimnz: Many of the various helper funcs defined on the description types that touch the schema description use the same pattern. I likely just copied it from one of those and made the various changes.

Certainly can remove it, but ill mark this as out of scope, as there is other parts in the `descriptions.go` that would need to change, and don't want to mix and match.

db/fetcher/fetcher.go
jsimnz: Implemented now

planner/scan.go
jsimnz: Mentioned in another comment, but posting here for completeness. The `order` check/loop is now removed here.

planner/type_join.go
jsimnz: no idea, thought it was weird when I was making the change. Will remove

db/fetcher/fetcher.go
```diff
@@ -316,7 +451,18 @@ func (df *DocumentFetcher) processKV(kv *core.KeyValue) error {
 	if !df.isReadingDocument {
 		df.isReadingDocument = true
 		df.doc.Reset()
+
+		// re-init doc state
+		// @todo: convert to util method
```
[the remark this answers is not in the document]
jsimnz: Was more of a "nice to have", removed the todo.

planner/mapper/mapper.go
AndrewSisley: > All 3 cases of this if block are covered by the existing tests.

The lines are executed, but not necessarily tested.  Does reverting back to `joinField := condition.Fields[0]` cause the existing tests to fail?  If so, all good here :)

planner/mapper/mapper.go
AndrewSisley: cheers :)

planner/scan.go
AndrewSisley: All the relevant fields on any child stuff, should be present at the parent level - all the needs you just described should be handled by the `case *mapper.Field:` case within this switch

planner/scan.go
AndrewSisley: Nice - thanks John

planner/scan.go
jsimnz: *Should be* as in that is how you believe it currently works, or.. *Should be* I should make the necessary changes to make this work?

request/graphql/parser/filter.go
jsimnz: Updated to actually have an error return, so the entire call stack also has an error path

planner/scan.go
AndrewSisley: Both.  I believe it does this now.  And conceptually it should do that anyway - the whole point of mapper is to produce a data structure that is ready to use by planner so it doesnt have to bother doing any of this stuff and can instead focus on planner stuff.

jsimnz: The mapper doesn't do this at the moment, as it was written when the fetcher would always return all fields, so even the most basic of selects would gurantee we get all dependent fields that any of the filters, orders, childs, groups, etc would ever need, so it didn't need to be overly accurate. All my changes to the mapper have been to add much of this functionality into the mapper.

In any case, the goal of the mapper is to provide a structured way to access information in the query/document using array indexes, hence *mapper* (mapping between maps and slices).

Don't think of this as mapper code in the planner, this collecting of fields would happen even if the mapper didn't exist, and we still used the old map based representation for things. The mapper structures all the necessary information (for the most part) so that it can be consumed in w.e way, but each respective node in the plan graph that consumes mapper structures has to grab its own specific information. The `scanNode` just has to grab the most, as there are various ways that dependent fields can sneak into what a query needs.

IMO this code isn't doing the job of the mapper, we still want the fields for a `_group` for example, within a child select type, because the `groupNode` needs it that way, or the `typeJoinNode`, etc. This is just pulling the now more structured and more explicit info from the `mapper.Select` into the scanNode and fetcher

AndrewSisley: If a field isn't present in mapper.IndexesByName, it shouldn't be fetched/scanned, if it is, it should be.  That really is all the logic that should be here. I think you've even done most/all the work required on the mapper side.

jsimnz: We had previously talked about updating the `IndexesByName` in the `core.DocumentMapping` object that gets passed around, but, for a bunch of different reasons I decided that wasn't the best approach. Most notably, the mapping types are passed around, so there aren't that many unique ones, where as there are several instances where the scanNode/fetcher are using the same collection, but for different reasons, with different fields, filters, etc. So it would be really awkward if the mapping is shared among all the `mapper.Select` types to try to convey this information via the `DocumentMapping`.

The `mapper.Select` are already unique, and contain the various specific filter/order information, so the goal was to just make use of that structure to convey all the information regarding dependent fields.

AndrewSisley: Ah fair enough, I see where you are coming from now.  I think I would probably have still used a shared `IndexesByName` and over-fetched fields that may be required by one Requestable but not another, I think that the maintenance savings  would be worth the performance hit.

Happy as is then :)

jsimnz:
cc: @AndrewSisley @fredcarle 

I believe all your concerns / notes have been addressed. Decided to add a more minified/targeted change to this PR regarding the susbcription tests flakeyness instead of a full implementation in a diff PR.

This PR also indirectly fixes #1582.

Lastly @shahzadlone (mentioned in DMs) reworked the way the alias group by is being converted in the mapper, as it wasn't accounting for all cases. Not, it just uses a direct comparison from the Schema/Field descriptions to determine if its an object type, and will alias accordingly, instead of checking the doc mapper.

AndrewSisley (dismissed):
Looks good to me, but there are still a handful of todos, including an integration test gap that really needs to be filled before merge.

shahzadlone:
> Lastly @shahzadlone (mentioned in DMs) reworked the way the alias group by is being converted in the mapper, as it wasn't accounting for all cases. Not, it just uses a direct comparison from the Schema/Field descriptions to determine if its an object type, and will alias accordingly, instead of checking the doc mapper.

LGTM left some todos

db/fetcher/encoded_doc.go
fredcarle: Is still still relevant? I can't see if there was any changes as a result.

fredcarle (approved):
LGTM. Just one todo related to commented code.

PR #1518, opened 2023-05-18
fix: Handle SIGHUP signal in start command

### LINKED ISSUE #1517: Handle SIGHUP signal

orpheuslummis:
## Relevant issue(s)

Resolves #1517

## Description

Add the handling of SIGHUP signal. SIGHUP (Signal Hang Up) is a signal sent to a process by the operating system when the controlling terminal or session of the process is disconnected or closed.

I'm not sure about this PR. SIGHUP is sometimes used for "reload of configuration".

## How has this been tested?

Manual


fredcarle:
What is the manual testing that you did for this one? I'm curious as to why you think we need to handle that signal. For example, the process exits on closing the terminal even without SIGHUP.

PR #1538, opened 2023-05-26
ci(i): Respect ignore decorator on PR title validation

### LINKED ISSUE #1176: Support ignoring PR commits from changelog using decorators
e.g. `fix(ignore)`

shahzadlone:
## Relevant issue(s)
Resolves #1176

## Description
- Any previous valid label can now also support an additional decorator like so: `(i)` that will help signal that this commit/pr-title can be ignored from change-log.

## How has this been tested?
Unit tests and CI

### DIFF
--- a/tools/scripts/scripts_test.sh
+++ b/tools/scripts/scripts_test.sh
@@ -23,22 +23,33 @@ TestReturnCode() {
 
 # Test the script that is responsible for the validation of pr title.
 readonly T1="./validate-conventional-style.sh"
+
 TestReturnCode "${T1}" 2;
+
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
I like this :) I suggest waiting for others to chime in on this in case they see things differently.

shahzadlone:
> I like this :) I suggest waiting for others to chime in on this in case they see things differently.

Pretty easy to change/reverse/disable this if there is pushback, is safe to merge IMO as we do know we want a way to filter PR merge commits that are internal (this is a step towards that).

Issue #1540, opened 2023-05-26
Explain incorrectly reporting filter for related types.


jsimnz:
A query with a related type, with a filter applied to the related type, is being incorrectly reported.

Eg:
```graphql
query {
  User(friends: {name: { _eq: "John"}}) {
    ...
  }
}
```

Will assign the filter `fiends: {name: {_eq: "John"}}` to the `scanNode` on the `root` of the `typeJoin`. But it should be reported on the `selectNode` above the `typeJoin`.

The actual filter is correctly applies to the `selectNode`, but the way the filters are rendered, misrepresents this.

Related to #1539

PR #1541, opened 2023-05-26
fix: Filter splitting and rendering for related types

### LINKED ISSUE #1539: Filter splitting between scan and select nodes doesn't handle multiple related types
Given the following query
```graphql
query {
  User(filter: { friends: { name: { _eq: "John"}}, address: { street: { _eq: "Example St" }}}) {
    ...
  }
}
```

Where `friends` and `address` are related types within `User`. The planner will incorrectly split the filter and only apply the second of the two ie: `address: { _eq: "Example St" }`. Resulting in an incorrect filter evaluation and subsequent query results.

jsimnz:
## Relevant issue(s)

Resolves #1539, #1540

## Description

Fixes the filter splitting of related types issue. This issue was also directly related to the rendering of these related filters within the explain requests, so this PR covers both.

## Tasks

- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, etc...

## How has this been tested?
Manually, CI

Extensive manual testing with varied queries, extended integration suite with both query and explain tests. Lastly verified against a complex partner query that caused the investigation into the bug originally.

- Ubuntu (WSL2)


tests/integration/explain/default/with_filter_test.go
```diff
@@ -391,3 +391,210 @@ func TestDefaultExplainRequestWithMatchInsideList(t *testing.T) {
...
+				`{
+					"name": "Cornelia Funke",
+					"age": 62,
+					"verified": false
+				}`,
+			},
+		},
+
+		// @shahzadlone: Do I need this?
+		// ExpectedPatterns: []dataMap{basicPattern},
```
shahzadlone, line 557: answer: Yes, you should assert that the order of the plan graph nodes is what we expect.

planner/mapper/targetable.go
```diff
@@ -86,6 +86,52 @@ func NewFilter() *Filter {
...
+				outmap[outkey] = filterObjectToMap(childMapping, subObj)
+			}
+
+		case *Operator:
+			switch keyType.Operation {
+			case "_and", "_or":
+				v := v.([]any)
+				logicMapEntries := make([]any, len(v))
+				for i, item := range v {
+					itemMap := item.(map[connor.FilterKey]any)
```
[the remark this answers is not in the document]
jsimnz: I did have a whole type checks for all the type casting in this function. But there really isn't a way for them, to not be the respective types, unless there is a notable bug, in which case it should be caught much higher in the call stack.

tests/integration/explain/default/with_filter_test.go
```diff
@@ -391,3 +391,210 @@ func TestDefaultExplainRequestWithMatchInsideList(t *testing.T) {
...
+				// bae-b769708d-f552-5c3d-a402-ccfd7ac7fb04
+				`{
+					"name": "Cornelia Funke",
+					"age": 62,
+					"verified": false
+				}`,
+			},
+		},
+
+		// @shahzadlone: Do I need this?
```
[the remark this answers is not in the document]
jsimnz: Will do

tests/integration/explain/default/with_filter_test.go
jsimnz: Will do

jsimnz:
> Was a good catch, overall looks good. I answered the questions that were left as a comment.
> 
> Moreover, you should probably also do the fix on the `group` filter. More specifically in `(n *groupNode) simpleExplain()`.

Good catch, will do

fredcarle (approved):
LGTM. Thanks John!

tests/integration/explain/default/with_filter_test.go
```diff
@@ -391,3 +391,268 @@ func TestDefaultExplainRequestWithMatchInsideList(t *testing.T) {
 
 	runExplainTest(t, test)
 }
+
+var singleTypeJoinPattern = dataMap{
```
shahzadlone, line 395: todo: Just use the previous fixture

shahzadlone (changes requested):

shahzadlone (changes requested):
Left some todos. In addition I would add the tests you are adding to a new file as the test files are split according to different patterns, the patterns of tests you added are more similar to the type join tests, so make a file called ".../default/type_join_with_filter_test.go" and add these two there.

All the todos and the above suggestion have been done here in this commit, feel free to copy: https://github.com/shahzadlone/defradb/commit/e57dff549fe621c80d3b4ea1c8046863b3dcc19b

shahzadlone (approved):
Thanks for doing the suggestions :)

planner/group.go
```diff
@@ -236,10 +236,10 @@ func (n *groupNode) simpleExplain() (map[string]any, error) {
 				childExplainGraph["docKeys"] = nil
 			}
 
-			if c.Filter == nil || c.Filter.ExternalConditions == nil {
+			if c.Filter == nil {
 				childExplainGraph[filterLabel] = nil
 			} else {
-				childExplainGraph[filterLabel] = c.Filter.ExternalConditions
+				childExplainGraph[filterLabel] = c.Filter.ToMap(n.documentMapping)
```
shahzadlone, line 242: question: Does this need a new test as well?

PR #1545, opened 2023-05-29
test(i): Integrate explain tests into test action system

### LINKED ISSUE #1243: Integrate the new explain test setup into the new test action system
2 things happened to our testing frameworks:
- We introduced a new explain testing framework, that we are still not fully migrated to (https://github.com/sourcenetwork/defradb/pull/949).
- We introduced a new general integration testing action based setup in (https://github.com/sourcenetwork/defradb/pull/1089)

The new action based setup is not as rigid as our previous setup which couldn't be extended, so should integrate the new explain testing framework as an action of the new general integration testing setup.

If this is done before the conversion of old explain tests, would help save time having to convert these tests twice.

Side Notes:
for explain(type:execute) tests, introduce a way to omit `Elapsed Time` like metrics that will vary every execution (i.e. can't assert).

shahzadlone:
Resolves #1243

## Description:
- [x] Integrates the new explain test setup with the new testing with action based setup.
- [x] Converts all simple explain tests.
- [x] Converts all default explain tests.
- [x] Converts all execute explain tests.
- [x] Removes the documents for default and simple tests as they aren't worth maintaining (it doesn't effect simple explain testing).
- [x] Fixed a test that was being omitted due to it's file not having the extension "*_test.go".

### For Reviewers:
- I believe commit by commit review should be fairly clean.
- Mostly concerned about the commit(s) with `(CORE)` commit label decorator.
- Commits with `(RENAME)` or `(UNRELATED)` label decorators can be ignored.
- Commits with`(PREP)` label decorators are preparations needed before cleanly doing the integrations/conversions.
- Is there any preference to moving the test cases to the test_cases.go utility files?
- Perhaps soon we can have some sort of `assertion` package to store all the testing assertions as if I used the `utils2.go` file instead of `explain.go` file to put the assertion code, would have ended up with with 1700-1800 lines of code in that file, which I am not a big fan of.

## How has this been tested?
I have tried many permutations of manually inputting incorrect result to ensure tests don't automatically pass.


AndrewSisley (approved):
Looks good to me, good job getting all this sorted!

shahzadlone:
> Looks good to me, good job getting all this sorted!

Speedy, I didn't even ping any reviewers yet haha.

BTW I think the new integration runs change detection on explain (and fails). I forgot if turning it off previously was intentional.

fredcarle (approved):
I like to see more lines removed than added 🙌

AndrewSisley:
> Speedy, I didn't even ping any reviewers yet haha.

I thought you had - I got an email notifiication lol

> BTW I think the new integration runs change detection on explain (and fails). I forgot if turning it off previously was intentional.

They should pass, I was having a look as to why they are failing

shahzadlone:
> I thought you had - I got an email notifiication lol

LOL odd, I don't think I did. But worked for the best :)

> They should pass, I was having a look as to why they are failing

Documented, this is probably because explain tests didn't run with change detection before (were skipped).

Issue #1448, opened 2023-05-03
Add schema migration system to Defra


AndrewSisley:
Part of https://github.com/sourcenetwork/defradb/issues/1002

Using Lens.

Copy-pasted from discord thread `dev-db`.`LensVM on read`(https://discord.com/channels/427944769851752448/1102730749007904798):
> - Lens on local read, and on update (from anywhere, local or P2P), transform outcome to be stored (cached) in datastore
> - Can tolerate discarding P2P updates/creates for unknown schema within release cycle, but must not discard stuff by the time release is cut.  By that point unknown schema version updates need to be held in a hidden (from normal queries) branch
> - Current datastore schema version tracked per document initially, long term we may wish to offer this per field, but not right now  

From discord discussion RE unique secondary indexes:
> - The loss of datastore uniqueness can be tolerated in the short term.
> - Long term, this will likely use the same mechanics as the secondary index backfill stuff, as the problem also exists when defining new unique indexes on existing fields. It could be that if a Lens migration is for an object/field with a unique index it is eagerly evaluated (probably in the background).
> - Something worth noting, that I had forgotten is that non-unique secondary indexes are being implemented first, it could be that this will not even be a problem within 0.6, depending on when unique secondary indexes are implemented

From discord discussion RE wasm module persistence:
> We could just rely on the user/admins preserving wherever they exist at time of migration registration (1).  But we could also copy them somewhere defra specific (2), or store them as blobs within a datastore (3). 
> - (1) is fine short term, and is likely the easiest, it is not a blocker for release
> - (2) is preferred short term if low cost
> - going from (1) to (2) should be fairly cheap, so if (1) is implemented first and everyone hates it, it shouldnt block merge and moving to (2) can likely be done immediately after within the release if required
> - (3) is Andy's and John's preferred long term solution, but:
> - (2) and (3) actually share the same public interfaces, and so the differences between the two can really be seen as implementation details

Personal/wip thoughts:
- `wasmer.NewModule(store, content)` is slow (40-50ms), and should not be done in user/query time.  This will mean they need to be cached somewhere.
- Lens module instances are not threadsafe, and will probably need some worker-like channel stuff to ensure that db actions on different threads do not concurrently access the same lens module.
- A Lens module instance should not be shared between multiple defra requests/actions/contexts at the same time.  This is to protect against the leaking of state between contexts.  This is particularly important the proposed Lens enumerable stuff, but is otherwise just a very good thing to have anyway.
[...]

PR #1561, opened 2023-06-05
feat: Track schema version history

### LINKED ISSUE #1556: Track schema version history
Part of https://github.com/sourcenetwork/defradb/issues/1002

At the moment users need to specify the destination and source schema versions for their migrations.  This is error prone, can create divergent/overlapping migrations and is unnecessary hassle for them.

They should only need to provide the destination schema version. 

To do this we can create an index, and move the existing schema index (which stores the schema json representation), roughly as follows:
```
// this already exists, without the `/v` (CollectionSchemaVersionKey)
/collection/version/v/[newerSchemaID] => [jsonRepresentation] 
// new key tracking history
/collection/version/p/[newerSchemaID] => [priorSchemaVersionID]
```
by including the newerSchemaID in the key and excluding the priorSchemaVersionID we can ensure that duplicates are not created. `p` and `v` are before the schema ids to allow for more efficient scans of one or the other.  It may be desirable to include `[schemaID]` in the history key to allow for more efficient queries of the history when migrating a record across multiple versions i.e.: `/collection/version/p/[schemaID]/[newerSchemaID] => [priorSchemaVersionID]`.

AndrewSisley:
## Relevant issue(s)

Resolves #1556

## Description

Tracks schema version history via dedicated index.

This is required by https://github.com/sourcenetwork/defradb/issues/1448 as the fetcher must know the path through the schema version history if migrating across multiple schema versions.  For example if a document is at schema version 1, and it needs to return it at schema version 3, it needs to know that it has to migrate to schema version 2 first.

The structure should allow for manual correction of divergent schemas, as the history can be overwritten with this change in a semi-destructive way.  Migrations will still work up from orphaned schema versions, but not down. This is an unlikely edge case, and I feel like this is a good compromise in the short-medium term, instead of over engineering to fully handle divergent schema migrations. 

~~I think I'm going to leave this as a draft for now, as I'd have to go out of my way (and add more features) in order to test this properly as it is internal only and the new keys are never read.  I would like feedback nowish though as the Defra-Len PR will be based off this work (and will properly test it).~~ I am now confident that this is required, and am looking to merge it. 

The change detector should hopefully fail, I wont fix (i.e. document) until shortly before merge.

### DIFF
--- a/core/key.go
+++ b/core/key.go
@@ -113,8 +113,8 @@ var _ Key = (*CollectionSchemaVersionKey)(nil)
 // If a SchemaHistoryKey does not exist for a given SchemaVersionID it means
 // that that SchemaVersionID is for the latest version.
 type SchemaHistoryKey struct {
-	SchemaID             string
-	PriorSchemaVersionId string
+	SchemaID                string
+	PreviousSchemaVersionId string
 }
 
 var _ Key = (*SchemaHistoryKey)(nil)
@@ -223,10 +223,10 @@ func NewCollectionSchemaVersionKey(schemaVersionId string) CollectionSchemaVersi
 	return CollectionSchemaVersionKey{SchemaVersionId: schemaVersionId}
 }
 
-func NewSchemaHistoryKey(schemaId string, priorSchemaVersionId string) SchemaHistoryKey {
+func NewSchemaHistoryKey(schemaId string, previousSchemaVersionId string) SchemaHistoryKey {
 	return SchemaHistoryKey{
-		SchemaID:             schemaId,
-		PriorSchemaVersionId: priorSchemaVersionId,
+		SchemaID:                schemaId,
+		PreviousSchemaVersionId: previousSchemaVersionId,
 	}
 }
 
@@ -428,8 +428,8 @@ func (k SchemaHistoryKey) ToString() string {
 		result = result + "/" + k.SchemaID
 	}
 
-	if k.PriorSchemaVersionId != "" {
-		result = result + "/" + k.PriorSchemaVersionId
+	if k.PreviousSchemaVersionId != "" {
+		result = result + "/" + k.PreviousSchemaVersionId
 	}
 
 	return result
--- a/core/key.go
+++ b/core/key.go
@@ -384,7 +384,7 @@ func (k CollectionSchemaKey) ToDS() ds.Key {
 }
 
 func (k CollectionSchemaVersionKey) ToString() string {
-	result := COLLECTION_SCHEMA_VERSION
+	result := COLLECTION_SCHEMA_VERSION + "/" + string(ValueKey)
 
 	if k.SchemaVersionId != "" {
 		result = result + "/" + k.SchemaVersionId
--- a/core/key.go
+++ b/core/key.go
@@ -41,13 +41,14 @@ const (
 )
 
 const (
-	COLLECTION                = "/collection/names"
-	COLLECTION_SCHEMA         = "/collection/schema"
-	COLLECTION_SCHEMA_VERSION = "/collection/version"
-	SEQ                       = "/seq"
-	PRIMARY_KEY               = "/pk"
-	REPLICATOR                = "/replicator/id"
-	P2P_COLLECTION            = "/p2p/collection"
+	COLLECTION                        = "/collection/names"
+	COLLECTION_SCHEMA                 = "/collection/schema"
+	COLLECTION_SCHEMA_VERSION         = "/collection/version/v"
+	COLLECTION_SCHEMA_VERSION_HISTORY = "/collection/version/h"
+	SEQ                               = "/seq"
+	PRIMARY_KEY                       = "/pk"
+	REPLICATOR                        = "/replicator/id"
+	P2P_COLLECTION                    = "/p2p/collection"
 )
 
 // Key is an interface that represents a key in the database.
@@ -404,7 +405,7 @@ func (k CollectionSchemaKey) ToDS() ds.Key {
 }
 
 func (k CollectionSchemaVersionKey) ToString() string {
-	result := COLLECTION_SCHEMA_VERSION + "/" + string(ValueKey)
+	result := COLLECTION_SCHEMA_VERSION
 
 	if k.SchemaVersionId != "" {
 		result = result + "/" + k.SchemaVersionId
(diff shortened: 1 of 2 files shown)


core/key.go
```diff
@@ -384,7 +404,7 @@ func (k CollectionSchemaKey) ToDS() ds.Key {
 }
 
 func (k CollectionSchemaVersionKey) ToString() string {
-	result := COLLECTION_SCHEMA_VERSION
+	result := COLLECTION_SCHEMA_VERSION + "/" + string(ValueKey)
```
AndrewSisley, line 407: Note: I am reusing the `InstanceType` values (directly) here, I like the consistency and think it fits, but I can also see how it may be seen as pointless coupling and abuse - do let me know :)

fredcarle: Can't you explain why you want to use the value key in the path? At first glance it seems unnecessary.

db/collection.go
```diff
@@ -236,6 +236,7 @@ func (db *db) updateCollection(
 	if err != nil {
 		return nil, err
 	}
+	priorSchemaVersionID := desc.Schema.VersionID
```
fredcarle, line 239: suggestion: This could probably be defined closer to where it is used bellow.

AndrewSisley: `desc.Schema.VersionID` gets overwritten two lines down, it needs to be here (or one line down)

fredcarle: Oh yes I missed that sorry.

core/key.go
AndrewSisley: Not a problem, will add a comment (and on the history key which does the same)

core/key.go
```diff
@@ -106,6 +106,19 @@ type CollectionSchemaVersionKey struct {
...
 
+// SchemaHistoryKey holds the pathway through the schema version history for
+// any given schema.
+//
+// The keys point to the schema version id of the next version of the schema.
+// If a SchemaHistoryKey does not exist for a given SchemaVersionID it means
+// that that SchemaVersionID is for the latest version.
+type SchemaHistoryKey struct {
+	SchemaID             string
+	PriorSchemaVersionId string
```
jsimnz, line 117: nitpick: Slight preference for `Previous` vs `Prior`, I feel its a little more common and is often used in the programming context in regards to `next` / `prev` (previous).

core/key.go
jsimnz: Unless there's a necessity for this I don't see, I would suggestion removing it.

core/key.go
```diff
@@ -401,6 +421,28 @@ func (k CollectionSchemaVersionKey) ToDS() ds.Key {
 	return ds.NewKey(k.ToString())
 }
 
+func (k SchemaHistoryKey) ToString() string {
+	result := COLLECTION_SCHEMA_VERSION + "/" + string(PriorityKey)
```
jsimnz, line 425: question: Same here as the above `value` suffix.

db/collection.go
```diff
@@ -264,6 +265,12 @@ func (db *db) updateCollection(
 		return nil, err
 	}
 
+	schemaVersionHistoryKey := core.NewSchemaHistoryKey(desc.Schema.SchemaID, schemaVersionID)
+	err = txn.Systemstore().Put(ctx, schemaVersionHistoryKey.ToDS(), []byte(priorSchemaVersionID))
```
jsimnz, line 269: question: This is somewhat confusing. `SchemaHistoryKey` is defined with `SchemaID` and `PriorSchemaVersion`, but here you are building the `NewSchemaHistoryKey` with the *current* `schemaVersionID`, and then using this key to point to a value, and that value is the `priorSchemaVersionID`. 

Feels like there is some inconsistent naming going on. Feel free to correct me.

jsimnz (changes requested):
Only one confusing item. However, I don't really understand the use of value/priority here, as it hasn't been used in the system store, and the names "value" and "priority" A) don't really make sense even if we do want to include them B) Why do we need to include them?

AndrewSisley:
> I don't really understand the use of value/priority here, as it hasn't been used in the system store, and the names "value" and "priority" A) don't really make sense even if we do want to include them B) Why do we need to include them?

It is so that we can distinguish between the two keys, as otherwise they would have the same prefix, and the same structure.

The alternative to using them (or other identifiers) would to not store one of them under `/collection/version`, but I think that there is the most sensible location for both.

db/collection.go
AndrewSisley: SchemaHistoryKey is an index of the next version, v1 points to v2 etc. This is documented on the key, and I cannot think of a better name. Am very open to alternatives though.

core/key.go
AndrewSisley: Can do, not a problem

- [ ] Rename prior to previous

core/key.go
AndrewSisley: note: Discussion and resolution will be found against the other comment and applied here.

core/key.go
AndrewSisley: I thought about this a bit more, I'm going to split the const in to `/collection/version/v` and `/collection/version/h` (dropping the runtime concat of the InstanceType).  It has tripped two readers up slightly, and you cant see the reason for it reading each ToString implementation in isolation (without comment).  As the consts will be declared next to each other they should be pretty obvious as to why they exist too.

- [ ] Split const, rm use of InstanceType

core/key.go
```diff
@@ -106,6 +107,19 @@ type CollectionSchemaVersionKey struct {
 
 var _ Key = (*CollectionSchemaVersionKey)(nil)
 
+// SchemaHistoryKey holds the pathway through the schema version history for
+// any given schema.
+//
+// The keys point to the schema version id of the next version of the schema.
+// If a SchemaHistoryKey does not exist for a given SchemaVersionID it means
+// that that SchemaVersionID is for the latest version.
```
islamaliev, line 115: I find the comment somewhat confusing. It says "schema version id of the next version" but the member's name is `PreviousSchemaVersionId` (btw use "ID" please, not "Id"). Where is this next version? Does `SchemaID` hold this information? Then maybe there is some inconsistency in using terms schema id and schema version id?

AndrewSisley: - [ ] Fix typo `keys point` => `key points`

- [ ] Capitalize `ID`...

> Where is this next version?

The key points to it. The next version is the value component of the store key-value item.

> PreviousSchemaVersionId

Would you prefer just `SchemaVersionID`?

islamaliev: ~~Not sure. I'm still confused with the comment. It says "next" while the value that is stored is previous~~

oh, I see it's just a key in KV storage that points to the next version. I overlooked this part.
Never mind mind comment then.

AndrewSisley:
I've closed this PR.  Feedback has been applied, and the branch cherry-picked into the main defra-lens PR here: https://github.com/sourcenetwork/defradb/pull/1564

PR #1563, opened 2023-06-08
feat: Add ability to explain all nodes (debug)

### LINKED ISSUE #948: Explain Request - Debug
Sub Task of https://github.com/sourcenetwork/defradb/issues/35

Spec: https://source.getoutline.com/doc/explain-queries-jtaGX0L995

shahzadlone:
Resolves #948 

## Description
Implements a new type of explain called `debug` which dumps all plan nodes (even non-explainable nodes) as a graph (has no attributes).

Usage is similar to other explain types, but would just do `@explain(type: debug)` instead of `@explain(type: simple)` or `@explain(type: execute)`.

Example Request:
```
query @explain(type: debug) {
	Author (groupBy: [age]) {
		age
		_group {
			name
		}
	}
}
```

Response:
```
{
	"data": [
		"explain": {
			"selectTopNode": {
				"groupNode": {
					"selectNode": {
						"pipeNode": {
							"scanNode": {}
						}
					}
				}
			}
		}
	]
}
```

## How has this been tested?
Integration tests

- WSL2 (Manjaro)


AndrewSisley:
> > thought: It feels quite odd to me that `debug` explain outputs less stuff than simple explain.  Why would we not want to output the attributes for the nodes that have them?
> 
> We already have access to the attributes with the simple explain call. The debug explain (atleast the first iteration of it) should visualize the full plan graph nodes. i.e. the `pipeNode` and `multiScanNode`.
> 
> However this makes me think that similar to how we were gunna add an option to `execute` explain like `showSimpleAttributes` to enable showing simple explain attributes for execute explain nodes, we can in future do the same thing for `debug` as well.

I'm not sure you are understanding me.  Typically `debug` provides the same information as `simple`/`info`/`default`/etc, **and** more, additional info (never losing the `simple` info in the `debug`).  It would also annoy me as a user if I couldnt get the full graph with attributes from a single explain call.

shahzadlone:
> Typically `debug` provides the same information as `simple`/`info`/`default`/etc, **and** more, additional info (never losing the `simple` info in the `debug`).

Is "simple" an actual log level? I have never seen that, if it is then I can see the confusion. The different explain types aren't log/verbosity levels, they do different things. For example `execute` explain will return the execution information, simple will return the attributes and only the explain nodes. Upon thinking further I would be okay to name this to something else, maybe `type: internal`, this explain should be very rarely used by non-internal-devs as the graph includes nodes that are too into the weeds of defradb.

> It would also annoy me as a user if I couldnt get the full graph with attributes from a single explain call.

That will be possible though for both `execute` and this explain with `showSimpleAttributes`.

shahzadlone:
@AndrewSisley made a ticket to reflect our standup discussion: https://github.com/sourcenetwork/defradb/issues/1570

planner/explain.go
```diff
@@ -56,14 +56,114 @@ const (
...
 	filterLabel         = "filter"
-	keysLabel           = "_keys"
 	idsLabel            = "ids"
+	joinRootLabel       = "root"
+	joinSubTypeLabel    = "subType"
+	keysLabel           = "_keys"
 	limitLabel          = "limit"
 	offsetLabel         = "offset"
 	sourcesLabel        = "sources"
 	spansLabel          = "spans"
```
AndrewSisley, line 66: suggestion (out of scope): These are all public, and I think they should live in `client/request/explain.go` - might be worth adding a ticket if you agree.

Same also goes for the attribute tags and node names (if they are in planner atm)

AndrewSisley (approved):
Is a mad amount of tests that you have managed to write for this, I skimmed through them and they all appeared sensible to have - thanks a bunch for adding them all :)

Prod code looked good too, minus the feature-change requested RE type that we discussed in the standup.  Just added an out of scope suggestion for the future that I spotted whilst reading.

planner/explain.go
shahzadlone: I appreciate this suggestion! Would be much nicer. Will open a ticket.

shahzadlone: ticket: https://github.com/sourcenetwork/defradb/issues/1571

islamaliev:
bug bash result:
ran local defradb node, create different collection with relationships.
Used Altair GraphQL client to query many different types of queries and mutations (filters, groups, orders, limits and others).
For all queries got a corresponding explain graph

PR #1575, opened 2023-06-13
feat: Playground

### LINKED ISSUE #884: Bundled graphql playground in defra binary.
Bundle the GraphiQL editor as part of the binary (static linked assets), so that we can access `http(s)://<defra_host:port>/playground` or something like that. 

https://github.com/graphql/graphiql

Therefore we can have a consistent experience readily bundled with defra accessible from the browser.

Overtime we can customize the behavior of the playground, especially for testing with a MVP version of digital signatures/state proofs that can transparently generate and manage signatures/proofs.

Future ideas:
- include schema editor/manager
- policy editor/manager
- peer insights
- metrics
- etc

nasdf:
## Relevant issue(s)

Resolves #884 

## Description

This PR adds a web based playground to DefraDB.

To use the embedded playground run the following.

```
# build playground
make deps:playground

# enable embedded playground
export BUILD_TAGS=playground

make start
# or 
make build
```

The playground is served from the same port as the API: `http://localhost:9181`

- [x] View & query DefraDB types
- [x] Create & patch schemas
- [x] Relational field support
- [x] Embed playground in binary

localhost_5173_

## Tasks

## How has this been tested?

Testing was done manually.

- MacOS


api/http/router.go
```diff
@@ -35,38 +34,22 @@ const (
...
+	router.Get(BlocksPath+"/{cid}", getBlockHandler)
+	router.Get(GraphQLPath, execGQLHandler)
+	router.Post(GraphQLPath, execGQLHandler)
+	router.Get(SchemaPath, listSchemaHandler)
+	router.Post(SchemaPath, loadSchemaHandler)
+	router.Patch(SchemaPath, patchSchemaHandler)
+	router.Post(IndexPath, createIndexHandler)
+	router.Delete(IndexPath, dropIndexHandler)
+	router.Get(IndexPath, listIndexHandler)
+	router.Get(PeerIDPath, peerIDHandler)
```
fredcarle, line 52: thought: I'm not a big fan of a global var router and setting the routes on package init. I think it would be better for future maintenance (and other beneficial reasons) to keep this under the `setRoutes` function. Although you can simply return the router.

That being said, I like how you moved the handle method to a middleware and that it avoids wrapping the handlers in the `handle` method.

fredcarle: `playground.go:init` can also be moved inside the `setRoutes` function.

api/http/playground.go
```diff
@@ -0,0 +1,18 @@
+//go:build playground
```
shahzadlone, line 1: todo: Should add a copyright header (I am also wondering why the linter didn't complain about the copyright header)

playground/playground.go
```diff
@@ -0,0 +1,10 @@
+//go:build playground
```
shahzadlone, line 1: todo: Missing header

playground/src/components/Plugin.tsx
```diff
@@ -0,0 +1,47 @@
+import { useQuery } from '@tanstack/react-query'
```
shahzadlone, line 1: todo: Header

playground/src/components/SchemaLoadForm.tsx
```diff
@@ -0,0 +1,71 @@
+import { useState, useEffect } from 'react'
```
shahzadlone, line 1: todo: Header

playground/src/components/SchemaPatchForm.tsx
```diff
@@ -0,0 +1,107 @@
+import { useState } from 'react'
```
shahzadlone, line 1: todo: Header

playground/src/index.css
```diff
@@ -0,0 +1,93 @@
+html, body, #root {
```
shahzadlone, line 1: todo: Header?

api/http/router.go
nasdf: I wanted to keep the `setRoutes` method but I couldn't think of another way to do build time route registration. I'm using the `playground.go:init` function because the playground can be enabled/disabled using the build tag. I'm happy to move everything back if we can think of an alternative that works with build tags.

api/http/playground.go
nasdf: Maybe due to the build tag here? I'm not exactly sure either.

api/http/router.go
nasdf: Thanks! My brain was stuck on that for a while. I think I came up with a good solution. The route registration is moved back to `setRoutes` and the playground handler is set at build time via `init`. I made sure it works when not including the playground as well.

PR #1603, opened 2023-06-29
test: Rework transaction test framework capabilities

### LINKED ISSUE #1602: Improve test action txn capabilities
Instead of a dedicated txn action (like we currently have), we can instead follow a similar pattern to node ids where any action can specify a transaction id.

AndrewSisley:
## Relevant issue(s)

Resolves #1602

## Description

Instead of a dedicated txn action (like we currently have), we can instead follow a similar pattern to node ids where any action can specify a transaction id. I would like to use this on other actions, most immediately for Lens testing, although it will likely be very useful elsewhere too.


islamaliev (approved):
LGTM

PR #1621, opened 2023-07-11
fix: Avoid duplicated requestable fields

### LINKED ISSUE #1615: resolveFilterDependencies() in mapper.go produces duplicated fields
`resolveFilterDependencies` in `planner/mapper/mapper.go` returns duplicated fields if query with filter requests only top-level fields.

For example, in `tests/integration/query/ont_to_many/with_filter_test.go:TestQueryOneToManyWithNumericGreaterThanFilterOnParent` test with nothing changed, i.e. with this query:
```graphql
query {
	Author(filter: {age: {_gt: 63}}) {
		name
		age
		published {
			name
			rating
		}
	}
}
```
it will produce a result as expected, i.e. for Author collection:

Image

But with this query:
```graphql
query {
	Author(filter: {age: {_gt: 63}}) {
		name
		age
	}
}
```
there will be a bunch of duplicates:

Image

islamaliev:
## Relevant issue(s)

Resolves #1615

## Description

Makes sure the requestable fields aren't duplicated 

## Tasks

## How has this been tested?

Intergration tests

- MacOS

### DIFF
--- a/planner/mapper/mapper.go
+++ b/planner/mapper/mapper.go
@@ -795,6 +795,7 @@ func resolveInnerFilterDependencies(
 ) ([]Requestable, error) {
 	newFields := []Requestable{}
 
+sourceLoop:
 	for key := range source {
 		if strings.HasPrefix(key, "_") && key != request.KeyFieldName {
 			continue
@@ -844,15 +845,10 @@ func resolveInnerFilterDependencies(
 			// If the key index is outside the bounds of the child mapping array, then
 			// this is not a relation/join and we can add it to the fields and
 			// continue (no child props to process)
-			isDuplicate := false
 			for _, field := range existingFields {
 				if field.GetIndex() == keyIndex {
-					isDuplicate = true
-					break
+					continue sourceLoop
 				}
-			}
-			if isDuplicate {
-				continue
 			}
 			newFields = append(existingFields, &Field{
 				Index: keyIndex,
--- a/planner/mapper/mapper.go
+++ b/planner/mapper/mapper.go
@@ -844,6 +844,16 @@ func resolveInnerFilterDependencies(
 			// If the key index is outside the bounds of the child mapping array, then
 			// this is not a relation/join and we can add it to the fields and
 			// continue (no child props to process)
+			isDuplicate := false
+			for _, field := range existingFields {
+				if field.GetIndex() == keyIndex {
+					isDuplicate = true
+					break
+				}
+			}
+			if isDuplicate {
+				continue
+			}
 			newFields = append(existingFields, &Field{
 				Index: keyIndex,
 				Name:  key,


islamaliev:
> Does this impact publicly visible behaviour? If so please add tests documenting this change.

So far I didn't see any public impact. I'm not sure why. The collected fields are used internally and I assume that so far the code didn't rely on uniqueness of the fields in the list.

AndrewSisley:
> > Does this impact publicly visible behaviour? If so please add tests documenting this change.
> 
> So far I didn't see any public impact. I'm not sure why. The collected fields are used internally and I assume that so far the code didn't rely on uniqueness of the fields in the list.

Then why make the change?

islamaliev:
> > > Does this impact publicly visible behaviour? If so please add tests documenting this change.
> > 
> > 
> > So far I didn't see any public impact. I'm not sure why. The collected fields are used internally and I assume that so far the code didn't rely on uniqueness of the fields in the list.
> 
> Then why make the change?

I could have added this code to the part related to secondary indexes, but I though that it would be beneficial for us to have a generic solution. Or you are asking why a separate PR? I though it was quite related to my stuff so that the score is limited. I can include this fix in secondary indexes task. Not a problem.

planner/mapper/mapper.go
```diff
@@ -844,6 +844,16 @@ func resolveInnerFilterDependencies(
 			// If the key index is outside the bounds of the child mapping array, then
 			// this is not a relation/join and we can add it to the fields and
 			// continue (no child props to process)
+			isDuplicate := false
+			for _, field := range existingFields {
+				if field.GetIndex() == keyIndex {
+					isDuplicate = true
```
AndrewSisley, line 850: suggestion: You can drop the `isDuplicate` var by naming the outer loop and continuing here:

```
sourceLoop:
    for key := range source {
         ...
         if field.GetIndex() == keyIndex {
             continue sourceLoop
        }
```

AndrewSisley (approved):
> I could have added this code to the part related to secondary indexes, but I though that it would be beneficial for us to have a generic solution. Or you are asking why a separate PR? I though it was quite related to my stuff so that the score is limited. I can include this fix in secondary indexes task. Not a problem.

I was asking why you are making the change at all, what it does, and what you have done to check that it does what you want it to do :)  In the absence of tests it is quite hard to tell.

Approving with a minor suggestion, and assuming that you have checked it has the result you think it has (going by the linked ticket).  I would prefer that there were tests for this, but going by your previous answers it looks like secondary indexes is not working because of this, and integration tests for secondary indexes will flag any regressions here?

planner/mapper/mapper.go
islamaliev: yeah, I though about it but the label would be way up there... Anyway I will switch to use the label

planner/mapper/mapper.go
```diff
@@ -844,6 +845,11 @@ func resolveInnerFilterDependencies(
 			// If the key index is outside the bounds of the child mapping array, then
 			// this is not a relation/join and we can add it to the fields and
 			// continue (no child props to process)
+			for _, field := range existingFields {
```
fredcarle, line 848: question: You are removing duplicate fields that didn't impact query results at the expense of an extra for loop for every field. Could you explain why you had to do this? Or guide me towards the explanation if it exists already.

AndrewSisley: The linked issue does cover this (#1615), but it does require some imagination in order to connect the dots.

fredcarle: It doesn't. It show state of variables. I understand that there was duplicates and doing this removes the duplicates. I'd like to know if it's just for cleanliness of if it has a direct impact on something else which could help support the extra for loop.

islamaliev: I already replied to Andy on this matter. Here is my quote:
> So far I didn't see any public impact. I'm not sure why. The collected fields are used internally and I assume that so far the code didn't rely on uniqueness of the fields in the list.
I noticed this by observing weird behaviour while working on indexes.

PR #1625, opened 2023-07-11
feat: Schema list API

### LINKED ISSUE #1624: Schema list API
We could use an API for returning user created schemas from DefraDB.
This will make it easier for users to export and update schemas.
The API should return all collections and fields excluding auto generated ones.

nasdf:
## Relevant issue(s)

Resolves #1624 

## Description

This PR adds an API for listing schemas.

### HTTP

`curl http://localhost:9181/api/v0/schema/list`

#### Output

```
{
  "data": {
    "collections": [
      {
        "name": "User",
        "id": "bafkreibpnvkvjqvg4skzlijka5xe63zeu74ivcjwd76q7yi65jdhwqhske",
        "fields": [
          {
            "id": "1",
            "name": "age",
            "kind": "Int"
          },
          {
            "id": "2",
            "name": "name",
            "kind": "String"
          },
          {
            "id": "3",
            "name": "points",
            "kind": "Float"
          },
          {
            "id": "4",
            "name": "verified",
            "kind": "Boolean"
          }
        ]
      }
    ],
    "result": "success"
  }
}
```

### CLI

`defradb client schema list`

#### Output

```
type User {
    name: String 
    age: Int 
    verified: Boolean 
    points: Float
}
```

## Tasks

## How has this been tested?

Manual testing

- MacOS


api/http/handlerfuncs.go
```diff
@@ -155,9 +156,64 @@ func execGQLHandler(rw http.ResponseWriter, req *http.Request) {
...
+	Kind string `json:"kind"`
+}
+
+type collectionResponse struct {
+	Name   string          `json:"name"`
+	ID     string          `json:"id"`
+	Fields []fieldResponse `json:"fields,omitempty"`
+}
+
+func listSchemaHandler(rw http.ResponseWriter, req *http.Request) {
```
fredcarle, line 171: todo: I would be really nice to have a unit test for this handler to see what the expected output is and that we catch future changes to the schema structure. Ideally the error paths would also be tested. You can base yourself off of the other unit test for this.

api/http/handlerfuncs.go
```diff
@@ -155,9 +156,64 @@ func execGQLHandler(rw http.ResponseWriter, req *http.Request) {
...
+			Name:   col.Name(),
+			ID:     col.SchemaID(),
+			Fields: fields,
+		}
+	}
+
+	sendJSON(
+		req.Context(),
+		rw,
+		simpleDataResponse("result", "success", "collections", colResp),
```
fredcarle, line 214: suggestion: You can probably remove the `result` and `success` for the response as those are superfluous in this case.

api/http/router.go
```diff
@@ -30,6 +30,7 @@ const (
 	DumpPath        string = versionedAPIPath + "/debug/dump"
 	BlocksPath      string = versionedAPIPath + "/blocks"
 	GraphQLPath     string = versionedAPIPath + "/graphql"
+	SchemaListPath  string = versionedAPIPath + "/schema/list"
```
fredcarle, line 33: suggestion: We started changing out paths to be a bit more RESTful with the introduction of the `index` handlers. If you don't mind, please convert the 3 schema paths to a single `SchemaPath string = versionedAPIPath + "/schema"` and use the relevant http verbs in the `setRoutes` method bellow (Get, Post, Patch).

cli/schema_list.go
```diff
@@ -0,0 +1,84 @@
...
+	} `json:"data"`
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+func MakeSchemaListCommand(cfg *config.Config) *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "list",
+		Short: "List schema types from DefraDB",
```
fredcarle, line 46: suggestion: `List schema types with their respective fields`. No a strong opinion but I though it would be a bit more descriptive of what it does.

fredcarle (changes requested):
This is an nice functionality. I like the improvements from what you had for it on the playground PR :)

I do have a few things I'd like to address as you'll see bellow. Nothing major though and once they're resolved we should be good to merge.

api/http/handlerfuncs.go
nasdf: Good call. I added tests for the error path as well as the happy path.

api/http/router.go
nasdf: I simplified the paths like you mentioned. Do we need to make an announcement when changing the API?

fredcarle: At this stage I don't think we do as most users probably use the cli for this but the python sdk might need to be updated.

fredcarle (approved):
Good job Keenan :) Thanks for the changes.

islamaliev:
bug bash result:
ran locally defradb nodes with different `rootdir` with different data. Ran `list` CLI command for them and observed the list of collections. Also added a new collection to make sure the result is not cached.

PR #1630, opened 2023-07-12
feat: Add simple data import and export

### LINKED ISSUE #1544: Simple version migration import/export
Export the data in a way that it can be imported without taking the DAG into account. This is to help partners update from one version of Defra to an other during development.

fredcarle:
## Relevant issue(s)

Resolves #1544

## Description

This PR adds import and export functionality to the http api and cli. It can export to json to reduce the potential file size. At this stage csv output was not implemented as it would require extensive type casting (everything in csv is a string) on both writing and reading from the csv.

## Tasks

## How has this been tested?

make test and manual testing

- MacOS


cli/db_export.go
```diff
@@ -0,0 +1,158 @@
...
+
+	httpapi "github.com/sourcenetwork/defradb/api/http"
+	"github.com/sourcenetwork/defradb/config"
+	"github.com/sourcenetwork/defradb/logging"
+)
+
+const cborFileType = "cbor"
+const jsonFileType = "json"
+
+func MakeDBExportCommand(cfg *config.Config) *cobra.Command {
```
AndrewSisley, line 32: todo: This looks awfully expensive in terms of memory, With the results being fully held in memory from the http call, unmarshalled, then remarshalled, before being written to file. For all collections at the same time in the database.

Worse, the entire database is copied into memory server-side too, and could do quite a lot of damage.

Whilst we might be okay tolerating this for a short time, I think the command documentation should come with a very large warning. And another should be added to the func documentation of the http export command.

AndrewSisley (changes requested):
I think the documentation at least needs to change, documenting the memory cost of this operation, if the team is happy with the current implementation in the short-term.

api/http/handlerfuncs_export.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"net/http"
+
+	"github.com/fxamacker/cbor/v2"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+func exportHandler(rw http.ResponseWriter, req *http.Request) {
```
AndrewSisley, line 21: suggestion: It does feel a bit odd that this is only exposed over the http client, and not via the `client` package too.  It might be nicer if the bulk of the code was pulled out into its own package too (regardless), for greater visibility and to make it easier to expose via other locations at a later date should we want to.

api/http/handlerfuncs_export.go
```diff
@@ -0,0 +1,188 @@
...
+	getDocs := func(rw http.ResponseWriter, req *http.Request, col client.Collection) (docs []map[string]any, ok bool) {
+		keysCh, err := col.GetAllDocKeys(req.Context())
+		if err != nil {
+			handleErr(req.Context(), rw, err, http.StatusInternalServerError)
+			return nil, false
+		}
+
+		for key := range keysCh {
+			doc, err := col.Get(req.Context(), key.Key, showDeleted)
+			if err != nil {
```
AndrewSisley, line 43: todo: This looks like it might error should the document have been deleted (with `showDeleted == false`) between here and the retrieval of the key.  It also looks as if it will error out on any transaction conflict.

I don't think either of those things should result in the entire export being cancelled. Such behaviour might make it highly impractical to ever export some production databases.

cli/db_export.go
fredcarle: I was going to bring this up during standup. There is a simple solution to this which is to let the defra instance write/read the file directly like postgres and other DBs do.

api/http/handlerfuncs_export.go
fredcarle: I'm ok with moving it to the client package but I don't think its worth having it's own package.

api/http/handlerfuncs_export.go
fredcarle: I can change it to use a single transaction for the whole export which would be a better approach anyways.

cli/db_export.go
AndrewSisley: Just noting that the CLI or DB behaviour doesnt actually need to change to not load everything into memory. Network stuff might need to change though if you want the total IM size to be less than 1 DB.

api/http/handlerfuncs_export.go
AndrewSisley: I think a single transaction would be even worse, you'd be as good as locking the whole database for the duration of the export, and any transaction already in progress would result in a conflict (also cancelling the whole export)

fredcarle: Why would the export be canceled? We can't get a transaction conflict on read. Thinking about it more, why would a deleted document cause any problem if the getting of the documents is done on a single transaction?

cli/db_export.go
fredcarle: We could stream every document across one at a time.

api/http/handlerfuncs_export.go
AndrewSisley: > We can't get a transaction conflict on read.

We never get them on read? Only update?  Is not a problem them, from that perspective.  The very large transaction might cause some memory problems of its own though - worth testing that.

Delete would never be an issue with a single txn.

Would also likely run into the problem that a single txn would complicate the solving of the memory issue however, as we currently have no means to preserve a txn between remove calls (i.e. the whole server side op would be forced to copy everything into memory until/unless we implement that too).

shahzadlone (commented):
Is a nice first step and I think provides quality of life improvements for users.

I do have a slight nitpick for the terminology of `backup/recover` over `import/export` but I think that might be just me :P

suggestion: Would be nice to perhaps include a simple cli usage example in the `README.md`

fredcarle:
Note: The tests will fail as I didn't update them for the most recent changes. I want to ensure that we agree with the approach before I put time into the tests which I'll do as soon as consensus is reached.

The change moves the handling of the import and export to the `db` package as it's a `db` specific action. It can be done programatically or via CLI. The new approach builds the export file document by document and reads the import file document by document. As such, the memory usage should remain quite low. The obvious downside is that we will only support JSON import/export for now but I feel like that is less important than the benefit of the reduction in memory usage.

Let me know what you guys think :)

db/backup.go
nasdf, line 1: I think the streaming approach here is solid, but I think some things could be simplified with the built in `json.Decoder`. Specifically the `Decode` func see the example here https://pkg.go.dev/encoding/json#Decoder.Decode

fredcarle: That's a very nice suggestion. I'll try to use that approach.

api/http/handlerfuncs_export.go
```diff
@@ -0,0 +1,188 @@
...
+
+	sendJSON(
+		req.Context(),
+		rw,
+		DataResponse{docs},
+		http.StatusOK,
+	)
+}
+
+func importHandler(rw http.ResponseWriter, req *http.Request) {
```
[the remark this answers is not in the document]
fredcarle: it's now in `handlerfuncs_backup.go`

cli/db_import_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestDBImportCmd_WithEmptyDatastore_ReturnError(t *testing.T) {
+	cfg, _, close := startTestNode(t)
+	defer close()
+
+	filePath := t.TempDir() + "/test.json"
+
+	err := os.WriteFile(
+		filePath,
+		[]byte(`{"User":[{"_key":"bae-e933420a-988a-56f8-8952-6c245aebd519","_newKey":"bae-e933420a-988a-56f8-8952-6c245aebd519","age":30,"name":"John"}]}`),
+		0664,
```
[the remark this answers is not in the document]
fredcarle: I'm not of fan of this and it's now really something you see in Go codebases.

db/backup.go
fredcarle: I've applied the suggestion and it does simplify the code. It took me a little while to understand the behaviour since the documentation isn't super clear unfortunately. Thanks again for suggesting it.

db/backup.go
```diff
@@ -0,0 +1,266 @@
...
+			}
+		}
+	}()
+
+	// open the object
+	_, err = f.WriteString("{")
+	if err != nil {
+		return NewErrFailedToWriteString(err)
+	}
+	if config.Pretty {
```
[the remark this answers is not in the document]
AndrewSisley: second related but independent suggestion: This func does a lot of:
```
if pretty {
    _, err = f.WriteString("some pretty string")
    if err != nil {
	    return NewErrFailedToWriteString(err)
    }
} else {
    _, err = f.WriteString("some ugly string")
    if err != nil {
	    return NewErrFailedToWriteString(err)
    }
}
```
I think it would be quite a lot nicer if instead you just did:
```
var foo string
if pretty {
    foo = "some pretty string"
} else {
    foo = "some ugly string"
}
_, err = f.WriteString(foo)
if err != nil {
	 return NewErrFailedToWriteString(err)
}
```
This way the differences between the two blocks are much clearer, and there is less duplication.

AndrewSisley (changes requested):
Looks good, but I'd like a few more changes before merge please.  I'm only part way through review, but thought I'd submit what I have now and let you get cracking.

cli/db_export.go
```diff
@@ -0,0 +1,142 @@
...
+		Short: "Export the database to a file",
+		Long: `Export the database to a file.
+		
+If the --collection flag is provided, only the data for that collection will be exported.
+Otherwise, all collections in the database will be exported.
+
+If the --pretty flag is provided, the JSON will be pretty printed.
+
+Example: export data for the 'Users' collection:
+  defradb client export --collection Users user_data.json`,
```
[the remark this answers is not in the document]
AndrewSisley: If not tested already, we should probably assert that behaviour does not change either, if we are explicitly noting it in the documentation.

db/backup.go
```diff
@@ -0,0 +1,266 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (db *db) basicExport(ctx context.Context, txn datastore.Txn, config *client.BackupConfig) (err error) {
+	cols := []client.Collection{}
```
AndrewSisley, line 94: todo: These are not scoped to a the input txn param, and as such, actions on them, such as `GetAllDocKeys` will not be scoped to the provided transaction either.  Please scope these to the txn.

db/backup.go
```diff
@@ -0,0 +1,266 @@
...
+	if t != json.Delim('{') {
+		return ErrExpectedJSONObject
+	}
+	for d.More() {
+		t, err := d.Token()
+		if err != nil {
+			return err
+		}
+		colName := t.(string)
+		col, err := db.getCollectionByName(ctx, txn, colName)
```
AndrewSisley, line 51: todo: These are not scoped to a the input txn param, and as such, actions on them, such as `Create` will not be scoped to the provided transaction either.  Please scope these to the txn.

db/backup.go
AndrewSisley, line 1: todo: This definitely should be integration tested, not just for test longevity, but also to test using multiple stores (and later, file systems) and with the change detector.

As time is short, I think creating a ticket noting the test gap (also for CLI ideally) should be fine for now and they can be added later.

db/errors.go
```diff
@@ -59,12 +60,24 @@ const (
 	errIndexDescriptionHasNoFields    string = "index description has no fields"
 	errIndexDescHasNonExistingField   string = "index description has non existing field"
 	errFieldOrAliasToFieldNotExist    string = "The given field or alias to field does not exist"
+	errCreateFile                     string = "failed to create the file"
```
AndrewSisley, line 63: nitpick: `the` in these error messages feels quite odd to me. It might also be misleading if the filepath is not noted in the error (e.g. in case error was from a internal-file and not the user-specified path).

client/document.go
```diff
@@ -283,6 +283,11 @@ func (doc *Document) setAndParseType(field string, value any) error {
 		if err != nil {
 			return err
 		}
+	case uint64:
```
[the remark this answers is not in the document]
fredcarle: Because of the json marshalling and unmashalling of the documents. Type int can be interpreted as such when imported. If I remove this case, this is the error we get: `unexpected type. Property: age, Actual: uint64`

api/http/router.go
```diff
@@ -33,6 +33,8 @@ const (
 	SchemaMigrationPath string = SchemaPath + "/migration"
 	IndexPath           string = versionedAPIPath + "/index"
 	PeerIDPath          string = versionedAPIPath + "/peerid"
+	ExportPath          string = versionedAPIPath + "/export"
```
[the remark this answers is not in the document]
fredcarle: It's the same everywhere. The difference is the CLI splits them into two file and `db` and `http` bundle the export and import into a single file called backup. Backup is the category and import and export are the individual actions.

client/mocks/DB.go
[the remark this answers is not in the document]
fredcarle: It already there.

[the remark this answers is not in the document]
fredcarle: Added by you btw ;)

db/backup.go
fredcarle: This is not an obvious behaviour. I was sure that since the call to get collection was scoped to a txn, the return collections and their associated operations would also automatically be scoped to the same transaction.

db/backup.go
```diff
@@ -0,0 +1,266 @@
...
+		for _, colName := range config.Collections {
+			col, err := db.getCollectionByName(ctx, txn, colName)
+			if err != nil {
+				return NewErrFailedToGetCollection(colName, err)
+			}
+			cols = append(cols, col)
+		}
+	}
+
+	f, err := os.Create(config.Filepath)
```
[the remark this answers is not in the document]
fredcarle: I don't think this one needs to be scoped to the txn. The txn on export cannot fail or have conflict. However, if there is an operational error, the file will be removed anyways.

db/backup.go
fredcarle: I'm not sure there is any real benefit to integration testing with this one. There is already 3 layers of test. In the `db` package, then in the `api/http` package which also tests what's in the `db` package and then in the `cli` package which tests both what's int the `api/http` package and in the `db` package. The integration tests all run on the same file-system as the unit tests and it's not these functionalities jobs to ensure that docs can be read or written to different datastores.

client/document.go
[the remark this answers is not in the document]
AndrewSisley: Okay, this needs to change though, because now you are persisting JSON parsing logic into CBOR database storage.  Lens had this issue too.

The type/values that we store in the database stores should not be dependent on external data formats.

api/http/router.go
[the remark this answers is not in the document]
AndrewSisley: If backup category, should this not be reflected in the CLI and http path?  `backup export` and `versionedAPIPath + "/backup/export"` (refactored, like migrations is)?

client/mocks/DB.go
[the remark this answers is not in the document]
AndrewSisley: :grin: I swear this file had a new file icon when I reviewed it lol.  Sorry :laughing:

api/http/router.go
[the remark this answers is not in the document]
fredcarle: I feel like thats just superfluous words when wanting to access the functionality in the url and the cli command. I prefer the direct action in those two cases.

db/backup.go
AndrewSisley: No, sadly not - it is an ambiguity in the client package that we have not yet fixed

db/backup.go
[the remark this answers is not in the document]
AndrewSisley: There are many errors that can occur between calling `Create` (and truncating an existing file if there is one), and finishing the export.  Including file-write errors.  atm a partial success seems very possible, and potentially quite destructive.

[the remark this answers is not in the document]
fredcarle: I see what you mean. I'll look for an different approach.

db/backup.go
fredcarle: ✅

db/backup.go
fredcarle: ✅

client/backup.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// BackupConfig holds the configuration parameters for database backups.
+type BackupConfig struct {
+	Filepath string `json:"filepath"`
+	// only JSON is supported at the moment
+	Format string `json:"format"`
+	// pretty print JSON
+	Pretty      bool     `json:"pretty"`
+	Collections []string `json:"collections"`
```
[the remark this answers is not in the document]
fredcarle: ✅

db/backup.go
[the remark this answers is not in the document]
fredcarle: ✅

cli/db_export.go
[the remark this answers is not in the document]
fredcarle: ✅

client/backup.go
```diff
@@ -0,0 +1,34 @@
...
+	// BasicImport imports a json dataset.
+	// filepath must be accessible to the node.
+	BasicImport(ctx context.Context, filepath string) error
+	// BasicExport exports the current data or subset of data to file in json format.
+	BasicExport(ctx context.Context, config *BackupConfig) error
+}
+
+// BackupConfig holds the configuration parameters for database backups.
+type BackupConfig struct {
+	Filepath string `json:"filepath"`
```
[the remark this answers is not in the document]
fredcarle: ✅

api/http/router.go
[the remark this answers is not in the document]
AndrewSisley: It is slightly, but it means the 3 access points are all structured and named the same way, making it much easier to understand and translate between them.  It confused me, and will probably throw externals too.

db/errors.go
fredcarle: ✅

db/backup.go
AndrewSisley: Will raise in standup.  I strongly disagree with this.

api/http/router.go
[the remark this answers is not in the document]
fredcarle: But there is nowhere that the functionalities need to write backup other than for `client.BackupConfig`. The only use of backup is for grouping under a file or for grouping the methods under a narrow interface type. I don't really understand where the confusion is.

I'd much rather type `defradb client export myfile.json` than `defradb client backup export myfile.json` The two are just as clear in my opinion.

db/backup.go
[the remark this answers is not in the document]
fredcarle: ✅

db/backup.go
fredcarle: Consensus reached in standup. Integration tests will be added.

api/http/router.go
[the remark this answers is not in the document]
fredcarle: Change the cli and api pattern to fit your suggestion.

✅

client/document.go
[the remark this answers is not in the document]
fredcarle: Can you please point me towards what you did with Lens for this situation?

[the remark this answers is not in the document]
AndrewSisley: It was the following lines in `core.DecodeFieldValue`:
```
case client.FieldKind_INT:
	switch v := val.(type) {
	case float64:
		if v >= 0 {
			return uint64(v), nil
		}
		return int64(v), nil
	}
}
```
This ensured that Json-parsed values were decoded to standard (i.e. CBOR) types.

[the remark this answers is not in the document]
fredcarle: added temporary fix until #1681 is resolved.

✅

AndrewSisley (approved):
LGTM - cheers for all your efforts here Fred!

Just to reiterate - I personally am totally fine with any integration tests being done in a new ticket post merge

fredcarle:
> LGTM - cheers for all your efforts here Fred!
> 
> Just to reiterate - I personally am totally fine with any integration tests being done in a new ticket post merge

Thanks Andy! 

Integration tests just landed and I'm finalizing one final change on export

db/backup.go
fredcarle: ✅

tests/integration/backup/one_to_one.go/export_test.go
shahzadlone, line 1: todo: Why is the folder called `one_to_one.go` lol, please fix.

AndrewSisley: good spot :grin:

tests/integration/backup/one_to_many/utils.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1: todo: Fix copyright year.

```suggestion
// Copyright 2023 Democratized Data Foundation
```

tests/integration/backup/one_to_one.go/utils.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1: todo: Fix copyright year.

```suggestion
// Copyright 2023 Democratized Data Foundation
```

tests/integration/backup/simple/utils.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1: todo: Fix copyright year.

```suggestion
// Copyright 2023 Democratized Data Foundation
```

cli/backup.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 Democratized Data Foundation
```
shahzadlone, line 1: todo: Fix copyright year.

```suggestion
// Copyright 2023 Democratized Data Foundation
```

tests/integration/backup/one_to_one.go/export_test.go
fredcarle: LOL! Oups..

shahzadlone (approved):
LGTM. If possible I would like a usage example in the `README.md` for new users who are might be wondering if a tool or utility to import and export exists, or under some FAQ section where a question like so is being answered "How can I import and export defradb data".

tests/integration/backup/one_to_one.go/export_test.go
fredcarle: ✅

api/http/handlerfuncs_backup_test.go
```diff
@@ -0,0 +1,743 @@
...
+// 	errResponse := ErrorResponse{}
+// 	testRequest(testOptions{
+// 		Testing:        t,
+// 		DB:             db,
+// 		Method:         "GET",
+// 		Path:           ExportPath,
+// 		Body:           nil,
+// 		ExpectedStatus: 500,
+// 		ResponseData:   &errResponse,
+// 	})
```
shahzadlone, line 215: todo: uncomment?

api/http/handlerfuncs_backup_test.go
```diff
@@ -0,0 +1,743 @@
...
+// 	t.Cleanup(CleanupEnv)
+// 	env = "dev"
+
+// 	testError := errors.New("test error")
+// 	db := mocks.NewDB(t)
+// 	txn := dsMocks.NewTxn(t)
+// 	store := mockStore{
+// 		DB:  db,
+// 		Txn: txn,
+// 	}
```
shahzadlone, line 607: todo: remove or uncomment?

README.md
```diff
@@ -387,6 +389,25 @@ defradb start --allowed-origins=http://localhost:3000
 
 The catch-all `*` is also a valid origin. 
 
+## Backing up and restoring
+
+It is currently no possible to do a full backup of DefraDB that includes the history of changes through the Merkle DAG. However, DefraDB currently supports a simple backup of the data in JSON format that can be used to seed a database or help with transitioning from one DefraDB version to another.
```
shahzadlone, line 394:
```suggestion
It is currently not possible to do a full backup of DefraDB that includes the history of changes through the Merkle DAG. However, DefraDB currently supports a simple backup of the current data state in JSON format that can be used to seed a database or help with transitioning from one DefraDB version to another.
```

api/http/handlerfuncs_backup_test.go
fredcarle: They need to be removed. I had kept them as reference for tests I moved to the `db` package. Thanks for pointing this out.

README.md
fredcarle: ✅

README.md
```diff
@@ -387,6 +389,25 @@ defradb start --allowed-origins=http://localhost:3000
 
 The catch-all `*` is also a valid origin. 
 
+## Backing up and restoring
```
shahzadlone, line 392: thought: I like this backing up and restoring analogy better :), but wondering if should be consistent as currently we have files with the following names:
- `backup_export.go`
- `backup_import.go`

Maybe instead can be:
- `backup.go`
- `restore.go`

README.md
```diff
@@ -387,6 +389,25 @@ defradb start --allowed-origins=http://localhost:3000
...
+
+It is currently not possible to do a full backup of DefraDB that includes the history of changes through the Merkle DAG. However, DefraDB currently supports a simple backup of the current data state in JSON format that can be used to seed a database or help with transitioning from one DefraDB version to another.
+
```
shahzadlone, line 395: thought: If the above thought is expanded then we can have the cli be:
```
defradb client backup path/to/backup.json
defradb client restore path/to/backup.json
```

Instead of:
```
defradb client backup export path/to/backup.json
defradb client backup import path/to/backup.json
```

README.md
fredcarle: The way I went about it is: it's the backup feature and it can export and import data. I wanted a category of actions.

README.md
fredcarle: I had 
```
defradb client export path/to/backup.json
defradb client import path/to/backup.json
```
before and it was suggested that grouping them under a command / api path would be clearer.

README.md
shahzadlone: Yep I get that, I feel like backup is being used as an umbrella term. IMO commands will be simpler too if we can pick up one concise word for each action. As pointed out below

README.md
shahzadlone: Hmm, I do like the one with less words to type. But if there seems to be a consensus reached already then will just follow that (also is a non-blocker thought).

api/http/router.go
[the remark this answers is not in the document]
shahzadlone: > But there is nowhere that the functionalities need to write backup other than for `client.BackupConfig`. The only use of backup is for grouping under a file or for grouping the methods under a narrow interface type. I don't really understand where the confusion is.
> 
> I'd much rather type `defradb client export myfile.json` than `defradb client backup export myfile.json` The two are just as clear in my opinion.

Sorry for the delay, I missed this convo. Just casting my vote that I do prefer:
```
defradb client restore myfile.json
defradb client backup myfile.json
```
OR
```
defradb client export myfile.json
defradb client import myfile.json
```
OVER the longer version

PR #1631, opened 2023-07-12
feat: Add `_not` operator

### LINKED ISSUE #1542: No tests for `_not` operator in filters.
We currently have no tests that test for the `_not` operator in filters. I further suspect that its actually broken based on my work for #1541 . Haven't verified if its working. 

If its not working, its seems like the cause is within the mapper `ToFilter` not correctly handling the `_not` operator.

fredcarle:
## Relevant issue(s)

Resolves #1542 

## Description

This PR add the `_not` operator as an option for filters in graphql queries.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/connor/not.go
+++ b/connor/not.go
@@ -3,11 +3,9 @@ package connor
 // not is an operator which performs object equality test
 // and returns the inverse of the result.
 func not(condition, data any) (bool, error) {
-	if m, err := eq(condition, data); err != nil {
+	m, err := eq(condition, data)
+	if err != nil {
 		return false, err
-	} else if m {
-		return false, nil
 	}
-
-	return true, nil
+	return !m, nil
 }
--- a/connor/not.go
+++ b/connor/not.go
@@ -1,7 +1,7 @@
 package connor
 
-// like is an operator which performs string equality
-// tests.
+// not is an operator which performs object equality test
+// and returns the inverse of the result.
 func not(condition, data any) (bool, error) {
 	if m, err := eq(condition, data); err != nil {
 		return false, err
--- a/connor/not.go
+++ b/connor/not.go
@@ -1,0 +1,13 @@
+package connor
+
+// like is an operator which performs string equality
+// tests.
+func not(condition, data any) (bool, error) {
+	if m, err := eq(condition, data); err != nil {
+		return false, err
+	} else if m {
+		return false, nil
+	}
+
+	return true, nil
+}
--- a/connor/not_test.go
+++ b/connor/not_test.go
@@ -28,7 +28,7 @@ func TestNot_WithEmptyCondition_ReturnError(t *testing.T) {
 }
 
 type operator struct {
-	// The filter operation string that this Operator represents.
+	// The filter operation string that this `operator`` represents.
 	//
 	// E.g. "_eq", or "_and".
 	Operation string
--- a/connor/not_test.go
+++ b/connor/not_test.go
@@ -1,0 +1,50 @@
+package connor
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNot_WithNotAndNotNot_NoError(t *testing.T) {
+	const testString = "Source is the glue of web3"
+
+	// not equal
+	result, err := not(testString, testString)
+	require.NoError(t, err)
+	require.False(t, result)
+
+	// not not equal
+	result, err = not("Source is the glue", testString)
+	require.NoError(t, err)
+	require.True(t, result)
+}
+
+func TestNot_WithEmptyCondition_ReturnError(t *testing.T) {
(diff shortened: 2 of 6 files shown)


connor/not.go
```diff
@@ -0,0 +1,13 @@
+package connor
+
+// like is an operator which performs string equality
```
shahzadlone, line 3: todo: copy paste typo

fredcarle: thanks for catching!

fredcarle: ✅

tests/integration/query/simple/with_filter/with_not_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2022 Democratized Data Foundation
```
AndrewSisley, line 1: todo: Whilst they are a pain to write, I do think it is worth adding at least one schema test.  Particularly given the order bug Keenan found today.

tests/integration/query/simple/with_filter/with_not_test.go
```diff
@@ -0,0 +1,109 @@
...
+//
+// Use of this software is governed by the Business Source License
```
AndrewSisley, line 3: todo: I think a test for an empty `_not` clause is required.

AndrewSisley (changes requested):
Production code looks good, I think more tests are required though.

connor/not_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestNot_WithEmptyCondition_ReturnError(t *testing.T) {
+	const testString = "Source is the glue of web3"
+
+	_, err := not(map[FilterKey]any{&operator{"_some"}: "test"}, testString)
+	require.ErrorIs(t, err, ErrUnknownOperator)
+}
+
+type operator struct {
+	// The filter operation string that this Operator represents.
```
shahzadlone, line 31: nitpick:
```suggestion
	// The filter operation string that this `operator` represents.
```

connor/not.go
```diff
@@ -0,0 +1,13 @@
...
+// and returns the inverse of the result.
+func not(condition, data any) (bool, error) {
+	if m, err := eq(condition, data); err != nil {
+		return false, err
+	} else if m {
+		return false, nil
+	}
+
+	return true, nil
+}
```
shahzadlone, line 13: nitpick: Could just be:

```suggestion
func not(condition, data any) (bool, error) {
	if m, err := eq(condition, data); err != nil {
		return false, err
	}
	return !m, nil
}
```

shahzadlone (approved):
LGTM! Thanks for taking care of this. I think some more tests would be nice, but not a blocker for me and the implementation seems straight forward.

connor/not.go
fredcarle: I like this. Thank you for the suggestion.

tests/integration/query/simple/with_filter/with_not_test.go
fredcarle: What do you mean by a schema test?

tests/integration/query/simple/with_filter/with_not_test.go
fredcarle: will do :)

connor/not_test.go
fredcarle: ✅

connor/not.go
fredcarle: ✅

tests/integration/query/simple/with_filter/with_not_test.go
fredcarle: ✅

tests/integration/query/simple/with_filter/with_not_test.go
fredcarle: If so I don't think this PR has any impact on those.

AndrewSisley: > If so I don't think this PR has any impact on those.

It does, as the feature will actually work now, so we should ensure that it stays working.  I had a look though, and `_not` is already covered by them.

AndrewSisley (approved):
LGTM, cheers Fred

tests/integration/query/simple/with_filter/with_not_test.go
```diff
@@ -0,0 +1,200 @@
+// Copyright 2022 Democratized Data Foundation
```
AndrewSisley, line 1: todo: Because we mentioned this on our call, here is the `2022` :laughing:

fredcarle: Thanks for catching my sloppy copy pasting ;)

fredcarle: ✅

PR #1635, opened 2023-07-13
refactor: Remove redundant txn param from fetcher start

### LINKED ISSUE #1623: Remove `txn` param from Fetcher.Start
As of https://github.com/sourcenetwork/defradb/pull/1564 it should be taken from the `Init` call.

AndrewSisley:
## Relevant issue(s)

Resolves #1623

## Description

Removes the redundant txn param from fetcher start as it is always the same value as fetcher.Init

### DIFF
--- a/db/collection_get.go
+++ b/db/collection_get.go
@@ -63,7 +63,7 @@ func (c *collection) get(
 	// construct target key for DocKey
 	targetKey := base.MakeDocKey(*desc, key.DocKey)
 	// run the doc fetcher
-	err = df.Start(ctx, txn, core.NewSpans(core.NewSpan(targetKey, targetKey.PrefixEnd())))
+	err = df.Start(ctx, core.NewSpans(core.NewSpan(targetKey, targetKey.PrefixEnd())))
 	if err != nil {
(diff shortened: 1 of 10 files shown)


fredcarle (approved):
LGTM.

It seems that a more appropriate name for the PR would be `Move txn from fetcher start to init` don't you think?

AndrewSisley:
> LGTM.
> 
> It seems that a more appropriate name for the PR would be `Move txn from fetcher start to init` don't you think?

I disagree with you on that :)  Your suggestion suggests it doesnt already exist on init

fredcarle:
> > LGTM.
> > It seems that a more appropriate name for the PR would be `Move txn from fetcher start to init` don't you think?
> 
> I disagree with you on that :) Your suggestion suggests it doesnt already exist on init

That's my point. It didn't.

fredcarle:
I mean the it was part of the Init param but nothing was done with it.

AndrewSisley:
> That's my point. It didn't.

I think you are mistaken.  It was added to Init in an earlier Lens PR.

PR #1645, opened 2023-07-17
test: Bundle test state into single var

### LINKED ISSUE #1644: Collect integration test state vars into single context-like var
There are now many vars that the suite has to track, and they currently have to be passed individually into each action-func that requires   them.  Roll them up into a single type and each action can receive it without worry which sub-props it actually needs.

This can later allow actions to be interface-driven instead of the switch, which is also starting to become unwieldy.

AndrewSisley:
## Relevant issue(s)

Resolves #1644

## Description

Bundles all test state into single var.  Stuff was starting to get too messy as the action system expanded.  This makes all the stuff consistent, in name and location and removes the need to pass the individual items around as and when needed. All props have been documented.

It also simplifies refreshing state, as return values are no longer needed.  It opens the way to making the action system interface-based, which we can look at in the future if we want but not here in this PR.

### DIFF
--- a/tests/integration/explain.go
+++ b/tests/integration/explain.go
@@ -16,12 +16,12 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/sourcenetwork/immutable"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/logging"
-	"github.com/sourcenetwork/immutable"
 )
 
 var (
--- a/tests/integration/explain.go
+++ b/tests/integration/explain.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/logging"
-	"github.com/sourcenetwork/defradb/net"
 	"github.com/sourcenetwork/immutable"
 )
 
@@ -103,7 +102,6 @@ type ExplainRequest struct {
 
 func executeExplainRequest(
 	s *state,
-	nodes []*net.Node,
 	action ExplainRequest,
 ) {
 	// Must have a non-empty request.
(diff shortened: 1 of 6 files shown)


tests/integration/utils2.go
```diff
@@ -1311,73 +1204,71 @@ func executeRequest(
...
-				data = append(data, sData...)
-
-				if len(data) >= len(action.Results) {
-					expectedDataRecieved = true
-				}
-
-			case <-allActionsDone:
-				allActionsAreDone = true
-			}
+	for _, node := range getNodes(action.NodeID, s.nodes) {
```
fredcarle, line 1212: question: It's not quite clear to me why we now have a for loop here. Could you give a quick explanation please?

fredcarle (approved):
LGTM. There is just one change that I can't easily understand why it was done.

tests/integration/explain.go
```diff
@@ -97,41 +101,40 @@ type ExplainRequest struct {
...
 
-	result := db.ExecRequest(ctx, action.Request)
-	assertExplainRequestResults(
-		ctx,
-		t,
-		description,
-		&result.GQL,
-		action,
-	)
+	for _, node := range getNodes(action.NodeID, s.nodes) {
```
shahzadlone, line 128: question: Can you write a line here to document what the addition of this for loop enables?

tests/integration/utils2.go
AndrewSisley: All of the other actions do this.  A test case may involve many nodes, is just this one and a couple of other stragglers hadn't implemented it yet.

tests/integration/explain.go
AndrewSisley: This loop is a copy-paste of the same line in the other 20 or so actions.  I don't really think it needs to chew up anymore eyespace.  It is not unusual.

tests/integration/utils2.go
fredcarle: But it still worked and didn't affect outcome?

AndrewSisley: The NodeID property was added along with the loop here.  Before it was essentially hardcoded to only ever act on the first node, regardless of the test configuration.

fredcarle: These test were never meant to have more than one node but it could be useful in the future 👍

PR #1647, opened 2023-07-17
fix: Return error when relation is not defined on both types

### LINKED ISSUE #1455: AddSchema does not error if only given half of relationship
For example the below should error upon definition as the relationship has not been fully defined and will not be usable:
```
type Users {
	 b: [Vote]
 }
type Vote {
	c: String
}
```

nasdf:
## Relevant issue(s)

Resolves #1455 

## Description

This fixes an issue where creating schemas with only one side of a relation defined would not return an error.

## Tasks

## How has this been tested?

Added a unit test.

- MacOS


AndrewSisley (approved):
Looks good to me!  Surprised how clean the fix looks.

However, I think you may only have solved part of the linked issue.  This problem likely affects both sides of a many-one, and one-one relationships.  Super happy for this to be merged as-is, but please leave the issue open (with note RE this PR) if you don't sort out the rest here :)

fredcarle (approved):
Good job on finding this fix Keenan :) Just one todo related to schema definition.

To go with what Andy is saying, it would be a nice addition to document if the other relationship options behave correctly (Can be done in a separate PR). 

This is the one you tested for:
```graphql
# Bad relationship definition 1-many
type Dog {
  name: String
}
type User {
  dogs: [Dog]
}

# Good relationship definition 1-many
type Dog {
  name: String
  user: User
}
type User {
  dogs: [Dog]
}
```

These are the others that might be nice to test for:
```graphql
# Bad relationship definition 1-1
type Dog {
  name: String
}
type User {
  dog: Dog
}

# Still bad relationship definition 1-1
type Dog {
  name: String
  user: User
}
type User {
  dog: Dog
}

# Good relationship definition 1-1
type Dog {
  name: String
  user: User @primary # primary can be set on whichever side of the relationship
}
type User {
  dog: Dog
}
```

shahzadlone (approved):
LGTM overall, just had similar points to Andy and Fred

nasdf:
Thanks everyone for the feedback. I've updated this fix to return an error when any relationship type is one sided. I've also added more test cases and moved them all to a new file.

These changes did cause one of the test cases to fail because we check relationships before field types.  If you try to create a schema with an unknown type it will now instead return a one sided relationship error.

As an example. The schema below will return an error `relation must be defined on both schemas` instead of `no type found for given name`.
```
type User {
  name: InvalidField
}
```

tests/integration/schema/simple_test.go
```diff
@@ -150,23 +150,6 @@ func TestSchemaSimpleCreatesSchemaWithDefaultFieldsGivenEmptyType(t *testing.T)
 	testUtils.ExecuteTestCase(t, []string{"Users"}, test)
 }
 
-func TestSchemaSimpleErrorsGivenTypeWithInvalidFieldType(t *testing.T) {
```
fredcarle, line 153: suggestion: Instead of removing that test, it would be nice to keep but change to the new expected error

fredcarle (approved):

PR #1673, opened 2023-07-20
fix: Fix fetcher close

### LINKED ISSUE #1672: AssertFail in badger file on db close
Has started happening in the change detector recently, happens in `TestQueryOneToOneWithGroupRelatedIDFromSecondary` most (but not all) of the time.  Removing the last query removes the assert fail.

*Might* be a badger bug, as they seem to have had issues in the past: https://github.com/dgraph-io/badger/issues/1350

If the request in the test is modified:

This fails:
```
Request: `query {
	Book {
		name
	}
}`,
``` 
This works:
```
Request: `query {
	Author {
		name
	}
}`,
``` 
- Removing either one of the Author create actions removes/hides the problem. 
- Removing lines 874-891 (where secondary id creation is handled) from collection.go has no impact.
- Removing the Lens fetcher has no impact.
- Updating to badger v4.1.0 has no impact
- Exiting before calling `fetcher.FetchNextDoc` in scan node removes the error, suggesting that `fetcher.Init` and `fetcher.Start` are not relevant in this issue.
- Exiting immediately after the `FetchNext` call in `FetchNextDoc` has no impact, suggesting `FetchNext` is responsible
- Removing the filter stuff from `FetchNext` has no impact, suggesting it's irrelevance
- Returning an error anywhere within the last 4 lines of `fetcher.nextKey` removes the assert failure
- Calling `nextKV` 8 times results in the failure, calling it 7 times does not.  The key on the 8th value is `/1/v/bae-d6627fea-8bf7-511c-bcf9-bac4212bddd6/v`
- Calling `df.kvResultsIter.NextSync()` in `nextKVRaw` a 9th time results in the problem, regardless of whether the value is ignored or not

AndrewSisley:
## Relevant issue(s)

Resolves #1672

## Description

Fixes the fetcher close logic, removing the too-early exits.

Lucky guess, but it seems to have fixed the change detector issue.

- [ ] Don't forget to remove the temp commit :) Will also need to add a breaking change doc to allow merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,7 @@ test\:coverage-html:
 .PHONY: test\:changes
 test\:changes:
 	@$(MAKE) deps:lens
-	env DEFRA_DETECT_DATABASE_CHANGES=true gotestsum -- ./... -shuffle=on -p 1
+	env DEFRA_DETECT_DATABASE_CHANGES=true DEFRA_TARGET_BRANCH=1672-assert-true-chd-db-close gotestsum -- ./... -shuffle=on -p 1
 
 .PHONY: validate\:codecov
 validate\:codecov:
--- a/db/fetcher/fetcher.go
+++ b/db/fetcher/fetcher.go
@@ -656,22 +656,18 @@ func (df *DocumentFetcher) FetchNextDoc(
 
 // Close closes the DocumentFetcher.
 func (df *DocumentFetcher) Close() error {
-	if df.kvIter == nil {
-		return nil
+	if df.kvIter != nil {
+		err := df.kvIter.Close()
+		if err != nil {
+			return err
+		}
 	}
 
-	err := df.kvIter.Close()
-	if err != nil {
-		return err
-	}
-
-	if df.kvResultsIter == nil {
-		return nil
-	}
-
-	err = df.kvResultsIter.Close()
-	if err != nil {
-		return err
+	if df.kvResultsIter != nil {
+		err := df.kvResultsIter.Close()
+		if err != nil {
+			return err
+		}
 	}
 
 	if df.deletedDocFetcher != nil {


shahzadlone (approved):
Lgtm

Makefile
```diff
@@ -272,7 +272,7 @@ test\:coverage-html:
 .PHONY: test\:changes
 test\:changes:
 	@$(MAKE) deps:lens
-	env DEFRA_DETECT_DATABASE_CHANGES=true gotestsum -- ./... -shuffle=on -p 1
+	env DEFRA_DETECT_DATABASE_CHANGES=true DEFRA_TARGET_BRANCH=1672-assert-true-chd-db-close gotestsum -- ./... -shuffle=on -p 1
```
fredcarle, line 275: question: Why do we need to set the target branch?

db/fetcher/fetcher.go
```diff
@@ -656,22 +656,18 @@ func (df *DocumentFetcher) FetchNextDoc(
...
-	}
-
-	err := df.kvIter.Close()
-	if err != nil {
-		return err
-	}
-
-	if df.kvResultsIter == nil {
-		return nil
+	if df.kvIter != nil {
```
nasdf, line 659: Should these iterators be set to `nil` after closing to avoid calling `Close` again?

Makefile
AndrewSisley: For testing, please see the commits.

db/fetcher/fetcher.go
AndrewSisley: Is an interesting thought, but am leaning towards 'I don't think so'.  I see it as the norm for Close func to assume that they will only ever be called once, and whilst you could argue that setting stuff to nil and allowing the GC to dispose of the objects is within the scope of a func call, it still feels slightly off to me.

Have you seen such things done much before?

AndrewSisley:
Still broken in CI, had it consistently passing locally, so this is a bit odd :(  Sorry for reviewer time

PR #1678, opened 2023-07-21
fix: Remove duplicated `peers` in peerstore prefix

### LINKED ISSUE #1260: Peer store is prefixed twice with `peers` key
https://github.com/sourcenetwork/defradb/blob/c9ef176d561ec0f085eb4efc1cf528842122a40b/node/node.go#L97

```
/peers/peers/addrs/AASAQAISEAOKGS54....
/peers/peers/keys/AASAQAISEAOKGS54.../priv": ...
/peers/peers/keys/AASAQAISEAOKGS54.../pub": ...
```
suggestion: it could be called `p2p` like the comment is hinting at.

fredcarle:
## Relevant issue(s)

Resolves #1260 

## Description

This PR changes the peerstore prefix from `/peers/peers` to `/db/peers`

## How has this been tested?

manual `defradb client dump`

- MacOS

### DIFF
--- a/net/node.go
+++ b/net/node.go
@@ -85,7 +85,7 @@ func NewNode(
 	// create our peerstore from the underlying defra rootstore
 	// prefixed with "p2p"
 	rootstore := db.Root()
-	pstore := namespace.Wrap(rootstore, ds.NewKey("peers"))
+	pstore := namespace.Wrap(rootstore, ds.NewKey("/db"))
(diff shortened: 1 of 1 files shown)


AndrewSisley (approved):
Approved, with the out of scope thought that pstore should probably be part of multistore.

fredcarle:
> Approved, with the out of scope thought that pstore should probably be part of multistore.

Issue created #1680

PR #1685, opened 2023-07-22
fix: Fix invalid (badger) datastore state

### LINKED ISSUE #1672: AssertFail in badger file on db close

AndrewSisley:
## Relevant issue(s)

Resolves #1672

## Description

Fixes invalid datastore state where somehow the iterator is not closed properly, even though Close is called.  I am guessing that something is getting GC'd before it can be closed, as closing it immediately after use appears to remove the issue.

Bug discovered testing with badger-file, but I cannot be sure where the GC'd object is, and so it could be within the ipfs code, which could mean other datastore implementations were exposed to this.

Fix is commit `Close iterator on spanDone` *only*, the others are small improvements to the codebase I found along the way.

### DIFF
--- a/db/fetcher/fetcher.go
+++ b/db/fetcher/fetcher.go
@@ -332,6 +332,10 @@ func (df *DocumentFetcher) nextKey(ctx context.Context, seekNext bool) (spanDone
 
 	df.kvEnd = spanDone
 	if df.kvEnd {
+		err = df.kvResultsIter.Close()
+		if err != nil {
+			return false, false, err
+		}
 		moreSpans, err := df.startNextSpan(ctx)
 		if err != nil {
 			return false, false, err
--- a/db/fetcher/fetcher.go
+++ b/db/fetcher/fetcher.go
@@ -656,22 +656,18 @@ func (df *DocumentFetcher) FetchNextDoc(
 
 // Close closes the DocumentFetcher.
 func (df *DocumentFetcher) Close() error {
-	if df.kvIter == nil {
-		return nil
+	if df.kvIter != nil {
+		err := df.kvIter.Close()
+		if err != nil {
+			return err
+		}
 	}
 
-	err := df.kvIter.Close()
-	if err != nil {
-		return err
-	}
-
-	if df.kvResultsIter == nil {
-		return nil
-	}
-
-	err = df.kvResultsIter.Close()
-	if err != nil {
-		return err
+	if df.kvResultsIter != nil {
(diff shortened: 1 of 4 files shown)


db/fetcher/fetcher.go
```diff
@@ -332,6 +332,10 @@ func (df *DocumentFetcher) nextKey(ctx context.Context, seekNext bool) (spanDone
 
 	df.kvEnd = spanDone
 	if df.kvEnd {
+		err = df.kvResultsIter.Close()
```
shahzadlone, line 335: Question: where was this close happening before? Do we have to worry about double closing now?

shahzadlone (approved):

db/fetcher/fetcher.go
AndrewSisley: Double closing is a non-issue, the docs say you can close as often as you like.  Before it was happening in fetcher.Close, and fetcher.startNextSpan.  It is still done in those places. In some situations (like the uncommented out test) Close will now be called multiple times.

Issue #1697, opened 2023-07-24
Backup-Import cannot handle self referencing documents 


AndrewSisley:
As documented by `TestBackupSelfRefImport_SelfReferencing_NoError` and `TestBackupSelfRefImport_SplitPrimaryRelationWithSecondCollection_NoError`.

Whilst it is very difficult to achieve by import alone due to difficulties in generating the relation-id, this issue does mean that an export of such a document cannot be imported correctly.

PR #1699, opened 2023-07-24
test: Expand backup integration tests

### LINKED ISSUE #1696: Self-reference import missing integration tests

AndrewSisley:
## Relevant issue(s)

Resolves #1696 

## Description

Expands the backup integration tests, and fixes an invalid schema declaration.

Documents https://github.com/sourcenetwork/defradb/issues/1697

I went over the CLI tests too, but found no gaps.

### DIFF
--- a/tests/integration/backup/self_reference/import_test.go
+++ b/tests/integration/backup/self_reference/import_test.go
@@ -13,6 +13,8 @@ package backup
 import (
 	"testing"
 
+	"github.com/sourcenetwork/immutable"
+
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
 )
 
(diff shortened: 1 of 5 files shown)


tests/integration/backup/self_reference/import_test.go
```diff
@@ -0,0 +1,303 @@
...
+	test := testUtils.TestCase{
+		Actions: []any{
+			testUtils.BackupImport{
+				// Note: Whilst generating the `boss_id` field value manually would be really difficult,
+				// the document may be exported with `_key`/`_newKey` values, resulting in an export that
+				// could not be fully imported, although all the information required is there.
+				ImportContent: `{
+					"User":[
+						{
+							"_key":"bae-790e7e49-f2e3-5ad6-83d9-5dfb6d8ba81d",
```
fredcarle, line 80: suggestion: Since `_newKey` will be used to check if it's a self reference, I suggest we change `_key` for `_newKey` or have both.

tests/integration/backup/simple/import_test.go
```diff
@@ -151,3 +151,56 @@ func TestBackupImport_WithMultipleNoKeys_NoError(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	executeTestCase(t, test)
+}
+
+func TestBackupImport_WithMultipleNoKeysAndInvalidField_NoError(t *testing.T) {
```
fredcarle, line 180: suggestion: Change function name to end with `_ReturnError` as that is what is expected of the function under test.

tests/integration/backup/self_reference/import_test.go
```diff
@@ -0,0 +1,303 @@
...
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, []string{"Author", "Book"}, test)
+}
+
+// This test documents undesirable behaviour, as the documents are not linked.
+// https://github.com/sourcenetwork/defradb/issues/1697
+func TestBackupSelfRefImport_SplitPrimaryRelationWithSecondCollection_NoError(t *testing.T) {
```
fredcarle, line 241: suggestion: This test would work if the import content was taken from the generated export. I suggest either it documents an invalid import content and we remove the linked issue above the function name or we change the test to use a valid import content.

tests/integration/backup/self_reference/import_test.go
AndrewSisley: > `_newKey` will be used to check if it's a self reference

It would?  I always saw it the other way round, as any exported relation id fields would contain the old key, not the new one.

Happy to have both.

- [ ] add _newKey

tests/integration/backup/simple/import_test.go
AndrewSisley: My bad, will change, thanks

- [ ] Fix test name

tests/integration/backup/self_reference/import_test.go
AndrewSisley: That's fair, and it is more descriptive, even if a little longer/more-complex. Will change.

- [ ] import an export

tests/integration/backup/self_reference/import_test.go
fredcarle: The relation ID will be updated on export.

AndrewSisley: Converted to the import-export flow suggested in another comment

tests/integration/backup/self_reference/import_test.go
```diff
@@ -0,0 +1,365 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, []string{"Author", "Book"}, test)
+}
+
+// This test documents undesirable behaviour, as the documents are not linked.
+// https://github.com/sourcenetwork/defradb/issues/1697
```
fredcarle, line 267: note: This test isn't linked to an import error. It's an export error. It's having trouble setting the right expected book ID because they both have a primary field related to each other.

fredcarle (approved):
LGTM. Just notice my note on the import error that you linked.

tests/integration/backup/self_reference/import_test.go
AndrewSisley: Ah nice spot - would be good to fix that too if you have time, I'm not sure your fix-PR solves for this kind of schema anyway and it is hard to test fully independently of the import

fredcarle: We can't test it on import. It has to be tested on export. The json object representing the exported data is clearly wrong so we could correct it manually and assert an error instead, Showing a wrong output after import is fine. Just a little less clear as to what the problem is.

AndrewSisley: Ah true - good shout.  Tweaking now, will like to a new ticket

PR #1701, opened 2023-07-24
fix(i): Add self reference import handling

### LINKED ISSUE #1697: Backup-Import cannot handle self referencing documents

fredcarle:
## Relevant issue(s)

Resolves #1697 

## Description

This PR add the capability to handle self referencing docs on import.

## Tasks

## How has this been tested?

added integration tests

- MacOS

### DIFF
--- a/db/backup.go
+++ b/db/backup.go
@@ -69,6 +69,19 @@ func (db *db) basicImport(ctx context.Context, txn datastore.Txn, filepath strin
 				return NewErrJSONDecode(err)
 			}
 
+			// check if self referencing and remove from docMap for key creation
+			resetMap := map[string]any{}
+			for _, field := range col.Schema().Fields {
+				if field.Kind == client.FieldKind_FOREIGN_OBJECT {
+					if val, ok := docMap[field.Name+request.RelatedObjectID]; ok {
+						if docMap["_newKey"] == val {
+							resetMap[field.Name+request.RelatedObjectID] = val
+							delete(docMap, field.Name+request.RelatedObjectID)
+						}
+					}
+				}
+			}
+
 			delete(docMap, "_key")
 			delete(docMap, "_newKey")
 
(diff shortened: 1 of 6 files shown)


db/backup.go
```diff
@@ -69,6 +69,19 @@ func (db *db) basicImport(ctx context.Context, txn datastore.Txn, filepath strin
 				return NewErrJSONDecode(err)
 			}
 
+			// check if self referencing and remove from docMap for key creation
+			resetMap := map[string]any{}
```
AndrewSisley, line 73: todo: This does not look like it will work with `TestBackupSelfRefImport_SplitPrimaryRelationWithSecondCollection_NoError` - your `TestBackupImport_DoubleReletionshipWithUpdate_NoError` test has defined the primary relations all on one object meaning it is not really self-referencing, although it is a gap I have missed in my PR and should be kept.

Either way I think you should wait for that PR to be merged in order to make use of the tests that flagged this error to be sure it  is fixed.

AndrewSisley (changes requested):
I'm not sure this fully solves the issue, and I think it should wait for the more comprehensive tests added in the PR that found the bug.

db/backup.go
fredcarle: It won't but it's a different problem. The error in `TestBackupSelfRefImport_SplitPrimaryRelationWithSecondCollection_NoError` is on export as explained in the note I left on your PR.

AndrewSisley: (noted over there, but repeating here for ease of reference)

Was a really good spot by you there, but I still don't think the import code will work once the export is fixed.

tests/integration/backup/one_to_one/import_test.go
```diff
@@ -237,3 +237,58 @@ func TestBackupImport_WithMultipleNoKeyAndMultipleCollectionsAndMultipleUpdatedD
 
 	executeTestCase(t, test)
 }
+
+func TestBackupImport_DoubleReletionshipWithUpdate_NoError(t *testing.T) {
```
shahzadlone, line 241: nitpick:
```suggestion
func TestBackupImport_DoubleRelationshipWithUpdate_NoError(t *testing.T) {
```

jsimnz (approved):
Lgtm. Didn't dive too much into the root problem so my review is mostly on the superficial code quality aspect.

PR #1712, opened 2023-07-25
fix: Fix os.Exit(1) on badger close

### LINKED ISSUE #1702: TestQueryOneToOneRelations flaky in CI
`TestQueryOneToOneRelations` often fails on the change detector on the CI. It's not clear yet what is causing this.

It looks to be #1672 - only now flaky instead of consistently failing (and on a different, but similar test).

Possibly significant is that both this and #1672 involved tests with one-one relations.

- Executing `TestQueryOneToOneRelations` only does not resolve the issue (other tests do not appear to be a factor)
- Removing the create statements for either `Publishers` or `Authors` appears to remove the issue
- Reducing the query to just `Book {name}` appears to resolve the issue - this is unlike #1672 which persisted on a simplified query
     - Reducing the query to `Book {name author {name}}` (removing the publisher join) also resolves the issue
     - Reducing the query to `Book {name publisher {name}}` (removing the author join) *does not* resolve the issue. This interestingly is the one-one join (same as #1672)
- Setting `n.subtype.Spans()` does not resolve the issue (is a difference between the primary and secondary)
- No errors in `valuesSecondary` appear to be generated, despite #1711
- Returning early from `valuesSecondary` immediately after `n.subType.Init()` resolves the issue, this is before `n.subType.Next()`
- Returning early from `valuesSecondary` immediately after `n.subType.Next()` does not resolve the issue, suggesting `Next` is the trigger (same as in #1672)
- Removing `n.subType.Init()` from `typeJoinOne.Init()` does not resolve the issue (although it is a duplicate, pointless call I think)
- Updating to badger v4 still has no affect
- ~~Removing the 'extra' `df.kvResultsIter.Close()` call added to the fetcher in #1672  appears to resolve the issue~~ It did for a while, and then broke following a git push

AndrewSisley:
## Relevant issue(s)

Resolves #1702

## Description

Fixes os.Exit(1) on badger close by manually triggering the badger GC before closing.  Seems to solve the problem.

Todo:
- [ ] Drop the temp commits once happy is consistently passing in the CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,7 +273,7 @@ test\:coverage-html:
 .PHONY: test\:changes
 test\:changes:
 	@$(MAKE) deps:lens
-	env DEFRA_DETECT_DATABASE_CHANGES=true gotestsum -- ./tests/integration/query/one_to_many_to_one --run TestQueryOneToOneRelations -shuffle=on -p 1
+	env DEFRA_DETECT_DATABASE_CHANGES=true DEFRA_TARGET_BRANCH=1702-flaky-close-issue gotestsum -- ./tests/integration/query/one_to_many_to_one --run TestQueryOneToOneRelations -shuffle=on -p 1
 
 .PHONY: validate\:codecov
 validate\:codecov:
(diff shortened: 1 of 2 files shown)


AndrewSisley:
I think I got lucky locally (20 odd passes in a row), fix seems no good now

Issue #1637, opened 2023-07-14
Add fieldFetched metric to @explain


islamaliev:
As each document in defradb consist of several fields with each of them being stored separately. Hence fetching 1 doc that has only 1 field and fetching 1 doc that has dozens of fields has a big difference.
That's why it would be a valuable information to add the number of fields fetched to the @explain query.

PR #1713, opened 2023-07-26
feat: Make fetcher calculate docFetches and fieldFetches

### LINKED ISSUE #1636: Fix docFetched metric for @explain
At the moment docFetched metric is counter (incremented) in the scan node (scan node calls the fetcher, when it's done scan node increments).
The problem is that the fetcher can perform any number of doc fetches inside before it returns.
That means the fetcher should be responsible for counting the number of docs fetched.

islamaliev:
## Relevant issue(s)

Resolves #1636 and #1637

## Description

With this change the calculation of stats for fetching documents is moved to fetcher.

This was necessary because each call to `fetcher.FetchNext...` could potentially perform multiple doc fetches.
But scanNode (previously responsible for this) would just increment by 1.
Moreover, scanNode would incorrectly count end of fetching (returned nil doc) as 1 as well.

Also this fix introduces another (probably more important) metric `fieldFetches` which better indicates 
the execution cost because it much closer correlates with number of IO operations, as eace document's
field is stored in a separate KV pair.

## Tasks

## How has this been tested?

Integration tests

- MacOS


db/fetcher/fetcher.go
```diff
@@ -27,6 +27,23 @@ import (
...
+	return Stats{
+		DocsFetched:   s.DocsFetched + other.DocsFetched,
+		FieldsFetched: s.FieldsFetched + other.FieldsFetched,
+	}
+}
+
+func (s *Stats) Reset() {
+	s.DocsFetched = 0
+	s.FieldsFetched = 0
+}
```
fredcarle, line 45: todo: Since these are public types, please add documentation :)

planner/scan.go
```diff
@@ -21,12 +21,13 @@ import (
...
 
+// scanExecInfo contains information about the execution of a scan.
 type scanExecInfo struct {
 	// Total number of times scan was issued.
 	iterations uint64
 
-	// Total number of times attempted to fetch documents.
-	docFetches uint64
+	// Information about fetches.
+	fetches fetcher.Stats
```
shahzadlone, line 30: suggestion: Rename

```suggestion
	fetches fetcher.ExecInfo
```

fredcarle (approved):
LGTM!

db/fetcher/fetcher.go
```diff
@@ -27,6 +27,23 @@ import (
 	"github.com/sourcenetwork/defradb/request/graphql/parser"
 )
 
+type Stats struct {
```
shahzadlone, line 31: suggestion: Perhaps more consistent naming, I wanted to avoid the use of the word `metrics` specifically as metrics are a bit more higher level and now days mean something specific i.e usage with instrumentation of histograms, gauge and etc. For consistency can name this `ExecInfo`. Or if you want to keep the word Stats (I don't mind this word) then perhaps `ExecStats` or `ExecutionStats` and rename all other instances where we use the `ExecInfo` terminology.

tests/integration/explain/execute/query_deleted_docs_test.go
```diff
@@ -0,0 +1,70 @@
...
+package test_explain_execute
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	explainUtils "github.com/sourcenetwork/defradb/tests/integration/explain"
+)
+
+func TestExecuteExplainQueryDeletedDocs(t *testing.T) {
```
shahzadlone, line 20: praise: Appreciate you adding this one!

db/fetcher/fetcher.go
```diff
@@ -617,45 +646,41 @@ func (df *DocumentFetcher) FetchNextDoc(
...
+			if df.kvEnd ||
+				(df.reverse && ddf.kv.Key.DocKey > df.kv.Key.DocKey) ||
+				(!df.reverse && ddf.kv.Key.DocKey < df.kv.Key.DocKey) {
+				var stats Stats
+				encdoc, stats, err = ddf.FetchNext(ctx)
+				if err != nil {
+					return nil, core.Doc{}, Stats{}, err
 				}
+				status = client.Deleted
+				resultStats = resultStats.Add(stats)
```
shahzadlone, line 658: praise: Nice refactor

db/fetcher/fetcher.go
```diff
@@ -27,6 +27,28 @@ import (
...
+	FieldsFetched uint64
+}
+
+// Add adds the other stats to the current stats.
+func (s Stats) Add(other Stats) Stats {
+	return Stats{
+		DocsFetched:   s.DocsFetched + other.DocsFetched,
+		FieldsFetched: s.FieldsFetched + other.FieldsFetched,
+	}
+}
```
shahzadlone, line 44: todo: Can just mutate `s`, similar to `Reset()`. Then update the usage in other places, as you are reassigning back to `s` anyway in all your usages.

```suggestion
func (s *Stats) Add(other Stats) {
	s.DocsFetched = s.DocsFetched + other.DocsFetched
	s.FieldsFetched = s.FieldsFetched + other.FieldsFetched
}
```

shahzadlone (changes requested):
I would argue this isn't a `fix`, this is either a `refactor` or `feat`.

Thanks for bringing back the granularity of fetching datapoints :) left some comments.

PR #1716, opened 2023-07-26
fix: Return errors from typeJoinOne

### LINKED ISSUE #1711: typeJoinOne silently discards errors in `Next`
`valuesPrimary` and `valuesSecondary` do this, and there appears to be no good reason for it (likely historical)

AndrewSisley:
## Relevant issue(s)

Resolves #1711

## Description

Returns errors from typeJoinOne instead of ignoring them.

I assume this is a historical accident.

### DIFF
--- a/planner/type_join.go
+++ b/planner/type_join.go
@@ -16,7 +16,6 @@ import (
 	"github.com/sourcenetwork/defradb/connor"
 	"github.com/sourcenetwork/defradb/core"
 	"github.com/sourcenetwork/defradb/db/base"
-	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/planner/mapper"
 	"github.com/sourcenetwork/defradb/request/graphql/schema"
 )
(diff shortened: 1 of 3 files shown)


planner/type_join.go
```diff
@@ -367,32 +373,31 @@ func (n *typeJoinOne) valuesSecondary(doc core.Doc) core.Doc {
...
 	if err != nil {
-		return core.Doc{}
+		return core.Doc{}, err
 	}
 
 	// We have to reset the scan node after appending the new key-filter
 	if err := n.subType.Init(); err != nil {
-		log.ErrorE(n.p.ctx, "Sub-type initialization error at scan node reset", err)
-		return doc
+		return doc, errors.Wrap("sub-type initialization error at scan node reset", err)
```
fredcarle, line 381: todo: Add to `errors.go`

planner/type_join.go
```diff
@@ -408,8 +413,7 @@ func (n *typeJoinOne) valuesPrimary(doc core.Doc) core.Doc {
 
 	// re-initialize the sub type plan
 	if err := n.subType.Init(); err != nil {
-		log.ErrorE(n.p.ctx, "Sub-type initialization error at scan node reset", err)
-		return doc
+		return doc, errors.Wrap("sub-type initialization error at scan node reset", err)
```
fredcarle, line 416: todo: After adding the above to `errors.go`, you can use the defined error function here :)

planner/type_join.go
AndrewSisley: sorted

planner/type_join.go
AndrewSisley: sorted

PR #1718, opened 2023-07-26
test(i): Remove collection names param from ExecuteTestCase

### LINKED ISSUE #1717: Remove integration test collectionNames param
It is annoying and ugly to have to define this when the info is all in the testCase.

AndrewSisley:
## Relevant issue(s)

Resolves #1717

## Description

Removes collection names param from testUtils.ExecuteTestCase and instead figures stuff out from the test actions.

It has irked me for a while that we had to declare this, and it is only going to keep getting harder and harder to remove it as we add more tests.

I have a preference to use simple string stuff instead of any ast magic for this, as we'd be using the production code under test to test itself, and it is a bit heavy going for now.  If the basic string stuff becomes a hassle we can switch over to ast stuff or similar without too much trouble at a later date.

~~This is opened as a draft PR as I have 375 references to update and I'd prefer to get feedback on this before updating all of them~~

~~`ExecuteTEMP` is temporary, allowing me to keep testing as I update all the refs, instead of doing all 365 before realising I have messed something up.  Once all the refs have been updated `ExecuteTestCase` will lose it's collectionsNames param, and I will find+replace all references to `ExecuteTEMP` back to `ExecuteTestCase` and then delete `ExecuteTEMP` before converting this draft to a full PR.~~

Todo:
- [x] Remove `ExecuteTEMP`

### DIFF
--- a/tests/integration/utils2.go
+++ b/tests/integration/utils2.go
@@ -267,6 +267,20 @@ func GetDatabase(ctx context.Context, t *testing.T, dbt DatabaseType) (client.DB
 //
 // Will also attempt to detect incompatible changes in the persisted data if
 // configured to do so (the CI will do so, but disabled by default as it is slow).
+func ExecuteTEMP(
+	t *testing.T,
+	testCase TestCase,
+) {
+	collectionNames := getCollectionNames(testCase)
+
+	ExecuteTestCase(t, collectionNames, testCase)
+}
+
+// ExecuteTestCase executes the given TestCase against the configured database
+// instances.
+//
+// Will also attempt to detect incompatible changes in the persisted data if
+// configured to do so (the CI will do so, but disabled by default as it is slow).
 func ExecuteTestCase(
 	t *testing.T,
 	collectionNames []string,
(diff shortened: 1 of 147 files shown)


tests/integration/utils2.go
```diff
@@ -262,6 +262,20 @@ func GetDatabase(ctx context.Context, t *testing.T, dbt DatabaseType) (client.DB
 	return nil, "", nil
 }
 
+// ExecuteTestCase executes the given TestCase against the configured database
+// instances.
+//
+// Will also attempt to detect incompatible changes in the persisted data if
+// configured to do so (the CI will do so, but disabled by default as it is slow).
+func ExecuteTEMP(
```
AndrewSisley, line 270: note: Please read PR descriptino before commenting on this function name :)

tests/integration/utils2.go
```diff
@@ -417,6 +418,57 @@ func executeTestCase(
...
+
+	for _, a := range testCase.Actions {
+		switch action := a.(type) {
+		case SchemaUpdate:
+			if action.ExpectedError != "" {
+				// If an error is expected then no collections should result from this action
+				continue
+			}
+
+			splitByType := strings.Split(action.Schema, "type ")
```
nasdf, line 441: This is an edge case but the following schema type will cause problems here.

```
type user_type {
  name: String
}
```

nasdf (commented):
This is a great usability change! Only issue is the name parsing code has an edge case.

tests/integration/utils2.go
fredcarle: Since this is in our test suite and we define the type names, I don't mind this too much. Not sure it's worth trying to be clever in this case.

shahzadlone:
The pr label can be `test(i)` instead of `test`

AndrewSisley:
> The pr label can be `test(i)` instead of `test`

Good shout, cheers Shahzad - I need to get into the habit of remembering that :)

tests/integration/utils2.go
AndrewSisley: Cheers Guys, that was a silly mistake of mine :)

> Since this is in our test suite and we define the type names, I don't mind this too much. Not sure it's worth trying to be clever in this case.

I think I agree with you, but I'll document this limitation somewhere obvious so it doesn't cost us too much time if we ever feel the need to define a collection of such a name. 

- [ ] Doc limitation

source-devs (approved):
Shahzad: LGTM

Sorry didn't realize I was signed into the `source-devs` account

tests/integration/utils2.go
AndrewSisley: @nasdf Give me a shout if you are not happy with the documentation-only-for-now approach and I can tweak this post merge

Issue #1719, opened 2023-07-27
Handle downgrading of datastore versions of P2P synced docs


AndrewSisley:
Part of #1002

PR #1720, opened 2023-07-27
fix(i): Allow documents synced from P2P to migrate

### LINKED ISSUE #1591: Use Datastore version with Lens when going across P2P
The P2P system is currently unaware of the schema version of documents transiting through it. This means that upon arrival Defra assumes they are on the latest version and does not migrate them.  This is incorrect behaviour and should be fixed.

This also means that downgrading a schema version is not currently testable, as the only route through which it would be possible is when syncing a document with a higher schema version to a node on a lower version via P2P.

AndrewSisley:
## Relevant issue(s)

Resolves #1591

## Description

Allows documents synced from P2P to migrate (up) by storing the correct schema version id in the datastore on sync.

### DIFF
--- a/tests/integration/schema/migrations/query/with_p2p_test.go
+++ b/tests/integration/schema/migrations/query/with_p2p_test.go
@@ -30,6 +30,7 @@ func TestSchemaMigrationQueryWithP2PReplicatedDocAtOlderSchemaVersion(t *testing
 				Schema: `
 					type Users {
 						name: String
+						verified: Boolean
 					}
 				`,
 			},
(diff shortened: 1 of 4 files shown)


tests/integration/schema/migrations/query/with_p2p_test.go
```diff
@@ -38,15 +39,15 @@ func TestSchemaMigrationQueryWithP2PReplicatedDocAtOlderSchemaVersion(t *testing
 				NodeID: immutable.Some(1),
 				Patch: `
 					[
-						{ "op": "add", "path": "/Users/Schema/Fields/-", "value": {"Name": "verified", "Kind": "Boolean"} }
+						{ "op": "add", "path": "/Users/Schema/Fields/-", "value": {"Name": "email", "Kind": "String"} }
```
AndrewSisley, line 42: note: This test setup was changed to allow for an easier pairing when downgrading documents in #1719

shahzadlone (approved):
LGTM! However I must say that, even though I understand conceptually the addition of `unknownVersionPipe`, but don't fully understand the `how` it's going to be used.

AndrewSisley:
> LGTM! However I must say that, even though I understand conceptually the addition of `unknownVersionPipe`, but don't fully understand the `how` it's going to be used.

If P2P syncs a doc over of an unknown version, it will be fed into that.  It does this now.  This ensures it doesn't skip ahead and mess with the order of anything still lingering in the pipe-system

PR #1721, opened 2023-07-27
feat: Enable downgrading of documents via Lens inverses

### LINKED ISSUE #1719: Handle downgrading of datastore versions of P2P synced docs

AndrewSisley:
## Relevant issue(s)

Resolves #1719

## Description

Enables downgrading of documents via Lens inverses.  Inverse migrations are optional.

This PR is based off of https://github.com/sourcenetwork/defradb/pull/1720 - please review only the last 2 commits here.

There are so far fewer tests than I thought there would be, but I am a bit tired atm, and I may add more whilst waiting on review.  Please suggest any where you think there may be gaps.

~~This IMO breaches the code freeze, but it could be very nice to get it in the release if everyone is happy.~~

Discussed in standup, this should not be merged before the release.

### DIFF
--- a/lens/lens.go
+++ b/lens/lens.go
@@ -182,18 +182,14 @@ func (l *lens) Next() (bool, error) {
 
 				historyLocation = historyLocation.next.Value()
 			} else {
-				// The pipe head then becomes the schema version migration to the next version
-				// sourcing from any documents at schemaVersionID, or lower schema versions.
-				// This also ensures each document only passes through each migration once,
-				// in order, and through the same state container (in case migrations use state).
-				pipeHead, err = l.lensRegistry.MigrateDown(junctionPipe, historyLocation.schemaVersionID)
+				// Aquire a lens migration from the registery, using the junctionPipe as its source.
+				// The new pipeHead will then be connected as a source to the next migration-stage on
+				// the next loop.
+				pipeHead, err = l.lensRegistry.MigrateDown(junctionPipe, historyLocation.previous.Value().schemaVersionID)
 				if err != nil {
 					return false, err
 				}
 
-				// Aquire a lens migration from the registery, using the junctionPipe as its source.
-				// The new pipeHead will then be connected as a source to the next migration-stage on
-				// the next loop.
 				historyLocation = historyLocation.previous.Value()
 			}
 		}
--- a/lens/registry.go
+++ b/lens/registry.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/ipfs/go-datastore/query"
 	"github.com/lens-vm/lens/host-go/config"
+	"github.com/lens-vm/lens/host-go/config/model"
 	"github.com/lens-vm/lens/host-go/engine/module"
 	"github.com/lens-vm/lens/host-go/runtimes/wazero"
 	"github.com/sourcenetwork/immutable"
@@ -45,7 +46,8 @@ type lensRegistry struct {
 	modulesByPath map[string]module.Module
 	moduleLock    sync.Mutex
 
-	lensPoolsBySchemaVersionID map[string]*lensPool
+	lensPoolsBySchemaVersionID     map[string]*lensPool
+	reversedPoolsBySchemaVersionID map[string]*lensPool
 
 	// lens configurations by source schema version ID
 	configs map[string]client.LensConfig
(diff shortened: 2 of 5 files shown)


lens/lens.go
```diff
@@ -128,7 +130,18 @@ func (l *lens) Next() (bool, error) {
 		// up to the output via any intermediary pipes.
 		inputPipe = p
 	} else {
-		historyLocation := l.schemaVersionHistory[doc.SchemaVersionID]
+		historyLocation, ok := l.schemaVersionHistory[doc.SchemaVersionID]
+		if !ok {
```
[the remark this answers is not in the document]
AndrewSisley: This is already tested, for example by `TestP2PPeerUpdateWithNewFieldSyncsDocsToOlderSchemaVersionMultistep`.  The commit message went into why this is changing now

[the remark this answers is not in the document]
fredcarle: The commit message didn't help me understand this sorry.

If the result of `TestP2PPeerUpdateWithNewFieldSyncsDocsToOlderSchemaVersionMultistep` is the same with and without this change, then it's hard to know that this is being tested and that it's even needed (I understand why it's needed conceptually).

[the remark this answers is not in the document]
AndrewSisley: > The commit message didn't help me understand this sorry.

Ah okay - I'll try add more next time :P It is a non-obvious change, essentially pre-empting the next commit breaking things

lens/lens.go
```diff
@@ -182,18 +182,14 @@ func (l *lens) Next() (bool, error) {
...
 			} else {
-				// The pipe head then becomes the schema version migration to the next version
-				// sourcing from any documents at schemaVersionID, or lower schema versions.
-				// This also ensures each document only passes through each migration once,
-				// in order, and through the same state container (in case migrations use state).
-				pipeHead, err = l.lensRegistry.MigrateDown(junctionPipe, historyLocation.schemaVersionID)
+				// Aquire a lens migration from the registery, using the junctionPipe as its source.
+				// The new pipeHead will then be connected as a source to the next migration-stage on
+				// the next loop.
+				pipeHead, err = l.lensRegistry.MigrateDown(junctionPipe, historyLocation.previous.Value().schemaVersionID)
```
shahzadlone, line 188: question: After this line does `historyLocation.previous.Value()` change? if not why not move the declaration of `historyLocation` from below to above here and use here.

AndrewSisley: `Value()` is a very cheap getter, and I probably thought nothing about calling it twice.

I probably bounce around between which I prefer here, as a lot of the time I will prefer the double call over having an extra variable floating around.

lens/registry.go
```diff
@@ -98,9 +101,53 @@ func (r *lensRegistry) SetMigration(ctx context.Context, txn datastore.Txn, cfg
 }
 
 func (r *lensRegistry) cacheLens(txn datastore.Txn, cfg client.LensConfig) error {
-	locker, lockerAlreadyExists := r.lensPoolsBySchemaVersionID[cfg.SourceSchemaVersionID]
-	if !lockerAlreadyExists {
-		locker = r.newPool(r.poolSize, cfg)
+	inversedModuleCfgs := make([]model.LensModule, len(cfg.Lenses))
+	for i, moduleCfg := range cfg.Lenses {
+		// Reverse the order of the lenses for the inverse migration.
+		inversedModuleCfgs[len(cfg.Lenses)-i-1] = model.LensModule{
```
shahzadlone, line 107: question: If `len(cfg.Lenses) == 1` then wouldn't this be `inversedModuleCfgs[-1]` for first index (`i=0`)?

lens/lens.go
shahzadlone: Is totes fine either way, just thought if there is a var then to use it (or remove it, and call `Value()` everywhere)

shahzadlone (approved):
Looks good to best of my ability as I don't understand some chunks in depth, specially rust code haha.

lens/registry.go
AndrewSisley: :) I don't think so:

`len(cfg.Lenses) = 1`
`i = 0`
Results in:
`inversedModuleCfgs[1-0-1]`

AndrewSisley:
> Looks good to best of my ability as I don't understand some chunks in depth, specially rust code haha.

Lol, the Rust code isn't really important to review, is just an example/test lens.  Is worth at the func signature though, as it show cases that the inverse is optional, and lens devs don't actually need to declare them

lens/registry.go
shahzadlone: I needed sleep, rip haha

PR #1739, opened 2023-08-01
tools(i): Migrate mockery to use package level config file

### LINKED ISSUE #1735: Mockery v2 is deprecated, move to packages configuration.
We are only the last version that is supported by the non-package level legacy configuration by mockery. Move to the new packages configuration as it is much faster and easier to manage.

From their github repo:

[v3](https://github.com/vektra/mockery/projects/3) will include a ground-up overhaul of the entire codebase and will completely change how mockery works internally and externally. The highlights of the project are:

Moving towards a package-based model instead of a file-based model. mockery currently iterates over every file in a project and calls package.Load on each one, which is time-consuming. Moving towards a model where the entire package is loaded at once will dramatically reduce runtime, and will simplify logic. Additionally, supporting only a single mode of operation (package mode) will greatly increase the intuitiveness of the software.
Configuration-driven generation. v3 will be entirely driven by configuration, meaning:
You specify the packages you want mocked, instead of relying on it auto-discovering your package. Auto-discovery in theory sounds great, but in practice it leads to a great amount of complexity for very little benefit.
Package- or interface-specific overrides can be given that change mock generation settings on a granular level. This will allow your mocks to be generated in a heterogeneous manner, and will be made explicit by YAML configuration.
Proper error reporting. Errors across the board will be done in accordance with modern Golang practices
Variables in generated mocks will be given meaningful names.

shahzadlone:
## Relevant issue(s)
Resolves #1735 

## Description
- Setup the new and faster mocking configuration.
- Call the configuration from `Makefile`.
- Removes old mocks.
- Adds new mocks (biggest difference is file names and removal of mockery version string from the header).

## How has this been tested?
Locally, and CI

- Arch Linux


shahzadlone:
question: I don't know if this was available before as well, but I see now that most usage examples Prepend(or Append if we want) `Mock` to the created mock object. Example instead of:
```
type Collection struct {
	mock.Mock
}
```

```
type MockCollection struct {
	mock.Mock
}
```

Do we want that?

Another question, the files also by the default configurations does add `mock_` before the file name, is that something we want?
Example: instead of `db.go` having `mocks_db.go`

AndrewSisley (approved):
LGTM :)

PR #1740, opened 2023-08-01
chore: Update Badger version to v4

### LINKED ISSUE #1702: TestQueryOneToOneRelations flaky in CI

fredcarle:
## Relevant issue(s)

Resolves #1702

## Description

This PR updates the Badger version to v4.

It includes a temporary fix to our badger `os.Exit` issue that is costing us time when managing PRs. This should be reverted or properly fixed before releasing v0.7.

## Tasks

## How has this been tested?

multiple runs of `env DEFRA_DETECT_DATABASE_CHANGES=true gotestsum -- ./tests/integration/query/one_to_many_to_one -shuffle=on -p 1`

- MacOS


go.mod
```diff
@@ -202,6 +201,7 @@ require (
 
 // SourceNetwork fork og graphql-go
 replace (
+	github.com/dgraph-io/badger/v4 => github.com/fredcarle/badger/v4 v4.0.0-20230801145501-d3a57bd4c2ec
```
AndrewSisley, line 204: question: I thought we agreed to use a source fork instead of a personal one? I do have a preference for a sourcenetwork fork.

AndrewSisley:
question: Has the fork been reviewed?

go.mod
fredcarle: Used mine because I didn't have the rights to push to our fork. I can update it now that I do have rights but it's temporary anyways.

fredcarle:
> question: Has the fork been reviewed?

Initial review has been done and it's not going to be the way to fix this but the badger team is also looking into it. Hopefully we get a proper fix before the end of this cycle.

go.mod
AndrewSisley: Cheers :)

AndrewSisley (approved):
LGTM, thanks for switching to the source fork :)

PR #1768, opened 2023-08-10
fix: Handle the querying of secondary relation id fields

### LINKED ISSUE #1654: Secondary relation id is unavailable when querying one-one relations
See documenting tests:
- `TestQueryOneToOneWithGroupRelatedIDFromSecondary`
- `TestQueryOneToOneWithGroupRelatedIDAliasFromSecondary`

Likely affects all stuff (order, filter) not just group by.
- https://github.com/sourcenetwork/defradb/issues/1670
- https://github.com/sourcenetwork/defradb/issues/1671

AndrewSisley:
## Relevant issue(s)

Resolves #1654

## Description

Handles the querying of secondary relation id fields.

It does this by joining the related object and copying the dockey value into the secondary related id field.

- ~~Is currently blocked by https://github.com/sourcenetwork/defradb/issues/1769 but can still be reviewed now~~ ~~PR is now based off the fix for that issue (do not review the first commit here).  Will rebase off develop branch before merge.~~ Fix merged to develop, this branch has been rebased onto it.

### DIFF
--- a/planner/mapper/mapper.go
+++ b/planner/mapper/mapper.go
@@ -804,39 +804,12 @@ sourceLoop:
 		propertyMapped := len(mapping.IndexesByName[key]) != 0
 
 		if !propertyMapped {
-			index := mapping.GetNextIndex()
-
-			dummyParsed := &request.Select{
-				Field: request.Field{
-					Name: key,
-				},
-			}
-
-			childCollectionName, err := getCollectionName(descriptionsRepo, dummyParsed, parentCollectionName)
+			dummyJoin, err := constructDummyJoin(descriptionsRepo, parentCollectionName, mapping, key)
 			if err != nil {
 				return nil, err
-			}
-
-			childMapping, _, err := getTopLevelInfo(descriptionsRepo, dummyParsed, childCollectionName)
-			if err != nil {
-				return nil, err
-			}
-			childMapping = childMapping.CloneWithoutRender()
-			mapping.SetChildAt(index, childMapping)
-
-			dummyJoin := &Select{
-				Targetable: Targetable{
-					Field: Field{
-						Index: index,
-						Name:  key,
-					},
-				},
-				CollectionName:  childCollectionName,
-				DocumentMapping: childMapping,
 			}
 
 			newFields = append(newFields, dummyJoin)
-			mapping.Add(index, key)
 		}
 
 		keyIndex := mapping.FirstIndexOfName(key)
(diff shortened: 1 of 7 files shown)


tests/integration/query/one_to_one/simple_test.go
fredcarle, line 1: suggestion: Can you please add a test in `one_to_one_to_one` with a query on `Book` joining both `Author` and `Publisher`?

AndrewSisley: I'm curious, not argumentative - how do you see that breaking/being-relevant?

fredcarle: I want a test confirming that two or more joins on the one object works as expected. Pretty much to ensure `makeTypeJoinOne` works in these kinds of situations.

AndrewSisley: As-in work in general? I hope we already have tests that properly cover one-one-one joins

AndrewSisley: Confirmed, we already cover all possible direction combinations for one-one-one joins.

fredcarle: Work in general yes and with this change. There aren't that many in the package `one_to_one_to_one` and none with the pattern I'd like to see. Maybe there are some elsewhere.

AndrewSisley: > none with the pattern I'd like to see

Which pattern are you after?

fredcarle:
```graphql
Book {
  name
  publisher {
    name
  }
  author {
    name
  }
}
```

AndrewSisley: Ah okay spreading out from the middle obj? I'll add that in another PR/branch real quick. But not here.

fredcarle: Well it's the "middle obj" in this case but what matters to me is the dual join on the one object. I think it's quite relevant here so I would like to see it added.

AndrewSisley: This PR doesn't change any of the join mechanics.  It uses a join, but it doesn't change how they work.

If a bug is found in the test gap you spotted it should not block or impact the merging of this work.

https://github.com/sourcenetwork/defradb/issues/1792

fredcarle: fair enough :)

planner/mapper/mapper.go
```diff
@@ -928,6 +906,131 @@ sourceLoop:
 	return newFields, nil
 }
 
+// constructDummyJoin constructs a valid empty join with no requested fields.
+func constructDummyJoin(
```
fredcarle, line 910: suggestion: I think we should avoid using the work `Dummy` as it feels more like something you'd use just for testing purposes. In this case this is a function that is used in a critical path. Maybe `constructEmptyJoin` would be more appropriate.

AndrewSisley: Yeah good shout - I'll change it

- [ ] Rename Dummy func

fredcarle (approved):
Approving with one nitpick.

Tests have been added with two different test name format. I don't want to force you to change that but keep in mind for the future that we agreed to try to have `TestName_Details_Error` as the test name format.

PR #1770, opened 2023-08-10
fix(i): Require secondary id field for one-ones

### LINKED ISSUE #1769: Patch one-one relation field to existing schema does not require both id fields
Sorry, I misread the code and only thought the primary was required.  The secondary is also needed.

Blocking https://github.com/sourcenetwork/defradb/pull/1768

AndrewSisley:
## Relevant issue(s)

Resolves #1769

## Description

Require secondary id field for one-ones.

Simple silly bug caused by developer misunderstanding. Simple fix.


shahzadlone (approved):
LGTM

PR #1776, opened 2023-08-11
refactor: HTTP client interface

nasdf:
## Description

This PR is split into two to make reviewing easier: #1839

This PR adds an HTTP client and server implementation that implements the `client.DB` interface and runs through the existing integration test suite.

## Tasks

## How has this been tested?

`make test`

- MacOS


datastore/memory/txn.go
```diff
@@ -123,9 +123,9 @@ func (t *basicTxn) GetSize(ctx context.Context, key ds.Key) (size int, err error
 
 // Has implements ds.Has.
 func (t *basicTxn) Has(ctx context.Context, key ds.Key) (exists bool, err error) {
-	t.ds.closeLk.RLock()
-	defer t.ds.closeLk.RUnlock()
-	if t.ds.closed {
+	t.closeLk.RLock()
```
AndrewSisley, line 126: For reference, for myself and other reviewers, the rest of the funcs in `basicTxn` do this (`t.closeLk`) and this appears to be a bug fix.

tests/integration/backup/one_to_one/import_test.go
```diff
@@ -237,7 +237,8 @@ func TestBackupImport_DoubleRelationshipWithUpdate_NoError(t *testing.T) {
 						},
 					},
 					{
-						"name": "Game of chains",
+						"name":   "Game of chains",
+						"author": nil,
```
AndrewSisley, line 241: question: Why has this been added?

nasdf: I believe this was due to the JSON serialization not dropping the nil field here. Potentially it could be an incorrect case within the new `resultsAreEqual` function.

nasdf: It turns out these test case changes are no longer necessary. I've reverted this and a few other changes.

tests/integration/explain.go
```diff
@@ -312,24 +313,41 @@ func findTargetNode(
 			}
 		}
 
+	case []any:
+		return findTargetNodeFromArray[any](targetName, toSkip, includeChildNodes, r)
+
```
shahzadlone, line 318: praise+question: interesting and this sorted out what you needed?

nasdf: Yeah we lose type info when serializing to and from JSON. This was the cleanest solution I could find for handling generic arrays.

tests/integration/backup/one_to_one/import_test.go
AndrewSisley: Nice, cheers Keenan

tests/integration/utils2.go
```diff
@@ -58,6 +61,13 @@ const (
 	badgerFileType DatabaseType = "badger-file-system"
 )
 
+type ClientType string
+
+const (
+	goClientType   ClientType = "go"
+	httpClientType ClientType = "http"
```
AndrewSisley, line 68: todo: I think the new consts would benefit from a line or two of documentation each

tests/integration/utils2.go
```diff
@@ -1652,13 +1707,7 @@ func assertRequestResults(
...
 	log.Info(ctx, "", logging.NewKV("RequestResults", result.Data))
 
-	// compare results
-	assert.Equal(t, len(expectedResults), len(resultantData), description)
-	if len(expectedResults) == 0 {
-		// Need `require` here otherwise will panic in the for loop that ranges over
-		// resultantData and tries to access expectedResults[0].
-		require.Equal(t, expectedResults, resultantData)
-	}
+	require.Equal(t, len(expectedResults), len(resultantData), description)
```
AndrewSisley, line 1710: praise: cheers :laughing:

tests/integration/results.go
```diff
@@ -0,0 +1,164 @@
+// Copyright 2023 Democratized Data Foundation
```
AndrewSisley, line 1: praise: Thanks for moving this to a new file, is much tidier, and I think Shahzad had been wanting this for a while :)

http/http_client.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (c *httpClient) request(req *http.Request) ([]byte, error) {
+	c.setDefaultHeaders(req)
+
+	res, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close() //nolint:errcheck
```
AndrewSisley, line 59: todo: I can't spot the reason as to why ignoring the error here is okay/desirable - can you please document this in the code with a comment if you feel like we want to do this, or handle the error.

http/handler_tx.go
```diff
@@ -0,0 +1,95 @@
...
+	db := req.Context().Value(dbContextKey).(client.DB)
+	txs := req.Context().Value(txsContextKey).(*sync.Map)
+	readOnly, _ := strconv.ParseBool(req.URL.Query().Get("read_only"))
+
+	tx, err := db.NewTxn(req.Context(), readOnly)
+	if err != nil {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+		return
+	}
+	txs.Store(tx.ID(), tx)
```
AndrewSisley, line 40: question: I can't spot anything that 'expires' long-lived transactions which have likely been forgotten by the client.  Is that true?

I think some datastore implementations do this (not fully sure thought), meaning that the server will think a txn is alive, when the datastore has decided it is dead - are we relying on this to minimise expenses and handle errors?

question: If the above is true, are we sure we are happy tolerating a forever growing map of txns in the short-medium term?

http/middleware.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2023 Democratized Data Foundation
```
AndrewSisley, line 1: praise: I really like how you have solved some of the complexities with Lens/Collections/txns/etc in this file :)

todo: I do think the documentation within this file needs to be expanded though, particularly on the stuff that is accessed from within the handlers (i.e. the `fooContextKey` variables).  It is not obvious when consuming them whether stuff like txns are respected.

http/handler_collection.go
```diff
@@ -0,0 +1,343 @@
...
+func (s *collectionHandler) Create(rw http.ResponseWriter, req *http.Request) {
+	col := req.Context().Value(colContextKey).(client.Collection)
+
+	var body any
+	if err := requestJSON(req, &body); err != nil {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+		return
+	}
+
+	switch t := body.(type) {
```
AndrewSisley, line 48: suggestion:  I found the split here to Create/CreateMany strange, and I had to check few a few files to be sure where and why it was like this.  I understand it in the short term, but still think it is strange, and think you are kind of breaking the `client.Collection` interface here - the `http.Client` type implements the interface, but the http endpoint does not - this will be visible to users when not using the Go http-client wrapper.

I suggest you add a new ticket to change the `client.Collection` interface to remove the `CreateMany` func.  Create would then need to handle 0..n docs - I would leave it to whoever picks up that ticket how to handle that (could be an `any` type, `...Document`, or a generic constrait even).

In the new ticket please reference this code area to ensure it gets cleaned up along the way.

http/client.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	var txRes CreateTxResponse
+	if err := c.http.requestJson(req, &txRes); err != nil {
+		return nil, err
+	}
+	return &Transaction{txRes.ID, c.http}, nil
+}
+
+func (c *Client) WithTxn(tx datastore.Txn) client.Store {
+	client := c.http.withTxn(tx.ID())
```
AndrewSisley, line 87: thought: I'm just highlighting here that our current transaction ID implementation is a bit of a hack, and exposing it more here increases the need to improve the way we manage these IDs.

http/client.go
```diff
@@ -0,0 +1,412 @@
...
+func (c *Client) execRequestSubscription(ctx context.Context, r io.ReadCloser) *events.Publisher[events.Update] {
+	pubCh := events.New[events.Update](0, 0)
+	pub, err := events.NewPublisher[events.Update](pubCh, 0)
+	if err != nil {
+		return nil
+	}
+
+	go func() {
+		eventReader := sse.NewReadCloser(r)
+		defer eventReader.Close() //nolint:errcheck
```
AndrewSisley, line 362: todo: Same as other comment, please handle the error or document why not

http/client.go
```diff
@@ -0,0 +1,412 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package http
```
AndrewSisley, line 11: question: The current (develop branch) `http` package has a playground.go file, I'm not familiar with the playground stuff at all, has this been forgotten about? Is playground.go deadcode? Have I just missed it somewhere in this PR?

AndrewSisley (commented):
I really like this, but have a few questions and todos for you before I'm ready to approve.

tests/integration/lens.go
```diff
@@ -75,9 +75,34 @@ func getMigrations(
 
 		configs, err := db.LensRegistry().Config(s.ctx)
 		require.NoError(s.t, err)
-
-		// The order of the results is not deterministic, so do not assert on the element
-		// locations.
-		assert.ElementsMatch(s.t, configs, action.ExpectedResults)
+		require.Equal(s.t, len(configs), len(action.ExpectedResults))
```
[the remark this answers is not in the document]
nasdf: This is due to the http client losing type info when serializing to json. The newly added `assertResultsEqual` function must be used to properly compare the two values.

tests/integration/query/one_to_many/simple_test.go
```diff
@@ -167,7 +167,7 @@ func TestQueryOneToManyWithNonExistantParent(t *testing.T) {
 			{
 				"name":   "Painted House",
 				"rating": 4.9,
-				"Author": nil,
+				"author": nil,
```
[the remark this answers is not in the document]
nasdf: It failed at first but now it passes with either version.

tests/integration/results.go
```diff
@@ -0,0 +1,164 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// resultsAreEqual returns true if the expected and actual results are of equal value.
+//
+// NOTE: Values of type json.Number and immutable.Option will be reduced to their underlying types.
+func resultsAreEqual(expected any, actual any) bool {
```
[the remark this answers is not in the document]
nasdf: This is due to the fact that we lose type info when serializing to json. The simple fix is to compare them using `assert.JSONEq`, but I don't think everyone would be happy with that solution.

tests/integration/lens.go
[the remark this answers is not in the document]
AndrewSisley: You mean the http stuff is returning values of a type other than the one expected in the test(s)?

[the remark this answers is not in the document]
nasdf: The underlying values are the same but the types differ slightly. 

A good example is when you serialize a number to json. It loses its type value of uint64 or float64 and becomes a string representation. When it's deserialized it is stored as an intermediary type `json.Number`. 

This allows the test suite to check the expected type and attempt to convert the `json.Number` to the correct type before comparing.

tests/integration/utils2.go
```diff
@@ -172,6 +188,11 @@ func init() {
 		badgerFile = false
 		inMemoryStore = true
 	}
+	// default is to run against all
+	if !goClient && !httpClient && !DetectDbChanges {
```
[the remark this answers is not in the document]
nasdf: I cleaned this up a bit and changed it to only enable the `goClient` when running the change detector.

tests/integration/utils2.go
nasdf: Added

http/http_client.go
nasdf: Added

tests/integration/lens.go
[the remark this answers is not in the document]
AndrewSisley: Asserting the type is pretty important, particularly for the embedded Go client.  Changing the return type is a breaking change.

It also at glance looks like you are equating zero and option.None too?  This is also incorrect, including within http/gql contexts (think zero attendees vs undefined/nil - zero represents the knowledge that no one is attending, nil means that it is unknown).

http/client.go
nasdf: Added comment

http/client.go
```diff
@@ -0,0 +1,412 @@
...
+	res, err := c.http.client.Do(req)
+	if err != nil {
+		result.GQL.Errors = []error{err}
+		return result
+	}
+	if res.Header.Get("Content-Type") == "text/event-stream" {
+		result.Pub = c.execRequestSubscription(ctx, res.Body)
+		return result
+	}
+	defer res.Body.Close() //nolint:errcheck
```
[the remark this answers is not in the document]
nasdf: Added comment

http/middleware.go
nasdf: Added more documentation clarifying what each context key will be set to.

http/handler_tx.go
nasdf: This is a great point. I think a good long term solution would be having a shared transaction pool within the database that allows the user to specify limits and timeouts.

This will also enable us to fix the "leaky" `datastore.Txn` interface. Exposing the `MultiStore` interface on transactions is currently a pain point for the http client.

In the short-term I think this will be okay. I think only advanced users will interact with the transaction system, and we can make a note of the current short-comings somewhere.

http/handler_collection.go
nasdf: I had made this trade-off to keep the API more RESTful.

I do like the idea of merging the `Create` and `CreateMany` methods.

http/client.go
nasdf: The follow up PR addresses this. 

https://github.com/sourcenetwork/defradb/blob/19c734bc7a6da1eba2e2ee108182a14bd41f5f39/http/handler_playground.go

tests/integration/lens.go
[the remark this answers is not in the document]
nasdf: I've updated the `assertResultsEqual` to use the new "relaxed" comparison only when the client type is not `goClientType`.

https://github.com/sourcenetwork/defradb/blob/5f96d78080d65af8b7b24e787e19f62c1b9b21c5/tests/integration/results.go#L26-L38

http/client.go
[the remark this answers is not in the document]
AndrewSisley: :) don't forget to push

http/handler_tx.go
AndrewSisley: Sounds sensible, sorry, I could of asked this earlier but didnt - could you please open a ticket to sort this out?

http/handler_collection.go
AndrewSisley: > I had made this trade-off to keep the API more RESTful.

I understood this :) Please don't forget to create that ticket though.

http/client.go
AndrewSisley: Cheers

tests/integration/lens.go
```diff
@@ -75,9 +75,34 @@ func getMigrations(
 
 		configs, err := db.LensRegistry().Config(s.ctx)
 		require.NoError(s.t, err)
-
-		// The order of the results is not deterministic, so do not assert on the element
```
AndrewSisley, line 79: todo: I think this comment still adds value and should be preserved somewhere.  It is not obvious reading the code as to why the order of the elements is not asserted.

http/client_lens.go
```diff
@@ -0,0 +1,133 @@
...
+	body, err := json.Marshal(src)
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, methodURL.String(), bytes.NewBuffer(body))
+	if err != nil {
+		return nil, err
+	}
+	var result enumerable.Enumerable[map[string]any]
+	if err := c.http.requestJson(req, &result); err != nil {
```
AndrewSisley, line 79: question: Are you 100% sure this works? You appear to be unmarshalling json to an interface.  If it does work, how, and what is the resultant concrete type?  If it does not work, did the tests miss this, or did you forget to run them?  If they missed it, how?

`MigrateUp`, `MigrateDown` on `LensRegistry` and `lensHandler` appear to do this.

tests/integration/lens.go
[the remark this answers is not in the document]
AndrewSisley: This still makes me uncomfortable:

1) The type assertion is opt-in, not opt-out, it would be very easy to accidentally not test the typing of new client interface implementations.  It would prefer that the `default` is `assert.Contains`.

2) The return types of all client interface implementations is important, and should be deterministic and consistent.  The Go client is just the most likely to be used directly by users.  That said, `http`, and later, `cli`, are new implementations, and it might be sensible to add a new ticket to the backlog to revisit this later, as to not block the main body of work in this PR.

tests/integration/utils2.go
```diff
@@ -131,59 +148,57 @@ var previousTestCaseTestName string
...
-}
 
-func getBool(val string) bool {
-	switch strings.ToLower(val) {
-	case "true":
-		return true
 	default:
-		return false
+		// Default is to test all client types.
+		if !goClient && !httpClient {
```
AndrewSisley, line 192: todo: I disagree with this default.  It will essentially double the stuff being done when running `go test` or `make test`, and will significantly degrade the dev-ex of anyone working in the codebase.

Can you please default to the go client only, and then maybe add a couple of make actions to make it easy to run http?

tests/integration/utils2.go
```diff
@@ -131,59 +148,57 @@ var previousTestCaseTestName string
...
-		// Testing against the file system is off by default
-		badgerFile = false
-		inMemoryStore = true
-	}
-
-	if DetectDbChanges {
+	// Set default values for the specified testing mode.
+	switch {
+	case DetectDbChanges:
+		// Change detector runs using only the go client type.
```
AndrewSisley, line 185: todo: I think the code should change.  Whilst we currently think that running the change detector via the http client is pointless, it is not technically impossible and we shouldnt prevent it.  The environment variables should still be respected, and it will cause confusion (and bugs) if they are not.

Can you please instead change the `make test:changes` action to instead explicitly run using just the Go client.  Regardless of what the default clients are (see other comment about defaults).

http/client.go
AndrewSisley: Please push it :)

http/middleware.go
AndrewSisley: Looks good - thanks Keenan

http/client_lens.go
AndrewSisley: I think I know why this was missed.

Previously we have been able to rely on the fact that we only ever had one implementation of the client interfaces, this allowed us to get away with not directly testing most functions implemented, so long as they were called internally  by functions that were directly tested.

These functions (as do most of them, our testing is horrible across the go-embedded stuff) fall into that category. They have no tests targeting them. 

I think this requires discussion at standup, as it means that significant chunks of this PR are untested, as will likely much of the CLI stuff be. There is a significant risk of other bugs besides the probable one here that I have flagged.

http/handler_tx.go
nasdf: Created issue #1843

AndrewSisley: Cheers :)

http/handler_collection.go
nasdf: Created issue #1844

AndrewSisley: Thanks :)

tests/integration/lens.go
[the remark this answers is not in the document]
nasdf: I've updated the default comparisons like you recommended.

I'm still thinking of a solution for your second point, but it will most likely be addressed in a follow up PR.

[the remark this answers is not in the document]
AndrewSisley: Cheers Keenan, and really no pressure to sort out (2) within this PR.

AndrewSisley (approved):
Cheers Keenan - looks really good :)  I'm approving but, would recommend you quickly check in with another dev too if any are interested as it is a significant change and they may have different opinions to mine (and catch stuff I have missed).

tests/integration/utils2.go
```diff
@@ -131,45 +148,46 @@ var previousTestCaseTestName string
...
-	SetupOnly = getBool(setupOnlyValue)
-
-	if !repositorySpecified {
+	httpClient, _ = strconv.ParseBool(os.Getenv(clientHttpEnvName))
+	goClient, _ = strconv.ParseBool(os.Getenv(clientGoEnvName))
+	badgerFile, _ = strconv.ParseBool(os.Getenv(fileBadgerEnvName))
+	badgerInMemory, _ = strconv.ParseBool(os.Getenv(memoryBadgerEnvName))
+	inMemoryStore, _ = strconv.ParseBool(os.Getenv(inMemoryEnvName))
+	DetectDbChanges, _ = strconv.ParseBool(os.Getenv(detectDbChangesEnvName))
+	SetupOnly, _ = strconv.ParseBool(os.Getenv(setupOnlyEnvName))
```
shahzadlone, line 157: question: what made you switch from `LookupEnv` to `Getenv`. 

IRRC `Getenv` returns an empty string, even in cases where the environment variable is not set and `LookupEnv` returns the value and a bool to indicate if it was set or not.

tests/integration/utils2.go
```diff
@@ -131,45 +148,46 @@ var previousTestCaseTestName string
...
+	var targetBranchValue string
+	if value, ok := os.LookupEnv(targetBranchEnvName); ok {
+		targetBranchValue = value
+	} else {
 		targetBranchValue = "develop"
 	}
 
-	if mutationTypeSpecified {
-		mutationType = MutationType(mutType)
+	if value, ok := os.LookupEnv(mutationTypeEnvName); ok {
```
shahzadlone, line 173: question: What was the benefit to moving the env lookup here? I did like how they were all in one place above.

shahzadlone (commented):
LGTM but I have some questions I asked and some non-blocking comments / thoughts:

1) The PR was quite easy to review, however I think it could have been even easier/clearer if:
- The testing framework changes could have been isolated in a separate PR.
- Rebase flow was used instead of the multiple merges from develop into your branch, which create those non-linear commit history with extra merge commits that aren't useful for reviewers.

2) CodeCov shows that code coverage has dropped by `-5.40%`, that is quite a lot, I am not quite sure why that is, I know sometimes it can be inaccurate and show a different number on merge, but usually it is not off by that much. This isn't a blocker but might be worth exploring / looking into it why that is.

shahzadlone:
@nasdf I noticed that everything under `http/` has 0 coverage, is that expected for now?
image

tests/integration/utils2.go
nasdf: We weren't using the second return value from `LookupEnv` for the boolean variables. The behavior should be the same but is a bit easier to read in my opinion.

tests/integration/utils2.go
nasdf: I think this makes it more obvious that we have a default value for these variables. We also don't have to keep track of extra variables since the lookup return values are now scoped to the if statement.

nasdf:
> @nasdf I noticed that everything under `http/` has 0 coverage, is that expected for now? image

I think it's not picking up the coverage from the integration tests correctly.

AndrewSisley:
> I think it's not picking up the coverage from the integration tests correctly.

It wont do, this is expected, as the codecov job will only be running with the go client for now.  We need to come up with a performant way of handling this, for both the multi-clients, the mutations, and change detector - I think we spoke about this in a recent standup.

tests/integration/utils2.go
shahzadlone: I thought we were using the second value for some of them. However I don't mind this minor change just wanted to know the reason.

tests/integration/utils2.go
shahzadlone: I think this is subjective, I do like if all the lookups are done in the beginning and then the condition flow is executed, rather than lookup and condition checking happening at the same step. But not a strong pref, so is a non-blocking one.

shahzadlone:
Sorry I had a small subjective nitpick I forgot to mention, regarding some function names (if these are added by you and not existent from before). I do like boolean returning functions to start with `is` or `are` to make it clear that a "question"/"equality" is being "answered"/"compared", for example:
- resultArraysAreEqual
- resultOptionsAreEqual
- resultsAreEqual
- resultsAreAnyOf


Could be:
- areResultArraysEqual
- areResultOptionsEqual
- areResultsEqual
- areResultsAnyOf

shahzadlone (approved):
Posted some nitpicks(non-blocking). All G otherwise

PR #1777, opened 2023-08-11
feat: Support PatchSchema relational field kind substitution

### LINKED ISSUE #1771: Extend PatchSchema field Kind substitution magic to foriegn relations
Part of #1002 

For example:
```
{ "op": "add", "path": "/Users/Schema/Fields/-", "value": {
	"Name": "foo", "Kind": "Users", "RelationType": 133, "RelationName": "foo"
}},
```
Should be converted to:
```
{ "op": "add", "path": "/Users/Schema/Fields/-", "value": {
	"Name": "foo", "Kind": 16, "RelationType": 133, "Schema": "Users", "RelationName": "foo"
}},
```
And similar for Many relations (e.g. `"Users"`)

AndrewSisley:
## Relevant issue(s)

Resolves #1771

## Description

Adds support for PatchSchema relational field `Kind` substitution, and the auto setting of the `Schema` property.

### DIFF
--- a/db/errors.go
+++ b/db/errors.go
@@ -45,6 +45,7 @@ const (
 	errInvalidCRDTType                    string = "only default or LWW (last writer wins) CRDT types are supported"
 	errCannotDeleteField                  string = "deleting an existing field is not supported"
 	errFieldKindNotFound                  string = "no type found for given name"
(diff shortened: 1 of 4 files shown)


jsimnz (approved):
LGTM! Awesome, with the follow up work, this will be really powerful!

PR #1788, opened 2023-08-14
refactor(i): Unify collection.save and .applyMerge logic

### LINKED ISSUE #1786: Unify collection.save and collection.applyMerge field logic
Is a notable source of bugs when stuff only gets done in one of them, and a time sink.

AndrewSisley:
## Relevant issue(s)

Resolves #1786

## Description

Unifies collection.save and collection.applyMerge field logic, leaving us with one function through which all document mutations take place.

Also fixes the slightly broken logic for determining whether a field was a foreign relation or not (relying on the field name ending in `_id` is not good enough).

Also expands the value types that `client.Document` can handle.  This is however not the goal of this PR, and they are undocumented (or poorly documented) and likely still incomplete.  Another work item can target this area at some other point in time.

### DIFF
--- a/client/document.go
+++ b/client/document.go
@@ -283,6 +283,12 @@ func (doc *Document) setAndParseType(field string, value any) error {
 		if err != nil {
 			return err
 		}
+	case uint64:
+		err := doc.setCBOR(LWW_REGISTER, field, int64(val))
+		if err != nil {
+			return err
+		}
+
 	case float64:
 		// case int64:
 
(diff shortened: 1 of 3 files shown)


client/document.go
```diff
@@ -300,7 +306,7 @@ func (doc *Document) setAndParseType(field string, value any) error {
 		}
 
 	// string, bool, and more
-	case string, bool, []any:
+	case string, bool, int64, []any, []bool, []*bool, []int64, []*int64, []float64, []*float64, []string, []*string:
```
fredcarle, line 309: todo: Please add a separate case for int64. Note that the type of val here will be `any`.

fredcarle (approved):
Minor todo. Otherwise looks good to me.

client/document.go
fredcarle: Nevermind.  I think it makes no difference since it setCBOR takes an `any`

Issue #298, opened 2022-03-15
Sec. Indexes: Implement unique indexes


AndrewSisley:
Part of https://github.com/sourcenetwork/defradb/issues/201

Use composite keys

PR #1790, opened 2023-08-15
fix: Prevent multiple docs from being linked in one one

### LINKED ISSUE #1646: bug: Updating one-to-one relation link makes it act like a one-to-many
Consider one-to-one relation between `Book` and `Author`, Schema:

```gql

type Author {
	name: String
	published: Book @primary
}

type Book {
	name: String
	author: Author
}

```

Consider the following collections to have these documents respectively:

```
Author {
	{name: "author1"} // _docID = "a1"
	{name: "author2"} // _docID = "a2"
}

Book {
	{name: "book1", author: "a1"} // _docID = "b1"
}
```

Doing an `update_Author` query to update it's published book to be `book1` now like so:
`update_Author( id: "a2", data: {published_id: "b1"} )`

### Sets book of both authors as "b1".

Note: similar situation happens even if we do a `update_Book` instead:
 `update_Book(id: "b1", data: {author_id: "a2"})`
 
 
 Some actual tests that document this in our source code:
 - `TestMutationUpdateOneToOne_RelationIDToLinkFromPrimarySide`
 - `TestMutationUpdateOneToOne_RelationIDToLinkFromSecondarySide`
 - `TestBackupImport_WithMultipleNoKeyAndMultipleCollectionsAndMultipleUpdatedDocs_NoError`

AndrewSisley:
## Relevant issue(s)

Resolves #1646

## Description

Prevents multiple docs from being linked in one one.

It does this by scanning if from primary, or a point-lookup if from secondary.  By going via the `Fetcher` interface and `newFetcher` func this should respect Lens, and any indexes once the are implemented.

A handful of import/export tests got removed as they would no longer have anything to do with import/export once corrected (the error occurs before that action).


db/collection.go
```diff
@@ -1082,6 +1088,108 @@ func (c *collection) save(
...
+	if !ok {
+		return client.NewErrFieldNotExist(strings.TrimSuffix(fieldDescription.Name, request.RelatedObjectID))
+	}
+	if !objFieldDescription.RelationType.IsSet(client.Relation_Type_ONEONE) {
+		return nil
+	}
+
+	fetcher := c.newFetcher()
+
+	err := fetcher.Init(ctx, txn, &c.desc, []client.FieldDescription{fieldDescription}, nil, nil, false, false)
```
fredcarle, line 1116: question: Wouldn't it help to use a filter here instead of having to filter bellow? This way you would only need to check if it's self and if otherwise documents are returned, a relationship has already been set.

AndrewSisley: Yes, but a filter requires a mapping, which would need to be constructed, and then a manually constructed filter. Atm it really isn't worth that hassle I think as I think it would be uglier, and I'm doubtful it would have any meaningful impact on performance (atm) either.

Maybe in the future.

db/collection.go
```diff
@@ -1082,6 +1088,108 @@ func (c *collection) save(
...
+		props, err := doc.Properties(false)
+		if err != nil {
+			closeErr := fetcher.Close()
+			if closeErr != nil {
+				return errors.Wrap(err.Error(), closeErr)
+			}
+			return err
+		}
+
+		for field, fetchedValue := range props {
```
fredcarle, line 1169: question: Given that you requested only one field, there should only ever be one prop, correct?

AndrewSisley: I think at the moment all will be returned as Lens had to disable that.  Ignoring that issue, I'm not sure but _key and _deleted might be returned too but I'm not sure.

fredcarle: ok thanks. Are we tracking this disabled feature?

AndrewSisley: I don't know

fredcarle: Since you had to disable it with Lens, do you mind taking care of tracking it? Having it disabled has the side effect of using a lot more memory on queries. It's important that we don't forget it.

AndrewSisley: I was incorrect, this was changed.  Lens only disables this if there are migrations defined for the collection, and if there are there is little we can do here with the current public interfaces.  At a later date we could allow the users to control this, but there are probably more significant performance savings to be made before we get round to introducing such a thing.

fredcarle: Thanks for looking into it.

db/collection_update.go
```diff
@@ -364,13 +365,45 @@ func (c *collection) patchPrimaryDoc(
...
+	}
 
-	_, err = primaryCol.UpdateWithKey(
+	doc, err := primaryCol.Get(
 		ctx,
 		primaryDockey,
-		fmt.Sprintf(`{"%s": "%s"}`, primaryField.Name+request.RelatedObjectID, docKey),
+		false,
 	)
+	if err != nil && !errors.Is(err, ds.ErrNotFound) {
```
fredcarle, line 383: question: Shouldn't we return an error if we are trying to set a relation that doesn't exists?

AndrewSisley: We don't do this at all for anything at the moment and I would rather we added such behaviour deliberately and consistently instead of piecemeal as we happen upon such locations in the codebase

fredcarle: Alright. Just wanted to make sure that that is what we should be doing. Certainly out of scope here.

fredcarle (approved):
LGTM

db/errors.go
```diff
@@ -600,3 +602,12 @@ func NewErrDocCreate(inner error) error {
 func NewErrDocUpdate(inner error) error {
 	return errors.Wrap(errDocUpdate, inner)
 }
+
+func NewErrOneOneAlreadyLinked(documentId, targetId, relationName string) error {
```
shahzadlone, line 606: Nitpick: rename to `documentID`, `targetID`

shahzadlone (approved):

db/collection.go
jsimnz: cc: @AndrewSisley @fredcarle 

There is currently, and will continue to be *huge* savings both in time and space to use the Fetcher with the filter, and let the fetcher figure out how to effeciently give you the results.

In its current form, it will drastically reduce bytes copied, and in its future form it will drastically reduce the time to find all the docs matching the given filter.

At no point should any caller of the fetcher be doing its own filtering / field selection, unless it needs to merge/join with data from another collection/fetcher.

jsimnz:
This is a concerningly inefficient approach here. This should very much be refactored once #298 is completed, where each relation gets its own unique index. 

This would be analgeous to an SQL foreign key unique constraint. It will ofc be a space/time tradeoff, indexes using more space, but will turn this operation into an O(1) instead of O(n), at the cost of O(n) space.

This tradeoff can be exposed via the `@relation` directive at the time of schema creating by the dev/user if they wish to preserve space.

db/collection.go
fredcarle: I find that the fact that we have to do new,  Init, Start and FetchNext here is quite suboptimal. The field selection is done on Init so the other problem is the filtering that is happening within this function.  This will absolutely need to be refactored at some point in the near future but given that we don't have a friendly way to get with filter, I think this is acceptable for this PR.

jsimnz: > I find that the fact that we have to do new, Init, Start and FetchNext here is quite suboptimal.

The Fetcher is a fairly substantial system, and as such, it needs to fit a lot of different use cases, and functionality. Which is why things are seperated into `Init`, `Start`. The `FetchNext` is irrelvant, as that would be the case regardless of how the fetcher would be constructed. And the `newFetcher` stuff is just a shim so we can redirect which type of fetcher we're going to use. So really, it just comes down to `Init` and `Start`, which is seperated, so we can do proper multi-range fetches without rebuilding the fetcher.

> it really isn't worth that hassle

I find this somewhat dubious, as this is a wildly ineffecient approach for starters (doing a full table scan, and collecting every single field of every single document, just to determine we're not duplicating one-one links), so the hassle here would result in drastically reducing both the time and space complexity of this approach (only getting the field we care about, the relation id field, and only returning the docs that actually match the filter, and seeking past the rest, where `Seek` is magnitudes faster than a `Get`)

> but given that we don't have a friendly way to get with filter
This is a function of the current mapping system, and at the moment we have to live with it. If we're really promoting a much less effecient and out right wasteful approach, simply cus we don't want to touch the mapper, then we need to shift priorities to address that.

For reference, the Type Join code in the planner does exactly this, constructing a new filter, from the mapper, and passes that to the Fetcher, so it can let the fetcher do the hard work of effeciently seeking and filters docs.
```go
// line 347 of type_join.go
fkIndex := &mapper.PropertyIndex{
    Index: n.subType.DocumentMap().FirstIndexOfName(n.subTypeFieldName + "_id"),
}
filter := map[connor.FilterKey]any{
    fkIndex: doc.GetKey(),
}
```

Additionally, there is a func in the `mapper` package that makes this slightly less annoying, producing:
```go
rawFilter := map[string]any{
    relatedFieldName + "_id": map[string]any{"_eq": dockey}
}

filter := mapper.ToFilter(immutable.Some(request.Filter{filter}), mapping)
```

Both are a hassle, and we can put whatever syntactic surgar in between. But choosing to use the Fetcher in an inefficient way to avoid having to write an addition ~6 lines, and instead manually writing ~4 lines to do your own filter check and spend more time and space, seems... wrong.

--
As [mentioned](https://github.com/sourcenetwork/defradb/pull/1500#discussion_r1192614605) in the Fetcher refactor PR, the ugliness in constructing the filter is based on the necessity of the filter needing the doc mapper.

fredcarle: If you're happy with doing the whole fetcher manipulation here in the long run then it's really just adding the filter. The field selection is already passed to Init so not collecting every single field.

At the moment it has to be a whole table scan. We don't have indexes on the relational fields.

It would be nice to have a convenience function wrapping new, init and start but that would only really be useful if we're going to have multiple other locations that work directly with the fetcher like this.

fredcarle:
really@@hi

PR #1805, opened 2023-08-21
perf(i): Use a filter when checking output in one-one val

### LINKED ISSUE #1798: make use of filter in `validateOneToOneLinkDoesntAlreadyExist`
We currently manually filter outside of the fetcher. We should instead pass a filter to the fetcher to make use of seeking.

AndrewSisley:
## Relevant issue(s)

Resolves #1798

## Description

Uses a filter when checking output in one-one val instead of checking a fetcher result in the function directly.

Using the planner for this adds a small overhead, but avoids the need to expose the DocumentMapper and deal with the mess that that will entail.

Will fetch all values on a found document if one is found, but that is just zero-one doc and I don't think that is terribly important.


fredcarle:
> Will fetch all values on a found document if one is found, but that is just zero-one doc and I don't think that is terribly important.

We don't have a straight forward way from there to just select the key field?

AndrewSisley:
> > Will fetch all values on a found document if one is found, but that is just zero-one doc and I don't think that is terribly important.
> 
> We don't have a straight forward way from there to just select the key field?

Semi-straight forward, would need about 10 more lines of code.  I'd rather have the simpler code atm, the performance cost will b very low here.

PR #1806, opened 2023-08-21
fix: Error when saving a deleted document

### LINKED ISSUE #1787: Collection.Save does nothing if the doc was deleted
No error, no indication that nothing was done.  It just returns early.

AndrewSisley:
## Relevant issue(s)

Resolves #1787

## Description

Error when saving a deleted document, instead of doing nothing.

Previous behaviour documented in the commit history.


fredcarle (approved):
LGTM

PR #1807, opened 2023-08-22
feat: Auto-create relation id fields via PatchSchema

### LINKED ISSUE #1773: Auto create foreign id fields in PatchSchema
Part of #1002

Only if they are not explicitly defined by the user.

AndrewSisley:
## Relevant issue(s)

Resolves #1773

## Description

Auto-creates relation id fields via PatchSchema if not explicitly defined by the user.

### DIFF
--- a/db/collection.go
+++ b/db/collection.go
@@ -244,6 +244,20 @@ func (db *db) updateCollection(
 		return db.getCollectionByName(ctx, txn, desc.Name)
 	}
 
+	for _, field := range desc.Schema.Fields {
+		if field.RelationType.IsSet(client.Relation_Type_ONE) {
+			idFieldName := field.Name + "_id"
+			if _, ok := desc.Schema.GetField(idFieldName); !ok {
+				desc.Schema.Fields = append(desc.Schema.Fields, client.FieldDescription{
+					Name:         idFieldName,
+					Kind:         client.FieldKind_DocKey,
+					RelationType: client.Relation_Type_INTERNAL_ID,
+					RelationName: field.RelationName,
+				})
+			}
+		}
+	}
+
 	for i, field := range desc.Schema.Fields {
 		if field.ID == client.FieldID(0) {
 			// This is not wonderful and will probably break when we add the ability
@@ -443,24 +457,22 @@ func validateUpdateCollectionFields(
 			if proposedField.Kind == client.FieldKind_FOREIGN_OBJECT {
 				idFieldName := proposedField.Name + "_id"
 				idField, idFieldFound := proposedDesc.Schema.GetField(idFieldName)
-				if !idFieldFound {
-					return false, NewErrRelationalFieldMissingIDField(proposedField.Name, idFieldName)
-				}
+				if idFieldFound {
+					if idField.Kind != client.FieldKind_DocKey {
+						return false, NewErrRelationalFieldIDInvalidType(idField.Name, client.FieldKind_DocKey, idField.Kind)
+					}
 
-				if idField.Kind != client.FieldKind_DocKey {
-					return false, NewErrRelationalFieldIDInvalidType(idField.Name, client.FieldKind_DocKey, idField.Kind)
-				}
+					if idField.RelationType != client.Relation_Type_INTERNAL_ID {
+						return false, NewErrRelationalFieldInvalidRelationType(
+							idField.Name,
+							client.Relation_Type_INTERNAL_ID,
+							idField.RelationType,
+						)
+					}
 
-				if idField.RelationType != client.Relation_Type_INTERNAL_ID {
-					return false, NewErrRelationalFieldInvalidRelationType(
-						idField.Name,
-						client.Relation_Type_INTERNAL_ID,
-						idField.RelationType,
-					)
-				}
-
-				if idField.RelationName == "" {
-					return false, NewErrRelationalFieldMissingRelationName(idField.Name)
+					if idField.RelationName == "" {
+						return false, NewErrRelationalFieldMissingRelationName(idField.Name)
+					}
 				}
 			}
 
--- a/tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
+++ b/tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
@@ -115,32 +115,6 @@ func TestSchemaUpdatesAddFieldKindForeignObjectArray_MissingRelationName(t *test
 	testUtils.ExecuteTestCase(t, test)
 }
 
-func TestSchemaUpdatesAddFieldKindForeignObjectArray_MissingIDField(t *testing.T) {
-	test := testUtils.TestCase{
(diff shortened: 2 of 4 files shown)


tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
```diff
@@ -954,3 +928,79 @@ func TestSchemaUpdatesAddFieldKindForeignObjectArray_SecondaryObjectKindAndSchem
...
+				Request: `query {
+					Users {
+						name
+						foo {
+							name
+						}
+						foobar {
+							name
+						}
+					}
```
fredcarle, line 982: suggestion: To make it more explicit to the reader that `foobar_id` is set, I suggested returning it in the selection set.

tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
```diff
@@ -954,3 +928,79 @@ func TestSchemaUpdatesAddFieldKindForeignObjectArray_SecondaryObjectKindAndSchem
...
+
+	test := testUtils.TestCase{
+		Description: "Test schema update, add field with kind foreign object array (17), with auto id field generation",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: `
+					type Users {
+						name: String
+					}
+				`,
```
fredcarle, line 943: suggestion: Although it is interesting to show that this is possible, it would be nice to have a test also for two separate schemas. It would help to show that the `relation_id` field is only set on the primary.

tests/integration/schema/updates/add/field/kind/foreign_object_test.go
```diff
@@ -952,3 +894,159 @@ func TestSchemaUpdatesAddFieldKindForeignObject_ObjectKindAndSchemaMismatch(t *t
...
+
+	test := testUtils.TestCase{
+		Description: "Test schema update, add field with kind foreign object (16), with auto primary ID field creation",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: `
+					type Users {
+						name: String
+					}
+				`,
```
fredcarle, line 909: suggestion: Same as above. It would be nice to have a test for two separate schemas. In this case it would help show that the `relation_id` is set on both sides.

fredcarle (approved):
Approving as long as the comments get resolved :)

tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
AndrewSisley: Fair enough, will add.

- [ ] Add foobar_id to selection set

tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
AndrewSisley: > Although it is interesting to show that this is possible, it would be nice to have a test also for two separate schemas

I'm not sure I follow what makes this interesting, it is just a normal one-many relationship and adding a second schema to the test just complicates things.

> It would help to show that the relation_id field is only set on the primary.

I can't see how that would help, how do you see it?  I do think it is worth adding a test though to verify that the _id field does not exist on the many side.

- [ ] Assert _id field does not exist on many side.

shahzadlone (approved):
LGTM. 

Just 1 todo

tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
fredcarle: > I'm not sure I follow what makes this interesting, it is just a normal one-many relationship and adding a second schema to the test just complicates things.

It's interesting that we show that we can add self-referencing relationships. But that mean that everything is happening on the one collection. Having two separate collections would ensure that the operation isn't broken when it's not just one collection being manipulated.

db/collection.go
```diff
@@ -244,6 +244,20 @@ func (db *db) updateCollection(
 		return db.getCollectionByName(ctx, txn, desc.Name)
 	}
 
+	for _, field := range desc.Schema.Fields {
+		if field.RelationType.IsSet(client.Relation_Type_ONE) {
+			idFieldName := field.Name + "_id"
```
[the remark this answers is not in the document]
AndrewSisley: Will change, cheers.

- [ ] const _id

tests/integration/schema/updates/add/field/kind/foreign_object_array_test.go
AndrewSisley: I see what you mean, but that is not specifc to the changes in this PR (this feature actually is one of the least relevant, as it never cares which object it is hosted on). I'll add a ticket to close the test gap for other related behaviour though.

- [ ] Add test-ticket

fredcarle: In my mind it is very related as we need to ensure that the auto-creation is behaving properly for all objects involved.

fredcarle: I understand that the auto creation itself is specific to the one object but the patch operation is tested here through an integration test which involves 2 add operations. I feel like it's relevant to test that it works even with two distinct objects.

AndrewSisley: > integration test which involves 2 add operations

This is why it should be done in another ticket.  It should be tested with a simpler use case that isn't cluttered with irrelevancies.

AndrewSisley: It is irrelevant to the changes in this PR.

AndrewSisley: Isn't too different to trying to force a developer working on filters group by to test with a range of groupby order configurations, just because orderby isnt properly tested yet.  OrderBy should be tested, but in a simpler context where extras dont confuse the test.  It also means that if groupBy-filter is removed as a feature, groupBy-orderby is still tested.

And please remember, I am not objecting to the testing of PatchSchema relation definitions for multiple objects (even though it is technically not needed given the implementation).  I am say it should be done in another ticket, as it is not relevant here and will only complicate and dirty both work items.

AndrewSisley: https://github.com/sourcenetwork/defradb/issues/1809

fredcarle: I see it as related to the change in this PR. You're using an integration test to test the feature. The integration test doesn't care where the auto-create happens. It just cares that it happens. The most common way of having references is from one object to an other. Not from object to self. I would be quite appropriate to have a test that ensures that the feature works in the most common reference pattern where 2 objects are involved. It would also ensure that a change in the future doesn't result in only one side of the relationship having the relation id auto created.

If you really don't want to do it here, it should probably be done shortly so we don't forget about it.

PR #1810, opened 2023-08-23
feat: Allow field indexing by name in PatchSchema

### LINKED ISSUE #1169: Allow string/named substitution of field indexes
Part of #1002

I.e. instead of:
```
/Users/Schema/Fields/1
```
Users should be able to specify:
```
/Users/Schema/Fields/Age
```
Try and bear in mind whilst implementing that we'll likely want similar for collection name, kind and CrdtType.

Use of indexes should still be allowed else we'll break from the JSON patch spec - this should just be syntax sugar on top. Unless ofc we change the data model to map based, which could create a whole bunch of other issues (ordering, collisions, internal inefficiencies, etc) - I really do not want to go down that route, especially right now.

AndrewSisley:
## Relevant issue(s)

Resolves #1169

## Description

Allow field indexing by name in PatchSchema.


db/schema.go
```diff
@@ -171,23 +181,82 @@ func substituteSchemaPatch(
...
 	collectionsByName map[string]client.CollectionDescription,
 ) (jsonpatch.Patch, error) {
+	fieldIndexesByName := make(map[string]map[string]int, len(collectionsByName))
+	for colName, col := range collectionsByName {
+		colFieldIndexesByName := make(map[string]int, len(col.Schema.Fields))
+		fieldIndexesByName[colName] = colFieldIndexesByName
+		for i, field := range col.Schema.Fields {
+			colFieldIndexesByName[field.Name] = i
+		}
+	}
```
fredcarle, line 191: suggestion: This can be simplified to
```go
for colName, col := range collectionsByName {
    fieldIndexesByCollection[colName] = make(map[string]int, len(col.Schema.Fields))
    for i, field := range col.Schema.Fields {
        fieldIndexesByCollection[colName][field.Name] = i
    }
}
```

db/schema.go
```diff
@@ -171,23 +181,82 @@ func substituteSchemaPatch(
...
+						if colFieldIndexesByName, ok := fieldIndexesByName[desc.Name]; ok {
+							if i, ok := colFieldIndexesByName[fieldIndexer]; ok {
+								index = fmt.Sprint(i)
+							}
+						} else {
+							fieldIndexesByName[desc.Name] = map[string]int{
+								// The DocKey field is always at the zero index and we need to account for it
+								request.KeyFieldName: 0,
+							}
+						}
```
fredcarle, line 243: question: If the provided collection doesn't exist, shouldn't we return an error here?

db/schema.go
```diff
@@ -171,23 +181,82 @@ func substituteSchemaPatch(
...
+							}
+						} else {
+							field["Name"] = fieldIndexer
+						}
+						newPatchValue = immutable.Some[any](field)
+					}
+
+					desc := collectionsByName[splitPath[schemaNamePathIndex]]
+					var index string
+					if index == "" {
```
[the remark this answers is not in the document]
AndrewSisley: Cheers, I moved this block around a bit and missed this silliness.

- [ ] Remove extra check

db/schema.go
```diff
@@ -171,23 +181,82 @@ func substituteSchemaPatch(
 	patch jsonpatch.Patch,
 	collectionsByName map[string]client.CollectionDescription,
 ) (jsonpatch.Patch, error) {
+	fieldIndexesByName := make(map[string]map[string]int, len(collectionsByName))
```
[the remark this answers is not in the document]
AndrewSisley: I like that, will rename throughout - thanks Fred.

- [ ] Rename map

db/schema.go
AndrewSisley: It is one less line, but an extra call.  I prefer the current, there is actually less to read even though it takes up more eye-space.

db/schema.go
AndrewSisley: This is me falling into the trap of coding for the future and tolerating dead code to do so.

I was concerned about when PatchSchema gains the ability to add new collections, however this block sits within an `isFieldOrInner` check, and so when the time comes it would be a poor and surprising location to do this.

I will just delete the else block.

- [ ] remove dead code

AndrewSisley: Returning an error here would be dead code, there would be no way to reach it. If/when we allow new schema to be added via PatchSchema then we can add some logic somewhere to handle it, it would likely be located somewhere else.

fredcarle: Makes sense.

Issue #1816, opened 2023-08-24
collection.Save is not equivalent to request mutations


AndrewSisley:
We had a bug leaked recently (#1811) where an update via collection.Save was working, but update via a gql request was not.  Only the  collection.Save was tested.

We need to ensure that the behaviour is consistent.

AndrewSisley:
https://github.com/sourcenetwork/defradb/issues/1787 is also broken when given a gql request

PR #1817, opened 2023-08-24
fix: Only update updated fields via update requests

### LINKED ISSUE #1811: Document updates produce incorrect DAG state.
**Describe the problem**
DAG nodes from *update* mutations contain more than just the fields changed, they contain all the fields defined along with the field(s) changed.

**To Reproduce**
Create a document, update a document, run a commit request for the composite type. Will show a commit with all fields included in the update, when it 

**Expected behavior**
should only include the newest field(s) (delta)

**Additional context**
Potentially related to the LensFetcher, havent confirmed, works in 0.5.1, broken in 0.6.0

AndrewSisley:
## Relevant issue(s)

Resolves #1811

## Description

Only updates updated fields via update requests by returning a clean document from Decode.

## How has this been tested?

I have tested this via https://github.com/sourcenetwork/defradb/issues/1816 (https://github.com/sourcenetwork/defradb/compare/develop...AndrewSisley:1816-mutation-equivilency?expand=1) however this is a very simple and quite important fix deserving of its own commit/change-log entry in develop, and the test changes are complex and may bog down the merging of the fix.

### DIFF
--- a/db/fetcher/encoded_doc.go
+++ b/db/fetcher/encoded_doc.go
@@ -123,6 +123,11 @@ func Decode(encdoc EncodedDocument) (*client.Document, error) {
 
 	doc.SchemaVersionID = encdoc.SchemaVersionID()
 
+	// client.Document tracks which fields have been set ('dirtied'), here we
+	// are simply decoding a clean document and the dirty flag is an artifact
+	// of the current client.Document interface.
+	doc.Clean()
+
 	return doc, nil
 }
 


AndrewSisley:
> Can you please add an integration test that shows that this will result in the expected DAG state.

We have them, in the branch linked in the description.

fredcarle:
> We have them, in the branch linked in the PR description.

When I click on it I see no new tests. Nonetheless, we should test the changes that we make in the same PR as much as possible. Otherwise reviewers have to manually test the changed behaviour.

AndrewSisley:
> > We have them, in the branch linked in the PR description.
> 
> When I click on it I see no new tests. Nonetheless, we should test the changes that we make in the same PR as much as possible. Otherwise reviewers have to manually test the changed behaviour.

It doesnt add any new tests.  It allows all our existing tests to test this for us.

> Otherwise reviewers have to manually test the changed behaviour.

That depends on how much reviewers trust the author to make a one line change, tested in another branch.

fredcarle:
> That depends on how much reviewers trust the author to make a one line change, with the existing test gap provably closed in another branch.

The one line change doesn't obviously demonstrate the behaviour change. If I can't easily see what is happening in the code change, I have to have another way to confirm that the behaviour is as expected. In this case I had to manually run the queries.

AndrewSisley:
> > That depends on how much reviewers trust the author to make a one line change, with the existing test gap provably closed in another branch.
> 
> The one line change doesn't obviously demonstrate the behaviour change. If I can't easily see what is happening in the code change, I have to have another way to confirm that the behaviour is as expected. In this case I had to manually run the queries.

So read the description of the change in the PR header, or read the wip branch linked in it if you have too little faith in the author.

fredcarle (approved):
Next time please consider adding the tests with the PR.

jsimnz:
@AndrewSisley 

> We have them, in the branch linked in the PR description.

We actually don't *really* have them. I understand the benefit/goals of the linked branch running all mutations via both collection and GQL APIs, which I think is a nice and efficient way to ensure parity between the two APIs.

However, the linked branch doesn't add any tests that further gurantee the behavior of the original issue. It relies on the existing set of tests and runs those tests with the GQL mutation API. Unfortuently, the original tests only contain a ***single*** small test to check the issue, and its not even necessarily the core focus of the test (`TestQueryCommitsWithDockeyAndUpdateAndLinks`).

Based on how you've split the work, im very much OK to have the tests in the followup PR, but the tests in that branch need to be expanded to further cement coverage of the issue.

cc: @fredcarle

AndrewSisley:
> However, the linked branch doesn't add any tests that further gurantee the behavior of the original issue. It relies on the existing set of tests and runs those tests with the GQL mutation API. Unfortuently, the original tests only contain a single small test to check the issue, and its not even necessarily the core focus of the test (TestQueryCommitsWithDockeyAndUpdateAndLinks).

I believe you are mistaken here.  5 of our existing tests failed because of this bug, it is not just visible via links, but the field level commits.

PR #1850, opened 2023-09-07
test(i): Skip unsupported mutation types at test level

### LINKED ISSUE #1849: Fix change detector bug due to not skipping some mutation tests
The change detector action on develop seems to be failing, this is because some mutation tests should be skipped for the change detector but aren't because the skip is handled within the mutation action handlers (only ran on the setup side).

Failed Run:
https://github.com/sourcenetwork/defradb/actions/runs/6103214759/job/16563218995

AndrewSisley:
## Relevant issue(s)

Resolves #1849

## Description

Skips unsupported mutation types at test level.

Earlier version caused the change detector to fail, as the skip was only executed in the setup stage, not the main/assert stage resulting in test failures. I also think this location makes much more sense, as the skip is a test-level thing, not something that acts at the action-level.

Sorry about the bother caused by this, this should have been spotted by me when I wrote it.

### DIFF
--- a/tests/integration/collection/update/simple/with_deleted_test.go
+++ b/tests/integration/collection/update/simple/with_deleted_test.go
@@ -21,6 +21,10 @@ import (
 func TestUpdateSave_DeletedDoc_DoesNothing(t *testing.T) {
 	test := testUtils.TestCase{
 		Description: "Save existing, deleted document",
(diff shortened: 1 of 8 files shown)


shahzadlone (approved):
LGTM. Thanks

Issue #1852, opened 2023-09-07
Inconsistent errors returned from collection.save


AndrewSisley:
The collection api varients of `TestMutationUpdateOneToOne_RelationIDToLinkFromSecondarySideWithWrongField_Error` are flaky, as sometimes they return the expected error, and sometimes they return a different error.

Issue #1703, opened 2023-07-24
Field alias does not exist in some cases for update of one-one


AndrewSisley:
Documented by tests:
- TestMutationUpdateOneToMany_InvalidAliasRelationNameToLinkFromManySide_GQL
- TestMutationUpdateOneToMany_InvalidAliasRelationNameToLinkFromManySide_Collection
- TestMutationUpdateOneToMany_AliasRelationNameToLinkFromManySideWithWrongField_Error
- TestMutationUpdateOneToMany_AliasRelationNameToLinkFromManySide
- TestMutationUpdateOneToOne_AliasRelationNameToLinkFromPrimarySide
- TestMutationUpdateOneToOne_AliasRelationNameToLinkFromSecondarySide
- TestMutationUpdateOneToOne_AliasWithInvalidLengthRelationIDToLink_Error
- TestMutationUpdateOneToOne_InvalidAliasRelationNameToLinkFromSecondarySide_Error

PR #1853, opened 2023-09-07
test(i): Update mutation tests to make use of mutation system

### LINKED ISSUE #1851: Make existing mutation tests use mutation test system
Tweak them to make use of new mutation test system where possible.

AndrewSisley:
## Relevant issue(s)

Resolves #1851

## Description

Updates the mutation tests to make use of mutation system.

A handful of collection based tests still exist, as they test stuff that does not yet have support in the new mutation test system (filter, UpdateWithKey(s)).  These should be moved later when such support is added.

https://github.com/sourcenetwork/defradb/issues/1703 Was linked up to tests in this PR, and https://github.com/sourcenetwork/defradb/issues/1852 was discovered.  Fixing them is out of scope here.

Recommend reviewing commit by commit, nothing to fancy is going on, but it might make stuff more obvious.


tests/integration/mutation/update/field_kinds/one_to_many/with_alias_test.go
```diff
@@ -14,66 +14,108 @@ import (
...
+				),
 			},
-			testUtils.Request{ // NOTE: There is no `published_id` and so `published` alias is invalid use on book.
-				Request: fmt.Sprintf(
-					`mutation {
- 						update_Author(id: "%s", data: "{\"published\": \"%s\"}") {
- 							name
- 						}
- 					}`,
-					author2Key,
```
shahzadlone, line 76: question: why was `author2Key` deleted?

tests/integration/mutation/update/field_kinds/one_to_many/with_alias_test.go
```diff
@@ -14,66 +14,108 @@ import (
...
+			testUtils.UpdateDoc{
+				CollectionID: 1,
+				DocID:        1,
```
shahzadlone, line 113: question: Should DocID be the `author2Key` here?

AndrewSisley: no, it is and should be the (test) index of the document.

shahzadlone (approved):
The PR looks fine for what it does, however I have a non-blocking question, doesn't the use of mutation system over the actual gql queries like we were doing it before now introduce a gap between integration testing a client's actual end-to-end use case?

PR #1839, opened 2023-08-31
refactor: CLI client interface

nasdf:
## Relevant issue(s)

Closes #1472
Closes #1507
Closes #1860 

## Description

This is a follow up to #1776

This PR adds a CLI implementation that implements the client.DB interface and runs through the existing integration test suite.

- [x] Merge existing server config code
- [x] Refactor CLI to use new HTTP client
- [x] Remove `net/api` package
- [x] Remove `api/http` package
- [x] Lens tests are timing out in CI: fixed #1862
- [x] Code coverage is incorrectly reporting: fixed #1861
- [x] Flaky test causing failures: fixed #1912

Renamed Commands:
- `peerid` to `client peer info`
- `client p2pcollection` to `client p2p collection`
- `client replicator` to `client p2p replicator`
- `client schema list` to `client collection describe`

Removed Commands:
- `block get`
- `ping`
- `rpc`

Added Commands:
- `client collection create`
- `client collection delete`
- `client collection get`
- `client collection keys`
- `client collection update`
- `client tx create`
- `client tx discard`
- `client tx commit`
- `client schema migration up`
- `client schema migration down`
- `client schema migration reload`

**Notes for reviewers**:
- `.github` changes are merged from #1871
- `Makefile` most of these changes are also from #1871
- `docs/cli` ignore these changes, it will be updated next release
- sorry for all of the merge commits, I am working on learning rebase flow

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

`make test`

Specify the platform(s) on which this was tested:
- MacOS

PR #1862, opened 2023-09-08
ci: Test matrix

nasdf:
## Relevant issue(s)

N/A

## Description

This PR has been merged into #1839 due to flaky CLI tests causing failures.

This merges all test variants into a single test matrix and includes a few other updates:

- [x] Merge `code-test-coverage` into `run-tests` workflow
- [x] Batch upload code coverage from all jobs after a successful run
- [x] Replace go-acc with go 1.20 aggregate test coverage 
- [x] Run change detector as a matrix job #1871

## Tasks

## How has this been tested?

CI

- MacOS


.github/workflows/run-tests.yml
```diff
@@ -20,10 +20,24 @@ on:
 
 jobs:
   run-tests:
-    name: Run tests job
+    name: Run tests job matrix
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
```
AndrewSisley, line 28: praise: :heart_eyes: :heart_eyes: :heart_eyes:  I did not know we could do this.  And I really want it.  Thank you very much for opening this draft.

.github/workflows/run-tests.yml
```diff
@@ -20,10 +20,24 @@ on:
...
+        client-type: [go, http]
+        database-type: [badger-file, badger-memory]
+        mutation-type: [gql, collection-named, collection-save]
+        tests: [names, lens, cli]
```
AndrewSisley, line 32: suggestion: I would remove this line, and instead all items can just test `./...`.

.github/workflows/run-tests.yml
```diff
@@ -20,10 +20,24 @@ on:
...
+
```
AndrewSisley, line 33: thought: Once we remove the `-p 1` limitation, we can include the change detector in this file using an `include`, I think the below *might* work:
```
 include:
      - client-type: go
       database-type: badger-memory
        mutation-type: collection-save
        detect-changes: true
 ```

Comment (AndrewSisley) [inline .github/workflows/run-tests.yml:33]: thought: I think we should also include the codecov stuff in this file, and delete the codecov workflow. I don't want to mark this as a todo though, as I'm still playing around with it and it can be done after easily enough.

File: .github/workflows/run-tests.yml [hunk 4]
```diff
@@ -20,10 +20,24 @@ on:
 
 jobs:
   run-tests:
-    name: Run tests job
+    name: Run tests job matrix
 
     runs-on: ubuntu-latest
```
AndrewSisley, line 25: thought: We could also include this in the matrix at somepoint if we want (I would like this).

.github/workflows/run-tests.yml
AndrewSisley, line 28: thought: This will increase the visibility of any test flakiness.  This could be extremely annoying, but will force us to sort out flakiness.

.github/workflows/run-tests.yml
```diff
@@ -48,10 +47,50 @@ jobs:
...
+      - name: Install Go dependencies
         run: |
-          make deps:modules
-          make deps:test
-      
+          go mod download
+          go install gotest.tools/gotestsum@latest
+          go install github.com/ory/go-acc@latest
+
+      - name: Build Lens dependencies
```
AndrewSisley, line 56: praise: This is really nice that we only have to do this once :)

AndrewSisley, line 56: suggestion: However, I think it should probably be done via a `make` command, instead of defined directly in the github action file

.github/workflows/run-tests.yml
```diff
@@ -48,10 +47,50 @@ jobs:
...
+          cargo build --target wasm32-unknown-unknown --manifest-path "./tests/lenses/rust_wasm32_set_default/Cargo.toml"
+	        cargo build --target wasm32-unknown-unknown --manifest-path "./tests/lenses/rust_wasm32_remove/Cargo.toml"
+	        cargo build --target wasm32-unknown-unknown --manifest-path "./tests/lenses/rust_wasm32_copy/Cargo.toml"
+
       - name: Run the tests
-        run: make test:${{ matrix.tests }}
+        run: gotestsum ./... --format testname -- -race -shuffle=on -timeout 300s
+
+      - name: Generate test coverage
+        run: go-acc ./... --output=coverage.txt --covermode=atomic -- -failfast
```
AndrewSisley, line 67: todo/question: This looks like it is running the tests twice, once without code coverage with our test params (race/shuffle/etc), and once with code coverage but without the params?  Suggest just running the tests once

nasdf: I converted coverage to the new go 1.20 aggregate mode and it now only runs tests once.

.github/workflows/run-tests.yml
nasdf: Fixed

.github/workflows/run-tests.yml
nasdf: Added both the change detector and non change detector to the include so both will now run.

Issue #1869, opened 2023-09-11
Panic using compound relational filter in top level aggregate 


AndrewSisley:
Reading the code, this may affect any relational filter in any aggregate.  There also looks like there may be multiple separate code paths to achieve this flavour of bug.  Top level, normal aggregate, normal aggregate where relation is also requested (this likely works, but is untested and may do other strange stuff). Be very careful with testing.

Originally discovered bug documented with `TestQueryOneToOneWithCountWithCompoundOrFilterThatIncludesRelation`

PR #1870, opened 2023-09-11
test(i): Test top level agg. wt compound relational filter

### LINKED ISSUE #1868: Add tests for aggregates with compound relational filters

AndrewSisley:
## Relevant issue(s)

Resolves #1868

## Description

Adds a test for aggregates with a compound relational filter.

Documents https://github.com/sourcenetwork/defradb/issues/1869

I believe this predates the release, and may be a fiddly fix with lots more testing required, so I'm documenting it now and moving on - maybe it will be fixed before the release goes out, but will see.

### DIFF
--- a/tests/integration/utils2.go
+++ b/tests/integration/utils2.go
@@ -196,17 +196,24 @@ func init() {
 	}
 }
 
-// AssertPanicAndSkipChangeDetection asserts that the code of function actually panics,
+// AssertPanic asserts that the code inside the specified PanicTestFunc panics.
 //
-//	also ensures the change detection is skipped so no false fails happen.
+// This function is not supported by either the change detector, or the http-client.
+// Calling this within either of them will result in the test being skipped.
 //
-//	Usage: AssertPanicAndSkipChangeDetection(t, func() { executeTestCase(t, test) })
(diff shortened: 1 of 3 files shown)


AndrewSisley:
`AssertPanicAndSkipChangeDetection` does not work with the http client, the test will not pass until sorted.  Looking at how to handle that now.

tests/integration/utils2.go
```diff
@@ -196,17 +196,24 @@ func init() {
 	}
 }
 
-// AssertPanicAndSkipChangeDetection asserts that the code of function actually panics,
+// AssertPanic asserts that the code inside the specified PanicTestFunc panics.
```
shahzadlone, line 199: nitpick(non-blocking): Rename suggestion
```suggestion
// AssertPanicOrSkip asserts that the code inside the specified PanicTestFunc panics.
```

shahzadlone (approved):
LGTM. Just a non-blocking name suggestion.

tests/integration/utils2.go
AndrewSisley: :) I do disagree here, although I think my views over time have shifted, and I might have been the one to originally ask for the longer name.  Will leave as `AssertPanic` unless anyone else requests the change before I rebase and merge.

PR #1888, opened 2023-09-15
feat: Allow setting of default schema version

### LINKED ISSUE #1884: Allow users to set default schema version
Currently we default to the head schema version, this is limiting and may often be unwanted at times - forcing stuff to update immediately upon definition.

Allow users to set the default explicitly.  Consider adding a param to `PatchSchema` to allow users to set the default to the new one so that they dont have to make a second call.

E.g.:
```
PatchSchema(context.Context, string, bool)
SetDefaultSchemaVersion(context.Context, string, string)
```

AndrewSisley:
## Relevant issue(s)

Resolves #1884

## Description

Allows setting of default schema version, allowing rapid switching between (application) api/database versions.  It also allows them to define and apply schema updates eagerly, and then make the switch at a later date.

- [ ] Wait for the 0.7 release to be cut before merging


client/db.go
```diff
@@ -109,7 +110,16 @@ type Store interface {
...
+	PatchSchema(context.Context, string, bool) error
+
+	// SetDefaultSchemaVersion sets the default schema version to the ID provided.  It will be applied to all
+	// collections using the schema.
+	//
+	// This will affect all operations interacting with the schema where a schema version is not explicitly
+	// provided.  This includes GQL queries and Collection operations.
+	//
+	// It will return an error if the provided schema version ID does not exist.
+	SetDefaultSchemaVersion(context.Context, string) error
```
nasdf, line 122: This was confusing for me at first. What do you think about moving this to the Collection interface instead? I think it  would be more clear exactly which Collection you are setting a default for.

client/db.go
```diff
@@ -109,7 +110,16 @@ type Store interface {
 	//
 	// Field [FieldKind] values may be provided in either their raw integer form, or as string as per
 	// [FieldKindStringToEnumMapping].
-	PatchSchema(context.Context, string) error
+	PatchSchema(context.Context, string, bool) error
```
nasdf, line 113: What are your thoughts on inverting the boolean usage here? 

I think it'd be nice when using the http or cli interfaces that the lack of the boolean would indicate the original functionality of setting the default. 

Requiring an extra `--set-default-schema` cli flag or `SetDefaultSchema` http property would break backwards compatibility.

nasdf (commented):
I left some feedback about the public API portions.

client/db.go
AndrewSisley: I want switching schema versions to be an explicit choice - it is a major, and likely breaking, change to the user's database and I don't want them to do it accidentally via parameter omission.

tests/integration/utils2.go
```diff
@@ -418,6 +418,9 @@ func executeTestCase(
 		case SchemaPatch:
 			patchSchema(s, action)
 
+		case SetDefaultSchemaVersion:
```
nasdf, line 421: I think we'll want this action included in the change detector setup steps.

client/db.go
nasdf: The simplicity is nice but I can see many users getting confused. 

This is the use case I see being used most often when you don't know the schema version id:

```go
cols := db.GetAllCollections()

for _, col := range cols {
  if isMatchingCollection(col) {
    col.SetDefaultSchemaVersion()
  }
}

```

AndrewSisley: I do disagree :) I see the most common use case to within a migration script:

```
defradb patchSchema jsonStuff
...
defradb setDefaultSchemaVersion hjndsmakdgamkmhga
```
if it was on the collection object it would be a bit stranger no, would look something like the below no?
```
defradb collection hjndsmakdgamkmhga setDefaultSchemaVersion
```

tests/integration/utils2.go
AndrewSisley: I don't think we do.  Normal tests will just set the PatchSchema change to default using the bool, tests that use this action are perhaps going to benefit more from having this in the main (post setup) stage, as they are more likely to be testing that SetDefaultSchemaVersion works.

client/db.go
nasdf: It's definitely more verbose but it's also much easier to parse what that command is doing. 

No need to block the PR. I think we could add the shortcut function in the future if users have trouble with it.

nasdf (approved):
Looks good! From the discussion there's a minor API addition that we may revisit if needed.

PR #1915, opened 2023-09-27
fix: Infinite loop when updating one-one relation

### LINKED ISSUE #1914: Infinite loop when updating one-one self-ref. relation field
Depending on field names/order, updating a one-one self-referencing relationship from the secondary field can result in an infinite loop due to `CollectionDescription.GetRelation`

AndrewSisley:
## Relevant issue(s)

Resolves #1914 

## Description

Fixes an a bug where an infinite loop may be created when updating a self-referencing one-one relation from the secondary side.

The first commit documents the bug, the second commit fixes it.


shahzadlone (approved):
Cheers for sorting

PR #1920, opened 2023-10-02
fix(i): Remove collection name from schema ID generation

### LINKED ISSUE #1083: Make schemaID only dependant on the schema description.
At the moment, the schemaID is dependant on the collection name. If we eventually want to allow for the collection name to be different from the schema name but still reference the same schemaID, we need to decouple the two.

```go
schemaIDBuf, err := json.Marshal(col.desc.Schema)
```

AndrewSisley:
## Relevant issue(s)

Resolves #1083

## Description

Removes collection name from schema ID generation.

It is a schema ID, not a collection ID, including the collection name makes no sense and is likely a historical artefact.

### DIFF
--- a/api/http/handlerfuncs_test.go
+++ b/api/http/handlerfuncs_test.go
@@ -652,8 +652,8 @@ type group {
 			"collections": []any{
 				map[string]any{
 					"name":       "group",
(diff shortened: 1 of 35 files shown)


fredcarle (approved):
LGTM. Thanks for removing the collection name in the generation!

PR #1923, opened 2023-10-02
fix(i): Remove unused logging context

nasdf:
## Relevant issue(s)

N/A

## Description

This PR removes the unused context from all logging calls.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/cli/init.go
+++ b/cli/init.go
@@ -53,7 +53,6 @@ Passed flags will be persisted in the stored configuration.`,
 						}
 					} else {
 						log.FeedbackError(
-							cmd.Context(),
 							fmt.Sprintf(
 								"Configuration file already exists at %v. Consider using --reinitialize",
 								cfg.ConfigFilePath(),
--- a/cli/root.go
+++ b/cli/root.go
@@ -11,8 +11,6 @@
 package cli
 
 import (
-	"context"
-
 	"github.com/spf13/cobra"
 
 	"github.com/sourcenetwork/defradb/config"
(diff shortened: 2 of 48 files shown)


AndrewSisley (changes requested):
:grin: Controversial - this was deliberately included in the func sigs, so that the logger can include stuff from the context as/when we choose to do so.

I think this should be talked about in discord, to give it increased visibility.

jsimnz:
Agree with @AndrewSisley here, this is rather intentional design choice for the purposes of flexability/future upgradeability. I do actually think there is supposed to be a feature that uses this, but might not be implemented yet.

Basically, we should be able to create a general group of Logger KV Tags, and embed that into the context, so that any future call to the logger, that uses the same context will automatically pickup those tags. 

A usecase for this is things like PeerID, where at the top of the DB initialization, we can embed the PeerID into the context, and gurantee that all logger calls contain it.

```go
ctx := context.Background()
ctx = logger.ContextWithKV(ctx, logger.NewKV("peerID", peer.ID())

// ...

db.ExecRequest(ctx, query) // all logger calls inside of this call stack will print (peerID=QzABC...)
```

--

One such example/usecase for the context

nasdf:
> That only works on a log-line-by-line, if the context is always available certain key values can be appended to all log calls without polluting the core code, for example a user ID could be added to every log line.

I think the unused context is polluting the core code.

islamaliev (approved):
Looks neat with your change, but address please other comments first

PR #1924, opened 2023-10-03
refactor: P2P client interface

### LINKED ISSUE #1883: P2P database functions not calling peer logic
When calling the `client.P2P` interface functions it is expected that the peer connection logic will be called. This is currently not the case for `SetReplicator` and possibly others. 

The integration tests are currently not testing for this case as they use the `net.Node` and `net.Peer` structs instead.

Discussed a fix with @fredcarle 
- Remove `client.P2P` interface from `client.DB`
- Only `net.Node` implements `client.P2P`

nasdf:
## Relevant issue(s)

Resolves #1883 

## Description

Blocked by #1927 

This PR moves the `client.P2P` implementation from `client.DB` to `net.Node`. This fixes the problems mentioned in the issue above and should increase test coverage of the HTTP and CLI clients.

## Tasks

## How has this been tested?

`make test`

- MacOS


config/config.go
```diff
@@ -350,79 +349,38 @@ func (apicfg *APIConfig) AddressToURL() string {
 
 // NetConfig configures aspects of network and peer-to-peer.
 type NetConfig struct {
```
AndrewSisley, line 351: praise: Nice :grin: Glad this is simpler now

cli/p2p_replicator_delete.go
```diff
@@ -11,27 +11,36 @@
...
 
 	"github.com/sourcenetwork/defradb/client"
 )
 
 func MakeP2PReplicatorDeleteCommand() *cobra.Command {
+	var collections []string
 	var cmd = &cobra.Command{
-		Use:   "delete <peer>",
+		Use:   "delete [-c, --collection] <peer>",
 		Short: "Delete a replicator. It will stop synchronizing",
```
AndrewSisley, line 26: question: Should this (and `Long`), not now read `Delete replicators...`? It looks like it can now delete multiple.

AndrewSisley:
>  remove GRPC API

I really really think that such a change deserves its own PR.

nasdf:
> > remove GRPC API
> 
> I really really think that such a change deserves its own PR.

It's not used in the current develop branch. I had deprecated it when refactoring the CLI and forgot to remove it.

I'm happy to pull those changes into a new PR if you'd prefer.

client/p2p.go
```diff
@@ -12,9 +12,17 @@ package client
 
 import (
 	"context"
+
+	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+// P2P is a peer connected database implementation.
 type P2P interface {
+	DB
```
AndrewSisley, line 21: todo: It took me a while to figure it out, but you have removed functionality here.  `client.P2P` used to respect the transactionality of `client.Store`, but now it only works with implicit transactions.

I'm not sure we wish to lose this, and this change should be made very visible and open to discussion with the team.

AndrewSisley (changes requested):
I have a fairly major todo RE transactionality, I'll resume review once that conversation is resolved.

AndrewSisley:
> I'm happy to pull those changes into a new PR if you'd prefer.

I would prefer it, it is a big change, and should be in the commit history.  It also clutters up this PR a fair amount making review harder.

client/p2p.go
nasdf: I discussed this with @fredcarle and forgot to make a note. We came to the conclusion that peer operations only make sense when done implicitly. It would be pretty difficult to have the transaction also rollback the peer connections that happen during these operations.

nasdf:
> > I'm happy to pull those changes into a new PR if you'd prefer.
> 
> I would prefer it, it is a big change, and should be in the commit history. It also clutters up this PR a fair amount making review harder.

I'm working on this now and will update with a link when ready.

cli/p2p_replicator_delete.go
nasdf: If you are referring to deleting multiple collections replicated to a single peer you are correct.

client/p2p.go
fredcarle: We are not losing functionality. We are removing the need for the `Store` interface to support P2P related actions. The functionality still exists but is handled all within the `net` package. However, whatever implements the `P2P` interface needs access to a db instance and thus support the `DB` interface. Action related to the `P2P` interface can still use transactions.

AndrewSisley: > We came to the conclusion that peer operations only make sense when done implicitly.

I do disagree with this, with the obvious example being during an update to a new application version - there may be several of these operations (plus other stuff, like schema updates etc) and if any one of those operations fails, they should all fail so that the update is not partially applied.

I will not push back against the both of you on this, but I do think explicit transactions here are useful.

README.md
```diff
@@ -244,57 +244,73 @@ When starting a node for the first time, a key pair is generated and stored in i
 
 Each node has a unique `PeerID` generated from its public key. This ID allows other nodes to connect to it.
 
+To view your node's peer info:
```
AndrewSisley, line 247: praise: Thank you for adding this!

README.md
```diff
@@ -334,14 +350,20 @@ defradb client schema add --url localhost:9182 '
...
 ```shell
-defradb client rpc replicator set -c "Article" /ip4/0.0.0.0/tcp/9172/p2p/<peerID_of_nodeB>
+defradb client p2p info
 ```
 
-As we add or update documents in the "Article" collection on *nodeA*, they will be actively pushed to *nodeB*. Note that changes to *nodeB* will still be passively published back to *nodeA*, via pubsub.
+Set *nodeA* to actively replicate the Article collection to *nodeB*:
+
+```shell
+defradb client p2p replicator set -c Article <peer_info_of_nodeB>
```
AndrewSisley, line 362: suggestion: I had to check the help text of `defradb client p2p replicator set` to be sure that `<peer_info_of_nodeB>` meant the full json object - it might be worth renaming this to reflect that.

db/db.go
```diff
@@ -266,17 +262,17 @@ func (db *db) PrintDump(ctx context.Context) error {
 
 // Close is called when we are shutting down the database.
 // This is the place for any last minute cleanup or releasing of resources (i.e.: Badger instance).
-func (db *db) Close(ctx context.Context) {
-	log.Info(ctx, "Closing DefraDB process...")
+func (db *db) Close() {
+	log.Info(context.Background(), "Closing DefraDB process...")
```
AndrewSisley, line 266: question: Was this what prompted the attempted removal of the `ctx` from the log interface?

thought: This does *feel* incorrect, as context has been defined outside of this function, and whilst at the moment as far as I can see they are all untouched `context.Background()`s, the context is not owned by this function, and could be modified by those owning layers, at which point this function would become not just conceptually incorrect, but tangibly so too.  It might be worth chatting about this - whilst a minor/unimportant part of the codebase, this kind of feels like a drop in standards and as such I was temped to mark this as a todo.

tests/clients/cli/wrapper.go
```diff
@@ -27,54 +27,68 @@ import (
...
 
+func (w *Wrapper) PeerInfo() peer.AddrInfo {
+	args := []string{"client", "p2p", "info"}
+
+	data, err := w.cmd.execute(context.Background(), args)
+	if err != nil {
+		return peer.AddrInfo{}
+	}
+	var info peer.AddrInfo
+	if err := json.Unmarshal(data, &info); err != nil {
```
AndrewSisley, line 63: question: Here, and in the if above, we are ignoring the error. Other functions in this wrapper seem to return their errors.  How confident are you on the safety of this?  Could it not lead to the tests missing panics and errors that because of this only appear in the logs?

tests/integration/setup.go
```diff
@@ -0,0 +1,256 @@
+// Copyright 2023 Democratized Data Foundation
```
AndrewSisley, line 1: praise: I like that this is being broken out of the main (and large) utils2.go file

suggestion: I think it might be worth splitting it up again, I do dislike 'setup' files in general, as they often reduce the cohesion of the codebase, moving stuff away from where it is used.  However here we only really have two things, and they are independent.  We could split this file up into `clients.go` and `db.go` (or similar names), this would keep the database stuff together, and the client stuff together, both out of the main utils.go, and avoid us creating a general purpose, lets shove everything in it, setup file that risks growing horribly in the future if/when we add more stuff that requires setup.

There is also an argument to be had that 'setup' here is misleading (particularly considering the large init func at the top of the file), as it doesn't actually setup up the database or clients, it hosts the functions that the utils2.go calls to set them up.

AndrewSisley (commented):
Looks good, but I still have some small questions.

db/db.go
nasdf: Yeah this was the exact area where I ran into the log context. I plan on cleaning this up in the future when switching to the new `log/slog` go package.

tests/clients/cli/wrapper.go
nasdf: I could change the API signature to return an error here. My thought process was that an empty `peer.AddrInfo` struct represents an error and any function using that empty address should return an error.

db/db.go
AndrewSisley: Okay, I'm happy enough if it is left as-is in the short term then if others are :)

tests/integration/setup.go
nasdf: I like that idea. I split them into `db.go` and `client.go` and put the init back into `utils2.go`

tests/clients/cli/wrapper.go
AndrewSisley: It feels a bit wrong to change the production interface for the sake of a test wrapper (even if just to add an error return param).  Maybe we could just panic here for now (instead of returning empty)?

tests/integration/setup.go
AndrewSisley: I think you forgot to add+push the new files :) 

> and put the init back into utils2.go

Why did you chose to return the init?  I did picture that being split up too.

nasdf: Okay I see what you mean. I split the `init` logic but left the mutation variables in utils since they are not db or client specific.

README.md
nasdf: I updated it to `nodeB_peer_info_json` to hopefully make it more obvious.

AndrewSisley (approved):
LGTM, thanks Keenan :)

PR #1925, opened 2023-10-04
feat: Make queries utilise secondary indexes

### LINKED ISSUE #1555: Sec. Indexes: Make queries use indexes for fetching data

islamaliev:
## Relevant issue(s)

Resolves #1555

## Description

With this change the the secondary indexes are utilised during querying data.

A dedicated `Indexer` fetcher is implemented to perform fetching of values of indexed fields.

Now there is a separate `filter` package that houses a lot of methods for working with filters.

A new metric `indexesFetched` is introduced into `@explain` to provide information about how many indexes has been fetched.

It also includes an update to the testing framework to allow adding custom asserters.
The new ExplainResultsAsserter is used with this new feature.

## Benchmark

I ran some query benchmarks with regular fetching and indexed fetching on 10k docs and here is the summary:

**Time per Operation (ns/op)**
Indexed: Took approximately 2,196,481 nanoseconds (~2.2 ms) per operation
Original: Took approximately 74,037,628 nanoseconds (~74 ms) per operation
Boost Factor: Indexed is approximately 33.7 times faster than the original.

**Memory Allocation (B/op)**
Indexed: Allocated approximately 1,053,403 bytes (~1 MB) per operation
Original: Allocated approximately 45,218,327 bytes (~45 MB) per operation
Boost Factor: Indexed uses approximately 42.9 times less memory than the original.

**Number of Allocations (allocs/op)**
Indexed: Made 13,331 memory allocations per operation
Original: Made 494,494 memory allocations per operation
Boost Factor: Indexed has approximately 37.1 times fewer memory allocations than the original.

**Summary**
The indexed approach is 33.7 times faster, uses 42.9 times less memory, and has 37.1 times fewer memory allocations compared to the original approach.

## Tasks

## How has this been tested?

Integration tests

- MacOS


errors/defraError.go
```diff
@@ -58,13 +58,12 @@ func (e *defraError) Error() string {
...
 
 func (e *defraError) Is(other error) bool {
-	switch otherTyped := other.(type) {
-	case *defraError:
-		return e.message == otherTyped.message
-	default:
-		otherString := other.Error()
-		return e.message == otherString || e.Error() == otherString || errors.Is(e.inner, other)
+	var otherDefraError *defraError
+	if errors.As(other, &otherDefraError) {
```
islamaliev, line 62: I wasn't planning on changing this, but lint complained:
`type switch on error will fail on wrapped errors. Use errors.As to check for specific errors`

AndrewSisley (changes requested):
todo: The purpose of indexes is performance gains on read, at the cost of performance on write (and storage space).  I really think performance tests should be included in this PR.

I would like to see performance tests for non-indexed vs indexed queries, on both read and write.  Without these we are not really testing that this code change is doing anything useful, and could be seen as an explain-feature or a simple refactor.

praise: What I have seen so far of the production code looks good, but I have left a few comments dotted about.  Will finish my review a bit later once some of them have been addressed.

tests/integration/index/query_with_index_only_filter_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+package index
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestQueryWithIndex_WithNonIndexedFields_ShouldFetchAllOfThem(t *testing.T) {
```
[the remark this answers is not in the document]
islamaliev: I guess your concern mainly about `sendRequestAndExplain` function. Unfortunately go doesn't support named parameters, so we can write like this:
```go
sendRequestAndExplain(requestBody: `...`, expectedResult: []map[string]..., asserter: testUtil.New...)
```
I know this way it's not quite explicit when reading the first time, but this way we can avoid writing like this:
```go
...
testUtils.Request{
	Request: `
		query {
			User(filter: {name: {_eq: "Islam"}}) {
			        name
			        age
      			}
		}`,
	Results: []map[string]any{{
	        "name": "Islam",
	        "age":  uint64(32),
        }},,
},
testUtils.Request{
	Request: `
		query @explain(type: execute) {
			User(filter: {name: {_eq: "Islam"}}) {
			        name
			        age
      			}
		}`,
	Asserter:  testUtils.NewExplainAsserter().WithDocFetches(1).WithFieldFetches(2).WithIndexFetches(1),
},
```
This makes it repetitive and error-prone because one can change one request and forget to change the other.

Do you propose to use an arbitrary struct that will explicitly show the parameters? Or maybe a new test action, or to alter an existing one?

planner/planner.go
```diff
@@ -299,13 +301,67 @@ func (p *Planner) expandMultiNode(multiNode MultiNode, parentPlan *selectTopNode
...
+						filteredSubFields[fieldName] = subProp.Index
+					}
+				}
+			}
+		}
+	}
+	return filteredSubFields
+}
+
+func (p *Planner) expandTypeJoin(node *invertibleTypeJoin, parentPlan *selectTopNode) error {
```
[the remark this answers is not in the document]
islamaliev: I thought it was some kind of convention to prefix methods with "expand" for plan optimisation phase :)
But okay, I will add some info

islamaliev:
> I would like to see performance tests for non-indexed vs indexed queries, on both read and write. Without these we are not really testing that this code change is doing anything useful, and could be seen as an explain-feature or a simple refactor.

Do you mean adding tests to `tests/bench`?

planner/select.go
```diff
@@ -285,7 +286,29 @@ func (n *selectNode) initSource() ([]aggregateNode, error) {
...
+	return aggregates, nil
+}
+
+func findFilteredByIndexedField(scanNode *scanNode) immutable.Option[client.FieldDescription] {
+	if scanNode.filter != nil {
+		indexedFields := scanNode.desc.CollectIndexedFields()
+		for i := range indexedFields {
+			typeIndex := scanNode.documentMapping.FirstIndexOfName(indexedFields[i].Name)
+			if scanNode.filter.HasIndex(typeIndex) {
+				return immutable.Some(indexedFields[i])
```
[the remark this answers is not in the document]
islamaliev: because the first implementation should be relatively simple.
More sophisticated analysis and optimisations will come later.
There is a bunch of things that can be optimised here, but implementing them all at once would make this PR humongous.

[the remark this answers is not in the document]
AndrewSisley: Ah, this is us deliberately ignoring other indexes for now?

Can you please document that in the code?

AndrewSisley:
> Do you mean adding tests to `tests/bench`?

Adding to `tests/bench` would probably be the simplest for now, we could also assert on one being faster than the other (with appropriate protections against variation).

At the moment there is nothing that really tests the behaviour change that the users care about (query speed).

planner/planner.go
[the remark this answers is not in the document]
AndrewSisley: cheers Islam :)

db/fetcher/indexer.go
```diff
@@ -0,0 +1,165 @@
...
+		property := &encProperty{
+			Desc: f.indexedField,
+			Raw:  indexKey.FieldValues[0],
+		}
+
+		f.doc.key = indexKey.FieldValues[1]
+		f.doc.properties[f.indexedField] = property
+		f.execInfo.FieldsFetched++
+
+		if f.docFetcher != nil && len(f.docFields) > 0 {
```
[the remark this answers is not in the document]
islamaliev: No, I didn't run any performance checks because it's not the indexer's job to do it.
The indexer's job is to fetch the indexed field and determine the dockey it belongs to. The rest is supposed to be done by a dedicated `DocumentFetcher` that does exactly that.

Moreover, it does some other stuff that is necessary, like running filter on the rest of the fields.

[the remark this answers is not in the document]
AndrewSisley: > The indexer's job is to fetch the indexed field and determine the dockey it belongs to

That is not what it does though?  The indexer yields full documents.  Internally it is implemented so that it calls a private fetcher instance to get the other stuff non-indexed fields, but it does not need to be implemented that way.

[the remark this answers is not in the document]
islamaliev: > That is not what it does though? 

I do not quite follow. Yes, int the end the indexers does the whole job of fetching the full requested document, but fetching of non-indexed fields is delegated.

> but it does not need to be implemented that way.

Of course it doesn't have to be this way. There are many ways of implementing this.
Do you have a specific concern about this solution or a suggestion?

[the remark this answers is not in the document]
AndrewSisley: > Do you have a specific concern about this solution or a suggestion?

Yes, I was curious as to whether you had looked into the performance difference of using the inner fetcher vs just using txn.Get for each field, and considered whether such an implementation would be easier to read.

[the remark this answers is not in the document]
islamaliev: sorry Andy, I perceived your comment originally a bit sceptical. But now I gave it another thought and it seems to be an interesting option to consider. 
I think though that it should be part of another task as this one is already large enough.
I've created a task to look into it a bit later https://github.com/sourcenetwork/defradb/issues/1936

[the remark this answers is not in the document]
jsimnz: Commented on the new issue mentioned, not will add here.

I don't think a separate `Get()` method should be used. There's a number of things the fetcher can, should, and might do both now and in the future. I'd prefer not to implement the same or similar functionality in various places.

What is the concern for using the inner fetcher? Over fetching?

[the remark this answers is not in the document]
AndrewSisley: > What is the concern for using the inner fetcher?

No concern, but `txn.Get` would almost certainly be much simpler, and I was curious as to what the performance diff would be like (I'd expect slight gains for small numbers of returned docs, with increasing losses as the number of returned docs increases).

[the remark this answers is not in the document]
islamaliev: the whole idea of secondary indexes is about performance. If there a way to gain some more performance boost I think it should consider it.
But agree that adding a similar logic, or creating a new class is the most optimal way. But maybe we could add another configuration to the `DocumentFetcher` so that it fetches the value directly instead of seeking the next field every time. For some queries on a large data set it might prove quite beneficial I believe.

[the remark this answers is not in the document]
jsimnz: If my memory is correct, Get is only faster for a very low number of calls, compared to Iterate.

Secondary indexes are indeed about performance, and from the benches it's doing a good job of it. 

It seems needlessly complex/redundant to duplicate the fetcher work, especially as they're are already several items in the works to improve the efficiency and perf of the fetcher. Additionally, as we add more CRDT types, there may be a need to add some more custom logic to the document fetching process, which would again need to be duplicated.

The cost of the fetcher can also be amortized by fetching with multiple spans (each span as a  separate doc key).

So at the moment I don't see the immediate need/benefit.

tests/bench/fixtures/fixtures.go
```diff
@@ -104,7 +128,16 @@ func ExtractGQLFromType(t any) (string, error) {
...
+
+		directive := ""
+		if g.options != nil {
+			if dirsMap, ok := g.options.directives[name]; ok {
+				if dirs, ok := dirsMap[fname]; ok {
+					directive = " " + strings.Join(dirs, " ")
+				}
+			}
+		}
+		fmt.Fprintf(&buf, "\t%s: %s%s\n", fname, gqlType, directive)
```
AndrewSisley, line 140: todo: Please document this block, and maybe rename `directive` to `directives`, as it appears to hold multiple.

tests/bench/fixtures/fixtures.go
```diff
@@ -28,18 +29,41 @@ var (
...
 
-	schema string
-	types  []any
+	schema  string
+	types   []any
+	options *Options
 }
 
-func ForSchema(ctx context.Context, schemaName string) Generator {
+func ForSchema(ctx context.Context, schemaName string, options *Options) Generator {
```
AndrewSisley, line 61: suggestion: I think this would be much nicer if `options` was optional, instead of forcing the default to call it with an explicit nil param (there are lots of `ForSchema(foo, bar, nil)` calls now).  I'd also suggest changing `Options` to something array based, it would problably fit better.

```
type Options = func(Generator) Generator // or similar, perhaps you'd want an Options struct on Generator, but keep the func array as a param to ForSchema
func ForSchema(ctx context.Context, schemaName string, options ...Options)
```

tests/integration/utils2.go
```diff
@@ -369,110 +366,159 @@ func executeTestCase(
...
+	case SetupComplete:
+		// no-op, just continue.
+
+	default:
+		t.Fatalf("Unknown action type %T", action)
+	}
+}
+
+func benchmarkAction(
+	t *testing.T,
```
AndrewSisley, line 488: todo: Please remove the `t` param, it is on `state`

tests/integration/index/query_performance_test.go
```diff
@@ -0,0 +1,106 @@
...
+				Reps:   benchReps,
+				Action: testUtils.Request{Request: req},
+				Result: &benchResIndexed,
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test2)
+
+	for dbt, regularVal := range benchResRegular.ElapsedTime {
```
AndrewSisley, line 97: todo: This should be done within the `Benchmark` action, doing it here escapes the framework and will increase maintenance costs in the long run if we keep doing this.

I think you have also missused the framework somewhat by including two `testUtils.TestCase`s in the same test.  You should be able to remove the need for this by reworking the `Benchmark` action, perhaps so that the test case looks something like:
```
func TestQueryPerformance_Simple(t *testing.T) {
    test := testUtils.TestCase{
	    Actions: []any{
		    testUtils.SchemaUpdate{
			    Schema: `
				    type Users {
					    name: String
					    age: Int
					    email: String
					    verify: Boolean
				    }
			    `,
		    },
		    testUtils.SchemaUpdate{
			    Schema: `
				    type IndexedUsers {
					    name: String
					    age: Int @index
					    email: String
					    verify: Boolean
				    }
			    `,
		    },
		    testUtils.GenerateDocs{
                Count: 500,
            },
		    testUtils.Benchmark{
			    Reps:   benchReps,
			    Base: testUtils.Request{Request: "..."},
			    Comparee: testUtils.Request{Request: "..."}, // Comparee is not a word, try and find a better one :D
			    Factor: 5,
		    },
	    },
    }

}
```

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -0,0 +1,310 @@
...
+		User(filter: {
+			devices: {model: {_eq: "iPhone 10"}}
+		}) {
+			name
+		}
+	}`
+	test := testUtils.TestCase{
+		Description: "Filter on indexed relation field in 1-N relation",
+		Actions: []any{
+			createSchemaWithDocs(`
```
AndrewSisley, line 37: todo: I do think this is a bad use-case for auto-generating docs.  This tests (and many others here) are testing a very specific set of conditions and results, and really do care about what the initial state is.  At the moment, in this test (and others) I cannot see what that initial state is and if I want to know I have to spend considerable effort figuring that out.

Please change these, so that what state is being tested is very obvious to anyone reading.

tests/integration/index/query_performance_test.go
AndrewSisley: Once we are happy with the test action, I would want to know how expensive it is, and I would want more of them.  At the moment we only really have a single test that covers the behaviour that the users care most about.

tests/integration/utils2.go
```diff
@@ -1670,10 +1718,11 @@ func assertRequestResults(
 	result *client.GQLResult,
 	expectedResults []map[string]any,
 	expectedError string,
+	asserter ResultAsserter,
 	nodeID int,
 	anyOfByField map[docFieldKey][]any,
 ) bool {
-	if AssertErrors(s.t, s.testCase.Description, result.Errors, expectedError) {
+	if s.isBench || AssertErrors(s.t, s.testCase.Description, result.Errors, expectedError) {
```
AndrewSisley, line 1725: question: Why `s.isBench` here?

islamaliev: to skip assertion. If you run benchmark you don't bother asserting the query result, hence the expected (not set) value would be `nil`

shahzadlone: > to skip assertion. If you run benchmark you don't bother asserting the query result, hence the expected (not set) value would be `nil`

might be worth it to document that here.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: I strongly disagree. The initial state is put in a separate file you don't need to familiarise yourself with it every time you read another test case, as it's the same. Hence you do it once.
The tests with indexes (especially with relations) require a dozen (or better couple) of docs.
Soon I will be adding even more complex scenarios with one-to-many-to-many and other relations for which I think at least 40 docs would make sense. The test setup will take around 300 lines which I don't think anyone will have fun to read. Moreover, no developer wants to manually maintain all dockeys and relations.

AndrewSisley: > as it's the same

It is the same for a handful of index-explain tests.  If people only had to look at these tests that would be fine, but we have many other tests to look at too.  I'd really rather not have to learn stuff like this every time I enter a new set/sub-framework of integration tests.

It's not like you need hundreds of them here.  You are asserting that 3 very specific documents are being returned, and reading the test I have no idea whether they exist, only they exist, what their other field values are etc.  I can't even see the values that the filter is acting on.

AndrewSisley: > Moreover, no developer wants to manually maintain all dockeys and relations.

So improve the way we do that, we already have very similar logic in the test framework for handling this for mutations.  Don't hide the important stuff along with the dockeys.

datastore/util.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/ipfs/go-datastore/query"
+)
+
+// DeserializePrefix deserializes all elements with the given prefix from the given storage.
+// It returns the keys and their corresponding elements.
+func DeserializePrefix[T any](
+	ctx context.Context,
+	prefix string,
+	storage ds.Read,
```
fredcarle, line 27: suggetion: Rename param to store to match what we use elsewhere.

datastore/util.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, nil, err
+	}
+	return keys, elements, nil
+}
+
+// FetchKeysForPrefix fetches all keys with the given prefix from the given storage.
+func FetchKeysForPrefix(
+	ctx context.Context,
+	prefix string,
+	storage ds.Read,
```
fredcarle, line 61: suggestion: Rename param to store to match what we use elsewhere.

datastore/util.go
fredcarle, line 1: thought: I'm not a fan of `util.go` type files because they tend to become dump files overtime. 

suggestion: Move the content of this file to `store.go` or change the name of this file to something more descriptive of the content (`prefix_query.go`)

db/fetcher/encoded_doc.go
```diff
@@ -131,6 +131,24 @@ func Decode(encdoc EncodedDocument) (*client.Document, error) {
 	return doc, nil
 }
 
+// MergeProperties merges the properties of the given document into this document.
+// If values of the current document are overwritten.
```
fredcarle, line 135: todo: ~~.~~ // ~~I~~if values...

db/fetcher/indexer_iterators.go
```diff
@@ -0,0 +1,458 @@
...
+		var key connor.FilterKey
+		for key, filterVal = range condMap {
+			break
+		}
+		opKey := key.(*mapper.Operator)
+		op = opKey.Operation
+		break
+	}
+
+	if op == opEq || op == opGt || op == opGe || op == opLt || op == opLe || op == opNe {
```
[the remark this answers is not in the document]
fredcarle: I second this and would simplify une step further:
```go
switch op {
  case opEq:
    return &eqIndexIterator{
                indexKey:  indexDataStoreKey,
                filterVal: getCbor(),
                execInfo:  execInfo,
            }, nil
...
```

client/index.go
```diff
@@ -37,3 +37,22 @@ type IndexDescription struct {
 	// Fields contains the fields that are being indexed.
 	Fields []IndexedFieldDescription
 }
+
+// CollectIndexedFields returns all fields that are indexed by all collection indexes.
+func (d CollectionDescription) CollectIndexedFields() []FieldDescription {
```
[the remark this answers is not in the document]
islamaliev: It's not that straightforward, that part of code that call `CollectIndexedFields` don't have other references to schema. And making a lot of changes in this PR just to make it possible is not reasonable.
I think the second one to merge will have to adjust. I don't mind making this change, but I'd need the other PR to be merged.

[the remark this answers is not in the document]
AndrewSisley: I don't see what is complex about doing:
```
d.CollectionDescription(d.Schema)
```

tests/integration/index/query_with_relation_filter_test.go
islamaliev: > You are asserting that 3 very specific documents are being returned

This doesn't mean that the initial state needs only 3 docs. These 3 docs need to be selected from others, meaning it makes sense to have at least 5, each of them has related docs (1-N relation), which means on average 2 sub docs for every doc (to make index testing meaningful). 

Therefore roughly 15 docs would make sense to the test. It's not fun to deal with this ungrateful routine work. And having several of such test, would make a developer (at least) almost blind to the differences between them.

islamaliev: > dog: 2  // the doc index of Islam's dog  

it helps only a bit, there is still a lot of bookkeeping to do, and large test bodies are not very friendly.

db/fetcher/indexer_iterators.go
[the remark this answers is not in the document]
islamaliev: I like the way it looks, but reality will be more list this:
```go
switch op {
    case opEq:
        valueBytes, err := getCbor()
        if err != nil {
            return nil, err
        }
        return &eqIndexIterator{
	    indexKey:  indexDataStoreKey,
	    filterVal: valueBytes,
	    execInfo:  execInfo,
	}, nil
    case opGt:
        valueBytes, err := getCbor()
        if err != nil {
            return nil, err
        }
        return &filteredIndexIterator{
	    indexKey: indexDataStoreKey,
	    filter:   &gtIndexCmp{value: valueBytes},
	    execInfo: execInfo,
	}, nil
   case opIn ...
   default: 
       return nil, errors.New("invalid index filter condition")
}
```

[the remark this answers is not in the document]
islamaliev: I have changed it a bit. Check it please. Another option would be like this:
```go
	switch op {
	case opEq, opGt, opGe, opLt, opLe, opNe:
		writableValue := client.NewCBORValue(client.LWW_REGISTER, filterVal)

		valueBytes, err := writableValue.Bytes()
		if err != nil {
			return nil, err
		}

		if op == opEq {
			return &eqIndexIterator{
				indexKey:  indexDataStoreKey,
				filterVal: valueBytes,
				execInfo:  execInfo,
			}, nil
		}

		scanningIter := &scanningIndexIterator{indexKey: indexDataStoreKey, execInfo: execInfo}

		if op == opNe {
			scanningIter.matcher = &neIndexMatcher{value: valueBytes}
			return scanningIter, nil
		}

		byteValuesMatcher := &indexByteValuesMatcher{value: valueBytes}

		switch op {
		case opGt:
			byteValuesMatcher.evalFunc = func(res int) bool { return res > 0 }
		case opGe:
			byteValuesMatcher.evalFunc = func(res int) bool { return res > 0 || res == 0 }
		case opLt:
			byteValuesMatcher.evalFunc = func(res int) bool { return res < 0 }
		case opLe:
			byteValuesMatcher.evalFunc = func(res int) bool { return res < 0 || res == 0 }
		}

		scanningIter.matcher = byteValuesMatcher
		return scanningIter, nil
	case opIn, opNin:
		inArr, ok := filterVal.([]any)
		if !ok {
			return nil, errors.New("invalid _in/_nin value")
		}
		valArr := make([][]byte, 0, len(inArr))
		for _, v := range inArr {
			writableValue := client.NewCBORValue(client.LWW_REGISTER, v)
			valueBytes, err := writableValue.Bytes()
			if err != nil {
				return nil, err
			}
			valArr = append(valArr, valueBytes)
		}
		if op == opIn {
			return newInIndexIterator(indexDataStoreKey, valArr, execInfo), nil
		} else {
			return &scanningIndexIterator{
				indexKey: indexDataStoreKey,
				matcher:  newNinIndexCmp(valArr, false),
				execInfo: execInfo,
			}, nil
		}
	case opLike:
		return &scanningIndexIterator{
			indexKey: indexDataStoreKey,
			matcher:  newLikeIndexCmp(filterVal.(string), true),
			execInfo: execInfo,
		}, nil
	case opNlike:
		return &scanningIndexIterator{
			indexKey: indexDataStoreKey,
			matcher:  newLikeIndexCmp(filterVal.(string), false),
			execInfo: execInfo,
		}, nil
	}
```
Let me know if you like it

db/index_test.go
```diff
@@ -779,7 +780,7 @@ func TestGetIndexes_IfSystemStoreHasInvalidData_ReturnError(t *testing.T) {
 	mockedTxn.MockSystemstore.EXPECT().Query(mock.Anything, mock.Anything).Return(q, nil)
 
 	_, err := f.getAllIndexes()
-	assert.ErrorIs(t, err, NewErrInvalidStoredIndex(nil))
+	assert.ErrorContains(t, err, "invalid character")
```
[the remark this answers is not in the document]
islamaliev: because now `DeserializePrefix` is extracted to a common place and not just indexes specific, therefore it doesn't return index-specific error

tests/integration/utils2.go
islamaliev: You might want to benchmark an error path as well. Anyway I don't think asserting anything but performance is the job of `BenchmarkAction`

AndrewSisley: > Anyway I don't think asserting anything but performance is the job of BenchmarkAction

No, but it is always nice to know when you have messed up and broken the benchmark and caused it to error out instead of doing what you want it to.

If you never check that it has errored you will never know unless you happen to be debugging it (or you notice that performance is now 50 times better etc :)).

db/index_test.go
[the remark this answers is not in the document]
AndrewSisley: It should still return a consistent error though no?  Or is the error returned flaky?

db/fetcher/indexer_iterators.go
```diff
@@ -0,0 +1,458 @@
...
+
+func (cmp *likeIndexCmp) doesMatch(value string) bool {
+	switch {
+	case cmp.hasPrefix && cmp.hasSuffix:
+		return strings.Contains(value, cmp.filterValue)
+	case cmp.hasPrefix:
+		return strings.HasSuffix(value, cmp.filterValue)
+	case cmp.hasSuffix:
+		return strings.HasPrefix(value, cmp.filterValue)
+	case len(cmp.startAndEnd) == 2:
```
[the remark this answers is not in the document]
AndrewSisley: cheers :)

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: > It's not fun to deal with this ungrateful routine work

No it is not, but it is well worth a bit of upfront tedium to avoid future misery for maintainers later.

My problem is that if this test fails, potentially when working on an area unspecific to indexes, when the developer looks at this test instead of seeing the minimized set of documents required for this test, they have to instead navigate through `createSchemaWithDocs` (a very dense file), then to `docs.go`, and then they are presented with a massive dataset containing fields and documents that they really do not care about.  

Worse, they are not free to change anything in this file, as it will change X number of other tests that they may never have seen before and have no desire to change.  This is a very significant liability in the codebase.  Changing anything in `docs.go` is extremely unsafe.

jsimnz: I think we need better utilities for generating large sets of documents with specific properties or relationships.

I worked on something like this a long time ago when doing the initial benchmark suite, that did exactly this (although I'm not sure the code is a good enough state to be reused).

The crux here is that the tests don't care about The specifics of the values of the docs, only that they exist for numeric reasons (eg: number of fetches). So generating is a better route than explicitly writing it out.

However, based on my previous work, i know this to be a largish undertaking to create that kind of framework. 

I think it's important to have this kind of utility, but I think it's too large in scope to put in this PR.

I think we can define the ergonomics and design of such a framework and complete it after the PR, and leave this as is if we agree it's a temp measure.

AndrewSisley: > The crux here is that the tests don't care about The specifics of the values of the docs

These do.  They are filtering on specific values, and asserting that specific values are returned.  I am not against us having something like you described, or what Islam has built, but I am very much against it's use here.

db/fetcher/indexer_iterators.go
```diff
@@ -0,0 +1,458 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	i.resultIter = iter
+
+	return nil
+}
+
+type gtIndexCmp struct {
```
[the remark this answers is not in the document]
fredcarle: +1

tests/integration/index/query_with_relation_filter_test.go
islamaliev: > They are filtering on specific values, and asserting that specific values are returned

well, it's hard to assert on @explain results if you don't know exactly how many documents you are going to fetch and filter. Explain result here is the main indicator that the indexes are working and values are irrelevant there.
And the assertion on logical output is there just to make sure it's still working as expected, and partly to show why explain results are what they are.

> I think we can define the ergonomics and design of such a framework and complete it after the PR, and leave this as is if we agree it's a temp measure.

I also was thinking about making a framework out of it, but it doesn't cover yet some relation cases. So I might take care of this later.

planner/filter/unwrap_relation.go
```diff
@@ -0,0 +1,86 @@
...
+//		}
+//	}
+//
+// with given "published" field will return
+//
+//	{
+//		"rating": {
+//			"_gt": 4.0
+//		}
+//	}
```
fredcarle, line 35: praise: This doc is very nice 👌

tests/integration/utils2.go
islamaliev: if we assume that we will never benchmark an error path, then it makes sense to check error here.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: I've created a task https://github.com/sourcenetwork/defradb/issues/1962
I might have missed few details, so feel free to edit it

tests/bench/collection/simple_create_many_test.go
```diff
@@ -19,15 +19,15 @@ import (
 
 func Benchmark_Collection_UserSimple_CreateMany_Sync_0_10(b *testing.B) {
 	ctx := context.Background()
-	err := runCollectionBenchCreateMany(b, ctx, fixtures.ForSchema(ctx, "user_simple"), 0, 10, true)
+	err := runCollectionBenchCreateMany(b, ctx, fixtures.ForSchema(ctx, "user_simple", nil), 0, 10, true)
```
fredcarle, line 22: thought: It's an optional parameter so it doesn't need to explicitely be set to nil.

AndrewSisley: Just noting that I consider this a part of https://github.com/sourcenetwork/defradb/pull/1925#discussion_r1352853804 and these lines should now be reverted

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: LGTM, cheers Islam :)

tests/integration/utils2.go
AndrewSisley: > if we assume that we will never benchmark an error path, then it makes sense to check error here.

We might want to at somepoint, the bench action should probably allow this then.  Such a feature of the action could be added when the need arises though, not now.  I do think it is important to know right now if the benches stop working though.

fredcarle (approved):
LGTM. Nice work Islam. Just one minor thought on the extra param ion the bench tests.

fredcarle (approved):
LGTM. Nice work Islam. Just one minor thought on the extra param ion the bench tests.

Note: Make sure all conversations with Andy are resolved before merging.

db/index_test.go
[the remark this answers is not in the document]
AndrewSisley: todo: I would like this conversation resolved before merge please.

I also wish to clarify that I do see errors as public 'things', and thus changing an error is a public and often breaking change.

db/fetcher/indexer_iterators.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	res, err := f.matcher.Match(indexKey)
+	if err != nil {
+		f.err = err
+		return false
+	}
+	return res
+}
+
+type execInfoIndexMatcherDecorator struct {
```
AndrewSisley, line 199: todo: Sorry to be a bother, whilst this name is longer, I still have no idea what it is or what it is for - could you please document it?

db/fetcher/indexer_iterators.go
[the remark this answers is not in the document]
AndrewSisley: The new (pushed) code looks good, I still dislike the nested/duplicated switch, but you've had a play around and I assume the flatter option ended up looking messier.

AndrewSisley (approved):
LGTM, please resolve the remaining conversations before merge though.

I have a request for future PRs too - when resolving a conversation, can you please comment in the conversation/thread what you have done to resolve it.  It makes it easier to review, as I don't have to then go and re-find the location (which can sometimes be tricky due to github/other changes), and then figure out what has been done to resolve it (also sometimes tricky and occasionally requiring guesswork).

db/index_test.go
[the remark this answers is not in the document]
fredcarle: The error used to be wrapped in `DeserializePrefix`. It is no longer wrapped and returns an error from the underlying datastore and not one from Defra specifically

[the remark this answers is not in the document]
AndrewSisley: Then we can use error.Is on the error returned from the underlying datastore no?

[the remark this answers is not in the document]
fredcarle: Looked a bit closer and in this case it's an json Marshalling error. We could probably do:
```go
var sErr *json.SyntaxError
assert.ErrorAs(t, err, &sErr)
```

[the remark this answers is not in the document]
islamaliev: added a dedicated error `datastore.NewErrInvalidStoredValue`

tests/integration/utils2.go
islamaliev: okay, changed

db/fetcher/indexer_iterators.go
islamaliev: done

PR #1927, opened 2023-10-04
refactor: Remove net GRPC API

nasdf:
## Relevant issue(s)

N/A

## Description

This PR removes the GRPC API from the net package. The HTTP and CLI interfaces now include this functionality.

## Tasks

## How has this been tested?

`make test`

- MacOS


net/node_test.go
fredcarle, line 1: question: I've never seen these test being flaky. I'm not sure the tests are relevant anymore if we don't check the outcome. What do you think caused them to become flaky?

tests/integration/net/state/simple/replicator/with_create_update_test.go
```diff
@@ -141,17 +141,20 @@ func TestP2POneToOneReplicatorDoesNotUpdateDocExistingOnlyOnTarget(t *testing.T)
...
-				TargetNodeID: 1,
-			},
 			testUtils.CreateDoc{
 				// This document is created in all nodes
 				Doc: `{
 					"Name": "John",
 					"Age": 21
 				}`,
 			},
+			testUtils.ConfigureReplicator{
```
AndrewSisley, line 151: todo: This is changing what the test tests, and if there is flakiness here, it should probably be handled with a `WaitForSync` action instead.

That said, where have you seen this flakiness? Do you have an issue or an error log?

tests/integration/p2p.go
```diff
@@ -74,10 +76,10 @@ type SubscribeToCollection struct {
 	// them to this node.
 	NodeID int
 
-	// CollectionIDs are the collection IDs (indexes) of the collections to subscribe to.
+	// CollectionID is the collection ID (indexes) of the collection to subscribe to.
 	//
 	// A [NonExistentCollectionID] may be provided to test non-existent collection IDs.
-	CollectionIDs []int
+	CollectionID int
```
AndrewSisley, line 82: praise: This is a good change, thanks - the test action should reflect the structure of the production operation it represents.  `AddP2PCollection` takes a single id, so so should the action IMO

tests/integration/net/state/simple/peer/subscribe/with_add_test.go
```diff
@@ -224,103 +228,6 @@ func TestP2PSubscribeAddSingleErroneousCollectionID(t *testing.T) {
...
-				}`,
-				Results: []map[string]any{},
-			},
-		},
-	}
-
-	testUtils.ExecuteTestCase(t, test)
-}
-
-func TestP2PSubscribeAddValidThenErroneousCollectionID(t *testing.T) {
```
AndrewSisley, line 272: question: Why have these tests been removed?

net/node.go
```diff
@@ -52,7 +52,7 @@ var evtWaitTimeout = 10 * time.Second
 // Node is a networked peer instance of DefraDB.
 type Node struct {
 	// embed the DB interface into the node
-	client.DB
+	DB client.DB
```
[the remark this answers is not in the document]
nasdf: I had to temporarily because `client.DB` now has the same method signatures. For example `SetReplicator` here and in `client.DB` are the same.

net/node_test.go
nasdf: When it failed it was logging an extra line. My guess is that it was always prone to being flaky. Now that we run the parallel test matrix it just appears more often.

tests/integration/net/state/simple/replicator/with_create_update_test.go
nasdf: The CLI tests fail with the original test setup. If the replicator syncs the document before the second node attempts to create it, then it will fail. I will swap it for a `WaitForSync` like you suggested.

AndrewSisley: But only the CLI tests failed/were flaky?  That suggests a problem with the production code, not the test.  

If the other mutation types are not flaky, and the CLI is, the CLI is not behaving the same way as the others.

tests/integration/net/state/simple/peer/subscribe/with_add_test.go
nasdf: `TestP2PSubscribeAddSingleErroneousCollectionID` tests the same thing since only one subscription at a time is allowed.

AndrewSisley: `TestP2PSubscribeAddSingleErroneousCollectionID` does not test that other existing collection-peer subscriptions are unaffected by the attempt to configure a bad collection ID.

tests/integration/net/state/simple/replicator/with_create_update_test.go
nasdf: I believe this is due to the CLI being the slowest implementation. If you look through the original test steps you can see where the race condition happens. The document syncs from node 0 to node 1 before node 1 has the chance to create it.

tests/integration/p2p.go
nasdf: I'm happy to consider either option as long as it matches the client API.

tests/integration/net/state/simple/peer/subscribe/with_add_test.go
nasdf: My bad, I had misunderstood these tests. I restored them with one minor modification to make it pass.

tests/integration/net/state/simple/replicator/with_create_update_test.go
AndrewSisley: > I believe this is due to the CLI being the slowest implementation to test.

That does not make sense to me though, as the slowness is outside of the execution bounds in which speed matters (RE race), and if anything the CLI should be more resilient to flakiness here, as there is a larger gap between submitting the Create, and the Update.

Where x = main test thread, y = create async stuff, z = update async stuff, * = mutation type specific x stuff:
```
x * // Create doc action start
x *
x y // 
x y
x y
x y
x y // Create doc action end
x * // Update doc action start
x *
x z
x z
x z
``` 
When running CLI tests * may occupy more time, but that should only have the same effect as adding in a wait/sleep, and should reduce the risk of hitting a race condition (where y and z overlap).

From memory I also believe that the `ConfigureReplicator` action includes a wait anyway.

`TestP2PPeerReplicatorWithUpdate` also seems to run fine with the same create-config-update flow?

tests/integration/p2p.go
fredcarle: I'll let you decide which one you think creates a better user experience. 

```
defradb client p2pcollection add Users
defradb client p2pcollection add Address
```
vs
```
defradb client p2pcollection add Users,Address
```

tests/integration/net/state/simple/peer/subscribe/with_add_test.go
```diff
@@ -337,10 +341,6 @@ func TestP2PSubscribeAddNone(t *testing.T) {
 				SourceNodeID: 1,
 				TargetNodeID: 0,
 			},
-			testUtils.SubscribeToCollection{
```
AndrewSisley, line 340: todo: Either the name of this test should be changed slightly, or the test deleted (if this is tested elsewhere, which it might be).

tests/integration/net/state/simple/replicator/with_create_update_test.go
nasdf: I created an issue to start tracking tests with potential race condition #1935

AndrewSisley (approved):
LGTM, approving now, assuming Fred's comments are addressed sensibly, as well as/including the log test stuff (maybe split that out to another PR and merge this stuff now?). Plus the one new minor todo from me :)

Thanks Keenan :)

tests/integration/p2p.go
nasdf: I've updated it to support multiple collections like we discussed.

PR #1939, opened 2023-10-06
refactor: Deprecate CollectionDescription.Schema

### LINKED ISSUE #1913: Remove Schema from within CollectionDescription
Replace `CollectionDescription.Schema` with `CollectionDescription.SchemaVersionID`.

This will require changing a large number (+100) of references, PatchSchema (and tests), and the serialization/ID logic.  Try and avoid changing anything else in this ticket. 
```[tasklist]
### Tasks
- [ ] https://github.com/sourcenetwork/defradb/issues/1955
- [ ] https://github.com/sourcenetwork/defradb/issues/1956
- [ ] https://github.com/sourcenetwork/defradb/issues/1958
- [ ] https://github.com/sourcenetwork/defradb/issues/1963
- [ ] https://github.com/sourcenetwork/defradb/issues/1964
```

AndrewSisley:
## Relevant issue(s)

Partially Resolves #1913

## Description

Deprecate CollectionDescription.Schema.  Schema is not a sub-property of collection.

Removes as many references to CollectionDescription.Schema as possible without making any breaking changes.  Breaking changes will be made in a later PR. An exception has been made to the http API, which does have a breaking change in this PR - it can be pulled out of here if people want it more obvious in the develop commit history/change log.

This should minimise the conflicts generated by getting the large broad changes sorted out early, instead of being bogged down behind the more complex, breaking, changes coming later. 

Commits should be clean, and should bundle up the work into more readable chunks, would recommend reviewing them one by one.


cli/collection_describe.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/sourcenetwork/defradb/client"
 )
 
+type CollectionDefinition struct {
```
islamaliev, line 19: question: does it need to be a public definition of `cli` package?

Seems like it doesn't belong here

AndrewSisley: It is the public output type of the command.  And this is the only place that outputs it.

Where were you thinking it should be?

fredcarle: The struct doesn't need to be public. Only its  fields.

client/descriptions.go
```diff
@@ -41,12 +43,21 @@ func (col CollectionDescription) IDString() string {
 
 // GetFieldByID searches for a field with the given ID. If such a field is found it
 // will return it and true, if it is not found it will return false.
-func (col CollectionDescription) GetFieldByID(id FieldID) (FieldDescription, bool) {
-	if !col.Schema.IsEmpty() {
-		for _, field := range col.Schema.Fields {
-			if field.ID == id {
-				return field, true
-			}
+func (col CollectionDescription) GetFieldByID(id FieldID, schema *SchemaDescription) (FieldDescription, bool) {
```
fredcarle, line 46: question: This and the couple other methods bellow no longer need the `CollectionDescription`. Do you intend on turning those into methods of `SchemaDescription` in the future?

AndrewSisley: No, this and the others remain here as whilst they currently only access stuff on the schema, that is because the stuff they access is incorrectly on the schema.

For example, in this case we are getting fields by their ID, FieldID is a local concept that should exist only on collection, not on schema.

The function is located in the correct place, but for historical reasons the data structures are incorrect and need to be changed along with the implementation.

cli/collection_describe.go
fredcarle: It's specific to this CLI command and not used anywhere else. I really think this struct shouldn't be public. If you absolutely want it to be public, please at some documentation to the type.

client/descriptions.go
fredcarle: Thanks for clarifying

cli/collection_describe.go
AndrewSisley: When defining a http endpoint would you also declare their return types as private, hiding them from the godocs and any packages embedding the library?

fredcarle: Unless that type is returned by a public Go function, there is no use in making the type public. It adds no value to GoDocs as it's not relevant to any of the public functions. 

The HTTP API has to be documented in a specific API doc and the output JSON struct should be shared in that location.

AndrewSisley: :smile: That does surprise me :)

I do see these as the declaration of the data type/shape that is returned from the public endpoints, and as such they should be public.  Is just that the transmission of that shape is via []byte of json instead of []byte of the values only (which is what an instance of a struct really is, only masked/wrapped by the runtime).

I'll ping @nasdf to get his thoughts on this too, as we should be consistent with whatever we do here and the other non-embedded interfaces (including arguably LensVM).

fredcarle: It is returned from a public endpoint but that public endpoint isn't a public Go function. That endpoint won't show up in the godocs and the function itself is returning an error. The type/shape should be documented elsewhere but not with a public Go type.

nasdf: My preference would be to define it in the `client` package and reference it in the `cli` and `http`. That would also get rid of the duplicate definition in the `http` client.

db/collection.go
```diff
@@ -229,24 +207,38 @@ func (db *db) updateCollection(
 	ctx context.Context,
 	txn datastore.Txn,
 	existingDescriptionsByName map[string]client.CollectionDescription,
-	proposedDescriptionsByName map[string]client.CollectionDescription,
+	existingSchemaByName map[string]client.SchemaDescription,
+	proposedDescriptionsByName map[string]client.SchemaDescription,
```
fredcarle, line 211: thought(out of scope): I would like to point out that this in a really confusing set of params. We are updateing one collection but we are passing all collection, all schemas and all proposed schemas. In my opinion, only what is directly relevant should be passed to this function.

cli/collection_describe.go
fredcarle: I would be happy with that option :)

AndrewSisley: Nice okay - I'll try and figure something out.  Some other stuff might get re-jigged slightly as we already have a CollectionDefinition (interface) in client, but hopefully everything will be the better for it.

db/collection.go
AndrewSisley: This one is fiddly, and I agree that it is a bit of an eyesore. However they are all used and all need to be there (or within this func, we could probably (re)fetch them) due to relational fields.

It does get improved slightly in a later PR when we are able to change all their types to SchemaDescriptions (allowing us to only pass in two maps instead of 3).  Might be worth re-visiting this in https://github.com/sourcenetwork/defradb/pull/1965

AndrewSisley: I've added a todo in that PR, as one of the params still makes no sense there: https://github.com/sourcenetwork/defradb/pull/1965/files#r1358767728

Removing it leaves us with:
```
	ctx context.Context,
	txn datastore.Txn,
	existingSchemaByName map[string]client.SchemaDescription,
	proposedDescriptionsByName map[string]client.SchemaDescription,
	schema client.SchemaDescription,
	setAsDefaultVersion bool,
) (client.Collection, error) {
```
`existingSchemaByName` could be re-fetched inside `updateCollection`, but that feels fairly wasteful atm (although a planned repository mechanic would remove that issue).

fredcarle: They are used but in a wasteful way as we only care about one in each of the maps. The only proposedDescription that we care about is already passed as `schema` so that one can be removed completely. It should probably look like this:
```go
    ctx context.Context,
    txn datastore.Txn,
    existingSchema client.SchemaDescription,
    schema client.SchemaDescription,
    desc client.CollectionDescription
    setAsDefaultVersion bool,
) (client.Collection, error) {
```

Although the checks on the schema should probably be done before calling `updateCollection` and we would end up with just:
```go
    ctx context.Context,
    txn datastore.Txn,
    schema client.SchemaDescription,
    desc client.CollectionDescription
    setAsDefaultVersion bool, // if we call updateCollection, this should be implicit and removed from params.
) (client.Collection, error) {
```

AndrewSisley: > The only proposedDescription that we care about is already passed as schema so that one can be removed completely

This is not true, we must validate that all relational fields point to a valid schema.  This schema is rarely the one being updated.

> Although the checks on the schema should probably be done before calling updateCollection

I'me very much against that, as it moves the validation out of a request-language agnostic part of the codebase (db.collection.go) into a request-language specific area (patchSchema) of the codebase.  It would mean that if we add an alterative to patchSchema (including an embedded Go call) we'd have to rework the validation.

request/graphql/schema/collection.go
```diff
@@ -24,9 +24,32 @@ import (
 	"github.com/graphql-go/graphql/language/source"
 )
 
+type collectionDefinition struct {
```
fredcarle, line 27: thought: You'll be able to use the one in the client package after you move it there instead of redefining it here.

db/collection.go
fredcarle: Maybe the problem then is that the name of the function is `updateSchema` put it handles both schema and collection in the same function. Maybe once we get further in cleaning this up it won't be a problem anymore.

db/index_test.go
```diff
@@ -324,28 +304,6 @@ func TestCreateIndex_IfFieldHasNoDirection_DefaultToAsc(t *testing.T) {
 	assert.Equal(t, client.Ascending, newDesc.Fields[0].Direction)
 }
 
-func TestCreateIndex_IfNameIsNotSpecified_Generate(t *testing.T) {
```
islamaliev, line 327: question: Why did you remove this test?

http/handler_store.go
```diff
@@ -22,6 +22,11 @@ import (
 
 type storeHandler struct{}
 
+type CollectionDefinition struct {
```
islamaliev, line 25: note: whatever solution is for identical struct in `cli/collection_describe.go` that we discussed should also apply here I believe

db/index_test.go
AndrewSisley: This is documented in the commit message:

> One test was removed, TestCreateIndex_IfNameIsNotSpecified_Generate - this is because it only pretends to test what it tests - the name _is_ set later by the unit test framework, and as such is identical to other tests here.

db/collection_test.go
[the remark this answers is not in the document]
AndrewSisley: Yes

db/collection.go
AndrewSisley: for-reference: We spoke about this last Friday, and there is a task on a later PR to rename this function (when it stops being updateCollection)

cli/collection_describe.go
AndrewSisley: sorted, client.CollectionDefinition is now a struct

client/collection.go
```diff
@@ -16,24 +16,34 @@ import (
 	"github.com/sourcenetwork/defradb/datastore"
 )
 
+// CollectionDefinition contains the metadata defining what a Collection is.
+type CollectionDefinition struct {
```
fredcarle, line 20: thought: I'm a little worried that the name is so close to `CollectionDescription` that it will be a source of confusion. What do you think of `CollectionInfo`?

client/collection.go
```diff
@@ -16,24 +16,34 @@ import (
...
 	// Name returns the name of this collection.
 	Name() string
-	// Schema returns the SchemaDescription used to define this Collection.
-	Schema() SchemaDescription
 	// ID returns the ID of this Collection.
 	ID() uint32
 	// SchemaID returns the ID of the Schema used to define this Collection.
 	SchemaID() string
 
+	Definition() CollectionDefinition
```
fredcarle, line 41: suggestion: This should probably have documentation.

client/collection.go
AndrewSisley: I dislike `CollectionInfo` as it suffers from being a little ambiguous - could be read as containing info about what data is in the collection too, `CollectionDefinition` suffers less from that.

client/collection.go
AndrewSisley: will add

- [ ] add doc

client/collection.go
fredcarle: We should probably think of something better then.

AndrewSisley: > We should probably think of something better then.

I tried, and `CollectionDefinition` was the best I could come up with without spending an unreasonable amount of time on it.  I'm fairly happy with it given the problem.

islamaliev (approved):
LGTM. Nice work Andy!

db/collection.go
```diff
@@ -101,9 +94,11 @@ func (c *collection) newFetcher() fetcher.Fetcher {
 func (db *db) createCollection(
 	ctx context.Context,
 	txn datastore.Txn,
-	desc client.CollectionDescription,
-	schema client.SchemaDescription,
+	def client.CollectionDefinition,
 ) (client.Collection, error) {
+	schema := def.Schema
+	desc := def.Description
```
fredcarle, line 100: question: Why reallocate when you've already passed a copy of the definition? Using `def` directly would be clearer and you could assign to `col.def` directly from it.

AndrewSisley: It is a struct, and its children are structs, mutating them does not mutate the one on the parent, the parent must instead later be overwritten with a new inner.

```
// foo.bar.foobar is false
foo.bar.foobar = true
// foo.bar.foobar is still false, a new (and unreachable) bar instance has value true
```

fredcarle: I don't think that's true: https://go.dev/play/p/FHce-ajgFyr

AndrewSisley: :laughing: Maybe not, but something was very broken before I made that change. It is very deliberate and it doesnt work without it.  Could be something similar in the more complex real-world code.

client/collection.go
fredcarle: Just noting here for future reference:

I feel like `client.SchemaDescription` should just be `client.Schema` and `client.CollectionDescription` should be `client.Collection`. The interface should have a different name. Probably something like `client.CollectionAPI` or `client.CollectionHandler`. 

This way, `client.CollectionDefinition` would work.

db/collection.go
fredcarle: I just tested it and all tests are passing. Not sure what would have been broken.

AndrewSisley: I cannot see what you have changed, most likely it was schema that was important here, as that is what is being mutated most.

fredcarle (approved):
LGTM! Thanks Andy!

PR #1957, opened 2023-10-10
feat: Remove collection from patch schema

### LINKED ISSUE #1956: Remove collection from PatchSchema

AndrewSisley:
## Relevant issue(s)

Resolves #1956

## Description

Removes collectionDescription from patch schema.

PatchSchema now only uses collectionDescription internally to update it if requested (bool param).

This now means that the only place remaining (minus a util func that got missed) that refs collectionDescription.Schema should be the serialization.  This will be sorted out in a later PR, before removal of the property completely.


db/schema.go
```diff
@@ -193,13 +198,13 @@ func (db *db) getCollectionsByName(
...
 	patch jsonpatch.Patch,
-	collectionsByName map[string]client.CollectionDescription,
+	schemaByName map[string]client.SchemaDescription,
 ) (jsonpatch.Patch, error) {
-	fieldIndexesByCollection := make(map[string]map[string]int, len(collectionsByName))
-	for colName, col := range collectionsByName {
-		fieldIndexesByName := make(map[string]int, len(col.Schema.Fields))
-		fieldIndexesByCollection[colName] = fieldIndexesByName
-		for i, field := range col.Schema.Fields {
+	fieldIndexesByCollection := make(map[string]map[string]int, len(schemaByName))
```
fredcarle, line 203: suggestion: `fieldIndexesBySchema`

db/collection.go
```diff
@@ -228,12 +206,19 @@ func (db *db) createCollection(
 func (db *db) updateCollection(
```
AndrewSisley, line 206: todo: Spotted by Fred - this should no longer be called `updateCollection`.

- [ ] Rename to `updateSchema`

db/schema.go
```diff
@@ -128,31 +127,37 @@ func (db *db) patchSchema(ctx context.Context, txn datastore.Txn, patchString st
...
 		if err != nil {
 			return err
 		}
 
 		newCollections = append(newCollections, col)
-		newSchemaByName[col.schema.Name] = col.schema
 	}
 
 	for i, col := range newCollections {
 		col, err := db.updateCollection(
```
fredcarle, line 158: todo: Discussed via a video call but noting here. We should rename this function to `updateSchema`.

db/collection.go
fredcarle: Nothing that it should also be moved to `schema.go`

db/schema.go
fredcarle: resolving as it's been noted on the function itself.

db/schema.go
```diff
@@ -128,31 +127,37 @@ func (db *db) patchSchema(ctx context.Context, txn datastore.Txn, patchString st
...
+		if schema.Name == "" {
+			return ErrSchemaNameEmpty
+		}
+
+		collectionDescription, ok := collectionsByName[schema.Name]
+		if !ok {
+			return NewErrAddCollectionWithPatch(schema.Name)
+		}
+
+		col, err := db.newCollection(collectionDescription, schema)
```
fredcarle, line 149: thought: If we change the current `updateCollection` to `updateSchema`, we will still need an `updateCollection` function to handle the collection specific update to the new schema version. This part might be better off in that new `updateCollection` function.

db/collection.go
AndrewSisley: I think it should be co-located with the other functions that update the descriptions (including the future patchCollection), spreading those functions out will not aid readability. 

Your thought should be re-visited in https://github.com/sourcenetwork/defradb/issues/1963 once everything in this space makes a little more sense and we can move everything together.

db/schema.go
AndrewSisley: This will be handled in https://github.com/sourcenetwork/defradb/issues/1963 and https://github.com/sourcenetwork/defradb/issues/1963 - it was going to be done in https://github.com/sourcenetwork/defradb/pull/1965 but it proved to be a large undertaking to do it properly.

db/schema.go
AndrewSisley: cheers, changing :)

- [ ] Rename var to fieldIndexesBySchema

fredcarle (approved):
LGTM!

Issue #1964, opened 2023-10-13
Detangle collection storage


AndrewSisley:
The way we store collections makes no sense at the moment, and needs to be sorted out.

It should probably be done after https://github.com/sourcenetwork/defradb/issues/1963 though.

There will be the following keys:
```
[Collection.ID] => json // Json should be stored against an immutable index, i.e. not against `Name`
[Collection.Name] => [Collection.ID]  // This means collection names must be unique
```
We also need the following for the short-term, before Collection.GlobalID gets added, as otherwise if a collection is updated to a new schema version, we lose track of what collection it *could* (could as, this only narrows it down) be.
```
[Collection.SchemaVersionId]/[Collection.ID] => nil
```

jsimnz:
Can you add context here on what "no sense". The other issue regarding fetching has great additional info, let's spruce this one up too ;)

AndrewSisley:
> Can you add context here on what "no sense". The other issue regarding fetching has great additional info, let's spruce this one up too ;)

I don't plan on letting this issue live for very long, and it is both tricky and inexact to explain, and in some sense already solved in the WIP branch 1963-collection-feat-detangle

The keys and their values used to persist and index the collection description make no sense.  They are tangled, and with poor names and prefixes. They blend schema with collection very heavily.

jsimnz:
sounds good!

fredcarle:
Currently we have
```go
COLLECTION                        = "/collection/names" // {collectionName} -> schemaVersionID
COLLECTION_SCHEMA                 = "/collection/schema" // {rootSchema} -> schemaVersionID
COLLECTION_SCHEMA_VERSION         = "/collection/version/v" // {schemaVersionID} -> client.CollectionDescription
COLLECTION_SCHEMA_VERSION_HISTORY = "/collection/version/h" // {rootSchemaID}/{previousSchemaVersionID} -> schemaVersionID

SCHEMA_VERSION                    = "/schema/version" // {schemaVersionID} -> client.SchemaDescription 
```

This is what I think we should have:

```go
COLLECTION_NAME                   = "/collection/name" // {collectionName} -> client.CollectionDescription
COLLECTION_ID                     = "/collection/id" // {collectionID} -> collectionName

SCHEMA_VERSION                    = "/schema/version" // {schemaVersionID} -> client.SchemaDescription 
[...]

PR #1965, opened 2023-10-13
feat: Remove CollectionDescription.Schema

### LINKED ISSUE #1958: Store collection and schema descriptions against different keys
And sort out/split the various functions used to fetch and store them.

Consider moving both of them outside of `collection.go` (is a bit messy/busy there).

AndrewSisley:
## Relevant issue(s)

Resolves #1958

## Description

Removes CollectionDescription.Schema.

Also splits the storage of schema out from within collection.

The storage of schema has been broken out to a new sub-package of db, at the moment it is a very simple file, but collection will be moved there in https://github.com/sourcenetwork/defradb/issues/1964. I was planning on doing that in this PR (in part, to provide context for reviewers, as atm it is basically a single-file package), but it proved to be non-trivial due to some existing messiness in that space and was broken out to two more tasks.

I also wish for stuff in that directory to eventually follow a repository-like pattern, where stuff is cached (within a context/txn's context) instead of fetching from store on each call.

Moving this stuff out to a new directory instead of preserving it in the (already very large) db directory should make both db and the new sub-package a fair bit more cohesive and easier to read.


db/collection.go
```diff
@@ -228,25 +188,32 @@ func (db *db) createCollection(
 func (db *db) updateCollection(
 	ctx context.Context,
 	txn datastore.Txn,
-	existingDescriptionsByName map[string]client.CollectionDescription,
-	proposedDescriptionsByName map[string]client.CollectionDescription,
+	existingSchemaByName map[string]client.SchemaDescription,
+	proposedDescriptionsByName map[string]client.SchemaDescription,
 	desc client.CollectionDescription,
```
AndrewSisley, line 193: todo: Passing this in no longer makes any sense, not does fetching it in patchSchema.  Drop the patchSchema call to `getCollectionsByName`, and fetch it on demand in here.

See also https://github.com/sourcenetwork/defradb/pull/1939#discussion_r1358762878

- [ ] Remove unwanted complexity RE collectionDescription param

db/collection.go
```diff
@@ -151,53 +123,41 @@ func (db *db) createCollection(
...
-	collectionSchemaKey := core.NewCollectionSchemaKey(schemaID)
-	err = txn.Systemstore().Put(ctx, collectionSchemaKey.ToDS(), []byte(schemaVersionID))
+	collectionSchemaKey := core.NewCollectionSchemaKey(schema.SchemaID)
+	err = txn.Systemstore().Put(ctx, collectionSchemaKey.ToDS(), []byte(schema.VersionID))
 	if err != nil {
 		return nil, err
 	}
 
-	err = txn.Systemstore().Put(ctx, collectionKey.ToDS(), []byte(schemaVersionID))
+	err = txn.Systemstore().Put(ctx, collectionKey.ToDS(), []byte(schema.VersionID))
```
fredcarle, line 160: thought: These make no sense anymore. I think this is what you were referring to [here](https://github.com/sourcenetwork/defradb/issues/1964)

db/descriptions/schema.go
fredcarle, line 1: suggestion: Although I like the functions that you create here, I'm not a fan of the `descriptions` package. I would much rather see this under `db` and `db.schema.go` is still small enough to host the content of this file without causing trouble.

db/collection.go
AndrewSisley: Yes :)

db/descriptions/schema.go
fredcarle: I read the description. My comment still stands. I don't think that even after moving collection in there that it would make sense to me. I think having the content under a file with a representative name under `db` would be much better.

core/key.go
```diff
@@ -132,6 +133,12 @@ type CollectionIndexKey struct {
 
 var _ Key = (*CollectionIndexKey)(nil)
 
+type SchemaVersionKey struct {
```
islamaliev, line 136: todo: I think it needs a comment here

AndrewSisley: Cheers - I missed that, will add

- [ ] Doc key

core/key.go
```diff
@@ -257,6 +264,11 @@ func NewCollectionSchemaVersionKey(schemaVersionId string) CollectionSchemaVersi
 	return CollectionSchemaVersionKey{SchemaVersionId: schemaVersionId}
 }
 
+func NewCollectionSchemaVersionKeyFromString(key string) CollectionSchemaVersionKey {
```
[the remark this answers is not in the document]
AndrewSisley: I do not.  This is very simple, and tested by pretty much every integration test we have.  Unit tests would only slow down development (now and in the future).

db/descriptions/descriptions.go
islamaliev, line 1: suggestion: I think conventional way of naming packages in go is in singular form

db/descriptions/schema.go
fredcarle: I don't believe it would make any difference for maintainability. The schema stuff was already in there. You took it out of existing functions. If the files are well named and represent their contents, I don't really care how large the package is in terms of lines of code. Readability would be unchanged. I also care more about the size of functions than the size of files.

If you move collection to a sub package, we will end up with the sub-package having its parent as a dependency and eventually risk hitting circular dependency issues.

core/key.go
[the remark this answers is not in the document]
fredcarle: I had a closer look and I think that not unit testing our keys could cause bad changes to go unnoticed until the breaking change tests are executed which could be quite annoying.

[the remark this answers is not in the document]
AndrewSisley: I'm not really in the mood to get back into the integration vs unit test conversation at the moment.  I have written a large amount on this topic before and you know my opinion on this matter.

db/descriptions/descriptions.go
AndrewSisley: I think I always saw that as when talking about a single thing.  For example, the `db` package contains a single `db` definition, that can be used to instantiate multiple `db` instances.  In `descriptions` we have multiple `description` definitions (schema, collection, indexes - eventually), each of which can be used to instantiate multiple instances.

core/key.go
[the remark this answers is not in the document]
fredcarle: I don't think the mood of anyone is relevant to the conversation though. This is for the benefit of everyone on the team. And in this case a quick unit test would be a good approach.

db/descriptions/schema.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/ipfs/go-datastore/query"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/core"
+	"github.com/sourcenetwork/defradb/core/cid"
+	"github.com/sourcenetwork/defradb/datastore"
+)
+
+// CreateSchemaVersion creates and saves to the store a new schema version.
```
islamaliev, line 25: suggestion: "a new schema description version"

db/descriptions/schema.go
AndrewSisley: > If you move collection to a sub package, we will end up with the sub-package having its parent as a dependency and eventually risk hitting circular dependency issues.

Why would we do that?  There is nothing special about 'sub-packages', they are just packages, and we wont have circular dependencies here for the same reasons we don't have circular dependencies between any of our other packages.  That and there is absolutely no reason for this to reference anything minus stuff `client` and `core`, it should never be calling out to `db`, or `planner` etc.

> Readability would be unchanged.

This is simply not true, at least for me when I am reading a file that has lots of independent stuff all jumbled up within it.

I do not want a single file for this stuff, because schema should not be blended with collection or index, or likely, the repo mechanics (which will likely be shared by all of them).

`descriptions/schema.go` should not be mixed with `db/collections.go` or anything else in that package, it should also not be mixed in with `descriptions/collection.go`.  Keeping them separate will make my life easier.  Both in the short term as I finish sorting this stuff out, and long term.

db/descriptions/schema.go
```diff
@@ -0,0 +1,170 @@
...
+	var desc client.SchemaDescription
+	err = json.Unmarshal(buf, &desc)
+	if err != nil {
+		return client.SchemaDescription{}, err
+	}
+
+	return desc, nil
+}
+
+// GetSchemas returns the schema of all the default schemas in the system.
```
islamaliev, line 108: question: what is "default schema"? Is it a common term in defra?

AndrewSisley: It is a fairly new term that came to be when we allowed the toggling of schema versions.  There is a word missing here, it should read  `default schema versions`. A default schema version is a version that is the one used by a collection/query/etc when not explicitly requesting a version.

- [ ] Tweak doc wording

db/descriptions/schema.go
fredcarle: > it should never be calling out to `db`, or `planner` etc.

`collection` depends on db. Maybe if you don't bring that in to the sub package but then why scatter collection related functionality between packages?

> I do not want a single file for this stuff, because schema should not be blended with collection or index, or likely, the repo mechanics (which will likely be shared by all of them).

Why not just add `collection_description.go` and `schema_description.go` instead of creating a sub-package?

db/descriptions/schema.go
AndrewSisley: I do disagree, and don't think adding `description` there adds any clarity.  It does also clash with Fred's long-term naming suggestion where `SchemaDescription` will be renamed to `Schema` (or maybe `SchemaVersion`, either way `Description` will be dropped).

db/descriptions/schema.go
```diff
@@ -0,0 +1,170 @@
...
+	collectionSchemaVersionQuery, err := txn.Systemstore().Query(ctx, query.Query{
+		Prefix:   collectionSchemaVersionPrefix.ToString(),
+		KeysOnly: true,
+	})
+	if err != nil {
+		return nil, NewErrFailedToCreateSchemaQuery(err)
+	}
+	defer func() {
+		if err := collectionSchemaVersionQuery.Close(); err != nil {
+			log.Error(ctx, NewErrFailedToCloseSchemaQuery(err).Error())
```
islamaliev, line 123: todo: https://github.com/sourcenetwork/defradb/pull/1450#discussion_r1212151800 :)

db/descriptions/schema.go
AndrewSisley: > collection depends on db...

I do not understand what you mean here.  You might be misunderstanding what will be living in this package.

> Why not just add collection_description.go and schema_description.go instead of creating a sub-package?

Because they are going to be separated by 10 or so other files in the `db` package that are alphabetically larger than the first and smaller than the second, with `index_description.go` somewhere in between, plus whatever files the repo stuff will use, all causing cohesion to be lost.

db/descriptions/schema.go
AndrewSisley: Hahaha, nice :grin: Thanks Islam :laughing: 

- [ ] Return error

core/key.go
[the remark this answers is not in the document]
AndrewSisley: > And in this case a quick unit test would be a good approach.

You think that, and perhaps Islam thinks that.  But you also know that I really don't think that, and that I have had a lot of my time spent lately maintaining unit tests that have falsely failed when changing unrelated parts of the codebase.

Are you looking to re-open the conversation on unit vs integration tests here in this PR, or force me to make a change that you know I think will waste time?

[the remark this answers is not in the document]
fredcarle: I think your point can be debated for tests that need to interact with multiple functions/types. In that case it can potentially be subjective. In this case however, it's a direct test of the key generation and its methods and it's the most direct and explicit way to flag if something is wrong with the key generation. Like I said, only the change detector would be warn us of something wrong but it would be quite vague as to what is actually wrong. Probably a `key does not exist` error or something of the sort.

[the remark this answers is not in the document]
AndrewSisley: > In this case however, it's a direct test of the key generation and its methods and it's the most direct and explicit way to flag if something is wrong with the key generation.

A unit test of this function would not spot any failing in key generation, it would only re-describe in lengthier terms that this function does what the two lines in it say it does.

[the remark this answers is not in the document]
fredcarle: There is more than one function here. The comment was on one line but it applies to all the added functions.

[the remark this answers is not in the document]
fredcarle: > it would only re-describe in lengthier terms that this function does what the two lines in it say it does.

That is basically what every test we have does. Integration or otherwise. They describe in lengthier terms what functions and features do without needing a human to hand test everything.

We think testing these directly is relevant, quick, and the resulting test very easy to maintain. Not sure it's worth debating any further. Up to you.

db/descriptions/schema.go
fredcarle: If `schema_description.go` is next to `schema.go` and functions inside `schema.go` call functions inside `schema_description.go`, I think that's pretty good cohesion.

Why would having `schema_description.go` next to `collection_description.go` be more cohesive than the above?

core/key.go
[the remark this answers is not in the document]
islamaliev: I understand Andy your frustration dealing with some unit tests, especially with those that I wrote with mocks for testing secondary indexes. I think it was a combination of me zealously trying to achieve high test coverage (I'm rethinking it now) and not ideal design that didn't let easily to get to certain code parts.

This code with keys on the other hand has no dependancies and therefore very straightforward to test with low maintenance in the future.

db/descriptions/schema.go
islamaliev: it was just observation as a reader that I was looking for "description" word because the method is in `descriptions` package, but it wasn't clear to me at first sight why.

db/descriptions/descriptions.go
islamaliev: as I said, there is no official rule for it, it's just community convention. 
The main argument for it how you going to use it. Yes, the package might deal with several different descriptions, but for the most packages most of of their public methods deal with a single entity (a.k.a description).
Like `context.Context` allows you to create (or wrap) different types of contexts, but you handle a single one at a time.

db/descriptions/schema.go
islamaliev: all right, the term "default schema versions" I know :)

core/key.go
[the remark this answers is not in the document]
AndrewSisley: > That is basically what every test we have does. Integration or otherwise. They describe in lengthier terms what functions and features do without needing a human to hand test everything.

Yes, the difference between unit and integration tests though is that integration tests test behaviour that is meaningful to end users, and unit tests do not.  

It really does not matter what this function does, so long as it behaves correctly given the parameters the rest of the code will provide it in production.  What that behaviour is does not matter.  Other conditions do not matter.

If this function changes, it will be because what it needs to do has changed.  In such a situation they unit tests would be useless, as they only describe what it used to do, not what it used to need to do, nor what it now needs to do.  Changing the behaviour of this function would essentially be a rip and replace, and they unit tests would essentially be ripped and replaced too.  The would be replaced with more useless unit tests that only describe what the production code does.  

They do not, and will not ever describe what it needs to do, because what it needs to do is totally imaginary and is not grounded in any real-world/user-visible use case.  They would be a totally pointless, vague, re-wording of the contents of the function, and would only get in the way if that imaginary use-case changes.

> I understand Andy your frustration dealing with some unit tests...

Whilst there is some overlap in my opinion of both this, and the index unit tests, the index unit tests that were removed were removed for different reasons to the ones I have just outlined in this discussion.  I'd prefer we don't confuse the two issues and will happily discuss the index tests in more details elsewhere (in this PR).

db/descriptions/descriptions.go
AndrewSisley: > Like context.Context allows you to create (or wrap) different types of contexts, but you handle a single one at a time.

Yes, but a context.Context is still just a (singular) context.Context.  The descriptions in this package do (will) not share an interface, or base inherited type.  The description types are all public and independent, 'descriptions' is just a made up umbrella term because I needed to give the package a name - there is no `Description` interface or type.

Another example is the `enumerable` package, this is singular, because it describes a single thing, although there are many implementations, and some extensions of the base interface. 

Note: I'm mostly just chatting for fun here, I dont really care about the name of the package, and think `description`(s) is a bit awkward anyway.

db/descriptions/schema.go
AndrewSisley: > Why would having schema_description.go next to collection_description.go be more cohesive than the above?

The below is the definition of 'cohesive' returned from duck-duck-go.  By definition if two files not next to each other, they are less cohesive than two files that are next to each other.
  
> cohesive:
Holding the particles of a homogeneous body together; ; producing cohesion
Cohering, or sticking together, as in a mass; capable of cohering; tending to cohere.

For example, if the files were called `description_schema.go` and `description_collection.go`, they would be co-located, and it would be much easier to spot the shared concepts.  A directory is just that really but using a `/` instead of an `_`, with some special treatment from file browser - this also emphasises the fact that they are separate conceptually from the rest of the `db` package contents.

When I look at the codebase, I'd like to see the stuff defining how to save and fetch schema next to the code used to save and fetch collections, and I'd also want the two neatly organised so that the implementation to fetch schema isn't mixed up with the code to save collections.

This lets me make sure that they all look and behave in predictable ways, and if I want to change all of them it is much more obvious what 'all of them' is, so that I don't accidentally miss some similar stuff (e.g. index descriptions).

db/indexed_docs_test.go
```diff
@@ -661,86 +661,6 @@ func TestNonUniqueDrop_ShouldDeleteStoredIndexedFields(t *testing.T) {
 	assert.Len(t, f.getPrefixFromDataStore(prodCatKey.ToString()), 1)
 }
 
-func TestNonUniqueDrop_IfDataStorageFails_ReturnError(t *testing.T) {
```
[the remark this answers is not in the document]
AndrewSisley: > I know we already touched on this one, but it really hard to see these tests are being cut off.

I really do appreciate that, and it is a big personal reason why I do not want the removal of these tests to be blended in with the key.go unit test discussion.

> help us achieve a greater code coverage which we are striving for

1. Code coverage is not an end goal, it is just an indicator.  I do not strive for code coverage, I strive for correct and maintainable code.  Good tests contribute to code correctness and lower maintenance costs, bad tests can have a negative effect on both.
2. Removing these tests does not appear to have affected code coverage.

> I wonder how do you decide if some tests need to be removed.

:) I'm a lot more conservative than I want to be RE preserving unit tests.  I have tried several times before to remove many of the unit tests that I have spend notable time changing in these last few PRs.

The index unit tests however have cost me a lot of time the last 10 days or so.  It took me roughly a day of development to get them working in a previous PR, I didn't bother trying in this one and removed the ones that started failing.

I do not consider these to be worth their maintenance cost.

They are the only tests to change in this PR.

This PR does not change any public behaviour.

This PR does not touch any index related code.

Yet these tests started to fail. These tests inhibit useful work, they prevent us from making useful changes to the codebase.  Changes that are irrelevant to what they are testing for.

Almost everything that the tests test, is covered by integration tests.  Some tests in these files do test errors which do not yet have integration test coverage, however I consider testing those errors to be very low value.

As well as the above, fixing these once they start failing is particularly painful.  There is quite a lot of mock setup, and it is not easy code to read.  Fixing them requires guessing how many internal calls to various mocks used by the tests are the 'correct' number, with the 'correct' input params and responses.  

Somewhat related to the conversation we had RE the index-explain integration tests, these unit tests make unimportant stuff really important, and this wastes a lot of time.

In my opinion, the cost of these tests far outweighs any benefit of having them.  Good tests save developer time (including external users' time in the case of bugs), that is why we write them - these tests cost more time than they will ever save.

I am not against the unit testing of stuff.  Unit tests are useful in many cases.

I am not against mocks, they have their place, however here, with the amount of irrelevant stuff that they need to mock, they make these tests bad tests that cost more than they save.

If we wish to preserve tests similar in concept to the tests in this file they should be reworked somehow.  They really should not fail when making the changes I have made in this PR.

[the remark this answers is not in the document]
AndrewSisley: > Yet these tests started to fail.

To clarify on this, I changed the commit order - I try quite hard to make to all of my commits pass the CI, it makes a lot of this easier.  So they started to fail part way through this PR, I removed them, them moved the commit to the start.

db/descriptions/schema.go
fredcarle: Alright I'll get behind your approach. I would like that you name it in the singular if it's not too big an ask as pointed out by islam.

I am worried that we will eventually hit a circular dependency situation given the proximity to the types in the root db package. But we can cross that bridge when we get there.

AndrewSisley: Okay cheers, I'll add a task/note against the package name convo to rename it

db/descriptions/descriptions.go
AndrewSisley: Fred also prefers `description` over `descriptions`, renaming :)

- [ ] Rename package

core/key.go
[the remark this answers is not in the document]
fredcarle: We have two overall objective with testing:
1- Testing behaviour that is meaningful to the end user (like you mentioned above)
2- Reaching maximum code coverage so that all possible code paths have been tested (sometimes that is currently impossible to do just with integration tests). This is a goal that the team set for itself (significantly increase code coverage).

I'll end my input on this conversation with reminding you of the above objectives. If you believe that simply relying on integration tests in case contributes towards achieving the above goals, then I think we're good.

db/indexed_docs_test.go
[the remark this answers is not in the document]
islamaliev: hey Andy, thanks for the detailed reasoning. I appreciate it.

I agree, these tests turned out to be a bit too invasive and have high maintenance cost with relatively little benefit.

fredcarle (approved):
We're doing good progress on detangling collection and schema. Keep it up! 🙂

core/key.go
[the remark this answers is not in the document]
AndrewSisley: >  If you believe that simply relying on integration tests in case contributes towards achieving the above objectives

I very strongly believe so.  Sometimes the integration tests will need to be more imaginative than our current core suite, for example chaos monkey like magic, or providing a deliberately faulty/test datastore implementation (the root datastore is something that can be publicly provided), but ultimately everything should be testable via public only access points.  If it isn't, it is dead code.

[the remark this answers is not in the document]
islamaliev: just another note: it is practically impossible to reach a desired test coverage just with integration tests.
Just imagine setting up "a deliberately faulty/test datastore" that you want to trigger for a specific request at a specific code location. You saw it was hard to achieve it with unit tests, doing it with integration tests will be a nightmare.

And those tests helped me write better code and fix bugs (not reported) in the existing code. Like this:
```go
err := someAction()
if err != nil {
    return err
}
...
res :=rsultsIter.NextSync()
if res.Error != nil {
    return err // here is the problem
}
```
Usually we don't write test that such paths. But we can easily overlook the details and ship incorrect code.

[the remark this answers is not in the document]
AndrewSisley: Such resilience tests don't need to be coded like traditional do-this, do-that tests. You can use this https://en.wikipedia.org/wiki/Chaos_engineering approach instead.  This can allow the tests to be somewhat proactive (searching for bugs) instead reactive (testing known pathways where bugs are expected).

Also a lot of our errors can be reached my manipulating the datastore values directly, no need to mess around with mocks.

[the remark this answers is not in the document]
AndrewSisley: I also want to introduce a really magic testing layer, where randomized queries and setup data are fed in.  Traditionally you can do this using recorded real-world data/actions, but there has been significant movement in using 'AI' to generate stuff like queries for such things.

PR #1980, opened 2023-10-18
refactor: Add peerstore to multistore

### LINKED ISSUE #1680: Make peerstore part of multistore
Currently, the peerstore is defined in the `net` package. We should look into making it a part of multistore.

fredcarle:
## Relevant issue(s)

Resolves #1680

## Description

This small refactor adds the `peerstore` to the `multistore`. In doing so, we broaden the types of stores that can be part of the `multistore` by starting from a `ds.Datastore` instead of the more restrictive `DSReaderWriter`

## Tasks

## How has this been tested?

make test showed no change in behaviour.

- MacOS

### DIFF
--- a/datastore/multi.go
+++ b/datastore/multi.go
@@ -22,6 +22,7 @@ var (
 	dataStoreKey   = rootStoreKey.ChildString("data")
 	headStoreKey   = rootStoreKey.ChildString("heads")
 	blockStoreKey  = rootStoreKey.ChildString("blocks")
+	peerStoreKey   = rootStoreKey.ChildString("ps")
 )
 
 type multistore struct {
@@ -44,7 +45,7 @@ func MultiStoreFrom(rootstore ds.Datastore) MultiStore {
 		data: prefix(rootRW, dataStoreKey),
 		head: prefix(rootRW, headStoreKey),
 		// the `peers` prefix is assigned by the libp2p peerstore
-		peer:   namespace.Wrap(rootstore, rootStoreKey),
+		peer:   namespace.Wrap(rootstore, peerStoreKey),
 		system: prefix(rootRW, systemStoreKey),
 		dag:    NewDAGStore(prefix(rootRW, blockStoreKey)),
 	}
--- a/datastore/multi.go
+++ b/datastore/multi.go
@@ -12,21 +12,23 @@ package datastore
 
 import (
 	ds "github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/namespace"
 )
 
 var (
 	// Individual Store Keys
-	rootStoreKey   = ds.NewKey("/db")
-	systemStoreKey = rootStoreKey.ChildString("/system")
-	dataStoreKey   = rootStoreKey.ChildString("/data")
-	headStoreKey   = rootStoreKey.ChildString("/heads")
-	blockStoreKey  = rootStoreKey.ChildString("/blocks")
+	rootStoreKey   = ds.NewKey("db")
+	systemStoreKey = rootStoreKey.ChildString("system")
+	dataStoreKey   = rootStoreKey.ChildString("data")
+	headStoreKey   = rootStoreKey.ChildString("heads")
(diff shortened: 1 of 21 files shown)


datastore/multi.go
```diff
@@ -35,14 +37,15 @@ type multistore struct {
...
 	ms := &multistore{
-		root:   rootstore,
-		data:   prefix(rootstore, dataStoreKey),
-		head:   prefix(rootstore, headStoreKey),
-		system: prefix(rootstore, systemStoreKey),
-		dag:    NewDAGStore(block),
+		root:   rootRW,
+		data:   prefix(rootRW, dataStoreKey),
+		head:   prefix(rootRW, headStoreKey),
+		peer:   namespace.Wrap(rootstore, rootStoreKey),
```
AndrewSisley, line 46: question: Why is the peerstore not prefixed like the rest of the stores?

datastore/multi.go
```diff
@@ -12,21 +12,23 @@ package datastore
...
 )
 
 var (
 	// Individual Store Keys
-	rootStoreKey   = ds.NewKey("/db")
-	systemStoreKey = rootStoreKey.ChildString("/system")
-	dataStoreKey   = rootStoreKey.ChildString("/data")
-	headStoreKey   = rootStoreKey.ChildString("/heads")
-	blockStoreKey  = rootStoreKey.ChildString("/blocks")
+	rootStoreKey   = ds.NewKey("db")
```
AndrewSisley, line 20: question: Is this a bug fix?  Were we adding `//` to our keys instead of `/` here?

datastore/multi.go
fredcarle: I pushed the documentation of this at the same time you commented.

AndrewSisley: :grin: I thought I had gone mad/blind when I refreshed to add another comment and saw it lol.

question: Why do we want to rely on the libp2p peerstore to assign the prefix?

datastore/multi.go
fredcarle: `ChildString` does `NewKey(k.string + "/" + s)` and the `NewKey` function adds a `/` prefix if none are present and cleans the `//` that would have been present. Removing the `/` from the passed param makes it more "correct" according to what the function is doing.

datastore/multi.go
fredcarle: Because it's hard coded in the libp2p pstoreds package. If we assign a prefix of `peers` here we end up with what we had before `/db/peers/peers/addr/...`

datastore/multi.go
AndrewSisley: Ah nice - thanks Fred :)

datastore/multi.go
fredcarle: That's a fair observation. What if we make it specific? `db/ps/peers/...`

AndrewSisley: Totally happy with that :) I know it seems unlikely to ever cause problems, but it feels a bit wrong to rely on just giving it the root.

datastore/txn.go
```diff
@@ -66,7 +66,7 @@ func NewTxnFrom(ctx context.Context, rootstore ds.TxnDatastore, id uint64, reado
 		if err != nil {
 			return nil, err
 		}
-		multistore := MultiStoreFrom(rootTxn)
+		multistore := MultiStoreFrom(ShimTxnStore{rootTxn})
```
AndrewSisley, line 69: question: Why have you made this change?  There should be no need for it root is `IterableTxnDatastore`

datastore/multi.go
fredcarle: ✅ done

datastore/txn.go
fredcarle: `IterableTxnDatastore` does not implement `Close` and `MultiStoreFrom` now requires a `ds.Datastore`.

AndrewSisley: Ah okay - thanks for clarifying :)

AndrewSisley (approved):
LGTM - thanks Fred :)

PR #1984, opened 2023-10-19
feat: Change GetCollectionBySchemaFoo funcs to return many

### LINKED ISSUE #1963: Detangle collection fetching features
There are several features that make little sense in their current once schema is separated from collection, these need to be sorted out:

- In commits queries, collection ID can be requested and returned, this is impossible atm, as the DAG only contains global elements.  It is currently fetched by using the schema version ID in the DAG, but that does not define which collection.  If we end up storing a global collection ID in the DAG (which we may not want to), we could do this, but at the moment those things do not exist.
- `Store.GetCollectionByVersionID` fetches a single collection from a schema version ID
  - It should either be removed, or needs to change so that multiple collections are returned, as many collections can share the same schema version
  - This is also used in P2P, a hack to be added in their to use only the first whilst waiting on collection global IDs
  - This is also used in commit queries to get the collection ID
- `Store.GetCollectionBySchemaID` has the same problem as `Store.GetCollectionByVersionID` and is used heavily in P2P as well as being public

AndrewSisley:
## Relevant issue(s)

Resolves #1963

## Description

Changes GetCollectionBySchemaFoo function signatures to return many Collections, instead of single values.

Conceptually these functions were broken and relied on the fact that we cannot yet link multiple collections to the same schema(version)(s).  This changes their signatures and the code that calls them to reflect what they should do long term.

The code that calls these functions now hosts the 'broken' code.  These will still need to change once we add global IDs.

The implementation of these functions will change later in https://github.com/sourcenetwork/defradb/issues/1964.

### DIFF
--- a/cli/collection.go
+++ b/cli/collection.go
@@ -76,8 +76,8 @@ func MakeCollectionCommand(cfg *config.Config) *cobra.Command {
 				// Multiple params may have been specified, and in some cases both are needed.
 				// For example if a schema version and a collection name have been provided,
 				// we need to ensure tha a collection at the requested version is returned.
-				// Likewise we need to ensure that if a name and schema id are provided, but
-				// there are none matching both, that nothing is returned.
+				// Likewise we need to ensure that if a collection name and schema id are provided,
+				// but there are none matching both, that nothing is returned.
 				fetchedCols := cols
 				cols = nil
 				for _, c := range fetchedCols {
--- a/cli/collection.go
+++ b/cli/collection.go
@@ -64,6 +64,11 @@ func MakeCollectionCommand(cfg *config.Config) *cobra.Command {
 			}
 
 			if name != "" {
+				// Multiple params may have been specified, and in some cases both are needed.
+				// For example if a schema version and a collection name have been provided,
+				// we need to ensure tha a collection at the requested version is returned.
+				// Likewise we need to ensure that if a name and schema id are provided, but
+				// there are none matching both, that nothing is returned.
 				fetchedCols := cols
 				cols = nil
 				for _, c := range fetchedCols {
--- a/cli/collection.go
+++ b/cli/collection.go
@@ -63,6 +63,15 @@ func MakeCollectionCommand(cfg *config.Config) *cobra.Command {
 				return err
 			}
 
+			if schemaID != "" && versionID != "" && len(cols) > 0 {
+				if cols[0].SchemaID() != schemaID {
+					// If the a versionID has been provided that does not pair up with the given schemaID
+					// we should error and let the user know they have provided impossible params.
+					// We only need to check the first item - they will all be the same.
+					return NewErrSchemaVersionNotOfSchema(schemaID, versionID)
+				}
+			}
+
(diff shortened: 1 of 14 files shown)


planner/commit.go
```diff
@@ -353,13 +358,20 @@ func (n *dagScanNode) dagBlockToNodeDoc(block blocks.Block) (core.Doc, []*ipld.L
...
 	if err != nil {
 		return core.Doc{}, nil, err
 	}
+	if len(cols) == 0 {
+		return core.Doc{}, nil, client.NewErrCollectionNotFoundForSchemaVersion(schemaVersionId)
+	}
 
+	// WARNING: This will become incorrect once we allow multiple collections to share the same schema
+	// todo: link to tickets.
+	// We should instead fetch the collection be global collection ID:
```
fredcarle, line 371: nitpick: `by global...`

AndrewSisley: I managed to leave half the todo text in there too..., thanks for spotting Fred :)

- [ ] Fix warning text

cli/collection.go
```diff
@@ -60,6 +62,24 @@ func MakeCollectionCommand(cfg *config.Config) *cobra.Command {
...
+			if name != "" {
+				fetchedCols := cols
+				cols = nil
+				for _, c := range fetchedCols {
+					if c.Name() == name {
+						cols = append(cols, c)
+						break
+					}
+				}
+			}
```
fredcarle, line 75: thought: This seems superfluous. If `name != ""`, there will be exactly one collection anyways.

thought (out-of-scope): We should eventually ensure that the flags that are part of the switch statement are mutually exclusive.

cli/collection.go
```diff
@@ -60,6 +62,29 @@ func MakeCollectionCommand(cfg *config.Config) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			if name != "" {
+				// Multiple params may have been specified, and in some cases both are needed.
+				// For example if a schema version and a collection name have been provided,
+				// we need to ensure tha a collection at the requested version is returned.
+				// Likewise we need to ensure that if a name and schema id are provided, but
+				// there are none matching both, that nothing is returned.
```
fredcarle, line 71: thought: I don't think this comment reflect what happens in the switch statement. Only one of the three will ever be used.

AndrewSisley: Yes, but that is due to how they interact.

If name and another param is provided, the other param will be used and name applied after.
If schema version is provided, schema does not matter.  We could check that the schema version fetched is actually of the given schema though in case someone provides impossible params.

fredcarle: `// we need to ensure **that** a collection`

fredcarle (approved):
LGTM!

cli/collection.go
AndrewSisley: > Ah yes ok I see what you mean

Am adding an error for that schema-schema version id clash

- [ ] add error

> we need to ensure **that** a collection

Cheers, will change :)

- [ ] collection name

Issue #1085, opened 2023-02-08
Add a globalID to the collection description


fredcarle:
If we want to support application specific collision resistant collection IDs, we could specify a globalCollectionID which could be self generated on a trusted network or generated using SourceHub on a trustless network. This globalCollectionID can be assigned at creation time or post creation and would simply be stored on the collection description without affecting the local collectionID. If defined, the globalCollectionID would be used on the P2P network.

```graphql
type User @options(globalID: "ABC123") {
  Name: String
  Handle: String
  ...
}
```

or

```bash
defradb client collection options User globalID=ABC123
```

AndrewSisley:
Note: As of https://github.com/sourcenetwork/defradb/pull/1984 there are 2 places in the codebase that reference this issue, one in P2P, and one in the commit planner node.

A `Ctrl+f` for `https://github.com/sourcenetwork/defradb/issues/1085` in the codebase should find these locations.

AndrewSisley:
What this ticket will involve now depends on what we agree upon in https://github.com/sourcenetwork/SIPs/pull/19

PR #1988, opened 2023-10-20
refactor: Reorganise collection description storage

### LINKED ISSUE #1964: Detangle collection storage

AndrewSisley:
## Relevant issue(s)

Resolves #1964 #1913 

## Description

Reorganises collection description storage so that it actually makes sense.

The existing key/value setup has been replaced with the following keys:
```
[Collection.ID] => json // Json should be stored against an immutable index, i.e. not against `Name`
[Collection.Name] => [Collection.ID]  // This means collection names must be unique
```
We also need the following for the short-term, before Collection.GlobalID gets added, as otherwise if a collection is updated to a new schema version, we lose track of what collection it could be.  The locations that use this have been linked to https://github.com/sourcenetwork/defradb/issues/1085
```
[Collection.SchemaVersionId]/[Collection.ID] => nil
```

With this change the storage setup should fully support multiple collections from a single schema, however this is blocked off by `setDefaultSchemaVersion` (it doesn't allow users to explicitly specify a collection, so atm there is no way to try and do this). Later when we introduce `patchCollection` we will need to be mindful of this and block off user's ability to do this (if done before #1085).

Strongly recommend reviewing commit by commit, I've tried to add a fair bit of context in each commit message.

### DIFF
--- a/core/key.go
+++ b/core/key.go
@@ -42,18 +42,18 @@ const (
 )
 
 const (
-	COLLECTION                        = "/collection/value"
-	COLLECTION_NAME                   = "/collection/name"
-	COLLECTION_SCHEMA_VERSION         = "/collection/version/v"
-	COLLECTION_SCHEMA_VERSION_HISTORY = "/collection/version/h"
-	COLLECTION_INDEX                  = "/collection/index"
-	SCHEMA_MIGRATION                  = "/schema/migration"
-	SCHEMA_VERSION                    = "/schema/version"
-	SEQ                               = "/seq"
-	PRIMARY_KEY                       = "/pk"
-	DATASTORE_DOC_VERSION_FIELD_ID    = "v"
-	REPLICATOR                        = "/replicator/id"
-	P2P_COLLECTION                    = "/p2p/collection"
+	COLLECTION                     = "/collection/value"
+	COLLECTION_NAME                = "/collection/name"
+	COLLECTION_SCHEMA_VERSION      = "/collection/version"
+	COLLECTION_INDEX               = "/collection/index"
+	SCHEMA_MIGRATION               = "/schema/migration"
+	SCHEMA_VERSION                 = "/schema/version/v"
+	SCHEMA_VERSION_HISTORY         = "/schema/version/h"
+	SEQ                            = "/seq"
+	PRIMARY_KEY                    = "/pk"
+	DATASTORE_DOC_VERSION_FIELD_ID = "v"
+	REPLICATOR                     = "/replicator/id"
+	P2P_COLLECTION                 = "/p2p/collection"
 )
 
 // Key is an interface that represents a key in the database.
@@ -356,7 +356,7 @@ func NewSchemaVersionMigrationKey(schemaVersionID string) SchemaVersionMigration
 }
 
 func NewSchemaHistoryKeyFromString(keyString string) (SchemaHistoryKey, error) {
-	keyString = strings.TrimPrefix(keyString, COLLECTION_SCHEMA_VERSION_HISTORY+"/")
+	keyString = strings.TrimPrefix(keyString, SCHEMA_VERSION_HISTORY+"/")
 	elements := strings.Split(keyString, "/")
 	if len(elements) != 2 {
 		return SchemaHistoryKey{}, ErrInvalidKey
@@ -673,7 +673,7 @@ func (k SchemaVersionKey) ToDS() ds.Key {
 }
 
 func (k SchemaHistoryKey) ToString() string {
-	result := COLLECTION_SCHEMA_VERSION_HISTORY
+	result := SCHEMA_VERSION_HISTORY
 
 	if k.SchemaID != "" {
 		result = result + "/" + k.SchemaID
--- a/core/key.go
+++ b/core/key.go
@@ -11,6 +11,7 @@
 package core
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -41,8 +42,8 @@ const (
 )
 
 const (
-	COLLECTION                        = "/collection/names"
-	COLLECTION_SCHEMA                 = "/collection/schema"
+	COLLECTION                        = "/collection/value"
+	COLLECTION_NAME                   = "/collection/name"
 	COLLECTION_SCHEMA_VERSION         = "/collection/version/v"
 	COLLECTION_SCHEMA_VERSION_HISTORY = "/collection/version/h"
 	COLLECTION_INDEX                  = "/collection/index"
@@ -99,26 +100,32 @@ type HeadStoreKey struct {
 
 var _ Key = (*HeadStoreKey)(nil)
 
-// CollectionKey points to the current/'head' SchemaVersionId for
-// the collection of the given name.
+// CollectionKey points to the json serialized description of the
+// the collection of the given ID.
 type CollectionKey struct {
-	CollectionName string
+	CollectionID uint32
 }
 
 var _ Key = (*CollectionKey)(nil)
 
-// CollectionSchemaKey points to the current/'head' SchemaVersionId for
-// the collection of the given schema id.
-type CollectionSchemaKey struct {
-	SchemaId string
+// CollectionNameKey points to the ID of the collection of the given
+// name.
+type CollectionNameKey struct {
+	Name string
 }
 
-var _ Key = (*CollectionSchemaKey)(nil)
+var _ Key = (*CollectionNameKey)(nil)
 
-// CollectionSchemaVersionKey points to schema of a collection at a given
-// version.
+// CollectionSchemaVersionKey points to nil, but the keys/prefix can be used
+// to get collections that are using, or have used a given schema version.
+//
+// If a collection is updated to a different schema version, the old entry(s)
+// of this key will be preserved.
+//
+// This key should be removed in https://github.com/sourcenetwork/defradb/issues/1085
 type CollectionSchemaVersionKey struct {
 	SchemaVersionId string
+	CollectionID    uint32
 }
 
 var _ Key = (*CollectionSchemaVersionKey)(nil)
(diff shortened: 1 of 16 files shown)


core/key.go
```diff
@@ -41,18 +42,18 @@ const (
...
-	COLLECTION_SCHEMA_VERSION_HISTORY = "/collection/version/h"
-	COLLECTION_INDEX                  = "/collection/index"
-	SCHEMA_MIGRATION                  = "/schema/migration"
-	SCHEMA_VERSION                    = "/schema/version"
-	SEQ                               = "/seq"
-	PRIMARY_KEY                       = "/pk"
-	DATASTORE_DOC_VERSION_FIELD_ID    = "v"
-	REPLICATOR                        = "/replicator/id"
-	P2P_COLLECTION                    = "/p2p/collection"
+	COLLECTION                     = "/collection/value"
```
fredcarle, line 45: suggestion: I think `COLLECTION_NUMBER = "/collection/number"` would be a bit more representative in this case. I had mentionned previously that we could have `collection.Index` instead of the current `collection.ID` so that we could use `collection.ID` for the  global ID. The problem is that it would then create confusion with the secondary indexes. This is why I'm suggesting to use `collection.Number`.

suggestion: I think having `COLLECTION_ID = "/collection/id"` would be important to reference the collection with it's global ID.

db/description/collection.go
```diff
@@ -0,0 +1,214 @@
...
+	idBuf, err := json.Marshal(desc.ID)
+	if err != nil {
+		return client.CollectionDescription{}, err
+	}
+
+	nameKey := core.NewCollectionNameKey(desc.Name)
+	err = txn.Systemstore().Put(ctx, nameKey.ToDS(), idBuf)
+	if err != nil {
+		return client.CollectionDescription{}, err
+	}
```
fredcarle, line 49: suggestion: I would do the opposite here. I would store the description on the name key instead of the collection number key because when querying, users will do so using the collection name. This will prevent the extra key lookup on queries which is where speed is most important.

core/key.go
AndrewSisley: > The problem is that it would then create confusion with the secondary indexes. This is why I'm suggesting to use collection.Number.

Did I misread, or did you edit? :sweat_smile: 

I missed this when replying, `number` feels a bit strange, but I have no better suggestions yet.

fredcarle: > thought: colllection.Index will also be a confusing name for that reason. Maybe we can find something else when we get round to looking at that.

`collection.Number`. would be fine as noted above.

fredcarle: > Then again, I think this key prefix should align with that property, which means it should either be left as collection/id for now, or we need to solve the collection.ID property name problem here/now. Worth bearing in mind is that changing this is a breaking change, so it is better if we don't change it every other day.

I really think `collection.ID` should be used for the global ID.

fredcarle: > Did I misread, or did you edit? 😅

I did not edit :)

db/description/collection.go
AndrewSisley: I do disagree with you here, for the following reasons:

- Name is mutable, and if the description is stored against it, renaming the collection becomes more expensive and error prone as we have to move the definition and re-index everything.
- The performance cost should be very very small, it is an extra single-point lookup.
- I find it conceptually to make more sense to store the thing against it's fixed ID, and index that, than to store it against a mutable property on it, and then have everything else reference the mutable property.

I think the maintainability of storing against the ID far outweigh the performance gains here.

core/key.go
AndrewSisley: Discussed in stand-up, we are undecided but we set a deadline for noon Tuesday Ottawa time to pick something.  `Number` is still a bit weird, but otherwise received little criticism (unlike `LocalID`, which could be confusing for a few reasons.  If no consensus is reached by the deadline I will probably rename to `Number`.

- [ ] Wait/discuss/rename to `Number

db/description/collection.go
AndrewSisley: Ah, I should have noted this one in the PR description I think.

shahzadlone (approved):
I enjoyed all the improvement nuggets, cheers! As for the decision on picking a name for the Sequence, Index, ID, Number situation will the final decision be done in a later PR?

AndrewSisley:
> I enjoyed all the improvement nuggets, cheers! As for the decision on picking a name for the Sequence, Index, ID, Number situation will the final decision be done in a later PR?

Nice :)

Yes, that name can be revisited/removed when GlobalID gets implemented.

core/key.go
AndrewSisley: Leaving as is for now, hopefully this can be removed in https://github.com/sourcenetwork/defradb/issues/1085

fredcarle: Can we do something about the actual key string and though? Value isn't representative of the key.

AndrewSisley: Ah, true sorry - I'll change it to `collection/id`

- [ ] `collection/id`

db/description/collection.go
```diff
@@ -0,0 +1,214 @@
...
+	"encoding/json"
+
+	"github.com/ipfs/go-datastore/query"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/core"
+	"github.com/sourcenetwork/defradb/datastore"
+)
+
+func SaveCollection(
```
islamaliev, line 24: suggestion: would be nice to have a short documentation for this and other methods

fredcarle: A was going to point that out too. All the public functions should have documentation.

AndrewSisley: >  All the public functions should have documentation.

These arent public, they are internal to defra

db/description/collection.go
```diff
@@ -0,0 +1,214 @@
...
+	ctx context.Context,
+	txn datastore.Txn,
+	desc client.CollectionDescription,
+) (client.CollectionDescription, error) {
+	buf, err := json.Marshal(desc)
```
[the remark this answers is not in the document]
fredcarle: A go standard is also just to use `b` in a case like this.

core/key.go
```diff
@@ -41,18 +42,18 @@ const (
...
+	COLLECTION_SCHEMA_VERSION      = "/collection/version"
+	COLLECTION_INDEX               = "/collection/index"
+	SCHEMA_MIGRATION               = "/schema/migration"
+	SCHEMA_VERSION                 = "/schema/version/v"
+	SCHEMA_VERSION_HISTORY         = "/schema/version/h"
+	SEQ                            = "/seq"
+	PRIMARY_KEY                    = "/pk"
+	DATASTORE_DOC_VERSION_FIELD_ID = "v"
+	REPLICATOR                     = "/replicator/id"
+	P2P_COLLECTION                 = "/p2p/collection"
```
fredcarle, line 56: question: You didn't want to add the following as well as mentioned on the issue?
```golang
SCHEMA_ROOT_COLLECTION            = "/schema/root/collection" // "{rootSchemaID}/{collectionName} -> someEmptyMarker
SCHEMA_VERSION_COLLECTION         = "/schema/version/collection" // "{schemaVersionID}/{collectionName} -> someEmptyMarker
```

db/description/collection.go
fredcarle: They are very much public at the moment though. They will be accessible to users.

core/key.go
AndrewSisley: No, that is out of scope, and I think it would be dead code.

db/description/collection.go
```diff
@@ -0,0 +1,214 @@
...
+			if err := schemaVersionQuery.Close(); err != nil {
+				return nil, NewErrFailedToCloseSchemaQuery(err)
+			}
+			return nil, err
+		}
+
+		colIDs = append(colIDs, colSchemaVersionKey.CollectionID)
+	}
+
+	cols := make([]client.CollectionDescription, 0)
```
islamaliev, line 129: todo: `make([]client.CollectionDescription, 0, colIDs)`

core/key.go
fredcarle: It would be a public utility so not dead code. But I agree that it can be out-of-scope.

db/description/collection.go
AndrewSisley: I do not consider these to be public, and neither does at least one other dev.  I also think that conversation is out of scope here too.

db/description/collection.go
islamaliev, line 1: praise: it's great that you moved all these highly cohesive functions into a separate package. It makes the rest of code much cleaner.

db/description/collection.go
AndrewSisley: * `make([]client.CollectionDescription, len(colIDs))

db/description/collection.go
fredcarle: I strongly disagree with this one. It's either public or it's not. The functions are accessible by users and will be found in the go docs. They should be documented. If we want to keep them internal, they can be under an `internal` package.

db/description/collection.go
islamaliev: right :)

db/description/collection.go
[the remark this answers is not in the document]
islamaliev: I would argue that short var names like this are only acceptable for very short functions (I'd say up to 6 lines of code) otherwise it doesn't do any good to readability.

islamaliev (approved):
great job, Andy!

db/description/collection.go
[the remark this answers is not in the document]
fredcarle: To me as long as you can see the whole function on a smallish screen. then it's fine. That probably means around 50 lines. But definitely more than 6. As long as there aren't many more byte arrays defined. That being said, it was only a option. Your suggestions are good as well.

db/schema.go
```diff
@@ -149,11 +148,19 @@ func (db *db) patchSchema(ctx context.Context, txn datastore.Txn, patchString st
...
+
+	newCollections, err := db.getAllCollections(ctx, txn)
+	if err != nil {
+		return err
+	}
 
-		newCollections = append(newCollections, col.Definition())
+	definitions := make([]client.CollectionDefinition, len(newCollections))
+	for i, col := range newCollections {
+		definitions[i] = col.Definition()
```
fredcarle, line 160: tought: It would be nice to have a method `db.getAllCollectionDefinitions` that would return the definitions directly instead of having to loop through collections to get what we actually want.

db/description/collection.go
```diff
@@ -0,0 +1,214 @@
...
+
+	nameKey := core.NewCollectionNameKey(desc.Name)
+	err = txn.Systemstore().Put(ctx, nameKey.ToDS(), idBuf)
+	if err != nil {
+		return client.CollectionDescription{}, err
+	}
+
+	// The need for this key is temporary, we should replace it with the global collection ID
+	// https://github.com/sourcenetwork/defradb/issues/1085
+	schemaVersionKey := core.NewCollectionSchemaVersionKey(desc.SchemaVersionID, desc.ID)
```
fredcarle, line 53: tought: In its current form, this should probably not be removed when adding global collection ID since it will enable some nice DB admin feature where one can check which collections are at a given schema version.

db/description/collection.go
AndrewSisley: Discussed over discord, the `internal` package conversation will be re-opened outside of this PR.  I'll add some docs.

- [ ] Add docs for funcs

db/description/collection.go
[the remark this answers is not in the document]
AndrewSisley: We use `buf` all over the place in our codebase, and I'm pretty sure I've seen it for stuff like this in a fair few other codebases.  Islam, internally within `json.Marshal` the return value probably does fit your `some kind of preallocated memory` description and that's how I've seen this naming come from.

Let me know if either of you think this is important, as I don't, and without clear consensus between you guys I'll leave it as is otherwise.

db/schema.go
AndrewSisley: ah, I think that would only be used here no? And if it wasn't it may end up coupling totally unrelated parts of the codebase together - is not like this loop is particularly complex and in need of standardisation.

Leaving as is unless significant push back received.

db/description/collection.go
AndrewSisley: The replacement would handle that, there would be no loss of user/admin functionality.

db/description/collection.go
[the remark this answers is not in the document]
fredcarle: `buf` is misrepresenting what this is but we do have a few places where we use it probably by mistake including one that Islam wrote
Screenshot 2023-10-24 at 2 42 21 PM
but most were added by Andy in  #1965. `buf` is usually used for `bytes.Buffer` as we do a lot in the code base. 

My opinion is that it's the wrong name for the variable but I wouldn't say its important to me in this case. It was good to point it out though.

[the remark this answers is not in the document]
AndrewSisley: > but most were added by Andy in https://github.com/sourcenetwork/defradb/pull/1965.

nitpick: If you look at `db/collection.go` in that PR, you should see that these variables were `fooBuf` before :)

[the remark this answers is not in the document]
fredcarle: Fair enough 🙂

PR #2005, opened 2023-10-24
feat: Rename Schema.SchemaID to Schema.Root

### LINKED ISSUE #2004: Rename Schema.SchemaID to Schema.Root

AndrewSisley:
## Relevant issue(s)

Resolves #2004

## Description

Renames Schema.SchemaID to Schema.Root.

A `Schema` doesn't really exist outside of its version, and this name change will hopefully help clarify that.

Later we plan on renaming the struct (maybe to `SchemaVersion`), and `SchemaVersionID` to just `ID`, but not in this PR (different priorities, amongst other things).


core/crdt/composite.go
```diff
@@ -119,12 +119,12 @@ func NewCompositeDAG(
...
 }
 
-// GetSchemaID returns the schema ID of the composite DAG CRDT.
+// ID returns the schema ID of the composite DAG CRDT.
 func (c CompositeDAG) ID() string {
 	return c.key.ToString()
 }
 
-// GetSchemaID returns the schema ID of the composite DAG CRDT.
+// Value returns the schema ID of the composite DAG CRDT.
```
shahzadlone, line 127: question: Same comments for `ID()` and `Value()`?

shahzadlone (approved):
LGTM

core/crdt/composite.go
AndrewSisley: The comments, and the `Value` function implementation make no real sense to me anyway, but I felt the need to at least tie the comments to the functions upon which they sit. I see further changes here as out of scope.

PR #2006, opened 2023-10-26
feat: Add means to fetch schema

### LINKED ISSUE #1993: No way to fetch defined schema versions
There is actually no way to fetch the schema versions not actively in use in Defra.  If a collection is not using a schema version, it is impossible to view it.

We should add some commands to allow users to get them, by schema ID, version ID, name, and all.

AndrewSisley:
## Relevant issue(s)

Resolves #1993

## Description

Adds means to fetch schema.

Adds a bunch of funcs to the clients to allow fetching of schema(versions) defined in the system.  When the set-default stuff got merged, users were no longer guaranteed (at least on init) to be able to view the schema ids they have created, this change allows them to view them when ever they like.

## How has this been tested?

Todo:
- [x] Manually test the playground with open api stuff

### DIFF
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -48,7 +48,7 @@ Example: view schema by version id
 				if err != nil {
 					return err
 				}
-				schemas = []client.SchemaDescription{schema}
+				return writeJSON(cmd, schema)
 
 			case root != "":
 				s, err := store.GetSchemaByRoot(cmd.Context(), root)
@@ -70,10 +70,6 @@ Example: view schema by version id
 					return err
 				}
 				schemas = s
-			}
-
-			if len(schemas) == 1 {
-				return writeJSON(cmd, schemas[0])
 			}
 
 			return writeJSON(cmd, schemas)
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -23,19 +23,19 @@ func MakeSchemaDescribeCommand() *cobra.Command {
 
 	var cmd = &cobra.Command{
 		Use:   "describe",
-		Short: "View schema description.",
+		Short: "View schema descriptions.",
 		Long: `Introspect schema types.
 
-Example: view all schema
+Example: view all schemas
   defradb client schema describe
 		
-Example: view schema by name
+Example: view schemas by name
   defradb client schema describe --name User
 		
-Example: view schema by root
+Example: view schemas by root
   defradb client schema describe --root bae123
 		
-Example: view schema by version id
+Example: view a single schema by version id
   defradb client schema describe --version bae123
 		`,
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -51,21 +51,21 @@ Example: view schema by version id
 				return writeJSON(cmd, schema)
 
 			case root != "":
-				s, err := store.GetSchemaByRoot(cmd.Context(), root)
+				s, err := store.GetSchemasByRoot(cmd.Context(), root)
 				if err != nil {
 					return err
 				}
 				schemas = s
 
 			case name != "":
-				s, err := store.GetSchemaByName(cmd.Context(), name)
+				s, err := store.GetSchemasByName(cmd.Context(), name)
 				if err != nil {
 					return err
 				}
 				schemas = s
 
 			default:
-				s, err := store.GetAllSchema(cmd.Context())
+				s, err := store.GetAllSchemas(cmd.Context())
 				if err != nil {
 					return err
 				}
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -1,0 +1,74 @@
+// Copyright 2023 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+func MakeSchemaDescribeCommand() *cobra.Command {
+	var name string
+	var root string
+	var versionID string
+
+	var cmd = &cobra.Command{
+		Use:   "describe",
+		Short: "View schema description.",
+		Long: `Introspect schema types.
+
+Example: view all schema
+  defradb client schema describe
+		
+Example: view schema by name
+  defradb client schema describe --name User
+		
+Example: view schema by root
+  defradb client schema describe --root bae123
+		
+Example: view schema by version id
+  defradb client schema describe --version bae123
+		`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			store := mustGetStoreContext(cmd)
+
+			var schemas []client.SchemaDescription
+			switch {
+			case versionID != "":
+				schema, err := store.GetSchemaByVersionID(cmd.Context(), versionID)
+				if err != nil {
+					return err
+				}
+				schemas = []client.SchemaDescription{schema}
+
+			case root != "":
+				panic("todo")
+
+			case name != "":
+				panic("todo")
+
+			default:
+				panic("todo")
+			}
+
+			if len(schemas) == 1 {
+				return writeJSON(cmd, schemas[0])
+			}
+
+			return writeJSON(cmd, schemas)
+		},
+	}
+	cmd.PersistentFlags().StringVar(&name, "name", "", "Schema name")
+	cmd.PersistentFlags().StringVar(&root, "root", "", "Schema root")
+	cmd.PersistentFlags().StringVar(&versionID, "version", "", "Schema Version ID")
+	return cmd
+}
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -57,7 +57,11 @@ Example: view schema by version id
 				panic("todo")
 
 			default:
-				panic("todo")
+				s, err := store.GetAllSchema(cmd.Context())
+				if err != nil {
+					return err
+				}
+				schemas = s
 			}
 
 			if len(schemas) == 1 {
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -58,7 +58,11 @@ Example: view schema by version id
 				schemas = s
 
 			case name != "":
-				panic("todo")
+				s, err := store.GetSchemaByName(cmd.Context(), name)
+				if err != nil {
+					return err
+				}
+				schemas = s
 
 			default:
 				s, err := store.GetAllSchema(cmd.Context())
--- a/cli/schema_describe.go
+++ b/cli/schema_describe.go
@@ -51,7 +51,11 @@ Example: view schema by version id
 				schemas = []client.SchemaDescription{schema}
 
 			case root != "":
-				panic("todo")
+				s, err := store.GetSchemaByRoot(cmd.Context(), root)
+				if err != nil {
+					return err
+				}
+				schemas = s
 
 			case name != "":
 				panic("todo")
--- a/client/db.go
+++ b/client/db.go
@@ -167,6 +167,9 @@ type Store interface {
 	// Will return an error if it is not found.
 	GetSchemaByVersionID(context.Context, string) (SchemaDescription, error)
 
+	// GetSchemaByRoot returns the all schema versions for the given root.
+	GetSchemaByRoot(context.Context, string) ([]SchemaDescription, error)
+
 	// GetAllSchema returns all schema versions that currently exist within
 	// this [Store].
 	GetAllSchema(context.Context) ([]SchemaDescription, error)
--- a/client/db.go
+++ b/client/db.go
@@ -161,6 +161,9 @@ type Store interface {
 	// this [Store].
 	GetAllCollections(context.Context) ([]Collection, error)
 
+	// GetSchemaByName returns the all schema versions with the given name.
+	GetSchemaByName(context.Context, string) ([]SchemaDescription, error)
+
 	// GetSchemaByVersionID returns the schema description for the schema version of the
 	// ID provided.
 	//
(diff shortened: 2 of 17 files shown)


cli/schema_describe.go
```diff
@@ -0,0 +1,86 @@
...
+
+			default:
+				s, err := store.GetAllSchema(cmd.Context())
+				if err != nil {
+					return err
+				}
+				schemas = s
+			}
+
+			if len(schemas) == 1 {
```
nasdf, line 75: I think this will cause an issue since the output is expected to be an array in some of the cases.

client/db.go
```diff
@@ -161,6 +161,22 @@ type Store interface {
...
+	//
+	// Will return an error if it is not found.
+	GetSchemaByVersionID(context.Context, string) (SchemaDescription, error)
+
+	// GetSchemaByRoot returns the all schema versions for the given root.
+	GetSchemaByRoot(context.Context, string) ([]SchemaDescription, error)
+
+	// GetAllSchema returns all schema versions that currently exist within
+	// this [Store].
+	GetAllSchema(context.Context) ([]SchemaDescription, error)
```
nasdf, line 178: These are great additions to the API! It really helps clarify the workings of schemas and collections.

http/handler_store.go
```diff
@@ -411,6 +449,39 @@ func (h *storeHandler) bindRoutes(router *Router) {
...
+
+	schemaDescribe := openapi3.NewOperation()
+	schemaDescribe.OperationID = "schema_describe"
+	schemaDescribe.Description = "Introspect schema(s) by name, schema root, or version id."
+	schemaDescribe.Tags = []string{"schema"}
+	schemaDescribe.AddParameter(schemaNameQueryParam)
+	schemaDescribe.AddParameter(schemaSchemaRootQueryParam)
+	schemaDescribe.AddParameter(schemaVersionIDQueryParam)
+	schemaDescribe.AddResponse(200, schemaResponse)
+	schemaDescribe.Responses["400"] = errorResponse
```
nasdf, line 483: Thanks for keeping these in sync!

nasdf (changes requested):
Looks good! One minor fix in the CLI implementation.

cli/schema_describe.go
```diff
@@ -0,0 +1,86 @@
...
+	var name string
+	var root string
+	var versionID string
+
+	var cmd = &cobra.Command{
+		Use:   "describe",
+		Short: "View schema description.",
+		Long: `Introspect schema types.
+
+Example: view all schema
```
islamaliev, line 29: typo: "schemas"

client/db.go
islamaliev: hey @nasdf I just wanted to point out that we try to stick to the conventional commenting style https://conventionalcomments.org/

islamaliev, line 178: suggestion: rename to `GetAllSchemas` or `GetAllSchemaVersions`

db/description/schema.go
```diff
@@ -161,6 +203,47 @@ func GetSchemas(
 	return descriptions, nil
 }
 
+// GetSchemas returns all schema versions in the system.
+func GetAllSchemas(
+	ctx context.Context,
+	txn datastore.Txn,
+) ([]client.SchemaDescription, error) {
+	prefix := core.NewSchemaVersionKey("")
+	q, err := txn.Systemstore().Query(ctx, query.Query{
```
islamaliev, line 212: suggestion: I think you should use `datastore.DeserializePrefix[client.SchemaDescription]` for this

tests/clients/cli/wrapper.go
```diff
@@ -281,6 +281,65 @@ func (w *Wrapper) GetAllCollections(ctx context.Context) ([]client.Collection, e
...
+	}
+	var schema []client.SchemaDescription
+	if err := json.Unmarshal(data, &schema); err != nil {
+		return nil, err
+	}
+	return schema, err
+}
+
+func (w *Wrapper) GetAllSchema(ctx context.Context) ([]client.SchemaDescription, error) {
+	args := []string{"client", "schema", "describe"}
```
islamaliev, line 330: suggestion: 3 of these methods have almost identical bodies. They differ only by `args`. 

It will look cleaner if you extract the common code and pass only `args` there

islamaliev (commented):
Overall looks good. I just have few suggestions

client/db.go
fredcarle: +1

cli/schema_describe.go
```diff
@@ -0,0 +1,86 @@
...
+				schemas = []client.SchemaDescription{schema}
+
+			case root != "":
+				s, err := store.GetSchemaByRoot(cmd.Context(), root)
+				if err != nil {
+					return err
+				}
+				schemas = s
+
+			case name != "":
```
fredcarle, line 60: suggestion: It would be good to add a check after the switch to check if name and versionID were given that they are actually related like you did with collections.

cli/schema_describe.go
AndrewSisley: Ah okay, I'll change it.  Would you prefer an array all the time, even when `version` is provided, or should that still return a single item?

cli/schema_describe.go
AndrewSisley: Not a typo, but, after googling, I am incorrect anyway and this needs changing (possibly in a few places) - I actually thought `schema` was both plural and singular, like `sheep` and `sheep` :sweat_smile: 

- [ ] Schema is not plural

client/db.go
AndrewSisley: Cheers, will change.

- [ ] Rename to `GetAllSchemas`

db/description/schema.go
AndrewSisley: I only recently noticed that in one of your PRs and then forgot about it, will have a look, but there is a chance I'll want to leave it as is, as I might feel that calling that function hides too much (as the purpose of this file is to show all that stuff in one place).

- [ ] Checkout datastore.DeserializePrefix

tests/clients/cli/wrapper.go
AndrewSisley: I'm not sure that will be cleaner, as these are very simple functions that have no reason to be coupled via a private function, atm it is quite easy to see what they do, and a private func may damage that.

- [ ] Consider private func

http/handler_store.go
AndrewSisley: Thanks for making it simple enough to do so :)

cli/schema_describe.go
AndrewSisley: Yeah I was in two minds if this should behave like the collection stuff, this *is* different though, as they are not going to be acting on the collection (saving docs and the like). I'll revisit today and see what happens.

- [ ] Revisit multi param schema describe stuff

cli/schema_describe.go
```diff
@@ -0,0 +1,86 @@
...
+			switch {
+			case versionID != "":
+				schema, err := store.GetSchemaByVersionID(cmd.Context(), versionID)
+				if err != nil {
+					return err
+				}
+				schemas = []client.SchemaDescription{schema}
+
+			case root != "":
+				s, err := store.GetSchemaByRoot(cmd.Context(), root)
```
fredcarle, line 54: question: Why not assign directly to `schemas` when a slice is returned?

AndrewSisley: Compilation rules as `err` is not yet defined, the alternative would be to declare `err` outside the switch.

fredcarle: `err` outside the switch would be quite fine.

cli/schema_describe.go
fredcarle: Even if not acting on it, it would be weird to return a set of results event if the provided params make no sense.

db/description/schema.go
fredcarle: It would be just one click away from visibility though so I'm not sure the duplication would be worth it.

cli/schema_describe.go
AndrewSisley: :laughing: Both are quite fine and imperfect IMO.  Declaring `err` in the broader scope has the drawback of needlessly pulling a local variable with a short lifetime into a broader scope with a longer lifetime.

Are you asking me to change this?

http/handler_store.go
```diff
@@ -153,6 +153,41 @@ func (s *storeHandler) GetCollection(rw http.ResponseWriter, req *http.Request)
 	}
 }
 
+func (s *storeHandler) GetSchema(rw http.ResponseWriter, req *http.Request) {
+	store := req.Context().Value(storeContextKey).(client.Store)
+
+	switch {
```
fredcarle, line 159: suggestion: This switch should follow the same order as the one in the cli. Although we are constructing the request in the client to match exactly one of these cases, there is nothing preventing someone from manually writing a request with multiple parameters.

cli/schema_describe.go
nasdf: I think following the API would be best. So returning a single one for `version` and array for others.

db/description/schema.go
AndrewSisley: It is not just about that, but extracting this function to another shared function creates an oddity in this file, which itself hinders readability in some cases.

Worse though is that it inhibits change, if this func needs to change slightly (for whatever reason), it is currently very easy to do so - you just change it, however if it is calling `datastore.DeserializePrefix`, then either you have to risk serious damage by changing that function, or overcome a physcological barrier and un-refactor it and then make the change - that physcological barrier can be non-trivial in some cases and is frequent cause of large bloated shared functions.

http/handler_store.go
AndrewSisley: The client interface does not support that, so this does not support that.  Which means the CLI should not support that.  Which means I was wrong to ask Keenan to make the collection describe CLI stuff to support that.

cli/schema_describe.go
AndrewSisley: - [ ] Change array logic

cli/schema_describe.go
fredcarle: No. The assignment makes no difference in the number of allocation since a slice is passed by reference. I just wanted to know what was your reasoning behind it. Now I know :)

AndrewSisley: > The assignment makes no difference in the number of allocation since a slice is passed by reference

I wasn't talking in terms of performance, a local, short lived variable has a lower readability cost than a broadly scoped long lived one.  Especially in cases when there is no real reason for the increased scope/lifetime.

fredcarle: I know :)

But if there was a significant cost to the allocation, optimizing it would win over narrowing a variable scope. Imagine thousands of schema descriptions being allocated twice. I just mentioned it here to say that it wasn't something to consider and that your reasoning was good 👍

http/handler_store.go
fredcarle: This means that in both cases (CLI and HTTP for both schema and collection) we should make the params mutually exclusive.

cli/schema_describe.go
AndrewSisley: Ah got it :) Thanks for the explanation (and discussion) :)

http/handler_store.go
AndrewSisley: Yeah I think so :(

AndrewSisley: Although, I'd see changing the collection stuff as out of scope here, and with the code freeze coming up a rework could me that users end up getting nothing for 8 more weeks.

I think I'm in favour of leaving both Schema and Collection funcs as is in this PR, and then changing them both together in another ticket.

cli/schema_describe.go
AndrewSisley: I'm going to close this thread as it should match whatever we decide in https://github.com/sourcenetwork/defradb/pull/2006#discussion_r1374836866

http/handler_store.go
fredcarle: We could do that. It would simplify the interfaces but add a bit of complexity to the entry point. Not a big deal. I assume passing `nil` would return all schemas?

AndrewSisley: > I assume passing nil would return all schemas?

If we do this I would want to use `immutable.Option` for everything, types would be:

```
struct Filter {
  Name Option[String]
  Root Option[String]
  ID Option[String]
}
interface {
  GetSchemas(Option[Filter])
}
```

fredcarle: I think `Option[Filter]` is overkill and would be annoying more than anything. I would say it's overkill for all of this but I mind it less for the fields than for the function parameter.

AndrewSisley: > I think `Option[Filter]` is overkill and would be annoying more than anything. I would say it's overkill for all of this but I mind it less for the fields than for the function parameter.

Could have:
```
interface {
  GetSchemas(...Filter)
}
```
Maybe, but assuming you were happy enough with doing this in another ticket it is not important to sort it out here.

fredcarle: Totally ok with doing this in a different ticket. In the mean time do you mind just ordering the switch statement the same as the CLI so that they behave the same way?

AndrewSisley: > ordering the switch statement the same as the CLI

Yes of course, sorry I forgot about that.

- [ ] Change switch statement
- [ ] Open ticket for Filter stuff

db/description/schema.go
AndrewSisley: I had a look, and I'm still more comfy with the independent implementations.  Will be leaving as is.

tests/clients/cli/wrapper.go
AndrewSisley: I had a look, and I prefer the current setup.  Leaving as is.

http/handler_store.go
AndrewSisley: https://github.com/sourcenetwork/defradb/issues/2008

fredcarle (approved):
LGTM!

nasdf (approved):
Looks all good to me!

islamaliev (approved):
LGTM

PR #2016, opened 2023-10-30
refactor: CRDT merge direction

### LINKED ISSUE #1066: Incorrect CRDT Merge direction
At the moment we merge CRDT delta state starting from the `HEAD` and working backwards through the DAG, "merging" along the way. This is technically wrong, as it merges state backward to how its created.

However, the database still works because the primary CRDT type we use is the `LWWRegister`. Registers technically only care about the "last" value that is written, so if I set "name" to "Alice" *then* to "Bob", in reality, the in between state of "Alice" doesn't really matter that much.

However, the discussions related to #1064 its clear that this will breakdown for other CRDT types. Specifically this will already break down  for `Composite` CRDTs, where the current operations are only `CREATE`, and `UPDATE` effectively. But we want to introduce a new operation `DELETE`, which is order dependent in which direction we "merge". 

This will be necessary for #101 as well, since they'll likely be order dependant types.

fredcarle:
## Relevant issue(s)

Resolves #1066 

## Description

This PR change the CRDT merge direction. Originally, when an update arrived over the P2P network, each block was added to the DAG store and merged into the datastore in the order that they were received (newest to oldest block).

With this update, we start my receiving all blocks up to the version that we currently have (or until we reach the root) and then we apply the merge from oldest to newest block. This will enable more CRDT types to be added.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,11 @@ test\:coverage:
 	@$(MAKE) deps:lens
 	@$(MAKE) clean:coverage
 	mkdir $(COVERAGE_DIRECTORY)
+ifeq ($(path),)
 	gotestsum --format testname -- ./... $(TEST_FLAGS) $(COVERAGE_FLAGS)
+else
+	gotestsum --format testname -- $(path) $(TEST_FLAGS) $(COVERAGE_FLAGS)
+endif
 	go tool covdata textfmt -i=$(COVERAGE_DIRECTORY) -o $(COVERAGE_FILE)
 
 .PHONY: test\:coverage-func
@@ -266,8 +270,9 @@ test\:coverage-func:
 
 .PHONY: test\:coverage-html
 test\:coverage-html:
-	@$(MAKE) test:coverage
+	@$(MAKE) test:coverage path=$(path)
 	go tool cover -html=$(COVERAGE_FILE)
+	rm -rf ./coverage
 
 .PHONY: test\:changes
 test\:changes:
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,8 @@ test\:coverage-func:
 test\:coverage-html:
 	@$(MAKE) test:coverage path=$(path)
 	go tool cover -html=$(COVERAGE_FILE)
-	rm -rf ./coverage
+	@$(MAKE) clean:coverage
+	
 
 .PHONY: test\:changes
 test\:changes:
--- a/net/dag.go
+++ b/net/dag.go
@@ -20,8 +20,6 @@ import (
 	"github.com/ipfs/go-cid"
 	ipld "github.com/ipfs/go-ipld-format"
 
-	"github.com/sourcenetwork/defradb/core"
-	"github.com/sourcenetwork/defradb/datastore"
 	"github.com/sourcenetwork/defradb/logging"
 )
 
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -266,8 +270,9 @@ test\:coverage-func:
 
 .PHONY: test\:coverage-html
 test\:coverage-html:
-	@$(MAKE) test:coverage
+	@$(MAKE) test:coverage path=$(path)
 	go tool cover -html=$(COVERAGE_FILE)
+	rm -rf ./coverage
```
shahzadlone, line 275: todo: `make clean:coverage` instead?

net/process.go
```diff
@@ -29,57 +30,96 @@ import (
...
-		return nil, err
+	if err := txn.DAGstore().Put(ctx, nd); err != nil {
+		return err
 	}
 
-	delta, err := crdt.DeltaDecode(nd)
-	if err != nil {
-		return nil, errors.Wrap("failed to decode delta object", err)
+	if isComposite {
+		bp.composites.PushFront(nd)
```
AndrewSisley, line 67: todo:  This feels quite strange, as you are appending to a private property on an object that never reads from it (the only reading of it is done from `server.go`).  Given that this is the only property on this object it makes me wonder why `blockProcessor` exists as a type at all.

Also strange is that the block in `server.go` that does read from this, only uses stuff (`peer`) that is already on this object.

I suggest either removing `blockProcessor` as a type entirely (and just pass the list around), or move the below block from `server.go` into this type (in a `(bp *blockProcessor) ProcessBlocks()` func or similar):
```
for e := bp.composites.Front(); e != nil; e = e.Next() {
			nd := e.Value.(format.Node)
			err := s.peer.processBlock(ctx, txn, col, dsKey, nd, "")
			if err != nil {
				log.ErrorE(
					ctx,
					"Failed to process block",
					err,
					logging.NewKV("DocKey", dsKey.DocKey),
					logging.NewKV("CID", nd.Cid()),
				)
			}
		}
```

net/dag.go
```diff
@@ -125,44 +123,28 @@ func (p *Peer) dagWorker(jobs chan *dagJob) {
...
-				j.dsKey,
-				j.fieldName,
-				j.node,
-				children,
-				j.nodeGetter,
-			)
+			if j.nodeGetter != nil && j.cid.Defined() {
+				cNode, err := j.nodeGetter.Get(p.ctx, j.cid)
+				if err != nil {
+					log.ErrorE(p.ctx, "Failed to get node", err, logging.NewKV("CID", j.cid))
```
AndrewSisley, line 130: todo: Please define these errors (and others in other files) in an errors.go file

AndrewSisley (approved):
LGTM, just a couple of localised todos :)

shahzadlone (approved):
LGTM

Makefile
fredcarle: I do want the html output to have a visual.

net/dag.go
fredcarle: They are not returned errors. It's a log message. We don't predefine our log messages at all currently.

AndrewSisley: Our log messages are public, just like errors.  Most of the time the error messages will be exposed via logs (e.g. CLI/HTTP), the reasons for housing them in a dedicated and standardised file are exactly the same.

fredcarle: Although there is no value in defining them in the errors.go file. It's not an error that can be handled or compared. It's just a message in a log file.

shahzadlone: Time for a log.go file :,)

fredcarle: I'm ok with a solution like that. I'll leave it as is for now and create a separate PR to take care of log messages.

net/process.go
fredcarle: I move the block to process.go and simplified things a little more. `blockProcessor` had more fiels initially but I was able make the change leaner and it got kept around. I added some relevant fields to help pass things around. I like having the `blockProcessor` type as it keeps the processing under a specific domaine.

net/dag.go
AndrewSisley: > It's just a message in a log file.

nitpick: It is an error message in a log file, and indistinguishable to our users from any other 'kind' of error.

> I'll leave it as is for now and create a separate PR to take care of log messages

I'm fine with that, but I do think they should live in an `errors.go` file, because they are errors.

PR #2030, opened 2023-11-07
perf: Switch LensVM to wasmtime runtime

### LINKED ISSUE #2029: Switch Defra Lens to wasmtime runtime

AndrewSisley:
## Relevant issue(s)

Resolves #2029

## Description

Switches LensVM to the wasmtime runtime.

Should be a fair bit faster than wazero, and supports all the important build targets.  Also doesn't suffer from https://github.com/tetratelabs/wazero/issues/1818

Note on [wasmtime-go](https://github.com/bytecodealliance/wasmtime-go):
> This Go library uses CGO to consume the C API of the [Wasmtime project](https://github.com/bytecodealliance/wasmtime) which is written in Rust. Precompiled binaries of Wasmtime are checked into this repository on tagged releases so you won't have to install Wasmtime locally, but it means that this project only works on Linux x86_64, macOS x86_64 , and Windows x86_64 currently. Building on other platforms will need to arrange to build Wasmtime and use CGO_* env vars to compile correctly.

This is different to wazero, which as a pure Go project supports pretty much anything Go does without messing about.

Note: I had to change a test module, the new Lens version has closed a memory leak upon which the defra test module was relying on.

## Todo
- [x] Have a chat about cross compilation issues

## How has this been tested?

- [x] Debian Linux
- [x] Mac arm64
- [x] Windows


fredcarle (approved):
LGTM

shahzadlone (approved):
Comment from discord after passing `make test:lens` on windows:

Perhaps can add in README:
- Enable CGO
- Need GCC (for windows need mingw, i.e. `choco install mingw`)

jsimnz:
@shahzadlone 

CGO is needed regardless with this wasmtime integration. Is there something more specific you are referring to with the windows build?

shahzadlone:
> @shahzadlone 
> 
> CGO is needed regardless with this wasmtime integration. Is there something more specific you are referring to with the windows build?

Yea so normally on linux I have never had to worry about enabling cgo manually. 

However on windows, by default cgo is disabled. So it needs to be enabled and Inaddition the mingw dependency is needed for gcc tool chain.

Would be nice to document this in build requirements in README explicitly for windows.

fredcarle:
The need for cgo really sucks.Too bad Wasmer isn't more actively maintained.

jsimnz:
> The need for cgo really sucks. Too bad Wasmer isn't more actively maintained.

CGO applied to wasmer as well. Unless you mean wazero. But yes, not a fan of it in general, as it does some wonky things to the go runtime.

It shouldn't be hard to get wazero back to a place we can use it with the enumerator work. Then we can add a option as to which build users want. Wasmtime and speed, but with cgo vs wazero and slower, but no cgo

fredcarle:
I think I mixed the two yes 🙃

islamaliev (commented):
LGTM

PR #2033, opened 2023-11-10
ci: Add windows test runner

### LINKED ISSUE #2032: Add windows runner
These only need to run the simple set of tests. See also https://github.com/sourcenetwork/defradb/issues/2034

Tests fail on windows, so something will need to change.

AndrewSisley:
## Relevant issue(s)

Resolves #2032 

## Description

Adds a windows test run to our test matrix.

Strongly suggest reviewing commit by commit - I've tried to explain why a bunch of stuff I had to change has been changed in the commit bodies.

## Notes

- The windows run is consistently slower than the mac/ubuntu ones, worsening our ci wait times
- The below failure appears to happen occasionally:
> error generating coverage report: writing D:\a\defradb\defradb\coverage\covmeta.776b203c66aeebb53e5712fcea6cd9a0: rename from D:\a\defradb\defradb\coverage\tmp.covmeta.776b203c66aeebb53e5712fcea6cd9a01700087389181284400 failed: rename D:\a\defradb\defradb\coverage\tmp.covmeta.776b203c66aeebb53e5712fcea6cd9a01700087389181284400 D:\a\defradb\defradb\coverage\covmeta.776b203c66aeebb53e5712fcea6cd9a0: Access is denied.

I can't spot anything obvious searching for this, and I've only seen it maybe twice/thrice out 30-40 runs, so at the moment I'm in favour of merging as-is, and then we can disable the cod-cov for windows runs whilst we fix/shelve it - but do let me know.

### DIFF
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -28,6 +28,7 @@ jobs:
     name: Run tests matrix job
 
     strategy:
+      fail-fast: true
       matrix:
         os: [ubuntu-latest]
         client-type: [go, http, cli]
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -46,8 +46,18 @@ jobs:
             database-type: badger-memory
             mutation-type: collection-save
             detect-changes: false
+          - os: windows-latest
+            client-type: go
+            database-type: badger-memory
+            mutation-type: collection-save
+            detect-changes: false
 
     runs-on: ${{ matrix.os }}
+
+    # We run all runners via the bash shell to provide us with a consistent set of env variables and commands
+    defaults:
+      run:
+        shell: bash
 
     env:
       CGO_ENABLED: 1
--- a/datastore/badger/v4/datastore_test.go
+++ b/datastore/badger/v4/datastore_test.go
@@ -60,6 +60,10 @@ func TestNewDatastoreWithOptions(t *testing.T) {
 
 	s, err := NewDatastore(dir, &opt)
 	require.NoError(t, err)
+	defer func() {
+		err := s.Close()
+		require.NoError(t, err)
+	}()
 
 	s.Put(ctx, testKey1, testValue1)
 	s.Put(ctx, testKey2, testValue2)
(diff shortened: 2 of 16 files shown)


fredcarle:
The errors in the tests seem to all be related to incorrect file paths. Should be easy to fix.

datastore/badger/v4/datastore_test.go
```diff
@@ -8,6 +8,13 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+// The badger GC seems to struggle a bit on windows, and this file rapidly spins up and discards
+// databases resulting in out of memory and vlog cleanup issues, so we exclude this file from windows
+// builds and and rely on the integration tests.
```
islamaliev, line 13: typo: "and and"

AndrewSisley: Cheers :grin: 

- [ ] Fix typo

AndrewSisley: Was able to remove this restriction instead

.github/workflows/test-and-upload-coverage.yml
```diff
@@ -28,6 +28,7 @@ jobs:
     name: Run tests matrix job
 
     strategy:
+      fail-fast: true
```
shahzadlone, line 31: suggestion: making it false would let other runners complete even if one runner is flakey and fails. Otherwise all will fail upon one flakey runners and would need to re-run all failed runners instead of just the flakey one.

AndrewSisley: I think such a discussion/change is out of scope here, and does not need to interrupt the delivery of this PR.

shahzadlone: The addition of this line within this PR does affect / change the behavior of the ci.

AndrewSisley: Only if github changes it's default.  I will remove this line if you feel like this is a change, although I did bring it into this PR after you suggested I do so with this `true` value (in another PR).

shahzadlone: Okay let's keep it for some reason I thought I suggested the other way around. (I wanted to maintain the default github behavior, which seems like this maintains).

shahzadlone (approved):
Maybe @fredcarle can comment regarding the removed net test, all rest LGTM

AndrewSisley:
> Maybe @fredcarle can comment regarding the removed net test, all rest LGTM

This PR has idled for a while, can bring back/discuss the removed test post merge if need be.

PR #2035, opened 2023-11-10
ci: Add mac test runner

### LINKED ISSUE #2034: Add mac test runner
These only need to run the simple set of tests. See also https://github.com/sourcenetwork/defradb/issues/2032

AndrewSisley:
## Relevant issue(s)

Resolves #2034

## Description

Adds a mac (latest) test run to our test matrix.

This only use the lightest of configurations we have, I think this will catch any/99% issues that we would otherwise miss.

Windows has been broken out to a different ticket, as quite a lot of tests fail on windows and it will take more effort to get working: https://github.com/sourcenetwork/defradb/pull/2033

### DIFF
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -27,19 +27,26 @@ jobs:
   run-tests:
     name: Run tests matrix job
 
-    runs-on: ubuntu-latest
-
     strategy:
       matrix:
+        os: [ubuntu-latest]
         client-type: [go, http, cli]
         database-type: [badger-file, badger-memory]
         mutation-type: [gql, collection-named, collection-save]
         detect-changes: [false]
         include:
-          - client-type: go
+          - os: ubuntu-latest
+            client-type: go
             database-type: badger-memory
             mutation-type: collection-save
             detect-changes: true
+          - os: macos-latest
+            client-type: go
+            database-type: badger-memory
+            mutation-type: collection-save
+            detect-changes: false
+
+    runs-on: ${{ matrix.os }}
 
     env:
       DEFRA_CLIENT_GO: ${{ matrix.client-type == 'go' }}


shahzadlone (approved):
thought: Perhaps an additional check on line 83, where coverage reports are uploaded within `step.name: Upload coverage artifact` and line 75 within `step.name: Run integration tests` so that we don't generate and don't upload code coverage reports for the same type of test twice (as a ubuntu variant will already have had been uploaded by other run), and skip it for macos and windows like we do for change detection? Then you would need to update the ` steps.name: Run change detector tests` condition (line 79) to allow macos test.
If you do the above, would suggest renaming the following step names:
1) Rename "Run integration tests" -> "Run tests with coverage"
2) Rename "Run change detector tests" -> "Run tests without coverage"

Other than that rest LGTM! now all we need to do is update these 19 required matrix jobs lol with 20 new ones haha:

image

shahzadlone:
@AndrewSisley 

Once ready to merge, LMK I can remove the previous required checks.

.github/workflows/test-and-upload-coverage.yml
```diff
@@ -27,19 +27,26 @@ jobs:
   run-tests:
     name: Run tests matrix job
 
-    runs-on: ubuntu-latest
-
     strategy:
       matrix:
```
shahzadlone, line 31: thought(out of scope): While you are at it, perhaps an explicit `fail-fast: true` (which is the default behavior at the moment). To have it stay consistent incase GitHub decides to change the default in the future.

AndrewSisley:
@shahzadlone 

> Perhaps an additional check on line 83 (not uploading code-cov)

I do disagree with this, and I do want mac/windows to contribute to the code cov upload.  They may test OS specific lines, and the cost to including them should be tiny.  If the windows run turns out to be significantly slower and the primary bottleneck we can revisit, but even then I would be hesitant to support excluding it.

PR #2051, opened 2023-11-13
test: Add auto-doc generation

### LINKED ISSUE #1974: Random doc generation based on GraphQL schema
We want to be able to generate document from a GraphQL schema.
It should be possible to configure number of generated documents, number of related to documents in one-to-many relation and others

islamaliev:
## Relevant issue(s)

Resolves #1974

## Description

This pull request introduces `gen.AutoGenerateDocs` function, a dynamic and versatile tool designed for the automatic generation of documents based on specified collection definitions. This functionality is essential for testing and simulating various data models and scenarios, providing extensive customization and intelligent demand calculation for document generation.

### Key Features

**Dynamic Document Generation**: `AutoGenerateDocs` can generate documents adhering to a defined schema, interpreting types and relationships to create realistic, interconnected data structures.
**Intelligent Demand Calculation**: The function calculates the 'demand' for documents based on the configuration, adjusting the number of generated documents for consistency in relationships (e.g., one-to-one, one-to-many).
**Flexible Configuration**: Documents can be configured directly within the schema using annotations or via options passed to the function, with options taking precedence.
**Customization Options**: Includes capabilities like `WithTypeDemand`, `WithFieldRange`, `WithFieldLen`, `WithFieldGenerator`, and `WithRandomSeed` for extensive customization and deterministic output.

### Additional Functionality: Generation of Predefined Documents

**GenerateDocs Function**: Allows the generation of predefined documents, considering only the fields present in the schema and ignoring any extra fields in the `DocList`.

## Tasks

## How has this been tested?

Unit and integration tests.

- MacOS


tests/gen/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+```go
+schema := `
+type User {
+  name: String # len: 10
+  age: Int # min: 18, max: 50
+  verified: Boolean
+  rating: Float # min: 0.0, max: 5.0
+}`
+docs, _ := AutoGenerateDocs(schema, WithTypeDemand("User", 100))
```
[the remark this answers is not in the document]
AndrewSisley: Would probably be something like:
`AutoGenerateDocs(definitions, WithTypeDemand("User", 100))`

You could then have a syntax sugar layer on top with roughly the current signature that converts and SDL to definitions before processing internally, you'd probably want to add a `client.Store` param or something for that though to let the prod code handle the conversion from SDL to definition.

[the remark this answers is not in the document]
AndrewSisley: I'm tempted to see this as a todo, as there is a risk that this would become a major blocker and source of tech debt when we expose other means of creating schema/collections - we'll almost certainly want to include them in the test matrix, like we do for mutation and client types, and notable numbers of tests are using this (GQL SDL only) then they will inhibit full testing of that change.

We already exposed `db.CreateSchema(defintion)` briefly (earlier this year?), before hiding it (I cant remember why, I think it was a technical reason RE updating the GQL types), so I think there are decent odds of this being wanted again soon are reasonably high.

tests/gen/defs.go
islamaliev, line 1: I'm not sure `gen` package should be inside `tests` folder because it's not necessarily test-related.
If you know a better place, let me know. Another option would be the root.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,246 @@
...
+	defaultIntMax = 10000
+)
+
+type docRec struct {
+	doc    map[string]any
+	docKey string
+}
+
+// AutoGenerateFromSchema generates random documents from a schema.
+func AutoGenerateFromSchema(schema string, options ...Option) ([]GeneratedDoc, error) {
```
AndrewSisley, line 35: todo: I think `schema` is a misleading term here, this function takes a GQL SDL document that may contain many schema (and collection) definitions. `Schema` is too easily confused with actual defra `Schema`, and it is made worse by the param being of type `string` (is it json of a single defra schema?  Even reading the function source code it is not possible to tell given the (sensible) abstractions used).

I think this should be renamed to something else. Maybe `AutoGenerateFromGQLSDL`, although the two acronyms blend together with that, maybe `AutoGenerateFromSDL` is sufficient.

tests/gen/schema_parser.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, err
+	}
+	cols, err := parser.ParseSDL(context.Background(), schema)
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string]client.CollectionDefinition)
+	for i, col := range cols {
+		col.Description.ID = uint32(i)
```
AndrewSisley, line 33: thought: I think this line is a notable risk.  I'm guessing a little bit here, but I assume you use this as the collection index - used by documents in the integration test harness to attach themselves to a particular collection.

thought: *If* we ever want this tool to be moved out of the test package this will very likely need changing, as it then likely need to fully mimic the production code used when creating collections, which this code does not currently do as far as I can see.

suggestion: I suggest trying to find a way to remove need for this line somehow.

todo: If this line (or similar) is preserved, it very definitely needs to be documented.

tests/gen/util_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return doc.Key().String()
+}
+
+func getDocKeyFromDocMap(docMap map[string]any) string {
+	docJSON, err := json.Marshal(docMap)
+	if err != nil {
+		panic("can not marshal doc " + err.Error())
+	}
+	return getDocKeyFromDocJSON(string(docJSON))
```
AndrewSisley, line 88: todo: You appear to be converting from `[]byte` to `string` (this line) to `[]byte` (first line of `getDocKeyFromDocJSON`) here for no reason.  I'd suggest changing `getDocKeyFromDocJSON` to take a `[]btye` instead (I think that makes more sense anyway regardless of the double convert).

tests/gen/util_test.go
```diff
@@ -0,0 +1,89 @@
...
+				continue outer
+			}
+		}
+		return fmt.Sprintf("expected doc not found: %v", expected[i])
+	}
+
+	return ""
+}
+
+func getDocKeyFromDocJSON(docJSON string) string {
```
AndrewSisley, line 75: suggestion: I believe convention is that functions that deliberately panic instead of returning errors should be prefixed with `must`, i.e. `mustGetDocKeyFromDocJSON`

tests/gen/schema_parser.go
islamaliev: the result of docs generation is a list of `GeneratedDoc`:
```go
type GeneratedDoc struct {
	ColID uint32
	JSON  string
}
```
This is basically the only reason why I need the collection ID. It's not used for anything else. If you think some other format of output would suit better our needs, let me know. We could use instead the collection's name.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,246 @@
...
+
+import (
+	"math/rand"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/client/request"
+)
+
+const (
+	defaultNumDocs           = 10
```
AndrewSisley, line 21: suggestion: It might be nicer if these were public, particularly if we are considering moving this tool out of the test directory.

todo: Public or private, I think these consts could benefit from documentation.

tests/gen/schema_parser.go
AndrewSisley: I think what you are doing is incorrect here then.  The `CollectionID` on the `Create/UpdateDoc` is poorly named and misleading.  It is not `Collection.ID`, but is the index of the collection within the test framework, at the moment it happens to (mostly) match the `Collection.ID`, but that is really a coincidence.

Name or the full struct (expensive if not using a pointer), would be better here I think.

tests/integration/test_case.go
```diff
@@ -338,6 +339,41 @@ type Request struct {
 	ExpectedError string
 }
 
+// GenerateDocsFromSchema is an action that will trigger generation of documents for the given schema.
+type GenerateDocsFromSchema struct {
```
AndrewSisley, line 343: suggestion: I do think these actions do too much, and could be simplified a fair bit.  It does not make sense to me to force consumers of these to use them over the existing `SchemaUpdate` action, when they should not be mutually exclusive.  The current setup also makes it painful to use them on a patched schema.

I would want the tests to look something like this:
```
testUtils.SchemaUpdate{
				Schema: `
					type Users {
						name: String
					}
				`,
			},
			testUtils.GenerateDocs{
			   AutoGenOptions{
			      NumberOfDocs: 10,
			   }
			},
			testUtils.Request{
			...
			},
```
At the point at which the test framework processes the GenerateDocs action it already knows which collections exist, and has access to their `[]client.Collection`.  There appears to be no reason to have new actions responsible for saving schema in the database in addition to the one we already have.

A setup like this would also automatically work with schema and collection patches.

You could then have an optional prop where consumers can optionally provide names or indexes of collections they wish to generate docs for (or just have a `gen.Option` for that).

islamaliev: I like this idea very much. Wasn't sure myself if the current way if very optimal.
Will take care of it

client/descriptions.go
```diff
@@ -287,6 +287,24 @@ func (f FieldDescription) IsPrimaryRelation() bool {
 	return f.RelationType > 0 && f.RelationType&Relation_Type_Primary != 0
 }
 
+// IsRelation returns true if this field is a relation.
+func (f FieldDescription) IsRelation() bool {
+	return f.RelationType > 0
+}
+
+// IsArray returns true if this field is an array type.
```
AndrewSisley, line 295: todo: Looking at the implementation, it looks like this function will return true if it is an inline array, or the many side of a one-many.  I think it is really worth highlighting this in the documentation as it would be easy enough to assume otherwise.

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
+# Automatic Documents Generation 
+
+`AutoGenerate` and `AutoGenerateFromSDL` are a highly versatile function designed for dynamic document generation, perfect for testing and simulation purposes. 
+
+`AutoGenerateFromSDL` creates documents based on a specified schema, allowing for extensive customization of data generation.
```
AndrewSisley, line 5: todo: I think `documents based on a specified schema,` is incorrect, and should instead read something like `documents based on a specified GQL SDL, which may contain multiple schema/collection definitions,`.

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
+# Automatic Documents Generation 
+
+`AutoGenerate` and `AutoGenerateFromSDL` are a highly versatile function designed for dynamic document generation, perfect for testing and simulation purposes. 
+
+`AutoGenerateFromSDL` creates documents based on a specified schema, allowing for extensive customization of data generation. 
+
+The function generates documents adhering to a defined schema and it's configuration.
+It interprets the types and relationships within the schema to create realistic, interconnected data structures.
```
AndrewSisley, line 8: todo: I think `defined schema and` and `relationships within the schema to` are incorrect, and should really read `defined collection and` and `relationships within the collection to`.

The distinction is still a bit blurry atm, but we are planning moving relation definitions off the schema and onto the collection (leaving only the primary side field on schema).

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+`AutoGenerate` creates documents based on the provider collections' definitions (`[]client.CollectionDefinition`)
+
+### Demand Calculation:
+
+The functions calculate the 'demand' or the number of documents to generate based on the configuration provided.
+For related types within the schema, it intelligently adjusts the number of generated documents to maintain consistency in relationships (one-to-one, one-to-many, etc.).
```
AndrewSisley, line 15: todo: Similar to other points, but I think that `within the schema, ` is incorrect here, and should instead read something like `within the collection set, `

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+In the absence of explicit demands, it deduces demands from the maximum required by related types or uses a default value if no relation-based demands are present.
+
+The error will be returned if the demand for documents can not be satisfied. 
+For example, a document expects at least 10 secondary documents, but the demand for secondary documents is 5.
+
+## Configuration
+
+Both functions `AutoGenerate` and `AutoGenerateFromSDL` can be configured using options.
+
+Additionally, `AutoGenerateFromSDL` can be configured directly within the schema using annotations.
```
AndrewSisley, line 26: todo: Similar to other comments, but I think `within the schema using ` should instead read `within the SDL doc using `

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+- `WithFieldRange` and `WithFieldLen` override in-schema configurations for field ranges and lengths.
+- `WithFieldGenerator` provides custom value generation logic for specific fields.
+- `WithRandomSeed` ensures deterministic output, useful for repeatable tests.
+
+## Examples
+
+### Basic Document Generation:
+
+```go
+schema := `
```
AndrewSisley, line 51: nitpick: `sdl :=` might be better here

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+  age: Int # min: 18, max: 50
+  verified: Boolean
+  rating: Float # min: 0.0, max: 5.0
+}`
+docs, _ := AutoGenerateFromSDL(schema, WithTypeDemand("User", 100))
+```
+
+### Custom Field Range:
+
+Overrides the age range specified in the schema.
```
AndrewSisley, line 63: todo: Similar to others, I think `sdl doc` instead of `schema` would be better here

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+```go
+docs, _ := AutoGenerateFromSDL(schema, WithTypeDemand("User", 50), WithFieldRange("User", "age", 25, 30))
```
AndrewSisley, line 66: nitpick: I'd prefer `(sdl,` instead of `(schema,` here

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+### One-to-Many Relationship:
+
+Generates User documents each related to multiple Device documents.
+
+```go
+schema := `
```
AndrewSisley, line 74: nitpick: `sdl :=` might be better here

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Custom Value Generation:
+
+Custom generation for age field.
+
+```go
+nameWithPrefix := func(i int, next func() any) any {
+  return "user_" + next().(string)
+}
+docs, _ := AutoGenerateFromSDL(schema, WithTypeDemand("User", 10), WithFieldGenerator("User", "name", nameWithPrefix))
```
AndrewSisley, line 94: nitpick: I'd prefer `(sdl,` instead of `(schema,` here

tests/gen/README.md
```diff
@@ -0,0 +1,156 @@
+# Automatic Documents Generation
```
AndrewSisley, line 1: praise: I think this readme is very valuable and I thank you for taking the time to write it :)

tests/gen/defs.go
```diff
@@ -0,0 +1,38 @@
...
+//				"age":      20,
+//				"devices": []map[string]any{
+//					{
+//						"model": "iPhone Xs",
+//					},
+//				},
+//			},
+//		},
+type DocsList struct {
+	ColName string
```
AndrewSisley, line 30: nitpick: If you are looking to move this out of the `tests` directory and into a public space, these properties should also have documentation on them.  Might as well quickly add it now IMO.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// DefaultStrLen is the default length of a string to generate.
+	DefaultStrLen = 10
+	// DefaultIntMin is the default minimum value of an integer to generate.
+	DefaultIntMin = 0
+	// DefaultIntMax is the default maximum value of an integer to generate.
+	DefaultIntMax = 10000
+)
+
+type docRec struct {
```
AndrewSisley, line 34: todo: I have no idea what a `docRec` is, please rename and/or add some documentation :)

tests/gen/gen_auto_config.go
```diff
@@ -0,0 +1,172 @@
...
+
+// genConfig is a configuration for a generation of a field.
+type genConfig struct {
+	labels         []string
+	props          map[string]any
+	fieldGenerator GenerateFieldFunc
+}
+
+// configsMap is a map of type name to a map of field name to a generation configuration.
+type configsMap map[string]map[string]genConfig
```
AndrewSisley, line 27: praise: Thanks for aliasing and doucmenting this, it makes the referencing code much easier to read

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+	types        map[string]client.CollectionDefinition
+	config       configsMap
+	resultDocs   []GeneratedDoc
+	usageCounter typeUsageCounters
+	cols         map[string][]docRec
+	docsDemand   map[string]typeDemand
+	random       rand.Rand
+}
+
+func (g *randomDocGenerator) GenerateDocs(options ...Option) ([]GeneratedDoc, error) {
```
AndrewSisley, line 77: question: Is this function deliberately public? And if so, why?

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, def := range definitions {
+		typeDefs[def.Description.Name] = def
+	}
+	generator := randomDocGenerator{types: typeDefs}
+	return generator.GenerateDocs(options...)
+}
+
+type randomDocGenerator struct {
+	types        map[string]client.CollectionDefinition
+	config       configsMap
```
AndrewSisley, line 69: suggestion:  Given the complexity of this type, I think it might be worth highlighting that `types` and `config` are immutable properties defined on construction, and the rest are internal mutable state.

I would probably add a blank line between the two property sets, a `newRandomDocGenerator` constructor (also makes it easier to find where it is constructed), and some documentation - but have have no strong feelings on how you go about clarifying this (if you do).

I think it may also be worth highlighting somehow that whilst the construction suggests that `GenerateDocs` could be called multiple times, this is not true (as far as I can see).  Or perhaps better, but probably harder, change the implementation/structure so that multiple `GenerateDocs`  calls are possible, and are inherently safe (would probably mean splitting the type).

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+	random       rand.Rand
+}
+
+func (g *randomDocGenerator) GenerateDocs(options ...Option) ([]GeneratedDoc, error) {
+	g.cols = make(map[string][]docRec)
+	if g.config == nil {
+		g.config = make(configsMap)
+	}
+
+	configurator := newDocGenConfigurator(g.types, g.config)
```
AndrewSisley, line 83: nitpick: It feels strange that this line `newDocGenConfigurator` is done here, when the only input params are local constants externally provided on object construction - IMO this should be done in a `newRandomDocGenerator` func.

Same goes for `g.cols` and `g.config`.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, docsList := range docsLists {
+		typeDef := g.types[docsList.ColName]
+		for _, doc := range docsList.Docs {
+			g.resultDocs = append(g.resultDocs, GeneratedDoc{
+				ColName: typeDef.Description.Name,
+				JSON:    createDocJSON(&typeDef, doc),
+			})
+		}
+	}
+	return g.resultDocs, nil
```
AndrewSisley, line 108: suggestion: Having now read this function, I cannot really tell why the state-machine like pattern is used here.  I think it would be easier to read/maintain if instead of mutating the host object state you just returned a `results` struct, with whatever stuff needed on it.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+	generator := randomDocGenerator{types: typeDefs}
+	return generator.GenerateDocs(options...)
+}
+
+type randomDocGenerator struct {
+	types        map[string]client.CollectionDefinition
+	config       configsMap
+	resultDocs   []GeneratedDoc
+	usageCounter typeUsageCounters
+	cols         map[string][]docRec
```
AndrewSisley, line 72: todo: Related to my uncertainty RE what `docRec` is, I can't fully tell what this property is either.  If you don't feel like it is clear having renamed `docRec`, I would suggest renaming and/or documenting this property too.

tests/gen/defs.go
jsimnz: I like it where it is :)

jsimnz:
Haven't done a full review (and merging this isn't dependent on my review). Just want to chime in and say this looks really nice! Should really expand the kinds of tests we can write, without being exceedingly laborious/tedious. Additionally when we re-enable the benchmarks, can also be a really great data generation for that as well. 

Cool shit!

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,251 @@
...
+				} else {
+					fieldConf := g.config.ForField(typeName, field.Name)
+					newDoc[field.Name] = g.generateRandomValue(typeName, field.Kind, fieldConf)
+				}
+			}
+			docKey, err := g.getDocKey(&typeDef, newDoc)
+			if err != nil {
+				return nil, err
+			}
+			g.cols[typeName] = append(g.cols[typeName], docRec{doc: newDoc, docKey: docKey})
```
AndrewSisley, line 161: suggestion: It feels like this would be easier to read if the population of `g.cols` with `typeName` was done outside of the block that creates documents.  This thought might however be related to my uncertainty as to why `order` is important in this function, I assume it is, and I assume it is to do with the order in which documents must be generated, so I also suggest documenting why type order is important to this function.

AndrewSisley (changes requested):
I really like this, so please don't think otherwise based on the number of comments.  I'm only part way through my review and will continue after the standup.

tests/gen/defs.go
AndrewSisley: I'm happy with where it is for now, but this could be something that we treat as a public tool and as such I think I would want it somewhere else long term.

tests/gen/README.md
jsimnz: > but we are planning moving relation definitions off the schema

Not 100% accurate, the decision was to not require *both* sides of the schema relation to be defined as I understood it. 
At the moment, we define the primary and secondary side of a realtionship, which causes some confusion and is overall a bad DX/UX. My understanding of the discussion on this was that we would only require adding a relationship to the primary side that actually depends/needs it, and then on the collection definition on the secondary side create a "fake" field so you can still query from both sides.

AndrewSisley: > My understanding of the discussion on this was that we would only require adding a relationship to the primary side that actually depends/needs it, and then on the collection definition on the secondary side create a "fake" field so you can still query from both sides.

Yes, this is larger what I meant - I just don't see the primary field as being a relationship definition, just a type declaration that points 'somewhere else'.

client/descriptions.go
islamaliev: I'm not quite following. What do you mean by "inline array"? And it's not quite clear what can be assumed otherwise. It definitely can't be the one side of one-many.

tests/gen/gen_auto.go
islamaliev: thanks for pointing this out. I turns out the structure is not needed, just the dockey.

tests/gen/gen_auto.go
islamaliev: one is not supposed to call `GenerateDocs` of `randomDocGenerator` type multiple times, it's private after all. One should call `gen.AutoGenerate`. It is safe to call multiple time because the generated is instantiated anew.

client/descriptions.go
AndrewSisley: An inline array is an array defined on the object itself (not a relation), i.e. `FieldKind_INT_ARRAY` or `FieldKind_BOOL_ARRAY`.

Pretty much everywhere in our code base differentiates (often out of out of technical necessity) between inline arrays and relation-arrays.  It is strange and unexpected to find a function that does not, and as such adding a handful of words to the documentation to make this clear will likely save a headache or too in the future.

tests/gen/gen_auto.go
AndrewSisley: Cheers, I really do miss proper access modifiers :)

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func newRandomDocGenerator(types map[string]client.CollectionDefinition, config configsMap) *randomDocGenerator {
+	if config == nil {
+		config = make(configsMap)
+	}
+	configurator := newDocGenConfigurator(types, config)
+	return &randomDocGenerator{
+		configurator:     configurator,
+		generatedDocKeys: make(map[string][]string),
```
AndrewSisley, line 69: suggestion:  Related to other comments, but now, if you moved this one line into `generateDocs`, `generateDocs` very much looks like it could be used multiple times, in which case the other conversation RE it's name is bypassed/irrelevant

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+				} else {
+					fieldConf := g.configurator.config.ForField(typeName, field.Name)
+					newDoc[field.Name] = g.generateRandomValue(typeName, field.Kind, fieldConf)
+				}
+			}
+			docKey, err := g.getDocKey(&typeDef, newDoc)
+			if err != nil {
+				return nil, err
+			}
+			g.generatedDocKeys[typeName] = append(g.generatedDocKeys[typeName], docKey)
```
AndrewSisley, line 155: suggestion: A code comment here explaining why you are appending the dockey here may save some future head scratching - it is not obvious glancing over the code, especially given that the map is not read from in this function.  I'd recommend highlighting why it needs to be held on the host object and can't be a local var passed into `getNextPrimaryDocKey` too, as that is also not obvious (and I assume that it is needed, if not please remove from the host state).

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// getNextPrimaryDocKey returns the key of the next primary document to be used as a relation.
+func (g *randomDocGenerator) getNextPrimaryDocKey(secondaryType string, field *client.FieldDescription) string {
+	ind := g.configurator.usageCounter.getNextTypeIndForField(secondaryType, field)
+	docKey := g.generatedDocKeys[field.Schema][ind]
+	return docKey
+}
+
+func (g *randomDocGenerator) getDocKey(typeDef *client.CollectionDefinition, doc map[string]any) (string, error) {
```
AndrewSisley, line 120: todo: `getFoo` suggests quite a cheap operation, typically a key lookup or something, but this function is actually doing an awful lot of work, serializing to JSON, then deserializing to `client.Document`, then getting the key.

If this function remains as such, I'd suggest renaming it to `generateDocKey` or similar.

I would however instead recommend a rework - it feels very wasteful to do all this work, only to then redo the json serialisation later (`GeneratedDoc` returns JSON atm), and then have that json deserialized (again) to `client.Document` by the integration test suite when these auto generated documents are consumed.

I don't think now is the appropriate place to look at returning `client.Documents`, as that would likely involve changing the core test suite stuff, but we could at least remove the need to serialise the document twice to json.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		result = append(result, col)
+	}
+	return result, nil
+}
+
+func getRandomString(random *rand.Rand, n int) string {
+	const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	b := make([]byte, n)
+	for i := range b {
```
AndrewSisley, line 166: nitpick: I think random number generation is relatively slow, and apart from the risk of collsion, there does not seem to be a good reason to independently randomise each char in the string, you could just generate a random start index, and optionally a randomized offset, then construct the string from that:

```
charIndex := random.Intn(len(letterBytes))
offset := random.Intn(len(letterBytes)) // or whatever :) 
for i := range b {
   b[i] = letterBytes[charIndex]
   charIndex += offset // You'll need to wrap this, I'm too lazy to write that here
}
```
Marked as a nitpick, as I'm not sure how significant the performance difference would be given normal sized documents and the relative cost of inserting them into Defra etc.

Perhaps an even better approach would to only randomize the initial string in the  `generateDocs` call, and then just increment the characters for subsequent strings - then there would be no risk of collision, and only one random call in the batch.

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+			g.generatedDocKeys[typeName] = append(g.generatedDocKeys[typeName], docKey)
+			col.Docs = append(col.Docs, newDoc)
+		}
+		result = append(result, col)
+	}
+	return result, nil
+}
+
+func getRandomString(random *rand.Rand, n int) string {
```
AndrewSisley, line 163: todo: There appears to be no protection against collisions here, which exposes us to flaky tests.

Consider the schema:
```
type User {
   Name: String # max: 1 // I cant remember the exact syntax, is not important here
}
```
1 in 26*2 docs would be identical and would cause tests to fail if they try to insert them.  Yes the odds increase with the number of chars, but we run the CI workflow a lot, and each run runs all the tests ~20 times, and there will hopefully be more than just one test using this doc generation, so I'd still expect us to be impacted by this if all the tests had 6 or so char strings.

I think it would be worth ensuring that each call to `generateDocs` cannot return the same string twice, for at least short-medium length strings.

This problem will exist for other field types too. although as we don't support `int8` etc it is unlikely to be a real problem there atm (maybe for bool inline-arrays).

AndrewSisley (changes requested):
Still working my way though, but submitting comments so far

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,202 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/client/request"
+)
+
+// GeneratePredefinedFromSDL generates documents for a schema from a predefined list
```
AndrewSisley, line 21: nitpick:  I think you are missusing `schema` a few times in this documentation, for example  `for a schema from` should probably read `for the given collections from` or something.  Also `The schema is parsed ` is incorrect, the SDL doc is parsed, not the schema, same problem exists in `fields parsed from the schema.`.

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,202 @@
...
+		resultDocs = append(resultDocs, docs...)
+	}
+	return resultDocs, nil
+}
+
+type docGenerator struct {
+	types map[string]client.CollectionDefinition
+}
+
+func createDocJSON(typeDef *client.CollectionDefinition, doc map[string]any) string {
```
AndrewSisley, line 67: todo: It is not obvious as to why this function is not just:
```
func createDocJSON(doc map[string]any) string{
   return json.Marshal(doc)
}
```
Can you please document what this function does and why we need to have custom json serialization logic here.

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,202 @@
...
+		} else {
+			sb.WriteString(",\n")
+		}
+		sb.WriteString(fmt.Sprintf(format, fieldName, doc[fieldName]))
+	}
+	sb.WriteString("\n}")
+	return sb.String()
+}
+
+func toRequestedDoc(doc map[string]any, typeDef *client.CollectionDefinition) map[string]any {
```
AndrewSisley, line 92: todo: Please document what this does, and what the other functions below do, it is not obvious to me at the moment.

tests/integration/test_case.go
```diff
@@ -338,6 +339,32 @@ type Request struct {
...
+
+	// Options to be passed to the auto doc generator.
+	Options []gen.Option
+
+	// The list of collection names to generate docs for.
+	// If not provided, docs will be generated for all collections.
+	ForCollections []string
+}
+
+// GeneratePredefinedDocs is an action that will trigger generation of documents.
```
AndrewSisley, line 357: todo: It is not obvious to me what this action does, it says it generates documents, yet it takes a set of documents.  If you are providing it a set of documents, what does it generate, and how many of them?

tests/gen/schema_parser.go
```diff
@@ -0,0 +1,134 @@
...
+	schemaLines := strings.Split(schema, "\n")
+	for _, line := range schemaLines {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "type ") {
+			typeNameEndPos := strings.Index(line[5:], " ")
+			currentType = strings.TrimSpace(line[5 : 5+typeNameEndPos])
+			currentConfig = make(map[string]genConfig)
+			continue
+		}
+		if strings.HasPrefix(line, "}") {
```
AndrewSisley, line 53: todo: I assume that `HasPrefix` is used here in case the line has comments after the `}`?  Can you please document why the code is doing this, as it is not obvious to the reader and requires some guess-work.

tests/gen/schema_parser.go
```diff
@@ -0,0 +1,134 @@
...
+	genConfigs := make(map[string]map[string]genConfig)
+
+	var currentConfig map[string]genConfig
+	var currentType string
+
+	schemaLines := strings.Split(schema, "\n")
+	for _, line := range schemaLines {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "type ") {
+			typeNameEndPos := strings.Index(line[5:], " ")
```
AndrewSisley, line 48: todo: It looks like `5` is used because that is how many character are in the string `type `, however readers are left to guess this.  Can you please extract `5` to a variable, maybe:
```
typePrefixLen := len("type ")
```

AndrewSisley (changes requested):
I really like this, but I think the code needs a bit more documentation before it can be merged.

Please comment against any comment/discussion-threads you resolve, it makes it much easier for me to tell what (if anything) has been done to resolve them.

tests/gen/gen_auto.go
islamaliev: the generator doesn't support at the moment unique fields. The example you provided will just generate documents the have the same values for `Name` field.
I plan on extending the generators capabilities like:
```graphql
type User {
   Name: String # min: 1, max: 1, unique
}
```
And a corresponding new option `WithUniqueField("User", "Name")`
But this will be probable another PR

AndrewSisley: I don't understand - you want to merge this knowing that it will cause our tests to be flaky?

AndrewSisley: Just to clarify why this is a problem - you cannot create two documents in a defra db with the same field values - it would result in the same doc key and the second create command will result in a `document already exists` error

tests/gen/gen_auto.go
islamaliev: feels this this comment slightly contradicts the other comment you left regarding collisions https://github.com/sourcenetwork/defradb/pull/2051/files/26ef63c132318f71007f4cb0a1710e3354043f52#r1401023373

With the approach that you proposed, yes, we get getter performance but at the expense of randomness, which will drastically increase chances for collision.

Lets take a string of length 8. With the current approach there are 52^8 permutations available. With your approach it drop to 52^2 which is just 2704.

Moreover, I don't think it should be hard-coded into the framework to set a level of randomness. It should be up to the client to decide. If the client is concerned about the performance he can just require string length to be a certain number (with number 2 it will call random number generator twice). If the client wants to have a longer name, he can define a decorator on top of the generated string:
```go
WithFieldGenerator("User", "name", func(i int, next func() any) any {
	return "some_prefix_" + next().(string)
}),
```
The performance concern arises mostly if a large set of document needs to be generated, but in this case you surely need a uniformly randomized string.

tests/gen/gen_auto.go
islamaliev: I'm aware of this constrain. That's why it's a bit confusing when you talk about the field uniqueness rather than doc uniqueness.

I've created a task for field uniqueness https://github.com/sourcenetwork/defradb/issues/2069.

For document uniqueness the story is a bit different. There are different ways of solving it.

One of them is to store a map of all dockeys generated and make sure newly created doc doesn't have the same dockey, and if so we should generated it again until the dockey is unique. The problem with this is that if the number of available permutations is close to number of demanded docs the generator will have hard times picking a unique combination. We could mitigate it by running some kind analysis after doc demand calculation to determine how likely we to collide.

I have some other interesting options in mind but I don't think it's worth the trouble. All this additional bookkeeping will make the code much more complex and less performant.

I think if the client wants to make sure there is no (or almost no) collision, he should provide a reasonable setup.
If his SDL is 
```graphql
type User {
    name: String # len: 1
}
```
and her requests 100 docs to generate, he is intentionally looking for trouble.

You said the if we use any but 8-bit integer it should be fine, but with string length of 3 you get twice as many permutations as with 16-bit integer.

tests/gen/gen_predefined.go
islamaliev: there is no json handling in `gen` package anymore

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (g *randomDocGenerator) generateRandomDocs(order []string) ([]DocsList, error) {
+	result := []DocsList{}
+	for _, typeName := range order {
+		col := DocsList{ColName: typeName}
+		typeDef := g.configurator.types[typeName]
+
+		currentTypeDemand := g.configurator.docsDemand[typeName]
+		averageDemand := currentTypeDemand.getAverage()
+		for i := 0; i < averageDemand; i++ {
```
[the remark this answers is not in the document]
islamaliev: added a comment

tests/gen/gen_auto.go
islamaliev: refactored

tests/gen/gen_auto.go
islamaliev: refactored

tests/gen/gen_predefined.go
islamaliev: changed

tests/gen/gen_predefined.go
islamaliev: done

tests/gen/gen_auto.go
AndrewSisley: > The performance concern arises mostly if a large set of document needs to be generated, but in this case you surely need a uniformly randomized string.

Why?

More collisions can still be cheaper than a complex randomization algo.

And, are you against the below suggestion, and if so why?:

> Perhaps an even better approach would to only randomize the initial string in the generateDocs call, and then just increment the characters for subsequent strings - then there would be no risk of collision, and only one random call in the batch.

tests/gen/gen_auto.go
AndrewSisley: The problem is that the risk on collision needs to be very low for it to not annoy us by tripping up the CI every now and then.  Say we have 50 tests that use this, and they each generate 11 docs, each push will run that 20 odd times (os, mutation type, client type, etc), and maybe on average each PR will have the CI run 20 times.  That is roughly 50*10*20*20=200000 chances per PR for duplicate docs to cause the CI to fail - increasing as this tool gets used by more tests, and if/as we hire more devs (more PRs).

And we can avoid this by just generating sequential values, or tracking which field values have been created in a test (very cheap, so long as the test is not creating billions of documents).

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,178 @@
...
+		resultDocs = append(resultDocs, docs...)
+	}
+	return resultDocs, nil
+}
+
+type docGenerator struct {
+	types map[string]client.CollectionDefinition
+}
+
+// toRequestedDoc removes the fields that are not in the SDL of the collection.
```
AndrewSisley, line 66: todo: This comment does not make sense to me, do you mean `the schema of the collection. `/`the collection`s schema.`?  And if so, how did these fields find their way on to the document in the first place, if they are not part of that collection/schema why were they added?

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (g *randomDocGenerator) generateRandomDocs(order []string) error {
+	for _, typeName := range order {
+		col := DocsList{ColName: typeName}
+		typeDef := g.configurator.types[typeName]
+
+		currentTypeDemand := g.configurator.docsDemand[typeName]
+		averageDemand := currentTypeDemand.getAverage()
+		// we need to decide how many documents to generate for this type
+		// and if it's a range (say, 10-30) we take average (20).
```
AndrewSisley, line 132: question: Why would anyone specify a range, if an average is always used?

tests/gen/gen_auto.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	configurator := newDocGenConfigurator(types, config)
+	return &randomDocGenerator{
+		configurator:  configurator,
+		generatedDocs: make(map[string][]genDoc),
+	}
+}
+
+type genDoc struct {
+	docKey string
```
AndrewSisley, line 74: question: Why does `docKey` exist as a property here?  Is is different to `doc.Key()`?  And if so, why?

tests/gen/gen_auto.go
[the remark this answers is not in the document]
AndrewSisley: Cheers :)  I have a follow-up question though

tests/gen/gen_predefined.go
AndrewSisley: Nice - thanks Islam

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	doc, err := client.NewDocFromMap(requested)
+	if err != nil {
+		return nil, NewErrFailedToGenerateDoc(err)
+	}
+
+	result = append(result, GeneratedDoc{Col: &typeDef, Doc: doc})
+
+	var docKey string
+	for _, field := range typeDef.Schema.Fields {
```
AndrewSisley, line 125: question: Why do you loop through the fields outside of `generateSecondaryDocs`, but inside of `generatePrimary`?

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	for name, val := range doc {
+		if strings.HasSuffix(name, request.RelatedObjectID) {
+			result[name] = val
+		}
+	}
+	return result
+}
+
+// generatePrimary generates primary docs for the given secondary doc.
```
AndrewSisley, line 83: todo: I think this documentation needs to highlight that it will essentially mutate the given `docMap` document, appending the keys of any generated secondary documents and then returning a `client.document` that has additional field values to the input `docMap`, and as a result, a different dockey.

tests/gen/gen_predefined.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+	return requested, result, nil
+}
+
+// generateRelatedDocs generates related docs (primary and secondary) for the given doc.
+func (this *docGenerator) generateRelatedDocs(docMap map[string]any, typeName string) ([]GeneratedDoc, error) {
+	typeDef := this.types[typeName]
+
+	requested, result, err := this.generatePrimary(docMap, &typeDef)
```
AndrewSisley, line 113: todo: This call must be made before doing anything else in this function due to the way `generatePrimary` mutates the document, however it is not obvious from reading the code.  I think this temporal coupling needs to be highlighted with a code comment, it would be easy enough to miss and then waste time on fixing should anyone start moving stuff around in this function.

tests/gen/gen_auto.go
islamaliev: While more collisions might be cheaper computationally, they can compromise the integrity of the system if uniqueness is a requirement. The cost of handling collisions (both in terms of performance and complexity) might outweigh the benefits of a simpler randomization algorithm.

> Perhaps an even better approach would to only randomize the initial string in the generateDocs call, and then just increment the characters for subsequent strings - then there would be no risk of collision, and only one random call in the batch.

I've actually thought about this option myself but more as an additional option for clients to chose between randomness vs performance.

tests/gen/gen_auto.go
islamaliev:
```graphql
type User {
    name: String
    devices: [Device] # max: 1, max 5
}
...
```
If you demand 100 `User` to generate, then the demand for `Device` is between 100 and 500, which is range.

AndrewSisley: The range is never used though, in the above example there would never be any users with 1 or 5 devices, every user would have 3 devices no?

tests/gen/gen_auto.go
islamaliev: `Key.String()` also does some computation. So this one is kind of cache. Depending on the setup the same document can be linked multiple times to other ones.

AndrewSisley: > Key.String() also does some computation

Very very cheap computation I think no?

> Depending on the setup the same document can be linked multiple times to other ones.

I do not understand you here, are you suggesting that `docKey` is not always equal to `doc.Key().String()`?

tests/gen/gen_auto.go
islamaliev: no, that's not how it works. In the given example the average would be 300, which means that `Device` docs will be generated. These 300 devices will be randomly distributed across 100 users. So yes, each user will have from 1 to 5  devices randomly.

tests/gen/gen_auto.go
AndrewSisley: I got too curious and checked the odds, if the maths in `50*(11-1)*(11-1)*20*20` is correct, for a schema with a single int32 field (e.g. Age), the odds of the tests not flaking out in 1 PR (20 pushes, 50 tests, 11 docs per test), are `0.999069110971`, i.e. 1 in 1000 chance of failing.  If we have 10 PRs, the odds of no failure = `0.990730008019` (1 in 100).  100 PRs = `0.911072996106` (9 in 100), 1000 PRs = `0.39403222849` (6 out of 10 chance of failure).  The odds are worse for short strings.

tests/gen/gen_auto.go
AndrewSisley: Ah okay nice! Thanks :)

Where does that take place though?  This loop would always be `3` though, does this mean that every `Device` would be linked to exactly `3` Users?  Do you validate somewhere that the random stuff does not pick the same 3 Users for every `Device`?

tests/gen/schema_parser.go
islamaliev: restructured the parsing so that it now can handle unformatted SDL

tests/gen/gen_auto.go
islamaliev: > Where does that take place though? 

> Do you validate somewhere else that the random stuff does not pick the same 3 Users for every Device?

you wanna take a look `(relationUsage) useNextDocKey()`

> This loop would always be 3 though, does this mean that every Device would be linked to exactly 3 Users?

With the example above this loop will make 300 iterations to produce 300 devices. Users will be randomly picked from the pool of available users for this particular relation.

tests/gen/gen_auto.go
islamaliev: I've reviewed the calculations you provided regarding the probabilities of collisions in our CI tests. I noticed a small discrepancy in the multiplication that affects the final number of opportunities for collision.

> if the maths in 50*(11-1)*(11-1)*20*20 is correct

Seems like you incorrectly took a square of number of tests. It should be just `50*(11-1)*20*20`
Which gives us 200,000 instead of 2,000,000 chances for duplication.

With the corrected calculation, the probability of at least one collision happening across 100 PRs, when using a 5-character string, is approximately 5.12%.

To be honest, I'm not sure why we are so much concerned about this kind of flakiness. I can hardly remember any of my PRs that didn't fails for a not legit reason (changes-related), so I just had to restart the job.

I don't say that we should neglect the number (your or mine) but I'd say it's not that high of a priority now. I propose to add a task for different way of random value generation. Maybe with lexicographically incrementable strings as the default option.

tests/gen/gen_predefined.go
islamaliev: There is an example of it in README.md. You can provide a large predefined set to `gen.GeneratePredefined` with document that might contain properties that are not defined in the collection. They will be filtered out.

tests/gen/gen_predefined.go
islamaliev: I don't understand your question. What do you mean "inside"/"outside"?
There are loops inside of both

tests/gen/gen_predefined.go
islamaliev: extended doc

tests/gen/gen_predefined.go
islamaliev: I don't understand your point. `generatePrimary` is called before anything else is done.

tests/gen/gen_auto.go
AndrewSisley: Ah, I think it is (11-1)^2 because each document (n-1) has n-1 chances to collide with another doc.

>  I propose to add a task for different way of random value generation. Maybe with lexicographically incrementable strings as the default option.

I'm a bit uncomfortable with that, but can accept it - it isn't like this PR converts all our tests to using it, and it should be fairly quick to fix if/when it starts flaking out (although that will cause delay and annoyance when working on unrelated stuff).

tests/gen/gen_auto.go
AndrewSisley: > With the example above this loop will make 300 iterations to produce 300 devices. Users will be randomly picked from the pool of available users for this particular relation.

In that example would the input param `order` have `100` "User" strings in it?

tests/gen/gen_predefined.go
AndrewSisley: Okay fair enough, that is however unexpected - could you please amend this code comment so that it contains this information?  Or change the code so that fields that do not exist are not added to the doc in the first place. 

And please change the wording to something similar to the original suggestion - collections do not have SDLs, they have a schema.

tests/gen/gen_auto.go
AndrewSisley: Okay, fair enough :)

todo: Can you please document this on the property, making note that it should always equal `doc.Key().String()`, as at the moment readers either have to trace through all the references, guess, or ask you to know this.

tests/gen/gen_predefined.go
AndrewSisley: `generateRelatedDocs` contains:
```
// no loop
generatePrimary()
...

//loop
for _, field := range typeDef.Schema.Fields {
   if field.IsRelation() {
   ...
   generateSecondaryDocs
```

But `generatePrimary` contains:
```
for _, field := range typeDef.Schema.Fields {
   if field.IsRelation() {
   ...
   foo
```
Is there a reason for this inconsistency or is it just how the code happened to be written?

tests/gen/gen_predefined.go
AndrewSisley: Yes it is called before anything else is done, right now, in this version of the code.

However what is not clear from reading the code (besides the type change) is that it absolutely has to be called first because it mutates the document.

If anyone (including your later self) feels the need to change this function, it would be trivial to miss such a thing and then waste time trying to correct their mistake.

tests/gen/gen_auto.go
AndrewSisley: @fredcarle Tagging you here to get a third opinion on this conversation

fredcarle: I think that if we setup our doc generation properly we should have minimal impact due to collisions. My opinion is to give this a try and fix it if it becomes a problem.

tests/gen/gen_auto.go
islamaliev: > In that example would the input param order have 100 "User" strings in it?

`order` is the ordered list of collection names that need to be generated from left to right `[]string{"User", "Device", "Component"}`

tests/gen/gen_auto.go
islamaliev: added a task https://github.com/sourcenetwork/defradb/issues/2072

tests/gen/gen_predefined.go
islamaliev: > Or change the code so that fields that do not exist are not added to the doc in the first place.

Code doesn't add the fields. They are provided by the user, or a test. You can take a look at how it's used here tests/integration/index/query_with_relation_filter_test.go

tests/gen/gen_auto.go
islamaliev: added doc

tests/gen/gen_predefined.go
AndrewSisley: Ah okay - cheers, I think I get it now, this code is really removing the fields from the input 'seed' documents, not the resultant output?

If that's correct it might be worth noting that in the function documentation, something like:

```
// toRequestedDoc removes the fields that are not in the schema of the collection.
//
// This is typically called on user/test provided seed documents to remove any non-existent fields before generating
// documents from them. [And then perhaps clarify whether the input set is mutated, as I'm not sure]
```

tests/gen/gen_auto.go
AndrewSisley: In this User-Device example `averageDemand` would equal `3` though no?  And each iteration of this loop appears to only add a single document, where is this multiplied by 100?

client/descriptions.go
islamaliev: done

tests/gen/gen_predefined.go
islamaliev: done

tests/gen/gen_auto.go
islamaliev: I think you need to read the conversation again :) I already mentioned it:

> In the given example the average would be 300, which means that Device docs will be generated.

> With the example above this loop will make 300 iterations to produce 300 devices

`demand` is total demand of documents of a collection. It's not demand of subdocuments per document.

tests/integration/test_case.go
AndrewSisley: It doesn't multiply their number?  It just converts them from map to doc?

Can you please document this on the struct/action, I'd rather not have to read through the readme (which I may forget even exists) in order to understand this when reading tests.

tests/gen/gen_predefined.go
islamaliev: restructured

tests/integration/test_case.go
islamaliev: moved to another package to make is more distinctive

AndrewSisley (approved):
LGTM, thanks a bunch Islam! :)

PR #2091, opened 2023-11-28
feat: Add blob scalar type

### LINKED ISSUE #2090: Blob scalar type
Adding support for a blob scalar type will simplify storage of binary objects such as images, audio files, etc. in DefraDB.

nasdf:
## Relevant issue(s)

Resolves #2090 

## Description

This PR adds a blob scalar type to the schema system.

## Tasks

## How has this been tested?

Added integration tests.

- MacOS


request/graphql/schema/descriptions.go
```diff
@@ -14,6 +14,7 @@ import (
 	gql "github.com/sourcenetwork/graphql-go"
 
 	"github.com/sourcenetwork/defradb/client"
+	schemaTypes "github.com/sourcenetwork/defradb/request/graphql/schema/types"
```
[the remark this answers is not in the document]
nasdf: I'm open to that. Should we move the other types there as well?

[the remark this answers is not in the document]
AndrewSisley: This makes me nervous, I think we added DateTime to that package, but `graphql-go` is supposed to be a general purpose gql package, not a defra package.  And we may/do want to get rid of it/replace it at somepoint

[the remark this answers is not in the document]
jsimnz: > I think we added DateTime to that package

No, DateTime was already defined as a scalar in the `graphql-go` package, but there were some problems with it as it relates to Defra integration, there was PR that got merged on the `graphql-go` package related to the DateTime stuff, but it wasn't the implementation/definition of the DateTime scalar. (reference: https://github.com/sourcenetwork/defradb/pull/931 and https://github.com/sourcenetwork/graphql-go/pull/8)

I do agree though that we shouldn't have to define the scalar in the `graphql-go` package, it exposes the necessary types/functions for us to define our own scalars outside that package, exactly as they're being used here. Moreover, the fact that we have our own defined custom scalars implies that w.e package we may replace the `graphql-go` package with, should implement the "base" scalars as defined in the GQL spec, and all we have to worry about is the specific "custom" scalars that are neatly organized in a single place.

request/graphql/schema/descriptions.go
```diff
@@ -23,14 +24,15 @@ var (
...
-		&gql.List{}:   client.FieldKind_FOREIGN_OBJECT_ARRAY,
+		gql.ID:                      client.FieldKind_DocKey,
+		gql.Boolean:                 client.FieldKind_BOOL,
+		gql.Int:                     client.FieldKind_INT,
+		gql.Float:                   client.FieldKind_FLOAT,
+		gql.DateTime:                client.FieldKind_DATETIME,
+		gql.String:                  client.FieldKind_STRING,
+		&gql.Object{}:               client.FieldKind_FOREIGN_OBJECT,
+		&gql.List{}:                 client.FieldKind_FOREIGN_OBJECT_ARRAY,
+		schemaTypes.BytesScalarType: client.FieldKind_BYTES,
```
jsimnz, line 35: suggestion: Related to the other comment from Fred, I think it should be made clear that these are custom scalars, as the comment on line 36 suggests ("More custom ones to come"). Just a quick comment separating base scalars from custom scalars for clarity 👍

client/descriptions.go
```diff
@@ -204,6 +204,7 @@ var FieldKindStringToEnumMapping = map[string]FieldKind{
 	"String":     FieldKind_STRING,
 	"[String]":   FieldKind_NILLABLE_STRING_ARRAY,
 	"[String!]":  FieldKind_STRING_ARRAY,
+	"Bytes":      FieldKind_BYTES,
```
AndrewSisley, line 207: todo: I think this is a strange way of representing this, and that `[Byte]` is more sensible and fits much better with the rest of the types, and any potential future `Byte`, `[Byte!]` or `Byte!` types.

request/graphql/schema/types/scalars_test.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, c := range cases {
+		result := BytesScalarType.Serialize(c.input)
+		assert.Equal(t, c.expect, result)
+	}
+}
+
+func TestBytesScalarTypeParseValue(t *testing.T) {
+	input := "00ff"
+	output := []byte{0, 255}
+	invalid := "invalid"
```
AndrewSisley, line 43: question: Why is this invalid?

todo: Please document why this is invalid, if it remains invalid.

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,63 @@
...
+	ParseLiteral: func(valueAST ast.Value) any {
+		switch valueAST := valueAST.(type) {
+		case *ast.StringValue:
+			data, err := hex.DecodeString(valueAST.Value)
+			if err != nil {
+				return nil
+			}
+			return data
+		default:
+			return nil
```
AndrewSisley, line 60: todo: Similar to another comment, please document this - it looks like a silent error, but I only know that by guessing given the:
```
if err != nil {
	return nil
}
```
stuff in other code blocks

tests/integration/schema/updates/add/field/kind/invalid_test.go
```diff
@@ -64,30 +64,6 @@ func TestSchemaUpdatesAddFieldKind9(t *testing.T) {
 	testUtils.ExecuteTestCase(t, test)
 }
 
-func TestSchemaUpdatesAddFieldKind13(t *testing.T) {
```
AndrewSisley, line 67: todo: Please add a tests in `.../schema/updates/add/field/kind/foo.go` that tests that this field works with PatchSchema

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"encoding/hex"
+
+	"github.com/sourcenetwork/graphql-go"
+	"github.com/sourcenetwork/graphql-go/language/ast"
+)
+
+var BytesScalarType = graphql.NewScalar(graphql.ScalarConfig{
+	Name:        "Bytes",
+	Description: "The `Bytes` scalar type represents an array of bytes.",
```
AndrewSisley, line 22: question:  If it is an array of bytes, why does it only appear to accept string values?

todo: I'm not sure if it does only accept string values, could you add an integration test documenting the beheviour when provided an array of numbers:
```
testUtils.CreateDoc{
	Doc: `{
		"name": "John",
		"data": [0,1,2,3...]
	}`,
},
```

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,63 @@
...
+	Serialize: func(value any) any {
+		switch value := value.(type) {
+		case []byte:
+			return hex.EncodeToString(value)
+		case *[]byte:
+			return hex.EncodeToString(*value)
+		default:
+			return nil
```
AndrewSisley, line 30: todo: Please document this `return nil`

AndrewSisley (changes requested):
I have a handful of localized questions/todos that I would like resolving before merging, sorry about the hassle :)

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/sourcenetwork/graphql-go/language/ast"
+)
+
+var BytesScalarType = graphql.NewScalar(graphql.ScalarConfig{
+	Name:        "Bytes",
+	Description: "The `Bytes` scalar type represents an array of bytes.",
+	Serialize: func(value any) any {
+		switch value := value.(type) {
+		case []byte:
+			return hex.EncodeToString(value)
```
AndrewSisley, line 26: question: Why are you encoding/decoding at all here (including in ParseValue)? 

todo: The `graphql` library lacks any documentation here, but we can add some to our code to document why `Serialize` and `ParseValue` exist.

client/descriptions.go
nasdf: Updated to `Blob` as we discussed in the standup.

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,63 @@
...
+			}
+			return data
+		case *string:
+			data, err := hex.DecodeString(*value)
+			if err != nil {
+				return nil
+			}
+			return data
+		default:
+			return nil
```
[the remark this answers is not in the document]
nasdf: I'm just following how the other types work in the `graphql-go` library. It seems like `nil` is the return value when the type cannot be parsed.

request/graphql/schema/types/scalars.go
nasdf: I believe this is a bug in the`client.Document` API. You can also do this to a string field:

```
testUtils.CreateDoc{
		Doc: `{
			"name":  12345
		}`,
}
testUtils.CreateDoc{
		Doc: `{
			"name":  false
		}`,
}
```

AndrewSisley: Not quite a bug, we just dont do any checking of the saved value.  The read should fail though I think.  Does the read of `[0,1,2,3...]` fail?  Can we please document the behaviour, as it is a more likely question than assigning `1234` to a boolean :)

AndrewSisley (approved):
Looks good to me, thanks for this, and the extra tweaks Keenan - I'm approving now with one open/pending discussion RE why you are converting to hex and back that should be resolved before merge :)

nasdf:
> Looks good to me, thanks for this, and the extra tweaks Keenan - I'm approving now with one open/pending discussion RE why you are converting to hex and back that should be resolved before merge :)

Thanks! I refactored the parsing logic so that the value will always be a hex string.

request/graphql/schema/types/scalars.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+var BlobScalarType = graphql.NewScalar(graphql.ScalarConfig{
+	Name:        "Blob",
+	Description: "The `Blob` scalar type represents a binary large object.",
+	// Serialize converts the value to a hex string
+	Serialize: coerceBlob,
+	// ParseValue converts the value to a hex string
+	ParseValue: coerceBlob,
+	// ParseLiteral converts the ast value a hex string
```
fredcarle, line 55: nitpick: "to a hex string"?

islamaliev (commented):
Looks good, but one test is missing.

islamaliev (approved):
good job, Keenan!

PR #2096, opened 2023-12-01
feat: New cmd for docs auto generation

### LINKED ISSUE #2095: Auto-doc generator: create a cmd for doc generation
There is no way at the moment to use auto doc generator other then in tests. 
It should be possible to use the framework for generation of docs for a existing defradb storage.

islamaliev:
## Relevant issue(s)

Resolves #2095

## Description

This change adds a new command for generating documents automatically.

It also includes fixes for some overlooked edge cases.

## Tasks

## How has this been tested?

Unit tests and manual command execution with a running defradb node.

- MacOS

### DIFF
--- a/cli/gendocs.go
+++ b/cli/gendocs.go
@@ -1,0 +1,129 @@
+// Copyright 2023 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/config"
+	"github.com/sourcenetwork/defradb/tests/gen"
+)
+
+const gendocBatchSize = 1000
+
+func MakeGenDocCommand(cfg *config.Config) *cobra.Command {
+	var demandJSON string
+
+	var cmd = &cobra.Command{
+		Use:   "gendocs --demand <demand_json>",
+		Short: "Automatically generates documents for existing collections.",
+		Long: `Automatically generates documents for existing collections.		
+
+Example: generates 100 User documents and 500 Device documents:
+  defradb gendocs --demand '{"User": 100, "Device": 500 }'`,
+		ValidArgs: []string{"demand"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// cobra does not chain pre run calls so we have to run them again here
+			if err := loadConfig(cfg); err != nil {
+				return err
+			}
+			if err := setTransactionContext(cmd, cfg, 0); err != nil {
+				return err
+			}
+			if err := setStoreContext(cmd, cfg); err != nil {
+				return err
+			}
+			store := mustGetStoreContext(cmd)
+
+			demandMap := make(map[string]int)
+			err := json.Unmarshal([]byte(demandJSON), &demandMap)
+			if err != nil {
+				return err
+			}
+
+			collections, err := store.GetAllCollections(cmd.Context())
+			if err != nil {
+				return err
+			}
+
+			opts := []gen.Option{}
+			for colName, numDocs := range demandMap {
+				opts = append(opts, gen.WithTypeDemand(colName, numDocs))
+			}
+			docs, err := gen.AutoGenerate(colsToDefs(collections), opts...)
+			if err != nil {
+				return err
+			}
+
+			out := cmd.OutOrStdout()
+			_, err = out.Write([]byte("Generated " + strconv.Itoa(len(docs)) + " documents\n"))
+			if err != nil {
+				return err
+			}
+
+			for len(docs) > 0 {
+				thisBatch := gendocBatchSize
+				if len(docs) < gendocBatchSize {
+					thisBatch = len(docs)
+				}
+
+				colDocsMap := make(map[string][]*client.Document)
+				for _, doc := range docs[:thisBatch] {
+					colDocsMap[doc.Col.Description.Name] = append(colDocsMap[doc.Col.Description.Name], doc.Doc)
+				}
+
+				for colName, colDocs := range colDocsMap {
+					for i := range collections {
+						if collections[i].Description().Name == colName {
+							err = collections[i].CreateMany(context.Background(), colDocs)
+							if err != nil {
+								return err
+							}
+							break
+						}
+					}
+				}
+
+				docs = docs[thisBatch:]
+
+				reports := make([]string, 0, len(colDocsMap))
+				for colName, colDocs := range colDocsMap {
+					reports = append(reports, strconv.Itoa(len(colDocs))+" "+colName)
+				}
+
+				r := strings.Join(reports, ", ")
+				_, err = out.Write([]byte("Added " + strconv.Itoa(thisBatch) + " documents: " + r + "\n"))
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&demandJSON, "demand", "d", "", "Documents' demand in JSON format")
+
+	return cmd
+}
+
+func colsToDefs(cols []client.Collection) []client.CollectionDefinition {
+	var colDefs []client.CollectionDefinition
+	for _, col := range cols {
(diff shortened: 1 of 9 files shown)


cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+	"context"
+	"encoding/json"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/config"
+	"github.com/sourcenetwork/defradb/tests/gen"
```
AndrewSisley, line 23: todo: I really don't think any of our production code should ever reference anything in the `tests` directory ever.  Please move what ever you need out of the test directory.

AndrewSisley (changes requested):
Requesting changes super quick before continuing the review as the todo is a major red flag for me and I don't want this merged if approved by someone else whilst I'm still going through it.  Will continue the rest of the review now, as the todo doesn't really change anything else already in the review.

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+const gendocBatchSize = 1000
+
+func MakeGenDocCommand(cfg *config.Config) *cobra.Command {
```
AndrewSisley, line 28: todo: I am of the opinion that anything/everything in the CLI should be part of the client interface and that there should be no feature gaps between embedded-go, http, cli and any future clients. *If* we want this in the main cli tool, I think it needs to be added to the rest too, in this PR.

AndrewSisley, line 28: todo: I don't really think this should be part of the main CLI, and should instead either be in a standalone or tools-like CLI.  Could be good to chat about this in the standup.

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+	var demandJSON string
+
+	var cmd = &cobra.Command{
+		Use:   "gendocs --demand <demand_json>",
+		Short: "Automatically generates documents for existing collections.",
+		Long: `Automatically generates documents for existing collections.		
+
+Example: generates 100 User documents and 500 Device documents:
```
AndrewSisley, line 36: suggestion: `Example: The following command generates 100...` reads a bit better IMO.

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+
+				colDocsMap := make(map[string][]*client.Document)
+				for _, doc := range docs[:thisBatch] {
+					colDocsMap[doc.Col.Description.Name] = append(colDocsMap[doc.Col.Description.Name], doc.Doc)
+				}
+
+				for colName, colDocs := range colDocsMap {
+					for i := range collections {
+						if collections[i].Description().Name == colName {
```
AndrewSisley, line 91: nitpick: I think the following reads a fair bit better, there seems to be little need to trouble readers with indexes here:
```
for _, collection := range collections {
  if collection.Description()...
     err = collection.CreateMany(...
```

cli/gendocs.go
AndrewSisley, line 28: todo: This command appears to be untested.  Please write some tests for it :)

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+							err = collections[i].CreateMany(context.Background(), colDocs)
+							if err != nil {
+								return err
+							}
+							break
+						}
+					}
+				}
+
+				docs = docs[thisBatch:]
```
AndrewSisley, line 101: suggestion: I think splitting up the batching logic this much, with several blocks of other stuff in between makes for harder reading than it needs to be.  If you generate the batches first, and then process them it should be easier to follow the code.

I do have a preference for all the batches to be generated before processing them, and would find that the easiest to read and maintain, but you could also just evaluate each batch one by one in the for loop as you are doing now, just without the extra stuff done in the middle.

AndrewSisley (changes requested):
Added a few more comments as I have to context switch, will resume in a little bit.

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+			for colName, numDocs := range demandMap {
+				opts = append(opts, gen.WithTypeDemand(colName, numDocs))
+			}
+			docs, err := gen.AutoGenerate(colsToDefs(collections), opts...)
+			if err != nil {
+				return err
+			}
+
+			out := cmd.OutOrStdout()
+			_, err = out.Write([]byte("Generated " + strconv.Itoa(len(docs)) + " documents\n"))
```
AndrewSisley, line 73: suggestion:  This output feels potentially confusing, as you appear to be writing it before the documents have been written to the database.  The next output could be an error, and the docs generated disappear into the void.

I suggest either changing this line a little bit, or adding another line immediately after stating something along the lines of `Adding/saving generated documents to datastore`.

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+
+				docs = docs[thisBatch:]
+
+				reports := make([]string, 0, len(colDocsMap))
+				for colName, colDocs := range colDocsMap {
+					reports = append(reports, strconv.Itoa(len(colDocs))+" "+colName)
+				}
+
+				r := strings.Join(reports, ", ")
+				_, err = out.Write([]byte("Added " + strconv.Itoa(thisBatch) + " documents: " + r + "\n"))
```
AndrewSisley, line 109: suggestion: I think the format of the output here could be improved somewhat, as it is not super clear as to what the outputted number is (especially for low numbers where it may be seen as a collection id or something).  

I suggest clarifying each output, perhaps something like (new line for each collection) `"Added %v documents to collection of name %s"`

fredcarle:
I don't think we should have this as part of the main Defra CLI but lets discuss it during standup.

tests/gen/gen_auto_configurator.go
```diff
@@ -89,10 +89,10 @@ func (c *typeUsageCounters) getNextTypeIndForField(secondaryType string, field *
 type relationUsage struct {
 	// counter is the number of primary documents that have been used for the relation.
 	counter int
-	// minAmount is the minimum number of primary documents that should be used for the relation.
-	minAmount int
-	// maxAmount is the maximum number of primary documents that should be used for the relation.
-	maxAmount int
+	// minSecDocPerPrim is the minimum number of primary documents that should be used for the relation.
+	minSecDocPerPrim int
```
AndrewSisley, line 93: suggestion: I'm not sure the loss of characters is worth it here, and `minSecondaryDocsPerPrimary` would be worth the extra length?  Or maybe add an `s` to `Doc`: `minSecDocsPerPrim`?

AndrewSisley (changes requested):

cli/gendocs.go
```diff
@@ -0,0 +1,129 @@
...
+func MakeGenDocCommand(cfg *config.Config) *cobra.Command {
+	var demandJSON string
+
+	var cmd = &cobra.Command{
+		Use:   "gendocs --demand <demand_json>",
+		Short: "Automatically generates documents for existing collections.",
+		Long: `Automatically generates documents for existing collections.		
+
+Example: generates 100 User documents and 500 Device documents:
+  defradb gendocs --demand '{"User": 100, "Device": 500 }'`,
```
jsimnz, line 37: todo: Another point of confusion is that this example string references `defradb` as a binary and `gendocs` as a subcommand.

However this is a standalone binary.

cli/gendocs.go
islamaliev: there will be documents for different collections created.
At the moment it looks like this:
```
Generated 1500 documents
Added 1000 documents: 300 User, 700 Device
Added 500 documents: 500 Device
```

cli/gendocs.go
islamaliev: > If you generate the batches first

What do you mean? This is what I do: first generate docs (or batches), then add them to collection (or process).

Anyway I extracted some blocks into functions

AndrewSisley: On line 83 you trim the end off the main `docs` slice, then you do a bunch of stuff, then on line 99 you trim the (consumed batch) off the beginning of the `docs` slice.  By splitting these lines you make the code harder to follow, vs if they were next to each other.

It is also not really helped by the variable `thisBatch` not actually being the batch, but the length of the current batch.

If keeping the batch chunking in the loop that processes the batch (smallest adjustment, but not my personal preference), I would rename `thisBatch` to `batchLen`, then do the following:
```
batchLen := ...
thisBatch := docs[:batchLen]
// note: this mutation that you do is one of the reasons I dislike chunking in the processing loop, if keeping it
// I suggest a comment highlighting why you are doing this, just to make it super obvious
docs = docs[batchLen:] 

// process the batch
...
```

AndrewSisley (approved):
Looks really good Islam - thanks for this :)

PR #2111, opened 2023-12-05
refactor: Simplify Merkle CRDT workflow

### LINKED ISSUE #917: Write logic seems far harder to read than it needs to be
The call stack seems very deep, the code hard to find, and datastore operations are littered at many different levels in the callstack. The way multistore and factory and the crdt and clock packages constructors work also creates a significant disconnect between how we read data, and how it is written.

I believe we can improve this greatly, and do so should make it significantly easier to understand how defra works - that said, dont refactor this for the sake of refactoring, we have plenty else to do.

### LINKED ISSUE #2110: Simplify Merkle CRDT workflow.

fredcarle:
## Relevant issue(s)

Resolves #2110 
Possibly Resolves #917

## Description

This PR aims to simplify the CRDT packages ahead of the new CRDT types.

## Tasks

## How has this been tested?

make test

- MacOS


client/errors.go
```diff
@@ -47,6 +47,7 @@ var (
 	ErrMaxTxnRetries        = errors.New(errMaxTxnRetries)
 	ErrRelationOneSided     = errors.New(errRelationOneSided)
 	ErrCollectionNotFound   = errors.New(errCollectionNotFound)
+	ErrUnknownCRDT          = errors.New("unknown crdt")
```
AndrewSisley, line 50: suggestion: It might be worth including the unknowns crdt value in the error.

EDIT: I see that this is not a new error, and very happy for you to shrug this suggestion off as out of scope for now.

core/crdt/composite.go
```diff
@@ -92,39 +92,19 @@ func (delta *CompositeDAGDelta) Links() []core.DAGLink {
...
-	// schemaVersionKey is the schema version datastore key at the time of commit.
-	//
-	// It can be used to identify the collection datastructure state at time of commit.
-	schemaVersionKey core.CollectionSchemaVersionKey
-
-	fieldName string
+	baseCRDT
 }
 
-var _ core.ReplicatedData = CompositeDAG{}
```
AndrewSisley, line 105: question: Why has this been removed?

core/crdt/composite.go
```diff
@@ -92,39 +92,19 @@ func (delta *CompositeDAGDelta) Links() []core.DAGLink {
...
-
-// ID returns the schema ID of the composite DAG CRDT.
-func (c CompositeDAG) ID() string {
-	return c.key.ToString()
+	return CompositeDAG{newBaseCRDT(store, key, schemaVersionKey, fieldName)}
 }
 
-// Value returns the schema ID of the composite DAG CRDT.
+// Value is a no-op for a CompositeDAG.
 func (c CompositeDAG) Value(ctx context.Context) ([]byte, error) {
```
AndrewSisley, line 108: question: Does this function still need to exist if `CompositeDAG` no longer implements the `core.ReplicatedData` interface?

merkle/crdt/composite.go
```diff
@@ -64,25 +32,20 @@ type MerkleCompositeDAG struct {
...
 func NewMerkleCompositeDAG(
-	datastore datastore.DSReaderWriter,
-	headstore datastore.DSReaderWriter,
-	dagstore datastore.DAGStore,
+	txn datastore.Txn,
 	schemaVersionKey core.CollectionSchemaVersionKey,
-	uCh events.UpdateChannel,
-	ns,
 	key core.DataStoreKey,
 	fieldName string,
```
AndrewSisley, line 38: suggestion: It feels like (I haven't chased down the full code flow) that `fieldName` could also be removed as a parameter here?  And possibly removed from the core/crdt.`CompositeDAG` struct. On a composite commit this will always be `"C"` no?

AndrewSisley (approved):
Thank you for this Fred!  Changes look good, it should be much easier to follow now, just added a couple of extra suggestions/minor questions but am happy for this to be merged regardless of whether they are actioned.

I suggest closing #917  - the removal of the factory stuff in particular will make this much easier to follow, and we can always create a new issue if it remains a problem error instead of cluttering the backlog.

client/errors.go
fredcarle: I like that suggestion. I'll do the change 👍.

core/crdt/composite.go
fredcarle: Because it's already covered at the top of the file :)

core/crdt/composite.go
fredcarle: See above :)

merkle/crdt/composite.go
fredcarle: Yeah I though about doing that but I was trying to keep the arguments the same for both merkle crdt types to see if we could eventually simplify things even further. I'll give it some more thinking.

core/crdt/base.go
```diff
@@ -28,13 +28,26 @@ import (
 type baseCRDT struct {
 	store datastore.DSReaderWriter
 	key   core.DataStoreKey
+
+	// schemaVersionKey is the schema version datastore key at the time of commit.
+	//
+	// It can be used to identify the collection datastructure state at time of commit.
```
islamaliev, line 34: typo: "at the time"

Note: I don't care about this level or correctness, but it doesn't match the sentence above :)

shahzadlone:
Wow 👌 much cleaner. Not very well versed with this part of the code but LGTM from what I can see.

db/collection.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/events"
 	"github.com/sourcenetwork/defradb/lens"
-	"github.com/sourcenetwork/defradb/merkle/crdt"
+	merklecrdt "github.com/sourcenetwork/defradb/merkle/crdt"
```
islamaliev, line 37: question: why alias? there no other package that conflicts with it

merkle/crdt/merklecrdt.go
```diff
@@ -11,15 +11,16 @@
 /*
 Package crdt provides CRDT implementations leveraging MerkleClock.
 */
-package crdt
+package merklecrdt
```
islamaliev, line 14: suggestion: as far as I know the best practice in Go is to have package name match the folder it in.

I know that we have another `crdt` package and you renamed it here to make them more distinct. But maybe it makes sense to rename the folder as well `merkle/merklecrdt`?

Another great benefit of it is that it will automatically resolve imports.

islamaliev (approved):
LGTM!

Just added couple minor suggestions.

merkle/crdt/composite.go
AndrewSisley: Ah fair enough, don't go out of your way for this then, especially if it disrupts flow and/or you have other stuff based off of this branch.

merkle/crdt/merklecrdt.go
```diff
@@ -65,19 +61,28 @@ func (base *baseMerkleCRDT) Value(ctx context.Context) ([]byte, error) {
 	return base.crdt.Value(ctx)
 }
 
-func (base *baseMerkleCRDT) ID() string {
-	return base.crdt.ID()
-}
-
-// Publishes the delta to state.
-func (base *baseMerkleCRDT) Publish(
```
jsimnz, line 73: suggestion: I am hesitant to remove this Publish function as it keeps a cleaner seperation of concerns when writting MerkleCRDT delta-mutator functions.

The recipe is simple at the moment. Call the core crdt equivalent mutator, then call Publish(delta). 

Even if Publish is just calling `clock.AddDAGNode` under the hood, it feels like its needlessly bleeding abstractions, and lose on some of the purity from before.

merkle/crdt/composite.go
```diff
@@ -64,25 +32,20 @@ type MerkleCompositeDAG struct {
 // NewMerkleCompositeDAG creates a new instance (or loaded from DB) of a MerkleCRDT
 // backed by a CompositeDAG CRDT.
 func NewMerkleCompositeDAG(
-	datastore datastore.DSReaderWriter,
-	headstore datastore.DSReaderWriter,
-	dagstore datastore.DAGStore,
+	txn datastore.Txn,
```
jsimnz, line 35: suggestion: (This is changed in a few places, so I am just commenting here and applying the thought everywhere it has changed) I don't think we should be using `Txn` type here. Its a rather broad and overly powerful type to expose into the CRDT system, which was previously and purposefully simple in its dependencies, only specifying that it needed a datastore, headstore, and dagstore, which each of which could potentially come from different places, and only needed to implement the base `DSReaderWriter` (and `DAGStore`) interfaces.

We can probably drop the individual params as you are doing now, but it should be opened up to the entirety of the `Txn` type.

My suggestion would be to define a interface within the merklecrdt package that is a slimmed down version of the Multistore, specifically one that exposes 3 methods (`Datastore()`, `Headstore()`, and `DAGStore()`).

jsimnz (changes requested):
Gone over this twice, and will prob go over it again two more times 😂 as there is a few subtle changes I want to make sure don't cause any issues with current and future CRDT implementations.

Two current suggestions/todos noted.

db/collection.go
fredcarle: the alias is automatic when the package name is different than the folder name.

merkle/crdt/merklecrdt.go
fredcarle: It's best practice but often ignored when the folder name is not ideal. Like in cases where the folder name is `graphql-go` for example. and we just want to use `graphql`. As for `merkle/merklecrdt`, I much rather keep it as `merkle/crdt`.

net/process.go
```diff
@@ -140,24 +139,31 @@ func initCRDTForType(
...
+			txn,
+			core.NewCollectionSchemaVersionKey(col.Schema().VersionID, col.ID()),
+			key,
+			field,
+		), nil
 	}
+
+	fd, ok := col.Schema().GetField(field)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("Couldn't find field %s for doc %s", field, dsKey))
```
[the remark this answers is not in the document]
fredcarle: Will do.

merkle/crdt/merklecrdt.go
fredcarle: The way I saw it is that when we called `merkleCRDT.Set(...)` and there was a call to `Publish`, we had to go to `Publish` only to find out that it was only calling `clock.AddDAGNode` and I found it obstructive more than anything. The way it was been changed makes the flow very clear. We set the CRDT delta object and then send that object to the clock for insertion into the DAG. Can't get purer than that if you ask me.

merkle/crdt/composite.go
fredcarle: I see what you mean. I'll make the change 👍.

client/errors.go
fredcarle: ✅

merkle/crdt/merklecrdt.go
AndrewSisley: Just noting that I agree with Fred here - I find the new, less-abstract flow easier to read.

core/crdt/composite.go
fredcarle: ✅

core/crdt/base.go
fredcarle: ✅

net/process.go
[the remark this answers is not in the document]
fredcarle: actually this was already there before. I'll come back later to clean up errors that aren't extracted.

merkle/crdt/composite.go
fredcarle: ✅

jsimnz (approved):

Issue #2113, opened 2023-12-07
User may define 'reserved' fields on object but cant query them


AndrewSisley:
For example they can define:
```
type Author {
	name: String
	books: [Book]
	_count: String
}
```
but they cannot query `_count: String` as it is an aggregate system-field.  The definition of the collection should fail with a nice error explaining this.

PR #2114, opened 2023-12-08
feat: Allow users to add Views 

### LINKED ISSUE #2073: Add support for un-cached Lens-free Defra Views
Add support for un-cached Lens-free Defra Views.

Syntax discussed in https://discord.com/channels/427944769851752448/1168676792597495968

Will be:
```
CreateView(
  ctx context.Context,
  // Unrestricted in contents, this can include joins (including arbitrary joins when they come), aggregates, other views,
  // multiple-queries (when we support them), etc.
  gqlQuery string,
  // The SDL of the output type(s)
  // This can be made optional later, I think it will always be required for views using lens, and we can defer the implementation of deriving
  // this from the gqlQuery for now for non-lens views
  sdl string,
)
```
```
query {
    [ViewName] {
       [field1]
       [field2]
       [relationField1] {
          [field1]
       }
    }
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #2073

## Description

Allows users to add Views.

Defined views and embedded schema types are visible when calling GetCollections/Schema etc.

Does not provide a means to delete or update them.
Does not allow users to create documents via views.
Does not cache view results.
Does not allow users to define Lens transforms for their views.
Does not allow syncing of results over P2P.

Note: The tests are a bit light for this, covering the very basics only.   I would like to add some more complex testing in another PR - as views are as complex as regular queries, I think it would be best to add some magic to the test framework to auto test all normal queries via auto-generated views (a new test-matrix dimension, only needs to run against mem-store, one OS, save mutation type, etc), otherwise to properly test we'll end up duplicating most of our integration test suite.  We'll still want some dedicated view tests, for view-specific stuff, those I have added, as well as a few important ones (such as basic filter and aggregate tests) that could be ripped out once that test magic is in place.


planner/mapper/mapper.go
```diff
@@ -738,25 +733,38 @@ func getTopLevelInfo(
...
+			schemas, err := store.GetSchemasByName(ctx, collectionName)
+			if err != nil {
+				return nil, client.SchemaDescription{}, err
+			}
+			if len(schemas) == 0 {
+				return nil, client.SchemaDescription{}, NewErrTypeNotFound(collectionName)
+			}
+			// `schemas` will contain all versions of that name, as views cannot be updated atm this should
+			// be fine for now
+			schema = schemas[0]
```
fredcarle, line 760: question: Should we try to ensure it's the latest version of the schema?

request/graphql/schema/generate.go
```diff
@@ -98,8 +98,24 @@ func (g *Generator) generate(ctx context.Context, collections []client.Collectio
 		if err != nil {
 			return nil, err
 		}
-		queryType.AddFieldConfig(f.Name, f)
 		generatedQueryFields = append(generatedQueryFields, f)
+
+		var isEmbedded bool
+		for _, definition := range collections {
+			if t.Name() == definition.Schema.Name && definition.Description.Name == "" {
+				isEmbedded = true
```
fredcarle, line 106: question: You rather use this method than to use `continue` with a loop tag? The block content is short enough that if would still read easily. Fine either way. Just curious.

planner/mapper/mapper.go
AndrewSisley: I thought about that, but it felt like too much code right now.  If it was really cheap to do so I would probably have added it in, but it is ultimately dead-code atm given that there can only be one version. 

Better to wait until we actually implement something that makes this necessary (and testable), and then code it properly.

request/graphql/schema/generate.go
AndrewSisley: I see loop tags as a bit unusual and my brain has not get chosen to either ban them or always use them.  Sometimes I use them sometimes I don't, *I think* I mostly only use them in tiny (~10 line or less) functions where it is hard for them to get lost.

tests/integration/view/one_to_many/with_count_test.go
```diff
@@ -0,0 +1,156 @@
...
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+							AuthorView {
+								name
+								_count
+							}
+						}`,
+				ExpectedError: "aggregate must be provided with a property to aggregate",
```
islamaliev, line 79: question: I don't quite follow how it works. If the `_count` is specified in SDL why can't it be used here?
If you can't override a reserved property why doesn't it throw an error in`CreateView`?

tests/integration/view/one_to_many/with_count_test.go
```diff
@@ -0,0 +1,156 @@
...
+					"author_id": "bae-ef9cd756-08e1-5f23-abeb-7b3e6351a68d"
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+							AuthorView {
+								name
+								numberOfBooks
+							}
+						}`,
```
islamaliev, line 144: nitpick: I've seen a lot in the tests that indentation is correct for raw strings. Would be nice to fix them (for other occurrences as well). I don't think lint will report them.

islamaliev (commented):
Looks good so far. Just have few minor comments

client/request/select.go
```diff
@@ -107,3 +109,88 @@ func (s *Select) validateGroupBy() []error {
...
+	Offset      immutable.Option[uint64]
+	OrderBy     immutable.Option[OrderBy]
+	GroupBy     immutable.Option[GroupBy]
+	Filter      immutable.Option[Filter]
+	ShowDeleted bool
+
+	// Properties above this line match the `Select` object and
+	// are deserialized using the normal/default logic.
+	// Properties below this line require custom logic in `UnmarshalJSON`
+	// in order to be deserialized correctly.
```
shahzadlone, line 130: suggestion: Although I like this comment, I find that overtime this might be hard to enforce / maintain. Perhaps in addition to this add a different prefix to variable names to the ones that require custom logic? i.e. `CustomFields` instead of `Fields`

planner/view.go
shahzadlone, line 1: todo: Since a new "viewNode" is being added to the planner:

1) Must add it to the `allPlanNodeNames` variable in the `tests/integration/explain.go` file for debug node, which dumps the entire plan node graph (including non-explainable nodes).

2) Following up on (1) please do add a test for the debug explain under the folder `tests/integration/explain/debug`.

3) discussion/question: Do we need to make `viewNode` explainable? any attributes that might be important to view for the simple explain (non-debug explain) @jsimnz

request/graphql/schema/generate.go
```diff
@@ -365,14 +409,23 @@ func (g *Generator) buildTypes(
 		// will be reassigned before the thunk is run
 		collection := c
 		fieldDescriptions := collection.Schema.Fields
+		isEmbeddedObject := collection.Description.Name == ""
```
shahzadlone, line 412: suggestion: A brief note / concrete example here describing what is meant by an "embedded object" here would be nice.

cli/errors.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrNoLensConfig             = errors.New("lens config cannot be empty")
 	ErrInvalidLensConfig        = errors.New("invalid lens configuration")
 	ErrSchemaVersionNotOfSchema = errors.New(errSchemaVersionNotOfSchema)
+	ErrViewAddMissingArgs       = errors.New("please provide a base query and output SDL")
```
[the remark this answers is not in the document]
AndrewSisley: Not a problem, will add :)

- [ ] Tweak error text

client/db.go
```diff
@@ -123,6 +123,34 @@ type Store interface {
...
+	//
+	// Author {
+	//	 name
+	//	 books {
+	//	   name
+	//	 }
+	// }
+	//
+	//
+	// A GQL SDL that matches its output type must also be provided.  There can only be one `type` declaration,
```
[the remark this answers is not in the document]
AndrewSisley: > nitpick: 2 spaces

:laughing: https://www.independent.co.uk/tech/one-space-or-two-spaces-after-a-full-stop-scientists-have-finally-found-the-answer-a8337646.html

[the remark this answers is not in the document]
AndrewSisley: - [ ] schema-only

client/descriptions.go
```diff
@@ -30,6 +32,13 @@ type CollectionDescription struct {
 	// The ID of the schema version that this collection is at.
 	SchemaVersionID string
 
+	// BaseQuery contains the base query of this view, if this is collection is a view.
```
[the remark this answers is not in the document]
AndrewSisley: Cheers :)

- [ ] "is this collection is" typo

> Should it be "Actor-defined"?

I don't think so, why?

db/view.go
```diff
@@ -0,0 +1,100 @@
...
+
+	baseQuery, ok := req.Queries[0].Selections[0].(*request.Select)
+	if !ok {
+		return nil, NewErrInvalidViewQuery(inputQuery)
+	}
+
+	for i, definition := range newDefinitions {
+		definition.Description.BaseQuery = baseQuery
+		newDefinitions[i] = definition
+	}
```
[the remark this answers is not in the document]
fredcarle: +1

client/request/select.go
```diff
@@ -107,3 +109,88 @@ func (s *Select) validateGroupBy() []error {
...
+		fieldJson, err := json.Marshal(field)
+		if err != nil {
+			return err
+		}
+
+		var fieldValue Selection
+		// We detect which concrete type each `Selection` object is by detecting
+		// non-nillable fields, if the key is present it must be of that type.
+		// They must be non-nillable as nil values may have their keys omitted from
+		// the json. This als relies on the fields being unique.  We may wish to change
```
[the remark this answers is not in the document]
AndrewSisley: Cheers lol

- [ ] fix (another) typo in doc

db/view.go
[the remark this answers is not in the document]
AndrewSisley: lol, any reason why you see this as a suggestion and not a personal preference nitpick?

db/view.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	existingDefinitions := make([]client.CollectionDefinition, len(existingCollections))
+	for i := range existingCollections {
+		existingDefinitions[i] = existingCollections[i].Definition()
+	}
+
+	err = db.parser.SetSchema(ctx, txn, append(existingDefinitions, newDefinitions...))
```
fredcarle, line 74: thought: It would be much nicer to have a `AddSchema` method on parser instead of having to query existing schemas to add a new one.

http/handler_store.go
```diff
@@ -106,6 +106,25 @@ func (s *storeHandler) SetDefaultSchemaVersion(rw http.ResponseWriter, req *http
 	rw.WriteHeader(http.StatusOK)
 }
 
+func (s *storeHandler) CreateView(rw http.ResponseWriter, req *http.Request) {
```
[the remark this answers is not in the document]
AndrewSisley: Everything was called `CreateView` when I originally wrote it, then I realized `AddView` fitted much more nicely with out existing funcs and I had to rename everything - I must have missed this function :)

- [ ] Rename to `AddView`

db/view.go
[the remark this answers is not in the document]
fredcarle: For me it's the unnecessary allocation and that it makes it harder to read what's happening. It's just one extra line but you have to pause to think for a second what is happening instead of just seeing that you set something for every new definition.

planner/mapper/mapper.go
```diff
@@ -282,11 +281,7 @@ func resolveAggregates(
 			var hasHost bool
 			var convertedFilter *Filter
 			if childIsMapped {
-				var fieldDesc client.FieldDescription
-				var isField bool
-				if collection != nil {
-					fieldDesc, isField = collection.Schema().GetField(target.hostExternalName)
-				}
+				fieldDesc, isField := schema.GetField(target.hostExternalName)
```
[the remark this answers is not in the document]
fredcarle: I don't think it's necessary since `GetField` will simply return `FieldDescription{}, false` if lenght is 0. The result would be the same.

planner/mapper/mapper.go
```diff
@@ -738,25 +733,38 @@ func getTopLevelInfo(
...
 	if selectRequest.Root == request.ObjectSelection {
-		mapping.Add(core.DocKeyFieldIndex, request.KeyFieldName)
-
+		var schema client.SchemaDescription
 		collection, err := store.GetCollectionByName(ctx, collectionName)
 		if err != nil {
-			return nil, nil, err
+			// If the collection is not found, check to see if a schema of that name exists,
+			// if so, this must be an embedded object
+			schemas, err := store.GetSchemasByName(ctx, collectionName)
```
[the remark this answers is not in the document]
AndrewSisley: Yes, I couldn't actually find an easy, reliable way to check this (the current error returned is the horrible 'datastore not found error', which I do not consider a reliable).  I figured not checking the error at all was the safest atm.

I'll re-visit this, but do you have any off-the-top-of-your-head suggestions in the meantime?

- [ ] Revisit mapper error check stuff

request/graphql/schema/collection.go
```diff
@@ -430,6 +471,13 @@ func getRelationshipName(
 }
 
 func finalizeRelations(relationManager *RelationManager, definitions []client.CollectionDefinition) error {
+	embeddedObjNames := map[string]any{}
```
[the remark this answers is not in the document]
AndrewSisley: Cheers, don't know why I went for `any` here lol

- [ ] struct{}

tests/integration/view/one_to_many/with_count_test.go
AndrewSisley: This is https://github.com/sourcenetwork/defradb/issues/2113 linked in the description of the test func doc - `_count` is not protected by us and is half-overrideable.

tests/integration/view/one_to_many/with_count_test.go
AndrewSisley: Ah true, my bad, I'll try and fix the ones in this PR at least for now.  Agree a linter would be nice long term.

- [ ] Fix test query indentation

client/request/select.go
AndrewSisley: The current solution for sure has it's flaws, but I'm not sure I like that suggestion.  I'll add a task to re-visit and see how I feel about it again in a little bit :)

- [ ] Revisit this suggestion

planner/view.go
AndrewSisley: (3) Can be done in another PR, I can't see any reason for adding it to this.

question:  Why do you think (1) and (2) need to be done in this PR?  The tests pass atm, is production explain broken?  Is there a gap in explain integration tests?

request/graphql/schema/generate.go
AndrewSisley: I do see `embedded` object as a ubiquitous database/defra concept that needs no explanation against every variable declaration that mentions it.  I'd be quite against doing this.

db/view.go
AndrewSisley: Food for thought, but I'm not sure it is worth the hassle that would mean in the parser/gql code.  That hassle would also increase the cost of adding any new query languages who would need to re-write their own flavour of that hassle in order to satisfy the interface.  Performance isn't a concern in these functions.

fredcarle: At the moment it's a hassle that is needed every time there is a call to `SetSchema`. Don't you think it would be better to have that in just one location instead?

AndrewSisley: Not every time, the existing are required when updating schema (and in the future, views).  I also see replacing everything with a new, clean instance as safer than mutating them.

AndrewSisley: > Don't you think it would be better to have that in just one location instead?

We can do this without hosting it inside the Parser interface.

db/view.go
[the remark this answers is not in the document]
AndrewSisley: Fair enough :)  Will change.

- [ ] simplify loop

planner/mapper/mapper.go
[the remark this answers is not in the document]
AndrewSisley: I've added a todo and a new issue: https://github.com/sourcenetwork/defradb/issues/2146

client/request/select.go
AndrewSisley: I still don't like the suggestion, and will leave as-is unless anyone continues this thread :)

planner/view.go
shahzadlone: Because Debug explain needs to be aware of all nodes in the plan graph, otherwise it likely skips the node (it needs to know all names of the valid plan nodes).

shahzadlone: 3) is not required for this PR

AndrewSisley: > otherwise *it* likely skips the node

'It' being the production code or the test code?

AndrewSisley: > 'It' being the production code or the test code?

Reading the code referenced, I assume the test code.  The test you asked for in (2), this would in effect just be testing that `viewNode.Kind()` returns `"viewNode"` no?

shahzadlone: Yes for (1) I meant that it would break the testing framework, as it needs to know what is and isn't a valid node since we can target specific nodes, all you need to do is add it to the list, it shouldn't break or fail anything.

(2) Yea once you test with `@explain(type: debug)` on any of your view queries, all it would show is that "viewNode" is properly being populated and in the correct structure that you expect it to be in the full build graph.

tests/integration/explain/debug/with_view_test.go
shahzadlone, line 1: praise: Thanks for adding this! LGTM :)

planner/view.go
AndrewSisley: Nice, cheers Shahzad, sorry for the questions - is maybe the first time I'm adding an explain test and I want to make sure I understand it all :)

fredcarle (approved):
LGTM 🙂

shahzadlone (approved):
Thanks for adding the Debug Explain tests. LGTM

client/descriptions.go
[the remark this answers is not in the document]
islamaliev: > I don't think so, why do you think so?

Otherwise it the sentence sound unnatural (at least to my non-native ears). Why then you used past tense "Actor defined"?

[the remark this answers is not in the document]
AndrewSisley: > Why then you used past tense "Actor defined"?

They are aggregates that have been (past tense), by an actor.  Both with and without hyphen made sense to me, and I think I over-use special characters when writing, and so I try and avoid them.

Grammarly suggests using hyphens like you suggested here: https://www.grammarly.com/blog/hyphen/ though.  I'm not sure it is a hard grammar rule in English though, Collins (respected author of dictionaries) keeps their usage simple: https://grammar.collinsdictionary.com/easy-learning/when-do-you-use-a-hyphen-in-english.

db/view.go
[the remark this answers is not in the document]
AndrewSisley: > "you wrote shit, take this beauty instead"

:rofl: 

Generally, if I feel like something is just personal preference, and that I might be suggesting it just because I am more used to something, I'll mark it as a nitpick, and I actively try and limit the number of nitpicks I raise - for example IIRC in your PRs you seem to prefer setting the capacity with `make([]foo{}, 0, x)` and then using `append`, I find that a bit jarring to read and think of it as more error prone because it is more unusual, although there is nothing inherently wrong about it.

Todo's are anything I'd block a merge over (although can/should be argued against be the author if they disagree). Suggestions are everything else :)

client/descriptions.go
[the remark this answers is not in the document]
islamaliev: good arguments, thanks. I'm not trying to say that it's wrong or not. Just giving you my perception.
My judgement based mostly on my own understanding. If I had to read it few time to make sure I understand, then some others will do as well.

islamaliev:
Bug bush issue: https://github.com/sourcenetwork/defradb/issues/2182

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2183

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2189

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2190

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2191

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2201

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2206

islamaliev:
But bash issue: https://github.com/sourcenetwork/defradb/issues/2212

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2218

islamaliev:
Bug bash issue: https://github.com/sourcenetwork/defradb/issues/2219

PR #2130, opened 2023-12-11
feat: Add Big Int and Big Float scalars

### LINKED ISSUE #2129: Big number scalar support
Adding support for big number scalars will give developers more choice when it comes to number storage types.

nasdf:
## Relevant issue(s)

Resolves #2129 

## Description

This PR adds two new scalar types: `BigInt` and `BigFloat`.

Aggregate support was intentionally left out so we can figure it out in a future PR.

## Tasks

## How has this been tested?

Added tests

- MacOS

### DIFF
--- a/request/graphql/schema/types/scalars.go
+++ b/request/graphql/schema/types/scalars.go
@@ -12,6 +12,7 @@ package types
 
 import (
 	"encoding/hex"
+	"math/big"
 	"regexp"
 
 	"github.com/sourcenetwork/graphql-go"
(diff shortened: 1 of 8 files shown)


AndrewSisley:
todo: Please add some integration tests for:

- Create and Update doc actions (covering all mutation types)
- Filter
- Sort
- GroupBy
- Sum and Average aggregates (including top-level) (preferably including introspection tests for objects containing only a Big field, example below)

```
Author {
  name
  book
}
Book {
   // It is important to test with an object where this is the only sumable field on it to make sure the gql type is added (or not)
   pages BigInt
}

query {
  _sum(book:pages)
}
query {
  author{
    pagesWritten: _sum(book:pages)
  }
}
```

request/graphql/schema/types/scalars.go
```diff
@@ -63,3 +64,211 @@ var BlobScalarType = graphql.NewScalar(graphql.ScalarConfig{
 		}
 	},
 })
+
+// BigIntPattern is a regex for validating big int strings
+var BigIntPattern = regexp.MustCompile("^[0-9]+$")
+
+var BigIntScalarType = graphql.NewScalar(graphql.ScalarConfig{
+	Name:        "BigInt",
+	Description: "The `BigInt` scalar type represents an arbitrary precision integer.",
```
AndrewSisley, line 73: todo: It is unclear as to what you mean by `arbitrary precision` here, from scanning the code/tests it looks like the user has no way of setting the precision, and that internally everything is just int64?  Please clarify this, same goes for big float.

Please also (PR comment is fine) explain why you chose the name `BigFoo` instead of Int64 or other alternatives.

nasdf: The internal storage for BigInt is a string matching the pattern `^[0-9]+$`. If the number is larger than an `int64` or `uint64` it can be set using a string of the same format.

AndrewSisley: :sweat_smile: Please document that in the description.

That also sounds horribly inefficient.  Why are you doing it like that?

AndrewSisley: Discussed in standup, I'll write my understanding here in case I am unavailable later:

todo: store both types as byte arrays, not strings.

---------------------------

Not discussed in standup, but tidied up a bit from my prior ramblings/thoughts:

todo:  Please don't use the phrase `arbitrary precision` for int/decimal, it only makes sense for float.

suggestion: BigDecimal should also be added (another PR most likely).

suggestion: Break up this PR into one for Int and one for float, but as it is already together I won't push back at all if you don't want to.

nasdf:
Will revisit when I have a better idea of features.

PR #2131, opened 2023-12-12
feat: Add unique secondary index

### LINKED ISSUE #298: Sec. Indexes: Implement unique indexes

islamaliev:
## Relevant issue(s)

Resolves #298 

## Description

Introduces unique secondary indexes.

## Tasks

## How has this been tested?

Unit and integration tests

- MacOS


db/fetcher/indexer_iterators.go
```diff
@@ -45,45 +45,52 @@ const (
...
+	value    []byte
+	err      error
+}
+
 type queryResultIterator struct {
 	resultIter query.Results
 }
 
-func (i queryResultIterator) Next() (core.IndexDataStoreKey, bool, error) {
+func (i *queryResultIterator) Next() indexIterResult {
```
fredcarle, line 63: suggestion: I would leave the error as a seperate return value. This makes it more obvious to the caller that the error needs to be checked.

db/fetcher/indexer_iterators.go
```diff
@@ -92,16 +99,62 @@ func (i *eqIndexIterator) Init(ctx context.Context, store datastore.DSReaderWrit
 	return nil
 }
 
-func (i *eqIndexIterator) Next() (core.IndexDataStoreKey, bool, error) {
-	key, hasValue, err := i.queryResultIterator.Next()
-	if hasValue {
+func (i *eqPrefixIndexIterator) Next() indexIterResult {
```
fredcarle, line 102: suggestion: Same as above. It would be nicer if ther error was kept separate.

db/collection_index.go
```diff
@@ -222,7 +223,8 @@ func (c *collection) createIndex(
 	c.indexes = append(c.indexes, colIndex)
 	err = c.indexExistingDocs(ctx, txn, colIndex)
 	if err != nil {
-		return nil, err
+		removeErr := colIndex.RemoveAll(ctx, txn)
```
fredcarle, line 226: question: Since this operation is happening on a transaction, why do we have to bother removing the indexed documents if it fails? Can we not rely on the discarding of the transaction?

db/fetcher/indexer_iterators.go
islamaliev: I followed what seems to be a common way of dealing with iterators. Like `query.Result`.
I think the main advantage of it (besides having shorter returns) it will be easier to use such API with go channels.

db/collection_index.go
islamaliev: I thought about it myself, but wasn't sure if it would be correct to discard the whole transaction. After all, it theoretically might contain several operations (not only index creation) it should be up to the caller's site to decide what to do.
On the other hand, calling `colIndex.RemoveAll` is very localized and the collection has all necessary knowledge to make this decision. 
Maybe I'm missing something.
Note: here is the test that tests it https://github.com/sourcenetwork/defradb/pull/2131/files#diff-0a00fc4743373305afbef72902d95cdc5a9ddbda32c694a0c97c85d03efcde46R1034

db/fetcher/indexer_iterators.go
fredcarle: Do you think we will move to go channels for that API? I'm not sure a channel makes sense here.

fredcarle (approved):
Overall it look good. I'm approving assuming we will resolve the 2 conversations before merging.

Issue #2175, opened 2024-01-08
Updating a doc to violate unique index results in 'create' error


AndrewSisley:
When updating a document so that it violates a unique index, the following error message is returned `can not create doc that violates unique index` - it says cannot **create** doc, when an update is being updated. 

E.g.:
```
                        testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String 
						age: Int
					}
				`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"John",
						"age":	21
					}`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Andy",
						"age":	22
					}`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Shahzad",
						"age":	23
					}`,
			},
			testUtils.CreateIndex{
				CollectionID: 0,
				FieldName:    "age",
				Unique:       true,
			},
			testUtils.UpdateDoc{
				DocID: 2,
				Doc: `{
					"age": 22
				}`,
			},
```

PR #2178, opened 2024-01-09
fix: Handle multiple nil values on unique indexed fields

### LINKED ISSUE #2174: Violating a unique index with a nil value returns unhelpful error, or panics
For example, the below will return a `The given field does not exist. Name: age` error instead of an index violation error when attempting to create `Shahzad`.  Worse, if the index is created after all three documents (instead of before `Shahzad`), the code panics with `Unclosed iterator at time of Txn.Discard`:

Error:
```
                       testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String 
						age: Int
					}
				`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"John",
						"age":	21
					}`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Andy"
					}`,
			},
			testUtils.CreateIndex{
				CollectionID: 0,
				FieldName:    "age",
				Unique:       true,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Shahzad"
					}`,
			},
```
Panic:
```
                       testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String 
						age: Int
					}
				`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"John",
						"age":	21
					}`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Andy"
					}`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Shahzad"
					}`,
			},
			testUtils.CreateIndex{
				CollectionID: 0,
				FieldName:    "age",
				Unique:       true,
			},
```

islamaliev:
## Relevant issue(s)

Resolves #2174 #2175

## Description

Forbid `nil` values for fields that are unique-indexed.

## Tasks

## How has this been tested?

Integration and unit tests

- MacOS

### DIFF
--- a/db/errors.go
+++ b/db/errors.go
@@ -86,7 +86,8 @@ const (
 	errExpectedJSONArray                  string = "expected JSON array"
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
-	errCanNotIndexNonUniqueField          string = "can not create doc that violates unique index"
+	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
+	errCanNotIndexNilField                string = "can not index a doc's field with nil value"
 	errInvalidViewQuery                   string = "the query provided is not valid as a View"
 )
 
@@ -572,6 +573,14 @@ func NewErrCanNotIndexNonUniqueField(docID, fieldName string, value any) error {
 		errors.NewKV("DocID", docID),
 		errors.NewKV("Field name", fieldName),
 		errors.NewKV("Field value", value),
+	)
+}
+
+func NewErrCanNotIndexNilField(docID, fieldName string) error {
+	return errors.New(
+		errCanNotIndexNilField,
+		errors.NewKV("DocID", docID),
+		errors.NewKV("Field name", fieldName),
 	)
 }
 
--- a/db/errors.go
+++ b/db/errors.go
@@ -87,7 +87,6 @@ const (
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
 	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
-	errCanNotIndexNilField                string = "can not index a doc's field with nil value"
 	errInvalidViewQuery                   string = "the query provided is not valid as a View"
 )
 
@@ -576,19 +575,11 @@ func NewErrCanNotIndexNonUniqueField(docID, fieldName string, value any) error {
 	)
 }
 
-func NewErrCanNotIndexNilField(docID, fieldName string) error {
-	return errors.New(
-		errCanNotIndexNilField,
-		errors.NewKV("DocID", docID),
-		errors.NewKV("Field name", fieldName),
-	)
-}
-
 func NewErrInvalidViewQueryCastFailed(query string) error {
 	return errors.New(
 		errInvalidViewQuery,
 		errors.NewKV("Query", query),
-		errors.NewKV("Reason", "Internal errror, cast failed"),
+		errors.NewKV("Reason", "Internal error, cast failed"),
 	)
 }
 
--- a/db/index.go
+++ b/db/index.go
@@ -102,6 +102,7 @@ func NewCollectionIndex(
 		return nil, err
 	}
 	if desc.Unique {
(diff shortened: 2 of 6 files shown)


AndrewSisley:
This does feel like the wrong approach.   I think users should still be allowed nil values in indexed fields, especially considering that the GQL types *are* nillable (`Int` vs `INT!`).

The SQL standards for this allow multiple nil values in a column with a unique constraint.  Microsoft SQL Server naturally ignores this, and only permits a single nil value.

https://stackoverflow.com/questions/767657/how-do-i-create-a-unique-constraint-that-also-allows-nulls (check top answer).

AndrewSisley (changes requested):
This does feel like the wrong approach.   I think users should still be allowed nil values in indexed fields, especially considering that the GQL types *are* nillable (`Int` vs `Int!`).

The SQL standards for this allow multiple nil values in a column with a unique constraint.  Microsoft SQL Server naturally ignores this, and only permits a single nil value.

https://stackoverflow.com/questions/767657/how-do-i-create-a-unique-constraint-that-also-allows-nulls (check top answer).

tests/integration/index/create_unique_test.go
```diff
@@ -186,3 +186,90 @@ func TestUniqueIndexCreate_IfFieldValuesAreUnique_Succeed(t *testing.T) {
...
+				Unique:        true,
+				ExpectedError: db.NewErrCanNotIndexNonUniqueField("bae-caba9876-89aa-5bcf-bc1c-387a52499b27", "age", nil).Error(),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestUniqueIndexCreate_UponAddingDocWithExistingNilValue_ReturnError(t *testing.T) {
```
AndrewSisley, line 236: todo: Can you please add a test that shows a single document with a nil indexed value succeeding?  This way if this (multiple nil values) test is deleted (e.g. when switching to allowing multiple nils) we do not lose coverage of the single nil happy-case.

todo: Can you please add a test with a filter eq nil?

todo: Can you please add some tests for composite indexes with nil and non-nil values?

db/index.go
```diff
@@ -290,10 +290,16 @@ func (i *collectionUniqueIndex) newUniqueIndexError(
 	doc *client.Document,
 ) error {
 	fieldVal, err := doc.GetValue(i.fieldDesc.Name)
+	var val any
 	if err != nil {
-		return err
+		if !errors.Is(err, client.ErrFieldNotExist) {
```
AndrewSisley, line 295: todo: Can you please document why this if exists, it will not be obvious when reading the code at a later date.

db/errors.go
```diff
@@ -86,7 +86,7 @@ const (
 	errExpectedJSONArray                  string = "expected JSON array"
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
-	errCanNotIndexNonUniqueField          string = "can not create doc that violates unique index"
+	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
```
AndrewSisley, line 89: praise: Thanks for solving this one :)

AndrewSisley (changes requested):
Looks good, and I'm glad not much production code was required.  I do think more tests are required though.

tests/integration/index/create_unique_test.go
AndrewSisley: Could you please add it as an iteration test please?  It will make it much easier to discover, and it will be able to take advantage of all our integration test related perks (such as mutation and client types, change detection, etc)

AndrewSisley: > the current test already includes a successful addition of a document with a nil value:

Yes I saw that, but an independent successful test would be useful. If this (multiple nil values) test is deleted (e.g. when switching to allowing multiple nils) we do not lose coverage of the single nil happy-case.

islamaliev: added an explicit test

AndrewSisley (approved):
LGTM :)

Issue #2182, opened 2024-01-10
Schema View: 2 views with the same interface override each other


islamaliev:
When creating 2 view that have interfaces with the same name, then one overrides another and the overridden one is not available in the first view anymore.
```go
func TestView_Override(t *testing.T) {
	test := testUtils.TestCase{
		Description: "One to many view with alias on outer object",
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type Author {
						name: String
						books: [Book]
					}
					type Book {
						name: String
						author: Author
					}
				`,
			},
			testUtils.CreateView{
				Query: `
					Author {
						name
						books {
							name
						}
					}
				`,
				SDL: `
					type AuthorView {
						name: String
						books: [BookView]
					}
					interface BookView {
						name: String
					}
				`,
			},
			testUtils.CreateView{
				Query: `
					Author {
						authorName: name
						books {
							bookName: name
						}
					}
				`,
				SDL: `
					type AuthorAliasView {
						authorName: String
						books: [BookView]
					}
					interface BookView {
						bookName: String
					}
				`,
			},
			// bae-ef9cd756-08e1-5f23-abeb-7b3e6351a68d
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `{
					"name":	"Harper Lee"
				}`,
			},
			testUtils.CreateDoc{
				CollectionID: 1,
				Doc: `{
					"name":	"To Kill a Mockingbird",
					"author_id": "bae-ef9cd756-08e1-5f23-abeb-7b3e6351a68d"
				}`,
			},
			testUtils.Request{
				Request: `query {
							AuthorAliasView {
								authorName
								books {
									bookName
								}
							}
						}`,
				Results: []map[string]any{
					{
						"authorName": "Harper Lee",
						"books": []map[string]any{
							{
								"bookName": "To Kill a Mockingbird",
							},
						},
					},
				},
			},
			// This action results in error: Cannot query field "name" on type "BookView".
			/*testUtils.Request{
				Request: `query {
							AuthorView {
								name
								books {
									name
								}
							}
						}`,
				Results: []map[string]any{
					{
						"name": "Harper Lee",
						"books": []map[string]any{
							{
								"name": "To Kill a Mockingbird",
							},
						},
					},
				},
			},*/
			// This action returns books with "name" field although "bookName" is requested
			testUtils.Request{
				Request: `query {
							AuthorView {
								name
								books {
									bookName
								}
							}
						}`,
[...]

PR #2185, opened 2024-01-10
fix(i): Retain embedded schema within gql types on reload

### LINKED ISSUE #2184: Schema used in Views disappear when later updating the GQL types
As inner `interface` types are collection-less, they are not reloaded when further modifying the types.

For example the below will error as `BookView` is 'forgotten' when adding the `User` collection to the GQL types:
```
			testUtils.SchemaUpdate{
				Schema: `
					type Author {
						name: String
						books: [Book]
					}
					type Book {
						name: String
						author: Author
					}
				`,
			},
			testUtils.CreateView{
				Query: `
					Author {
						name
						books {
							name
						}
					}
				`,
				SDL: `
					type AuthorView {
						name: String
						books: [BookView]
					}
					interface BookView {
						name: String
					}
				`,
			},
			testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String
					}
				`,
			},
```

AndrewSisley:
## Relevant issue(s)

Resolves #2184 #2182

## Description

Retain embedded schema within gql types on reload and further GQL modification.

Also refactors the code a little to make it harder to re-introduce (calling `loadSchema` instead of sometimes relying on function vars).

Appears to accidentally solve #2182
Todo: 
- [x] Add proper tests for #2182

### DIFF
--- a/tests/integration/schema/simple_test.go
+++ b/tests/integration/schema/simple_test.go
@@ -78,7 +78,7 @@ func TestSchemaSimpleErrorsGivenDuplicateSchema(t *testing.T) {
 				Schema: `
 					type Users {}
 				`,
-				ExpectedError: "schema type already exists",
+				ExpectedError: "collection already exists",
 			},
 		},
 	}
@@ -94,7 +94,7 @@ func TestSchemaSimpleErrorsGivenDuplicateSchemaInSameSDL(t *testing.T) {
 					type Users {}
 					type Users {}
 				`,
-				ExpectedError: "schema type already exists",
+				ExpectedError: "collection already exists",
 			},
 		},
 	}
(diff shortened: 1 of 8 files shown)


tests/integration/schema/simple_test.go
```diff
@@ -78,7 +78,7 @@ func TestSchemaSimpleErrorsGivenDuplicateSchema(t *testing.T) {
 				Schema: `
 					type Users {}
 				`,
-				ExpectedError: "schema type already exists",
+				ExpectedError: "collection already exists",
```
AndrewSisley, line 81: note: These changed as the order in which `loadSchema` was called changed slightly.  I think the new error is more correct anyway.

fredcarle (approved):
LGTM

PR #2186, opened 2024-01-10
fix: Add entropy to counter CRDT type updates

### LINKED ISSUE #2179: Concurrent PNCounter is unpredictable over P2P
In a peer network of two nodes, where two concurrent updates are made to each node with the same value, sometimes the end result will be the summation/doubling of the update, and sometimes the result will be a single update.

For example in the below, the result will sometimes be `45`, and sometimes it will be `90`. `90` Appears to occur less frequently than `45`.
```
			testUtils.SchemaUpdate{
				Schema: `
					type Users {
						Name: String
						Age: Int @crdt(type: "pncounter")
					}
				`,
			},
			testUtils.CreateDoc{
				// Create John on all nodes
				Doc: `{
					"Name": "John",
					"Age": 0
				}`,
			},
			testUtils.ConnectPeers{
				SourceNodeID: 0,
				TargetNodeID: 1,
			},
			testUtils.UpdateDoc{
				NodeID: immutable.Some(0),
				Doc: `{
					"Age": 45
				}`,
			},
			testUtils.UpdateDoc{
				NodeID: immutable.Some(1),
				Doc: `{
					"Age": 45
				}`,
			},
			testUtils.WaitForSync{},
			testUtils.Request{
				Request: `query {
					Users {
						Age
					}
				}`,
			},
```

fredcarle:
## Relevant issue(s)

Resolves #2179 

## Description

Counters need to be incrementable independently from different nodes. Initially, if two nodes incremented by the same amount from the same synced state, then the update CID would be the same and thus taken as a the same operation when syncing. So starting from 10 if both node A and B increment by 5 simultaneously, because they would generate the same CID for the update, the end result would be 15. We want it to be 20.

This PR adds entropy to the PN Counter delta so that a situation like described above would result in 20.

Note that the entropy is only added on update so that a document can be instantiated on multiple nodes with exactly the same CID.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/core/crdt/pncounter.go
+++ b/core/crdt/pncounter.go
@@ -13,6 +13,9 @@ package crdt
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"math"
+	"math/big"
 
 	"github.com/fxamacker/cbor/v2"
 	dag "github.com/ipfs/boxo/ipld/merkledag"
@@ -45,6 +48,9 @@ type PNCounterDelta[T Incrementable] struct {
 	DocID     []byte
 	FieldName string
 	Priority  uint64
+	// Entropy is an added randomly generated number that ensures
+	// that each increment operation is unique.
+	Entropy int64
 	// SchemaVersionID is the schema version datastore key at the time of commit.
 	//
 	// It can be used to identify the collection datastructure state at the time of commit.
@@ -108,13 +114,30 @@ func (reg PNCounter[T]) Value(ctx context.Context) ([]byte, error) {
 }
 
 // Set generates a new delta with the supplied value
-func (reg PNCounter[T]) Increment(value T) *PNCounterDelta[T] {
+func (reg PNCounter[T]) Increment(ctx context.Context, value T) (*PNCounterDelta[T], error) {
+	// To ensure that the dag block is unique, we add a random number to the delta.
+	// This is done only on update (if the doc doesn't already exist) to ensure that the
+	// initial dag block of a document can be reproducible.
+	exists, err := reg.store.Has(ctx, reg.key.ToPrimaryDataStoreKey().ToDS())
+	if err != nil {
+		return nil, err
+	}
+	var entropy int64
+	if exists {
+		r, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+		if err != nil {
+			return nil, err
+		}
+		entropy = r.Int64()
+	}
+
 	return &PNCounterDelta[T]{
 		DocID:           []byte(reg.key.DocID),
 		FieldName:       reg.fieldName,
 		Data:            value,
 		SchemaVersionID: reg.schemaVersionKey.SchemaVersionId,
-	}
+		Entropy:         entropy,
+	}, nil
 }
 
 // Merge implements ReplicatedData interface.
--- a/core/crdt/pncounter.go
+++ b/core/crdt/pncounter.go
@@ -48,9 +48,9 @@ type PNCounterDelta[T Incrementable] struct {
 	DocID     []byte
 	FieldName string
 	Priority  uint64
-	// Entropy is an added randomly generated number that ensures
+	// Nonce is an added randomly generated number that ensures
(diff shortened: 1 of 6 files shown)


core/crdt/pncounter.go
```diff
@@ -45,6 +48,9 @@ type PNCounterDelta[T Incrementable] struct {
 	DocID     []byte
 	FieldName string
 	Priority  uint64
+	// Entropy is an added randomly generated number that ensures
+	// that each increment operation is unique.
+	Entropy int64
```
jsimnz, line 53: todo: Lets rename this to `Nonce` as its more cryptographically accurate, and slightly more common

core/crdt/pncounter.go
```diff
@@ -108,13 +114,30 @@ func (reg PNCounter[T]) Value(ctx context.Context) ([]byte, error) {
 }
 
 // Set generates a new delta with the supplied value
-func (reg PNCounter[T]) Increment(value T) *PNCounterDelta[T] {
+func (reg PNCounter[T]) Increment(ctx context.Context, value T) (*PNCounterDelta[T], error) {
+	// To ensure that the dag block is unique, we add a random number to the delta.
+	// This is done only on update (if the doc doesn't already exist) to ensure that the
+	// initial dag block of a document can be reproducible.
```
AndrewSisley, line 120: praise: Good catch :)

question: Do we have any tests that target this and ensure that the docID is consistent (e.g. by hard coding it in the test)? If not, please consider this a todo :)

AndrewSisley: Should have read first, there are tests that cover this, but they are complicated by cids and updates.  Can you please add a simple one that makes it clear why such a test is needed?

AndrewSisley (approved):
LGTM, one test todo first before merge though please :)

core/crdt/pncounter.go
```diff
@@ -108,13 +114,30 @@ func (reg PNCounter[T]) Value(ctx context.Context) ([]byte, error) {
...
+	exists, err := reg.store.Has(ctx, reg.key.ToPrimaryDataStoreKey().ToDS())
+	if err != nil {
+		return nil, err
+	}
+	var entropy int64
+	if exists {
```
jsimnz, line 126: question: When creating a new PNCounter field, this is going to produce an entropy value of `0`. Is this intended? Or do we want to omit the entropy field entirely when creating? (IE: not include in the cbor marshal)

jsimnz (changes requested):

core/crdt/pncounter.go
fredcarle: If the docID wasn't consistent, the updates wouldn't sync over P2P.

core/crdt/pncounter.go
fredcarle: This is indeed intended. Instead of having a different struct for create vs update, using 0 gives us the same result with less complexity.

core/crdt/pncounter.go
AndrewSisley: Lol... Whoever introduced that to crypto was probably not British... It is not very polite.  If it is only slightly more common I'd suggest sticking with `Entropy`.

core/crdt/pncounter.go
AndrewSisley: True :)  Thanks for pointing out

fredcarle: I'm adding a docID assertion on the mutation create test just to make it more obvious.

core/crdt/pncounter.go
fredcarle: A clarification. The entropy will be `0` only when creating the **document**.

core/crdt/pncounter.go
shahzadlone: FWIW `nonce` is fairly common in the cryptography world, so I have a slight pref for it.

shahzadlone (approved):
Neat fix, LGTM

jsimnz (approved):

PR #2202, opened 2024-01-12
fix(i): Remove default collection fields from gql view types

### LINKED ISSUE #2199: Base level view GQL types include auto generated fields
e.g. given the below setup, `_docID` is present on `BookView`, and I assume (unverified) that `_version` is too:
```

			testUtils.SchemaUpdate{
				Schema: `
					type Author {
						name: String
						books: [Book]
					}
					type Book {
						name: String
						author: Author
					}
				`,
			},
			testUtils.CreateView{
				Query: `
					Book {
						name
						author {
							name
						}
					}
				`,
				SDL: `
					type BookView {
						name: String
						author: AuthorView
					}
					interface AuthorView {
						name: String
					}
				`,
			},
```

AndrewSisley:
## Relevant issue(s)

Resolves #2199

## Description

Removes default collection fields from gql view types.

Views should not automatically have these fields.  It will also be impossible to guarantee that they exist once we allow Views to have Lens transforms.

### DIFF
--- a/request/graphql/schema/generate.go
+++ b/request/graphql/schema/generate.go
@@ -417,6 +417,7 @@ func (g *Generator) buildTypes(
 		collection := c
 		fieldDescriptions := collection.Schema.Fields
 		isEmbeddedObject := collection.Description.Name == ""
+		isViewObject := isEmbeddedObject || collection.Description.BaseQuery != nil
 
 		var objectName string
 		if isEmbeddedObject {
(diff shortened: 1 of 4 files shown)


nasdf:
What happens if the user explicitly includes the `_docID` in the view query?

```go
testUtils.CreateView{
  Query: `
	  Book {
                  _docID
		  name
	  }
  `,
  SDL: `
	  type BookView {
                   _docID: ID
		  name: String
	  }
  `,
}
```

AndrewSisley:
> What happens if the user explicitly includes the _docID in the view query?

At the moment it will reject it (error on view creation), as `_docID` is a reserved field name.  Users can work around that by aliasing - later we can perhaps allow them to declare it.

nasdf (approved):

Issue #1962, opened 2023-10-13
test: Improve index comparison benchmarks


islamaliev:
at the moment the way we assert performance of queries on indexed fields (in tests/integration/index/query_performance_test.go) is not as expressive as it could be.

We run 2 tests: one to run a normal query that asserts results, and another one with @explain to assert on metrics.
The first run is just there to make sure the query produces the same result as without indexes. The actual values don't really matter.

Instead of explicitly providing expected results (which we don't care about) we could run the query on a not-indexed set, and then the same query in a equivalent indexed set and to assert that both result are equal.

Another problem the we should address is that in our tests we don't want to be dependant on a specific data set, but if we don't know what the state of the database is we don't know which and how many documents are going to fetched, hence we can't assert on @explain result.

So one of the ideas is to configure documents auto-generation in a way the would randomise them in a predictable way that can be asserted agains.

PR #2226, opened 2024-01-17
feat: Add composite indexes

### LINKED ISSUE #299: Sec. Indexes: Implement composite indexes

islamaliev:
## Relevant issue(s)

Resolves #299 

## Description

This change introduces composite secondary indexes as well as unique-composite secondary indexes.
Note: different order direction for composite fields is not included in this change.

## Tasks

## How has this been tested?

Unit and integration tests.

- MacOS

### DIFF
--- a/client/index.go
+++ b/client/index.go
@@ -58,3 +58,13 @@ func (d CollectionDescription) CollectIndexedFields(schema *SchemaDescription) [
 	}
 	return fields
 }
+
+func (d CollectionDescription) CollectIndexesOnField(fieldName string) []IndexDescription {
+	result := []IndexDescription{}
+	for _, index := range d.Indexes {
+		if index.Fields[0].Name == fieldName {
+			result = append(result, index)
+		}
+	}
+	return result
+}
--- a/client/index.go
+++ b/client/index.go
@@ -59,7 +59,9 @@ func (d CollectionDescription) CollectIndexedFields(schema *SchemaDescription) [
 	return fields
 }
 
-func (d CollectionDescription) CollectIndexesOnField(fieldName string) []IndexDescription {
+// GetIndexesOnField returns all indexes that are indexing the given field.
+// If the field is not the first field of a composite index, the index is not returned.
+func (d CollectionDescription) GetIndexesOnField(fieldName string) []IndexDescription {
 	result := []IndexDescription{}
 	for _, index := range d.Indexes {
 		if index.Fields[0].Name == fieldName {
--- a/db/errors.go
+++ b/db/errors.go
@@ -87,6 +87,7 @@ const (
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
 	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
+	errCanNotIndexNonUniqueCombination    string = "can not index a doc's fields that violate unique index"
 	errInvalidViewQuery                   string = "the query provided is not valid as a View"
 )
 
@@ -573,6 +574,14 @@ func NewErrCanNotIndexNonUniqueField(docID, fieldName string, value any) error {
 		errors.NewKV("Field name", fieldName),
 		errors.NewKV("Field value", value),
 	)
+}
+
+func NewErrCanNotIndexNonUniqueCombination(docID string, fieldValues ...errors.KV) error {
+	kvPairs := make([]errors.KV, 0, len(fieldValues)+1)
+	kvPairs = append(kvPairs, errors.NewKV("DocID", docID))
+	kvPairs = append(kvPairs, fieldValues...)
+
+	return errors.New(errCanNotIndexNonUniqueCombination, kvPairs...)
 }
 
 func NewErrInvalidViewQueryCastFailed(query string) error {
--- a/db/errors.go
+++ b/db/errors.go
@@ -69,7 +69,6 @@ const (
 	errInvalidFieldValue                  string = "invalid field value"
 	errUnsupportedIndexFieldType          string = "unsupported index field type"
 	errIndexDescriptionHasNoFields        string = "index description has no fields"
-	errIndexDescHasNonExistingField       string = "index description has non existing field"
 	errFieldOrAliasToFieldNotExist        string = "The given field or alias to field does not exist"
 	errCreateFile                         string = "failed to create file"
 	errRemoveFile                         string = "failed to remove file"
@@ -466,16 +465,6 @@ func NewErrIndexDescHasNoFields(desc client.IndexDescription) error {
 	return errors.New(
 		errIndexDescriptionHasNoFields,
 		errors.NewKV("Description", desc),
-	)
-}
-
-// NewErrIndexDescHasNonExistingField returns a new error indicating that the given index
-// description points to a field that does not exist.
-func NewErrIndexDescHasNonExistingField(desc client.IndexDescription, fieldName string) error {
-	return errors.New(
-		errIndexDescHasNonExistingField,
-		errors.NewKV("Description", desc),
-		errors.NewKV("Field name", fieldName),
 	)
 }
 
--- a/db/errors.go
+++ b/db/errors.go
@@ -86,8 +86,7 @@ const (
 	errExpectedJSONArray                  string = "expected JSON array"
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
-	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
-	errCanNotIndexNonUniqueCombination    string = "can not index a doc's fields that violate unique index"
+	errCanNotIndexNonUniqueFields         string = "can not index a doc's field(s) that violates unique index"
 	errInvalidViewQuery                   string = "the query provided is not valid as a View"
 )
 
@@ -109,6 +108,7 @@ var (
 	ErrExpectedJSONObject             = errors.New(errExpectedJSONObject)
 	ErrExpectedJSONArray              = errors.New(errExpectedJSONArray)
 	ErrInvalidViewQuery               = errors.New(errInvalidViewQuery)
+	ErrCanNotIndexNonUniqueFields     = errors.New(errCanNotIndexNonUniqueFields)
 )
 
 // NewErrFailedToGetHeads returns a new error indicating that the heads of a document
@@ -567,21 +567,12 @@ func NewErrIndexDoesNotMatchName(index, name string) error {
 	)
 }
 
-func NewErrCanNotIndexNonUniqueField(docID, fieldName string, value any) error {
-	return errors.New(
-		errCanNotIndexNonUniqueField,
-		errors.NewKV("DocID", docID),
-		errors.NewKV("Field name", fieldName),
-		errors.NewKV("Field value", value),
-	)
-}
-
-func NewErrCanNotIndexNonUniqueCombination(docID string, fieldValues ...errors.KV) error {
+func NewErrCanNotIndexNonUniqueFields(docID string, fieldValues ...errors.KV) error {
 	kvPairs := make([]errors.KV, 0, len(fieldValues)+1)
 	kvPairs = append(kvPairs, errors.NewKV("DocID", docID))
 	kvPairs = append(kvPairs, fieldValues...)
 
-	return errors.New(errCanNotIndexNonUniqueCombination, kvPairs...)
+	return errors.New(errCanNotIndexNonUniqueFields, kvPairs...)
 }
 
 func NewErrInvalidViewQueryCastFailed(query string) error {
--- a/db/fetcher/indexer_iterators.go
+++ b/db/fetcher/indexer_iterators.go
@@ -89,10 +89,6 @@ type eqPrefixIndexIterator struct {
 	queryResultIterator
 }
 
-func (i *eqPrefixIndexIterator) SetKeyFieldValue(value []byte) {
-	i.keyFieldValue = value
-}
-
 func (i *eqPrefixIndexIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
 	i.indexKey.FieldValues = [][]byte{i.keyFieldValue}
 	resultIter, err := store.Query(ctx, query.Query{
@@ -123,22 +119,17 @@ func (i *eqPrefixIndexIterator) Next() (indexIterResult, error) {
 	}
 }
 
-type keyFieldIndexIterator interface {
-	indexIterator
-	SetKeyFieldValue([]byte)
-}
-
 type eqSingleIndexIterator struct {
-	indexKey      core.IndexDataStoreKey
-	keyFieldValue []byte
-	execInfo      *ExecInfo
+	indexKey       core.IndexDataStoreKey
+	keyFieldValues [][]byte
+	execInfo       *ExecInfo
 
 	ctx   context.Context
 	store datastore.DSReaderWriter
 }
 
 func (i *eqSingleIndexIterator) SetKeyFieldValue(value []byte) {
-	i.keyFieldValue = value
+	i.keyFieldValues = [][]byte{value}
 }
 
 func (i *eqSingleIndexIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
@@ -151,7 +142,7 @@ func (i *eqSingleIndexIterator) Next() (indexIterResult, error) {
 	if i.store == nil {
 		return indexIterResult{}, nil
 	}
-	i.indexKey.FieldValues = [][]byte{i.keyFieldValue}
+	i.indexKey.FieldValues = i.keyFieldValues
 	val, err := i.store.Get(i.ctx, i.indexKey.ToDS())
 	if err != nil {
 		if errors.Is(err, ds.ErrNotFound) {
@@ -169,7 +160,7 @@ func (i *eqSingleIndexIterator) Close() error {
 }
 
 type inIndexIterator struct {
-	keyFieldIndexIterator
+	indexIterator
 	keyFieldValues [][]byte
 	nextValIndex   int
 	ctx            context.Context
@@ -179,7 +170,7 @@ type inIndexIterator struct {
 
 func (i *inIndexIterator) nextIterator() (bool, error) {
 	if i.nextValIndex > 0 {
-		err := i.keyFieldIndexIterator.Close()
+		err := i.indexIterator.Close()
 		if err != nil {
 			return false, err
 		}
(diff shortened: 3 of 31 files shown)


db/fetcher/indexer_iterators.go
```diff
@@ -157,38 +160,33 @@ func (i *eqSingleIndexIterator) Close() error {
...
 		return false, nil
 	}
 
-	i.SetFilterValue(i.filterValues[i.nextValIndex])
-	err := i.filterValueIndexIterator.Init(i.ctx, i.store)
+	switch fieldIter := i.indexIterator.(type) {
+	case *eqPrefixIndexIterator:
+		fieldIter.keyFieldValue = i.keyFieldValues[i.nextValIndex]
+	case *eqSingleIndexIterator:
+		fieldIter.keyFieldValues[0] = i.keyFieldValues[i.nextValIndex]
```
AndrewSisley, line 187: question: I don't see a constructor for `eqSingleIndexIterator`, where is it guaranteed that the slice will have a `0` index?

db/fetcher/indexer_iterators.go
```diff
@@ -423,70 +401,22 @@ func createIndexIterator(
 			return nil, err
 		}
 
+		m := &indexByteValuesMatcher{value: valueBytes}
 		switch op {
```
AndrewSisley, line 405: praise: Thanks for tidying this switch up - looks much nicer :)

db/fetcher/indexer_iterators.go
```diff
@@ -423,70 +401,22 @@ func createIndexIterator(
...
-			} else {
-				return &eqPrefixIndexIterator{
-					indexKey: indexDataStoreKey,
-					filterValueHolder: filterValueHolder{
-						value: valueBytes,
-					},
-					execInfo: execInfo,
-				}, nil
-			}
+			m.evalFunc = func(res int) bool { return res == 0 }
```
AndrewSisley, line 407: question: As mentioned, I really do like the improvements to this switch, however it does feel a bit strange that you are returning an interface and instead of having lots of implementations for that interface, you are passing in a func to a private property in a concrete implementation of that interface and then returning that one 'super' type instance.

Why did you chose to do this?  Did it just kind of evolve as you were working in this space?

db/fetcher/indexer_iterators.go
```diff
@@ -501,38 +431,176 @@ func createIndexIterator(
...
-					execInfo: execInfo,
-				}
+		return newNinIndexCmp(valArr, op == opIn), nil
+	case opLike, opNlike:
+		return newLikeIndexCmp(filterVal.(string), op == opLike), nil
+	case opAny:
+		return &anyMatcher{}, nil
+	}
+
+	return nil, errors.New("invalid index filter condition")
```
AndrewSisley, line 441: todo: Please declare this error in the errors.go file :)

db/fetcher/indexer_iterators.go
```diff
@@ -501,38 +431,176 @@ func createIndexIterator(
...
+
+func (f *IndexFetcher) determineFieldFilterConditions() []fieldFilterCond {
+	result := make([]fieldFilterCond, 0, len(f.indexedFields))
+	for i := range f.indexedFields {
+		fieldInd := f.mapping.FirstIndexOfName(f.indexedFields[i].Name)
+		found := false
+		for filterKey, indexFilterCond := range f.indexFilter.Conditions {
+			propKey, ok := filterKey.(*mapper.PropertyIndex)
+			if !ok {
+				continue
```
AndrewSisley, line 469: todo: I can't spot any good reason for this ever happening, and codecov says it is untested.  Should this not be an error?  Or perhaps can we cast here without checking type?  If neither of those can you please add a comment documenting how this can happen, and why it is okay if it does.

EDIT: When looking at the next `fieldInd != propKey.Index` if, I think this is just you cycling through trying to find the correct property, however the cast raises eyebrows and it's intent is not that clear - if this the reason for this code, can I suggest that instead of the current you do the below, as it makes it clearer that the cast is just a requirement for finding the correct prop:

```
if !ok || fieldInd != propKey.Index {
    // this is not the property we are looking for (reword :))
    continue
}
```

AndrewSisley (commented):
Part way through my review, submitting comments so far :)

db/fetcher/indexer_iterators.go
```diff
@@ -501,38 +431,176 @@ func createIndexIterator(
...
+		}
+
+		// iterators for _in filter already iterate over keys with first field value
+		// matching the filter value, so we can skip the first matcher
+		if len(matchers) > 1 {
+			matchers[0] = &anyMatcher{}
+		}
+
+		var iter indexIterator
+		if isUniqueFetchByFullKey(&f.indexDesc, fieldConditions) {
```
AndrewSisley, line 577: todo: This looks like it will always be false no? As to get here at least one of the `fieldContitions` must be `opIn`, and `isUniqueFetchByFullKey` requires that all `fieldContitions` must be `opEq`.

Regardless of whether I'm right or wrong here, can you please document this.

db/index.go
```diff
@@ -90,16 +90,20 @@ func NewCollectionIndex(
...
-	var err error
-	base.validateFieldFunc, err = getFieldValidateFunc(field.Kind)
-	if err != nil {
-		return nil, err
+	base.validateFieldFuncs = make([]func(any) bool, 0, len(desc.Fields))
+	base.fieldsDescs = make([]client.FieldDescription, 0, len(desc.Fields))
+	for _, fieldDesc := range desc.Fields {
+		field, foundField := collection.Schema().GetField(fieldDesc.Name)
+		if !foundField {
+			return nil, NewErrIndexDescHasNonExistingField(desc, desc.Fields[0].Name)
```
AndrewSisley, line 99: suggestion:  I don't think the new error kind is worth it here, or at least it should wrap our `client.ErrFieldNotExist` error.  The more different kinds of errors we have in the system, the harder it will be for users to know/discover/handle them. Happy for this to be shrugged off as out of scope, as this was here before.

db/indexed_docs_test.go
```diff
@@ -94,11 +94,11 @@ func (b *indexKeyBuilder) Col(colName string) *indexKeyBuilder {
 	return b
 }
 
-// Field sets the field name for the index key.
+// Fields sets the fields' names for the index key.
```
AndrewSisley, line 97: nitpick: I think `fields' names` implies that each field has multiple names.  If each field only has one name then `field names` is more appropriate.  Same goes for the variable names below.

planner/filter/normalize.go
```diff
@@ -185,7 +185,12 @@ func normalizeProperties(parentKey connor.FilterKey, conditions []any) []any {
 	// if canMergeAnd is true, all _and groups will be merged
 	props := make(map[int][]any)
 	for _, c := range conditions {
-		for key, val := range c.(map[connor.FilterKey]any) {
+		cMap, ok := c.(map[connor.FilterKey]any)
```
AndrewSisley, line 188: question: Why has this changed?  What else could it be?

planner/select.go
```diff
@@ -296,20 +296,22 @@ func (n *selectNode) initSource() ([]aggregateNode, error) {
 	return aggregates, nil
 }
 
-func findFilteredByIndexedField(scanNode *scanNode) immutable.Option[client.FieldDescription] {
+func findFilteredByIndexedField(scanNode *scanNode) immutable.Option[client.IndexDescription] {
 	if scanNode.filter != nil {
```
AndrewSisley, line 300: suggestion: Inverting this if statement would make it easier to read IMO.

planner/select.go
```diff
@@ -296,20 +296,22 @@ func (n *selectNode) initSource() ([]aggregateNode, error) {
 	return aggregates, nil
 }
 
-func findFilteredByIndexedField(scanNode *scanNode) immutable.Option[client.FieldDescription] {
+func findFilteredByIndexedField(scanNode *scanNode) immutable.Option[client.IndexDescription] {
```
AndrewSisley, line 299: suggestion:  Is an old name I know, but it looks like it is missing a word? I suggest renaming to `findFilteredIndexByIndexedField`, or maybe `findIndexByFilteringField`.

planner/select.go
```diff
@@ -296,20 +296,22 @@ func (n *selectNode) initSource() ([]aggregateNode, error) {
...
-				return immutable.Some(indexedFields[i])
+		colDesc := scanNode.col.Description()
+
+		for _, field := range scanNode.col.Schema().Fields {
+			if _, isFiltered := scanNode.filter.ExternalConditions[field.Name]; !isFiltered {
+				continue
+			}
+			indexes := colDesc.CollectIndexesOnField(field.Name)
+			if len(indexes) > 0 {
+				return immutable.Some(indexes[0])
```
AndrewSisley, line 309: question:  This looks suspicious, why are you just returning the first?  What if there are multiple indexes using the same field? It that tested?

AndrewSisley (commented):
Nearly through it, submitting a few more comments.

tests/integration/index/query_with_composite_index_only_filter_test.go
```diff
@@ -0,0 +1,767 @@
...
+		User(filter: {name: {_eq: "Islam"}, age: {_eq: 66}}) {
+			name
+			age
+		}
+	}`
+	test := testUtils.TestCase{
+		Description: "Test filtering on composite index with _eq filter",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: schemaWithNameAgeIndex,
```
AndrewSisley, line 56: suggestion: I do feel like you are hiding the most important part of this test here.  The index is what you are testing, and as reader, what I am most interesting in seeing, and yet it is not visible within the test.

tests/integration/index/query_with_composite_index_only_filter_test.go
```diff
@@ -0,0 +1,767 @@
...
+		Description: "Test index filtering with _nin filter",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: schemaWithNameAgeIndex,
+			},
+			testUtils.CreatePredefinedDocs{
+				Docs: getUserDocs(),
+			},
+			testUtils.Request{
+				Request: req,
```
AndrewSisley, line 429: suggestion: I dislike how you had to pull the other super important part of the test out of the action flow, disrupting things for the reader.

I wonder if you could add an action that wraps the two requests perhaps, so the test could read:
```
testUtils.RequestWithExplain{
    Request: `query {
		User(filter: {age: {_nin: [20, 23, 28, 42]}, name: {_nin: ["John", "Andy", "Chris"]}}) {
			name
		}
	}`,
	Results: []map[string]any{
					{"name": "Roy"},
					{"name": "Islam"},
					{"name": "Keenan"},
				},
	Explanation: testUtils.NewExplainAsserter().WithFieldFetches(3).WithIndexFetches(10),
}
``` 
Internally it would split the action into the two Request Actions (that are currently here in the test) before sending them all into the main test executor func.

Happy if you prefer we dont do this in this PR.

tests/integration/test_case.go
```diff
@@ -246,6 +246,7 @@ type CreateIndex struct {
 	// The names of the fields to index. Used only for composite indexes.
 	FieldsNames []string
 	// The directions of the 'FieldsNames' to index. Used only for composite indexes.
+	// If not provided all fields will be indexed in ascending order.
 	Directions []client.IndexDirection
```
AndrewSisley, line 250: suggestion: It feels a bit clunky to have to separate fields from directions like this.  Happy for it to be done outside of this PR but I do think it would be nicer to have something like the below:
```
Fields []client.IndexedFieldDescription
```

db/fetcher/indexer_iterators.go
```diff
@@ -39,6 +39,7 @@ const (
 	opNin   = "_nin"
 	opLike  = "_like"
 	opNlike = "_nlike"
+	opAny   = "_any"
```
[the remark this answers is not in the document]
islamaliev: it's just there for composite indexes. We construct a slice of value matchers with every matcher being responsible for a corresponding field in the index to match. For some fields there might not be any criteria to match. For examples if you have composite index of /name/age/email/ and in the filter you specify only "name" and "email". Then the "_any" matcher will be used for "age".

I will add a comment to explain this.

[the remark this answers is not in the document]
AndrewSisley: Okay cheers, I think I'm understanding this better.  In that case, can I also suggestion:

suggestion: Can those that are actual request filters reference those consts, and those which are not (i.e. opAny) remain as strings - this might make it clearer that some are special (and will also may it easier and less error prone to change the public filter syntax as these values will update when those do).  e.g.:
```
const (
   opEq = request.FilterOpEq,
   ...
   opAny   = "_any"
)
```
Although I found that `connor`/etc doesn't have a const yet for these, which would mean you have the hassle of changing all of those too, so feel very free to shrug this suggestion off as out of scope.

client/index.go
```diff
@@ -58,3 +58,13 @@ func (d CollectionDescription) CollectIndexedFields(schema *SchemaDescription) [
 	}
 	return fields
 }
+
+func (d CollectionDescription) CollectIndexesOnField(fieldName string) []IndexDescription {
```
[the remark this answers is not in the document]
islamaliev: added a comment, renamed

db/errors.go
```diff
@@ -87,6 +87,7 @@ const (
 	errOneOneAlreadyLinked                string = "target document is already linked to another document"
 	errIndexDoesNotMatchName              string = "the index used does not match the given name"
 	errCanNotIndexNonUniqueField          string = "can not index a doc's field that violates unique index"
+	errCanNotIndexNonUniqueCombination    string = "can not index a doc's fields that violate unique index"
```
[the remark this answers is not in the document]
islamaliev: merged 2 errors into 1

db/fetcher/indexer_iterators.go
islamaliev: it is created in `createIndexIterator` and it's guaranteed because it's the field value we are seeking and the value is being  set upon creation.

db/fetcher/indexer_iterators.go
islamaliev: the interface is different kinds of matchers there is byte-matcher, in-matcher, like-matcher and new any-matcher. All of them (except for any) are configurable. Byte-matcher just a bit more: it compares bytes and we can configure what is the expected result of the comparison. 

I think this way it's a bit more concise, but I don't have a strong preference here.

db/fetcher/indexer_iterators.go
islamaliev: done

db/fetcher/indexer_iterators.go
AndrewSisley: Ah okay, there are still a fair few concrete types then :)

> I think this way it's a bit more concise, but I don't have a strong preference here.

Keep it as-is then, unless anyone else has a really strong preference - I have nothing really against this, it just struck me as unusual and worth highlighting/chatting about :)  Is a fairly nice way of reducing the number to types required.

db/fetcher/indexer_iterators.go
islamaliev: merged ifs and added a comment

db/fetcher/indexer_iterators.go
islamaliev: added a comment to `isUniqueFetchByFullKey`

db/index.go
islamaliev: replaced the error

db/indexed_docs_test.go
islamaliev: adjusted

planner/filter/normalize.go
islamaliev: `normalizeProperty` method that calls this `normalizeProperties` checks is given condition is `map[connor.FilterKey]any` or an array of conditions which has a type of `[]any`. But as it recursively checks all nested conditions and values one of the values (not conditions) can match `[]any` (for example "_in" op). That's why we need to check if it's really a conditions or just value.

planner/select.go
islamaliev: adjusted. Not part of this PR though

planner/select.go
islamaliev: it will be part of another PR. 

Added a comment.

tests/integration/index/query_with_composite_index_only_filter_test.go
islamaliev: done

tests/integration/index/query_with_composite_index_only_filter_test.go
islamaliev: we've got a task for it https://github.com/sourcenetwork/defradb/issues/1962

tests/integration/test_case.go
islamaliev: I don't think I agree. This will make every such test action more wordy. There will be a few tests that test different directions.

I just added a comment there. You were not supposed to comment on this part :D 

The next task I'm working on involves different directions. I'd suggest we check how it's going to look in those tests and discuss it there.

planner/filter/normalize.go
AndrewSisley: Ah okay - this was a bug fix?  Did you add an integration test that covers this?

planner/select.go
AndrewSisley: cheers Islam

planner/select.go
AndrewSisley: Okay cheers.  Can you please make sure that there are integration tests that make sure that if there are multiple overlapping indexes (e.g. two composites, a simple and a composite) that queries still work (no need for explain test documenting this issue, but that would be nice if you can do it reliably without much effort too)?

tests/integration/index/query_with_composite_index_only_filter_test.go
AndrewSisley: Ah nice, my bad and thanks :)

tests/integration/test_case.go
AndrewSisley: > I just added a comment there. You were not supposed to comment on this part :D

:grin: :grin: 

> The next task I'm working on involves different directions. I'd suggest we check how it's going to look in those tests and discuss it there.

That sounds sensible, I'm happy with that :)

db/fetcher/indexer_iterators.go
AndrewSisley: Ah okay, I would prefer a constructor for stuff like that, but `createIndexIterator` is a factory-like func and  the concrete a private type, so I am happy for this to be shrugged off as my personal preference here :grin:

db/fetcher/indexer_iterators.go
AndrewSisley: cheers, looks good

db/fetcher/indexer_iterators.go
AndrewSisley: Ah of course, cheers - that comment really helps, I didn't spot it skipping the first index :)

AndrewSisley (approved):
Looks good to me, thanks Islam :)

I'm approving now, as all but one of my remaining comments are very optional within this PR. I do think we need the following though before merge please:

> Can you please make sure that there are integration tests that make sure that if there are multiple overlapping indexes (e.g. two composites, a simple and a composite) that queries still work (no need for explain test documenting this issue, but that would be nice if you can do it reliably without much effort too).

And preferably at least one for what looks like a bug fix in `filter/normailze.go` (if it is a bug fix :))

planner/filter/normalize.go
islamaliev: an integration test found this bug :)

planner/select.go
islamaliev: This line of code is not new. There are integration tests that cover it

PR #2242, opened 2024-01-22
ci(i): Fix and document release workflow logic

### LINKED ISSUE #2244: Fix and document release workflow logic
Fix and document release workflow logic

shahzadlone:
## Relevant issue(s)

Resolves #2244 

## Description
Fixes the broken id usage (i.e. `cache`) which didn't exist, use the newer `save` and `restore`, add some conditions and flags to fail early if the release fails, and document the cache saving logic (which is not so obvious).


fredcarle (approved):
LGTM. Thanks for cleaning this up Shahzad. 

Please add an related issue before merging.

PR #2246, opened 2024-01-23
refactor(i): Delete dead code in `db.addSchema` func

### LINKED ISSUE #2251: Delete dead code in db.addSchema func
Delete dead code in db.addSchema func

shahzadlone:
## Relevant issue(s)

Resolves #2251

## Description
Seems like dead code. Not sure if this was the outcome of a partial implementation or just some code that wasn't removed before being merged.

### DIFF
--- a/db/schema.go
+++ b/db/schema.go
@@ -39,16 +39,6 @@ func (db *db) addSchema(
 	txn datastore.Txn,
 	schemaString string,
 ) ([]client.CollectionDescription, error) {
-	existingCollections, err := db.getAllCollections(ctx, txn)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


AndrewSisley (approved):
It is historical, I would rather you avoid changing anything else in this space for a little while though as I am in the middle of restructuring it.

fredcarle (approved):
LGTM. Please link an issue before merging :)

shahzadlone:
> It is historical, I would rather you avoid changing anything else in this space for a little while though as I am in the middle of restructuring it.

I don't think I will be changing anything else here, the closest I would probably be changing is in the `parser` code I believe.

Issue #2255, opened 2024-01-24
Values in new fields are not visible if docs synced before receiving node updates schema


AndrewSisley:
I.e.:
```
func TestP2PPeerCreateWithNewFieldDocSyncedBeforeReceivingNodeSchemaUpdatedDoesNotReturnNewField(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.RandomNetworkingConfig(),
			testUtils.RandomNetworkingConfig(),
			testUtils.SchemaUpdate{
				Schema: `
					type Users {
						Name: String
					}
				`,
			},
			testUtils.ConnectPeers{
				SourceNodeID: 1,
				TargetNodeID: 0,
			},
			testUtils.SubscribeToCollection{
				NodeID:        1,
				CollectionIDs: []int{0},
			},
			testUtils.SchemaPatch{
				// Patch the schema on the first node only
				NodeID: immutable.Some(0),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "Email", "Kind": 11} }
					]
				`,
			},
			testUtils.CreateDoc{
				// Create the doc with a value in the new field on the first node only, and allow the values to sync
				NodeID: immutable.Some(0),
				Doc: `{
					"Name": "John",
					"Email": "imnotyourbuddyguy@source.ca"
				}`,
			},
			testUtils.WaitForSync{},
			testUtils.SchemaPatch{
				// Update the schema on the second node
				NodeID: immutable.Some(1),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "Email", "Kind": 11} }
					]
				`,
			},
			testUtils.Request{
				NodeID: immutable.Some(0),
				Request: `
					query {
						Users {
							Name
							Email
						}
					}
				`,
				Results: []map[string]any{
					{
						"Name":  "John",
						"Email": "imnotyourbuddyguy@source.ca",
					},
				},
			},
			testUtils.Request{
				NodeID: immutable.Some(1),
				Request: `
					query {
						Users {
							Name
							Email
						}
					}
				`,
				Results: []map[string]any{
					{
						"Name": "John",
						// The email should be returned but it is not
					},
				},
			},
		},
	}

	testUtils.ExecuteTestCase(t, test)
}
```

AndrewSisley:
[...]

PR #2256, opened 2024-01-24
test(i): Document doc sync before schema update bug

AndrewSisley:
## Relevant issue(s)

Documents #2255

## Description

Documents a bug (very likely cache related) when a doc syncs to a node on an older schema version before the schema is updated.


shahzadlone (approved):
LGTM

PR #2266, opened 2024-01-29
fix: Load root dir before loading config

### LINKED ISSUE #2264: Config file not being loaded on start
Changes to the config file are not being taken into consideration when starting a DefraDB node. This is caused by the root directory not being set when we check for an existing config file.

fredcarle:
## Relevant issue(s)

Resolves #2264 

## Description

This PR ensures that the root dir path gets defined prior to looking for a config file. It fixes an issue where the config file wasn't being loaded on node start.

## Tasks

## How has this been tested?

manually tested that a change in the config file would be reflected in the node behaviour.

- MacOS


nasdf (approved):
Thanks for fixing!

PR #2276, opened 2024-01-31
feat: Multiple docs with nil value on unique-indexed field

### LINKED ISSUE #2180: Sec. indexes: Allow multiple fields with `nil` values with unique index
It should be possible to have multiple docs with `nil` values on a field that is indexed with a unique index.

islamaliev:
## Relevant issue(s)

Resolves #2180

## Description

Allow multiple docs with nil value on unique-indexed field.
This change also included few minor query optimizations that involve secondary indexes.

Explain request is slightly adjusted not to increase field fetches count upon fetching indexes.

## Tasks

## How has this been tested?

Unit and integration tests.

- MacOS

### DIFF
--- a/db/fetcher/indexer_iterators.go
+++ b/db/fetcher/indexer_iterators.go
@@ -492,7 +492,7 @@ func (f *IndexFetcher) determineFieldFilterConditions() []fieldFilterCond {
 
 // isUniqueFetchByFullKey checks if the only index key can be fetched by the full index key.
 //
-// This method ignores the first condition because it's expected to be called only
+// This method ignores the first condition (unless it's nil) because it's expected to be called only
 // when the first field is used as a prefix in the index key. So we only check if the
 // rest of the conditions are _eq.
 func isUniqueFetchByFullKey(indexDesc *client.IndexDescription, conditions []fieldFilterCond) bool {
@@ -500,7 +500,7 @@ func isUniqueFetchByFullKey(indexDesc *client.IndexDescription, conditions []fie
 	// if all fields of the index are specified in the filter
 	res := indexDesc.Unique && len(conditions) == len(indexDesc.Fields)
 
-	// if the first is _eq, val should be not nil
+	// first condition is not required to be _eq, but if is, val must be not nil
 	res = res && (conditions[0].op != opEq || conditions[0].val != nil)
 
 	// for the rest it must be _eq and val must be not nil
@@ -523,6 +523,8 @@ func getFieldsBytes(conditions []fieldFilterCond) ([][]byte, error) {
 	return result, nil
 }
 
+// newPrefixIndexIterator creates a new eqPrefixIndexIterator for fetching indexed data.
+// It can modify the input matchers slice.
 func (f *IndexFetcher) newPrefixIndexIterator(
 	fieldConditions []fieldFilterCond,
 	matchers []valueMatcher,
@@ -530,6 +532,8 @@ func (f *IndexFetcher) newPrefixIndexIterator(
 	keyFieldValues := make([][]byte, 0, len(fieldConditions))
 	for i := range fieldConditions {
 		if fieldConditions[i].op != opEq {
+			// prefix can be created only for subsequent _eq conditions
+			// if we encounter any other condition, we built the longest prefix we could
 			break
 		}
 
@@ -558,6 +562,8 @@ func (f *IndexFetcher) newPrefixIndexIterator(
 	}, nil
 }
 
+// newInIndexIterator creates a new inIndexIterator for fetching indexed data.
+// It can modify the input matchers slice.
 func (f *IndexFetcher) newInIndexIterator(
 	fieldConditions []fieldFilterCond,
 	matchers []valueMatcher,
--- a/db/fetcher/indexer_iterators.go
+++ b/db/fetcher/indexer_iterators.go
@@ -86,16 +86,14 @@ func (i *queryResultIterator) Close() error {
 }
 
 type eqPrefixIndexIterator struct {
-	indexKey      core.IndexDataStoreKey
-	keyFieldValue []byte
-	execInfo      *ExecInfo
-	matchers      []valueMatcher
+	indexKey core.IndexDataStoreKey
+	execInfo *ExecInfo
+	matchers []valueMatcher
 
 	queryResultIterator
 }
 
 func (i *eqPrefixIndexIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
-	i.indexKey.FieldValues = [][]byte{i.keyFieldValue}
 	resultIter, err := store.Query(ctx, query.Query{
 		Prefix: i.indexKey.ToString(),
 	})
@@ -166,11 +164,11 @@ func (i *eqSingleIndexIterator) Close() error {
 
 type inIndexIterator struct {
 	indexIterator
-	keyFieldValues [][]byte
-	nextValIndex   int
-	ctx            context.Context
-	store          datastore.DSReaderWriter
-	hasIterator    bool
+	inValues     [][]byte
+	nextValIndex int
+	ctx          context.Context
+	store        datastore.DSReaderWriter
+	hasIterator  bool
 }
 
 func (i *inIndexIterator) nextIterator() (bool, error) {
@@ -181,15 +179,15 @@ func (i *inIndexIterator) nextIterator() (bool, error) {
 		}
 	}
 
-	if i.nextValIndex >= len(i.keyFieldValues) {
+	if i.nextValIndex >= len(i.inValues) {
 		return false, nil
 	}
 
 	switch fieldIter := i.indexIterator.(type) {
 	case *eqPrefixIndexIterator:
-		fieldIter.keyFieldValue = i.keyFieldValues[i.nextValIndex]
+		fieldIter.indexKey.FieldValues[0] = i.inValues[i.nextValIndex]
 	case *eqSingleIndexIterator:
-		fieldIter.keyFieldValues[0] = i.keyFieldValues[i.nextValIndex]
+		fieldIter.keyFieldValues[0] = i.inValues[i.nextValIndex]
 	}
 	err := i.indexIterator.Init(i.ctx, i.store)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


tests/integration/index/query_with_unique_index_only_filter_test.go
```diff
@@ -532,3 +532,54 @@ func TestQueryWithUniqueIndex_WithEqualFilterOnNilValue_ShouldFetch(t *testing.T
...
+				Results: []map[string]any{
+					{"name": "Bob"},
+					{"name": "Alice"},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
```
AndrewSisley, line 585: praise: Testing looks good, thank you for taking the time :)

db/fetcher/indexer_iterators.go
```diff
@@ -520,9 +523,98 @@ func getFieldsBytes(conditions []fieldFilterCond) ([][]byte, error) {
...
 }
 
+func (f *IndexFetcher) newPrefixIndexIterator(
+	fieldConditions []fieldFilterCond,
+	matchers []valueMatcher,
+) (*eqPrefixIndexIterator, error) {
+	keyFieldValues := make([][]byte, 0, len(fieldConditions))
+	for i := range fieldConditions {
+		if fieldConditions[i].op != opEq {
+			break
```
AndrewSisley, line 533: suggestion: A comment here explaining why we'd want to break here could be really handy

db/fetcher/indexer_iterators.go
```diff
@@ -520,9 +523,98 @@ func getFieldsBytes(conditions []fieldFilterCond) ([][]byte, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		keyFieldArr = append(keyFieldArr, keyFieldBytes)
+	}
+
+	// iterators for _in filter already iterate over keys with first field value
+	// matching the filter value, so we can skip the first matcher
+	if len(matchers) > 1 {
+		matchers[0] = &anyMatcher{}
```
AndrewSisley, line 582: suggestion: `matchers` is not owned by this function, seeing an element in it overwritten like this sets off alarm bells in my head.  Especially when this function's signature does not suggest that it will be mutating any of its input params.

If you wish to keep the mutation, can I please ask that you document it on the function signature?

The same goes for `newPrefixIndexIterator` which does similar.

db/index.go
```diff
@@ -123,16 +123,24 @@ func (i *collectionBaseIndex) getDocFieldValue(doc *client.Document) ([][]byte,
 	result := make([][]byte, 0, len(i.fieldsDescs))
 	for iter := range i.fieldsDescs {
 		fieldVal, err := doc.GetValue(i.fieldsDescs[iter].Name)
+		isNil := false
 		if err != nil {
 			if errors.Is(err, client.ErrFieldNotExist) {
```
AndrewSisley, line 128: thought: :sweat_smile: Out of scope here perhaps, but we should add a `TryGetValue` func to `client.Document`.

db/index.go
```diff
@@ -123,16 +123,24 @@ func (i *collectionBaseIndex) getDocFieldValue(doc *client.Document) ([][]byte,
...
+			} else {
+				return nil, err
 			}
-			return nil, err
+		}
+		if fieldVal != nil && fieldVal.Value() == nil {
+			isNil = true
+		}
+		if isNil {
+			valBytes, err := client.NewFieldValue(client.LWW_REGISTER, nil).Bytes()
```
AndrewSisley, line 138: suggestion: This does look a bit strange, can we perhaps instead have a const in the client package for this:

```
// With documentation and probably a better name
const NIL_FIELD_BYTES []byte = ...
```

db/index.go
```diff
@@ -270,33 +278,46 @@ func (i *collectionSimpleIndex) deleteDocIndex(
 	return i.deleteIndexKey(ctx, txn, key)
 }
 
+// hasIndexKeyNilField returns true if the index key has a field with nil value
+func hasIndexKeyNilField(key *core.IndexDataStoreKey) bool {
+	const cborNil = 0xf6
+	for i := range key.FieldValues {
+		if len(key.FieldValues[i]) == 1 && key.FieldValues[i][0] == cborNil {
```
AndrewSisley, line 285: suggestion: This line earlier in the review threw me quite a bit, and it is repeated here.  It might be worth adding an `IsNill` func to `core.IndexDataStoreKey` with a bit of documentation.  

suggestion: It is also not great that we are having to duplicate the `cborNil` const in a few places. Can I suggest moving it to a shared place and making it accessible, and renaming it to something that does not include cbor in the name, so if it being cbor related changes we don't have to update a whole bunch of references to it (it being cbor also seems irrelevant to the referencing location).

db/index.go
```diff
@@ -270,33 +278,46 @@ func (i *collectionSimpleIndex) deleteDocIndex(
...
+}
+
 type collectionUniqueIndex struct {
 	collectionBaseIndex
 }
 
 var _ CollectionIndex = (*collectionUniqueIndex)(nil)
 
-func (i *collectionUniqueIndex) Save(
+func (i *collectionUniqueIndex) saveRecord(
```
AndrewSisley, line 298: nitpick: Why `saveRecord` and not `save`?

db/index.go
```diff
@@ -321,36 +342,66 @@ func (i *collectionUniqueIndex) newUniqueIndexError(
...
+) (core.IndexDataStoreKey, []byte, error) {
+	key, err := i.getDocumentsIndexKey(doc)
+	if err != nil {
+		return core.IndexDataStoreKey{}, nil, err
+	}
+	if hasIndexKeyNilField(&key) {
+		key.FieldValues = append(key.FieldValues, []byte(doc.ID().String()))
+		return key, []byte{}, nil
+	} else {
+		return key, []byte(doc.ID().String()), nil
```
AndrewSisley, line 356: suggestion: There is a `Bytes()` func on `client.DocID`, no need to cast it twice :)

(same thing a few lines above)

db/index.go
```diff
@@ -321,36 +342,66 @@ func (i *collectionUniqueIndex) newUniqueIndexError(
...
 
+func (i *collectionUniqueIndex) getDocumentsIndexRecord(
+	doc *client.Document,
+) (core.IndexDataStoreKey, []byte, error) {
+	key, err := i.getDocumentsIndexKey(doc)
+	if err != nil {
+		return core.IndexDataStoreKey{}, nil, err
+	}
+	if hasIndexKeyNilField(&key) {
+		key.FieldValues = append(key.FieldValues, []byte(doc.ID().String()))
```
AndrewSisley, line 353: question:  This looks strange, why are you doing it?

AndrewSisley (approved):
Looks good to me, I mostly just have documentation requests :)

db/index.go
islamaliev: if an index key has a nil value we tread the key as the same way as non-unique index, i.e. append docID

AndrewSisley: Why is this not done within `getDocumentsIndexKey`?  In the current setup `getDocumentsIndexKey` appears to be returning an incorrect key.

db/index.go
islamaliev: it's not used in the code that's why I was hesitant. And it seems to produce different result, than normal "bae-..."

AndrewSisley: If it is weird and we don't want to use it, why should we expect our users to use it?

Best to either make it not weird and use it, or delete it I think.

db/index.go
islamaliev: I share your concerns, I had the same thought. 
But I'm currently working on encoding on another branch. All this will be gone anyway.
That's why I didn't bother too much with this part.
If you still think I should do this in this PR, I can do it.

db/index.go
islamaliev: this will be gone as well with new encoding. I can add a comment though.

db/index.go
AndrewSisley: Ah okay nice!  If you are confident it will be removed soon I'm happy for it to remain as-is here :)

db/index.go
AndrewSisley: Nice - thanks Islam, same as comment on `if len(key.FieldValues[i]) == 1 && key.FieldValues[i][0] == cborNil` - I'm happy as-is if you are killing it/reworking it soon :)

db/index.go
islamaliev: This is actually what I wanted to ask on upcoming sync-up. 

Deleting it wasn't part of the PR.

fredcarle: In a separate PR, it might be worth modifying the `Bytes()` function to do what we want it to do if we don't use it for anything else.

db/index.go
islamaliev: `getDocumentsIndexKey` is called on `collectionBaseIndex`. It doesn't return incorrect key, it returns incomplete key that is being complete in this function.

AndrewSisley: If I have a `collectionUniqueIndex` and I want to get the key for a document, and I see a function called `getDocumentsIndexKey` I'm going to call it, and I'd expect a complete, valid key (or an error) returned.

Does `getDocumentsIndexKey` have a `WARNING` in the documentation that says it returns incomplete keys?

db/fetcher/indexer_iterators.go
islamaliev: added a comment

db/index.go
islamaliev: I like the idea.

db/index.go
islamaliev: no parcilar reason. Changed

tests/integration/index/create_unique_test.go
```diff
@@ -295,8 +293,6 @@ func TestUniqueIndexCreate_UponAddingDocWithExistingNilValue_ReturnError(t *test
 					{
 						"name":	"Andy"
 					}`,
-				ExpectedError: db.NewErrCanNotIndexNonUniqueFields(
-					"bae-2159860f-3cd1-59de-9440-71331e77cbb8", errors.NewKV("age", nil)).Error(),
 			},
 		},
 	}
```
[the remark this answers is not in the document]
islamaliev: added a test action `GetIndexes` that is run after `CreateIndex` while having exinsting docs with multiple `nil` values.
For the rest of the test it pointless because they create index along with the collection and there are already tests for this.
Note: creating docs can not delete index.

Issue #2257, opened 2024-01-26
[EPIC] Config refactor


nasdf:
The config package is difficult to maintain at the moment and can be improved in a few areas.

- Decouple the config from the consumer packages (net, http, etc.)
- Validation and parsing should be handled by the consumers of the config values.
- Remove config items that are not used or not relevant.
- Merge config into cli package. (The cli is the only place where the config is used.)

```[tasklist]
### Tasks
- [ ] https://github.com/sourcenetwork/defradb/issues/2263
- [ ] https://github.com/sourcenetwork/defradb/issues/2277
- [ ] https://github.com/sourcenetwork/defradb/issues/2295
```

PR #2278, opened 2024-02-01
refactor: HTTP config

### LINKED ISSUE #2277: HTTP config refactor
The HTTP config can be simplified by moving support for autotls to an external tool such as https://go-acme.github.io/lego/

nasdf:
## Relevant issue(s)

Resolves #2277 

## Description

This PR is a refactor of the HTTP config and is part of the larger config refactor #2257 

## Tasks

## How has this been tested?

make test

- MacOS


http/server.go
```diff
@@ -13,304 +13,171 @@ package http
...
+
+// ListenAndServe listens for and serves incoming connections.
+func (s *Server) ListenAndServe() error {
+	listener, err := net.Listen("tcp", s.options.Address)
+	if err != nil {
+		return err
 	}
+	// ignore close errors as they cannot be handled
+	// from the caller of this method
+	defer listener.Close() //nolint:errcheck
```
AndrewSisley, line 166: question: Why can they not be handled but other errors appear to be so?  And is that not up to any callers of this public function to decide?  Should the choice to discard the error not be made by the caller?

http/server.go
```diff
@@ -13,304 +13,171 @@ package http
...
-	}
-
-	// Save the address on the server in case the port was set to random
-	// and that we want to see what was assigned.
-	s.address = s.listener.Addr().String()
-
-	return nil
+// Server struct holds the Handler for the HTTP API.
+type Server struct {
+	address atomic.Value
```
AndrewSisley, line 119: question: Why is this atomic? It only appears to be set once, in `ListenAndServe`, and that function cannot be called multiple times concurrently no?

todo: If it remains atomic, please document why.

AndrewSisley (approved):
praise:  Code looks much simpler, thank you :)

Only a couple of comments from me, but please address them before merge :)

http/server.go
nasdf: I've added documentation to clarify the use of the atomic value.

AndrewSisley: Ah, you want `AssignedAddr` to return empty if `ListenAndServe` is in progress?  Would it not be better to lock, and block `AssignedAddr` from returning until the address has been assigned? (i.e. no atomic, but a RWLock)

http/server.go
nasdf: I've removed this `Close` call as it is redundant. After taking a closer look the `Serve` and `ServeTLS` functions will close the listener.

http/server.go
nasdf: I agree that would be better, but I realized we don't use this method anywhere outside of `server_test.go`. I've removed the function. 

As a side note: tests that need a random port assignment should use the `httptest` package instead.

fredcarle: > As a side note: tests that need a random port assignment should use the httptest package instead.

That's a really nice change. Thanks for removing the need to keep track of the randomly assigned port on the http server :)

Issue #1772, opened 2023-08-11
Add PatchSchema syntax sugar for RelationType in foreign relations


AndrewSisley:
Part of #1002 

- The ONE/MANY bit should be auto set from kind
- The ONEONE/ONEMANY bit should be auto set from kins (need to look at both sides)
- The primary bit needs to follow the GQL default logic (move?), and have a means to be explicitly set by the user without defining the full bitmap
- The InternalId should be auto set from Kind (should be done independently of auto-create of _id field, as the user may have reason to explicitly define part of the _id field - particularly if in the future we decouple the name from function) 

See also semi-related: https://github.com/sourcenetwork/defradb/issues/2287

AndrewSisley:
Under discussion

PR #2288, opened 2024-02-07
feat: Replace FieldDescription.RelationType with IsPrimary

### LINKED ISSUE #2287: Clean up RelationType
It being a bitmap is a pain, for us, and the users.  Break it up.

1) Relation_Type_INTERNAL_ID is to be removed, if a field is part of a relation and Kind is ID, it must be the ID field
2) Relation_Type_ONE and Relation_Type_MANY are to be removed, this is also already handled by Kind.
3) Relation_Type_MANYMANY is to be removed, as it is not supported atm
4) Relation_Type_ONEONE and Relation_Type_ONEMANY will probably be removed, as it looks like we only ever care about on half anyway, and that is covered by Kind
5) Relation_Type_Primary will become a bool, it might be that we can remove this later though (when allowing some fields to be declared on the collection, not schema) 

This does mean that the schema ID will not change depending on whether it is the primary side of a one-one or a one-many.

See also semi-related https://github.com/sourcenetwork/defradb/issues/1772

AndrewSisley:
## Relevant issue(s)

Resolves #2287 #1772

## Description

Replaces `FieldDescription.RelationType` with an `IsPrimary` boolean.

One notable effect of this that may not be easily visible in the code-diff is that the schema ID will not change depending on whether it is the primary side of a one-one or on the one side of a one-many - now the schema only cares about the shape of the data, not what is on the other side of the relationship.

Commits should be clean, it might be easier reviewing them one by one.


islamaliev (approved):
Looks much cleaner now. Good job!

islamaliev:
bug bash result: tested by creating different types and 1-to-1 and 1-no-many relationships and querying on them. Everything works

PR #2296, opened 2024-02-08
refactor: Node config

### LINKED ISSUE #2295: Node config refactor
Node configuration can be refactored into a new package that simplifies the setup of a DefraDB instance with optional sub-systems.

nasdf:
## Relevant issue(s)

Resolves #2295 

## Description

This PR refactors the node setup logic into a new `node` package.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -1,0 +1,42 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package node
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/defradb/db"
+)
+
+func TestNodeStart(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	opts := []Opt{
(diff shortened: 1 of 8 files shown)


node/node_test.go
```diff
@@ -0,0 +1,42 @@
...
+		WithDatabaseOpts(db.WithUpdateEvents()),
+	}
+
+	node, err := NewNode(ctx, opts...)
+	require.NoError(t, err)
+
+	err = node.Start(ctx)
+	require.NoError(t, err)
+
+	<-time.After(5 * time.Second)
```
AndrewSisley, line 38: question:  How much value do you think this test provides?  Is it not covering something that is covered by pretty much every integration test we have, but with a 5 second cost?

node/store_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+package node
+
+import (
+	"crypto/rand"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
AndrewSisley, line 19: question: These tests do not appear to actually test their config works, only that it does not error - they (the WithFoo funcs) could be empty funcs for all these tests care.  Do you think they are worth keeping and maintaining?

AndrewSisley (approved):
Production code looks nicer :), but I am uncertain as to the value provided by the new tests.

node/node_test.go
nasdf: I wanted to have something similar to the `start-binary` workflow. I think it would be valuable to catch those errors before that workflow is run.

node/store_test.go
nasdf: I've updated the tests. They should make more sense now.

PR #2297, opened 2024-02-08
fix: Fix compound relational filters in aggregates

### LINKED ISSUE #1869: Panic using compound relational filter in top level aggregate

AndrewSisley:
## Relevant issue(s)

Resolves #1869

## Description

Fixes compound relational filters in aggregates where their dependencies were not already part of the query.


shahzadlone (approved):
LGTM

PR #2306, opened 2024-02-12
fix: Add missing delta payload

### LINKED ISSUE #2299: delta field on commit/version queries returns incorrect result
**Describe the problem**
When requesting the `delta` field on either ` _version` object or a `commit` query will return incorrect values.

**To Reproduce**
- Create any schema
- Create any document with that schema
- Query the `_version` field or `commit` query specifying the `delta` in the selection set

**Expected behavior**
Return the base64 encode delta payload, instead it just seems to return the string "Data"

**Platform**
Linux WSL

fredcarle:
## Relevant issue(s)

Resolves #2299

## Description

This PR fixes a regression bug and adds integration tests to ensure that future changes don't break the expected query results

## Tasks

## How has this been tested?

make test and manual queries

- MacOS

### DIFF
--- a/request/graphql/schema/types/commits.go
+++ b/request/graphql/schema/types/commits.go
@@ -47,11 +47,11 @@ var (
 		Name:        request.CommitTypeName,
 		Description: commitDescription,
 		Fields: gql.Fields{
-			"height": &gql.Field{
+			request.HeightFieldName: &gql.Field{
 				Description: commitHeightFieldDescription,
 				Type:        gql.Int,
 			},
-			"cid": &gql.Field{
+			request.CidFieldName: &gql.Field{
 				Description: commitCIDFieldDescription,
 				Type:        gql.String,
 			},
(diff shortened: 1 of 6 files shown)


tests/integration/query/commits/simple_test.go
```diff
@@ -315,3 +313,118 @@ func TestQueryCommitsWithFieldIDFieldWithUpdate(t *testing.T) {
...
+								name
+							}
+						}
+					}
+				`,
+				Results: []map[string]any{
+					{
+						"cid":          "bafybeiddpjl27ulw2yo4ohup6gr2wob3pwagqw2rbeaxxodv4ljelnu7ve",
+						"collectionID": int64(1),
+						"delta":        []uint8([]byte{0x16}),
```
nasdf, line 354: question: should this be a base64 encoding instead?

request/graphql/schema/types/commits.go
```diff
@@ -47,47 +47,47 @@ var (
 		Name:        request.CommitTypeName,
 		Description: commitDescription,
 		Fields: gql.Fields{
-			"height": &gql.Field{
+			request.HeightFieldName: &gql.Field{
```
nasdf, line 50: praise: this is much cleaner

nasdf (approved):
Fix looks good! Just one question about encoding.

request/graphql/schema/types/commits.go
AndrewSisley: It looks nicer, but now we have lost protection against these changing.  I have mixed feelings about this change.

AndrewSisley (approved):
Looks good to me, and thank you for the tests :)

I am a bit uncomfortable about the use of production constants in the expected test results, but not uncomfortable enough to block it :)

tests/integration/query/commits/simple_test.go
fredcarle: yes. Trying to figure something out for this. It si base64 when using CLI and HTTP clients but not when using the go client.

fredcarle: So it's base64 with CLI and HTTP clients because that's how `json.Marshal` will convert `[]uint8` to `string` representation. I've added support for `[]uint8` result assertion when using the CLI and HTTP clients.

shahzadlone (approved):
Thanks for this fix! LGTM

PR #2310, opened 2024-02-13
refactor: Make config internal to CLI

### LINKED ISSUE #2257: [EPIC] Config refactor

nasdf:
## Relevant issue(s)

Resolves #2257

## Description

This PR refactors the config into a simpler implementation that exists within the `cli` package.

Notable changes:

- ~~`datastore.store` config key has been replaced with `datastore.badger.inMemory`~~
- ~~`--store` cli flag has been replaced with `--in-memory`~~
- `log.overrides` config map has been added. (contains same keys as `log` but for individual loggers)
- `--logger` cli flag has been temporarily removed (will be replaced in upcoming log library)
- `init` command has been removed. (can be replaced with `rm -rf ~/.defradb` or similar)

## Tasks

## How has this been tested?

`make test`

- MacOS


cli/config.go
```diff
@@ -0,0 +1,178 @@
...
+	cfg.SetDefault("log.output", "stderr")
+	cfg.SetDefault("log.noColor", false)
+	cfg.SetDefault("log.caller", false)
+
+	return cfg
+}
+
+// createConfig writes the default config file if one does not exist.
+func createConfig(rootdir string) error {
+	cfg := defaultConfig()
```
fredcarle, line 91: suggestion: If you add the `bind cli flags to config keys` loop to `createConfig`, you will be able to remove all the `cfg.SetDefault` for the keys in `configFlags`. The flags default will be used instead and it will remove the need to maintain 2 sets of default values.

fredcarle (commented):
A few comments while it's in draft :)

cli/config.go
```diff
@@ -0,0 +1,178 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	_ "embed"
```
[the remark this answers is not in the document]
nasdf: Yeah good catch!

cli/utils.go
```diff
@@ -18,16 +18,20 @@ import (
...
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
 	"github.com/sourcenetwork/defradb/client"
-	"github.com/sourcenetwork/defradb/config"
 	"github.com/sourcenetwork/defradb/datastore"
 	"github.com/sourcenetwork/defradb/http"
 )
 
 type contextKey string
```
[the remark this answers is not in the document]
nasdf: I used a custom type here because it is recommended to avoid name collisions.

https://pkg.go.dev/context#WithValue

cli/utils.go
```diff
@@ -60,19 +64,47 @@ func mustGetP2PContext(cmd *cobra.Command) client.P2P {
 	return cmd.Context().Value(dbContextKey).(client.P2P)
 }
 
+// mustGetConfigContext returns the config for the current command context.
+//
+// If a config is not set in the current context this function panics.
+func mustGetConfigContext(cmd *cobra.Command) *viper.Viper {
```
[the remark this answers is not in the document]
nasdf: I agree that is confusing. I've updated all of the function names.

cli/config.go
nasdf: I was hoping there was a way to do this but I didn't think of that. Thanks for the suggestion!

cli/utils.go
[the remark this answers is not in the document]
fredcarle: Ah yes. Sorry I should have thought about that. Thanks for the reminder.

cli/config.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if err := bindConfigFlags(cfg, flags); err != nil {
+		return err
+	}
+	// make sure rootdir exists
+	if err := os.MkdirAll(rootdir, 0755); err != nil {
+		return err
+	}
+	err := cfg.SafeWriteConfig()
+	if _, ok := err.(viper.ConfigFileAlreadyExistsError); ok { //nolint:errorlint
```
AndrewSisley, line 87: todo: Can you please document the reasons for the `nolint` stuff with code-comments please (in all locations).

cli/config.go
```diff
@@ -0,0 +1,173 @@
...
+	err := cfg.ReadInConfig()
+	if _, ok := err.(viper.ConfigFileNotFoundError); err != nil && !ok { //nolint:errorlint
+		return nil, err
+	}
+	// bind cli flags to config keys
+	if err := bindConfigFlags(cfg, flags); err != nil {
+		return nil, err
+	}
+
+	// make paths relative to the rootdir
```
AndrewSisley, line 108: praise: Thank you for comments like this, it helps a lot when the code itself is non-obvious.

cli/config.go
```diff
@@ -0,0 +1,173 @@
...
+		level = logging.Error
+	case "fatal":
+		level = logging.Fatal
+	default:
+		level = logging.Info
+	}
+
+	var format logging.EncoderFormat
+	switch value := cfg.GetString("format"); value {
+	case "json": //nolint:goconst
```
AndrewSisley, line 157: suggestion: It is annoying that the linter flags this, but I think that using a const (even though that is also bad here) is slightly less bad that a `nolint`.

cli/tx_commit.go
```diff
@@ -15,22 +15,23 @@ import (
...
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			cfg := mustGetContextConfig(cmd)
+
 			id, err := strconv.ParseUint(args[0], 10, 64)
 			if err != nil {
 				return err
 			}
-			tx, err := http.NewTransaction(cfg.API.Address, id)
+			tx, err := http.NewTransaction(cfg.GetString("api.address"), id)
```
AndrewSisley, line 34: question: This feels like it should be a shared const somewhere, not hardcoded here - do you have reasons for declaring it here like this?

(same for the other locations where this is done)

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+## `datastore.store`
+
+Store can be badger or memory. Defaults to `badger`.
+
+- badger: fast pure Go key-value store optimized for SSDs (https://github.com/dgraph-io/badger)
+- memory: in-memory version of badger
+
+## `datastore.maxtxnretries`
+
+Maximum number of times to retry a failed transaction. Defaults to `5`.
```
AndrewSisley, line 16: todo: I don't think this line fully explains this option, from my memory this is for P2P only txn commit attempts, but that is not clear here.

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+The default DefraDB directory is `$HOME/.defradb`. It can be changed via the --rootdir CLI flag.
+
+Relative paths are interpreted as being rooted in the DefraDB directory.
+
+## `datastore.store`
+
+Store can be badger or memory. Defaults to `badger`.
+
+- badger: fast pure Go key-value store optimized for SSDs (https://github.com/dgraph-io/badger)
+- memory: in-memory version of badger
```
AndrewSisley, line 12: thought: `memory` here confuses me a little bit:

1. We have our own memory store, and I though that was what `datastore.store=memory` was
2. Badger memory is driven by the badger memory boolean option, if I am incorrect with (1), then why are we duplicating the option

suggestion: Unless I'm really confused/incorrect here, and given that we do document some badger config options in this file, I think it might be worth documenting the badger memory config prop.

AndrewSisley (approved):
Looks good to me, very happy with the amount of deleted code :)

I have a couple of small comments that I would like resolving before merge, and a small concern around the below:

> --logger cli flag has been temporarily removed (will be replaced in upcoming log library)

But if you and everyone else is comfy with that, I'm happy enough.

cli/config.go
nasdf: I moved all of them to `const` for now. They should be removed when the new logging package is ready.

cli/tx_commit.go
nasdf: I wanted to avoid having const keys or a pre-defined config struct so that the config can be modified more easily. The places that access the `api.address` specifically should have good test coverage since they use random port assignments in the tests.

docs/config.md
nasdf: I've updated it with a bit more text from the db documentation.

docs/config.md
AndrewSisley: Ah fair enough.

> We could swap it to the memory datastore, but I'd prefer to do that in a new PR.

I agree that *if* we want to make that change it should be done in another PR.

docs/config.md
AndrewSisley: Cheers, new text looks good.

fredcarle (approved):
LGTM. Thanks for applying the changes :)

shahzadlone (approved):
LGTM, do suggest making the commit/PR Title abit more detailed rather than just `Config` after the commit labal

PR #2311, opened 2024-02-13
feat: Add support for views with Lens transforms

### LINKED ISSUE #2070: Add support for un-cached Lens-driven Defra Views
Add support for un-cached Lens-driven Defra Views.

Persistence/caching of the result will be added later. Mutations through the view (to underlying documents) will be added later. P2P stuff will be added later.

AndrewSisley:
## Relevant issue(s)

Resolves #2070

## Description

Adds support for views with Lens transforms.

### DIFF
--- a/planner/lens.go
+++ b/planner/lens.go
@@ -1,0 +1,174 @@
+// Copyright 2023 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package planner
+
+import (
+	"github.com/sourcenetwork/immutable/enumerable"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/client/request"
+	"github.com/sourcenetwork/defradb/core"
+)
+
+// viewNode applies a lens transform to data yielded from the source node.
+//
+// It may return a different number of documents to that yielded by its source,
+// and there is no guarentee that those documents will actually exist as documents
+// in Defra (they may be created by the transform).
+type lensNode struct {
+	docMapper
+	documentIterator
+
+	p          *Planner
+	source     planNode
+	collection client.CollectionDescription
+
+	input  enumerable.Queue[map[string]any]
+	output enumerable.Enumerable[map[string]any]
+}
+
+func (p *Planner) Lens(source planNode, docMap *core.DocumentMapping, col client.Collection) *lensNode {
+	return &lensNode{
+		docMapper:  docMapper{docMap},
+		p:          p,
+		source:     source,
+		collection: col.Description(),
+	}
+}
+
+func (n *lensNode) Init() error {
+	n.input = enumerable.NewQueue[map[string]any]()
+
+	pipe, err := n.p.db.LensRegistry().MigrateUp(n.p.ctx, n.input, n.collection.ID)
+	if err != nil {
+		return err
+	}
+
+	n.output = pipe
+
+	return n.source.Init()
+}
+
+func (n *lensNode) Start() error {
+	return n.source.Start()
+}
+
+func (n *lensNode) Spans(spans core.Spans) {
+	n.source.Spans(spans)
+}
+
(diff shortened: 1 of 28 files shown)


tests/lenses/rust_wasm32_filter/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+static PARAMETERS: RwLock<StreamOption<Parameters>> = RwLock::new(None);
+
+#[no_mangle]
+pub extern fn alloc(size: usize) -> *mut u8 {
+    lens_sdk::alloc(size)
+}
+
+#[no_mangle]
```
fredcarle, line 49: what is no_mangle

fredcarle (approved):
LGTM. I've create a few follow up issues that stemmed from reviewing this.

- https://github.com/sourcenetwork/defradb/issues/2314
- https://github.com/sourcenetwork/defradb/issues/2315
- https://github.com/sourcenetwork/defradb/issues/2316

planner/lens.go
shahzadlone, line 1: todo: Since a new plan node is being added, I would make the same comment as: https://github.com/sourcenetwork/defradb/pull/2114#discussion_r1432460060

tldr: Please add the new name to `allPlanNodeNames` variable in the `tests/integration/explain.go` file for debug node and add a test for the debug explain under the folder `tests/integration/explain/debug` similar to `tests/integration/explain/debug/with_view_test.go`. You don't have to make it explainable however in this PR, that discussion can be had elsewhere but would suggest opening an issue for it.

shahzadlone (changes requested):
Mostly looks good, just the one main concern regarding the explain debug test similar to this comment when view node was added: https://github.com/sourcenetwork/defradb/pull/2114#discussion_r1432460060

shahzadlone:
> LGTM. I've create a few follow up issues that stemmed from reviewing this.
> 
> * [Add explain method to lens node #2314](https://github.com/sourcenetwork/defradb/issues/2314)
> * [Add explain method to view node #2315](https://github.com/sourcenetwork/defradb/issues/2315)
> * [Add support for foreign objects in document creation. #2316](https://github.com/sourcenetwork/defradb/issues/2316)

Ahh cheers for making the explain issues. Are the issues for discussion if we need lens and view nodes, or we already decided that we need explain for them?

planner/lens.go
AndrewSisley: - [ ] Add explain test stuff

AndrewSisley:
> Are the issues for discussion if we need lens and view nodes, or we already decided that we need explain for them?

I don't think that they needing explain is a given, although I think the fact that we have this question and hide some nodes from our users is an unwanted complication for both us and them and plan on chatting about this in the standup (it came up in some of my other work).

AndrewSisley:
> LGTM. I've create a few follow up issues that stemmed from reviewing this.

Cheers Fred :)

planner/lens.go
```diff
@@ -0,0 +1,174 @@
...
+func (n *lensNode) Start() error {
+	return n.source.Start()
+}
+
+func (n *lensNode) Spans(spans core.Spans) {
+	n.source.Spans(spans)
+}
+
+func (n *lensNode) Next() (bool, error) {
+	hasNext, err := n.output.Next()
```
islamaliev, line 69: nitpick: var with name `hasValue` would fit better here

planner/lens.go
```diff
@@ -0,0 +1,174 @@
...
+		// similar logic, for example when converting a core.Doc to a map before passing it
+		// into a lens transform.
+		fieldIndex := indexes[len(indexes)-1]
+
+		if len(properties) <= fieldIndex {
+			// Because the document is sourced from another mapping, we may still need to grow
+			// the resultant field set. We cannot use [append] because the index of each field
+			// must still correspond to it's field ID.
+			originalProps := properties
+			properties = make([]any, fieldIndex+1)
```
islamaliev, line 144: suggestion: it will be more efficient to run through `indexes` to find the maximum value and then allocate `properties`. It can be also done before the loop.
Otherwise there might be scenarios when `properties` can be allocated several times.

AndrewSisley: Which one is more efficient depends on the two sizes really.  Not every indexes key will be present in the map and your suggestion will over-allocate when that happens.

Your suggestion is probably better more of the time, but I don't think it is worth a new PR.

Issue #2333, opened 2024-02-20
FieldDescription IDs are dependent on which fields are currently on the schema


AndrewSisley:
This creates problems for schema branching and field removal.

For example, if two different fields are added in different schema branches, they will have the same ID (as the number of other fields is the same).  This means that data saved in documents on the two branches will share the same space/datastore-keys and do weird stuff.

WARNING: This cannot be solved simply by using a sequence for the IDs at the moment, as the ID field (local concept) is current held on the schema (global).  The fact that other fields have been added on different schema branches (incrementing the sequence) must not impact the schema version ids (via the field id).

PR #2336, opened 2024-02-21
fix: Move field id off of schema

### LINKED ISSUE #2334: Remove FieldDescription.ID from Schema
At the moment schema branching doesn't really work, due to #2333.  

We also we to move local field description stuff off the schema and on to the collection long term (so it's location makes sense, and doesn't impact the global schema).

Move FieldDescription.ID on to something on the CollectionDescription.  Do so whilst keeping in mind that other stuff (such as (perhaps) RelationName, and secondary relation fields).

AndrewSisley:
## Relevant issue(s)

Resolves #2334

## Description

Moves field id off of schema and onto collection.

It is now a local property and does not need to be the same across multiple nodes.  The new `Fields` property on `CollectionDescription` will be where other local field stuff is move too from the schema (relation_name, descriptions, secondary relation fields).

This change is a prerequisite of allowing fields to be deleted and schema branching (see https://github.com/sourcenetwork/defradb/issues/2333 for more info on that if currious)

### DIFF
--- a/client/definitions.go
+++ b/client/definitions.go
@@ -109,32 +109,7 @@ func NewFieldDefinition(local CollectionFieldDescription, global SchemaFieldDesc
 	}
 }
 
-// IsObject returns true if this field is an object type.
-func (f FieldDefinition) IsObject() bool {
-	return (f.Kind == FieldKind_FOREIGN_OBJECT) ||
-		(f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY)
-}
-
-// IsObjectArray returns true if this field is an object array type.
-func (f FieldDefinition) IsObjectArray() bool {
-	return (f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY)
-}
-
 // IsRelation returns true if this field is a relation.
 func (f FieldDefinition) IsRelation() bool {
 	return f.RelationName != ""
-}
-
-// IsArray returns true if this field is an array type which includes inline arrays as well
-// as relation arrays.
-func (f FieldDefinition) IsArray() bool {
-	return f.Kind == FieldKind_BOOL_ARRAY ||
-		f.Kind == FieldKind_INT_ARRAY ||
-		f.Kind == FieldKind_FLOAT_ARRAY ||
-		f.Kind == FieldKind_STRING_ARRAY ||
-		f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY ||
-		f.Kind == FieldKind_NILLABLE_BOOL_ARRAY ||
-		f.Kind == FieldKind_NILLABLE_INT_ARRAY ||
-		f.Kind == FieldKind_NILLABLE_FLOAT_ARRAY ||
-		f.Kind == FieldKind_NILLABLE_STRING_ARRAY
 }
--- a/client/definitions.go
+++ b/client/definitions.go
@@ -1,0 +1,132 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package client
+
+// CollectionDefinition contains the metadata defining what a Collection is.
+type CollectionDefinition struct {
+	// Description returns the CollectionDescription of this Collection.
+	Description CollectionDescription `json:"description"`
+	// Schema returns the SchemaDescription used to define this Collection.
+	Schema SchemaDescription `json:"schema"`
+}
+
+// GetFieldByName returns the field for the given field name. If such a field is found it
+// will return it and true, if it is not found it will return false.
+func (def CollectionDefinition) GetFieldByName(fieldName string) (FieldDefinition, bool) {
+	collectionField, ok := def.Description.GetFieldByName(fieldName)
+	if ok {
+		schemaField, ok := def.Schema.GetFieldByName(fieldName)
+		if ok {
+			return NewFieldDefinition(
+				collectionField,
+				schemaField,
+			), true
+		}
+	}
+	return FieldDefinition{}, false
+}
+
+// GetFields returns the combined local and global field elements on this [CollectionDefinition]
+// as a single set.
+func (def CollectionDefinition) GetFields() []FieldDefinition {
+	fields := []FieldDefinition{}
+	for _, localField := range def.Description.Fields {
+		globalField, ok := def.Schema.GetFieldByName(localField.Name)
+		if ok {
+			fields = append(
+				fields,
+				NewFieldDefinition(localField, globalField),
+			)
+		}
+	}
+	return fields
+}
+
+// FieldDefinition describes the combined local and global set of properties that constitutes
+// a field on a collection.
+//
+// It draws it's information from the [CollectionFieldDescription] on the [CollectionDescription],
+// and the [SchemaFieldDescription] on the [SchemaDescription].
+//
+// It is to [CollectionFieldDescription] and [SchemaFieldDescription] what [CollectionDefinition]
+// is to [CollectionDescription] and [SchemaDescription].
+type FieldDefinition struct {
+	// Name contains the name of this field.
+	Name string
+
+	// ID contains the local, internal ID of this field.
+	ID FieldID
+
+	// The data type that this field holds.
+	//
+	// Must contain a valid value. It is currently immutable.
+	Kind FieldKind
+
+	// Schema contains the schema name of the type this field contains if this field is
+	// a relation field.  Otherwise this will be empty.
+	Schema string
+
+	// RelationName the name of the relationship that this field represents if this field is
+	// a relation field.  Otherwise this will be empty.
+	RelationName string
+
+	// The CRDT Type of this field. If no type has been provided it will default to [LWW_REGISTER].
+	//
+	// It is currently immutable.
+	Typ CType
+
+	// If true, this is the primary half of a relation, otherwise is false.
+	IsPrimaryRelation bool
+}
+
+// NewFieldDefinition returns a new [FieldDefinition], combining the given local and global elements
+// into a single object.
+func NewFieldDefinition(local CollectionFieldDescription, global SchemaFieldDescription) FieldDefinition {
+	return FieldDefinition{
+		Name:              global.Name,
+		ID:                local.ID,
+		Kind:              global.Kind,
+		Schema:            global.Schema,
+		RelationName:      global.RelationName,
+		Typ:               global.Typ,
+		IsPrimaryRelation: global.IsPrimaryRelation,
+	}
+}
+
+// IsObject returns true if this field is an object type.
+func (f FieldDefinition) IsObject() bool {
+	return (f.Kind == FieldKind_FOREIGN_OBJECT) ||
+		(f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY)
+}
+
+// IsObjectArray returns true if this field is an object array type.
+func (f FieldDefinition) IsObjectArray() bool {
+	return (f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY)
+}
+
+// IsRelation returns true if this field is a relation.
+func (f FieldDefinition) IsRelation() bool {
+	return f.RelationName != ""
+}
+
+// IsArray returns true if this field is an array type which includes inline arrays as well
+// as relation arrays.
+func (f FieldDefinition) IsArray() bool {
+	return f.Kind == FieldKind_BOOL_ARRAY ||
+		f.Kind == FieldKind_INT_ARRAY ||
+		f.Kind == FieldKind_FLOAT_ARRAY ||
+		f.Kind == FieldKind_STRING_ARRAY ||
+		f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_BOOL_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_INT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_FLOAT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_STRING_ARRAY
+}
--- a/client/definitions.go
+++ b/client/definitions.go
@@ -11,6 +11,10 @@
 package client
 
 // CollectionDefinition contains the metadata defining what a Collection is.
+//
+// The definition types ([CollectionDefinition], [FieldDefinition]) are read-only types returned
+// from various functions as a convienient means to access the computated convergence of schema
+// and collection descriptions.
 type CollectionDefinition struct {
 	// Description returns the CollectionDescription of this Collection.
 	Description CollectionDescription `json:"description"`
@@ -58,6 +62,10 @@ func (def CollectionDefinition) GetFields() []FieldDefinition {
 //
 // It is to [CollectionFieldDescription] and [SchemaFieldDescription] what [CollectionDefinition]
 // is to [CollectionDescription] and [SchemaDescription].
+//
+// The definition types ([CollectionDefinition], [FieldDefinition]) are read-only types returned
+// from various functions as a convienient means to access the computated convergence of schema
+// and collection descriptions.
 type FieldDefinition struct {
 	// Name contains the name of this field.
 	Name string
--- a/client/descriptions.go
+++ b/client/descriptions.go
@@ -74,23 +74,29 @@ func (col CollectionDescription) IDString() string {
 
 // GetFieldByName returns the field for the given field name. If such a field is found it
 // will return it and true, if it is not found it will return false.
-func (def CollectionDefinition) GetFieldByName(fieldName string) (FieldDescription, bool) {
-	return def.Description.GetFieldByName(fieldName, &def.Schema)
+func (def CollectionDefinition) GetFieldByName(fieldName string) (FieldDefinition, bool) {
+	collectionField, ok := def.Description.GetFieldByName(fieldName)
+	if ok {
+		schemaField, ok := def.Schema.GetFieldByName(fieldName)
+		if ok {
+			return NewFieldDefinition(
+				collectionField,
+				schemaField,
+			), true
+		}
+	}
+	return FieldDefinition{}, false
 }
 
 // GetFieldByName returns the field for the given field name. If such a field is found it
 // will return it and true, if it is not found it will return false.
-func (col CollectionDescription) GetFieldByName(fieldName string, schema *SchemaDescription) (FieldDescription, bool) {
-	for _, localField := range col.Fields {
-		if localField.Name == fieldName {
-			globalField, ok := schema.GetFieldByName(fieldName)
-			return NewFieldDescription(
-				localField,
-				globalField,
-			), ok
+func (col CollectionDescription) GetFieldByName(fieldName string) (CollectionFieldDescription, bool) {
+	for _, field := range col.Fields {
+		if field.Name == fieldName {
+			return field, true
 		}
 	}
-	return FieldDescription{}, false
+	return CollectionFieldDescription{}, false
 }
 
 // GetFieldByName returns the field for the given field name. If such a field is found it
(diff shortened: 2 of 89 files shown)


client/descriptions.go
```diff
@@ -312,19 +331,54 @@ func (f FieldID) String() string {
...
+	// ID contains the local, internal ID of this field.
+	ID FieldID
+}
+
+// FieldDescription describes the combined local and global set of properties that constitutes
+// a field on a collection.
+//
+// It draws it's information from the [CollectionFieldDescription] on the [CollectionDescription],
+// and the [SchemaFieldDescription] on the [SchemaDescription].
+type FieldDescription struct {
```
fredcarle, line 377: suggestion: It really threw me off to see `FieldDescription` still in existance. It's an exact duplicate of `SchemaFieldDescription` with tha addition of the `ID` field. Every function that returns `FieldDescription` or uses it is a method of `CollectionDescription` or `collection` and it's odd that when `FieldDescription` is returned you have to construct the struct with `NewFieldDescription`. A much cleaner approach would be to embed the `SchemaFieldDescription` into the `CollectionFieldDescription`.

db/collection.go
```diff
@@ -264,6 +286,27 @@ func (db *db) updateSchema(
 				},
 			}
 
+			for i, globalField := range schema.Fields {
+				isNew := true
+				for _, localField := range col.Fields {
+					if localField.Name == globalField.Name {
```
fredcarle, line 292: thought: There will be limitations here if we allow a field name to be changed locally. The way things are currently saved, a safer way to do this would be to use the global field index and the local field ID.

planner/mapper/mapper.go
```diff
@@ -771,27 +771,34 @@ func getTopLevelInfo(
...
 			}
 			if len(schemas) == 0 {
-				return nil, client.SchemaDescription{}, NewErrTypeNotFound(collectionName)
+				return nil, client.CollectionDefinition{}, NewErrTypeNotFound(collectionName)
+			}
+
+			for i, f := range schemas[0].Fields {
+				// As embedded objects do not have collections/field-ids, we just take the index
+				mapping.Add(int(i), f.Name)
+			}
```
fredcarle, line 783: question: Why wasn't this needed before?

fredcarle (commented):
A few comments to resolve before I can approve :)

client/descriptions.go
```diff
@@ -69,26 +72,52 @@ func (col CollectionDescription) IDString() string {
...
 }
 
-// GetFieldByID searches for a field with the given ID. If such a field is found it
+// GetFieldByName returns the field for the given field name. If such a field is found it
 // will return it and true, if it is not found it will return false.
-func (col CollectionDescription) GetFieldByID(id FieldID, schema *SchemaDescription) (FieldDescription, bool) {
-	for _, field := range schema.Fields {
-		if field.ID == id {
-			return field, true
+func (def CollectionDefinition) GetFieldByName(fieldName string) (FieldDescription, bool) {
```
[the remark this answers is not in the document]
AndrewSisley: Okay, I did like having all the `GetFieldByName` next to each other, but that's partly because the Descriptions are all jumbled up in this one file anyway, and in a way its odd that Definition isn't.

Will let this discussion sit for a while and you/anyone-else talk some more before I try and make a choice RE its location :)

- [ ] Wait and see

client/descriptions.go
AndrewSisley: > It really threw me off to see FieldDescription still in existance.

That's not good - do you have a rough idea as to how much of that was caused my terrible setting of expectations in discord, vs the actual change?

> Every function that returns FieldDescription or uses it is a method of CollectionDescription

Yes, that kind of felt like it was breaking expectations when I was consuming it.  It kind of flows better at the moment IMO, but that is very possibly because of the limited number of properties on collection-field.

> A much cleaner approach would be to embed the SchemaFieldDescription into the CollectionFieldDescription

I do very strongly disagree with this.  It creates a large amount of ambiguity in the data model and is kind of reverting back to the starting point where collection had a pointer to schema on it. Amongst other things, it would be very very confusing when mutating collection field stuff as to what you could and couldn't mutate - it is bad with a pointer, and much worse when embedding.

`CollectionFieldDescription` does not have a `SchemaFieldDescription`, and in the possibly quite near future, not all `CollectionFieldDescription` will even have a corresponding `SchemaFieldDescription` (think secondary fields in a relation).

----

I did originally have a todo here about moving this struct to the `core` package, but decided it was nice having the funcs that return it as public.  Maybe this choice should be revisited, along with the fact that the functions returning it are hosted on `CollectionFieldDescription` (which feels uglier the longer I look at it, even though it is kind of convenient atm).

- [ ] Discuss further, maybe move stuff

db/collection.go
AndrewSisley: I can't see us ever wanting to change field names cant be changed locally to differ from schema field names (guess we could have a local `label` property, but that is not its name).

At the moment the field name is it's global primary identity if/when(?) that changes we can/should of course change the way everything references global field definitions.

planner/mapper/mapper.go
fredcarle: I understand that but I don't understand why now it's added to the mapping and not before. Maybe if you pointed me to a location that uses the mapping instead of the schema.field.id I would understand better.

AndrewSisley: Ah - sorry I understand you now. It was added to the mapper before, on the moved line `mapping.Add(int(f.ID), f.Name)` (old line number 794, new line number 800).

client/descriptions.go
fredcarle: > That's not good - do you have a rough idea as to how much of that was caused my terrible setting of expectations in discord, vs the actual change?

 I though `FieldDescription` would become `SchemaFieldDescription` and `CollectionFieldDescription`. Not have the two new struct plus `FieldDescription`.
 
 > I do very strongly disagree with this. It creates a large amount of ambiguity in the data model
 
 It feels very ambiguous the way it is now in my opinion.  
 
 > is kind of reverting back to the starting point where collection had a pointer to schema on it
 
I disagree. Everywhere `FieldDescription` is used, both collection and schema description are needed to build the struct. It would be simple to remove the confusion and the have the same functionality you have now with something like this
 
```go
type CollectionFieldDescription struct {
  Name string
  ID FieldID

  schemaField immutable.Option[SchemaFieldDescription] `json:"-"`
```
}

db/collection.go
fredcarle: Ah I had missed a key part of the CollectionFieldDescription.Name doc. Thanks for clarifying.

client/descriptions.go
AndrewSisley: > I disagree. Everywhere FieldDescription is used, both collection and schema description are needed to build the struct. It would be simple to remove the confusion and the have the same functionality you have now with something like this

Only at the moment, because we have only moved one property.

planner/mapper/mapper.go
fredcarle: Oh yes I see now. It was done outside the if/else statement before and now done explicitly in each block.

AndrewSisley: Yeah, the rendering of the diff is kind of mangled there

client/descriptions.go
fredcarle: As far as I can tell it would only get worse with more fields in the `CollectionFieldDescription`.

AndrewSisley: The suggestion is totally corrupting the purpose of `CollectionFieldDescription`.  It is no longer modelling just collection fields, and is instead becoming the kind of 'God class' that schema field is/was (and the old collection description with pointer).

There is a loss of separation in the model.  It is dirty and ambiguous.

Just because you dislike FieldDescription (essentially a utility type that could be made internal, it is not actually part of the model) in it's current form, does not mean the only alternative is to blend everything back together again on the public model.

fredcarle: > The suggestion is totally corrupting the purpose of CollectionFieldDescription. It is no longer modelling just collection fields, and is instead becoming the kind of 'God class' that schema field is/was (and the old collection description with pointer).

I don't understand why you are saying this. Everywhere `FieldDescription` is needed, we need both `CollectionFieldDescription` and `SchemaDescription`. Also, everywhere it is used is in a collection context which means a `CollectionFieldDescription` makes a lot more sense here. Since you are saying that at some point a `CollectionFieldDescription` could exists without a relation to a `SchemaFieldDescription`, then it renders `FieldDescription` even more confusing in its usage. However, having an optional `SchemaFieldDescription` in the `CollectionFieldDescription` makes it a lot more obvious what is happening and removes the ambiguity that `FieldDescription` brings in collection contexts.

> There is a loss of separation in the model. It is dirty and ambiguous.

The separation is still there. The is only an optional value added when it makes sense to have it. Like I said above, much less ambiduous. Nothing dirty here.

> Just because you dislike FieldDescription

It's not that I dislike it, it's that it is extremely confusing to be using the `FieldDescription` in a collection context when a `CollectionFieldDescription` exists.

> The suggestion largely defeats the point of this PR and line of work.

It does not. You can still store the two separately and use `SchemaFieldDescription` and `CollectionFieldDescription` on their own. But when `CollectionFieldDescription` has a related `SchemaFieldDescription` that need to be present for different manipulations, we won't need to copy all the field to a new type just for that exact purpose.

client/descriptions.go
```diff
@@ -378,6 +447,36 @@ func (f FieldDescription) IsArray() bool {
...
+	return f.Kind == FieldKind_BOOL_ARRAY ||
+		f.Kind == FieldKind_INT_ARRAY ||
+		f.Kind == FieldKind_FLOAT_ARRAY ||
+		f.Kind == FieldKind_STRING_ARRAY ||
+		f.Kind == FieldKind_FOREIGN_OBJECT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_BOOL_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_INT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_FLOAT_ARRAY ||
+		f.Kind == FieldKind_NILLABLE_STRING_ARRAY
+}
```
fredcarle, line 478: thought: Another not ideal side effect of this is that we need to dupliucate these methods.

client/descriptions.go
```diff
@@ -69,26 +72,52 @@ func (col CollectionDescription) IDString() string {
...
+func (def CollectionDefinition) GetFieldByName(fieldName string) (FieldDescription, bool) {
+	return def.Description.GetFieldByName(fieldName, &def.Schema)
+}
+
+// GetFieldByName returns the field for the given field name. If such a field is found it
+// will return it and true, if it is not found it will return false.
+func (col CollectionDescription) GetFieldByName(fieldName string, schema *SchemaDescription) (FieldDescription, bool) {
+	for _, localField := range col.Fields {
+		if localField.Name == fieldName {
+			globalField, ok := schema.GetFieldByName(fieldName)
```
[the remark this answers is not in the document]
AndrewSisley: yes, I'm currently changing this

client/descriptions.go
fredcarle: If you did something like this https://go.dev/play/p/SwCupd3Ep2C it would eliminate the duplicated fields and methods. Doesn't remove the ambiguity in the collection context though.

AndrewSisley: I had that originally, I got rid of it because `Name` is duplicated/ambiguous, and that problem is moderately likely to happen with a few other props in the future.

Having to paste a couple of new props every now and then from Collection/SchemaFieldDescription onto FieldDescription is not worth dealing with that mess.

AndrewSisley: I have pushed changes that I am happy with.  I realized part way through that (what was) `FieldDescription` is basically what `CollectionDefinition` is, but at a field level.  So I ended up just cleaning up the `GetFieldByName` funcs and renaming it.

`CollectionDefinition` will eventually suffer from the same (and worse) ambiguity that `FieldDescription.Name` does when embedding, however at the moment we were able to get away with it because there is currently no overlap in properties.  This is almost certainly going to change, and `CollectionDefinition` will need to be changed for that (redeclaration of it's components in a way that eliminates ambiguity over what should be used, for example a single `Fields []FieldDefinition` property, containing a merged set of collection and schema fields).

client/descriptions.go
AndrewSisley: That is quite cheap, I'm not worried about that.

client/descriptions.go
AndrewSisley: > and that problem is moderately likely to happen with a few other props in the future

To expand on this, I see the following properties likely to clash, and they may have different values on schema vs collection:

- `Name` (if/when it becomes a label this will be much worse, where it's value may be different on schema vs collection)
- `Description`, I definitely want this on Collection, and there is value on having one on Schema too (overridable by collection
- `RelationName`, I want this on Collection, but we could also allow it to be defined on Schema in the case where users wish to include both halves of a relation on the global schema.  In such a case I'm not sure we'd want to allow different values to be set, but we could also allow it to be overrideable like `Description` and `Name`.
- Potentially ACP field level stuff too, I can imaging people might want to be able to set field level stuff on both the schema, and allow local collections to define additional rules on top of that.

client/definitions.go
```diff
@@ -0,0 +1,132 @@
...
+
+// FieldDefinition describes the combined local and global set of properties that constitutes
+// a field on a collection.
+//
+// It draws it's information from the [CollectionFieldDescription] on the [CollectionDescription],
+// and the [SchemaFieldDescription] on the [SchemaDescription].
+//
+// It is to [CollectionFieldDescription] and [SchemaFieldDescription] what [CollectionDefinition]
+// is to [CollectionDescription] and [SchemaDescription].
+type FieldDefinition struct {
```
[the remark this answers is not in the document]
AndrewSisley: Will add something (and maybe to CollectionDefinition too), thanks Islam :)

- [ ] expand FieldDefinition

client/descriptions.go
AndrewSisley: Islam (https://github.com/sourcenetwork/defradb/pull/2339#discussion_r1503957788):

todo: I feel like it's wrong having almost identical structs with `SchemaFieldDescription`.

Moreover, you draw an analogy with collectin/schema description, but these 2 are embedded into `CollectionDefinition`.
But this one is not. It's inconsistent. I you think that `CollectionFieldDescription` is too small and has only 1 unique field `ID` we can make it:
```go
type FieldDefinition struct {
    schemaField SchemaFieldDescription
    collectionFieldID int
}
```
And you won't need to copy&past all method below (`IsObject`, `IsObjectArray`, ...)

client/index.go
```diff
@@ -41,18 +41,19 @@ type IndexDescription struct {
...
-			for i := range schema.Fields {
-				colField := schema.Fields[i]
-				if field.Name == colField.Name && !fieldsMap[field.Name] {
-					fieldsMap[field.Name] = true
-					fields = append(fields, colField)
-					break
-				}
+			if fieldsMap[field.Name] {
+				// If the FieldDescription has already been added to the result do not add it a second time
+				// this can happen if a field if referenced by multiple indexes
```
AndrewSisley, line 51: Islam (https://github.com/sourcenetwork/defradb/pull/2339#discussion_r1503961988):

typo: "if a field referenced"

planner/mapper/mapper.go
```diff
@@ -267,7 +267,7 @@ func resolveAggregates(
 	inputFields []Requestable,
 	mapping *core.DocumentMapping,
 	collectionName string,
-	schema client.SchemaDescription,
+	schema client.CollectionDefinition,
```
AndrewSisley, line 270: Islam (https://github.com/sourcenetwork/defradb/pull/2339#discussion_r1503974819):

todo: rename the param to colDef or something similar please

AndrewSisley: My bad, thanks for spotting :)

- [ ] Rename param

client/descriptions.go
AndrewSisley: > Moreover, you draw an analogy with collectin/schema description, but these 2 are embedded into CollectionDefinition.

They shouldn't be long term, please see my earlier comment:

> This is almost certainly going to change, and CollectionDefinition will need to be changed for that (redeclaration of it's components in a way that eliminates ambiguity over what should be used, for example a single Fields []FieldDefinition property, containing a merged set of collection and schema fields).

> And you won't need to copy&past all method below (IsObject, IsObjectArray, ...)

I consider copy-paste here to be significantly cheaper than detangling inheritance hierarchies, especially coupled with how embedding structs in Go works (is half flattened, but not quite, so the end interface is not the same as declaring the properties in place).  And as mentioned further up, these property sets atm are very similar because only one has been moved, they will diverge more so in the future.

client/descriptions.go
islamaliev: I don't think you can justify copy&paste by performance characteristics.

I'd say you either make one struct embed the other or create something like `client.IsArrayKind(Kind)`

AndrewSisley: cheap in terms of dev cost, performance is not relevant here

AndrewSisley: `client.IsArrayKind(Kind)` or `Kind.IsArray()` is a nice thought though, can do the same for the others too.

- [ ] Move Kind related funcs

client/descriptions.go
AndrewSisley: > And you won't need to copy&past all method below (IsObject, IsObjectArray, ...)

Discussion point bypassed by Islam's suggestion of moving these functions onto FieldKind.

client/descriptions.go
```diff
@@ -245,6 +238,31 @@ func (f FieldKind) String() string {
 	}
 }
 
+// IsObject returns true if this FieldKind is an object type.
+func (f FieldKind) IsObject() bool {
+	return (f == FieldKind_FOREIGN_OBJECT) ||
+		(f == FieldKind_FOREIGN_OBJECT_ARRAY)
```
fredcarle, line 244: thought: I know this is copy-pasted but I don't think an array of object is an object. One contains the other but the one isn't the other.

fredcarle (approved):
I have to say I'm still struggling with the `FieldDefinition` but the latest changes to the PR make it overall much better than it was initially. I'll give you the benefit of the doubt as to the future benefits that `FieldDefinition` will bring.

Approving with a few minor comments to resolve before merge.

client/descriptions.go
```diff
@@ -245,6 +238,31 @@ func (f FieldKind) String() string {
...
+}
+
+// IsObjectArray returns true if this FieldKind is an object array type.
+func (f FieldKind) IsObjectArray() bool {
+	return (f == FieldKind_FOREIGN_OBJECT_ARRAY)
+}
+
+// IsArray returns true if this FieldKind is an array type which includes inline arrays as well
+// as relation arrays.
+func (f FieldKind) IsArray() bool {
```
islamaliev, line 254: praise: looks much better now

islamaliev (approved):
Overall looks good. But as with Fred, this `FieldDescription` makes me a bit uncomfortable.
Approving with the promise that it will get better in the near future.

client/descriptions.go
AndrewSisley: I'll quickly check if any tests start failing if I change this, otherwise I will leave it as-is for now to limit scope creep.

Although... Having written that, I would probably prefer this function to be removed if it is only checking a single Kind - will see what happens.

- [ ] Have a look, but don't spend too much time

client/descriptions.go
```diff
@@ -245,6 +238,31 @@ func (f FieldKind) String() string {
 	}
 }
 
+// IsObject returns true if this FieldKind is an object type.
+func (f FieldKind) IsObject() bool {
+	return (f == FieldKind_FOREIGN_OBJECT) ||
```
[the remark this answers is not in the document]
AndrewSisley: Will do

- [ ] Remove parentheses

client/descriptions.go
```diff
@@ -245,6 +238,31 @@ func (f FieldKind) String() string {
...
+		(f == FieldKind_FOREIGN_OBJECT_ARRAY)
+}
+
+// IsObjectArray returns true if this FieldKind is an object array type.
+func (f FieldKind) IsObjectArray() bool {
+	return (f == FieldKind_FOREIGN_OBJECT_ARRAY)
```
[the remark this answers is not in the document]
AndrewSisley: Will do

- [ ] Remove parentheses

client/descriptions.go
AndrewSisley: Lots of tests start failing, leaving as-is.  If we want to change that it can be done so in another PR.

fredcarle: I created an issue for it. https://github.com/sourcenetwork/defradb/issues/2356

PR #2337, opened 2024-02-23
feat: Introduce Access Control Policy

shahzadlone:
## Relevant issue(s)
Resolves #TBD

## Description
There are some pieces of the puzzle missing, PR is opened upon request for high-level feedback only:
- Signatures (That branch will merge here and likely will merge some of it's commits in a separate PR).
- Some CLI Implementations (add policy)
- Testing Combinations with other features
- Test persistence with DB restart.
- Remove logging from acp package to outside where the functions are called.
- Enable In-Memory ACP Support
  - Remove In-Memory Hack  
- Some other todos / history re-writes

## Tasks

## How has this been tested?
- [ ] Describe all types of tests

- Manjaro WSL2


db/db.go
```diff
@@ -71,12 +73,20 @@ type db struct {
...
+	// ACP module
+	acp immutable.Option[acp.ACPModule]
 }
 
 // Functional option type.
 type Option func(*db)
 
-const updateEventBufferSize = 100
+// WithACPModule enables access control. If path is empty, then acp module runs in-memory.
+func WithACPModule(path string) Option {
```
AndrewSisley, line 85: question: The initial ACP PR will only introduce a local variant of ACP, it is not making remote calls?

db/fetcher/fetcher_acp.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/sourcenetwork/defradb/acp"
+	"github.com/sourcenetwork/defradb/db/permission"
+)
+
+// runDocReadPermissionCheck handles the checking (while fetching) if the document has read access
+// or not, according to our access logic based on weather (1) the request is permissioned,
+// (2) the collection is permissioned (has a policy), (3) acp module exists.
+func (df *DocumentFetcher) runDocReadPermissionCheck(ctx context.Context) error {
+	hasPermission, err := permission.CheckDocPermissionedAccessOnCollection(
```
AndrewSisley, line 24: question: How expensive do you see this call being?  Do you ever see a need for it to be aysnc at somepoint?

suggestion: It does at first glance look like it might be better to instead do the acp stuff in a new node, the fetcher is already quite painful to maintain, and a new node feels more aligned with the spirit of the planner package. Less importantly, this also allows turning it on/off or replacing it to be evaluated during plan construction, instead of the main iteration loop (slightly easier for us, and the machine).

shahzadlone: Yes the asyc aspects are on the roadmap. For now this uses the embedding local acp module.

shahzadlone:
I force push some updated refs, which ended up closing the PR, I am not sure that happened.

shahzadlone:
I force push some updated refs, which ended up closing the PR, I am not sure that happened.

db/fetcher/fetcher_acp.go
AndrewSisley: Do you see the location code flow around the calling of this function to remain the same when that time comes? Or is the code flow here likely to be very temporary?  And do you have a rough idea of what that new code structure might look like? 

The thought of a blocking async call per document inside the fetcher is not a happy one.

shahzadlone:
Wont allow me to re-open, I am not sure if this is a bug with github or something is off

db/fetcher/fetcher_acp.go
shahzadlone: replying here: https://github.com/sourcenetwork/defradb/pull/2338

db/db.go
shahzadlone: Correct.

PR #2385, opened 2024-03-06
fix: Make returned collections respect explicit transactions

### LINKED ISSUE #2384: Collections returned from `db.getCollections` don't have a loaded transaction
`db.getCollections` take a transaction as a parameter but the returned collections aren't loaded with that transaction. This means that subsequent operations using the returned collection will operate on an implicit transaction.

AndrewSisley:
## Relevant issue(s)

Resolves #2384

## Description

Makes returned collections respect explicit transactions.

I have not added any new tests in this PR as there is not much time before the release, there is currently no test actions to do facilitate this, and it is a fairly safe change.  I might add tests in a new PR before 0.10.0 is cut if there is time.

### DIFF
--- a/client/db.go
+++ b/client/db.go
@@ -187,6 +187,9 @@ type Store interface {
 	// GetCollectionByName attempts to retrieve a collection matching the given name.
 	//
 	// If no matching collection is found an error will be returned.
+	//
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
LGTM!

PR #2390, opened 2024-03-07
fix: Make peers sync secondary index

### LINKED ISSUE #2379: Secondary indexes does not appear to work with P2P
For example, the final action below fails as the query returns no results.  Deleting the `CreateIndex` test action causes the test to pass.

The order in which the fields are declared in the index does not affect the results.  Using a single-field index still results in failure, suggesting this affects 0.9.0 and all indexes on P2P synced data.

Test is not commited to develop branch.

```
func TestP2PCreateWithP2PCollection2(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type Users {
						Name: String
						Age: Int
					}
				`,
			},
			testUtils.CreateIndex{
				CollectionID: 0,
				IndexName:    "name_age_index",
				Fields:       []testUtils.IndexedField{{Name: "Name"}, {Name: "Age"}},
			},
			testUtils.ConnectPeers{
				SourceNodeID: 1,
				TargetNodeID: 0,
			},
			testUtils.SubscribeToCollection{
				NodeID:        1,
				CollectionIDs: []int{0},
			},
			testUtils.CreateDoc{
				NodeID: immutable.Some(0),
				Doc: `{
					"Name": "Fred",
					"Age": 31
				}`,
			},
			testUtils.WaitForSync{},
			testUtils.Request{
				NodeID: immutable.Some(1),
				Request: `query {
					Users (filter: {Age: {_eq: 31}, Name: {_eq: "Fred"}}){
						Age
					}
				}`,
				Results: []map[string]any{
					{
						"Age": int64(31),
					},
				},
			},
		},
	}

}
```

islamaliev:
## Relevant issue(s)

Resolves #2379

## Description

Make peers updated update local storage when an update happens on docs.


client/db.go
```diff
@@ -215,6 +215,15 @@ type Store interface {
 	// GetAllIndexes returns all the indexes that currently exist within this [Store].
 	GetAllIndexes(context.Context) (map[CollectionName][]IndexDescription, error)
 
+	// CreateDocIndex creates an index for the given document.
+	CreateDocIndex(context.Context, Collection, *Document) error
```
AndrewSisley, line 219: suggestion: Are you sure you want 3 public functions for this?  Is there not only ever one valid index state for a document at any given time, and would this not be more safely handled by a `RefreshIndex(doc)`-like function, that decides for the caller what underlying action(s) should be taken?

islamaliev: it's not possible. Indexes operate on values, on ids. To update an indexed document you need to provide a doc with the old value to delete from index and a doc with a new value to add.
Proposed `RefreshIndex` receives only one doc.

AndrewSisley: Ahhh, true - that's a bit annoying.  I wonder if we can get around that somehow by using something clever like tracking the txn or using events, but that can be looked at later, not in the last couple of days before release :)

Thanks for the explanation :)

AndrewSisley, line 219: suggestion: It feels like hosting these functions on `client.Collection` would be nicer, given that they have to take a `Collection` as a param anyway.

db/collection_index.go
```diff
@@ -107,6 +107,26 @@ func (db *db) fetchCollectionIndexDescriptions(
 	return indexDescriptions, nil
 }
 
+func (db *db) getCollectionIndexes(
+	ctx context.Context,
+	txn datastore.Txn,
+	col client.Collection,
+) ([]CollectionIndex, error) {
+	indexDescriptions, err := db.fetchCollectionIndexDescriptions(ctx, txn, col.ID())
```
AndrewSisley, line 115: todo: Are you sure this is necessary?  The IndexDescriptions should already be on the `col.Description()` that you are passing into this function.

client/db.go
islamaliev: that was our initial approach (Fred also took part in this design), but with all these http and cli wrappers it turned out to be more complicated. That's why wee took this route

db/txn_db.go
```diff
@@ -174,14 +174,154 @@ func (db *implicitTxnDB) GetAllIndexes(
...
+		return err
+	}
+	defer txn.Discard(ctx)
+
+	indexes, err := db.getCollectionIndexes(ctx, txn, col)
+	if err != nil {
+		return err
+	}
+
+	for _, index := range indexes {
```
AndrewSisley, line 204: todo: Please don't include any non transaction related logic in this file - instead move this block into a `db.createDocIndex` func (along with the others).

This allows us to keep the transaction boiler-plate stuff separate from the more complex business logic.

http/client.go
```diff
@@ -434,3 +434,28 @@ func (c *Client) Events() events.Events {
 func (c *Client) MaxTxnRetries() int {
 	panic("client side database")
 }
+
+func (c *Client) CreateDocIndex(
+	ctx context.Context,
+	col client.Collection,
+	doc *client.Document,
+) error {
+	panic("client side database")
```
AndrewSisley, line 443: todo: Please implement these :)

tests/clients/cli/wrapper.go
```diff
@@ -522,3 +522,28 @@ func (w *Wrapper) WaitForPushLogByPeerEvent(id peer.ID) error {
 func (w *Wrapper) WaitForPushLogFromPeerEvent(id peer.ID) error {
 	return w.node.WaitForPushLogFromPeerEvent(id)
 }
+
+func (w *Wrapper) CreateDocIndex(
+	ctx context.Context,
+	col client.Collection,
+	doc *client.Document,
+) error {
+	panic("client side database")
```
AndrewSisley, line 531: todo: Please implement these :)

AndrewSisley (changes requested):
Looks good, but I think there are few bits that need sorting out before merge

client/db.go
AndrewSisley: Why was that more complicated?  The bulk of the logic for handling collection stuff already exists in http/cli

db/txn_db.go
islamaliev: done

tests/clients/cli/wrapper.go
islamaliev: this is not supposed to be called from outside

http/client.go
AndrewSisley: We have plenty of http/cli funcs that receive docs as params

tests/clients/cli/wrapper.go
AndrewSisley: Then it shouldn't be in the `client` package

db/collection_index.go
islamaliev: you are right. Adjusted. Nice catch!

client/db.go
islamaliev: moved

tests/clients/cli/wrapper.go
islamaliev: Added error as discussed in syncup

http/client.go
islamaliev: Added error as discussed in syncup

db/collection_index.go
AndrewSisley: nice! thanks :)

client/collection.go
```diff
@@ -169,6 +169,21 @@ type Collection interface {
...
 	// GetIndexes returns all the indexes that exist on the collection.
 	GetIndexes(ctx context.Context) ([]IndexDescription, error)
+
+	// CreateDocIndex creates an index for the given document.
+	// WARNING: This method is only for internal use and is not supposed to be called by the client
+	// as it might compromise the integrity of the database. This method will be removed in the future
+	CreateDocIndex(context.Context, *Document) error
+
+	// UpdateDocIndex updates the index for the given document.
+	// WARNING: This method is only for internal use and is not supposed to be called by the client
```
AndrewSisley, line 179: praise: These are suitably scary - thanks :grin:

fredcarle: I second that 👍

db/collection_index.go
```diff
@@ -107,6 +107,55 @@ func (db *db) fetchCollectionIndexDescriptions(
...
+	if err != nil {
+		return err
+	}
+	defer c.discardImplicitTxn(ctx, txn)
+
+	err = c.deleteIndexedDoc(ctx, txn, oldDoc)
+	if err != nil {
+		return err
+	}
+	err = c.indexNewDoc(ctx, txn, newDoc)
```
AndrewSisley, line 136: nitpick: I do see `delete+new` as still being business logic, but only at a nitpick level - so feel free to merge as is :)

fredcarle (approved):
LGTM!

net/server.go
```diff
@@ -338,6 +327,64 @@ func (s *server) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.PushL
...
+		},
+	)
+	if err != nil {
+		return nil, errors.Wrap(fmt.Sprintf("Failed to get collection from schemaRoot %s", schemaRoot), err)
+	}
+	if len(cols) == 0 {
+		return nil, client.NewErrCollectionNotFoundForSchema(schemaRoot)
+	}
+	var col client.Collection
+	for _, c := range cols {
```
AndrewSisley, line 348: praise: Thanks for moving this somewhere out of the way :)

AndrewSisley (approved):
LGTM - thanks Islam :)

db/collection_index.go
fredcarle: Can you expand on what is your nitpick concern here Andy?

AndrewSisley: I have none and thought I was in the txn_db.go file :)

AndrewSisley: Ah I see I how I got confused - they have moved to collection, and I new that they had moved, but my brain still thought of the implementation as being in the txn_db file when I saw the txn logic.  I wonder if we can do a similar job to collection that we did to client.Store and split out the txn boilerplate to another file sometime

PR #2395, opened 2024-03-09
fix(i): Move cache action after goreleaser action

### LINKED ISSUE #2394: Release CI workflow trying to save cache with no content

fredcarle:
## Relevant issue(s)

Resolves #2394 

## Description

This PR moves the cache action in the release workflow to be after the goreleaser action. After the last release, we changed the cache action to use `action/cache/save` instead of just `action/cache`. The difference is that `action/cache` acts like a defer function and  gets actioned at the end of a job. `action/cache/save` is meant to be more granular and tries to save when called. This caused a situation where it tried to save in the cache a path that didn't exist.

## How has this been tested?

tested the workflow on my personal fork

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,11 +71,17 @@ jobs:
         shell: bash
         run: echo "sha_short=$(git rev-parse --short HEAD)" >> ${GITHUB_ENV}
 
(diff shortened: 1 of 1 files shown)


shahzadlone (approved):
Thanks @fredcarle much better :)

LGTM

PR #2396, opened 2024-03-09
Release v0.10.0

fredcarle:
Release branch as of 21b30693fc04310bc900dbb9aebd18154bc147d1

Release tasks completed:
- [x] ran `go mod tidy`
- [x] regenerated in-repo docs (CLI and manpages)
- [x] bump BSL license parameters
- [x] update changelog
- [x] Made sure that the Playground can build

Only need to review the last commit :)

~~This is a redo since building the Playground failed when running the releaser.~~
This is a second redo due to a regression bug in our release workflow. Would have preferred to avoid doing this but since we use CGO, it's simpler to have the binaries compiled on the different architectures through the CI.


jsimnz (approved):
lgtm

shahzadlone (approved):
LGTM. LETS SHIP IT!

PR #2402, opened 2024-03-11
feat: Add PatchCollection

### LINKED ISSUE #2389: Add `PatchCollection` function to `Store`
Like `PatchSchema`, but for `CollectionDescription`s.

This is a prerequisite for moving more stuff from schema onto collection, as users will need a means to set the moved properties.

Should eventually allow the setting of indexes and ACP stuff too, but it is fine to block those off initially in order to reduce scope of the initial PR.

As the only to-be mutable property that is not index/acp and currently already on collection description is col/field name, and those properties comes with higher risk, it might be best to do this alongside https://github.com/sourcenetwork/defradb/issues/1386 or https://github.com/sourcenetwork/defradb/issues/1385 to keep things simple to start off with.

AndrewSisley:
## Relevant issue(s)

Resolves #2389

## Description

Adds the PatchCollection command.

Mutating anything but the collection name is currently disabled, we can expand this as we see fit, but for now I'd prefer to keep the initial PR small.

This change means that the Collection Name is no longer always going to be the same as the Schema Name.

I've manually tested the OpenApi stuff via playground.

### DIFF
--- a/cli/collection_patch.go
+++ b/cli/collection_patch.go
@@ -37,6 +37,7 @@ Example: patch from stdin:
   cat patch.json | defradb client collection patch -
 
 To learn more about the DefraDB GraphQL Schema Language, refer to https://docs.source.network.`,
+		Args: cobra.RangeArgs(0, 1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			store := mustGetContextStore(cmd)
 
@@ -54,7 +55,7 @@ To learn more about the DefraDB GraphQL Schema Language, refer to https://docs.s
 					return err
 				}
 				patch = string(data)
-			case len(args) >= 1:
+			case len(args) == 1:
 				patch = args[0]
 			default:
 				return fmt.Errorf("patch cannot be empty")
--- a/cli/collection_patch.go
+++ b/cli/collection_patch.go
@@ -31,7 +31,7 @@ Example: patch from an argument string:
   defradb client collection patch '[{ "op": "add", "path": "...", "value": {...} }]'
 
 Example: patch from file:
-  defradb client collection patch -f patch.json
+  defradb client collection patch -p patch.json
 
 Example: patch from stdin:
   cat patch.json | defradb client collection patch -
--- a/cli/collection_patch.go
+++ b/cli/collection_patch.go
@@ -1,0 +1,68 @@
+// Copyright 2023 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func MakeCollectionPatchCommand() *cobra.Command {
+	var patchFile string
+	var cmd = &cobra.Command{
+		Use:   "patch [patch]",
+		Short: "Patch existing collection descriptions",
+		Long: `Patch existing collection descriptions.
+
+Uses JSON Patch to modify collection descriptions.
+
+Example: patch from an argument string:
+  defradb client collection patch '[{ "op": "add", "path": "...", "value": {...} }]'
+
+Example: patch from file:
+  defradb client collection patch -f patch.json
+
+Example: patch from stdin:
+  cat patch.json | defradb client collection patch -
+
+To learn more about the DefraDB GraphQL Schema Language, refer to https://docs.source.network.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			store := mustGetContextStore(cmd)
+
+			var patch string
+			switch {
+			case patchFile != "":
+				data, err := os.ReadFile(patchFile)
+				if err != nil {
+					return err
+				}
+				patch = string(data)
+			case len(args) > 0 && args[0] == "-":
+				data, err := io.ReadAll(cmd.InOrStdin())
+				if err != nil {
+					return err
+				}
+				patch = string(data)
+			case len(args) >= 1:
+				patch = args[0]
+			default:
+				return fmt.Errorf("patch cannot be empty")
+			}
+
+			return store.PatchCollection(cmd.Context(), patch)
+		},
+	}
+	cmd.Flags().StringVarP(&patchFile, "patch-file", "p", "", "File to load a patch from")
+	return cmd
+}
--- a/cli/collection_patch.go
+++ b/cli/collection_patch.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Democratized Data Foundation
+// Copyright 2024 Democratized Data Foundation
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt.
--- a/client/db.go
+++ b/client/db.go
@@ -120,6 +120,17 @@ type Store interface {
 	// A lens configuration may also be provided, it will be added to all collections using the schema.
 	PatchSchema(context.Context, string, immutable.Option[model.Lens], bool) error
 
+	// PatchCollection takes the given JSON patch string and applies it to the set of CollectionDescriptions
+	// present in the database.
+	//
+	// It will also update the GQL types used by the query system. It will error and not apply any of the
+	// requested, valid updates should the net result of the patch result in an invalid state.  The
+	// individual operations defined in the patch do not need to result in a valid state, only the net result
+	// of the full patch.
+	//
+	// Currently only the collection name can be modified.
+	PatchCollection(context.Context, string) error
+
 	// SetActiveSchemaVersion activates all collection versions with the given schema version, and deactivates all
 	// those without it (if they share the same schema root).
 	//
--- a/db/collection.go
+++ b/db/collection.go
@@ -13,10 +13,13 @@ package db
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	"github.com/ipfs/go-cid"
 	ds "github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/query"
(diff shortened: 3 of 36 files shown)


cli/collection_patch.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2023 Democratized Data Foundation
```
shahzadlone, line 1: todo: 
```suggestion
// Copyright 2024 Democratized Data Foundation
```

cli/collection_patch.go
```diff
@@ -0,0 +1,68 @@
...
+				}
+				patch = string(data)
+			case len(args) > 0 && args[0] == "-":
+				data, err := io.ReadAll(cmd.InOrStdin())
+				if err != nil {
+					return err
+				}
+				patch = string(data)
+			case len(args) >= 1:
+				patch = args[0]
```
shahzadlone, line 58: question: In what case can `len(args)` > 1?

client/db.go
```diff
@@ -120,6 +120,17 @@ type Store interface {
 	// A lens configuration may also be provided, it will be added to all collections using the schema.
 	PatchSchema(context.Context, string, immutable.Option[model.Lens], bool) error
 
+	// PatchCollection takes the given JSON patch string and applies it to the set of CollectionDescriptions
+	// present in the database.
+	//
+	// It will also update the GQL types used by the query system. It will error and not apply any of the
+	// requested, valid updates should the net result of the patch result in an invalid state.  The
+	// individual operations defined in the patch do not need to result in a valid state, only the net result
```
shahzadlone, line 128: question: what is meant by `valid state` here.
```suggestion
	// individual operations defined in the patch do not need to result in a `valid state`, only the net result
```

client/db.go
```diff
@@ -120,6 +120,17 @@ type Store interface {
 	// A lens configuration may also be provided, it will be added to all collections using the schema.
 	PatchSchema(context.Context, string, immutable.Option[model.Lens], bool) error
 
+	// PatchCollection takes the given JSON patch string and applies it to the set of CollectionDescriptions
+	// present in the database.
+	//
+	// It will also update the GQL types used by the query system. It will error and not apply any of the
+	// requested, valid updates should the net result of the patch result in an invalid state.  The
```
shahzadlone, line 127: todo: This sentence reads very odd to me "It will error and not apply any of the requested, valid updates should the net result of the patch result in an invalid state."

Please reword.

db/collection.go
```diff
@@ -13,10 +13,13 @@ package db
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
+	jsonpatch "github.com/evanphx/json-patch/v5"
```
shahzadlone, line 22: question: I forgot if we cared about camel casing package import alias names. If we do then `jsonPatch`?

db/collection.go
```diff
@@ -526,6 +529,283 @@ func validateUpdateSchemaFields(
 	return hasChanged, nil
 }
 
+func (db *db) patchCollection(
```
shahzadlone, line 532: suggestion: I would much rather than patch stuff be moved to a new file, like: `db/collection_patch.go`

db/collection.go
```diff
@@ -526,6 +529,283 @@ func validateUpdateSchemaFields(
...
+	oldColsByID map[uint32]client.CollectionDescription,
+	newColsByID map[uint32]client.CollectionDescription,
+) error {
+	for _, newCol := range newColsByID {
+		oldCol, ok := oldColsByID[newCol.ID]
+		if !ok {
+			continue
+		}
+
+		// DeepEqual is temporary, as this validation is temporary
```
shahzadlone, line 684: question: Why is this temporary? Will the use of reflection package be removed?

db/collection.go
```diff
@@ -526,6 +529,283 @@ func validateUpdateSchemaFields(
...
+
+func validateCollectionNotRemoved(
+	oldColsByID map[uint32]client.CollectionDescription,
+	newColsByID map[uint32]client.CollectionDescription,
+) error {
+oldLoop:
+	for _, oldCol := range oldColsByID {
+		for _, newCol := range newColsByID {
+			// It is not enough to just match by the map index, in case the index does not pair
+			// up with the ID (this can happen if a user moves it)
```
shahzadlone, line 797: question: `moves it` moves what?

tests/integration/collection_description/updates/add/collections_test.go
```diff
@@ -0,0 +1,85 @@
...
+					type Users {}
+				`,
+			},
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{ "op": "add", "path": "/2", "value": {"Name": "Dogs"} }
+					]
+				`,
+				ExpectedError: "collection ID cannot be zero",
```
shahzadlone, line 33: Would add another test where it is explicitly `0`:
`{ "op": "add", "path": "/2", "value": {"ID": 0, "Name": "Dogs"} }`

tests/integration/test_case.go
```diff
@@ -97,6 +97,17 @@ type SchemaPatch struct {
 	ExpectedError string
 }
 
+type PatchCollection struct {
+	// NodeID may hold the ID (index) of a node to apply this patch to.
+	//
+	// If a value is not provided the patch will be applied to all nodes.
+	NodeID immutable.Option[int]
+
+	Patch string
```
shahzadlone, line 106: todo: document please

cli/collection_patch.go
AndrewSisley: :grin: Will do

- [ ] 2024

cli/collection_patch.go
AndrewSisley: I don't really care :)  But I've added a param limit, arg length is now 0 or 1.

client/db.go
AndrewSisley: A valid CollectionDescription

client/db.go
AndrewSisley: What is odd about it?  Do you have a rough alternative in mind?

I think it is important to highlight that the full patch will be rolled back if an given patch item fails - especially given that SQLs do not do this and require manual rollback when given multiple DDL statements.

db/collection.go
AndrewSisley: The validation (function) is temporary, which of course means that if this function is removed, the call to DeepEqual will be removed with it.

db/collection.go
AndrewSisley: The collection relative to it's index in the map, I'll replace `it`

- [ ] Replace `it`

tests/integration/collection_description/updates/add/collections_test.go
AndrewSisley: Will add

- [ ] Add explicit zero add test

tests/integration/test_case.go
AndrewSisley: My bad :)  Of course :)

- [ ] Doc test action Patch prop

db/collection.go
AndrewSisley: I don't know either, my IDE did this, and I don't see it as being unreadable

db/collection.go
AndrewSisley: I've opened a ticket: https://github.com/sourcenetwork/defradb/issues/2407

nasdf (changes requested):
One minor fix from me.

client/db.go
shahzadlone: This sounds more clear when you said: `full patch will be rolled back if an given patch item fails` 

This is very wordy: `... apply any of the requested, valid updates should the net resulted`

db/collection.go
shahzadlone: > The validation (function) is temporary, which means that if this function is removed, the call to DeepEqual will be removed with it.

Why is it temporary?

client/db.go
AndrewSisley: `rolled back` is incorrect/lazy lol - the patches are aggregated before being applied - hence the use of `net result of the patch`

tests/integration/collection_description/updates/replace/id_test.go
```diff
@@ -0,0 +1,146 @@
...
+				`,
+			},
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{ "op": "replace", "path": "/1/ID", "value": 2 },
+						{ "op": "replace", "path": "/2/ID", "value": 1 }
+					]
+				`,
+				ExpectedError: "collection sources cannot be mutated.",
```
fredcarle, line 90: thought: The error is a little weird here. It's talking about mutating sources when the patch is dealing with the collection ID.

fredcarle (commented):
Just a few comments to resolve before approval :)

client/db.go
shahzadlone: > the patches are aggregated before being applied

This will be nice to document somewhere too

cli/collection_patch.go
```diff
@@ -0,0 +1,69 @@
...
+			case len(args) == 1:
+				patch = args[0]
+			default:
+				return fmt.Errorf("patch cannot be empty")
+			}
+
+			return store.PatchCollection(cmd.Context(), patch)
+		},
+	}
+	cmd.Flags().StringVarP(&patchFile, "patch-file", "p", "", "File to load a patch from")
```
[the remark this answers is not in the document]
shahzadlone: Don't all other file consumption commands use `--file |-f` why do we want to have a different flag name for this arg (does it clash with `-f` somewhere?)

shahzadlone (approved):
LGTM assuming you will resolve other comments

db/collection.go
AndrewSisley: Because the function is validating that indexes have not been modified, and we very much want them to be mutable in the future.

client/db.go
AndrewSisley: > I though I understood the bahaviour from the prior sentence but this one gets me confused. How can the net result be valid if an individual operation creates an invalid state?

There are many reasons why we want this to be, and many of the tests cover this, for example:

> { "op": "copy", "from": "/Users/Fields/1", "path": "/Users/Fields/2" },
{ "op": "replace", "path": "/Users/Fields/2/Name", "value": "age" },

The first field is copied (invalid as it is now a duplicate), and then renamed (net is now valid).  This allows existing stuff (including full schema/collections) to be used as templates.  There are other use cases too.

> This will be nice to document somewhere too

It is, we are talking about the documentation that currently documents it.

tests/integration/collection_description/updates/replace/id_test.go
AndrewSisley: Yes, unfortunately when mutating IDs it becomes impossible for us to tell what the original object was (and that the ID was mutated), so it will often get caught by other validation rules.

db/collection.go
shahzadlone: Thanks that makes so much sense.

Please document this

tests/integration/collection_description/updates/copy/name_test.go
```diff
@@ -0,0 +1,68 @@
...
+				`,
+				SetAsDefaultVersion: immutable.Some(false),
+			},
+			testUtils.PatchCollection{
+				// Activate the second collection by setting its name to that of the first,
+				// then decativate the original collection version by removing the name
+				Patch: `
+					[
+						{ "op": "copy", "from": "/1/Name", "path": "/2/Name" },
+						{ "op": "remove", "path": "/1/Name" }
```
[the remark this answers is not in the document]
AndrewSisley: I think the value is low, as duplicate names are tested elsewhere, but I will add

- [ ] add copy name test

db/collection.go
```diff
@@ -526,6 +529,283 @@ func validateUpdateSchemaFields(
...
+
+func validateCollectionNotRemoved(
+	oldColsByID map[uint32]client.CollectionDescription,
+	newColsByID map[uint32]client.CollectionDescription,
+) error {
+oldLoop:
+	for _, oldCol := range oldColsByID {
+		for _, newCol := range newColsByID {
+			// It is not enough to just match by the map index, in case the index does not pair
+			// up with the ID (this can happen if a user moves the collection within the map)
```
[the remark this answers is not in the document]
AndrewSisley: > { "op": "move", "from": "/1", "path": "/2" }

This moves the collection within the map, without mutating the ID.

cli/collection_patch.go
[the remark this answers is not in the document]
AndrewSisley: `PatchSchema` uses `p` because it it takes two kinds of files.  I think it is more important to be consistent with `PatchSchema` than anywhere else here.

db/collection.go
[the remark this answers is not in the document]
fredcarle: Thanks for clarifying.

cli/collection_patch.go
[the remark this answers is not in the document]
shahzadlone: How does it take two files? I see 1 so far with -p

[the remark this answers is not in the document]
AndrewSisley: `PatchSchema` takes two files, I assume you are looking at `PatchCollection`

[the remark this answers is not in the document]
fredcarle: Maybe we should change the flag for patch schema to also be -f and make it consistent throughout. The lens file can stay with -t or maybe -l would make more sense.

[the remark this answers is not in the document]
AndrewSisley: `f` and `file` is ambiguous for a command that takes two different files.  I do not like that.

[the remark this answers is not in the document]
fredcarle: @nasdf your opinion would be appreciated here when you have a chance.

nasdf (approved):

cli/collection_patch.go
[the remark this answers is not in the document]
nasdf: I think its fine to keep it `-p` for consistency with the other patch method. Using `-f` when there are multiple files is worse in my opinion.

PR #2404, opened 2024-03-12
refactor: Add NormalValue

### LINKED ISSUE #2403: Have a single place in code that manages standard (or normal) types
At the moment we have few places in code that deal with values of types like `int64`, `int32`, `int`... and converts it to standard on `int64`.

We should have a single place in code that takes care of converting a value of any type into a standard (used by GraphQL) type. So that the value that is being passed around is known be of standard (or normal) type and there is no need to do lengthy type assertions.

islamaliev:
## Relevant issue(s)

Resolves #2403

## Description

This change introduces `NormalValue` type that carries a standard (or normal) value that doesn't need to be type-asserted with all possible types.


client/normal_value.go
```diff
@@ -0,0 +1,591 @@
...
+
+import (
+	"time"
+
+	"github.com/sourcenetwork/immutable"
+	"golang.org/x/exp/constraints"
+)
+
+type NormalValue struct {
+	value any
```
AndrewSisley, line 21: suggestion: It might be worth chatting over discord about this before commiting too much time to polishing it, I have a few concerns that need talking over and others may have some too.  My biggest are below:

1. This means a pointer for every field value
2. Loss of compile time safety

You might actually be able to remove the second concern by using a constrained generic instead of `any`, in most langs this would also resolve (1), but IIRC in Go all generic variables are pointers anyway (I'm not sure, worth checking):
```
type NormalType interface {
  int64 | float64 | string | etc
}

type NormalValue[T NormalType]  struct {
  value T
}
```

fredcarle: > in Go all generic variables are pointers anyway

I don't think that's true. Do you remember where you might have seen this?

islamaliev: > It might be worth chatting over discord

I actually wanted to bring it up yesterday, but the time switch caught me off the guard. And then we want over the time.

> Loss of compile time safety

We use `any` everywhere and so far it wasn't a concern.
In this case `NormalValue` is meant to provide some type safety in more compact and reliable way.
It it mostly a wrapper over the value held by `client.FieldValue` which is also of `any` type.

fredcarle: > It is mostly a wrapper over the value held by client.FieldValue which is also of any type.

Could we not apply all the `NormalValue` methods directly to `FieldValue`?

islamaliev: > in Go all generic variables are pointers anyway

I might need to do some additional research on this, but I hardly believe.
In any language with decent support of generics (like C++ or Rust) they are all resolved at compile time and every instantiation of a generic type is a distinct type.

AndrewSisley: https://deepsource.com/blog/go-1-18-generics-implementation

> In Boxing, "values" are boxed and passed around as references of the polymorphic type, generally using a pointer table, commonly called virtual table. This is how Go's interface are implemented. This generally produces a smaller binary and requires short compile-times but may have a toll on the runtime performance.
> 
> Go generics combines concepts from "monomorphisation" (stenciling) and "boxing" (dynamic dispatch) and is implemented using [GCshape stenciling and dictionaries](https://github.com/golang/proposal/blob/master/design/generics-implementation-dictionaries-go1.18.md). This allows Go to have fast compile times and smaller binaries while having generics.

Go generics are not the same generics that you find in Rust/C#/C++/etc - they have a runtime cost here :(

AndrewSisley: > We use any everywhere and so far it wasn't a concern.

This is true, when I think about it our code is all very reliant on boxing anyway, however I think the proposed code might result in double boxing - e.g. in a `core.Doc.Fields` we have an array of `any`, which here would box `NormalValue`, which in turn boxes the field value.

If we are able to change the type of `core.Doc.Fields` to `NormalValue` we might only have a single box (my preference being the generic, not an any anyway).

client/normal_value.go
```diff
@@ -0,0 +1,1531 @@
...
+		if v, ok := arr[i].(T); ok {
+			result[i] = immutable.Some(v)
+		} else {
+			return nil, NewCanNotNormalizeValue(arr)
+		}
+	}
+	return newNormalNillableArr(result), nil
+}
+
+func NewNormalNil() NormalValue {
```
AndrewSisley, line 1138: thought: I don't think this should exist no?  If a `nil` is required, of an unknown type, would `NormalVoid` not be more appropriate?

client/normal_value.go
```diff
@@ -0,0 +1,1531 @@
...
+	return normalNil{}
+}
+
+func NewNormalBool(val bool) NormalValue {
+	return normalBool{baseNormalValue[bool]{val: val}}
+}
+
+func NewNormalInt[T constraints.Integer | constraints.Float](val T) NormalValue {
```
AndrewSisley, line 1146: praise: This is a really nice func signature :)

client/normal_value.go
```diff
@@ -0,0 +1,1531 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package client
```
AndrewSisley, line 11: suggestion: This file is pretty hard to read, and I can see it getting very disorganized over time, and very hard to find stuff/see what exists.  I suggest splitting it up by type in its own directory within client:

``` client/value
void.go
nil.go
int.go
// etc
```
If you really want the types directly in the client package, you could always re-export them(or just the constructors+interface?) Rust style from a mod.rs style file (client/normal_value.go):
```
type NormalValue = value.NormalValue
func NewNormalInt(...) { // I don't know if there is a better way than this in Go for funcs?
  return value.NormalValue()
}
```

client/normal_value.go
```diff
@@ -0,0 +1,1531 @@
...
+
+import (
+	"time"
+
+	"github.com/sourcenetwork/immutable"
+	"golang.org/x/exp/constraints"
+)
+
+type NormalValue interface {
```
AndrewSisley, line 20: todo: Please document all the public stuff before merge.  If you end up breaking up this file, a doc.go with package level documentation might be nice too.

AndrewSisley (changes requested):
I'm still reviewing, but submitting a few suggestions now that may aid further review :)

fredcarle (commented):
Currently reviewing but I just want to point out that this PR is in conflict with #2414. #2414 should merge first and the conflict can be resolved here.

client/normal_value.go
islamaliev: I also thought about it but then it would need access to `client.Document` and `client` package will also depend on it. You have a suggestion?

client/normal_value.go
islamaliev: will do

client/normal_value.go
fredcarle:
```
type NormalValue = value.NormalValue  
func NewNormalInt(...) { // I don't know if there is a better way than this in Go for funcs?  
  return value.NormalValue()  
} 
```

Please don't 😅.

I'm fine if we go for a client sub-package. `client/normalized` would be my choise. Then we could have the interface`normalized.Value` and the types `normalized.String`, `normalized.Int`, etc.

fredcarle: > I also thought about it but then it would need access to client.Document and client package will also depend on it. You have a suggestion?

No need to bring `client.Document` into the sub package. Just have the `normilized.Value` interface implement for `client.Document` where `client.Document` is defined.

client/normal_value.go
islamaliev: NormalVoid is like a dummy returning default value like `false`. It there just to rid other type of necessary to implement all methods.
And I think having a dedicated `NewNormalNil` more explicitly conveys the intent.

client/normal_value.go
fredcarle: Sorry. I need to point out that for the last comment to work we need to keep the interface in client. So the `normalized` package wouldn't have the `normalized.Value` interface. It would just hold the types.

islamaliev: > No need to bring client.Document into the sub package. Just have the normilized.Value interface implement for client.Document where client.Document is defined.

Then it won't be possible to get normal value from `NewNormalValue` passing in a `*client.Document`. Not sure it's a good trade-off.

client/normal_value.go
AndrewSisley: > NormalVoid is like a dummy returning default value like false.

Yes, that's what I mean, is why I'd prefer it over `NormalNil` - `NormalNil` looks like it is typed, when it is not really - it also returns true for IsNillable no? (unwanted IMO if the type is undefined).

client/normal_value.go
AndrewSisley: An alternative could be to namespace the file names within the client package: `_normal_int.go`, `_normal_value.go`, etc - please dont do this without a namespace though, or we wont be able to find anything else in the directory.

fredcarle: That might be the more sensible approach. No need for the `_` prefix though.

islamaliev: is "normal" bad namespace? `normal_int.go`, `normal_nillable_int.go`...?

fredcarle: > is "normal" bad namespace? `normal_int.go`, `normal_nillable_int.go`...?

I much prefer just `normal_...` 👍

fredcarle: > I would prefer that they are all either at the end alphabetically, or at the start - otherwise the other files will be split on either side.

I'm pretty sure the compiler ignores files with `_` prefixes just like it ignores  files with a `.` prefix.

fredcarle: I confirm https://github.com/golang/go/blob/master/src/go/build/build.go#L548

AndrewSisley: :(

`zzz_normal_int.go` :laughing: ? I will have to tolerate them being in the middle of the package them :(

client/normal_value.go
fredcarle: It seems that everywhere this is used, we have access to the field kind which means that we can get the nillable supported type directly instead of a catch all `nil` type. I think all we would need is an extra `NewNormalValueFromKind` function.

fredcarle: `DecodeFieldValue` would need to take a FieldKind as parameter as a side effect.

tests/integration/index/query_with_index_only_field_order_test.go
```diff
@@ -108,7 +108,7 @@ func TestQueryWithIndex_IfFloatFieldInDescOrder_ShouldFetchInRevertedOrder(t *te
 			testUtils.Request{
 				Request: `
 					query {
-						User(filter: {iq: {_gt: 1}}) {
+						User(filter: {iq: {_lt: 1}}) {
```
fredcarle, line 111: thought: That was a sneeky bug that we should have caught 😅

AndrewSisley (changes requested):
Submitting a handful of documentation requests now, interface looks good :)

client/normal_value.go
```diff
@@ -0,0 +1,134 @@
...
+// All nillable values are represented as immutable.Option[T].
+type NormalValue interface {
+	// Any returns the underlying value.
+	// If the value is nillable the result will be of type `immutable.Option[T]`.
+	Any() any
+	// Unwrap returns the underlying value.
+	// For not nillable values it will act as `Any()`
+	// For nillable values it will return result of `Value()` of `immutable.Option` if it
+	// has value, otherwise it will return `nil`.
+	Unwrap() any
```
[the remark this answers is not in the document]
fredcarle: > I would not expect it to return `immutable.Option`

It doesn't. Tt returns the value of the `immutable.Option`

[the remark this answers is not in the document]
AndrewSisley: Yes... How did I miss read that - maybe, my eyes half read the docs for Any part way through...

client/normal_value.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// IsNil returns if the value is nil.
+	IsNil() bool
+	// IsNillable returns if the value can be nil.
+	IsNillable() bool
+
+	// Bool returns the value as a bool.
+	Bool() (bool, bool)
+	// Int returns the value as an int64.
```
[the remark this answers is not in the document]
islamaliev: nothin is converted there because the value is stored as `int64` all conversions happen either in `NewNormalValue` or `NewNormalInt`

client/normal_value.go
```diff
@@ -0,0 +1,134 @@
...
+	// String returns the value as a string.
+	String() (string, bool)
+	// Bytes returns the value as a byte slice.
+	Bytes() ([]byte, bool)
+	// Time returns the value as a time.Time.
+	Time() (time.Time, bool)
+	// Document returns the value as a *Document.
+	Document() (*Document, bool)
+
+	// NillableBool returns the value as a nillable bool.
```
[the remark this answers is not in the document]
islamaliev: there is actually a note in the documentation of the interface itself. I thought it would be enough.

client/normal_value.go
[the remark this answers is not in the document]
AndrewSisley: Yes, but can can you please document that on the interface?

client/normal_value.go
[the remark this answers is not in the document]
AndrewSisley: Ah true, I know it is a hassle, but can you please document it on the funcs too - I think that is best practice, and it is the first place where people will look most of the time to try and find that out.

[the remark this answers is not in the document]
AndrewSisley: I think the time spent now will be much smaller than the sum of all the head scratching (or bugs) that could take place if people have to chase down this info (or guess and potentially get it wrong)

client/value.go
```diff
@@ -62,17 +67,16 @@ func (val FieldValue) Bytes() ([]byte, error) {
 	}
 
 	var value any
-	switch tempVal := val.value.(type) {
```
AndrewSisley, line 65: thought: This seems very broken and very much looks like a bug to me.  Please let me know if you know it is not a bug.

core/encoding.go
```diff
@@ -245,51 +244,33 @@ func DecodeIndexDataStoreKey(
 
 		i := len(key.Fields)
 		descending := false
+		var kind client.FieldKind = client.FieldKind_DocID
 		// If the key has more values encoded then fields on the index description, the last
 		// value must be the docID and we treat it as a string.
+		if i < len(fields) {
```
AndrewSisley, line 250: suggestion: This check seems unnecessary - did something break without it?  How could it ever happen?

Something would have to be very very broken for this to ever be false no?

I suggest removal of the `if`.

todo: The comment above has been moved away from the line upon which it is commenting, please move it back :)

core/encoding.go
```diff
@@ -245,51 +244,33 @@ func DecodeIndexDataStoreKey(
...
-		var val any
-		data, val, err = encoding.DecodeFieldValue(data, descending)
+		var val client.NormalValue
+		data, val, err = encoding.DecodeFieldValue(data, descending, kind)
 		if err != nil {
 			return IndexDataStoreKey{}, err
 		}
+		if i == len(indexDesc.Fields) {
+			if _, ok := val.String(); !ok {
+				return IndexDataStoreKey{}, NewErrInvalidFieldValue("docID is not a string")
```
AndrewSisley, line 266: todo: Please document how/when this can happen, or remove it.  I am struggling to imagine how this might happen (besides manually messing around in the datastore).

core/key_test.go
```diff
@@ -326,14 +332,14 @@ func TestDecodeIndexDataStoreKey(t *testing.T) {
 }
 
 func TestDecodeIndexDataStoreKey_InvalidKey(t *testing.T) {
-	replace := func(b []byte, i int, v byte) []byte {
+	/*replace := func(b []byte, i int, v byte) []byte {
```
AndrewSisley, line 335: todo: Please remove or uncomment the commented out tests.

client/normal_value.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// NormalValue is the interface for the normal value types.
+// It is used to represent the normal (or standard) values across the system and to avoid
+// asserting all possible types like int, int32, int64, etc.
+//
+// All methods returning a specific type returns the value and the second boolean flag indicating
+// if the value is of the requested type.
+//
+// All nillable values are represented as immutable.Option[T].
```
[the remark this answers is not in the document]
islamaliev: ok

client/normal_value.go
```diff
@@ -0,0 +1,131 @@
...
+	NillableString() (immutable.Option[string], bool)
+	// NillableBytes returns the value as a nillable byte slice.
+	NillableBytes() (immutable.Option[[]byte], bool)
+	// NillableTime returns the value as a nillable time.Time.
+	NillableTime() (immutable.Option[time.Time], bool)
+	// NillableDocument returns the value as a nillable *Document.
+	NillableDocument() (immutable.Option[*Document], bool)
+
+	// IsArray returns if the value is an array.
+	IsArray() bool
```
AndrewSisley, line 70: suggestion: `IsArray` might be slightly nicer to have futher up this definition, next to `IsNillable`.

db/fetcher/indexer_iterators.go
```diff
@@ -274,108 +274,75 @@ func (iter *scanningIndexIterator) Next() (indexIterResult, error) {
...
-	if !ok {
-		return false, NewErrUnexpectedTypeValue[int64](value)
+func (m *intMatcher) Match(value client.NormalValue) (bool, error) {
+	intVal, ok := value.Int()
+	if ok {
+		return m.evalFunc(intVal, m.value), nil
+	}
+	intOptVal, ok := value.NillableInt()
+	if ok {
+		return m.evalFunc(intOptVal.Value(), m.value), nil
```
AndrewSisley, line 292: question: This looks like a potential behaviour change to me, we are not checking for `nil` here - is that alright here?

suggestion: Maybe a func like the below would be handy (annoyingly generic funcs cant be defined on Go interfaces or we could replace most funcs on it with a couple of generic ones...):
```
func AsT[T client.NormalValueConstraints](v NormalValue) (T, bool) {
  any := v.Unwrap()
  return any.(T)
}
```
It could either live here as a private func, or in the client package, then in the funcs here you wouldn't need two ifs, you could just:
```
func (m *intMatcher) Match(value client.NormalValue) (bool, error) 
  intVal, ok := value.AsT[int64]()
  if ok {
    return m.evalFunc(intVal, m.value), nil
  }
  return false, NewErrUnexpectedTypeValue[int64](value)
}
```

client/normal_nil.go
```diff
@@ -0,0 +1,48 @@
...
+	case FieldKind_NILLABLE_BOOL_ARRAY:
+		return NewNormalBoolNillableArray(immutable.None[[]bool]()), nil
+	case FieldKind_NILLABLE_INT_ARRAY:
+		return NewNormalIntNillableArray(immutable.None[[]int64]()), nil
+	case FieldKind_NILLABLE_FLOAT_ARRAY:
+		return NewNormalFloatNillableArray(immutable.None[[]float64]()), nil
+	case FieldKind_NILLABLE_STRING_ARRAY:
+		return NewNormalStringNillableArray(immutable.None[[]string]()), nil
+	default:
+		return nil, NewCanNotMakeNormalNilFromFieldKind(kind)
```
AndrewSisley, line 46: suggestion: Maybe returning `NormalVoid` would be nicer than an error here? I'm fairly torn though, so don't do this if you are even slightly preferring the error :)

encoding/field_value.go
```diff
@@ -81,9 +102,9 @@ func DecodeFieldValue(b []byte, descending bool) ([]byte, any, error) {
 			b, v, err = DecodeVarintAscending(b)
 		}
 		if err != nil {
-			return nil, nil, NewErrCanNotDecodeFieldValue(b, client.FieldKind_NILLABLE_INT, err)
```
AndrewSisley, line 84: question: Why did you remove the Kind from the log, it is still known here and might be useful no?

client/normal_value.go
```diff
@@ -0,0 +1,131 @@
...
+
+// NormalValue is the interface for the normal value types.
+// It is used to represent the normal (or standard) values across the system and to avoid
+// asserting all possible types like int, int32, int64, etc.
+//
+// All methods returning a specific type returns the value and the second boolean flag indicating
+// if the value is of the requested type.
+//
+// All nillable values are represented as immutable.Option[T].
+type NormalValue interface {
```
AndrewSisley, line 27: suggestion: A `Kind() FieldKind` func might be handy to have on this.

AndrewSisley (commented):
Looks good :)  Just a couple of small suggestions :)

client/normal_value.go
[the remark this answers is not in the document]
islamaliev: > document whether they return nillable values for none nillable `NormalValues`

I don't think we should document on what it returns in case it's nillable or not nillable. It's irrelevant. If the requested method does not match exactly the underlying type, the second param will be false (even if both underlying and requested are nillable). And in this case the first parameter is the default value as with standard Go type assertion. We don't document that the first value of `Int()` returns 0, because it should be clear and nobody is supposed to use this value anyway.

client/normal_value.go
```diff
@@ -0,0 +1,134 @@
...
+	NillableBytes() (immutable.Option[[]byte], bool)
+	// NillableTime returns the value as a nillable time.Time.
+	NillableTime() (immutable.Option[time.Time], bool)
+	// NillableDocument returns the value as a nillable *Document.
+	NillableDocument() (immutable.Option[*Document], bool)
+
+	// IsArray returns if the value is an array.
+	IsArray() bool
+
+	// BoolArray returns the value as a bool array.
```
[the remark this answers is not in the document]
islamaliev: as I mentioned earlier, I don't think it's worth documenting on the first value if the second returns false.
It acts as Go's type assertion that in this case returns default values (or zero). For arrays it's nil

client/normal_value.go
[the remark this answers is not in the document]
AndrewSisley: > I don't think we should document on what it returns in case it's nillable or not nillable. It's irrelevant. If the requested method does not match exactly the underlying type, the second param will be false (even if both underlying and requested are nillable).

This is not clear from the function signature.  It is not unreasonable to expect `NillableInt` to return Some when calling it on a none nillable int type.

client/normal_void.go
```diff
@@ -0,0 +1,200 @@
...
+
+package client
+
+import (
+	"time"
+
+	"github.com/sourcenetwork/immutable"
+)
+
+// NormalVoid is dummy implementation of NormalValue to be embedded in other types.
```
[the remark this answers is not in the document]
islamaliev: no, it's not used for anything else. And it doesn't implement `Unwrap` method (former `Any()`) to enforce override if `NormalVoid` is ever embedded by a custom type.

client/value.go
islamaliev: where exactly is do you think there can be a bug? switching from go's type assertion to `NormalValue` type assertion? They are identical.

AndrewSisley: There is very likely a bug in develop branch.

db/fetcher/indexer_iterators.go
fredcarle: > annoyingly generic funcs cant be defined on Go interfaces or we could replace most funcs on it with a couple of generic ones...

They can be defined on interfaces: https://go.dev/play/p/gt5p6hQsq_Q

AndrewSisley: > They can be defined on interfaces:

Only if the type param is declared on the interface, which you dont want to do here.

islamaliev: I don't a benefit here. Why can't you just write `v, ok := value.Unwrap().(int64)`?

AndrewSisley: > Why can't you just write v, ok := value.Unwrap().(int64)?

That is my suggestion, only wrapped up into a func so you don't have to cast in a whole bunch of places.

encoding/field_value.go
islamaliev: I removed it because at some point `cliend.FieldKind` was removed from the func signature.
I bought it back.

client/value.go
AndrewSisley: I just looked it over - I misread, it (`convertImmutable`) is weirdly written (probably by me...) and does append the nil values instead of setting by index.  Resolving this conversation.

core/key_test.go
islamaliev: ok

core/encoding.go
islamaliev: no, nothing breaks. The two slices should have the same len.
I just thought some abnormal scenario (that I can't think of) these might be different.
But I think you are right, it is not necessary. I bought it back.

core/encoding.go
islamaliev: removed

client/normal_nil.go
islamaliev: I'm not sure about it at the moment. I think we don't have enough use cases to see if it makes sense. I'd say lets stay on the safe side and keep the error for now.

client/normal_value.go
islamaliev: good suggestion but we don't have all `FieldKind` to cover all possible normal values. For example, we don't have `FieldKind_INT`.

client/normal_new.go
```diff
@@ -0,0 +1,465 @@
...
+
+import (
+	"time"
+
+	"github.com/sourcenetwork/immutable"
+)
+
+// NewNormalValue creates a new NormalValue from the given value.
+// It will normalize all known types that can be converted to normal ones.
+// For example, is the given type is `[]int32`, it will be converted to `[]int64`.
```
fredcarle, line 21: todo: "`if` the given type is..."

client/normal_new.go
```diff
@@ -0,0 +1,465 @@
...
+// If the given value is of type `[]any` is will to through every element and try to convert it
```
fredcarle, line 22: todo: "`it` with `go` through..."

client/normal_new.go
```diff
@@ -0,0 +1,465 @@
...
+// minimal common type and normalizes it.
```
fredcarle, line 23: question: Is `minmal` the right word here? There might be something missing to make it make sense.

client/normal_new.go
```diff
@@ -0,0 +1,465 @@
...
+// For examples, the following conversions will be made:
+//   - `[]any{int32(1), int64(2)}` -> `[]int64{1, 2}`.
+//   - `[]any{int32(1), int64(2), float32(1.5)}` -> `[]float64{1.0, 2.0, 1.5}`.
+//   - `[]any{int32(1), nil}` -> `[]immutable.Option[int64]{immutable.Some(1), immutable.None[int64]()}`.
+//
+// This function will not check if the given value is `nil`. To normalize a `nil` value use the
+// `NewNormalNil` function.
+func NewNormalValue(val any) (NormalValue, error) {
```
fredcarle, line 31: thought: If we passed `FieldKind` in here it would make it even more strongly typed given that `[]any{int32(1), int64(2)}` could very well be `[]immutable.Option[int64]{immutable.Some(1), immutable.Some(2)}` instead of `[]int64{1, 2}`. This can be done in a separate PR though.

client/normal_util.go
```diff
@@ -0,0 +1,118 @@
...
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package client
+
+// ToArrayOfNormalValues converts a NormalValue into a slice of NormalValue is the given value
```
fredcarle, line 13: todo: "`if` the given value..."

client/normal_value.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/sourcenetwork/immutable"
+)
+
+// NormalValue is the interface for the normal value types.
+// It is used to represent the normal (or standard) values across the system and to avoid
+// asserting all possible types like int, int32, int64, etc.
+//
+// All methods returning a specific type returns the value and the second boolean flag indicating
+// if the value is of the requested type. They act similar to the Go's type assertion.
```
fredcarle, line 24: todo: "to Go's type assertion" or "to the Go type assertion"

db/fetcher/indexer_iterators.go
fredcarle: `v, ok := value.Unwrap().(int64)` this wouldn't work as it wouldn't check for `nil` passing `0` to the `evalFunc` is no a correct behaviour if the actual value is `nil`.

fredcarle (approved):
Approving and assuming you'll resolve all comments before merge :)

This is overall really nice and brings us to a place where we can easily add on if we need to as we go.

db/fetcher/indexer_iterators.go
AndrewSisley: Ah true, the private func would need to be:
```
func AsT[T client.NormalValueConstraints](v NormalValue) (T, bool) {
  if v.IsNil() {
    return false
  }
  return v.Unwrap().(T), true
}
```

client/normal_value.go
AndrewSisley: I think that can be sorted out when we add support for `FieldKind_INT`

client/normal_value.go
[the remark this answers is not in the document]
islamaliev: done

client/normal_value.go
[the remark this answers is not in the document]
islamaliev: done

client/normal_value.go
islamaliev: this is somewhat involved addition. I've created another task for it https://github.com/sourcenetwork/defradb/issues/2463

client/normal_new.go
islamaliev: this is somewhat involved addition. I've created another task for it https://github.com/sourcenetwork/defradb/issues/2463

db/fetcher/indexer_iterators.go
islamaliev: I just changed code to check it explicitly. You suggestion wouldn't work as it will return 0 for `nil` values which will match if the user runs a filter `_eq: 0`

PR #2406, opened 2024-03-13
refactor: Replace logging package with corelog

### LINKED ISSUE #2405: Replace logging package with corelog
Standardize logs by using the corelog library.

nasdf:
## Relevant issue(s)

Resolves #2405

## Description

This PR replaces the logging package with the corelog library.

## Tasks

## How has this been tested?

Manually tested

- MacOS


nasdf:
> And has the `corelog` lib been reviewed by anyone? Should we also review it in this PR?

I think it'd be nice to have them both reviewed before merging.

AndrewSisley:
Thanks Keenan :)

> context for logging calls has been moved to a context specific variant. For example Info has a InfoContext variant.

What is the reasoning behind this change?  Is there a Source use-case for calling `Info` without a `ctx`?

And did you consider calling it `InfoCtx` instead?  Do you know of other logging packages using this naming?

nasdf:
> Thanks Keenan :)
> 
> > context for logging calls has been moved to a context specific variant. For example Info has a InfoContext variant.
> 
> What is the reasoning behind this change? Is there a Source use-case for calling `Info` without a `ctx`?

There's a few cases within DefraDB that did not have a proper context and I wanted it to be flexible for the other projects.

> And did you consider calling it `InfoCtx` instead? Do you know of other logging packages using this naming?

This is the same API as the [`slog`](https://pkg.go.dev/log/slog#DebugContext) package.

http/logger.go
```diff
@@ -28,14 +27,14 @@ type logEntry struct {
...
-	log.Info(
+	log.InfoContext(
 		e.req.Context(),
 		"Request",
-		logging.NewKV("Method", e.req.Method),
-		logging.NewKV("Path", e.req.URL.Path),
-		logging.NewKV("Status", status),
-		logging.NewKV("LengthBytes", bytes),
-		logging.NewKV("ElapsedTime", elapsed.String()),
+		"Method", e.req.Method,
```
AndrewSisley, line 33: todo: These no longer look 'structured' looking at the call `"Method": methodValue, "Path": pathValue, etc`, instead it looks like they would be logged as equal and independent values `"Method", methodValue, "Path", pathValue, ...`?

I think that is a loss, and one worth chatting about with the team.

AndrewSisley (commented):
The code changes in this repo look good, but I want the coreLog lib to be reviewed before this is merged, and there is a design change that I would like the team to chat about and agree upon too (RE log args).

http/logger.go
nasdf: It is possible to use `slog.String("key", "val")` or any of the other variants here in place of the alternating key value pairs. I have a slight preference for the alternating key values but happy to hear more opinions.

datastore/blockstore.go
```diff
@@ -64,7 +64,7 @@ func (bs *bstore) HashOnRead(enabled bool) {
 // Get returns a block from the blockstore.
 func (bs *bstore) Get(ctx context.Context, k cid.Cid) (blocks.Block, error) {
 	if !k.Defined() {
-		log.Error(ctx, "Undefined CID in blockstore")
+		log.ErrorContext(ctx, "Undefined CID in blockstore")
```
fredcarle, line 67: suggestion: We should remove the error logging here since an error is returned to the caller.

http/logger.go
fredcarle: I'm also in favour of using `slog.String("key", "val")`. Is there a reason for your preference of the alternating key value pairs that could help change our minds?

net/server.go
```diff
@@ -496,55 +495,55 @@ func (s *server) publishLog(ctx context.Context, topic string, req *pb.PushLogRe
...
+		log.ErrorContextE(s.peer.ctx, "Failed to unmarshal pubsub message %s", err)
 		return nil, err
 	}
 
 	ctx := grpcpeer.NewContext(s.peer.ctx, &grpcpeer.Peer{
 		Addr: addr{from},
 	})
 	if _, err := s.PushLog(ctx, req); err != nil {
-		log.ErrorE(ctx, "Failed pushing log for doc", err, logging.NewKV("Topic", topic))
+		log.ErrorContextE(ctx, "Failed pushing log for doc", err, "Topic", topic)
```
fredcarle, line 525: todo: This error log should just be removed.

fredcarle (commented):
> Feedback methods have been removed in favor of restricting the output to stderr and stdout

The Feedback methods allowed sending the logs to a log file but also return useful messages to the terminal. Any plans to support the same functionality? Or is there a more devex appropriate way to do this?

nasdf:
> > Feedback methods have been removed in favor of restricting the output to stderr and stdout
> 
> The Feedback methods allowed sending the logs to a log file but also return useful messages to the terminal. Any plans to support the same functionality? Or is there a more devex appropriate way to do this?

My main concern with logging to a file is that we don't have a log rotation system. Cloud deployments usually come with a way to automatically capture logs, docker has a built in logging driver, and linux deployments can use syslog or systemd / journalctl.

http/logger.go
nasdf: > Is there a reason for your preference of the alternating key value pairs that could help change our minds?

It's easier for me to read, but I don't mind the other option.

AndrewSisley: > It's easier for me to read, but I don't mind the other option.

Might be worth chatting with the protocol (and the rest of the db team) about this - whilst the current interface allows for both to be used, we might want to force `slog.Foo("Key", val)` like the old logging package used to.

fredcarle:
I just tested out what it looks like in the terminal.
```
time=2024-03-15T13:55:36.073-04:00 level=INFO msg="Providing HTTP API at http://127.0.0.1:9181." name=node
time=2024-03-15T13:57:05.898-04:00 level=INFO msg=Request name=http Method=POST Path=/api/v0/graphql Status=200 LengthBytes=26 ElapsedTime=4.096792ms
```

One thing that I would prefer is the have the `name` field before or after `level` but definitely before `msg` since I feel like it's confusing the focus of trying to understand what the logs is trying to tell me. I know it doesn't matter for a log capturing tool that can re organise everything but when looked at raw by devs, in my case anyways, once I hit the `msg`, I want only `msg` related information past that point. I hope that makes sense. Does anyone else feel this way?

AndrewSisley:
> One thing that I would prefer is the have the name field before or after level but definitely before msg since I feel like it's confusing the focus of trying to understand what the logs is trying to tell me. I know it doesn't matter for a log capturing tool that can re organise everything but when looked at raw by devs, in my case anyways, once I hit the msg, I want only msg related information past that point. I hope that makes sense. Does anyone else feel this way?

Very nice spot :)  I think it should be before `level` too.  It is particularly bad in its current location as it now shares the same format as the structured params (this also feels problematic from a collision perspective).

nasdf:
> I just tested out what it looks like in the terminal.
> 
> ```
> time=2024-03-15T13:55:36.073-04:00 level=INFO msg="Providing HTTP API at http://127.0.0.1:9181." name=node
> time=2024-03-15T13:57:05.898-04:00 level=INFO msg=Request name=http Method=POST Path=/api/v0/graphql Status=200 LengthBytes=26 ElapsedTime=4.096792ms
> ```
> 
> One thing that I would prefer is the have the `name` field before or after `level` but definitely before `msg` since I feel like it's confusing the focus of trying to understand what the logs is trying to tell me. I know it doesn't matter for a log capturing tool that can re organise everything but when looked at raw by devs, in my case anyways, once I hit the `msg`, I want only `msg` related information past that point. I hope that makes sense. Does anyone else feel this way?

It's possible to change the order but we'll need to rename the variables. This should also help avoid attribute name collisions. Which version does everyone prefer?

```
_time=2024-03-15T12:12:06.883-07:00 _name=test _level=INFO _msg=message
```

```
$time=2024-03-15T12:14:35.429-07:00 $name=test $level=INFO $msg=message
```

AndrewSisley:
> It's possible to change the order but we'll need to rename the variables. 

Weirdly I think I prefer the `$` prefix, but no strong preference.

fredcarle:
> > It's possible to change the order but we'll need to rename the variables.
> 
> Weirdly I think I prefer the `$` prefix, but no strong preference.

Same 👍

AndrewSisley (approved):
LGTM, thanks a bunch Keenan :)

fredcarle (approved):
LGTM!

PR #2414, opened 2024-03-15
refactor: Unify Field Kind and Schema properties

### LINKED ISSUE #2409: Unify SchemaFieldDescription Kind and Schema
They are really mutually exclusive, and can be combined into a single property.

Design discussed in discord dev-db thread `Collection vs Schema`:
```
type SchemaFieldDescription struct {
    Kind FieldKind
    ...
}

type FieldKind interface {
    String() string
    IsNillable() bool
    IsObject() bool
    IsObjectArray() bool
    IsArray() bool
}

type ScalarKind uint8
type ScalarArrayKind uint8
type ObjectKind string
type ObjectArrayKind string

func (k ScalarKind) IsObject() bool {
    return false
}
...

func (f SchemaFieldDescription) UnmarshalJson() {
   ...
   // As well as deciding which concrete type to parse into, this function could also handle parsing `"[Boolean]"` into `18` etc
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #2409

## Description

Unifies the Field Kind and Schema properties.

Commits should be clean, although most of the work is in the last.  Tests have been modified/deleted in other ways besides cid changes as this PR changes public behaviour (heads up in case the first few files make it look like the usual cid change stuff).

### DIFF
--- a/client/schema_field_description.go
+++ b/client/schema_field_description.go
@@ -10,10 +10,36 @@
 
 package client
 
-import "fmt"
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+)
 
 // FieldKind describes the type of a field.
-type FieldKind uint8
+type FieldKind interface {
+	// String returns the string representation of this FieldKind.
+	String() string
+
+	// Underlying returns the unterlying Kind as a string.
+	//
+	// If this is an array, it will return the element kind, else it will return the same as
+	// [String()].
+	Underlying() string
+
+	// IsNillable returns true if this kind supports nil values.
+	IsNillable() bool
+
+	// IsObject returns true if this FieldKind is an object type, or an array of object types.
+	IsObject() bool
+
+	// IsObjectArray returns true if this FieldKind is an object array type.
+	IsObjectArray() bool
+
+	// IsArray returns true if this FieldKind is an array type which includes inline arrays as well
+	// as relation arrays.
+	IsArray() bool
+}
 
 // SchemaFieldDescription describes a field on a Schema and its associated metadata.
 type SchemaFieldDescription struct {
@@ -26,10 +52,6 @@ type SchemaFieldDescription struct {
 	//
 	// Must contain a valid value. It is currently immutable.
 	Kind FieldKind
-
-	// Schema contains the schema name of the type this field contains if this field is
-	// a relation field.  Otherwise this will be empty.
-	Schema string
 
 	// RelationName the name of the relationship that this field represents if this field is
 	// a relation field.  Otherwise this will be empty.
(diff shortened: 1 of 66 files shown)


client/schema_field_description.go
```diff
@@ -10,10 +10,36 @@
...
 // FieldKind describes the type of a field.
-type FieldKind uint8
+type FieldKind interface {
+	// String returns the string representation of this FieldKind.
+	String() string
+
+	// Underlying returns the unterlying Kind as a string.
+	//
+	// If this is an array, it will return the element kind, else it will return the same as
+	// [String()].
```
fredcarle, line 27: thought: I think that this notation is used in some other places as well but here it threw me off given the context. I was wondering why calling `Undelying()` on a non array would return `String()` wrapped in an array.

client/schema_field_description.go
```diff
@@ -169,3 +280,74 @@ var FieldKindStringToEnumMapping = map[string]FieldKind{
...
+		if err != nil {
+			return err
+		}
+		switch intKind {
+		case uint8(FieldKind_BOOL_ARRAY), uint8(FieldKind_INT_ARRAY), uint8(FieldKind_FLOAT_ARRAY),
+			uint8(FieldKind_STRING_ARRAY), uint8(FieldKind_NILLABLE_BOOL_ARRAY), uint8(FieldKind_NILLABLE_INT_ARRAY),
+			uint8(FieldKind_NILLABLE_FLOAT_ARRAY), uint8(FieldKind_NILLABLE_STRING_ARRAY):
+			f.Kind = ScalarArrayKind(intKind)
+		default:
+			f.Kind = ScalarKind(intKind)
```
fredcarle, line 327: thought: This as the potential to be quite annoying to maintain as we update out list of supported array types. I'm wondering if we should take the hit now and change the assigned number ranges to something like:
- [0:100] -> Scalar types
- [100:200] -> Array types

We wont be using the full range but at least if we ever need to do another switch statement it would be a much cleaner check.

fredcarle (approved):
LGTM. I like that it simplifies the code in some areas. Good job!

client/schema_field_description.go
AndrewSisley: Square brackets are how to correctly link stuff in Go docs, although I'm not really sure it works for funcs.  I used to use backticks as that feels more normal (and is what rust uses), but the Go devs went with square brackets in 1.19:

https://tip.golang.org/doc/comment#links

client/schema_field_description.go
AndrewSisley: I dislike the current too, however I don't like the range option, and would probably prefer a custom encoding: First bit for nillable, second bit for array, then the scalar id.

As a proper solution would require a bit of thought and an agreed design, I'd prefer if this conversation was out of scope here.

PR #2421, opened 2024-03-18
feat: Add Defra-Lens support for branching schema

### LINKED ISSUE #2293: Add Defra-Lens support for branching schema
https://github.com/sourcenetwork/defradb/pull/2286 allowed schema branching to work locally.  It shouldn't be too much extra effort to get migrations to work for it too - just a bit of extra pathing logic in len/history.go

AndrewSisley:
## Relevant issue(s)

Resolves #2293

## Description

Adds Defra-Lens support for branching schema.

Defra will now be able to handle migrations for a branching schema version DAG.

### DIFF
--- a/lens/history.go
+++ b/lens/history.go
@@ -36,10 +36,8 @@ type schemaHistoryLink struct {
 }
 
 // targetedSchemaHistoryLink represents an item in a particular schema's history, it
-// links to the previous and next version items if they exist.
-//
-// It also contains a vector which describes the distance and direction to the
-// target schema version (given as an input param on construction).
+// links to the previous and next version items if they exist and are on the path to
+// the target schema version.
 type targetedSchemaHistoryLink struct {
 	// The collection as this point in history.
 	collection *client.CollectionDescription
@@ -51,13 +49,6 @@ type targetedSchemaHistoryLink struct {
 	// The link to the previous schema version, if there is
 	// one (for the initial schema version this will be None).
 	previous immutable.Option[*targetedSchemaHistoryLink]
-
-	// The distance and direction from this history item to the target.
-	//
-	// A zero value indicates that this is the target item. A positive value
-	// indicates that the target is more recent. A negative value indicates
-	// that the target predates this history item.
-	targetVector int
 }
 
 // getTargetedSchemaHistory returns the history of the schema of the given id, relative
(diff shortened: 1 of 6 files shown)


lens/history.go
```diff
@@ -180,25 +150,28 @@ func getSchemaHistory(
...
 	if err != nil {
 		return nil, err
 	}
 
-	history := map[collectionID]*schemaHistoryLink{}
+	history := map[schemaVersionID]*schemaHistoryLink{}
+	schemaVersionsByColID := map[uint32]schemaVersionID{}
 
 	for _, c := range cols {
 		col := c
```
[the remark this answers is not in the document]
AndrewSisley: I have no idea, might have made sense at one point, or it might have been a brain-fart.

I'll remove it shortly before merge.

- [ ] Remove odd var copy

lens/history.go
```diff
@@ -76,99 +67,78 @@ func getTargetedSchemaHistory(
...
 
-	result := map[schemaVersionID]*targetedSchemaHistoryLink{}
-
-	for _, item := range history {
-		result[item.collection.SchemaVersionID] = &targetedSchemaHistoryLink{
-			collection: item.collection,
-		}
+	targetHistoryItem, ok := history[targetSchemaVersionID]
+	if !ok {
+		// If the target schema version is unknown then there are possible no migrations
```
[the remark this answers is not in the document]
AndrewSisley: cheers, is actually in the wrong place, should be on the left side of `no`

- [ ] Move word

lens/history.go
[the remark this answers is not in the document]
AndrewSisley: I found out, it is because we want a pointer to it, and the way Go works means that without the copy `col` would always be the last col in `cols`.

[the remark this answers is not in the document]
fredcarle: Thanks for reminding me of this. Could you add a `TODO remove when Go 1.22` comment above it please.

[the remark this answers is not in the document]
AndrewSisley: Ah nice, glad that will fix it - will do

- [ ] Add too comment

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/lens-vm/lens/host-go/config/model"
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/lenses"
+)
+
+func TestSchemaMigrationQueryWithP2PReplicatedDocOnOtherSchemaBranch(t *testing.T) {
```
fredcarle, line 24: thought: This test should propably be in the net replicator test no?

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
```diff
@@ -0,0 +1,138 @@
...
+						Users {
+							name
+							phone
+						}
+					}
+				`,
+				Results: []map[string]any{
+					{
+						// name has been cleared by the inverse of the migration from version 1 to 2
+						"name": nil,
```
fredcarle, line 127: thought: This is an unexpected behaviour. In my mind this should have had the value set by the first migration. I'm might be missing why we are doing the inverse migration here.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: It fits in both directories just as well IMO - I probably prefer it here because I see it more from a schema migration perspective, and you probably see it in the `net` directory as you see it more from a P2P perspective.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: I'm not sure I follow - we defined the inverse/migration action because we want to test it.

The reason it executes is because the document was defined on schema version 2, and to get to schema version 3 we need to inverse to 1, then migrate to 2.  The schema version DAG looks like the below:

```
       (1)
     |     |
   (2)     (3) 
```

fredcarle: Ah. I see where I missed. In `TestSchemaMigrationQuery_WithBranchingSchema` we set the active version back to the first one so it was clearer a little bit. Here, each patch is done on a separate node. That's what I missed. All good 👍

fredcarle: Note: I'm looking forward to the day this test can revert back to `John` instead of `nil`.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
fredcarle: In a sense, everything that is in the `net` directory also relate to some other test directory. In my mind, as soon as a test is interacting with the P2P system, it would be better placed in the `net` package.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: > Note: I'm looking forward to the day this test can revert back to John instead of nil.

I don't understand - `nil` is the desired outcome.  We will never go directly from schema version 2 to 3 with a schema history like this - we always need to respect the dag history.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
```diff
@@ -0,0 +1,138 @@
...
+					Users {
+						name
+					}
+				}`,
+				Results: []map[string]any{
+					{
+						"name":     "John",
+						"verified": true,
+					},
+				},
```
fredcarle, line 111: thougth: This seems to document a bug in our test system. Why is `verified` returned when only `name` is requested. Also, we should be able to return the `email` field here.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
fredcarle: I would think the desired outcome is `John` since that is what it was set to on a field that exists on all versions. And didn't need to go through the migration at any point.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: > This seems to document a bug in our test system.

True, I forgot to add that to the query
- [ ] Fix test query

> Why is verified returned when only name is requested.

It wont be returned, I think it is far more likely that our test system is not asserting that all expected fields are returned.

> Also, we should be able to return the email field here.

Of course, but it is not relevant to this test and would just be noise here .

fredcarle: > Of course, but it is not relevant to this test and would just be noise here .

That is subjective. I find that since a patch was applied to the schema, I would like to see that the field is available to query. Otherwise I'm not sure why just the name was queried. One could argue that the comment above is misleading as it's not true that the newer version is unknown to this node. It is in fact on a newer schema version. Technically just as new to the node as the other version is to the other node.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
fredcarle: You raise an interesting point. This implies that migration can be destructive when sharing data across nodes that aren't on the same schema version for a collection if they know of the migration. Setting a migration without the schema involved existing on the node should come with a big warning. I know the DAG isn't affected but the view is and that may cause user to panic when they see (because the view is what they see) their data disappear.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: > I would like to see that the field is available to query.

There are other tests that test this, without complicating things with Lens transforms.

> One could argue that the comment above is misleading as it's not true that the newer version is unknown to this node

I don't understand you here. schema version 3 is unknown, it doesnt exist on this node, only the migration exists not the definition.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
AndrewSisley: > This implies that migration can be destructive when sharing data across nodes that aren't on the same schema version for a collection if they know of the migration

Lens transforms in general can be destructive, there is nothing special/new here.  The lens transform choses to override existing data, that is a choice available to all Lenses, since Lens has existed.  Not warning to this specific flavour here is required, as it is just an inherent fact about Lens.

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
```diff
@@ -0,0 +1,138 @@
...
+				// Create John on the first (source) node only, and allow the value to sync
+				NodeID: immutable.Some(0),
+				Doc: `{
+					"name": "John",
+					"verified": true
+				}`,
+			},
+			testUtils.WaitForSync{},
+			testUtils.Request{
+				// Node 0 should yield results as they were defined, as the newer schema version is
```
AndrewSisley, line 98: todo:  change `as the newer schema version` to `as schema version 3...`

tests/integration/schema/migrations/query/with_p2p_schema_branch_test.go
fredcarle: Note: We had a live conversation about this and the conclusion is that Lens migrations can be destructive but that is simply a result of their implementation flexibility and that is reflected in the Lens migration used in this test. 

All good 👍

fredcarle (approved):
LGTM! Thanks for answering all my comments and questions.

AndrewSisley:
> LGTM! Thanks for answering all my comments and questions.

Thanks for finding a bunch of stuff (including the ones raised on discord) :)

PR #2423, opened 2024-03-18
feat: Encryption at rest

### LINKED ISSUE #1987: Encryption at rest
Badger has support for encryption at rest, but there is no documentation for how to enable it for DefraDB.

nasdf:
## Relevant issue(s)

Resolves #1987 

## Description

This PR enables encryption at rest. Encryption can only be enabled at initialization. The `init` command has a shortcut that simplifies the process of setting up DefraDB with encryption `defradb init -e`. This PR does not include a mechanism to rotate encryption keys.

## Tasks

## How has this been tested?

`make test`

- MacOS


cli/init.go
```diff
@@ -0,0 +1,64 @@
...
+
+			if encrypted {
+				// generate a random encryption key
+				key := make([]byte, 32)
+				if _, err := rand.Read(key); err != nil {
+					return err
+				}
+				cfg.Set("datastore.badger.encryptionkey", hex.EncodeToString(key))
+				// set index cache size to improve read performance.
+				cfg.Set("datastore.badger.indexcachesize", 10<<20)
```
AndrewSisley, line 56: todo: Is this not overwriting/ignoring anything the user may have explicitly set for `datastore.badger.indexcachesize` in their config?  Can we only do this if they have not explicitly provided a value?  Can this be documented in a user-visible space if we keep it like this?

node/store.go
```diff
@@ -56,12 +60,19 @@ func WithValueLogFileSize(size int64) StoreOpt {
...
 
 // WithEncryptionKey sets the badger encryption key.
-func WithEncryptionKey(encryptionKey []byte) StoreOpt {
+func WithEncryptionKey(encryptionKey string) StoreOpt {
 	return func(o *StoreOptions) {
 		o.encryptionKey = encryptionKey
 	}
 }
 
+// WithIndexCacheSize sets the index cache size.
```
AndrewSisley, line 69: suggestion:  This may be confused with thinking we have a cache for secondary indexes somewhere, suggest changing to  `WithIndexCacheSize sets the rootstore index cache size`

cli/init.go
```diff
@@ -0,0 +1,64 @@
...
+			cfg.AddConfigPath(rootdir)
+
+			err := bindConfigFlags(cfg, cmd.Root().PersistentFlags())
+			if err != nil {
+				return err
+			}
+
+			if encrypted {
+				// generate a random encryption key
+				key := make([]byte, 32)
```
AndrewSisley, line 50: suggestion:  I think it is worth documenting the `32`, as if it is too small it could be insecure no?

thought: Is it worth us taking ownership over key generation like we are here?  Might it not be better to just let users provide their own via a CLI arg?

cli/init.go
nasdf: This only happens when the config is first created. They can change the value after that.

cli/init.go
nasdf: > thought: Is it worth us taking ownership over key generation like we are here? Might it not be better to just let users provide their own via a CLI arg?

This is just for convenience. It is possible to run `defradb init` and then provide your own encryption key in the config.

cli/init.go
AndrewSisley: What happens if `init` is called on an existing database?

cli/init.go
AndrewSisley: Oh, I missed something - we are just storing the key in plain text in the config... Are we sure we want to do that even in the short term?

cli/init.go
nasdf: It will error if it already exists. [`SafeWrite`](https://pkg.go.dev/github.com/spf13/viper#SafeWriteConfig) ensures we don't overwrite an existing config.

cli/init.go
nasdf: I think in the future we could use Orbis for key management. I don't see an issue with storing it in the config. My thought process being: if you can access the config you can access the data files as well.

fredcarle: The issue I see with keeping it in the config is that it might make it more difficult for projects that like to store config in their repos. This would could lead to keys being shared with people that shouldn't have access to them.

AndrewSisley: >  if you can access the config you can access the data files as well

The point of this PR is that the data is encrypted on disk, so if someone gains access to the machine they cant access the data no?  And we are storing the key in plain text next to the 'protected' data.

nasdf: There is also the automatic environment variable for that case.

AndrewSisley: > There is also the automatic environment variable for that case.

That has similar problems to storing it in the config, if someone has access to the file system, they presumably have access to the env vars too?

fredcarle: > That has similar problems to storing it in the config, if someone has access to the file system, they presumably have access to the env vars too?

I think the key should be in a separate file to save the potential config headache (key in repo for as stated above). That will have the same issue as storing it in the `env` variables though. One that we should not think we are responsible for. The level of safety needed depends on the deployment environment and the developers should be the one responsible for their usage of secret keys. All that we need to worry about is providing options for them to pass their keys along to Defra.

jsimnz: Discussed in the discord thread, but small note for @AndrewSisley

> suggestion: I think it is worth documenting the 32, as if it is too small it could be insecure no?

32bytes is the correct key size for AES256. It ideally runs with 128, 192, or 256 *bits*, which 32bytes*8 = 256 bits

AndrewSisley: > 32bytes is the correct key size for AES256. It ideally runs with 128, 192, or 256 bits, which 32bytes*8 = 256 bits

Yes, I trusted the current number was appropriate, I'm more worried about documenting it so that nobody tries to set it to something smaller.

nasdf:
Closing until key management is implemented.

PR #2424, opened 2024-03-18
feat: Allow mutation of collection sources via PatchCollection

### LINKED ISSUE #2422: Allow mutation of collection sources via PatchCollection

AndrewSisley:
## Relevant issue(s)

Resolves #2422

## Description

Allows the mutation of collection sources via `PatchCollection`.

Users can now set view and migration Lens transforms, and update view queries via json patch.

Validation is non-existant RE the Query struct, as it is in develop branch.  An invalid config will result in errors when querying the view.  I think we can improve on this later, this PR does however increase the likelihood of invalid view queries being constructed (by users).

It also massively increases the exposure of the `client/request` package, which whilst already public, is quite poorly documented.  Similar to validation, this PR increases the priority of documenting it.


db/collection.go
```diff
@@ -661,10 +702,28 @@ func validateSourcesNotModified(
...
+					oldColSources[i].SourceCollectionID,
+				)
+			}
+		}
+
+		newQuerySources := newCol.QuerySources()
+		oldQuerySources := oldCol.QuerySources()
+
+		if len(newQuerySources) != len(oldQuerySources) {
+			return NewErrCollectionSourcesCannotBeAddedRemoved(newCol.ID)
```
fredcarle, line 726: todo: This validation is no longer doing what it was doing before. This means that the name is no longer accurate. Sources can be modified. We can't add or remove them or change their ID, be they can be modified. Maybe change the name to `validateSourcesModification`.

tests/integration/collection_description/updates/replace/query_source_transform_test.go
```diff
@@ -0,0 +1,113 @@
...
+						]
+					`,
+					newTransformCfgJson,
+				),
+			},
+			testUtils.CreateDoc{
+				// Set the `name` field only
+				Doc: `{
+					"firstName":	"John",
+					"lastName":	"S"
```
fredcarle, line 92: nitpick: If you don't mind removing the extra spaces that would be great.

db/collection.go
AndrewSisley: Good spot, will do.

- [ ] Rename validation func

fredcarle (approved):
Approving assuming the minor todo will be resolved. The tests are really nice in helping to visualize what's happening.

PR #2443, opened 2024-03-21
refactor: Clean up client/request package

### LINKED ISSUE #2425: Clean up client/request package

AndrewSisley:
## Relevant issue(s)

Resolves #2425

## Description

Cleans up client/request package, removing some extra stuff, and documenting most of the remaining.

It is not perfect, but it should hopefully be better.

### DIFF
--- a/client/request/select.go
+++ b/client/request/select.go
@@ -35,10 +35,10 @@ type Select struct {
 	Offsetable
 	Orderable
 	Filterable
+	DocIDsFilter
 	Groupable
 
-	DocIDs immutable.Option[[]string]
-	CID    immutable.Option[string]
+	CID immutable.Option[string]
 
 	// Root is the top level type of parsed request
 	Root SelectionType
@@ -130,8 +130,8 @@ type selectJson struct {
 	Offsetable
 	Orderable
 	Filterable
+	DocIDsFilter
 	Groupable
-	DocIDs      immutable.Option[[]string]
 	CID         immutable.Option[string]
 	Root        SelectionType
 	ShowDeleted bool
--- a/client/request/select.go
+++ b/client/request/select.go
@@ -31,6 +31,7 @@ type Select struct {
 	Field
 
 	Limitable
+	Offsetable
 
 	DocIDs immutable.Option[[]string]
 	CID    immutable.Option[string]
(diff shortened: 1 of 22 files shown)


client/request/select.go
```diff
@@ -12,39 +12,36 @@ package request
...
-	GroupBy immutable.Option[GroupBy]
-	Filter  immutable.Option[Filter]
-
+// ChildSelect represents a type with selectable child properties.
+//
+// At least one child must be selected.
+type ChildSelect struct {
+	// Fields contains the set of child properties to return.
+	//
+	// At least one child propertt must be selected.
```
fredcarle, line 43: todo: `property`

AndrewSisley: Lol

- [ ] Fix typo

client/request/select.go
```diff
@@ -111,25 +108,20 @@ func (s *Select) validateGroupBy() []error {
...
-
-	// Properties above this line match the `Select` object and
-	// are deserialized using the normal/default logic.
-	// Properties below this line require custom logic in `UnmarshalJSON`
-	// in order to be deserialized correctly.
-
-	Fields []map[string]json.RawMessage
 }
 
 func (s *Select) UnmarshalJSON(bytes []byte) error {
```
fredcarle, line 127: thought: Since `ChildSelect` has a custom `Unmarshaller`, this should now be unecessary.

fredcarle (approved):
Thank you for adding all that documentation and even creating the `Fooable`s. It adds to the clarity of what the request types contain.

client/request/select.go
AndrewSisley: Was documented in the commit message:

> The use of a ChildSelect object allows the custom code for desrialising fields to live on a dedicated object. Sadly, due to a very annoying 'quirk' of the json lib, embedded UnmarshalJSON functions are not picked up, and so we still have to do some manual property mapping of the parent.

fredcarle: It does pick it up. I added print statements in the ChildSelect unmarshal method and it does get in there. Something else is behaving weirdly though.

fredcarle: So it's not that the emvbedded UnmarshalJSON is not picked up. It's that it ignores the other fields: https://stackoverflow.com/a/38982567

fredcarle: Either way, you have to keep it like it is now. All good 👍.

PR #2445, opened 2024-03-21
refactor: Rewrite convertImmutable

### LINKED ISSUE #2444: Rewrite convertImmutable
It is too horrible, and quick to do.

AndrewSisley:
## Relevant issue(s)

Resolves #2444

## Description

Rewrites convertImmutable.

### DIFF
--- a/client/value.go
+++ b/client/value.go
@@ -79,13 +79,11 @@ func (val FieldValue) Bytes() ([]byte, error) {
 }
 
 func convertImmutable[T any](vals []immutable.Option[T]) []any {
-	var out []any
-	for _, val := range vals {
-		if !val.HasValue() {
-			out = append(out, nil)
-			continue
+	out := make([]any, len(vals))
+	for i := range vals {
+		if vals[i].HasValue() {
+			out[i] = vals[i].Value()
 		}
-		out = append(out, val.Value())
 	}
 	return out
 }


shahzadlone (approved):
LOL at issue description `It is too horrible` 😆 

LGTM

client/value.go
```diff
@@ -79,13 +79,11 @@ func (val FieldValue) Bytes() ([]byte, error) {
...
 func convertImmutable[T any](vals []immutable.Option[T]) []any {
-	var out []any
-	for _, val := range vals {
-		if !val.HasValue() {
-			out = append(out, nil)
-			continue
+	out := make([]any, len(vals))
+	for i := range vals {
+		if vals[i].HasValue() {
+			out[i] = vals[i].Value()
```
islamaliev, line 85: suggestion: this is a small change with potentially significant impact. I wonder why no tests cover this.
Would be great to add some tests.

islamaliev (approved):

client/value.go
AndrewSisley: It is tested, I was mistaken in your PR - I just rewrote it because it was really ugly

fredcarle: I don't think this change has any impact other than improving the efficiency of the code. The returned result is the same.

PR #2482, opened 2024-04-02
feat: Add P Counter CRDT

### LINKED ISSUE #2116: P Counter CRDT

fredcarle:
## Relevant issue(s)

Resolves #2116 

## Description

This PR adds a Positive Counter CRDT type. It converts the PNCounter implementation to a Counter implementation that supports both PN and P counters.

## Tasks

## How has this been tested?

make test

- MacOS


merkle/crdt/merklecrdt.go
```diff
@@ -79,21 +79,23 @@ func InstanceWithStore(
...
+				ctype == client.PN_COUNTER,
 			), nil
 		case client.FieldKind_NILLABLE_FLOAT:
-			return NewMerklePNCounter[float64](
+			return NewMerkleCounter[float64](
 				store,
 				schemaVersionKey,
 				key,
 				fieldName,
+				ctype == client.PN_COUNTER,
```
shahzadlone, line 98: nitpick: change to `cType`

tests/integration/mutation/update/crdt/pcounter_test.go
```diff
@@ -0,0 +1,264 @@
...
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents what happens when an overflow occurs in a P Counter with Int type.
+func TestPCounterUpdate_IntKindWithPositiveIncrementOverflow_RollsOverToMinInt64(t *testing.T) {
+	test := testUtils.TestCase{
+		Description: "Positive increments of a P Counter with Int type causing overflow behaviour",
+		SupportedMutationTypes: immutable.Some([]testUtils.MutationType{
+			// GQL mutation will return an error
+			// when integer type overflows
```
shahzadlone, line 124: suggestion: add that GQL test as well (separately if not added already).

tests/integration/mutation/update/crdt/pcounter_test.go
```diff
@@ -0,0 +1,264 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents what happens when an overflow occurs in a P Counter with Float type.
+// In this case it is the same as a no-op.
```
shahzadlone, line 221: question: In this case gql doesn't behave differenty?

tests/integration/net/state/simple/peer/crdt/pcounter_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/sourcenetwork/immutable"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestP2PUpdate_WithPCounter_NoError(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			testUtils.RandomNetworkingConfig(),
```
shahzadlone, line 24: question: out of scope, but I saw this in a few tests, wonder why we have to do this `testUtils.RandomNetworkingConfig()` twice?

shahzadlone (commented):
Praise: Nice tests

Mostly questions. Looks good in first pass

merkle/crdt/counter.go
```diff
@@ -21,40 +21,41 @@ import (
...
+	return &MerkleCounter[T]{
 		baseMerkleCRDT: base,
 		reg:            register,
 	}
 }
 
-// Save the value of the PN Counter to the DAG.
-func (mPNC *MerklePNCounter[T]) Save(ctx context.Context, data any) (ipld.Node, uint64, error) {
+// Save the value of the  Counter to the DAG.
+func (mC *MerkleCounter[T]) Save(ctx context.Context, data any) (ipld.Node, uint64, error) {
```
islamaliev, line 50: question: is it conventional to name vars like this? don't we usually do something like "mc"?

tests/integration/mutation/update/crdt/pcounter_test.go
```diff
@@ -0,0 +1,264 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents what happens when an overflow occurs in a P Counter with Int type.
+func TestPCounterUpdate_IntKindWithPositiveIncrementOverflow_RollsOverToMinInt64(t *testing.T) {
```
islamaliev, line 119: question: is this what users would expect?
I guess any attempt to restrict this would make it hard to sync p2p.
But is this behaviour well documented?

tests/integration/mutation/update/crdt/pcounter_test.go
```diff
@@ -0,0 +1,264 @@
...
+					Users {
+						name
+						points
+					}
+				}`,
+				Results: []map[string]any{
+					{
+						"name": "John",
+						// Note the lack of precision of float types.
+						"points": 20.299999999999997,
```
islamaliev, line 210: suggestion: increment by values that sum up to something more precise. Maybe 10.25 and 10.5

tests/integration/mutation/update/crdt/pcounter_test.go
```diff
@@ -0,0 +1,264 @@
...
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents what happens when an overflow occurs in a P Counter with Float type.
+// In this case it is the same as a no-op.
+func TestPCounterUpdate_FloatKindWithPositiveIncrementOverflow_NoOp(t *testing.T) {
```
islamaliev, line 222: todo: is this well documented?

tests/integration/schema/updates/add/field/crdt/pcounter_test.go
```diff
@@ -0,0 +1,73 @@
...
+				Schema: `
+					type Users {
+						name: String
+					}
+				`,
+			},
+			testUtils.SchemaPatch{
+				Patch: `
+					[
+						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "foo", "Kind": 2, "Typ": 5} }
```
islamaliev, line 65: nitpick: it would be easier to read if instead of Kind as a number we specified it as a string: "Int".
This way we don't need to remember or look up the number.
It's my personal preference. Feel free to ignore it.

islamaliev (approved):
LGTM. Just have a few minor comments

merkle/crdt/counter.go
fredcarle: I agree. This is from copy pasting and then a simple search a replace. I'll change them to be idiomatic.

tests/integration/mutation/update/crdt/pcounter_test.go
fredcarle: > But is this behaviour well documented?

Can you clarify your question please. Do you mean internally or as a behaviour of the language? This not an implementation specific issue so I do not think it need to be documented elsewhere.

tests/integration/mutation/update/crdt/pcounter_test.go
fredcarle: it doesn't :)

tests/integration/mutation/update/crdt/pcounter_test.go
fredcarle: I don't think it would help with anything since the reason for the GQL test returning an error is that the GQL Int type is an int32 and here we are sending and int64 to test the internal overflow. I'll update the comment to be more appropriate.

merkle/crdt/merklecrdt.go
fredcarle: ✅

merkle/crdt/counter.go
fredcarle: ✅

tests/integration/mutation/update/crdt/pcounter_test.go
fredcarle: I added a comment to the `Increment` method on the `Counter` type and added to the crdt type descriptions for GQL introspection. Let me know if that's what you wanted :)

tests/integration/mutation/update/crdt/pcounter_test.go
fredcarle: Answered above so resolving this one.

tests/integration/mutation/update/crdt/pcounter_test.go
islamaliev: looks great. 
One more question: wouldn't it make sense to set a float value to max float in case of overflow?

fredcarle: I'm not sure what is the most appropriate solution. It might even be — both for Int and Float — to return an error when the increment will overflow. That would be fine for direct mutation but for P2P sync its probably sub optimal.
@jsimnz what is your opinion on this one.

PR #2495, opened 2024-04-04
test: Add flag to skip network tests

### LINKED ISSUE #2494: Add test flag to disable network tests

nasdf:
## Relevant issue(s)

Resolves #2494 

## Description

This PR adds a test flag set via an environment variable to skip any tests that involve network actions.

## Tasks

## How has this been tested?

`make test`

- MacOS

### DIFF
--- a/tests/integration/test_case.go
+++ b/tests/integration/test_case.go
@@ -593,3 +593,35 @@ type BackupImport struct {
 	// contains this string.
 	ExpectedError string
 }
+
+// IsNetworkAction returns true if the given action involves the network subsystem.
+func IsNetworkAction(act any) bool {
+	switch act.(type) {
+	case ConfigureNode:
+		return true
+
+	case ConnectPeers:
+		return true
+
+	case ConfigureReplicator:
+		return true
+
+	case DeleteReplicator:
+		return true
+
+	case SubscribeToCollection:
+		return true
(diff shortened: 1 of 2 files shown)


tests/integration/test_case.go
```diff
@@ -593,3 +593,35 @@ type BackupImport struct {
 	// contains this string.
 	ExpectedError string
 }
+
+// IsNetworkAction returns true if the given action involves the network subsystem.
+func IsNetworkAction(act any) bool {
+	switch act.(type) {
+	case ConfigureNode:
```
AndrewSisley, line 600: suggestion: I don't think you need to check all of these?  `ConfigureNode` is required for all P2P tests IIRC.  Expecting all future network test actions to consistently be added to this switch seems a little optimistic to me and I would rather avoid the need if we can. 

suggestion: If keeping all the actions in the switch, I think collapsing them into a single case might be easier to read:
```
switch act.(type) {
	case ConfigureNode,
	        ConnectPeers,
	        ConfigureReplicator,
	        etc...:
	      return true
}
```

nasdf: I simplified it to just check for `ConfigureNode`.

AndrewSisley (approved):
LGTM :)

PR #2511, opened 2024-04-09
fix: Allow update when updating non-indexed field

### LINKED ISSUE #2510: Updating non indexed field of doc with unique index fails
Updating unique indexed field on a given doc succeeds but updating a non indexed field of a doc with a unique index fails. The following should succeed but it fails with the error:
```
can not index a doc's field(s) that violates unique index. DocID: bae-b0a35b3c-486b-54fc-8736-eeff036414ef, name: Fred.
```
```go
func TestUniqueIndexUpdate_UponUpdatingDocNonIndexedField_ShouldSucceed(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String @index(unique: true)
						age: Int
					}
				`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"Fred",
						"age":	36
					}`,
			},
			testUtils.UpdateDoc{
				CollectionID: 0,
				DocID:        0,
				Doc: `
					{
						"age":	37
					}`,
			},
		},
	}

}
```

There is also a similar behaviour for composite indexes where the following should succeed but fails with the error:
```
can not index a doc's field(s) that violates unique index. DocID: bae-cae3deac-d371-5a1f-93b4-ede69042f79b, name: John, age: 21
```
```go
func TestUniqueCompositeIndexUpdate_UponUpdatingDocWithExistingFieldValue_ShouldSucceed(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type User @index(unique: true, fields: ["name", "age"]) {
						name: String 
						age: Int 
						email: String
					}
				`,
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `
					{
						"name":	"John",
						"age":	21,
						"email": "email@gmail.com"
					}`,
			},
			testUtils.UpdateDoc{
				CollectionID: 0,
				DocID:        0,
				Doc: `
					{
						"email": "another@gmail.com"
					}`,
			},
		},
	}

}
```

fredcarle:
## Relevant issue(s)

Resolves #2510 

## Description

This PR adds a check to the index validation step where if the resulting KV pair is the same on update as the previous KV pair, the outcome will be an no-op instead of an index existing error.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/db/collection_index.go
+++ b/db/collection_index.go
@@ -196,13 +196,9 @@ func (c *collection) updateIndexedDoc(
 	}
 	txn := mustGetContextTxn(ctx)
 	for _, index := range c.indexes {
-		// We only need to update the index if one of the indexed fields
-		// on the document has been changed.
-		if isUpdatingIndexedFields(index, oldDoc, doc) {
-			err = index.Update(ctx, txn, oldDoc, doc)
-			if err != nil {
-				return err
-			}
+		err = index.Update(ctx, txn, oldDoc, doc)
+		if err != nil {
+			return err
 		}
 	}
 	return nil
@@ -593,26 +589,4 @@ func generateIndexName(col client.Collection, fields []client.IndexedFieldDescri
 		sb.WriteString(strconv.Itoa(inc))
 	}
 	return sb.String()
-}
-
-func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Document) bool {
-	for _, indexedFields := range index.Description().Fields {
-		oldVal, getOldValErr := oldDoc.GetValue(indexedFields.Name)
-		newVal, getNewValErr := newDoc.GetValue(indexedFields.Name)
-
-		// GetValue will return an error when the field doesn't exist.
-		// This will happen for oldDoc only if the field hasn't been set
-		// when first creating the document. For newDoc, this will happen
-		// only if the field hasn't been set when first creating the document
-		// AND the field hasn't been set on the update.
-		switch {
-		case getOldValErr != nil && getNewValErr != nil:
-			continue
-		case getOldValErr != nil && getNewValErr == nil:
-			return true
-		case oldVal.Value() != newVal.Value():
-			return true
-		}
-	}
-	return false
 }
--- a/db/collection_index.go
+++ b/db/collection_index.go
@@ -196,9 +196,13 @@ func (c *collection) updateIndexedDoc(
 	}
 	txn := mustGetContextTxn(ctx)
 	for _, index := range c.indexes {
-		err = index.Update(ctx, txn, oldDoc, doc)
-		if err != nil {
-			return err
+		// We only need to update the index if one of the indexed fields
+		// on the document has been changed.
+		if isUpdatingIndexedFields(index, doc) {
+			err = index.Update(ctx, txn, oldDoc, doc)
+			if err != nil {
+				return err
+			}
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


db/collection_index.go
```diff
@@ -588,3 +592,18 @@ func generateIndexName(col client.Collection, fields []client.IndexedFieldDescri
 	}
 	return sb.String()
 }
+
+func isUpdatingIndexedFields(index CollectionIndex, doc *client.Document) bool {
```
islamaliev, line 596: suggestion: I think it's better to compare 2 docs here. What if an update is requested, but to the same value? Will `val.IsDirty` return true?

You can instead pass here 2 docs and do something like:
```go
for _, indexedFields := range index.Description().Fields {
	newVal := newDoc.GetValue(indexedFields.Name)
	oldVal := oldDoc.GetValue(indexedFields.Name)
	if newVal.Value() != oldVal.Value() {
		return true
	}
}
```
This way you don't need to iterate through all `doc.Fields()` but only indexed ones.

tests/integration/index/update_unique_composite_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+package index
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestUniqueCompositeIndexUpdate_UponUpdatingDocWithExistingFieldValue_ShouldSucceed(t *testing.T) {
```
islamaliev, line 19: praise: thanks for adding a test for the composite one.

islamaliev (approved):
Good catch and good job fixing it! I just have some clarifying comments.

db/collection_index.go
fredcarle: > What if an update is requested, but to the same value? Will val.IsDirty return true?
Yes.

I like that suggestion. It will ensure we don't try to update the index if the indexed field values are set to the same value 👍

fredcarle: ✅

fredcarle: I'll wait for you to check the change before merging.

db/collection_index.go
```diff
@@ -197,9 +197,13 @@ func (c *collection) updateIndexedDoc(
 		return err
 	}
 	for _, index := range c.indexes {
-		err = index.Update(ctx, txn, oldDoc, doc)
-		if err != nil {
-			return err
+		// We only need to update the index if one of the indexed fields
+		// on the document has been changed.
+		if isUpdatingIndexedFields(index, oldDoc, doc) {
```
jsimnz, line 202: question(cc: @islamaliev): Should this be done inside the `index.Update` func, instead of here? This will A) allow different index types to handle this differently B) Gurantees this behavior is consistent regardless of where `index.Update` is called from (vs doing it here which is only fixed for this particular call path).

db/collection_index.go
fredcarle: @islamaliev still waiting for you to check the change :)

db/collection_index.go
islamaliev: I also thought about this, but though it would be unlikely to have any other update behaviour.
Although, thinking about it again we don't know what will be in the future especially when we decide to index array or json fields or some other fancy stuff.

fredcarle: Moved ✅

db/index.go
```diff
@@ -381,6 +381,11 @@ func (index *collectionUniqueIndex) Update(
 	oldDoc *client.Document,
 	newDoc *client.Document,
 ) error {
+	// We only need to update the index if one of the indexed fields
+	// on the document has been changed.
+	if !isUpdatingIndexedFields(index, oldDoc, newDoc) {
```
islamaliev, line 386: suggestion: I think it also makes sense to do this check for regular indexes otherwise they will do unnecessary work with storage.

islamaliev (commented):
LGTM! Just have a final suggestion.

db/index.go
fredcarle: I'll open a new PR to cover that case 👍

PR #2513, opened 2024-04-10
refactor: DB transactions context

### LINKED ISSUE #2512: Move db transactions to context

### LINKED ISSUE #2516: Implicit Collection transactions aren't respected.
**Describe the problem**
Some collection call paths don't respect the intended transaction scope. Haven't looked all every method yet, but in can be seen in the `collection.updateWithDocID` (private func), which is given a transaction in its arguments, but then makes a call to `collection.Get` (public func), which doesn't get a transaction. This only affects the *implicit* transaction that collections will create if they haven't *explicity* (ie: `collection.WithTxn`) been given one.

This is because the `collection.getTxn` will create a new transaction on every call unless there is an existing transaction on the `collection` struct.

**To Reproduce**
Any public Collection call which uses implicit transactions that *then* calls another public Collection function.

**Expected behavior**
All DB calls respect the implicit or explicit transaction for the entire lifecycle/scope of that call.

nasdf:
## Relevant issue(s)

Resolves #2512
Resolves #2516

## Description

This PR moves the db transactions to the context.

Notable Changes:
- moved `txn_db.go` to `store.go`
- ~replaced `explicitTxnDB` and `implicitTxnDb` with `store`~
- ~added `db/session.go` to simplify setting context~
- added `db/context.go` to manage context values

## Tasks

## How has this been tested?

`make test`

- MacOS


db/session/session.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/sourcenetwork/defradb/datastore"
+)
+
+type contextKey string
+
+const (
+	TxnContextKey = contextKey("txn")
```
fredcarle, line 22: suggestion: To avoid allocation on the key, you can simply use
```go
type TxnContextKey struct{}
```
https://pkg.go.dev/context#WithValue

db/collection.go
```diff
@@ -1240,11 +1230,11 @@ func (c *collection) GetAllDocIDs(
 	ctx context.Context,
 	identity immutable.Option[string],
 ) (<-chan client.DocIDResult, error) {
-	txn, err := c.getTxn(ctx, true)
+	ctx, err := ensureContextTxn(ctx, c.db, true)
 	if err != nil {
 		return nil, err
 	}
-
+	txn := mustGetContextTxn(ctx)
```
fredcarle, line 1237: thought: It's a little weird that whe need to extract the `txn` from the context and the next function call passes the context and the transaction.

question: Why not leave it to the function that needs the `txn` to get it from the context?

db/context.go
```diff
@@ -0,0 +1,72 @@
...
+// transactionDB is a db that can create transactions.
+type transactionDB interface {
+	NewTxn(context.Context, bool) (datastore.Txn, error)
+}
+
+// ensureContextTxn ensures that the returned context has a transaction.
+//
+// If a transactions exists on the context it will be made explicit,
+// otherwise a new implicit transaction will be created.
+func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (context.Context, error) {
```
fredcarle, line 44: suggestion: There is a lot of `ensureContextTxn` followed by `mustGetContextTxn`. To avoid having to type cast everytime, I would suggest returning `datastore.Txn` here. 
```go
func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (context.Context, datastore.Txn, error)
```

db/collection.go
nasdf: Someone had a preference for this in the standup discussion. If a function requires a transaction it should remain as a parameter and not come from context.

db/store.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/lens-vm/lens/host-go/config/model"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+var _ client.Store = (*store)(nil)
+
+type store struct {
```
fredcarle, line 25: question: Can you explain why you prefer wrapping the `db` instead of using it directly for the methods bellow?

nasdf: Good point. There's no need for a separate struct anymore. Would you prefer these functions remain in `store.go` or would it be better to move them into `db.go`?

http/handler_ccip.go
```diff
@@ -35,7 +35,7 @@ type CCIPResponse struct {
 
 // ExecCCIP handles GraphQL over Cross Chain Interoperability Protocol requests.
 func (c *ccipHandler) ExecCCIP(rw http.ResponseWriter, req *http.Request) {
-	store := req.Context().Value(storeContextKey).(client.Store)
+	db := req.Context().Value(dbContextKey).(client.DB)
```
fredcarle, line 38: thought: There is nothing wrong with keeping the more restrictive type casting of `client.Store` since that is really what is needed here.

net/server.go
```diff
@@ -349,15 +353,14 @@ func (s *server) syncIndexedDocs(
 	ctx context.Context,
 	col client.Collection,
 	docID client.DocID,
+	txn datastore.Txn,
```
fredcarle, line 356: todo: Remove unused param.

fredcarle (commented):
I like the changes. I'll just wait for the comments to be resolved before approving.

I'm wondering if we should remove the `txn` from function params when the context already holds it. Probably out of scope for this PR but could be done shortly after.

Edit: Will bring it up in the standup.

db/store.go
fredcarle: `store.go` is fine with me. We could even add documentation at the top saying that the fonction within implement the `client.Store` interface for `db`.

jsimnz:
I'm not a huge fan of this PR, not sure if its the specific implementation or the general concept (txn via context). This was my concern when discussing it in the ACP discord thread regarding the use of txn via the context making the depedency more opaque. It can make following the call stack of a given transaction harder to reason about and easier to mess up.

Heres an example in the `basicImport` func

It has the following function signature
image

It uses the `txn` as an argument to a follow-up DB call (note the private method being called which specifically has the txn argument)

image

Then makes a call to the a public function which doesn't use the txn

image.

It is not 100% clear the txn scope the last call is using, since the calling function (`basicImport`) is explicitly given a txn.

Before, the last line was `col.WithTxn(txn).Create(...)` which makes it very obvious the txn being used.

--

Theres a possible solution here:
All actual DB/Collection functionality is implemented in the private funcs (which we basicallly do now), all private funcs *only* call private funcs, and private funcs take `txn` as a input args.

***in fact*** after reviewing this, and the existing (develop) implementation, there is actually a bug where some call paths don't respect transaction scopes correctly because of the implicit `txn` creation on the `Collection` public calls.

Oddly enough, this PR actually fixes that without intending to because of the shared context.

Perhaps my issue isn't necessarily with the transaction context propagation, but the inconsistency we have between implicit and explicit transactions and how they are handled, both on the `db` object and the `Collection` object.

Happy to chat more during the dev call today (cc @nasdf @fredcarle )

db/context.go
```diff
@@ -0,0 +1,66 @@
...
+// TryGetContextTxn returns a transaction and a bool indicating if the
+// txn was retrieved from the given context.
+func TryGetContextTxn(ctx context.Context) (datastore.Txn, bool) {
+	txn, ok := ctx.Value(txnContextKey{}).(datastore.Txn)
+	return txn, ok
+}
+
+// SetContextTxn returns a new context with the txn value set.
+func SetContextTxn(ctx context.Context, txn datastore.Txn) context.Context {
+	return context.WithValue(ctx, txnContextKey{}, txn)
```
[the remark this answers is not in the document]
nasdf: The `ensureContextTxn` makes sure we don't overwrite an existing transaction. There is a case within `net/server.go` where we need the functionality to overwrite the transaction.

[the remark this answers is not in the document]
islamaliev: not sure if having `ensureContextTxn` is enough in this case. `SetContextTxn` is a public method and can be called from anywhere. If `ensureContextTxn` is required to call before calling this one, why not to make it part of it? Otherwise I see no link between the two.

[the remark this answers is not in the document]
nasdf: I'm not seeing the issue here. The lifetime of the transaction is managed from the caller. If they overwrite a transaction and don't discard it, then the bug is in the caller's code.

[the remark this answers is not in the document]
islamaliev: if the intention is to put this responsibility on the caller's side, it's fine, but then it needs to be documented.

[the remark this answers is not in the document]
nasdf: I've updated the function docs to be more clear.

fredcarle (approved):
LGTM

PR #2518, opened 2024-04-12
refactor(i): Remove datastore.Txn from private db methods

### LINKED ISSUE #2517: Remove `txn` parameter from private `db` funcs

nasdf:
## Relevant issue(s)

Resolves #2517 

## Description

This PR removes the `datastore.Txn` parameter from private db methods.

## Tasks

## How has this been tested?

`make test`

- MacOS


db/context.go
```diff
@@ -53,6 +53,11 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 	return SetContextTxn(ctx, txn), txn, nil
 }
 
+// mustGetContextTxn returns the transaction from the context or panics.
+func mustGetContextTxn(ctx context.Context) datastore.Txn {
+	return ctx.Value(txnContextKey{}).(datastore.Txn)
```
AndrewSisley, line 58: todo: This is a user input, and I think an error if not found would be much better than a panic - unless it is always guaranteed to exist at the point that this function is called, in which case please document this in this function.

db/context.go
```diff
@@ -53,6 +53,11 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 	return SetContextTxn(ctx, txn), txn, nil
 }
 
+// mustGetContextTxn returns the transaction from the context or panics.
+func mustGetContextTxn(ctx context.Context) datastore.Txn {
```
AndrewSisley, line 57: suggestion: This feels like an odd place for such a function, why is this defined privately in `db`, instead of internal-public in `core`?  Is the long term plan to move it, or to duplicate it? Or have we managed to remove all `txn` usage from all other packages?

db/context.go
```diff
@@ -53,6 +53,11 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 	return SetContextTxn(ctx, txn), txn, nil
```
AndrewSisley, line 53: suggestion: Sorry I missed the PR in which this was added, would you mind noting that a new `context.Context` object may be returned from this function (in the case of implicit txn)? It is not clear from the existing documentation.

Something along the lines of:
> ... otherwise a new implicit transaction will be created, and returned within a new context object also containing the copied values from the input context.

I know this is expected with Go contexts, but it is important and thus worth being nice and explicit here IMO.

db/context.go
nasdf: From the standup discussion, our consensus was to use this only within the `db` package. Other APIs will still use `txn` params. I agree that it should move to `core` if we decide to expand it to other areas.

db/context.go
nasdf: I've updated the docs to better reflect our usage pattern. It should be guaranteed to exist on any private function with `db`.

AndrewSisley: Nice - new doc changes look good, thanks Keenan :)

AndrewSisley (approved):
LGTM, thanks Keenan :)

db/context.go
jsimnz: Just adding some *context* here ;). A new `context.Context` will *always* be returned regardless of whether the implicit txn is created. Since the existing `txn` on the context is converted to an `explicityTxn` and then re-added to the context, thus creating a new context (since they are immutable).

PR #2528, opened 2024-04-15
refactor(i): DB identity context

### LINKED ISSUE #2439: Cleanup The Identity Param
Andy's suggestion in https://github.com/sourcenetwork/defradb/pull/2338#discussion_r1534171165:

> thought: This new param does dirty the interfaces a little bit IMO I'm fine with this being out of scope for now, but in a follow up I would like to see something like what we did for transactions - with a WithIndentity() DB func on the DB interface. Maybe add a ticket if not doing in this PR.

John's Suggestion:

> For this, you currently pass the "identity" as an argument, but in reality, it should be passed in as a Context value
eg: you have col.CreateMany(cmd.Context(), identity, docs), where you are passing in the identity value, as a param. But to make it cleaner, you should create a new context with the identity value injected into it

nasdf:
## Relevant issue(s)

Resolves #2439

## Description

This PR moves the identity param from the `client.DB` and `client.Collection` interfaces to the context.

Notable changes:
- `acp/identity.go` added `Identity` type alias to make future updates simpler

## Tasks

## How has this been tested?

`make test`

- MacOS

### DIFF
--- a/acp/identity/identity.go
+++ b/acp/identity/identity.go
@@ -18,17 +18,17 @@ package identity
 type Identity string
 
 var (
-	// NoIdentity is an empty identity.
-	NoIdentity = Identity("")
+	// None is an empty identity.
+	None = Identity("")
 )
 
-// New makes a new identity if the input is not empty otherwise, returns an empty Option.
+// New makes a new identity if the input is not empty otherwise, returns None.
 func New(identity string) Identity {
 	// TODO-ACP: There will be more validation once sourcehub gets some utilities.
 	// Then a validation function would do the validation, will likely do outside this function.
 	// https://github.com/sourcenetwork/defradb/issues/2358
 	if identity == "" {
-		return NoIdentity
+		return None
 	}
 	return Identity(identity)
 }
--- a/acp/identity/identity.go
+++ b/acp/identity/identity.go
@@ -18,18 +18,21 @@ import (
 	"github.com/sourcenetwork/immutable"
 )
 
+// Identity is the unique identifier for an actor.
+type Identity immutable.Option[string]
+
 var (
 	// NoIdentity is an empty identity.
-	NoIdentity = immutable.None[string]()
+	NoIdentity = Identity(immutable.None[string]())
 )
 
 // NewIdentity makes a new identity if the input is not empty otherwise, returns an empty Option.
-func NewIdentity(identity string) immutable.Option[string] {
+func NewIdentity(identity string) Identity {
 	// TODO-ACP: There will be more validation once sourcehub gets some utilities.
 	// Then a validation function would do the validation, will likely do outside this function.
 	// https://github.com/sourcenetwork/defradb/issues/2358
 	if identity == "" {
 		return NoIdentity
 	}
-	return immutable.Some[string](identity)
+	return Identity(immutable.Some[string](identity))
 }
--- a/acp/identity/identity.go
+++ b/acp/identity/identity.go
@@ -36,3 +36,13 @@ func NewIdentity(identity string) Identity {
 	}
 	return Identity(immutable.Some[string](identity))
 }
+
+// HasValue returns true if the identity is not empty.
+func (i Identity) HasValue() bool {
+	return immutable.Option[string](i).HasValue()
+}
+
+// Value returns the identity value.
+func (i Identity) Value() string {
+	return immutable.Option[string](i).Value()
+}
--- a/acp/identity/identity.go
+++ b/acp/identity/identity.go
@@ -14,23 +14,25 @@ Package identity provides defradb identity.
 
 package identity
 
+import "github.com/sourcenetwork/immutable"
+
 // Identity is the unique identifier for an actor.
 type Identity string
 
 var (
 	// None is an empty identity.
-	None = Identity("")
+	None = immutable.None[Identity]()
 )
 
 // New makes a new identity if the input is not empty otherwise, returns None.
-func New(identity string) Identity {
+func New(identity string) immutable.Option[Identity] {
 	// TODO-ACP: There will be more validation once sourcehub gets some utilities.
 	// Then a validation function would do the validation, will likely do outside this function.
 	// https://github.com/sourcenetwork/defradb/issues/2358
 	if identity == "" {
 		return None
 	}
-	return Identity(identity)
+	return immutable.Some(Identity(identity))
 }
 
 // String returns the string representation of the identity.
--- a/acp/identity/identity.go
+++ b/acp/identity/identity.go
@@ -14,35 +14,26 @@ Package identity provides defradb identity.
 
 package identity
 
-import (
-	"github.com/sourcenetwork/immutable"
-)
-
 // Identity is the unique identifier for an actor.
-type Identity immutable.Option[string]
+type Identity string
 
 var (
 	// NoIdentity is an empty identity.
-	NoIdentity = Identity(immutable.None[string]())
+	NoIdentity = Identity("")
 )
 
-// NewIdentity makes a new identity if the input is not empty otherwise, returns an empty Option.
-func NewIdentity(identity string) Identity {
+// New makes a new identity if the input is not empty otherwise, returns an empty Option.
+func New(identity string) Identity {
 	// TODO-ACP: There will be more validation once sourcehub gets some utilities.
 	// Then a validation function would do the validation, will likely do outside this function.
 	// https://github.com/sourcenetwork/defradb/issues/2358
 	if identity == "" {
 		return NoIdentity
 	}
-	return Identity(immutable.Some[string](identity))
+	return Identity(identity)
 }
 
-// HasValue returns true if the identity is not empty.
-func (i Identity) HasValue() bool {
-	return immutable.Option[string](i).HasValue()
-}
-
-// Value returns the identity value.
-func (i Identity) Value() string {
-	return immutable.Option[string](i).Value()
+// String returns the string representation of the identity.
+func (i Identity) String() string {
+	return string(i)
 }
--- a/db/collection.go
+++ b/db/collection.go
@@ -1228,7 +1228,7 @@ func (db *db) getAllActiveDefinitions(ctx context.Context) ([]client.CollectionD
 func (c *collection) GetAllDocIDs(
 	ctx context.Context,
 ) (<-chan client.DocIDResult, error) {
-	ctx, _, err := ensureContextValues(ctx, c.db, true)
+	ctx, _, err := ensureContextTxn(ctx, c.db, true)
 	if err != nil {
 		return nil, err
 	}
@@ -1342,7 +1342,7 @@ func (c *collection) Create(
 	ctx context.Context,
 	doc *client.Document,
 ) error {
-	ctx, txn, err := ensureContextValues(ctx, c.db, false)
+	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return err
 	}
@@ -1362,7 +1362,7 @@ func (c *collection) CreateMany(
 	ctx context.Context,
 	docs []*client.Document,
 ) error {
-	ctx, txn, err := ensureContextValues(ctx, c.db, false)
+	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return err
 	}
@@ -1445,7 +1445,7 @@ func (c *collection) Update(
 	ctx context.Context,
 	doc *client.Document,
 ) error {
-	ctx, txn, err := ensureContextValues(ctx, c.db, false)
+	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return err
 	}
@@ -1506,7 +1506,7 @@ func (c *collection) Save(
 	ctx context.Context,
 	doc *client.Document,
 ) error {
-	ctx, txn, err := ensureContextValues(ctx, c.db, false)
+	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return err
 	}
@@ -1780,7 +1780,7 @@ func (c *collection) Delete(
 	ctx context.Context,
 	docID client.DocID,
 ) (bool, error) {
-	ctx, txn, err := ensureContextValues(ctx, c.db, false)
+	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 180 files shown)


db/context.go
```diff
@@ -64,6 +76,14 @@ func mustGetContextTxn(ctx context.Context) datastore.Txn {
 	return ctx.Value(txnContextKey{}).(datastore.Txn)
 }
 
+// mustGetContextIdentity returns the identity from the context or panics.
+//
+// This should only be called from private functions within the db package
+// where we ensure an identity always exists.
+func mustGetContextIdentity(ctx context.Context) acpIdentity.Identity {
```
fredcarle, line 83: thought: I don't think we need the same pattern for identity as for transactions. There isn't a case where something is wrong if the identity hasn't been provided. I think we could get away with setting the identity as you do now and getting the identity returning a `NoIdentity` if one isn't present in the context. What do you think?

edit: Come to think of it, we could probably apply the same logic to transactions. We could have just a set and get function here as well with the get returning a new txn if one hasn't been set.

http/client_acp.go
```diff
@@ -23,6 +23,8 @@ import (
 type AddPolicyRequest struct {
 	// Policy body in JSON or YAML format.
 	Policy string `json:"policy"`
+	// CreateID is the acp identity of the policy creator.
+	CreateID string `json:"creator_id"`
```
fredcarle, line 27: todo: `CreateID` -> `CreatorID`

http/middleware.go
```diff
@@ -112,3 +123,23 @@ func CollectionMiddleware(next http.Handler) http.Handler {
 		next.ServeHTTP(rw, req.WithContext(ctx))
 	})
 }
+
+func IdentityMiddleware(next http.Handler) http.Handler {
```
fredcarle, line 127: praise: This is very nice 👍

fredcarle (changes requested):
One more negative line contribution PR adding to your total. I'm loving this haha!

A few comments to resolve before I make a final pass. Ping me if you want to discuss this live :)

db/collection.go
```diff
@@ -1227,18 +1227,16 @@ func (db *db) getAllActiveDefinitions(ctx context.Context) ([]client.CollectionD
 // it hits every key and will cause Tx conflicts for concurrent Txs
 func (c *collection) GetAllDocIDs(
 	ctx context.Context,
-	identity immutable.Option[string],
 ) (<-chan client.DocIDResult, error) {
-	ctx, _, err := ensureContextTxn(ctx, c.db, true)
+	ctx, _, err := ensureContextValues(ctx, c.db, true)
```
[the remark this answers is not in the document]
nasdf: I agree with the changes to `TryGetIdentity`. I do think the pattern of calling `ensureContextTxn` from the public functions helps with organization, but happy to reconsider.

db/context.go
nasdf: I agree, the Identity calls can be simplified. Transactions are a bit more complicated because we need to discard / commit the implicit ones. If we used that pattern for transactions it would be hard to keep track of where to discard / commit.

tests/integration/acp/register_and_delete_test.go
```diff
@@ -206,7 +206,8 @@ func TestACP_CreateWithoutIdentityAndDeleteWithIdentity_CanDelete(t *testing.T)
 func TestACP_CreateWithIdentityAndDeleteWithIdentity_CanDelete(t *testing.T) {
 	// OwnerIdentity should be the same identity that is used to do the registering/creation,
 	// and the final read check to see the state of that registered document.
-	OwnerIdentity := Actor1Signature
+	OwnerIdentity := Actor1Identity
+	OwnerSignature := Actor1Signature
```
shahzadlone, line 210: question: Why do we need both here? And all other similar instances?

tests/integration/acp/register_and_delete_test.go
```diff
@@ -215,7 +216,7 @@ func TestACP_CreateWithIdentityAndDeleteWithIdentity_CanDelete(t *testing.T) {
 		Actions: []any{
 			testUtils.AddPolicy{
 
-				Creator: OwnerIdentity,
+				Creator: OwnerSignature,
```
shahzadlone, line 219: todo: similar to previous todo update all occurrences of signature to identity.

tests/integration/utils2.go
```diff
@@ -1204,11 +1199,12 @@ func createDocViaColSave(
...
 	}
 
-	return doc, collections[action.CollectionID].Save(
-		s.ctx,
-		acpIdentity.NewIdentity(action.Identity),
-		doc,
-	)
+	txn := getTransaction(s, node, immutable.None[int](), action.ExpectedError)
+
+	ctx := db.SetContextTxn(s.ctx, txn)
```
shahzadlone, line 1204: question: Does setting order matter? 

suggestion: Even if order doesn't matter suggest, perhaps a function that sets all contexts so it's less error prone and consistent in all places we need to set both.

acp/identity/identity.go
```diff
@@ -18,18 +18,31 @@ import (
 	"github.com/sourcenetwork/immutable"
 )
 
+// Identity is the unique identifier for an actor.
+type Identity immutable.Option[string]
```
shahzadlone, line 22: todo: I dislike the introduction of this type for the following reasons:

1) This seems like an independent change to this PR, I don't think Identity Context propagation stuff needs this (correct me if I am wrong).

2) Identity being optional is not and should not be part of the Identity type.

3) `immutable.Option[string]` is more readable IMO. 

4) Introduction of wrapper `HasValue` and `Value` functions which immutable provides already.

5) Currently the `Identity` is just a string, and should stay like that. In future when authentication is implemented, the `Identity` type will evolve into a `struct` type which will then be used like `immutable.Option[Identity]`

6) The abstraction should likely be done when Identity type becomes more complex (likely when authentication is done).

I also noticed that testing frameworks Identity params changed as well, I share similar thoughts for them, they should just be a string for now.

shahzadlone (commented):
Cheers, mostly looks good. Appreciate you taking care of this.

Did first pass from my phone, got some todos for you.

acp/identity/identity.go
fredcarle: The usage of the `Identity` type is to avoid having to change func param types when we change from string to some other type. That being said, we could achieve the same thing and keep the immutability out of the type definition if we did the following instead:
```go
type Identity string
```
end then use `immutable.Option[Identity]` as param type.

tests/integration/acp/fixture.go
```diff
@@ -10,5 +10,13 @@
...
 
-var Actor1Signature = "cosmos1zzg43wdrhmmk89z3pmejwete2kkd4a3vn7w969"
-var Actor2Signature = "cosmos1x25hhksxhu86r45hqwk28dd70qzux3262hdrll"
+import (
+	"github.com/sourcenetwork/defradb/acp/identity"
+)
+
+var (
+	Actor1Signature = "cosmos1zzg43wdrhmmk89z3pmejwete2kkd4a3vn7w969"
+	Actor2Signature = "cosmos1x25hhksxhu86r45hqwk28dd70qzux3262hdrll"
```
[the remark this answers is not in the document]
fredcarle: That's not new, it comes from your ACP PR.

tests/integration/acp/register_and_delete_test.go
fredcarle: Currently, we represent the `identity` internally as a `string`. If that is what it will always be, then `identity` everywhere is fine. If we foresee that the type will evolve from a `string` to something more complex (which was previously discussed), then it doesn't make sense to call the Auth `string` that comes in with a request the `identity`. It has to have a different name and `signature` would make sense. Or maybe `token` would be appropriate.

tests/integration/utils2.go
fredcarle: The order does not matter. If we want to create a private function for the test utils and cli specifically that's fine with me. But a global function would be more error prone than the explicitness of setting each individually pretty much everywhere else we need them.

http/client_acp.go
```diff
@@ -33,7 +35,8 @@ func (c *Client) AddPolicy(
 	methodURL := c.http.baseURL.JoinPath("acp", "policy")
 
 	addPolicyRequest := AddPolicyRequest{
-		Policy: policy,
+		Policy:    policy,
+		CreatorID: creator,
```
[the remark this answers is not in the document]
nasdf: We discussed this in the standup. The main concern was that identity is required for this function but optional for all others. Most of us thought keeping it a param was the better option, but happy to continue the discussion here.

acp/identity/identity.go
nasdf: I agree the `immutable.Option` is the wrong type here. I would like to keep the custom type as it will make future changes simpler. I will look into changing the alias to `type Identity string`.

tests/integration/acp/fixture.go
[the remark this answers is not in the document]
shahzadlone: I missed some instances that should have been removed.

acp/identity/identity.go
shahzadlone: > The usage of the `Identity` type is to avoid having to change func param types when we change from string to some other type. That being said, we could achieve the same thing and keep the immutability out of the type definition if we did the following instead:
> ```go
> type Identity string
> ```
> end then use `immutable.Option[Identity]` as param type.

I am okay with this alias and param type suggestion.

http/client_acp.go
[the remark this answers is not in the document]
shahzadlone: Thanks for the update, just wondering though that how does a user then specify this through JWT ?

tests/integration/acp/register_and_delete_test.go
shahzadlone: I look at Identity as whatever bits and pieces belong to identifying an actor. For now that is just the Public Sourcehub Account Address (string). Soon this might be a combination of Priv/Pub key pair. So Identity IMO encapsulates this well.

Signature doesn't make sense if it can have multiple "signatures" in the identity struct. IIRC on protocol side of things signature means something else which would always confuse inter team communication when I kept referring to the sourcehub address as signature.

http/client_acp.go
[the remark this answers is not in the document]
jsimnz: > We discussed this in the standup. The main concern was that identity is required for this function but optional for all others.

Shahzad has a point here, I don't think the conversation convered the HTTP client flow per-se. The parameter can be left in for the `client.Store` interface consistency, but in general for the HTTP client it should be using a Auth Header to communicate the identity of the person making the request, which is then authenticated/verified on the DB (eg in the middleware).

Obviously at the moment we don't do any authentication/verification of the identity so the middleware would just make sure that the provided `AddPolicyRequest` creator matches the provded Auth Header creator.

tests/integration/acp/register_and_delete_test.go
jsimnz: Agree with shahzad here, "Identity" is already the umbrella term, when we get to adding proper JWS based request tokens, this is still "Identity" just a verifable one.

At the end of the day, for Identity, we only need a unique identifier, and a verified public key (and potentially a private key, but that is only for embedded database flows, which we can ignore for this convo)

acp/identity/identity.go
jsimnz: Seems like resolved thread but chiming in anyways ;). I agree with Fred/Keenan on this. Using a custom type for this future proofs the changes we know we need to make. But also, yes the type should just be `string` and not `Option[string]`. Cheers

http/client_acp.go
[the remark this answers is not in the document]
nasdf: > but in general for the HTTP client, it should be using a Auth Header to communicate the identity of the person making the request, which is then authenticated/verified on the DB (eg in the middleware).

We already do this for all other methods.

Because this method has an explicit identity / creator param it should be transmitted this way. If we want to use the context identity we should remove the param from the client interface.

acp/identity/identity.go
```diff
@@ -14,22 +14,26 @@ Package identity provides defradb identity.
...
-	// NoIdentity is an empty identity.
-	NoIdentity = immutable.None[string]()
+	// None is an empty identity.
+	None = Identity("")
 )
 
-// NewIdentity makes a new identity if the input is not empty otherwise, returns an empty Option.
-func NewIdentity(identity string) immutable.Option[string] {
+// New makes a new identity if the input is not empty otherwise, returns None.
+func New(identity string) Identity {
```
shahzadlone, line 26: question: why not:
```suggestion
func New(identity string) immutable.Option[Identity] {
```

shahzadlone (commented):
Would be nice to do: `make docs` command pre-merge @nasdf

shahzadlone (commented):
Would be nice to do: `make docs` command pre-merge @nasdf

shahzadlone (commented):
Would be nice to do: `make docs` command pre-merge @nasdf

acp/identity/identity.go
nasdf: I've made identity an `immutable.Option` everywhere again.

shahzadlone: Thanks buddy

fredcarle (approved):
LGTM. Nice work Keenan. Just one thought that might be worth confirming before merging.

lens/fetcher.go
```diff
@@ -61,7 +62,7 @@ func NewFetcher(source fetcher.Fetcher, registry client.LensRegistry) fetcher.Fe
 
 func (f *lensedFetcher) Init(
 	ctx context.Context,
-	identity immutable.Option[string],
+	id immutable.Option[identity.Identity],
```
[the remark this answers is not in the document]
nasdf: I agree `id` is probably too vague here. I don't like shadowing the `identity` package either. 

I think we either need to:
1. alias the `identity` package as `acpIdentity` 
2. merge the `identity` package into `acp` so we can instead use `acp.Identity`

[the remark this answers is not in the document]
shahzadlone: is why I had aliased all instances to acpIdentity previously. Not sure why that was changed.

I do have a preference to use identity over id.

I don't think that we should merge the identity package into acp package.

http/client_acp.go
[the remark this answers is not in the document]
shahzadlone: I thought about this a little. I think we can keep it the way it is now (without the auth header). Reworking it should be easy if need be

lens/fetcher.go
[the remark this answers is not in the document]
fredcarle: `acpIndentity` is not very useful if we take care of the shadowing in other ways.

No need to merge `indentity` into `acp`. We can just alias the `identity.Identity` into `acp.Indentity`

```go
// in /acp
type Identity = identity.Identity
```

This solve the issue very cleanly.

http/client_acp.go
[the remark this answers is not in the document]
fredcarle: @shahzadlone we decided in the standup to move it to the header and change the API method params.

lens/fetcher.go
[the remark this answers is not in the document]
nasdf: The alias makes it so we can't access the methods on the `Identity` without a type cast. With that in mind I decided to use the import alias.

PR #2530, opened 2024-04-16
fix(i): Correctly handle inverse lenses from multiple sources

### LINKED ISSUE #2433: Probable bug migrating backwards multiple versions
Fred spotted the below code in `lens/history.go`, it looks very suspicious and I suspect that there may be a test gap migrating backwards multiple versions.  Find/close the test gap and fix the bug.
```
historyItem.previous = append(
  historyItem.next,
  src,
)
```
Code almost certainly should read:
```
historyItem.previous = append(
  historyItem.previous,
  src,
)
```

AndrewSisley:
## Relevant issue(s)

Resolves #2433

## Description

Correctly handles inverse lenses from multiple sources.

PR includes a new test for verifying that inversing across multiple versions does work, it is not relevant to the bug fix, but I wrote it to make sure I want reading the code wrong (as I first did when creating the issue).

The bug is not testable atm, as this would only have become a problem when introducing multi-source collections (atm both `next` and `previous` slices are always empty before calling this line).  Was likely copy-paste code.

### DIFF
--- a/lens/history.go
+++ b/lens/history.go
@@ -174,7 +174,7 @@ func getCollectionHistory(
 			srcSchemaVersion := schemaVersionsByColID[source.SourceCollectionID]
 			src := history[srcSchemaVersion]
 			historyItem.previous = append(
-				historyItem.next,
+				historyItem.previous,
 				src,
 			)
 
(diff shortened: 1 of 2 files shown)


nasdf (approved):
LGTM

PR #2531, opened 2024-04-19
refactor: Merge collection UpdateWith and DeleteWith

### LINKED ISSUE #2457: Merge `UpdateWith` / `DeleteWith` functions
The client API currently has the following batch update / delete methods:

- `UpdateWith`
- `UpdateWithFilter`
- `UpdateWithDocID`
- `UpdateWithDocIDs`
- `DeleteWith`
- `DeleteWithFilter`
- `DeleteWithDocID`
- `DeleteWithDocIDs`

There is an opportunity to merge them into a compact form that keeps their functionality relatively the same using only filters:

- `UpdateWithFilter`
- `DeleteWithFilter`

nasdf:
## Relevant issue(s)

Resolves #2457 

## Description

This PR merges the `DeleteWith` and `UpdateWith` functions into a singular `DeleteWithFilter` and `UpdateWithFilter`.

## Tasks

## How has this been tested?

`make test`

- MacOS


islamaliev (approved):
LGTM

client/collection.go
```diff
@@ -78,21 +78,6 @@ type Collection interface {
...
-	//
-	// Returns an ErrInvalidUpdateTarget error if the target type is not supported.
-	// Returns an ErrInvalidUpdater error if the updater type is not supported.
-	UpdateWith(
-		ctx context.Context,
-		target any,
-		updater string,
-	) (*UpdateResult, error)
-
 	// UpdateWithFilter updates using a filter to target documents for update.
```
AndrewSisley, line 81: suggestion: Just `Update` and `Delete` would be better names IMO, especially given that the filtering is optional - they might be calling `UpdateWithFilter` with no filter.

planner/update.go
```diff
@@ -72,7 +72,15 @@ func (n *updateNode) Next() (bool, error) {
 			if err != nil {
 				return false, err
 			}
-			_, err = n.collection.UpdateWithDocID(n.p.ctx, docID, string(patch))
+			doc, err := n.collection.Get(n.p.ctx, docID, false)
```
AndrewSisley, line 75: todo: Why are you fetching the value here? You already have it `n.currentValue = n.results.Value()`

nasdf: I haven't worked with the planner much. Is there a way to create a `client.Document` from a `core.Doc`?

AndrewSisley: I missed that the types here were different, but yes you can - easiest way would probably be:
```
docMap := n.currentValue.ToMap()
clientDoc := client.NewDocFromMap(docMap, colDef)
```
There might be something nicer somewhere, but I couldn't spot anything in a 30sec scan.  I prefer the double-cast over fetching twice.

nasdf: Took a few tries but I think I found a good solution.

cli/collection_update.go
```diff
@@ -48,38 +48,14 @@ Example: update private docIDs, with identity:
...
-			case filter != "" && updater != "":
+			case len(filter) > 0 && len(updater) > 0:
 				res, err := col.UpdateWithFilter(cmd.Context(), filter, updater)
 				if err != nil {
 					return err
 				}
 				return writeJSON(cmd, res)
-			case len(argDocIDs) == 1 && len(args) == 1:
-				docID, err := client.NewDocIDFromString(argDocIDs[0])
+			case len(argDocID) > 0 && len(args) == 1:
```
fredcarle, line 57: todo: The use of `len(argDocID) > 0` is misleading here as it suggests that `argDocID` could be an array of `docID`s but then used as a string bellow. Please change this to `argDocID != ""`. Same thing in the `collection_delete.go` file :)

fredcarle (approved):
LGTM. Another negative line contribution PR. Keep at it haha.

Just one minor todo before merge.

planner/update.go
```diff
@@ -64,15 +62,18 @@ func (n *updateNode) Next() (bool, error) {
...
-				return false, err
+
+			docMap := map[string]any{
+				request.DocIDFieldName: n.currentValue.GetID(),
+			}
+			for k, v := range n.input {
+				docMap[k] = v
 			}
-			patch, err := json.Marshal(n.input)
+			doc, err := client.NewDocFromMap(docMap, n.collection.Schema())
```
AndrewSisley, line 72: todo: Now that the col vs schema PR is merged, this probably needs to change to:
```
doc, err := client.NewDocFromMap(docMap, n.collection)
```

AndrewSisley (approved):
LGTM, thanks Keenan :)

planner/update.go
nasdf: It's actually not possible to update a document without fetching it first. Partial document updates break some parts of our indexing system.

It's also not possible to reconstruct the document from the planner node because it only selects the field values the user has requested be returned.

I've reverted back to fetching the document first and then updating with the patch fields.

AndrewSisley: > Partial document updates break some parts of our indexing system.

Thats not good, and a regression IMO.  No worries though RE this PR, thanks for giving it a go :)

fredcarle: @AndrewSisley That's exactly what `UpdateWithDocID` did so the pattern is the same. No regression in that sense.

AndrewSisley: The regression was introduced with sec. indexes, I'm pretty sure that up until that point we could do partial updates via any public func, so long as the docID was provided.

PR #2535, opened 2024-04-19
test(i): Assert all expected props are returned in tests

### LINKED ISSUE #2435: Integration test suite does not test that all expected query result properties are returned
I recently forgot to add a field to a query, but expected the field in the results and the test still passed. For example:
```
testUtils.Request{
    Request: `query {
	    Users {
		    name
	    }
    }`,
    Results: []map[string]any{
	    {
		    "name":     "John",
		    "verified": true,
	    },
    },
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #2435

## Description

Asserts that all expected properties are returned in integration test queries.

Is quite bad that this was missed for so long, luckily none of the bad tests passing because of it failed when their queries were corrected :sweat_smile:

### DIFF
--- a/tests/integration/backup/one_to_one/import_test.go
+++ b/tests/integration/backup/one_to_one/import_test.go
@@ -237,7 +237,8 @@ func TestBackupImport_DoubleRelationshipWithUpdate_NoError(t *testing.T) {
 						},
 					},
 					{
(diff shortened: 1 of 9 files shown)


fredcarle (approved):
Thanks for fixing this Andy :)

PR #2555, opened 2024-04-25
refactor: Replace connor package with rust implementation

### LINKED ISSUE #2508: Convert a Go package into Rust
As a part of endeavour of transitioning to Rust we want to take a sample Go package, convert it to Rust and replace it's usage in Go code. This is the first exploratory task that's why we don't want to take the simplest package. We want to see where challenges might arise early on.

We decided to proceed with `connor` package which is somewhat small, but reasonably complex.

Here is the list of tasks that needs to be done
- [x] Convert `connor` to Rust
- [x] Define ABI boundaries
- [x] Replace existing `connor` code in Go with a layers that conducts communication between boudaries

After this the old `connor` package in Go won't exist. We don't want to maintain 2 packages.

islamaliev:
## Relevant issue(s)

Resolves #2508

## Description

Replaces go implementation of connor package with rust implementation. All the input into connor package is serialized into JSON format understandable by rust implementation and passed across ABI boundary.

JSON was chosen just to keep things as simple as possible. We are not going keep the current state long (hopefully).

Note: We shouldn't merge this PR unless we are ready to allocate some human resources into transitioning DefraDB to Rust.

## Tasks

- MacOS


AndrewSisley:
question: The CI is failing, are you wanting feedback before getting that to work?

jsimnz:
Only done an quick look, nothing immediate code wise. But can you create a follow-up issue to change the `null` behavior for Numeric comparison on the Go connor implementation too. So that it matches the new behavior defined by the udpated tests in this PR?

Makefile
```diff
@@ -32,6 +32,50 @@ else ifeq ($(OS_GENERAL),Windows)
 	OS_PACKAGE_MANAGER := choco
 endif
 
+# Detect the operating system
+UNAME_S := $(shell uname -s)
```
shahzadlone, line 36: todo: We already have this I believe in `OS_GENERAL`

Makefile
```diff
@@ -32,6 +32,50 @@ else ifeq ($(OS_GENERAL),Windows)
...
+
+# Directory where binaries will be placed
+BIN_DIR := ./libs
+
+# Set the library file name based on the operating system
+ifeq ($(UNAME_S),Linux)
```
shahzadlone, line 42: todo: this and all others.

```suggestion
ifeq ($(OS_GENERAL),Linux)
```

connor/ge.go
```diff
@@ -11,8 +11,7 @@ import (
 // value is strictly larger than or equal to another.
 func ge(condition, data any) (bool, error) {
 	if condition == nil {
-		// Everything is greater than or equal to nil
-		return true, nil
+		return false, nil
```
shahzadlone, line 14: question: What caused this change?

libs/.gitkeep
shahzadlone, line 1: question: Why not just `.keep`

islamaliev: can't answer your question because no real reason. I don't see any difference and can change to `.keep`. But I think `.gitkeep` is a bit more explicit.

Makefile
islamaliev: changed

Makefile
islamaliev: changed

islamaliev:
> Only done an quick look, nothing immediate code wise. But can you create a follow-up issue to change the `null` behavior for Numeric comparison on the Go connor implementation too. So that it matches the new behavior defined by the udpated tests in this PR?

I updated Go behaviour to match new expectations.

libs/.gitkeep
AndrewSisley: `.gitkeep` is the norm in the orgs I've worked worked with

connor/ge.go
AndrewSisley: I dont think this change should be made in this PR.  Besides the lack of transparency in the change log, and cleaner PRs, it is also a good way of making sure that the `ge` change doesnt get merged for a few months.

islamaliev: The rust implementation already behaves this way. I don't think it's a good idea to keep different behaviours. Otherwise you won't be able to switch between them without breaking existing tests.
And the idea behind it was anyway to get rid of golang version once we have Rust substitute because we don't want to maintain 2 implementations in parallel.

AndrewSisley: > The rust implementation already behaves this way.

The rust implementation is not used.  Change the Go one, then the Rust one will pair up with it.

PR #2557, opened 2024-04-26
feat: Keyring

### LINKED ISSUE #2556: CLI only keyring
Add a keyring to help with the management and encryption of local user keys. This should not be used anywhere outside of the CLI.

nasdf:
## Relevant issue(s)

Resolves #2556 

## Description

This PR adds a keyring to the cli.

Notable changes:
- created `keyring` package
  - file backed keystore uses a password based key generation to encrypt
  - system backed keystore uses the OS managed keyring 
- added `keyring` root command
  - `--no-keyring` disables the keyring and generates ephemeral keys
  - `--keyring-backend` flag sets the keyring backend (file or system)
  - `--keyring-path` flag sets the keyring directory when using file backend
  - `--keyring-namespace` flag sets the service name when using system backend
- added `keyring generate` generates required keys
  - `--no-encryption-key` skips generating encryption key (disables datastore encryption) 
- added `keyring import` import external keys  (hexadecimal for now)
- added `keyring export` export existing key (hexadecimal for now)
- moved key generation to `crypto` package

## Tasks

## How has this been tested?

Manually tested on MacOS.

- MacOS


cli/keyring.go
```diff
@@ -0,0 +1,127 @@
...
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"crypto/rand"
+	"errors"
+	"syscall"
+
+	"github.com/99designs/keyring"
```
AndrewSisley, line 18: question: How confident are we in this dependency?  The repo hasn't been updated in more than two years, and it is managed by a company that looks to have a focus way away from the security world and may lack the technical expertise or attention to keep it secure (assuming it and its dependencies were 2 years ago).

99Designs:
> Global creative platform for designers and clients

@jsimnz @nasdf /All

cli/keyring.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return keyring.Open(keyring.Config{
+		AllowedBackends:         allowedBackends,
+		ServiceName:             "defradb",
+		KeychainName:            "defradb",
+		KeychainPasswordFunc:    prompt,
+		FilePasswordFunc:        prompt,
+		FileDir:                 dir,
+		KeyCtlScope:             "user",
+		KeyCtlPerm:              0, // TODO
```
islamaliev, line 53: todo: Please create a task if this is something that needs to be handled later. Otherwise it will linger.

nasdf: I will fix before merging.

cli/keyring.go
islamaliev, line 1: suggestion: I wonder if all this logic with generating hashes should be scoped to this cli package.

Maybe it makes sense to extract it in another place so other that other packages can use it.

node/store.go
```diff
@@ -74,5 +74,12 @@ func NewStore(opts ...StoreOpt) (datastore.RootStore, error) {
 	badgerOpts.ValueLogFileSize = options.valueLogFileSize
 	badgerOpts.EncryptionKey = options.encryptionKey
 
+	if len(options.encryptionKey) > 0 {
```
islamaliev, line 77: suggestion: I prefer `if options.encryptionKey != ""` as it clearly shows that the type of the field is string, not a slice

cli/keyring.go
nasdf: I've replaced this dependency with a custom implementation using a newer library.

node/store.go
nasdf: I think you may have misread the type. `encryptionKey` is a `[]byte`

cli/keyring.go
nasdf: I think that would be nice. Would it make sense to have a `crypto` package?

cli/keyring.go
nasdf: I ended up replacing this dependency.

cli/keyring.go
AndrewSisley: Dependencies and code of the new one look much healthier, thanks Keenan :)

cli/start.go
```diff
@@ -84,23 +83,32 @@ func MakeStartCommand() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				netOpts = append(netOpts, net.WithPrivateKey(peerKey))
+
+				encryptionKey, err := loadOrGenerateAES256(keyring, badgerEncryptionKeyName)
+				if err != nil {
+					return err
+				}
+				storeOpts = append(storeOpts, node.WithEncryptionKey(encryptionKey))
```
jsimnz, line 109: question: This implies that if the keyring is enabled, then we also use use at-rest encryption. Should these technically be separate config options?

keyring/file.go
```diff
@@ -0,0 +1,67 @@
...
+	dir string
+	// key is the keystore encryption key
+	key []byte
+}
+
+func openFileKeyring(dir string, password []byte) (*fileKeyring, error) {
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, err
+	}
+	key := pbkdf2.Key(password, []byte("defradb"), 4096, 32, sha1.New)
```
jsimnz, line 38: todo: Shouldn't be using a static salt here. I believe if you use the `jwa.PBES2_HS512_A256KW` instead of `jwa.A256KW` here then the `Encrypt` and `Decrypt` calls will handle the KDF and salt for you, and the salt then gets included in the JWE headers.

jsimnz (changes requested):
Heading out, but wanted to get this TODO on your radar before this got merged. May have more comments soon.

keyring/file.go
jsimnz: then the `key` is just the `password` defined by the user, and you don't have to do the KDF manually.

cli/start.go
nasdf: Good catch! I've added a `datastore.encryptionDisabled` config item.

keyring/file.go
jsimnz: and the added benefit that it will use `sha256` instead of `sha1`

nasdf: I was hoping there was something like this in the library. I've replaced it as suggested.

keyring/system.go
```diff
@@ -0,0 +1,49 @@
...
+// built in key management system of the OS.
+type systemKeyring struct{}
+
+func openSystemKeyring() *systemKeyring {
+	return &systemKeyring{}
+}
+
+func (systemKeyring) Set(name string, key []byte) error {
+	enc := base64.StdEncoding.EncodeToString(key)
+	return keyring.Set(service, name, enc)
```
jsimnz, line 31: todo: Adding comment here, but it applies to all the system Get/Set calls.

By using a static service name here, then if you have two nodes running on the same machine, that both want to use a the system store, their keys will overlap.

Can either
A) Use a paramaterized service name via the config
B) Use the single service name as you have here, but add a namespace on the item names during the `Get/Set`, which is set via the config.
C) Anything else you think of ;)

jsimnz (changes requested):
one more, kk now im really heading out, will be AFK

keyring/system.go
nasdf: Decided to go with option A.

cli/config.go
```diff
@@ -75,6 +75,7 @@ func defaultConfig() *viper.Viper {
 
 	cfg.SetDefault("datastore.badger.path", "data")
 	cfg.SetDefault("datastore.encryptionDisabled", false)
+	cfg.SetDefault("keyring.service", "defradb")
```
jsimnz, line 78: suggestion: "namespace" might be a more approachable/understandable name from an external UX POV

.github/workflows/start-binary.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Attempt to start binary
         run: |
-          ./build/defradb start &
+          ./build/defradb start --no-keyring &
```
shahzadlone, line 51: question: Why do we need to start with this flag?

cli/root.go
```diff
@@ -139,5 +139,17 @@ Start a DefraDB node, interact with a local or remote node, and much more.
...
 
+	cmd.PersistentFlags().String(
+		"keyring-path",
+		"keys",
+		"Path to store encrypted keys",
+	)
+
+	cmd.PersistentFlags().Bool(
+		"no-keyring",
+		false,
```
shahzadlone, line 150: question: By default we want the keyring?

question: For my understanding, what happens to the flow for using keys when there is no keyring? Does it make sense to make weather keyring is disabled or active based on weather acp is on or off?

nasdf: I think enabled by default is a sensible option. The keyring also stores the peer key and encryption key. If you disable the keyring the peer key will be ephemeral and encryption disabled. I don't think the keyring should be dependent on ACP because of the other keys.

.github/workflows/start-binary.yml
shahzadlone: todo: Might be worth adding instructions for this extra configuration step in the root `README.md` (perhaps in the configuration section?) where as that is the default, and mention this flag to skip the keyring configuration.

cli/root.go
shahzadlone: Makes sense, would be nice to document this with my other `todo` in the README.md

cli/start.go
nasdf: Update: I've made the key optional instead of the config option.

cli/keyring_generate.go
```diff
@@ -0,0 +1,62 @@
...
+WARNING: This will overwrite existing keys in the keyring.
+
+Example:
+  defradb keyring generate
+
+Example: with no encryption key
+  defradb keyring generate --no-encryption-key
+
+Example: with system keyring
+  defradb keyring generate --backend system`,
```
fredcarle, line 36: todo: This example is wrong `Error: unknown flag: --backend`.

cli/keyring_export.go
```diff
@@ -0,0 +1,41 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func MakeKeyringExportCommand() *cobra.Command {
```
AndrewSisley, line 17: question: What are your plans for testing all of this? Are you looking to just rely on manual testing?

keyring/keyring.go
```diff
@@ -0,0 +1,21 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package keyring
+
+// Keyring provides a simple set/get interface for a keyring service.
+type Keyring interface {
+	// Set stores the given key in the keystore under the given name.
```
AndrewSisley, line 15: suggestion: I think it is worth explicitly defining what should happen when calling `Set` and a key of that name already exists, in the documentation on this interface.

suggestion: I think it is worth explicitly defining what should happen when calling `Get` and `Delete` and a name of a key that doesn't exist, in the documentation on this interface.

cli/keyring_generate.go
fredcarle: I think it's the global flag `--keyring-backend` that is correct here.

cli/keyring_export.go
AndrewSisley: Wouldn't need to mock to test it I don't think - tests can handle things like password prompts without to much hassle.

I'm a bit nervous about not pushing for this (either here or in a follow up PR), as apart from the openAPI stuff it would be the only thing in the whole repo that really requires manual testing.

cli/keyring_generate.go
```diff
@@ -0,0 +1,62 @@
...
+				}
+			}
+			peerKey, err := crypto.GenerateEd25519()
+			if err != nil {
+				return err
+			}
+			return keyring.Set(peerKeyName, peerKey)
+		},
+	}
+	cmd.Flags().BoolVar(&noEncryption, "no-encryption-key", false, "Skip generating an encryption key")
```
AndrewSisley, line 60: suggestion: I think it might be worth highlighting that this would disable doc-encryption in the description text.

cli/keyring_generate.go
nasdf: good catch!

cli/keyring_export.go
nasdf: I'll work on adding some tests. I think I have an idea on how to avoid the user input.

tests/integration/utils2.go
```diff
@@ -22,14 +22,14 @@ import (
 
 	"github.com/bxcodec/faker/support/slice"
```
AndrewSisley, line 23: todo: Can you please add some automated testing of encryption at rest?  Perhaps a new test environment variable, and a CI job that runs with it enabled.?

cli/root.go
```diff
@@ -139,5 +139,29 @@ Start a DefraDB node, interact with a local or remote node, and much more.
...
+	cmd.PersistentFlags().String(
+		"keyring-path",
+		"keys",
+		"Path to store encrypted keys when using the file backend",
+	)
+
+	cmd.PersistentFlags().Bool(
+		"no-keyring",
+		false,
+		"Disable the keyring and generate ephemeral keys",
```
AndrewSisley, line 163: question: Would this disable encryption at rest? If so, I suggest noting that in this description.

question: What happens if keyring commands are called if the keyring is disabled?

question: Is this flag available for all commands?  Not just `start`/`init`?  If so what happens if you disable it whilst/after it has been in use?

fredcarle: This is what happens when you use `--no-keyring` after it has been in use:
```
defradb start --no-keyring
2024-05-07T14:44:30.085-0400	ERROR	badger	v4@v4.2.1-0.20231113215945-a63444ca5276/db.go:263	Received err: Encryption key mismatch. Cleaning up...
Error: Encryption key mismatch
```

fredcarle: > question: What happens if keyring commands are called if the keyring is disabled?

`--no-keyring` is unused for keyring commands. It won't affect the outcome.

AndrewSisley: Calling it `start` is of less a worry to me than if someone uses it in an active instance, for example:
```
defradb client collection get --name User bae-123 --no-keyring
// or
defradb client collection create --name User '{ "name": "Bob" }'  --no-keyring
```

nasdf: > question: Would this disable encryption at rest? If so, I suggest noting that in this description.

Yes, I'll update the description.

> question: What happens if keyring commands are called if the keyring is disabled?

You can still use the keyring commands to manage keys. 

> question: Is this flag available for all commands? Not just start/init? If so what happens if you disable it whilst/after it has been in use?

This currently only applies to the start command. I'm planning on moving the flags that only apply to the start command in a future PR. 

If you disable the keyring after using it, the peer key will be randomly generated, and at rest encryption will be disabled. 

If at rest encryption was previously enabled, the datastore will fail to open due to the encryption key not matching.

AndrewSisley: > --no-keyring is unused for keyring commands. It won't affect the outcome.

It is a no-op?  Might it be worth excluding it as a flag from those commands then?

AndrewSisley: > This currently only applies to the start command. I'm planning on moving the flags that only apply to the start command in a future PR.

Okay nice, I'm happy with that - thanks Keenan :)

AndrewSisley (approved):
Thanks Keenan - code is really readable, and I feel like we are getting good value per line out of this :)

Thanks for answering my questions :)

I have a couple of todos around testing, but perhaps we can probably get away with them being done in follow-up PRs if not here (and if no push back :)).  If we want them to be done in new PRs, please create tickets for them before merge.

cli/keyring_export.go
nasdf: I've added unit tests for all keyring commands.

tests/integration/utils2.go
nasdf: Added a new variant for encrypted datastore to the test workflow

Issue #2566, opened 2024-04-29
Float64 PnCounter does not always eventually reach consistency when overflowing


AndrewSisley:
More info to come :)

PR #2567, opened 2024-04-29
test(i): Fix PnCounter overflow tests

### LINKED ISSUE #2565: PnCounter test changes
Ticket for test changes made whilst bug bashing PnCounter

AndrewSisley:
## Relevant issue(s)

Resolves #2565

## Description

Fixes a faulty PnCounter overflow test which was doing a no-op whilst pretending to increment.

Adds a test for negative float64 overflows, and two tests documenting the newly found https://github.com/sourcenetwork/defradb/issues/2566 in a new `issues` directory.

I don't think I see #2566 as a release blocker, but no strong feelings there yet.

There is more info in the commit messages RE each change.

## Todo

- [x] Tests are flaky in the CI
- [x] http client does not like the update mutations

### DIFF
--- a/tests/integration/issues/2566_test.go
+++ b/tests/integration/issues/2566_test.go
@@ -8,7 +8,7 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-package peer_test
+package issues
 
 import (
 	"fmt"
--- a/tests/integration/issues/2566_test.go
+++ b/tests/integration/issues/2566_test.go
@@ -23,6 +23,13 @@ import (
 // This test documents https://github.com/sourcenetwork/defradb/issues/2566
 func TestP2PUpdate_WithPNCounterSimultaneousOverflowIncrement_DoesNotReachConsitency(t *testing.T) {
 	test := testUtils.TestCase{
+		SupportedClientTypes: immutable.Some(
+			[]testUtils.ClientType{
+				// This test only supports the Go client at the moment due to
+				// https://github.com/sourcenetwork/defradb/issues/2569
+				testUtils.GoClientType,
+			},
+		),
 		Actions: []any{
 			testUtils.RandomNetworkingConfig(),
 			testUtils.RandomNetworkingConfig(),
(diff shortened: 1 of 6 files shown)


tests/integration/issues/2566_test.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package peer_test
```
nasdf, line 11: thought: unsure if this makes sense as the package name

tests/integration/issues/README.md
nasdf, line 1: praise: this is a nice convention and organizational change

nasdf (approved):
Looks good! I'm curious how or if we want to fix this issue.

tests/integration/issues/2566_test.go
AndrewSisley: My bad, I copy-pasted from somewhere else and forgot to change this, renaming

- [ ] Rename package

fredcarle:
I like the issue tracking tests. It's more work than adding a comment pointing to an issue where the test should ultimately be but I'm open to it. We should discuss it in the standup.

As for the test that is affected here, I don't really see the benefit of the change. It's a different test.
> Adding 1000 to float.MaxValue isn't going to change the value, you could add 1000 a trillion times and it would never change the value.  Note, the behaviour is not what was previously documented.

The behavior is exactly what was documented. I would just add the extra tests instead to show what the behavior is when in that specific case.

AndrewSisley:
> The behavior is exactly what was documented. I would just add the extra tests instead to show what the behavior is when in that specific case.

I don't think we need to test the adding of an insignificant value to a float - that is pretty universal computing behaviour.   And if we really want a test for that it should not use float.MaxValue as the base, and have 'overflow' in the test name and description because that is massively misleading - there is no overflow occurring.

shahzadlone (approved):
LGTM

PR #2573, opened 2024-05-01
fix: Add check to filter result for logical ops

### LINKED ISSUE #2571: Foreign object filter part of an `_or` set of filters doesn't seem to do anything
With a query of the form:
```graphql
query {
  Book (filter: {_or: [
    {title: {_ilike: "%Test3%"}},
    {author: {name: {_ilike: "%Test3%"}}},
    {subTitle: {_ilike: "%Test3%"}}
  ]}) {
    title
  }
}
```
The `{author: {name: {_ilike: "%Test3%"}}}` part of the filter doesn't seem to have any impact on the result. It seems like having this object filter inside an `_or` set doesn't work. The following does work however.
```graphql
query {
  Book (filter: {author: {name: {_ilike: "%Test3%"}}}) {
    title
  }
}
```

fredcarle:
## Relevant issue(s)

Resolves #2571

## Description

This PR adds a check post doc filtering in the fetcher to see if we need to rely on the external conditions instead to ensure a valid filtering operation. This is specially important when filtering with logical operators `_and` and `_or`.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/planner/filter/complex.go
+++ b/planner/filter/complex.go
@@ -17,7 +17,7 @@ import (
 
 // IsComplex returns true if the provided filter is complex.
 // A filter is considered complex if it contains a relation
-// object withing an _or or _not operator not necessarily being
+// object withing an _or, _and or _not operator not necessarily being
 // its direct  child.
 func IsComplex(filter *mapper.Filter) bool {
 	if filter == nil {
@@ -32,6 +32,7 @@ func isComplex(conditions any, seekRelation bool) bool {
 		for k, v := range typedCond {
 			if op, ok := k.(*mapper.Operator); ok {
 				if (op.Operation == request.FilterOpOr && len(v.([]any)) > 1) ||
+					(op.Operation == request.FilterOpAnd && len(v.([]any)) > 1) ||
 					op.Operation == request.FilterOpNot {
 					if isComplex(v, true) {
 						return true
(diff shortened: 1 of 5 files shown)


AndrewSisley (approved):
LGTM, thanks Fred!

planner/filter/complex.go
```diff
@@ -32,6 +32,7 @@ func isComplex(conditions any, seekRelation bool) bool {
 		for k, v := range typedCond {
 			if op, ok := k.(*mapper.Operator); ok {
 				if (op.Operation == request.FilterOpOr && len(v.([]any)) > 1) ||
```
nasdf, line 34: suggestion: I think this can be simplified a bit and avoid the potential panic from type casting.

```go
isOpComplex := false
switch op.Operation {
  case request.FilterOpOr:
    isOpComplex = isComplex(v, true)
  case request.FilterOpAnd:
    isOpComplex = isComplex(v, true)
  case request.FilterOpNot:
    isOpComplex = isComplex(v, true)
}
if isOpComplex {
  return true
}
continue
```

nasdf (approved):
LGTM. One minor suggestion.

shahzadlone (approved):
LGTM

planner/filter/complex.go
fredcarle: `v` can't be anything other than an `[]any` if it's an `_or` or `_and` operator and we don't want to classify them as complex if there is only one element in the array.

nasdf: I don't think this changes the logic though. It just adds another recursive step.

fredcarle: made a change in line with what you suggested. Let me know what you think.

Issue #2578, opened 2024-05-02
Secondary Indexes: many side of one-many joins


AndrewSisley:
This does not currently work and is documented by a handful of tests, e.g. `TestQueryWithIndexOnOneToMany_IfFilterOnIndexedRelation_ShouldFilterWithExplain`

Consider achieving this with a rewrite of type_join.go: https://github.com/sourcenetwork/defradb/issues/2577

PR #2579, opened 2024-05-02
fix: Return correct results from one-many indexed filter

### LINKED ISSUE #2574: `TestQueryWithIndexOnOneToTwoRelation_IfFilterOnIndexedRelation_ShouldFilter` is incorrect
It looks at first glance to be a good test, but it actually shows a one-many being yielded as a one-one (`devices` in the second request is a one-many and should be of type `[]map[string]any`).

Removing the index from the Device.model field results in the correct result.

Possibly related to https://github.com/sourcenetwork/defradb/issues/2577

AndrewSisley:
## Relevant issue(s)

Resolves #2574

## Description

Return correct results from one-many joins filtered on an indexed field on the many side.

The invertibleTypeJoin doesn't support the inversion of one-many joins and was converting them into one-one joins.

This means that the index is no longer used if targeting the many side: https://github.com/sourcenetwork/defradb/issues/2578 (this support can be added after the release).

### DIFF
--- a/planner/type_join.go
+++ b/planner/type_join.go
@@ -279,6 +279,7 @@ func (p *Planner) makeTypeJoinOne(
 			root:                source,
 			subType:             selectPlan,
 			subSelect:           subType,
+			subSelectFieldDef:   subTypeFieldDesc,
 			rootName:            immutable.Some(subTypeField.Name),
 			subTypeName:         subType.Name,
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
LGTM! Thanks for fixing this and the added tests.

Issue #2592, opened 2024-05-06
Add automatic secondary index testing to CI matrix


AndrewSisley:
We've had quite a lot of bugs recently where using an index returns different results compared to if the index did not exist at all.  It should be quite easy to automatically test this.

Add some magic to the test suite so that (given an env var) indexes are added to all fields in all collections before executing requests and add a job to the CI to run it in every PR.

islamaliev:
Though this will certainly increase overall robustness, but also significantly increase time to run all the tests.
I think we should anyway explicitly cover all index-related scenarios to prof not only the correctness, but also the effectiveness of sec. indexes.

AndrewSisley:
> Though this will certainly increase overall robustness, but also significantly increase time to run all the tests.

It really shouldn't, it will just be another item in the CI matrix and take roughly the same time as running the basic embedded Go `make test` command in the CI. 

> I think we should anyway explicitly cover all index-related scenarios to prof not only the correctness

To do this requires duplicating a massive amount of tests, and because doing that is unpleasant and time consuming it wasn't done, and wont be done, and even if it was done right now it wont be maintained properly to cover future cases.  And because of all of that, we have had a large number of bugs where adding an index causes queries to return the wrong results.

Query on an indexed field should always return the same results as if the index was not there at all, and it is really really easy for us to not need to multiply the number of manually written integration tests to enforce that.

islamaliev:
> Query on an indexed field should always return the same results as if the index was not there at all, and it is really really easy for us to not need to multiply the number of manually written integration tests to enforce that.

These tests will certainly make sure the indexes produce correct results (the same results as without indexes), but we still want to have dedicated tests to make sure indexes are being properly utilized.

It's not an argument against your proposal, it's just something to keep in mind.

AndrewSisley:
> These tests will certainly make sure the indexes produce correct results (the same results as without indexes), but we still want to have dedicated tests to make sure indexes are being properly utilized.

Ah got it, yeah I think we are aligned then :)

PR #2594, opened 2024-05-06
fix: Remove limit for fetching secondary docs

### LINKED ISSUE #2590: Sec. Index: query with filter on an indexed field of a primary subtype always returns only 1 doc.
schema:
```graphql
type User {
  name: String
  age: Int @index
  devices: [Device]
} 

type Device {
  model: String 
  owner: User
}
```
Query:
```graphql
query {
  Device(filter: {
    owner: {age: {_eq: 48}}
  }) {
    model
    owner {
      name
    }
  }
}
```
Here if a user with age 48 has more than 1 device, all of them should be returned with the same owner. At the moment only 1 device is returned.

Detected during testing of #2365

islamaliev:
## Relevant issue(s)

Resolves #2590 and #2600

## Description

Make join fetch all secondary docs of a fetched-by-index primary doc.

### DIFF
--- a/tests/integration/index/docs.go
+++ b/tests/integration/index/docs.go
@@ -216,7 +216,7 @@ func getUserDocs() predefined.DocsList {
 					},
 					{
 						"model": "Playstation 5",
-						"year":  2022,
+						"year":  2021,
 						"type":  "game_console",
 						"specs": map[string]any{
 							"CPU":     3.5,
--- a/tests/integration/index/query_with_relation_filter_test.go
+++ b/tests/integration/index/query_with_relation_filter_test.go
@@ -555,6 +555,8 @@ func TestQueryWithIndexOnOneToOne_IfFilterOnIndexedRelation_ShouldFilter(t *test
 }
 
 func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedField_ShouldFilterWithExplain(t *testing.T) {
+	// This query will fetch first a matching device which is secondary doc and therefore
+	// has a reference to the primary User doc.
 	req := `query {
 		Device(filter: {
 			year: {_eq: 2021}
@@ -619,9 +621,13 @@ func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedField_ShouldFilterWithExplai
 }
 
 func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain(t *testing.T) {
+	// This query will fetch first a matching user (owner) which is primary doc and therefore
+	// has no direct reference to secondary Device docs.
+	// At the moment the db has to make a full scan of the Device docs to find the matching ones.
+	// Keenan has 3 devices.
 	req := `query {
 		Device(filter: {
-			owner: {age: {_eq: 48}}
+			owner: {name: {_eq: "Keenan"}}
 		}) {
 			model
 		}
@@ -632,8 +638,7 @@ func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExp
 			testUtils.SchemaUpdate{
 				Schema: `
 					type User {
-						name: String
-						age: Int @index
+						name: String @index
 						devices: [Device]
 					} 
 
@@ -649,8 +654,8 @@ func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExp
 			testUtils.Request{
 				Request: req,
 				Results: []map[string]any{
-					{"model": "iPad Mini"},
 					{"model": "iPhone 13"},
+					{"model": "iPad Mini"},
 					{"model": "MacBook Pro"},
 				},
 			},
--- a/tests/integration/index/query_with_relation_filter_test.go
+++ b/tests/integration/index/query_with_relation_filter_test.go
@@ -265,7 +265,7 @@ func TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelatio
 			},
 			testUtils.Request{
 				Request:  makeExplainQuery(req2),
-				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(15).WithIndexFetches(3),
+				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(33).WithIndexFetches(3),
 			},
 		},
 	}
@@ -547,6 +547,116 @@ func TestQueryWithIndexOnOneToOne_IfFilterOnIndexedRelation_ShouldFilter(t *test
 			testUtils.Request{
 				Request:  makeExplainQuery(req),
 				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(2).WithIndexFetches(1),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedField_ShouldFilterWithExplain(t *testing.T) {
+	req := `query {
+		Device(filter: {
+			year: {_eq: 2021}
+		}) {
+			model
+			owner {
+				name
+			}
+		}
+	}`
+	test := testUtils.TestCase{
+		Description: "With filter on indexed field of secondary relation (N-1) should fetch secondary and primary objects",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: `
+					type User {
+						name: String
+						devices: [Device]
+					} 
+
+					type Device {
+						model: String 
+						year: Int @index
+						owner: User
+					}
+				`,
+			},
+			testUtils.CreatePredefinedDocs{
+				Docs: getUserDocs(),
+			},
+			testUtils.Request{
+				Request: req,
+				Results: []map[string]any{
+					{
+						"model": "Playstation 5",
+						"owner": map[string]any{
+							"name": "Islam",
+						},
+					},
+					{
+						"model": "Playstation 5",
+						"owner": map[string]any{
+							"name": "Addo",
+						},
+					},
+					{
+						"model": "iPhone 10",
+						"owner": map[string]any{
+							"name": "Addo",
+						},
+					},
+				},
+			},
+			testUtils.Request{
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(9).WithIndexFetches(3),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain(t *testing.T) {
+	req := `query {
+		Device(filter: {
+			owner: {age: {_eq: 48}}
+		}) {
+			model
+		}
+	}`
+	test := testUtils.TestCase{
+		Description: "Upon querying secondary object with filter on indexed field of primary relation (in 1-N) should fetch all secondary objects of the same primary one",
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: `
+					type User {
+						name: String
+						age: Int @index
+						devices: [Device]
+					} 
+
+					type Device {
+						model: String 
+						owner: User
+					}
+				`,
+			},
+			testUtils.CreatePredefinedDocs{
+				Docs: getUserDocs(),
+			},
+			testUtils.Request{
+				Request: req,
+				Results: []map[string]any{
+					{"model": "iPad Mini"},
+					{"model": "iPhone 13"},
+					{"model": "MacBook Pro"},
+				},
+			},
+			testUtils.Request{
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(44).WithIndexFetches(1),
 			},
 		},
 	}
--- a/tests/integration/index/query_with_relation_filter_test.go
+++ b/tests/integration/index/query_with_relation_filter_test.go
@@ -252,7 +252,11 @@ func TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelatio
 				},
 			},
 			testUtils.Request{
-				Request:  makeExplainQuery(req1),
+				Request: makeExplainQuery(req1),
+				// we make 1 index fetch to get the only address with city == "London"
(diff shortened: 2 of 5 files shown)


fredcarle:
@islamaliev are we sure the removal of the limit is the way to go? It feels like it's a bug with passing the correct limit along.

islamaliev:
> @islamaliev are we sure the removal of the limit is the way to go? It feels like it's a bug with passing the correct limit along.

That particular limit was a way to differentiate between 1-to-1 and 1-to-many relations. All previous tests passed before because they happened to have (or filtered out all but) one related object, so the limit accidentally matched.
You can see from the tests how few tests I had to adjust to accommodate this change.
This needs to be solved anyway in a more intelligent way by indexing relations, which I wanted to discuss today.

tests/integration/index/docs.go
```diff
@@ -216,7 +216,7 @@ func getUserDocs() predefined.DocsList {
 					},
 					{
 						"model": "Playstation 5",
-						"year":  2022,
+						"year":  2021,
```
AndrewSisley, line 219: question: Why has this changed?

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -553,3 +553,113 @@ func TestQueryWithIndexOnOneToOne_IfFilterOnIndexedRelation_ShouldFilter(t *test
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedField_ShouldFilterWithExplain(t *testing.T) {
```
AndrewSisley, line 557: question: Was there a behaviour change here?  Before the production code changed what would the results have been?

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -553,3 +553,113 @@ func TestQueryWithIndexOnOneToOne_IfFilterOnIndexedRelation_ShouldFilter(t *test
...
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(9).WithIndexFetches(3),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain(t *testing.T) {
```
AndrewSisley, line 621: todo: This test is really painful to try and read, to try and figure out what has/hasn't been filtered out I am required to read (and hold in memory) nearly 500 lines worth of documents.  Can you please simplify this a little bit so that we can see what is going on.

AndrewSisley (changes requested):
Can you please clarify a few things regarding the public behaviour?  It is hard to understand at the moment.

tests/integration/index/docs.go
islamaliev: so that this query can fetch it:
```graphql
query {
  Device(filter: {
    year: {_eq: 2021}
  }) {
    model
    owner {
      name
    }
  }
}
```
And also to make it consistent with another "Playstation 5".
It's not used in other tests anyway.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: no, there was no change here. The test was passed in the beginning, but I realized that an explicit test was missing for this scenario.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: I'm not quite sure what you mean by "simplify". The whole test is around 40 lines of code.
I added some comments.

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: What has changed RE publicly visible production behaviour?

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: I cannot readily see the relationship between the (hidden) inputs, and the asserted outputs.  At the moment this test is not very valuable as documentation - it only asserts that the current (obscured) behaviour does not change, it does not describe what the current behaviour is (in a way that is readable to humans).

One of the bugs that was fixed last week was leaked due to a similar failing in the tests.  It was tested, however the test asserted the behaviour that was observed, not the behaviour that was desired - and because the test was impossible to read the bug was discovered by a partner instead of us during development.

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -265,7 +265,7 @@ func TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelatio
 			},
 			testUtils.Request{
 				Request:  makeExplainQuery(req2),
-				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(15).WithIndexFetches(3),
+				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(33).WithIndexFetches(3),
```
AndrewSisley, line 268: question: Did this change because of the production code change, or because of the Playstation 5 year change?

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: > The whole test is around 40 lines of code.

If you don't care about what data the test is querying, if you do, it is closer to 500 lines.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: it's because the production code changed. The planner now doesn't stop after finding first matching related secondary doc and continues until all sec. docs are exhausted.

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: My second question was not targeted at this particular test, I should have made that clearer.

You have answered that question for this test, and the existing explain one, but what about `TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain` - does that behave the same way before and after the production code change?

tests/integration/index/query_with_relation_filter_test.go
islamaliev: > At the moment this test is not very valuable as documentation

no tests are particularly valuable as documentation if they contain over 100 lines of code with most of them being copy&paste of one another.

> it does not describe what the current behaviour is (in a way that is readable to humans).

Are you sure you are referring to the right test? The test under question initially failed, which means it wasn't an observable behaviour. In order to make it green - so that it behaves in the desired way - I had to change production code.

> One of the bugs that was fixed last week was leaked due to a similar failing in the tests. It was tested, however the test asserted the behaviour that was observed, not the behaviour that was desired - and because the test was impossible to read the bug was discovered by a partner instead of us during development.

I'm not sure if you intentionally phrased it this way oversimplifying. I can give you a dozen of other factor thats contribute to it.

The test are possible to read. I would kindly ask you to write next time "it was had for me to read" so that we don't spend time arguing about subjective matter and try find a solution instead.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: no, it doesn't. I had to change production code to make it pass

AndrewSisley: What were the results before the production code changed?

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: > Are you sure you are referring to the right test?

This test, `TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain`.  It is hard to tell why the results that are returned, are returned, and why those devices not returned are not returned - because the documents under test are not visible.

It probably seems far more readable to you because you wrote it, and have a stronger memory of the 500 lines of documents under test.  Not everyone, including reviewers, and your future self will have that benefit.

The amount of extra clutter in the test documents also clutters debugging - having a lot of extra stuff (e.g. addresses) floating around in the system really does not help narrowing down the source of any future bugs.

> I can give you a dozen of other factor thats contribute to it.

Sure, and we should work on reducing those too, but that doesn't mean we cant also improve the tests.

> The test are possible to read.

For you, the author, and that's great, but the code needs to be read and maintained by others too.

islamaliev: I agree that it makes debugging less trivial. And I understand that it might be a bit annoying to have to jump to another place where all the docs are defined.
We discussed it already in a big round about the different styles that we prefer and there was obvious winner.
I'm willing to stick more to your preferred way, but for the indexes I chose this way because it's not enough to just have the documents that are involved in queries, but also a number of other documents in the db to make sure those are not involved, i.e. to see clear number in `execInfo`.

Anyway I think whatever we want to do should not be part of this PR.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: it returned only 1 related secondary object.

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: The test(s) in *this* PR are not particularly readable.   It don't particularly care which 'style' they are written, so long as they are as easy to read and maintain as they can be.  This means testing with a dataset containing only useful elements, and having both the input and outputs visible in the test.

I'm pushing back hard now, because I feel like we have been far too lax lately, and a lot of expensive developer time has been wasted recently by sub-optimal testing.

Having a personal style is fine, so long as it doesn't waste your colleagues' (and users') time.  We all need to be able to read and maintain the tests, not just you.

tests/integration/index/query_with_relation_filter_test.go
AndrewSisley: You mean primary object? A `Device`?

islamaliev: primary is `User`, `Device` is secondary.

AndrewSisley: It is not, device references user, not the other way around.  Not really important though :)

tests/integration/index/query_with_relation_filter_test.go
islamaliev: I told you about the reasons why "only useful elements" is not good enough for testing secondary indexes. You ignored this part.
This means besides "useful elements" there should be nearly the same amount of other not-so-useful document.

The test that you changed recently `TestQueryWithIndexOnOneToMany_IfFilterOnIndexedRelation_ShouldFilterWithExplain` uses only "useful elements" which doesn't really show the power of indexes as it touches all documents in the db the same way full scan does. I didn't have manage to review it before it was merge, otherwise I would point it out.

> The test(s) in this PR are not particularly readable. 

I'd like to get somebody's opinion on this. @fredcarle do you also feel this way? If yes, I will just change. Tired of discussing this.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: it's not important indeed, but I'd like to get the terminology straight.
We usually apply these terms to relations 'User.devices' or 'Device.owner'.
In these examples `User.devices` is secondary and `Device.owner` is primary, right? (I hope I'm not mixing it up)
For me it means is a relation is primary, it refers to primary document, as opposed to being a field of a primary document.

AndrewSisley: > In these examples User.devices is secondary and Device.owner is primary

Yes :) The side that contains the id of the other document is the primary side - it is perhaps easier to see in one-ones.

> For me it means is a relation is primary, it refers to primary document, as opposed to being a field of a primary document.

Ah I see how you are seeing it, that makes sense to me.  We are just talking about different things :)  The primary side of the relation is on `Device`, but the primary object could be seen as the `User`.

islamaliev: That means we agree that a self-sufficient object (meaning it has no references to another one) is considered primary. In our case `User`.
And the object that has a reference (as primary relation field) to another object is considered secondary. In this case `Device`.

tests/integration/index/query_with_relation_filter_test.go
islamaliev: > Do you think addresses and device specs are useful to this test? They are never queried. Most of the fields on the two collections used are not touched either. Anyone reading or debugging the test is forced to deal with them.

holy cow! Now I see why you are concerned so much about the setup. 
But this is not how generation of predefined documents work. All the extra collections and even fields of collections that are not defined in the schema (given to `SchemaUpdate`) are ignored. Please refer the the readme of the package https://github.com/sourcenetwork/defradb/tree/develop/tests/predefined

AndrewSisley: Hahahaha :grin: That does make things an awful lot better than I thought :grin: 

It doesn't really remove the readabilty problem though (and in some ways makes it worse, although the trade off is well worth it in order to trim the data down).

I think documentation on the test action needs to be expanded, that readme is not as useful as it could be if it were linked to from the action, with a brief 2 line summary (out of scope here, not asking for that in this PR :))

That's nice though, we can largely ignore my prior debugging concerns here :)

I think with https://github.com/sourcenetwork/defradb/issues/2592 the importance of the normal non-explain request would be greatly reduced, as the X results are returned would be covered by our normal tests, so perhaps we can focus on explaining the explain tests here.

suggstion: Could you perhaps add a comment explaining the `44` and the `1` explain request expected values? Both here and in the other 2 tests in this PR.  Given the size of `getUserDocs` and the magic of `CreatePredefinedDocs` it remains very hard to understand what those values should be and how they are what they are.

AndrewSisley: Created an issue for the action documentation here: https://github.com/sourcenetwork/defradb/issues/2600

AndrewSisley (approved):
Approving now, assuming https://github.com/sourcenetwork/defradb/issues/2592 gets picked up soon, and some comments are added for the explain asserts.

Thanks for the fix Islam, and explaining it all to me :)

tests/integration/index/query_with_relation_filter_test.go
islamaliev: added comments to tests and to predefined docs action

AndrewSisley: > added comments to tests and to predefined docs action

They look good, and thanks for sorting out #2600 here too!

PR #2596, opened 2024-05-06
docs(i): Import docs from doc repo

### LINKED ISSUE #2595: Move doc repo docs into defra repo
Copy https://github.com/sourcenetwork/docs.source.network/tree/master/docs/defradb into a single directory in the Defra repo (preserve the existing sub-directory structure).

John will sort out the docs repo magic at somepoint to pull these in to the docs website, don't worry about all that in this ticket.

AndrewSisley:
## Relevant issue(s)

Resolves #2595

## Description

Imports the docs from doc repo (https://github.com/sourcenetwork/docs.source.network/tree/master/docs/defradb).

The first commit imports the linked directory as-is - nothing has changed.

The second commit adds a very simple readme to highlight that they are used for the docs website - it can be expanded over time.

The commits after that import the 0.10.0 and 0.11.0 change logs into the directory, and tweak them to match the ones already in the repo.

Note: The other change logs (0.9 and earlier) include the below snip-it at the top of the change log files with different numerical values, I don't understand it and have not added it to 0.10.0 and 0.11.0 - I'm happy to do so in this PR if someone tells me what values it should have :)  Otherwise it can be done later.


nasdf (approved):
LGTM. We'll need to update the CLI doc generation so we don't have duplicates. We should also consider adding OpenAPI docs using the `genopenapi` command.

fredcarle (approved):
LGTM. Please add the metadata as described bellow before merge.

> Note: The other change logs (0.9 and earlier) include the below snip-it at the top of the change log files with different numerical values, I don't understand it and have not added it to 0.10.0 and 0.11.0 - I'm happy to do so in this PR if someone tells me what values it should have :) Otherwise it can be done later.

Since the software orders sidebar items alphanumerically by default, the sidebar position metadata at the top of the file helps enforce a given order. The higher the number, the higher in the list it will be listed. Without it, the list would be:
- v0.10.0
- v0.11.0
- v0.2.0
- v0.2.1
- ...
What we actually want is:
- v0.11.0
- v0.10.0
- v0.9.0
- ...
So the metadata needs to be added to v0.10.0 and v0.11.0 with values
`sidebar_position: 100` and `sidebar_position: 110` respectively.

AndrewSisley:
> So the metadata needs to be added to v0.10.0 and v0.11.0 with values
sidebar_position: 100 and sidebar_position: 110 respectively.

Nice, thanks Fred :)

- [ ] Add sidebar position

PR #2599, opened 2024-05-07
refactor: Move internal packages to internal dir

### LINKED ISSUE #2507: Move internal packages into `internal` directory
They should not be consumed by non-source code, and are not tested or documented as public functions.  We do not wish to support a 500 func long public API.  The fact that they are all public atm is a source of continuous confusion even amongst Source employees and wastes a lot of dev time.

IIRC only the following packages should be public:
- `client`
- `db`
- `datastore` (and children, excluding `mocks`)
- `events`

AndrewSisley:
## Relevant issue(s)

Resolves #2507

## Description

Moves our internal packages to an internal directory.

No code has been changed, only moved, and import paths have been updated.

Note, I would like to move most of  `db`'s child packages, however that is a little bit more fiddly as a lot of them have stuff imported by the shared top-level `internal` packages and I don't want to complicate this PR.

Playground is still a top level directory.  So is `tests`. `cli` and `cmd` have been made internal. `http` is still public.

That `cli` is internal and `http` is public sounds weird writing it down, but feels right in my gut.  Maybe because users should be free to spin up our http server themselves - please voice your thoughts here :)

## How has this been tested?

I ran a bunch of make commands locally on my Linux machine.  Plus the CI stuff


fredcarle:
@AndrewSisley I don't think `cmd` can be internal. It won't build if I recall.

fredcarle:
> Note, I would like to move most of db's child packages, however that is a little bit more fiddly as a lot of them have stuff imported by the shared top-level internal packages and I don't want to complicate this PR.

You don't want to move all of `db` into `internal`?

AndrewSisley:
> You don't want to move all of `db` into `internal`?

The constructor needs to be public, otherwise users can't instantiate an embedded Go client.  We could of course chose to wrap that function in a public package, but that's not the only option.

AndrewSisley:
> Edit: It can be internal but we need to update the build paths.

Ah I see them, I gave a quick scan of the make file, but obviously not a good enough one :)

- [ ] fix `cmd` paths in make file

islamaliev:
it feels really strange to have `cli` and `http` on different levels. Could it be that your gut misleading you this time? :)

AndrewSisley:
> it feels really strange to have `cli` and `http` on different levels. Could it be that your gut misleading you this time? :)

Do you have a preference?  Make `cli` public, or `http` internal?  Feels a bit strange to have `cli` public, although if `cmd` is kept internal, it would perhaps allow users to include defra commands within their own CLI (although maybe only if they use `cobra too?).

nasdf:
> The constructor needs to be public, otherwise users can't instantiate an embedded Go client. We could of course chose to wrap that function in a public package, but that's not the only option.

The `node` package currently does this.

jsimnz:
I don't think `cli` or `cmd` should be internal, these are common, public entry points, and with `cli` being public, anyone can integrate the entire defra cli into any other binary very easily. As for `cmd` again its a public entry point, and standard practise to have this in the root of the repo. Otherwise, when you install it would be `go install github.com/sourcenetwork/defradb/internal/cmd/defradb` which is not ideal.

fredcarle:
> I don't think `cli` or `cmd` should be internal, these are common, public entry points, and with `cli` being public, anyone can integrate the entire defra cli into any other binary very easily. As for `cmd` again its a public entry point, and standard practise to have this in the root of the repo. Otherwise, when you install it would be `go install github.com/sourcenetwork/defradb/internal/cmd/defradb` which is not ideal.

I'm of the same opinion.

AndrewSisley:
> I don't think cli or cmd should be internal,

Okay nice, I'll make them public then :)

- [ ] cli or cmd stay public

fredcarle (approved):
LGTM 👍

jsimnz:
wait why is the `db` package internal? Its also very much a public entry point, how else are ppl supposed to use the embedded API?

Issue #2431, opened 2024-03-20
Update Go version to 1.22


AndrewSisley:
When 1.23 is released.

There are a handful of todos that can be sorted out when this is done - Ctrl+f for this ticket link and try and resolve them.

PR #2604, opened 2024-05-09
refactor: Change from protobuf to cbor for IPLD

### LINKED ISSUE #1065: Write data models are not used on read
Stuff like `LWWRegDelta` is used on write, but not on read.  We should have a shared set of models so we more can safely parse persisted stuff and see where it is being used.  It should probably live in `core` or similar.

Most/all should have some separation between use and storage (i.e. don't move LWWRegDelta to core, define a new, shared struct there and map (like an entity)).

### LINKED ISSUE #2603: Change IPLD protobuf to cbor

fredcarle:
## Relevant issue(s)

Resolves #2603 

## Description

This PR changes the encoding of the blocks in the blockstore to use cbor instead of protobof. As part of this, we move away from encoding/decoding individual deltas (no more Marshalling and Unmarshalling deltas) and instead, encode/decode at the block level. To do this, we use the new `github.com/ipld/go-ipld-prime` package that uses IPLD schemas to map the encoded blocks to the appropriate fields and CRDT types. The new `core/block` package handles that part of the change.

The `net` package is also affected by that change with the `DAGSyncer` being replaced by the new `ipld/linking.LinkSystem`. I expect further usage of the `LinkSystem` in a couple subsequent PRs.

## Tasks

## How has this been tested?

make test

- MacOS


internal/core/block/block.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// New creates a new block with the given delta and links.
+func New(delta core.Delta, links []DAGLink, heads ...cid.Cid) *Block {
+	blockLinks := make([]DAGLink, 0, len(links)+len(heads))
+	for _, head := range heads {
+		blockLinks = append(
+			blockLinks,
+			DAGLink{
+				Name: "_head",
```
AndrewSisley, line 128: question: Was this renamed from `head` in this PR, or am I misreading the deleted code?

internal/db/collection_delete.go
```diff
@@ -148,15 +151,15 @@ func (c *collection) applyDelete(
 		return err
 	}
 
-	dagLinks := make([]core.DAGLink, len(cids))
+	dagLinks := make([]coreblock.DAGLink, len(cids))
```
AndrewSisley, line 154: thought: It feels like this is leaking abstractions/dependencies - it might be good to move this into the `merkle` pkg or similar at some point

internal/db/fetcher/versioned.go
```diff
@@ -307,33 +306,27 @@ func (vf *VersionedFetcher) seekNext(c cid.Cid, topParent bool) error {
 	}
 
 	// decode the block
-	nd, err := dag.DecodeProtobuf(blk.RawData())
+	block, err := coreblock.GetFromBytes(blk.RawData())
 	if err != nil {
 		return NewErrVFetcherFailedToDecodeNode(err)
 	}
 
-	// subDAGLinks := make([]cid.Cid, 0) // @todo: set slice size
```
AndrewSisley, line 315: praise: Thanks for removing this :)

internal/planner/commit.go
```diff
@@ -238,7 +242,8 @@ func (n *dagScanNode) Next() (bool, error) {
 		n.queuedCids = append(make([]*cid.Cid, len(heads)), n.queuedCids...)
 
 		for i, h := range heads {
-			n.queuedCids[len(heads)-i-1] = &h.Cid
+			link := h // TODO remove when Go 1.22
```
AndrewSisley, line 245: todo: Please link this todo to https://github.com/sourcenetwork/defradb/issues/2431

internal/planner/commit.go
```diff
@@ -289,46 +294,28 @@ which returns the current dag commit for the stored CRDT value.
 All the dagScanNode endpoints use similar structures
 */
 
-func (n *dagScanNode) dagBlockToNodeDoc(block blocks.Block) (core.Doc, []*ipld.Link, error) {
+func (n *dagScanNode) dagBlockToNodeDoc(block *coreblock.Block) (core.Doc, []cidlink.Link, error) {
```
AndrewSisley, line 297: praise: Much, much, nicer - thanks Fred :)

internal/planner/commit.go
```diff
@@ -345,22 +332,23 @@ func (n *dagScanNode) dagBlockToNodeDoc(block blocks.Block) (core.Doc, []*ipld.L
...
+		field, ok := cols[0].Definition().GetFieldByName(fName)
 		if !ok {
-			return core.Doc{}, nil, client.NewErrFieldNotExist(fieldName.(string))
+			return core.Doc{}, nil, client.NewErrFieldNotExist(fName)
 		}
 		fieldID = field.ID.String()
 	}
-
+	d := block.Delta.GetData()
+	if d != nil {
```
AndrewSisley, line 342: question: This if-else does not make any sense to me, why are you doing it?

net/peer.go
```diff
@@ -21,10 +21,10 @@ import (
 	"github.com/ipfs/boxo/bitswap/network"
 	"github.com/ipfs/boxo/blockservice"
 	exchange "github.com/ipfs/boxo/exchange"
-	dag "github.com/ipfs/boxo/ipld/merkledag"
+	"github.com/ipfs/boxo/fetcher"
```
AndrewSisley, line 24: suggestion: I think it might be worth aliasing this import with a comment explaining why, when I was first reading the consuming code I thought it was working with the `db/fetcher` package - `fetcher` has quite a strong meaning in our codebase.

Same in process.go too.

net/process.go
```diff
@@ -18,26 +18,26 @@ import (
...
 type blockProcessor struct {
 	*Peer
-	txn    datastore.Txn
-	col    client.Collection
-	dsKey  core.DataStoreKey
-	getter ipld.NodeGetter
+	txn     datastore.Txn
+	col     client.Collection
+	dsKey   core.DataStoreKey
+	fetcher fetcher.Fetcher
```
AndrewSisley, line 40: suggestion: Similar to my import alias suggestion, I suggest not calling this property `fetcher` - it has a very strong association with out database fetchers and confused me for a while when reading.

net/server.go
```diff
@@ -262,20 +262,17 @@ func (s *server) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.PushL
 		}
 
 		// Create a new DAG service with the current transaction
-		var getter format.NodeGetter = s.peer.newDAGSyncerTxn(txn)
-		if sessionMaker, ok := getter.(SessionDAGSyncer); ok {
-			getter = sessionMaker.Session(ctx)
-		}
+		fetcher := s.peer.newDAGSyncerTxn(txn)
```
AndrewSisley, line 265: suggestion: Same as other comments, I'd prefer it if we didnt call this `fetcher`

tests/integration/events/simple/with_update_test.go
```diff
@@ -66,14 +66,14 @@ func TestEventsSimpleWithUpdate(t *testing.T) {
 		ExpectedUpdates: []testUtils.ExpectedUpdate{
 			{
 				DocID: immutable.Some(docID1),
-				Cid:   immutable.Some("bafybeif757a4mdwimqwl24ujjnao6xlajiajz2hwuleopnptusuttri6zu"),
+				Cid:   immutable.Some("bafyreifvrmwmlwtglxe3afki36spu6d5qs6vvza57kxs4giyi53r5vbbnu"),
```
AndrewSisley, line 69: question: I skimmed very quickly over the tests, at a glance they all seemed to be cid changes only.  Are there any public behaviour changes in this PR?

AndrewSisley (approved):
Looks really good Fred, thanks :) Code is much nicer.  I only have a handful of small comments for you :)

internal/core/block/block.go
```diff
@@ -0,0 +1,342 @@
...
+				Link: cidlink.Link{Cid: head},
+			},
+		)
+	}
+
+	// make sure the links are sorted lexicographically by CID
```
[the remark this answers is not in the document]
fredcarle: I'm not sure the order is that important. We loop through the links in all cases without even a requirement for the head to be at index 0. I think it's just nice to have things nicely ordered.

internal/core/block/block.go
fredcarle: It was not renamed. I'm curious what makes you think that.

internal/core/block/block.go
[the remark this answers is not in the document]
fredcarle: Actually, I think this is done to make sure the block CID is deterministic. I will add documentation.

internal/db/collection_delete.go
fredcarle: Can you explain why you're saying this? I don't think the DAGLink should move to the `merkle` package.

tests/integration/events/simple/with_update_test.go
fredcarle: There are no public behaviour changes :)

internal/core/block/block.go
```diff
@@ -0,0 +1,342 @@
...
+
+// Schema is the IPLD schema type that represents a `Block`.
+var Schema schema.Type
+
+func init() {
+	Schema = mustSetSchema()
+}
+
+func mustSetSchema() schema.Type {
+	ts, err := ipld.LoadSchemaBytes([]byte(`
```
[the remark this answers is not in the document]
fredcarle: ✅

internal/core/block/block.go
```diff
@@ -0,0 +1,342 @@
...
+
+		type CounterDeltaInt struct {
+			docID     		Bytes
+			fieldName 		String
+			priority  		Int
+			nonce 			Int
+			schemaVersionID String
+			data            Int
+		}
+
```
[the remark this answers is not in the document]
fredcarle: ✅

internal/core/block/block.go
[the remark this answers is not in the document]
fredcarle: ✅

internal/core/block/block.go
```diff
@@ -0,0 +1,342 @@
...
+	switch delta := delta.(type) {
+	case *crdt.LWWRegDelta:
+		block.Delta = CRDT{LWWRegDelta: delta}
+	case *crdt.CompositeDAGDelta:
+		block.Delta = CRDT{CompositeDAGDelta: delta}
+	case *crdt.CounterDelta[int64]:
+		block.Delta = CRDT{CounterDeltaInt: delta}
+	case *crdt.CounterDelta[float64]:
+		block.Delta = CRDT{CounterDeltaFloat: delta}
```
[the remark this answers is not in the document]
fredcarle: ✅

internal/planner/commit.go
fredcarle: ✅

net/peer.go
fredcarle: ✅

net/process.go
fredcarle: ✅

net/server.go
fredcarle: ✅

jsimnz:
reviewing now

internal/core/block/block.go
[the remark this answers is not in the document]
AndrewSisley: Nice, cheers Fred :)

internal/core/block/block.go
AndrewSisley: ln 219 of the old internal/core/crdt/composite.go had a `"head"` string, that might have been a bug or dead code though, as everywhere else seems to say `"_head"`.  If you've copied that line somewhere (maybe to here, I have forgotten) - can you try and see if we can remove it?

internal/db/collection_delete.go
AndrewSisley: It feels out of place to have the `db` package mess around directly with DAGLinks, and it is only done here (and on save).  The way they are interacted with looks easy to move out of the package too

tests/integration/events/simple/with_update_test.go
AndrewSisley: Nice, cheers :)

AndrewSisley (approved):
LGTM, one tiny todo in the PR, and another possible small one messaged over discord :)

internal/core/block/block.go
fredcarle: The `DeltaDecode` method that you are referring to was only used in the versioned fetcher and if I think that if that line actually worked, it would have been a bug because the next function that was called was `ProcessNode` and it needed the `_head` links to work properly. Adding the `_head` links here cannot be removed.

AndrewSisley: Nice, thanks for the explanation :)

internal/db/collection_delete.go
fredcarle: Ah I see what you mean. I tried that before but I don't think it's possible to avoid it given that on save we need to keep track of the link names. It would be weird to build a different array only to convert it in the `merkleCRDT` package.

internal/core/block/block.go
```diff
@@ -0,0 +1,361 @@
...
+	// This will be either the field name of the CRDT delta or "_head" for the head link.
+	Name string
+	// Link is the CID link to the object.
+	cidlink.Link
+}
+
+// IPLDSchemaBytes returns the IPLD schema representation for the DAGLink.
+//
+// This needs to match the [DAGLink] struct or [mustSetSchema] will panic on init.
+func (l DAGLink) IPLDSchemaBytes() []byte {
```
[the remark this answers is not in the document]
fredcarle: This is already taken care of by a compile time check on line 37 of `block.go`. No need to worry about doing that for every type individually. It will automatically be taken care of for every new type that we add to `mustSetSchema`.

net/server_test.go
```diff
@@ -247,6 +250,27 @@ func TestDocQueue(t *testing.T) {
 	q.mu.Unlock()
 }
 
+func getHead(ctx context.Context, db client.DB, docID client.DocID) (cid.Cid, error) {
```
jsimnz, line 253: todo: We shouldn't have a duplicate `getHead` function, with its own implementation, since there is already all this functionality within the `merkle/clock` package to be able to get a specific head CID of a document. 

Its not the nicest to use, but if there is going to be some util function that takes a store and a docID and returns the head CID, it should be along side the current implementation in the other package.

jsimnz: Note: We also probably don't need the updated `Headstore` method on the `client.DB` interface, since the `datastore.Multistore` already has this method, and the entire `client.DB` interface shouldn't be a argument to this function just to get a head CID of a document.

jsimnz (changes requested):
overall, big fan of these changes, and even moreso knowing what your next steps are, especially with the `net` package.

Couple of things ive noted in this PR, nothing giant, but one or two notable items.

internal/core/block/block.go
```diff
@@ -0,0 +1,361 @@
...
+	// Links are the links to other blocks in the DAG.
+	Links []DAGLink
+}
+
+// IPLDSchemaBytes returns the IPLD schema representation for the block.
+//
+// This needs to match the [Block] struct or [mustSetSchema] will panic on init.
+func (b Block) IPLDSchemaBytes() []byte {
+	return []byte(`
+	type Block struct {
```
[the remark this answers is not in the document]
jsimnz: This one is worth a large discussion likely

internal/core/delta.go
```diff
@@ -10,27 +10,9 @@
 
 package core
 
-import (
-	cid "github.com/ipfs/go-cid"
-)
-
 // Delta represents a delta-state update to delta-CRDT.
 // They are serialized to and from Protobuf (or CBOR).
 type Delta interface {
```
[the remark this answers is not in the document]
AndrewSisley: I have a preference to have two different interfaces for this, `Delta`, and whatever `IPLDSchemaBytes` is defined on.  They represent two different concepts that happen to be implemented by the same concrete types atm.  If we really want them on one interface, that interface should embed `Delta` and `IPLDSchemaBytes`'s host-interface.

net/server_test.go
AndrewSisley: `net` package test code should not be referencing the `merkle/clock` package.  If I had to chose between duplicating code and referencing an internal method, I would overwhelmingly support duplicating the code.

internal/core/block/block.go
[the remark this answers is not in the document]
nasdf: It might be simpler for the version to be specified as a CID multicodec. I think its worth a larger discussion like you said.

internal/core/block/block.go
```diff
@@ -0,0 +1,361 @@
...
+	return cidlink.LinkPrototype{Prefix: cid.Prefix{
+		Version:  uint64(multicodec.Cidv1),
+		Codec:    uint64(multicodec.DagCbor),
+		MhType:   uint64(multicodec.Sha2_256),
+		MhLength: 32,
+	}}
+}
+
+// CRDT is a union type that can hold any of the CRDT deltas.
+type CRDT struct {
```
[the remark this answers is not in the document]
fredcarle: I might explore that.

internal/core/crdt/composite.go
```diff
@@ -41,13 +36,23 @@ type CompositeDAGDelta struct {
...
 
-var _ core.CompositeDelta = (*CompositeDAGDelta)(nil)
+var _ core.Delta = (*CompositeDAGDelta)(nil)
+
+// IPLDSchemaBytes returns the IPLD schema representation for the type.
+//
+// This needs to match the [CompositeDAGDelta] struct or [coreblock.mustSetSchema] will panic on init.
+func (delta *CompositeDAGDelta) IPLDSchemaBytes() []byte {
+	return []byte(`
+	type CompositeDAGDelta struct {
```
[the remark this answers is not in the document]
fredcarle: Since it's a small amount of code/text, I find it more convenient to have it right there as you read the Go code. I don't see the benefit of separating them into their own files.

internal/core/delta.go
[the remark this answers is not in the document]
fredcarle: I prefer to keep them separate. Like Andy said, they represent two different concepts.  It is much better in my opinion to keep interfaces as small as possible. We already suffer from bloated interfaces or at least using large interfaces too liberally.

internal/core/crdt/composite.go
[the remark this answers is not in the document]
AndrewSisley: If we were to split them into their own files we'd probably want to force a one-one file correspondence with the Go code too otherwise which `.proto` file goes with which `.go` file would probably be a bit confusing - and I'm not sure we'd want to force that on the Go code, and splitting the go files would be better done in another PR anyway (I've not checked to be sure, but I don't think they are one type per file atm)

internal/merkle/crdt/merklecrdt.go
```diff
@@ -34,31 +35,38 @@ type Stores interface {
 // so it can be merged with any given semantics.
 type MerkleCRDT interface {
 	core.ReplicatedData
-	Clock() core.MerkleClock
-	Save(ctx context.Context, data any) (ipld.Node, uint64, error)
+	Clock() MerkleClocker
+	Save(ctx context.Context, data any) (cidlink.Link, []byte, error)
+}
+
+// MerkleClocker is the logical clock implementation that manages writing to and from
```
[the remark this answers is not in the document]
fredcarle: I moved it from `core` to `merkle/crdt` and added the -er for convention since we also have a `MerkleClock` struct type. But since the struct lives in `merkle/clock`, there wouldn't be any conflict to also naming the interface `MerkleClock`.

internal/core/crdt/composite.go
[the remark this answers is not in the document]
fredcarle: They are one type per file :)

net/server_test.go
fredcarle: There are a few problems here. One is like Andy mentioned, needing to deal with `internal/merkle/clock` for `net` tests is not ideal. Also, before this PR, the `Headstore` was only available through `MultiStore` with an explicit `txn` which those test don't use. Saying the entire `client.DB` shouldn't be an argument to this function is correct and that is something I want to fix in a lot of places in our codebase. We often use large interfaces when a locally specific one would be more appropriate.

I'd rather not get into this within this PR but I will try to improve the `Headstore` utilities shortly.

fredcarle: Issue created: https://github.com/sourcenetwork/defradb/issues/2621

internal/core/block/block.go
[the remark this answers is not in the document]
fredcarle: Issue created: https://github.com/sourcenetwork/defradb/issues/2622

We can figure this out before the end of the cycle. Let's not block this PR for this.

internal/core/crdt/composite.go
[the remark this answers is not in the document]
jsimnz: Its just A) Somewhat ugly B) not well structured, and doesn't lend itself to working with all the IPLD schemas independent of the delta implementations

net/dag.go
```diff
@@ -108,8 +91,8 @@ func (p *Peer) dagWorker(jobs chan *dagJob) {
 		}
 
 		go func(j *dagJob) {
-			if j.bp.getter != nil && j.cid.Defined() {
-				cNode, err := j.bp.getter.Get(p.ctx, j.cid)
+			if j.bp.dagSyncer != nil && j.cid.Defined() {
+				nd, err := j.bp.dagSyncer.BlockOfType(p.ctx, cidlink.Link{Cid: j.cid}, coreblock.SchemaPrototype)
```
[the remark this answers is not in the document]
fredcarle: ✅

internal/merkle/crdt/merklecrdt.go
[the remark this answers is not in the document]
fredcarle: ✅

internal/core/crdt/composite.go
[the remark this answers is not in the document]
fredcarle: A) That is subjective. 
B.1) It's well structured in the sense that it is right next to the Go type that represents it. 
B.2) Why would we want to use them independently of the delta structs? They are merely a representation of the same thing. The IPLD schema is used to help translate from an IPLD node representation to the corresponding Go struct.

net/server_test.go
jsimnz: Usage within a test or not doesn't make much difference here. This function is reimplementing a somewhat sensitive flow and is relying on the behavior of getHead to be inline with the DB. 

If somewhat makes changes to the actual headset or head store for whatever reason, it's not necessarily going to be clear when they run the tests and the net package tests are failing. 

The fact that the merkleclock is internal or not also has no bearing.

We have a specific implementation, that is rather core to the entire correctness of the DB with the headset implementation, and that behavior must be consistent.

fredcarle: We agree that it needs to be consistant. This will be fixed in a subsequent PR.

internal/core/crdt/composite.go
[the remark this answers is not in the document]
jsimnz: A) absolutely 
B1) I meant not well structured as it's just some arbitrary inline byte string inside a method
B2) Now that we have dedicated IPLD schemas defined, it opens up a whole world of things we can do with them. Eg generate non Go serialization bindings, use the schemas in docs,  can have util functions that operate on the schemas themselves, etc.

The reason I made the proto analogy, is that much like protobuf it actually has nothing to do with Go. IPLD schemas are general and language independent, and we're losing out on some potential functionally to have them inline like this.

This isn't a PR blocker, just a general design point for this new concept being brought into the codebase with this PR

net/server_test.go
jsimnz: Yupp perfect

(Note my comment was primarily in response to that it's a test and it's ok to duplicate the functionality, not the need for it to be in the pr)

Cheers

internal/core/delta.go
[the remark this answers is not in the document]
jsimnz: The only problem here is that unless we add more compile time interface to checks to the deltas, we lose out on the compile time safety.

It's not the biggest issue, but this change now requires that deltas have a corresponding IPLD schema.

We do still preserve some compile time safety, since if you pass the delta into the mustSchemas func it'll complain. 

But there's no clear statement in the code that deltas much have IPLD schemas, and if it's a contract we are requiring, that contract should be expressed in the code, ideally with compile time enforcement

internal/core/block/block.go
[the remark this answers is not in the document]
fredcarle: ✅

jsimnz (approved):
Lgtm thanks for extracting the follow up items from my review

Issue #2366, opened 2024-03-04
Add support for P2P and ACP 


shahzadlone:
~~Implement (and integrate) Defra SourceHub ACP module.~~

There is a bunch of todos in our codebase and temp errors in our codebase preventing ACP from working with P2P.  Resolve them.

https://github.com/sourcenetwork/defradb/issues/2836 has been broken out of this ticket.

AndrewSisley:
Note: This is currently under discussion in the `dev-db.ACP` discord thread: https://discord.com/channels/427944769851752448/1200142876307374100

fredcarle:
<img width="694" alt="Screenshot 2024-04-17 at 12 13 06 PM" src="https://github.com/sourcenetwork/defradb/assets/4949751/22066791-db5e-4c52-bc25-aa31394dc876">

<img width="556" alt="Screenshot 2024-04-17 at 12 13 45 PM" src="https://github.com/sourcenetwork/defradb/assets/4949751/cece2a61-4f52-46e4-8601-99360cca14a5">

AndrewSisley:
note: Unassigning myself for the cooldown, I intend to pick this up again for v0.12, but anyone else is very welcome to steal it in the meantime, or if I am unavailable in 2 weeks.

AndrewSisley:
Bruno is currently working on a new client for SourceHub that will make things much nicer for us from Defra.  It will hopefully be implemented within about 3 weeks.  The new client stuff should not impact the spinning up of SourceHub test instances, so that can be worked on a bit before actually having to worry about writing the production code.  The interface will be defined before he implements it, but working from that too early is risky and may potentially be changed as work on the implementation progresses.

I plan on leaving this ticket for another week or so, before checking in with Bruno RE the status of the new client.

PR #2633, opened 2024-05-21
feat: Inject ACP instance into the DB instance

### LINKED ISSUE #2632: Pass `acp.ACP` into `db.NewDB`
Pass `acp.ACP` into `db.NewDB` like we do with rootstore.

AndrewSisley:
## Relevant issue(s)

Resolves #2632

## Description

Injects the ACP instance into the DB instance, like how rootstore is provided.

Also makes the `acp` package public, and moves the ACP options off of `db` and into `node` (similar to rootstore).

Prerequisite for https://github.com/sourcenetwork/defradb/issues/2366 (as well as being something I think we inherently wanted).


shahzadlone:
question: Is acp in mem and non-mem usage now dependent on the path config?

AndrewSisley:
> question: Is acp in mem and non-mem usage now dependent on the path config?

As far as I could see it was always so.

node/node.go
```diff
@@ -112,7 +121,7 @@ func NewNode(ctx context.Context, opts ...NodeOpt) (*Node, error) {
 	if err != nil {
 		return nil, err
 	}
-	db, err := db.NewDB(ctx, rootstore, options.dbOpts...)
+	db, err := db.NewDB(ctx, rootstore, acp.NoACP, options.dbOpts...)
```
shahzadlone, line 124: question: Why not also dependent on the default options like for `rootstore` above?

node/acp.go
```diff
@@ -0,0 +1,83 @@
...
+		opt(options)
+	}
+
+	switch options.acpType {
+	case NoACPType:
+		return acp.NoACP, nil
+
+	case LocalACPType:
+		var acpLocal acp.ACPLocal
+		acpLocal.Init(context.Background(), options.path)
```
nasdf, line 75: todo: use context from func signature

node/acp.go
```diff
@@ -0,0 +1,83 @@
...
+	NoACPType    ACPType = 0
+	LocalACPType ACPType = 1
+)
+
+// ACPOptions contains ACP configuration values.
+type ACPOptions struct {
+	acpType ACPType
+
+	// Note: An empty path will result in an in-memory ACP instance.
+	path string
```
shahzadlone, line 33: question: would `target` make more sense? as path is specific to local acp, I wonder if this will cause confusion for remote acp as it might maybe have a target/endpoint or something to sourcehub, not necessarily a path?

node/acp.go
```diff
@@ -0,0 +1,83 @@
...
+		return acp.NoACP, nil
+
+	case LocalACPType:
+		var acpLocal acp.ACPLocal
+		acpLocal.Init(context.Background(), options.path)
+		return immutable.Some[acp.ACP](&acpLocal), nil
+
+	default:
+		var acpLocal acp.ACPLocal
+		acpLocal.Init(context.Background(), options.path)
```
nasdf, line 80: todo: use context from func signature

nasdf (changes requested):
Two small todos, but overall looks great!

net/peer_test.go
shahzadlone, line 1: suggestion: Perhaps not in this PR, but wonder if we should have a `peer_acp_test.go` as I see some tests here use node with acp and some without (maybe can test both variants).

node/acp.go
AndrewSisley: :) My bad, I should have reviewed this PR more carefully before opening it (is the second silly mistake here) - thanks

- [ ] ctx

cli/server_dump.go
shahzadlone, line 1: question: This makes me think, would we want server dump command to be under AAC? or is bypassing acp here somewhat safe?

node/acp.go
AndrewSisley: - [ ] Fix ctx

node/acp.go
AndrewSisley: `path` is consistent with the root store equivalent, I have a preference for us to use the same terminology

net/peer_test.go
AndrewSisley: Not in this PR, off the top of my head I would prefer to delete this file than to grow it (or add another similar)

cli/server_dump.go
AndrewSisley: I would be surprised if we wouldn't want it under AAC

cli/start.go
```diff
@@ -140,7 +140,7 @@ func MakeStartCommand() *cobra.Command {
 				// TODO-ACP: Infuture when we add support for the --no-acp flag when admin signatures are in,
 				// we can allow starting of db without acp. Currently that can only be done programmatically.
 				// https://github.com/sourcenetwork/defradb/issues/2271
-				dbOpts = append(dbOpts, db.WithACP(rootDir))
+				acpOpts = append(acpOpts, node.WithACPPath(rootDir))
```
shahzadlone, line 143: question/nitpick: Why not maintain the InMem and non InMem variants? if getting rid because it seems specific to local acp, then might as well also change `node.WithACPPath` to  `node.WithACPTarget`?

node/acp.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// Note: An empty path will result in an in-memory ACP instance.
+func WithACPPath(path string) ACPOpt {
+	return func(o *ACPOptions) {
+		o.path = path
+	}
+}
+
+// NewACP returns a new ACP module with the given options.
+func NewACP(ctx context.Context, opts ...ACPOpt) (immutable.Option[acp.ACP], error) {
```
shahzadlone, line 63: suggestion: Can preserve / modify the tests that were removed in `internal/db/config_test.go` to test the public functions in this file. Perhaps in a `node/acp_test.go` file or something?

cli/start.go
AndrewSisley: > Why not maintain the InMem and non InMem variants?

It just didn't seem to be worth keeping atm (increased surface area, and we would never reference it ourselves).  Long-term it's current implementation doesn't really make that much sense either as it doesn't have to be local-acp specific either.  Better to have a proper think about it if/when the demand for it ever arises.

node/acp.go
AndrewSisley: I would much rather spend time fixing the integration tests to cover this properly than to add unit tests here. https://github.com/sourcenetwork/defradb/issues/2634

cli/start.go
shahzadlone: I guess that makes sense, unless a user wanted to use specifically in-mem or persistence acp model that didn't match the type of db persistence we want.

For now we can just assume, for an in-mem DB, we use in mem ACP, and vice-versa

node/acp.go
shahzadlone: Right I get that, I am just wondering if path would still make sense for remote ACP?

AndrewSisley: No idea, if we want to rename it then we can do though

node/acp.go
shahzadlone: Thanks for making the issue. I do believe the tests that were moved were very recently added, I forget why that extra effort was put and if there was a reason for them.

cli/server_dump.go
shahzadlone: Actually then in that case wouldn't we want to enable acp, or if turning it off then add a todo linked to AAC issue saying that we will re-enable when implemented.

nasdf (approved):

cli/server_dump.go
AndrewSisley: I do not know, we can think about this when we get round to covering this with AAC :)

shahzadlone: The things we don't know should IMO have acp enabled.

fredcarle (approved):
LGTM

cli/server_dump.go
AndrewSisley: It wasn't enabled before.  This PR is not the place to enable it here. (note how this is bypassing `node` like the integration tests, and not supplying any options, `db.acp` would remain a `None` value).

shahzadlone (approved):
Cheers for the answers and doing the requested changes =) LGTM

PR #2644, opened 2024-05-23
test(i): Don't setup http and p2p by default in test suite

### LINKED ISSUE #2643: Integration tests `setupDatabase` is spinning up p2p and http servers
And not closing them down.

Introduced by https://github.com/sourcenetwork/defradb/pull/2641

AndrewSisley:
## Relevant issue(s)

Resolves #2643

## Description

Don't setup http and p2p by default in test suite.

This is a regression I introduced in https://github.com/sourcenetwork/defradb/pull/2641 (sorry for the bother) - we set p2p and http up later.  As the new, unused, P2P and http servers werent cleaned up at the end of the test, this had quite a bad effect on test suite memory consumption.

Later, I think the surrounding code that sets up p2p and http could probably be changed later to make better use of `node.New`, but IMO this is good enough for now (we have other things to do).

### DIFF
--- a/tests/integration/db.go
+++ b/tests/integration/db.go
@@ -109,7 +109,14 @@ func setupDatabase(s *state) (client.DB, string, error) {
 	}
 	storeOpts := []node.StoreOpt{}
 	acpOpts := []node.ACPOpt{}
-	opts := []node.NodeOpt{}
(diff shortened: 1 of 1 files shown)


nasdf (approved):
LGTM

PR #2647, opened 2024-05-23
ci(i): Add workflow to ensure cli docs are in sync

### LINKED ISSUE #2165: Assert output of `make docs` is up to date in CI

shahzadlone:
## Relevant issue(s)
Resolves #2165

## Description
- Detects if cli documents need updating.
- If Documentation is up to date the action passes, else fails.
- Run `make docs` or `make docs:cli` to generate the documentation.
- Introducing as a non-required check, but would argue it should be required.

## Tasks

## How has this been tested?
- Manually testing through action run and by using `act` utility.
- Failed action: https://github.com/sourcenetwork/defradb/actions/runs/9214068142/job/25349425467?pr=2647

- WSL2

### DIFF
--- a/.github/workflows/check-cli-documentation.yml
+++ b/.github/workflows/check-cli-documentation.yml
@@ -1,0 +1,61 @@
+# Copyright 2024 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+# This workflow checks that all CLI documentation is up to date.
+# If the documentation is not up to date then this action will fail.
+name: Check CLI Documentation Workflow
+
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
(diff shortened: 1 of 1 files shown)


.github/workflows/check-cli-documentation.yml
```diff
@@ -0,0 +1,59 @@
...
+  check-cli-documentation:
+    name: Check cli documentation job
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code into the directory
+        uses: actions/checkout@v3
+
+      - name: Check no changes exist initially
```
AndrewSisley, line 38: question: Why is this step needed?

todo: If keeping, please document why we want this :)

fredcarle (approved):
Approving assuming Andy's comment gets resolved.

.github/workflows/check-cli-documentation.yml
shahzadlone: Incase upon checkout some dirty state exists (i don't think it's going to happen, but just for extra caution). It's be nice to catch that inconsistency.

Happy to remove if there is preference/consensus.

AndrewSisley: Okay, I'm somewhat sceptical if a check here is the right place for that, but it is a cheap action and not worth spending time on :)

Please do document this though before merge.

AndrewSisley (approved):
Approving assuming that the 2nd step is documented before merge - thanks Shahzad :)

Issue #2357, opened 2024-02-28
ACP - Panic If Invalid Source Address being passed to AddPolicy


shahzadlone:
This is a bug on sourcehub, which bruno will fix soon.

This will be solved when his fix is merged on `dev` branch of sourcehub, or if this issue is solved first.

```
func TestACP_AddPolicy_InvalidCreatorIdentityWithValidPolicy_Error(t *testing.T) {
	test := testUtils.TestCase{
		Description: "Test acp, adding policy, with invalid creator, with valid policy, return error",

		Actions: []any{
			testUtils.AddPolicy{
				IsYAML: true,

				Creator: "invalid",

				Policy: `
                    description: a basic policy that satisfies minimum DPI requirements

                    actor:
                      name: actor

                    resources:
                      users:
                        permissions:
                          read:
                            expr: owner
                          write:
                            expr: owner

                        relations:
                          owner:
                            types:
                              - actor

                `,

				ExpectedError: "policy creator can not be empty",
			},
		},
	}

	testUtils.AssertPanic(t, func() { testUtils.ExecuteTestCase(t, test) })
}

```

PR #2649, opened 2024-05-25
feat: Add authentication for ACP

### LINKED ISSUE #2017: ACP - Full Identity Authentication
ACP - Full Identity Authentication

Initially we only have simple identity

nasdf:
## Relevant issue(s)

Resolves #2017

## Description

This PR adds ACP identity authentication via HTTP.

Notable changes:
- `acp/identity` has been replaced with the `acp.Identity` struct
  - `identity.PrivateKey` is the private key of the identity
  - `identity.PublicKey` is the public key of the identity
  - `identity.Address` is the bech32 formatted address for the identity
  - keys are all `secp256k1`
- `http` can authenticate requests using a jwt bearer token
  - ~a random `audience` value is generated on every http server startup~
  - audience must be set to the defradb host name
  - ~api route `/audience` returns the random audience value~
  - `http.Client` will create a signed token if an `acp.PrivateKeyIdentity` is set
  - jwt token subject is the identity public key
- `cli` `--identity` flag is now a hex encoded private key

Todo:
  - [x] ensure acp docs are updated

## Tasks

## How has this been tested?

`make test`

- MacOS


acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// NoIdentity specifies an anonymous actor.
+var NoIdentity = immutable.None[Identity]()
+
+// Identity uniquely identifies an acp actor.
+type Identity interface {
+	// PublicKey returns the public key of the identity.
+	PublicKey() *secp256k1.PublicKey
+	// Address returns the bech32 address of the identity.
```
AndrewSisley, line 27: suggestion: The callers of this function (with a lazy ctrl+f in the browser) are in relatively generalized parts of the codebase where maintainers do not need to understand or know what a `bech32 address` and why that is significant.

At the moment this comment/documentation does not describe if or why it being `bech32` is significant, and the calling sites appear to be ignorant of that too.

If its `bech32`ness is important, please describe why.  If it is not I suggest either removing it from the comment, or reducing its importance by moving it onto a secondary line (e.g. below).
```
// Address returns the address of the identity.  // the address where (maybe also note if this is in sourcehub/keyring/whereever)
//
// note on it being in bech32 format
```

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/sourcenetwork/immutable"
+)
+
+// NoIdentity specifies an anonymous actor.
+var NoIdentity = immutable.None[Identity]()
+
+// Identity uniquely identifies an acp actor.
+type Identity interface {
+	// PublicKey returns the public key of the identity.
+	PublicKey() *secp256k1.PublicKey
```
AndrewSisley, line 26: question: `secp256k1.PublicKey` is a very specific type to be returning from an otherwise quite generic looking interface, is there another slightly more abstract type that we can return instead to save coupling all consumers of this function to a very specific key type?

http/handler.go
```diff
@@ -74,15 +76,26 @@ func NewHandler(db client.DB) (*Handler, error) {
...
 		r.Use(
 			ApiMiddleware(db, txs),
 			TransactionMiddleware,
-			IdentityMiddleware,
+			AuthMiddleware(audience),
 		)
+		r.Get("/audience", func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Add("Content-Type", "text/plain")
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(audience)) //nolint:errcheck
```
AndrewSisley, line 97: todo: Please document why it is safe to ignore this error, if it is safe.

tests/integration/acp/add_policy/with_invalid_creator_arg_test.go
```diff
@@ -51,22 +49,19 @@ func TestACP_AddPolicy_InvalidCreatorIdentityWithValidPolicy_Error(t *testing.T)
 		},
 	}
 
-	//TODO-ACP: https://github.com/sourcenetwork/defradb/issues/2357
-	testUtils.AssertPanic(t, func() { testUtils.ExecuteTestCase(t, test) })
```
AndrewSisley, line 55: todo: You appear to have changed the test so that it no longer tests what the name and description say it does.

The linked ticket https://github.com/sourcenetwork/defradb/issues/2357 suggests that this is because of a sourcehub bug, which presumably is not fixed by this PR.

Is it now impossible to provide an invalid identity to defra? (this seems unlikely to me).  If it is still possible to provide an invalid identity, please preserve this testing what it was.   Otherwise it might be best to delete it and close the ticket.

Please reply to this message with what you have done when you have done it, so I don't forget or lose track of it :)

http/auth_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/sourcenetwork/defradb/acp"
+	"github.com/sourcenetwork/defradb/crypto"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildAuthToken(t *testing.T) {
```
AndrewSisley, line 26: question: How is authentication tested?  Is it just this file, or is it automagically covered by the http/cli integration tests?  Is authentication now always required when ACP is enabled?  Is there now no easy way for users of the embedded Go client to provide an ACP identity (is this untested?)?

AndrewSisley (changes requested):
Looks good, but I have a handful of questions/todos for you at the moment :)

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	// Address returns the bech32 address of the identity.
+	Address() string
```
AndrewSisley, line 28: question: Why are you returning a `bech32` string instead of a `"github.com/cosmos/cosmos-sdk/crypto/types"`.`Address`?

tests/integration/acp/add_policy/with_empty_args_test.go
```diff
@@ -42,8 +42,6 @@ func TestACP_AddPolicy_EmptyPolicyCreator_Error(t *testing.T) {
 
 		Actions: []any{
 			testUtils.AddPolicy{
-				Identity: "",
```
shahzadlone, line 45: todo: I don't like these changes, the tests should explicitly show the identity being empty.

The previous tests should stay as much as possible in my opinion. I think what you want to test in this PR and what the previous tests test are 2 fundamentally separate things.

The previous tests assert that the RAW identity (without being wrapped in auth token) work. Which should continue behaving like they did.

You should introduce a new Variable called `IdentityWithAuth` within the respective `testUtils.XYZ` util functions where you want to test the authentication, then copy the tests you think are relavent from `tests/integration/acp` and add them to a dir called `test/integration/acp/auth`. This way we don't mix the testing of authentication and the acp logic.

So al/mostl the previous tests should be reverted and unchanged in this PR and new one introduced with using `IdentityWithAuth` instead of `Identity` (you this way also don't have to re-test  all acp cases, but only the authentication functionality.

shahzadlone (changes requested):

acp/identity/identity.go
shahzadlone, line 1: issue/conflict: You have removed the identity package, while my branch has expanded it (with some mostly similar functionality to your identity stuff). We should definitely coordinate on the structure.

I was thinking to actually more the `indentity` package perhaps out of the `acp` package into it's own thing. Can discuss that.

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	return i.pubKey
+}
+
+func (i PublicKeyIdentity) Address() string {
+	pubKey := cosmosSecp256k1.PubKey{Key: i.pubKey.SerializeCompressed()}
+	return types.MustBech32ifyAddressBytes("cosmos", pubKey.Address().Bytes())
+}
+
+// PrivateKeyIdentity is an identity with both a private and public key.
+type PrivateKeyIdentity struct {
```
shahzadlone, line 54: todo: I don't like this name it's a bit confusing as to what different type of identities there are now? I think all you wanted was a struct for private key?

tests/integration/acp/add_policy/with_empty_args_test.go
AndrewSisley: > The previous tests assert that the RAW identity (without being wrapped in auth token) work. Which should continue behaving like they did.

This is related to one of my questions I think.  We might want to chat as a team as to whether we want to support unauthenticated identities in the future.  The jwt token is also required for SourceHub as far as I can tell, and the usefulness of local-ACP is debated atm.

tests/integration/acp/add_policy/with_invalid_creator_arg_test.go
AndrewSisley: > The linked ticket https://github.com/sourcenetwork/defradb/issues/2357 suggests that this is because of a sourcehub bug, which presumably is not fixed by this PR.

note: This is actually fixed in the `feat/sdk-pkg` branch of SourceHub, and maybe the latest `dev` branch

acp/identity.go
fredcarle: I think it's because `acpSystem.RegisterDocObject` takes a string and that method is used for that.

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/sourcenetwork/immutable"
+)
+
+// NoIdentity specifies an anonymous actor.
+var NoIdentity = immutable.None[Identity]()
+
+// Identity uniquely identifies an acp actor.
+type Identity interface {
```
shahzadlone, line 24: conflict: My branch introduces similar functionality but instead of using `Identity` as an interface and then two different structs implementing it, there is just one `Identity` struct that has three components:

``` go
// Identity holds the bits that make up an actor's unique identity.
type Identity struct {
	// An actor's private key. Uses secp256k1 algorithm.
	PrivateJWK string

	// An actor's corresponding public key address.
	PublicKey string

	// An actor's DID. Generated from the public key address.
	DID string
}
```

Note: `did` is what it will be after my stuff merges not a sourcehub address.

IMO we don't need to over abstract the identity with an interface, just a struct like above would suffice where what would have as `PrivateKeyIdentity` would just be the `Identity` struct above with all non-empty fields, and `PublicKeyIdentity` would the `Identity` struct above with just `PrivateJWK` as empty. One liner documenting this on the struct would make it clear between these differences.

todo: unabstract interface identity implementation

crypto/crypto.go
```diff
@@ -13,17 +13,29 @@ package crypto
 import (
 	"crypto/ed25519"
 	"crypto/rand"
+
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
```
shahzadlone, line 17: discussion:

I think I was using `	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"`

Any differences you know between them?

crypto/crypto.go
```diff
@@ -13,17 +13,29 @@ package crypto
...
 )
 
+// GenerateSecp256k1 generates a new secp256k1 private key.
+func GenerateSecp256k1() (*secp256k1.PrivateKey, error) {
```
shahzadlone, line 21: question: You only need this for testing right? not for any production code?

shahzadlone (changes requested):
Did a first pass quickly, left some todos, will review again tonight.

Can you please also update the acp readme with authentication cli examples now that we will always have acp with authentication. It will help better understand for reviewers as well, how acp with auth works.

tests/integration/acp/add_policy/with_empty_args_test.go
shahzadlone: When I said 

> Which should continue behaving like they did.

I meant the access control part should behave normally as it did, sorry. Not that we should keep non-auth identity functionality.

In the comment above I am suggesting to not mix the "authentication" testing with the "access control" testing, and keep them separate.

As for weather we want unauthenticated identities, I really can't think of a good reason why we would want to keep that if we have auth implemented.

As for:

> the usefulness of local-ACP is debated atm

I don't see how it's debated haha, IMO is really useful for anyone who wants to use defradb like a classic/normal use case (i.e. without a sourcehub-like instance). We can test in standup though, maybe I am missing something.

acp/identity.go
shahzadlone: I don't think this matters as address will be replaced with `did` in a follow up work

tests/integration/acp/add_policy/with_empty_args_test.go
AndrewSisley: > In the comment above I am suggesting to not mix the "authentication" testing with the "access control" testing, and keep them separate.

How would you keep them separate without permitting auth-free identities?

tests/integration/acp/add_policy/with_invalid_creator_arg_test.go
shahzadlone: When we bump to latest sourcehub commit on `dev` it I believe fixes this panic.

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	return immutable.Some(Identity(identity))
+}
+
+func (i PublicKeyIdentity) PublicKey() *secp256k1.PublicKey {
+	return i.pubKey
+}
+
+func (i PublicKeyIdentity) Address() string {
+	pubKey := cosmosSecp256k1.PubKey{Key: i.pubKey.SerializeCompressed()}
+	return types.MustBech32ifyAddressBytes("cosmos", pubKey.Address().Bytes())
```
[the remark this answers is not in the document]
nasdf: Since we are going between well known formats this should never panic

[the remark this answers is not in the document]
AndrewSisley: Can you please add a comment explaining that :)

http/auth_test.go
nasdf: > How is authentication tested?

It's tested by the integration tests as well. Any place we are using an identity in a test will use authentication when using the http / cli clients.

> Is there now no easy way for users of the embedded Go client to provide an ACP identity (is this untested?)?

This doesn't change anything with how we use identity in the Go client.

```go
identity := acp.IdentityFromPublicKey(...)
ctx := db.SetContextIdentity(ctx, identity)
```

tests/integration/acp/add_policy/with_empty_args_test.go
nasdf: > The previous tests assert that the RAW identity (without being wrapped in auth token) work. Which should continue behaving like they did.

The Go client doesn't create auth tokens. Only the HTTP / CLI clients use auth tokens.

The change here was made because the new identity requires a valid public or private key and an empty string is not a valid identity.

acp/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	// PublicKey returns the public key of the identity.
+	PublicKey() *secp256k1.PublicKey
+	// Address returns the bech32 address of the identity.
+	Address() string
+}
+
+var _ (Identity) = (*PublicKeyIdentity)(nil)
+
+// PublicKeyIdentity is an identity with only a public key.
+type PublicKeyIdentity struct {
```
AndrewSisley, line 34: question: I can't spot where this is used, is it dead code?

crypto/crypto.go
nasdf: Yes, I'm planning on adding a way to generate and store them in the keyring as well.

acp/identity.go
nasdf: It's a client facing utility, but will be removed when I merge in Shahzad's identity changes.

AndrewSisley: What will replace it?  Is there any value to it living for just the (short?) interval between merging this PR and Shahzad's changes?

I'm asking partly because SourceHub ACP needs stuff to be signed (client-side), which requires the private key.

nasdf: > What will replace it?

Something like this

```go
type Identity struct {
	// PublicKey is the identity public key.
	PublicKey *secp256k1.PublicKey
	// PrivateKey is the identity private key.
	PrivateKey *secp256k1.PrivateKey
	// Address is the identity address.
	Address string
}
```

> Is there any value to it living for just the (short?) interval between merging this PR and Shahzad's changes?

It's a quick change and Shahzad will most likely want this done before approving.

acp/identity.go
nasdf: I've flattened like the identity interface to struct like you suggested. Let me know which package you would like for it to exist in.

acp/identity.go
nasdf: The keys will be changed to a JWK shortly after this PR. It's probably not worth the effort to change them in the mean time.

shahzadlone (commented):
Some more questions.

Question: Was the ability to send requests without an identity (optionality of an identity being set or specified) removed? IMO we need to maintain optional identity.

Question: If Identity is still optional, when a verification fails, do you error out or assume that no identity was provided and still execute request? IMO should be an error even if optional identity is stripped.

internal/db/db.go
```diff
@@ -169,12 +169,11 @@ func (db *db) AddPolicy(
 	}
 	identity := GetContextIdentity(ctx)
 
-	var creatorID string
-	if identity.HasValue() {
```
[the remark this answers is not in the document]
nasdf: This should be the same behavior as before. If the identity is empty the creator address will be an empty string.

tests/clients/cli/wrapper_cli.go
```diff
@@ -60,8 +59,8 @@ func (w *cliWrapper) executeStream(ctx context.Context, args []string) (io.ReadC
 		args = append(args, "--tx", fmt.Sprintf("%d", tx.ID()))
 	}
 	id := db.GetContextIdentity(ctx)
```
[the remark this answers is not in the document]
nasdf: Since identity is a struct it will always have a default value. In this case we only care about the private key being non nil.

nasdf:
> Question: Was the ability to send requests without an identity (optionality of an identity being set or specified) removed? IMO we need to maintain optional identity.

No

> Question: If Identity is still optional, when a verification fails, do you error out or assume that no identity was provided and still execute request? IMO should be an error even if optional identity is stripped.

If verification fails the request returns a `403` status code / forbidden error.

AndrewSisley (changes requested):
Re-requesting changes, you appear to have forgotten about/forgotten to push a few of my previous todos.

tests/integration/acp/add_policy/with_invalid_creator_arg_test.go
nasdf: I've restored the tests but limited them to go client only.

AndrewSisley: Thanks Keenan, and thanks for the code-comments explaining why :)

acp/identity/identity.go
```diff
@@ -8,34 +8,48 @@
...
 }
 
-// String returns the string representation of the identity.
-func (i Identity) String() string {
-	return string(i)
+// AddressFromPublicKey returns the unique address of the given public key.
+func AddressFromPublicKey(publicKey *secp256k1.PublicKey) string {
+	pub := cosmosSecp256k1.PubKey{Key: publicKey.SerializeCompressed()}
+	// conversion from well known types should never cause a panic
+	return types.MustBech32ifyAddressBytes("cosmos", pub.Address().Bytes())
```
AndrewSisley, line 54: thought:  This looks like it might actually panic if `"cosmos"` is no longer the prefix (it is actually being changed at the moment, so that's not without precedent).

todo: There is a const for this that you can use instead of a magic string, please use `"github.com/sourcenetwork/sourcehub/app"`.`AccountAddressPrefix`

http/handler.go
```diff
@@ -74,15 +76,28 @@ func NewHandler(db client.DB) (*Handler, error) {
...
 			ApiMiddleware(db, txs),
 			TransactionMiddleware,
-			IdentityMiddleware,
+			AuthMiddleware(audience),
 		)
+		r.Get("/audience", func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Add("Content-Type", "text/plain")
+			w.WriteHeader(http.StatusOK)
+			// ignore errors because the headers have
+			// already been written to the response stream
```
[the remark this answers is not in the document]
nasdf: I've updated both response writer utils to log any errors.

acp/identity/identity.go
nasdf: Using that causes a panic `panic: invalid Bech32 prefix; expected cosmos, got source`

AndrewSisley: Ah okay, using the const might have to wait on Shahzad's PR then, and I've been chatting with Bruno today - he might change the source hub sdk so that Defra controls the prefix (in which case we can either own a const, or have a config option for this).

AndrewSisley (approved):
LGTM, thanks Keenan! I would prefer that you wait on Shahzad's approval before merge though :)

acp/identity/identity.go
```diff
@@ -8,34 +8,48 @@
...
-Package identity provides defradb identity.
-*/
-
 package identity
 
-import "github.com/sourcenetwork/immutable"
+import (
+	cosmosSecp256k1 "github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
```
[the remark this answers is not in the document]
nasdf: Discussed in call: `github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1` will be removed when we switch to DIDs.

cli/acp_policy_add.go
```diff
@@ -26,6 +26,7 @@ func MakeACPPolicyAddCommand() *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "add [-i --identity] [policy]",
 		Short: "Add new policy",
+		//nolint:lll
```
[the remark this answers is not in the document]
nasdf: I've fixed the linter error issues.

acp/identity/identity.go
```diff
@@ -8,34 +8,48 @@
...
-var (
-	// None is an empty identity.
-	None = immutable.None[Identity]()
-)
+type Identity struct {
+	// PublicKey is the identity public key.
+	PublicKey *secp256k1.PublicKey
+	// PrivateKey is the identity private key.
+	PrivateKey *secp256k1.PrivateKey
+	// Address is the identity address.
```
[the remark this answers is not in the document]
nasdf: I've updated the documentation here as suggested

acp/identity/identity.go
```diff
@@ -8,34 +8,48 @@
...
+var None = immutable.None[Identity]()
 
-var (
-	// None is an empty identity.
-	None = immutable.None[Identity]()
-)
+type Identity struct {
+	// PublicKey is the identity public key.
+	PublicKey *secp256k1.PublicKey
+	// PrivateKey is the identity private key.
```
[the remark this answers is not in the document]
nasdf: I've updated the documentation here as suggested

acp/identity/identity.go
```diff
@@ -8,34 +8,48 @@
...
-type Identity string
+// None specifies an anonymous actor.
+var None = immutable.None[Identity]()
 
-var (
-	// None is an empty identity.
-	None = immutable.None[Identity]()
-)
+type Identity struct {
+	// PublicKey is the identity public key.
```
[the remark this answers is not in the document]
nasdf: I've updated the documentation here as suggested

jsimnz:
reviewing :) 👀

tests/clients/cli/wrapper_cli.go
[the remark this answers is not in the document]
jsimnz: the point of the `Option` is to express the zero/empty value explicitly, rather than implicitly with Go zero values (like a default struct).

If we're going to wrap the identity in the `Option` type, we should at least use it correctly.

jsimnz (approved):
Obviously there is on going work with this part of the codebase, approving now so I don't further block the merging. Only one todo, and some thoughts/questions

acp/identity/identity.go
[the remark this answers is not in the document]
shahzadlone: > Discussed in call: `github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1` will be removed when we switch to DIDs.

But we still need it for the one case of the hardcoded source address so I think it will remain :/

acp/README.md
```diff
@@ -145,9 +145,40 @@ Here are some valid expression examples. Assuming these `expr` are under a requi
...
+    14:1e:40:b8:75:62:04:8e:dd:a0:7d:41:32:c6:10:
+    7d:5a:9f:c9:e5:8f:6a:e3:95:88:88:ef:86:e1:86:
+    45:d6:84:dc:79
+ASN1 OID: secp256k1
+```
+
+Authenticate with the identity:
+
+```sh
+defradb client ... --identity e3b722906ee4e56368f581cd8b18ab0f48af1ea53e635e3f7b8acd076676f6ac
```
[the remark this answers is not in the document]
nasdf: I'm planning on adding keyring support for identity keys in a follow up PR. I think Shahzad is working on a command that will generate an identity with a private key as well.

http/handler.go
```diff
@@ -74,15 +76,26 @@ func NewHandler(db client.DB) (*Handler, error) {
 	if err != nil {
 		return nil, err
 	}
+	// generate a secure random audience value
+	audienceBytes, err := crypto.RandomBytes(64)
```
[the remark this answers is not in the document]
nasdf: The problem I ran into is that the peer ID can't easily be retrieved when p2p is disabled.

acp/README.md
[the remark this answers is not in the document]
shahzadlone: Thinking again the `:` is not nice. I am wondering if the `util` command should generate a `Hex` version of the private key as well. I do know that it's temporary so perhaps until key ring is implemented the `Hex` version of private key can be in the util command.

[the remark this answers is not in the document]
shahzadlone: technically can also use this command, to get the trimmed hex without `:` private key:

```
openssl ecparam -name secp256k1 -genkey | openssl ec -text -noout | head -n5 | tail -n3 | tr -d '\n:\ '
```

tests/integration/acp/fixture.go
```diff
@@ -10,7 +10,33 @@
...
+)
+
 var (
-	Actor1Identity = "cosmos1zzg43wdrhmmk89z3pmejwete2kkd4a3vn7w969"
-	Actor2Identity = "cosmos1x25hhksxhu86r45hqwk28dd70qzux3262hdrll"
+	Actor1Identity immutable.Option[acpIdentity.Identity]
+	Actor2Identity immutable.Option[acpIdentity.Identity]
 )
+
+func init() {
```
shahzadlone, line 27: question: Why does this need to be in an `init` function?

todo: I would much prefer a util function some what like this:

```go
func toIdentity(privateHex string) immutable.Option[acpIdentity.Identity] {
	privateKeyBytes, err := hex.DecodeString(privateHex)
	if err != nil {
		panic(err)
	}

	return acpIdentity.FromPrivateKey(
		secp256k1.PrivKeyFromBytes(
			privateKeyBytes,
		),
	)
}

var (
	Actor1Identity = toIdentity("028d53f37a19afb9a0dbc5b4be30c65731479ee8cfa0c9bc8f8bf198cc3c075f")
	Actor2Identity = toIdentity("4d092126012ebaf56161716018a71630d99443d9d5217e9d8502bb5c5456f2c5")
)

```

suggestion: Something similar can replace the `loadIdentity` function you have in the other test as well.

tests/integration/acp/index/fixture.go
```diff
@@ -39,7 +39,7 @@ resources:
           - actor
 `
 
-// policy id: "68a4e64d5034b8a0565a90cd36483de0d61e0ea2450cf57c1fa8d27cbbf17c2c"
+// policy id: "e3c35f345c844e8c0144d793933ea7287af1930d36e9d7d98e8d930fb9815a4a"
```
shahzadlone, line 42: praise: Thanks for maintaining the documentation accuracy here, would also be worth fixing the user facing examples in the example directory and the acp readme as well (I think I mentioned in other comments).

shahzadlone (changes requested):
Thanks Keenen for doing the changes. I have some more comments that I have posted, please ping me once ready for re-review, and will give another final review :) Looks really nice now and I am super excited to have this in.

Sorry I forgot to post the action item points from our zoom call yesterday, I know you have done some of those items, but just leaving them here for reference to knock them all down pre-merge.

- Explicit empty identity in the tests that it was removed from
- Use identity option before checking private key, or re-think the option data struct now that identity must have private before verification and will only have public key after verfication.
- Documentation requests:
    - Document how to generate hex from private key for CLI
    - Document why we use HEX for CLI
    - Document HEX type for decoding and encoding
    - Document HTTP auth-flow 
    - Document command using `/audiance` endpoint/domain for HTTP
    - Document how to get the signed token with HTTP
    - Document where to put the signed token with HTTP (i.e. `Bearer` often times people confuse this with GQL Variable haha)

net/peer_test.go
```diff
@@ -1079,3 +1084,10 @@ func TestHandleDocUpdateLog_WithExistingSchemaTopic_TopicExistsError(t *testing.
 	})
 	require.ErrorContains(t, err, "topic already exists")
 }
+
+func loadIdentity(t *testing.T) immutable.Option[acpIdentity.Identity] {
+	privKeyBytes, err := hex.DecodeString("028d53f37a19afb9a0dbc5b4be30c65731479ee8cfa0c9bc8f8bf198cc3c075f")
+	require.NoError(t, err)
+	privKey := secp256k1.PrivKeyFromBytes(privKeyBytes)
+	return acpIdentity.FromPrivateKey(privKey)
+}
```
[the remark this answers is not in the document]
shahzadlone: or a helper like `toIdentity` I suggested in: https://github.com/sourcenetwork/defradb/pull/2649#discussion_r1621484244

cli/collection_delete.go
```diff
@@ -28,7 +28,8 @@ Example: delete by docID:
   defradb client collection delete  --name User --docID bae-123
 
 Example: delete by docID with identity:
-  defradb client collection delete -i cosmos1f2djr7dl9vhrk3twt3xwqp09nhtzec9mdkf70j --name User --docID bae-123
+  defradb client collection delete --name User --docID bae-123 \
+  	-i 028d53f37a19afb9a0dbc5b4be30c65731479ee8cfa0c9bc8f8bf198cc3c075f
```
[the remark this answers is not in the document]
nasdf: It works for me

http/handler.go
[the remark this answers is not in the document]
nasdf: Audience is now the api host name. Will update to include peer ID in a follow up PR.

shahzadlone (approved):
Awesome work @nasdf appreciate you doing all the changes :)

One last nitpick, before merge if you can change the title of the PR to have the first word as an action verb instead of `ACP` it would be nice.

For example, `feat: Add Authentication for ACP`

PR #2656, opened 2024-05-29
refactor: Extract Defra specific logic from ACPLocal type

### LINKED ISSUE #2655: Extract Defra specific logic from ACPLocal type
So that it can also be used by SourceHub ACP.

AndrewSisley:
## Relevant issue(s)

Resolves #2655

## Description

Extracts Defra specific logic from the ACPLocal type so that it may be re-used for SourceHub ACP.

None of the code has changed, it's just moved around.

I'm opening a PR now, as a lot of us (including Bruno) are working in this space, and merging refactors like this asap tends to minimise conflicts.

The (WIP) source hub implementation can be found here: https://github.com/sourcenetwork/defradb/pull/2657

### DIFF
--- a/acp/source_hub_client.go
+++ b/acp/source_hub_client.go
@@ -21,10 +21,10 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 )
 
-// sourcehubClient is a private abstraction to allow multiple ACP implementations
+// sourceHubClient is a private abstraction to allow multiple ACP implementations
 // based off of the SourceHub libraries to share the same Defra-specific logic via the
 // sourceHubBridge.
-type sourcehubClient interface {
+type sourceHubClient interface {
 	// Init initializes the acp, with an absolute path. The provided path indicates where the
 	// persistent data will be stored for acp.
 	//
@@ -88,10 +88,10 @@ type sourcehubClient interface {
 	Close() error
 }
 
-// sourceHubBridge wraps a sourcehubClient, hosting the Defra-specific logic away from client-specific
+// sourceHubBridge wraps a sourceHubClient, hosting the Defra-specific logic away from client-specific
 // code.
 type sourceHubBridge struct {
-	client sourcehubClient
+	client sourceHubClient
 }
 
 var _ ACP = (*sourceHubBridge)(nil)
--- a/acp/source_hub_client.go
+++ b/acp/source_hub_client.go
@@ -1,0 +1,323 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package acp
+
+import (
+	"context"
+
+	protoTypes "github.com/cosmos/gogoproto/types"
+	"github.com/sourcenetwork/corelog"
+	"github.com/sourcenetwork/sourcehub/x/acp/types"
+	"github.com/valyala/fastjson"
+
+	"github.com/sourcenetwork/defradb/errors"
+)
+
+// sourcehubClient is a private abstraction to allow multiple ACP implementations
+// based off of the SourceHub libraries to share the same Defra-specific logic via the
+// sourceHubBridge.
+type sourcehubClient interface {
+	// Init initializes the acp, with an absolute path. The provided path indicates where the
+	// persistent data will be stored for acp.
+	//
+	// If the path is empty then acp will run in memory.
+	Init(ctx context.Context, path string)
+
+	// Start starts the acp, using the initialized path. Will recover acp state
+	// from a previous run if under the same path.
+	//
+	// If the path is empty then acp will run in memory.
+	Start(ctx context.Context) error
+
+	// AddPolicy attempts to add the given policy. Upon success a policyID is returned,
+	// otherwise returns error.
+	AddPolicy(
+		ctx context.Context,
+		creatorID string,
+		policy string,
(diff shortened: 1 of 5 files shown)


acp/source_hub_client.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/sourcenetwork/sourcehub/x/acp/types"
+	"github.com/valyala/fastjson"
+
+	"github.com/sourcenetwork/defradb/errors"
+)
+
+// sourcehubClient is a private abstraction to allow multiple ACP implementations
+// based off of the SourceHub libraries to share the same Defra-specific logic via the
+// sourceHubBridge.
+type sourcehubClient interface {
```
shahzadlone, line 27: todo:

```suggestion
type sourceHubClient interface {
```

acp/source_hub_client.go
```diff
@@ -0,0 +1,323 @@
...
+
+	protoTypes "github.com/cosmos/gogoproto/types"
+	"github.com/sourcenetwork/corelog"
+	"github.com/sourcenetwork/sourcehub/x/acp/types"
+	"github.com/valyala/fastjson"
+
+	"github.com/sourcenetwork/defradb/errors"
+)
+
+// sourcehubClient is a private abstraction to allow multiple ACP implementations
```
shahzadlone, line 24:
```suggestion
// sourceHubClient is a private abstraction to allow multiple ACP implementations
```

acp/source_hub_client.go
```diff
@@ -0,0 +1,323 @@
...
+	) (bool, error)
+
+	// Close closes any resources in use by acp.
+	Close() error
+}
+
+// sourceHubBridge wraps a sourcehubClient, hosting the Defra-specific logic away from client-specific
+// code.
+type sourceHubBridge struct {
+	client sourcehubClient
```
shahzadlone, line 94: todo:
```suggestion
	client sourceHubClient
```

shahzadlone (commented):
I wonder if we are over abstracting things by having both the `ACP` and `sourcehubClient` interfaces. This introduces a complexity which I am not sure we need.

From what I understand so far (correct me if missing something):
- The `ACP` interface was combining the "wrapping of sourcehub" + "defra specific acp logic"
- But rather than re-implementing the wrapping of sourcehub logic for remote acp and the defra specific acp logic you thought perhaps you can separate the sourcehub client to re-use those parts? Hence the current bridge and new interface?

- The `sourcehubClient` just seems like a lean wrapper extracting the functions we need from sourcehub. Couldn't this just be a struct? What other thing would implement it other than the `sourceHubBridge`

AndrewSisley:
> The ACP interface was combining the "wrapping of sourcehub" + "defra specific acp logic"

Not the ACP interface, the localACP implementation of that interface was.

> But rather than re-implementing the wrapping of sourcehub logic for remote acp and the defra specific acp logic you thought perhaps you can separate the sourcehub client to re-use those parts? Hence the current bridge and new interface?

If by `those parts`, you mean the defra specific logic, yes.  It is shared 'business logic' that should be the same regardless of what the underlying sourcehub/acp tech is.  Later it can be extracted further so that the wrapper does not only function for sourcehub lib based clients, but any acp client - however I don't want to code for the imaginary and over engineer something that wont get used as such for a while. 

> The sourcehubClient just seems like a lean wrapper extracting the functions we need from sourcehub. Couldn't this just be a struct? What other thing would implement it other than the sourceHubBridge

I think you are confused. `sourceHubBridge` implements `acp.ACP`.  `sourcehubClient` is implemented by `acpLocal` and `acpSourceHub`.

acp/source_hub_client.go
AndrewSisley: not a problem.

- [ ] rename

acp/source_hub_client.go
AndrewSisley: covered by earlier comment

acp/source_hub_client.go
AndrewSisley: covered by earlier comment

shahzadlone:
Thanks for the clarification, so to understand the full picture:
- `sourcehubClient` will be implemented by `acpLocal` and `acpRemote`. 
- `ACP` interface will continue to be the "sourcehub wrapper / bridge (local or global) + defra specfic acp logic"

Is that correct?

AndrewSisley:
> Is that correct?

Yes.  We shouldn't expose the internals (client vs defra logic spit) outside of the acp package.  You can see the source hub implementation of the client in the PR linked in this PR's description

shahzadlone:
Thanks for the explanation here and on Discord.

I think ACP interface should be a struct rather than an interface with the new changes, I don't like abstractions at this stage when we only have sourcehub acp that we need to worry about now, mostly for the sake of simplicity.

But the preference isn't a hard blocker however. So will give an approval, would be nice to hear from others too.

shahzadlone (approved):

AndrewSisley:
> I think ACP interface should be a struct rather than an interface with the new changes, 

I do strongly disagree with that.  `acp.ACP` is a public type and users should be free to provide their own.  The only implementation we have is very much an internal construct too, and we have plans to add more implementations.  We also really don't want to couple every referencing location to a concrete type - it makes changing the acp implementation much more expensive.

AndrewSisley:
> would be nice to hear from others too.

I'm very happy to continue this discussion post-merge, changing that after this goes in will probably cause far less problems than delaying this merge.  I would like to merge this before I start my weekend too (I don't want to have this PR in my head next week).

jsimnz:
Want to briefly chime in and say I agree with @AndrewSisley regarding the interface vs struct approach.

I'm not a huge fan of the internals of some of this PR regarding the client approach, but it's fine and wouldn't have caused me to block the merge anyway.

PR #2658, opened 2024-05-31
refactor: DAG sync and move merge outside of net package

### LINKED ISSUE #2624: Move post DAG-sync merge mechanics outside the `net` package.

fredcarle:
## Relevant issue(s)

Resolves #2624 

## Description

This PR  simplifies the DAG sync process withing the `net` package and moves the merge functionality to the `db` package. The merge is now initiated via an event channel.

Note: I did a search and replace for `SchemaVersionId` to `SchemaVersionID`. It's in its own commit. I've also remove the `tests/integration/net/order` tests as they are now annoying to maintain an will become even more irrelevant when we refactor the WaitForSync functionality of our test framework.

Another note: I've reduced the severity of the race condition on my Mac. We had a lot of leaking go routines and what is left of them is WaitForSync methods that sometimes seem to leak and also badger cache and libp2p transport that seem to leak go routines on close but I'm not sure how to handle these last two.

## Tasks

## How has this been tested?

make test

- MacOS


internal/db/merge.go
```diff
@@ -0,0 +1,361 @@
...
+	ctx, txn, err := ensureContextTxn(ctx, db, false)
+	if err != nil {
+		return err
+	}
+	defer txn.Discard(ctx)
+	mp, err := db.newMergeProcessor(ctx, dagMerge.Cid, dagMerge.SchemaRoot)
+	if err != nil {
+		return err
+	}
+	mt, err := mp.getHeads(ctx)
```
AndrewSisley, line 77: suggestion: The flow and concept-status of `mergeProcessor` could be improved I think.  At the moment it has the following problems:

1. It is only called in this function, and given that it's functions do not return anything at any stage, it is unclear why it is not wrapped up into a single func on `db`
2. Party because most of the functions do not return anything, it is unclear without reading the implementation of `mergeProcessor` as to whether the order in which each function is called matters or not, and if assuming it does matter, what that order needs to be.
3. It is strange that `getComposites` takes the return value of `getHeads`, when the host object is essentially a state-machine whose primary reason to be appears to manage state produced by it's host functions.
4. That `getComposites` does not return anything, despite being called `getFoo` confuses things further, and casts more doubt onto how many side affects each function has, making me want to read the implementation of `getHeads` to see if it also stores state in the state-machine as well as returning `mt`.
5. Partly because of the uncertainty around state-machineness and `mt`, and partly because it is quite dense and recursive code, I found it quite hard to be sure what `getComposites` was actually doing when reading it's implementation.

Overall I think `mergeProcessor` needs some work, although what kind of work is required I am not fully sure.  Maybe removing the semi-statemachineness and some more documentation would be enough.

AndrewSisley (changes requested):
I like it, and it is very definitely an improvement, but I think `mergeProcessor` needs some love before this can be merged.

internal/db/merge.go
```diff
@@ -0,0 +1,361 @@
...
+func (db *db) handleMerges(ctx context.Context, merges events.Subscription[events.DAGMerge]) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case merge, ok := <-merges:
+			if !ok {
+				return
+			}
+			go func() {
```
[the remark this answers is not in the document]
fredcarle: > This is very resource greedy and inefficient
I very much disagree. Go routines that are non blocking are very much ok. It will exit safely on its own when it's done processing the merge. Not doing this would make our merge process single threaded and super slow in heavy usage environments.  HTTP servers do exactly this when receiving request. One go routine per request.

The problems with the leaking go routines that we are seeing are caused by them requiring a signal to exit. Those can be more problematic if not managed properly.

internal/db/merge.go
fredcarle: I agree with this. I pushed it as is because I was tired of all the troubleshooting I did but I will improve it before merge.

internal/db/merge.go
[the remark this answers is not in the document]
AndrewSisley: > make our merge process single threaded

To clarify, I am very much not suggesting that we do this.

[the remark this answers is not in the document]
AndrewSisley: > > This is very resource greedy and inefficient
> 
> I very much disagree. 

At the moment, our user-space requests operate on a single routine (roughly).  Each request can spawn a large number of child routines for stuff that does not take place in user-time.  This means that operations taking place in user time are going to be down-prioritised by the machine, when in fact they are very likely the most time sensitive.

> The problems with the leaking go routines that we are seeing are caused by them requiring a signal to exit.

This is one of several problems, all of which are IMO caused by a culturally relaxed attitude towards them.

internal/db/merge.go
```diff
@@ -0,0 +1,364 @@
...
+		ctx,
+		client.CollectionFetchOptions{
+			SchemaRoot: immutable.Some(rootSchema),
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	col := cols[0].(*collection)
```
AndrewSisley, line 136: todo: Please document this `[0]`: why it is okay to do so now, why it is safe (and we are sure that there will be at least one), and why it might need to change in the future.

internal/db/merge.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	return &mergeProcessor{
+		ctx:    ctx,
+		txn:    txn,
+		ls:     ls,
+		docID:  docID,
+		mCRDTs: make(map[uint32]merklecrdt.MerkleCRDT),
+		col:    col,
+		schemaVersionKey: core.CollectionSchemaVersionKey{
```
AndrewSisley, line 149: suggestion: This is only consumed in `initCRDTForType` - it could be removed from `mergeProcessor` state and become a temporary variable within `initCRDTForType`

fredcarle: ✅

internal/db/merge.go
fredcarle: I added a comment. Let me know what you think.

nasdf (approved):
Really nice changes! Should make future network improvements much easier.

internal/db/merge.go
```diff
@@ -0,0 +1,418 @@
...
+	block *coreblock.Block,
+	blockLink cidlink.Link,
+) error {
+	crdt, err := mp.initCRDTForType(block.Delta.GetFieldName())
+	if err != nil {
+		return err
+	}
+
+	// If the CRDT is nil, it means the field is not part
+	// of the schema and we can safely ignore it.
```
AndrewSisley, line 250: question: When can this happen?  The only time I can think of is if the block was created on a higher schema version with a new field, and then we would still want the block synced - I cant remember if this was fully working, or if there is a ticket for handling the datastore, but I thought the blocks would still be synced (they affect the composite cid, so I don't think it is right to drop them even if we cant yet handle the datastore side).

AndrewSisley (approved):
Looks really good Fred - thanks a load for this :)  Just one more question for you before merge :)

internal/db/merge.go
AndrewSisley: LGTM, thanks :)

internal/db/merge.go
fredcarle: This is not the block sync process. That has already been taken care of in the `net` package. Here we merge the synced blocks into the datastore. Like you said, this can only happen if the block was created on a different schema version. If the node merge doesn't know about this field, it can't merge it into the datastore.

internal/db/merge.go
```diff
@@ -0,0 +1,418 @@
...
+		mCRDTs:     make(map[string]merklecrdt.MerkleCRDT),
+		col:        col,
+		dsKey:      dsKey,
+		composites: list.New(),
+	}, nil
+}
+
+type mergeTarget struct {
+	heads      map[cid.Cid]*coreblock.Block
+	headHeigth uint64
```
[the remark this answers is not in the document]
fredcarle: ✅

net/process.go
```diff
@@ -13,229 +13,146 @@
...
-		select {
-		case bp.sendJobs <- job:
-		case <-bp.ctx.Done():
-			return // jump out
+	for _, link := range links {
+		bp.queuedChildren.Remove(link)
+	}
+}
+
+type cidSafeSet struct {
```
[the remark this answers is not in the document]
fredcarle: Good idea. I made the change. I often forget that this exists now.

jsimnz (commented):
submitting now (will continue later today)

events/dag_sync.go
```diff
@@ -0,0 +1,31 @@
...
+
+// DAGMerge is a notification that a merge can be performed up to the provided CID.
+type DAGMerge struct {
+	// Cid is the id of the composite commit that formed this update in the DAG.
+	Cid cid.Cid
+	// SchemaRoot is the root identifier of the schema that defined the shape of the document that was updated.
+	SchemaRoot string
+	// MergeCompleteChan is a channel that will be closed when the merge is complete
+	// allowing the caller to optionnaly block until the merge is complete.
+	MergeCompleteChan chan struct{}
```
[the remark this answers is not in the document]
fredcarle: I was thinking of making it an error channel initially but ended up just using a struct because I think the merge process should be non blocking in the future. I agree that a `sync.Wait` would preferable here.

internal/db/merge.go
```diff
@@ -0,0 +1,364 @@
...
+
+	block, err := coreblock.GetFromNode(nd)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := mt.heads[blockCid]; ok {
+		// We've already processed this block.
+		return nil
+	}
```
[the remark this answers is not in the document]
fredcarle: I thought I had already fixed this. Thanks for flagging.

net/node.go
```diff
@@ -101,8 +102,11 @@ func NewNode(
 
 	fin := finalizer.NewFinalizer()
 
+	ctx, cancel := context.WithCancel(ctx)
+
 	peerstore, err := pstoreds.NewPeerstore(ctx, db.Peerstore(), pstoreds.DefaultOpts())
 	if err != nil {
+		cancel()
```
islamaliev, line 109: suggestion: instead of calling `cancel` on every return you could name the return args `(nodeResult *Node, err error)` and call
```go
defer func() {
  if resultNode == nil {
    cancel()
  }
}
```

And then in the end `nodeResult = n`

islamaliev (approved):
LGTM! just have couple questions and suggestions

internal/db/merge.go
```diff
@@ -0,0 +1,420 @@
...
+
+	if block.Delta.GetPriority() >= mt.headHeight {
+		mp.composites.PushFront(block)
+		for _, link := range block.Links {
+			if link.Name == core.HEAD {
```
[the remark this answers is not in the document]
fredcarle: Only the genesis block won't have a head link. If there is no head in the list of links, it just means that we've reached the genesis block and we can exit from there.

net/node.go
fredcarle: I try to avoid named return variables as much as possible and I don't mind the explicitness of calling cancel wherever it's needed. I'll think about it while I wait the the rest of John's review.

islamaliev: I also don't like named returns, but this method has quite some occurrences of `return`. So another person might add some other error check later and forget to add `cancel`.

Issue #2577, opened 2024-05-02
Join code is difficult to read


AndrewSisley:
Please either rework it a little bit, or go through it very carefully and document it so that maintainers can understand why it is doing what it is doing.  Be careful as it is quite a complex area, and different bits seem to represent different things depending on context.

Review the tests before starting any code change, we've had a few bugs here recently, and historically we have had quite poor coverage here.

PR #2670, opened 2024-06-03
feat: Sec. indexes on relations

### LINKED ISSUE #2601: Sec. index: enable indexing of related objects
At the moment with requests like:
```graphql
query {
  Device(filter: {
    owner: {name: {_eq: "Keenan"}}
  }) {
    model
  }
}
```
and schema:
```graphql
type User {
  name: String @index
  devices: [Device]
} 

type Device {
  model: String 
  owner: User
}
```
the index will find and fetch the owner right away but having no reference to secondary objects (devices) we have to do full scan of devices to see which of them have corresponding `owner_id`.

The use cases from the following (potentially others as well) tests suffer from not utilizing indexes on related objects.
`TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelation_ShouldFilter`
`TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedRelationWhileIndexedForeignField_ShouldFilter`
`TestQueryWithIndexOnOneToMany_IfFilterOnIndexedRelation_ShouldFilterWithExplain`
`TestQueryWithIndexOnOneToOne_IfFilterOnIndexedRelation_ShouldFilter`
`TestQueryWithIndexOnManyToOne_IfFilterOnIndexedRelation_ShouldFilterWithExplain`

This task potentially covers #2578

islamaliev:
## Relevant issue(s)

Resolves #2601 #2578 #2577

## Description

Enables fetching related objects via secondary indexes.

It also fixes a bug with queries that contain multiple aggregates on the same collection.

Note: I will add some more tests

### DIFF
--- a/internal/planner/explain.go
+++ b/internal/planner/explain.go
@@ -92,8 +92,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 		var explainGraphBuilder = map[string]any{}
 
 		// If root is not the last child then keep walking and explaining the root graph.
-		if node.root != nil {
-			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.root)
+		if node.parentPlan != nil {
+			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentPlan)
 			if err != nil {
 				return nil, err
 			}
@@ -101,8 +101,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 			explainGraphBuilder[joinRootLabel] = indexJoinRootExplainGraph
 		}
 
-		if node.subType != nil {
-			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.subType)
+		if node.childPlan != nil {
+			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childPlan)
 			if err != nil {
 				return nil, err
 			}
@@ -117,8 +117,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 		var explainGraphBuilder = map[string]any{}
 
 		// If root is not the last child then keep walking and explaining the root graph.
-		if node.root != nil {
-			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.root)
+		if node.parentPlan != nil {
+			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentPlan)
 			if err != nil {
 				return nil, err
 			}
@@ -128,8 +128,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 			explainGraphBuilder[joinRootLabel] = nil
 		}
 
-		if node.subType != nil {
-			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.subType)
+		if node.childPlan != nil {
+			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childPlan)
 			if err != nil {
 				return nil, err
 			}
--- a/internal/planner/explain.go
+++ b/internal/planner/explain.go
@@ -92,8 +92,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 		var explainGraphBuilder = map[string]any{}
 
 		// If root is not the last child then keep walking and explaining the root graph.
-		if node.parentPlan != nil {
-			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentPlan)
+		if node.parentSide.plan != nil {
+			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentSide.plan)
 			if err != nil {
 				return nil, err
 			}
@@ -101,8 +101,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 			explainGraphBuilder[joinRootLabel] = indexJoinRootExplainGraph
 		}
 
-		if node.childPlan != nil {
-			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childPlan)
+		if node.childSide.plan != nil {
+			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childSide.plan)
 			if err != nil {
 				return nil, err
 			}
@@ -117,8 +117,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 		var explainGraphBuilder = map[string]any{}
 
 		// If root is not the last child then keep walking and explaining the root graph.
-		if node.parentPlan != nil {
-			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentPlan)
+		if node.parentSide.plan != nil {
+			indexJoinRootExplainGraph, err := buildDebugExplainGraph(node.parentSide.plan)
 			if err != nil {
 				return nil, err
 			}
@@ -128,8 +128,8 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
 			explainGraphBuilder[joinRootLabel] = nil
 		}
 
-		if node.childPlan != nil {
-			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childPlan)
+		if node.childSide.plan != nil {
+			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.childSide.plan)
 			if err != nil {
 				return nil, err
 			}
--- a/internal/planner/mapper/mapper.go
+++ b/internal/planner/mapper/mapper.go
@@ -987,6 +987,7 @@ func resolveInnerFilterDependencies(
 				return nil, err
 			}
 
+			childSelect.SkipResolve = true
 			newFields = append(newFields, childSelect)
 		}
 
--- a/internal/planner/mapper/mapper.go
+++ b/internal/planner/mapper/mapper.go
@@ -413,25 +413,23 @@ func resolveAggregates(
 				childMapping = childMapping.CloneWithoutRender()
 				mapping.SetChildAt(index, childMapping)
 
-				if !childIsMapped {
-					filterDependencies, err := resolveFilterDependencies(
-						ctx,
-						store,
-						rootSelectType,
-						childCollectionName,
-						target.filter,
-						mapping.ChildMappings[index],
-						childFields,
-					)
-					if err != nil {
-						return nil, err
-					}
-					childFields = append(childFields, filterDependencies...)
-
-					// If the child was not mapped, the filter will not have been converted yet
-					// so we must do that now.
-					convertedFilter = ToFilter(target.filter.Value(), mapping.ChildMappings[index])
+				filterDependencies, err := resolveFilterDependencies(
+					ctx,
+					store,
+					rootSelectType,
+					childCollectionName,
+					target.filter,
+					mapping.ChildMappings[index],
+					childFields,
+				)
+				if err != nil {
+					return nil, err
 				}
+				childFields = append(childFields, filterDependencies...)
+
+				// If the child was not mapped, the filter will not have been converted yet
+				// so we must do that now.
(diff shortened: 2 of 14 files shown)


internal/planner/mapper/mapper.go
```diff
@@ -413,25 +413,23 @@ func resolveAggregates(
 				childMapping = childMapping.CloneWithoutRender()
 				mapping.SetChildAt(index, childMapping)
 
-				if !childIsMapped {
```
islamaliev, line 416: this was a nasty bug that took many hours to spot and fix.
It because visible for `TestOneToManyToOneWithSumOfDeepFilterSubTypeOfBothDescAndAsc` test where 2 aggregates are used on the same related object:
```graphql
query {
    Author {
        name
        s1: _sum(book: {field: rating, filter: {publisher: {yearOpened: {_eq: 2013}}}})
        s2: _sum(book: {field: rating, filter: {publisher: {yearOpened: {_ge: 2020}}}})
    }
}
```
After the first aggregate is processed with correct `[]Requestable` fields, the second would reuse the results of the frst which doesn't include resolving of fields. So `Fields` field would have a `nil` value. Which happened to be interpreted by the fetcher as all-fields-requested.
So when dealing with secondary indexes I explicitly added to the `fields` array a relation id field (like `book_id`). That made the fetcher to fetch only the related id field and would prevent the filter from letting `yearOpened: {_ge: 2020}` condition to pass, because it's not fetched.

That's why filter dependencies should be resolved for every aggregate.

Let me know if you think there is a better solution.

internal/planner/select.go
```diff
@@ -315,6 +315,19 @@ func findIndexByFilteringField(scanNode *scanNode) immutable.Option[client.Index
...
 }
 
+func findIndexByFieldName(col client.Collection, fieldName string) immutable.Option[client.IndexDescription] {
+	for _, field := range col.Schema().Fields {
+		if field.Name != fieldName {
+			continue
+		}
+		indexes := col.Description().GetIndexesOnField(field.Name)
+		if len(indexes) > 0 {
+			return immutable.Some(indexes[0])
```
AndrewSisley, line 325: suggestion: I think I understand why `0` is not a problem and that all we risk is it being potentially less optimised than it could be in the fetcher (e.g. if the `0` index is a composite of 3, and the `1` index is a simple single field index), but I think it is worth a comment here to highlight/confirm/reject this thought :)

internal/planner/type_join.go
```diff
@@ -568,54 +666,90 @@ func (join *invertibleTypeJoin) Next() (bool, error) {
...
+
+	return true, nil
+}
+
+func (join *invertibleTypeJoin) nextJoinedSecondaryDoc() (bool, error) {
+	firstSide := join.getFirstSide()
+	secondSide := join.getSecondSide()
+
+	secondaryDocID := getForeignKey(firstSide.plan, firstSide.relFieldDef.Name)
+	// TODO: add some tests with filter on nil relation
```
AndrewSisley, line 698: todo: Mentioned in the description, just adding a comment so I don't forget (if you do :)) (I assume you are resolving this before merge)

internal/planner/type_join.go
```diff
@@ -568,54 +666,90 @@ func (join *invertibleTypeJoin) Next() (bool, error) {
...
+		for i := range join.encounteredDocIDs {
+			if join.encounteredDocIDs[i] == secondaryDocID {
+				return join.Next()
+			}
 		}
+		join.encounteredDocIDs = append(join.encounteredDocIDs, secondaryDocID)
 	}
 
-	join.docsToYield = append(join.docsToYield, join.root.Value())
+	// check if there can ever be false (a.k.a. hasDoc = false)
```
AndrewSisley, line 716: question: This comment looks like a temp dev-todo, is it done, should it be removed?

internal/planner/mapper/select.go
```diff
@@ -38,6 +38,10 @@ type Select struct {
 	// These can include stuff such as version information, aggregates, and other
 	// Selects.
 	Fields []Requestable
+
+	// SkipResolve is a flag that indicates that the fields in this Select don't need to be resolved.
+	// It is used to avoid resolving related objects if they are used only in a filter and not requested in a response.
+	SkipResolve bool
```
AndrewSisley, line 44: suggestion: It took a bit of head-scratching and reading to be sure, but I think I understand what you mean by resolve - however the term (within this context) was new to me, and the comment doesn't define what it means.

I think adding an explanation of that to this comment could be quite handy in the future.

internal/planner/type_join.go
```diff
@@ -452,108 +452,206 @@ func fetchPrimaryDoc(node, subNode planNode, parentProp string) (bool, error) {
...
 }
 
 func (join *invertibleTypeJoin) Spans(spans core.Spans) {
-	join.root.Spans(spans)
+	join.parentSide.plan.Spans(spans)
+}
+
+func (join *invertibleTypeJoin) Source() planNode { return join.parentSide.plan }
+
+type primaryObjectsFetcher struct {
```
AndrewSisley, line 502: thought: I have mixed feelings about this being called `fetcher`, but not strong enough to request a change, flagging here in case we want a discussion or if others have stronger thoughts.

tests/integration/index/query_with_compound_filter_relation_test.go
```diff
@@ -164,6 +164,7 @@ func TestIndex_QueryWithIndexOnOneToManyRelationAndFilter_Data(t *testing.T) {
 			},
 			testUtils.CreateDoc{
 				CollectionID: 0,
+				// TODO: Ask Andy of the docID is intentionally wrong (doesn't exist)
```
AndrewSisley, line 167: answer: It shouldn't be wrong, it should be the doc id of `ESA` - sorry if it is wrong, please correct it if it is :sweat_smile:

AndrewSisley (commented):
praise: Looks really really good Islam, and I can really imagine how much of a pain that must have been to find and fix.  Fantastic job.

I gave the production code a look over, and wanted to submit before the standup in case I get distracted etc.  No major suggestions, will likely be a quick approval once I finish my review.

Thanks a bunch for this, and sorry about the pain you must have gone through in the mapper :)

internal/planner/type_join.go
islamaliev: I have the same feelings. It used to be called "....Joiner" but I extracted join functionality out of this struct, so it is left with fetching-related functionality.
Some suggestions: `primaryObjectsFetchCoordinator`, `primaryObjectsRetriever`
Do you have preferences or suggestions?

internal/planner/mapper/select.go
islamaliev: I thought it was a common term in graphql space which refers to a resolver that is supposed to set a value for a requested object field.
If you have a better suggestion, let me know please. Otherwise I will just extend the documentation.

internal/planner/type_join.go
AndrewSisley: Out of those two I think I prefer `primaryObjectsRetriever`, although it does feel like a slightly force synonym of `fetcher` :)  Still no strong feelings from me here though.

internal/planner/mapper/select.go
AndrewSisley: I'm happy with the current name, documentation would be nice though - thanks :)

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -17,13 +17,15 @@ import (
 )
 
 func TestQueryWithIndexOnOneToManyRelation_IfFilterOnIndexedRelation_ShouldFilter2(t *testing.T) {
+	// 3 users have a MacBook Pro: Islam, Shahzad, Keenan
```
AndrewSisley, line 20: praise: Thanks for these, is useful info :)

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -210,14 +208,87 @@ func TestQueryWithIndexOnOneToOnesSecondaryRelation_IfFilterOnIndexedRelation_Sh
...
+					{"name": "John"},
+					{"name": "Fred"},
+					{"name": "Shahzad"},
+				},
+			},
+			testUtils.Request{
+				Request: makeExplainQuery(req2),
+				// we make 3 index fetches to get the 3 address with city == "Montreal"
+				// and 3 more index fetches to get the corresponding users
+				// then 3 field fetches to get the name of each user
```
AndrewSisley, line 273: praise: These are great - thanks Islam!

AndrewSisley (approved):
LGTM - thanks Islam, it looks really good.  Approving now but please resolve/argue against all the minor comments from me before merge :)

internal/planner/type_join.go
```diff
@@ -452,31 +447,34 @@ func fetchPrimaryDoc(node, subNode planNode, parentProp string) (bool, error) {
...
+	// secondNode is the node that is executed second.
+	secondNode planNode // Device
+	// subRelField is a field name of a secondary doc that refers to the primary docID (like author_id).
+	subRelField string // owner_id
+	// topRelField is a field name of the primary doc that refers to the secondary docID (like author_id).
+	topRelField string // devices_id
+	// isInvertable indicates if the join can be inverted.
+	isInvertable bool
+	// isInverted indicates if the join direction is inverted.
+	isInverted bool
```
shahzadlone: praise: Thanks for the documentation

internal/planner/explain.go
```diff
@@ -92,17 +92,17 @@ func buildDebugExplainGraph(source planNode) (map[string]any, error) {
...
 			if err != nil {
 				return nil, err
 			}
 			// Add the explaination of the rest of the explain graph under the "root" graph.
 			explainGraphBuilder[joinRootLabel] = indexJoinRootExplainGraph
 		}
 
-		if node.subType != nil {
-			indexJoinSubTypeExplainGraph, err := buildDebugExplainGraph(node.subType)
+		if node.childSide.plan != nil {
```
shahzadlone, line 104: praise: I like these!

internal/planner/type_join.go
```diff
@@ -147,45 +147,41 @@ func (n *typeIndexJoin) simpleExplain() (map[string]any, error) {
...
-			simpleExplainMap[joinDirectionLabel] = joinDirectionSecondaryLabel
-		} else {
-			simpleExplainMap[joinDirectionLabel] = joinDirectionPrimaryLabel
-		}
-
+	addExplainData := func(j *invertibleTypeJoin) error {
 		// Add the attribute(s).
-		simpleExplainMap[joinRootLabel] = joinType.rootName
-		simpleExplainMap[joinSubTypeNameLabel] = joinType.subTypeName
+		simpleExplainMap[joinRootLabel] = immutable.Some(j.childSide.relFieldDef.Name)
```
shahzadlone, line 152: question: Instead of a string, this will now be an optional type. Does the caller that receives the result of `simpleExplain()` handle this by checking `HasValue` idk why I missed seeing that check. or does it rely on the default string value being `""` if it is missing? another question, can it be missing?

shahzadlone (approved):
Looks really good :)

internal/planner/select.go
islamaliev: done

internal/planner/type_join.go
islamaliev: removed

internal/planner/mapper/select.go
islamaliev: done

internal/planner/type_join.go
islamaliev: added tests

internal/planner/type_join.go
islamaliev: renamed

internal/planner/type_join.go
islamaliev: `rootName` was also `immutable.Option`

shahzadlone:
Bug bash: got a crash course on this PR and tested some edge case tests (like `TestQueryWithIndexOnOneToMany_IfFilterOnIndexedRelation_ShouldFilterWithExplain`) with more fields indexed. LGTM

PR #2682, opened 2024-06-04
test: Add relation substitute mechanic to tests

### LINKED ISSUE #2681: Introduce substitution mechanic for relations in tests

AndrewSisley:
## Relevant issue(s)

Resolves #2681

## Description

Adds a relation substitute mechanic to tests, allowing us to not worry about doc ids in tests that don't need to care about the exact string.

Only converts a handful of tests to the new system, I don't think we should actively spend large chunks of time migrating to the new system - we can just write new tests using this where appropriate, and passively convert existing when convenient.


fredcarle (approved):
LGTM. This will be convenient :)

PR #2686, opened 2024-06-04
refactor: Replace subscription events publisher

### LINKED ISSUE #2685: Use event system for subsystem events only
The event system should be used for subsystem events only. This will make debugging and testing of the interactions between different subsystems much simpler. This also means that subscriptions will need a refactor so that they do not require an event subscription channel.

nasdf:
## Relevant issue(s)

Resolves #2685 

## Description

This PR replaces subscription event publishers with a simple go channel. This is a pre-requisite to a follow up events package refactor.

## Tasks

## How has this been tested?

`make test`

- MacOS

### DIFF
--- a/client/db.go
+++ b/client/db.go
@@ -265,9 +265,9 @@ type RequestResult struct {
 	// GQL contains the immediate results of the GQL request.
 	GQL GQLResult
 
-	// Pub contains a pointer to an event stream which channels any subscription results
-	// if the request was a GQL subscription.
-	Pub *events.Publisher[events.Update]
+	// Subscription is an optional channel which returns results
+	// from a subscription request.
+	Subscription <-chan GQLResult
 }
 
 // CollectionFetchOptions represents a set of options used for fetching collections.
--- a/internal/db/subscriptions.go
+++ b/internal/db/subscriptions.go
@@ -34,6 +34,8 @@ func (db *db) handleSubscription(ctx context.Context, r *request.Request) (<-cha
 	if !ok {
 		return nil, client.NewErrUnexpectedType[request.ObjectSubscription]("SubscriptionSelection", selections)
 	}
+	// unsubscribing from this publisher will cause a race condition
+	// https://github.com/sourcenetwork/defradb/issues/2687
 	pub, err := events.NewPublisher(db.events.Updates.Value(), 5)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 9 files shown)


client/db.go
```diff
@@ -265,9 +265,9 @@ type RequestResult struct {
 	// GQL contains the immediate results of the GQL request.
 	GQL GQLResult
 
-	// Pub contains a pointer to an event stream which channels any subscription results
-	// if the request was a GQL subscription.
-	Pub *events.Publisher[events.Update]
```
AndrewSisley, line 270: discussion: I think you are removing functionality here - IIRC the events package stuff allows for multiple concurrent readers of the stream.  The simple Go chan that you are replacing it lacks this functionality and will mean any concurrent readers will be competing with each other for each item.

The CLI/http clients are unaffected, but the embedded Go client is.

nasdf: It makes more sense to me that we assume one reader per request by default. The consumer of the channel then has the option to decide if they want to create an `events.Publisher` and publish the channel results to it.

internal/db/subscriptions.go
```diff
@@ -19,83 +19,75 @@ import (
...
-		}
-
-		return pub, subRequest, nil
+	selections := r.Subscription[0].Selections[0]
+	subRequest, ok := selections.(*request.ObjectSubscription)
+	if !ok {
+		return nil, client.NewErrUnexpectedType[request.ObjectSubscription]("SubscriptionSelection", selections)
+	}
+	// unsubscribing from this publisher will cause a race condition
+	// https://github.com/sourcenetwork/defradb/issues/2687
```
AndrewSisley, line 38: question: Is this a new issue, or was the old code affected to?

And where is the race condition?  Is it a test artefact, or production code?

nasdf: I believe this is an old issue. The previous code never called `Unsubscribe`. The race happens between the `db.Close` and the `pub.Unsubscribe` so it's possible to happen in production, but likely would cause no issues.

AndrewSisley (commented):
The code is easy to read, but I have a todo/question RE timeouts before I can approve :)

internal/db/subscriptions.go
```diff
@@ -19,83 +19,75 @@ import (
...
-			log.ErrorContext(ctx, err.Error())
-			continue
-		}
+		// listen for events and send to the result channel
+		for {
+			var evt events.Update
+			select {
+			case val := <-pub.Event():
+				evt = val
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
nasdf: Good catch! I've updated it to handle the channel closing.

AndrewSisley (approved):
LGTM, I'm curious as to what is coming next :)

PR #2547, opened 2024-04-23
feat: Update corelog to 0.0.7

nasdf:
## Relevant issue(s)

Resolves #2546

## Description

This PR updates `corelog` to version 0.0.7. It also adds the `log.nocolor` config.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

`make test`

Specify the platform(s) on which this was tested:
- MacOS

Issue #615, opened 2022-07-09
Panic reduction


orpheuslummis:
A tracking issue for minimizing panic.

The risk with `panic()` is that it leads to runtime crash.

Acceptable kinds of panic:
- In initialization as they are ran as part of process startup (AFAIK) and not expected to induce runtime crashes.
- When intentional, internal, and handled by `recover()`
- Crash intentionally when facing some kinds of "exceptional situations" 

Paths forward:
- convert to error
- avoid situation

Here are the current points calling `panic()` (as of commit 16084c89e3 ...), excluding tests.
I'm marking each as a task that can be marked off with a decision/change.

- [x]  https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/client/document.go#L397 and https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/client/document.go#L402
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/db/db.go#L224
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/db/container/container.go#L43
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/db/container/container.go#L67
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/net/utils/util.go#L31
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/core/data.go#L54 (although it is a marked as TODO)
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/core/data.go#L59 (although it is a marked as TODO)
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/logging/logger.go#L35 (might be idiomatic to panic like this, indicated by the use of `must`)
- [x] https://github.com/sourcenetwork/defradb/blob/16084c89e3421fabf3ab53bcd84a3804f2b27693/core/net/protocol.go#L42 (as part of init so IMO is fine)
[...]

PR #2691, opened 2024-06-06
refactor: Change local_acp implementation to use acp_core

### LINKED ISSUE #2694: Fix Defra's WASM compilation with the ACP Local feature
Currently, SourceHub is incompatible with WASM Compilation. Consequently, the introduction of SourceHub as a general dependency of Defra broke the WASM build. 

The proposed solution to this problem is to remove SourceHub as a dependency in the context of ACP Local and replace it with ACP Core

**To Reproduce**
Compile Defra for the WASM / js platform. The build should fail due to SourceHub / cosmos sdk dependencies.

**Expected behavior**
Defra should compile with WASM / JS Architecture and platform

**Platform**
n/a

**Additional context**
Related to #2440

Lodek:
## Relevant issue(s)

Also the related issue won't close upon merge, nor is it picked up by github at the moment, because you just have #2694 in the PR description.

So please don't forget to add the resolving key word (close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved) with it.

So you can put for example:

Resolve #2694

## Description

This PR changes the underlying implementation of the `ACPLocal` Reference Monitor from SourceHub to ACP Core.
The changes aim to allow the WASM build which was broken since the inclusion of SourceHub as a dependency.

Notable changes from this refactor include:
- Replace SourceHub as a dependency for acp_core for ACP Local 
- The field `name` in a `Policy` is now correctly marked as required. This was a consequence of an issue on SourceHub side which allowed unnamed policies to be included. Therefore, in order to correct the tests, a name was given to all policies in the codebase.
- The previous implementation of ACP Local made use of a special notion of account which had no sequence numbers. As such, re-submitting a Policy generated the same ID. This behavior isn't compatible with how SourceHub deals with IDs and the new implementation using ACP Core reflects that. That is, every created Policy, independent of its payload, will have unique IDs. 
- Updated the `Identity` struct to generate a DID instead of a SourceHub address

## Tasks

- [] I made sure the repository-held documentation is changed accordingly.

## How has this been tested?

Test suite ie `make test`

- Arch Linux


acp/acp_local.go
```diff
@@ -36,6 +37,7 @@ var _ sourceHubClient = (*ACPLocal)(nil)
 var _ Policy = (*localACPPolicyAdapter)(nil)
 var _ Resource = (*localACPResourceAdapter)(nil)
 var _ Permission = (*types.Permission)(nil)
+var errGeneratingDIDFromNonAccAddr = errors.New("cannot generate did if address is not prefixed")
```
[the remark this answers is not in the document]
AndrewSisley: With this Bruno, in the Defra repo we have effectively banned the use of the std `errors` package - instead we use the `errors` package defined in the repo root - from dedicated `errors.go` files in each package (`acp/errors.go` is a good example).

acp/acp_local.go
```diff
@@ -152,26 +239,56 @@ func (l *ACPLocal) VerifyAccessRequest(
...
-			AccessRequest: &types.AccessRequest{
-				Operations: []*types.Operation{
-					{
-						Object:     types.NewObject(resourceName, docID),
-						Permission: permission.String(),
-					},
-				},
-				Actor: &types.Actor{
-					Id: actorID,
+	// FIXME remove once Identity is refactored
```
AndrewSisley, line 242: todo: @Lodek any todos merged into the Defra repo must be linked to a github issue.  If/when creating one for this, please make sure it contains more info than this comment currently does :)  Atm it is unclear as to why it is temporary/can be removed after the refactor.

acp/acp_local.go
```diff
@@ -36,73 +81,94 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
 		)
+
 		if err != nil {
 			return NewErrInitializationOfACPFailed(err, "Local", l.pathToStore.Value())
 		}
 	}
 
-	l.localACP = &localACP
+	engine := engine.NewACPEngine(manager)
+	l.engine = engine
```
AndrewSisley, line 107: nitpick: `l.engine = engine.NewACPEngine(manager)` :)

acp/acp_local.go
```diff
@@ -36,73 +81,94 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
 
 func (l *ACPLocal) Start(ctx context.Context) error {
-	var localACP embedded.LocalACP
+	var manager runtime.RuntimeManager
 	var err error
 
 	if !l.pathToStore.HasValue() { // Use a non-persistent, i.e. in memory store.
-		localACP, err = embedded.NewLocalACP(
-			embedded.WithInMemStore(),
+		manager, err = runtime.NewRuntimeManager(
```
AndrewSisley, line 88: suggestion: I think the code would be a bit more readable if the `if-else` just handles the construction/appending of `runtime.Opt`s, and the `runtime.NewRuntimeManager` call is moved outside of the `if-else` (and the err handling).

AndrewSisley:
todo: Please sort out the PR description :)

Adding info regarding why this is being done would be/have-been particularly useful. And please make sure it is linked to a Defra repo github issue.

acp/source_hub_client.go
```diff
@@ -120,17 +117,10 @@ func (a *sourceHubBridge) AddPolicy(ctx context.Context, creatorID string, polic
 		return "", ErrPolicyDataMustNotBeEmpty
 	}
 
-	// Assume policy is in YAML format by default.
-	policyMarshalType := types.PolicyMarshalingType_SHORT_YAML
-	if isJSON := fastjson.Validate(policy) == nil; isJSON { // Detect JSON format.
-		policyMarshalType = types.PolicyMarshalingType_SHORT_JSON
```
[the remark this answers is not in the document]
shahzadlone: never mind I see this moved else where.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sourcenetwork/defradb
 
-go 1.21.3
+go 1.21.5
+
+toolchain go1.22.2
```
AndrewSisley, line 5: todo: Why has this been added?

acp/source_hub_client.go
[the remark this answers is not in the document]
Lodek: :saluting_face:

acp/acp_local.go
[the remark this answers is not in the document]
Lodek: :saluting_face:

replacement.py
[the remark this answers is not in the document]
Lodek: They shouldn't be part of the final PR. That's just a hacky script I wrote because I didn't want to relearn awk. I can move them to tools/ but I don't think it's very valuable. Should I?

tests/integration/acp/add_policy/basic_test.go
```diff
@@ -65,6 +66,7 @@ func TestACP_AddPolicy_BasicJSON_ValidPolicyID(t *testing.T) {
 
 				Policy: `
 					{
+					  "name": "test",
```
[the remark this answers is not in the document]
Lodek: It's regarding the issue you reported that an empty policy was being accepted. It's been added to the PR description

tests/integration/acp/add_policy/with_multi_policies_test.go
```diff
@@ -278,15 +278,15 @@ func TestACP_AddPolicy_AddMultipleDuplicatePolicies_Error(t *testing.T) {
...
+				ExpectedPolicyID: "ec02815cb630850678bda5e2d75cfacebc96f5610e32a602f7bfc414e21474ad",
 			},
 		},
 	}
 
 	testUtils.ExecuteTestCase(t, test)
 }
 
-func TestACP_AddPolicy_AddMultipleDuplicatePoliciesDifferentFmts_Error(t *testing.T) {
+func TestACP_AddPolicy_AddMultipleDuplicatePoliciesDifferentFmts_ProducesDifferentIDs(t *testing.T) {
```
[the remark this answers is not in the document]
Lodek: :saluting_face:

tests/integration/acp/add_policy/with_no_resources_test.go
```diff
@@ -74,7 +74,7 @@ func TestACP_AddPolicy_NoResourceLabel_ValidID(t *testing.T) {
 
 // Eventhough empty resources make no sense from a DefraDB (DPI) perspective,
 // it is still a valid sourcehub policy for now.
```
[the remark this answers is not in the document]
Lodek: :saluting_face:

acp/acp_local.go
Lodek: :saluting_face:

acp/acp_local.go
Lodek: good suggestion, thank you

go.mod
Lodek: Honestly, no idea. `go mod tidy` changed it. I'm not too familiar with the purpose of the `toolchain` directive. Has this happened before? I can look into or manually remove it

Lodek:
Is there any action required on my part regarding the `Check Vulnerabilities Workflow`?

AndrewSisley:
> Is there any action required on my part regarding the `Check Vulnerabilities Workflow`?

It looks like the vulnerability only exists in the runtime/toolchain version you tried to update to - so if you sort out the go.mod that job should pass anyway

go.mod
AndrewSisley: Ah fair enough, I'd suggest just reverting this and the `go` version then :)

acp/acp_local.go
Lodek: :saluting_face: Is there an issue for the work you were doing on refactoring the Identity to accept the DID form the JWS Shahzad?

go.mod
Lodek: I removed the toolchain directive, which fixed the vulnerability workflow. I left the minor bump as is, however I can revert it as well if that's preferred

AndrewSisley: I would prefer that we don't bump the minor version in random PRs.  Although I'm not really sure why we even specify the minor version.

acp/acp_local.go
shahzadlone: We can sync up on a call. The JWS stuff was done already you would just need to generate dids and plug them in now.

go.mod
shahzadlone: > I would prefer that we don't bump the minor version in random PRs. Although I'm not really sure why we even specify the minor version - might be a historical hangover that we havent yet removed from a previous version where we really wanted a minor version feature.

+1, also our `mod tidy` make rule (`make tidy`) ensures the modules are "clean" up to a minimum version. This was due to some pains that were encountered when a `mod tidy` command breaks and expects a minimum patch version. For example when the commit `c42e7ee6 feat: Update corelog to 0.0.7 (#2547)` made to develop it broke the go mod file. 

tldr: if you want to bump the version it should also update the `make tidy` Makefile rule and I would highly suggest to do this in a separate PR if you don't need the version bump here.

This reminds me, we should add a ci check to ensure our `make tidy` is never broken, made issue here: https://github.com/sourcenetwork/defradb/issues/2695

replacement.py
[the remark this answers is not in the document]
shahzadlone: na don't worry about it. I think I was reviewing commit by commit so missed you removed them. Cheers

acp/acp_local.go
```diff
@@ -36,73 +79,91 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
+	// FIXME remove once Identity is refactored.
+	// See genDIDFromSourceHubAddr docs and issue ###### for context
+	did, err := genDIDFromSourceHubAddr(creatorID)
+	if err != nil {
+		return "", err
+	}
+
+	principal, err := auth.NewDIDPrincipal(did)
+	if err != nil {
+		return "", newErrInvalidActorID(err, creatorID)
```
shahzadlone, line 125: suggestion: This case should be tested, but I think it needs the `did` hack to not be a hack :/

acp/acp_local.go
```diff
@@ -112,36 +173,58 @@ func (l *ACPLocal) RegisterObject(
...
-			CreationTime: creationTime,
-		},
-	)
+) (RegistrationResult, error) {
+	// FIXME remove once Identity is refactored
+	// See genDIDFromSourceHubAddr docs and issue ###### for context
+	did, err := genDIDFromSourceHubAddr(actorID)
 	if err != nil {
-		return types.RegistrationResult(0), err
+		return RegistrationResult_NoOp, err
```
shahzadlone, line 181: suggestion: similar to above this should be tested but we need to sort out the `did` hack

go.mod
Lodek: :saluting_face:

acp/acp_local_test.go
```diff
@@ -652,3 +654,46 @@ func Test_LocalACP_PersistentMemory_CheckDocAccess_TrueIfHaveAccessFalseIfNotErr
...
+		invalidIdentity,
+		validPolicy,
+	)
+
+	require.ErrorIs(t, err, ErrInvalidActorID)
+	require.Empty(t, policyID)
+
+	err = localACP.Close()
+	require.NoError(t, err)
+}
```
shahzadlone, line 677: todo: Add a counterpart test after this test like: `Test_LocalACP_PersistentMemory_AddPolicy_InvalidCreatorIDReturnsError` to test the same with persistence.

acp/acp_local_test.go
```diff
@@ -652,3 +654,46 @@ func Test_LocalACP_PersistentMemory_CheckDocAccess_TrueIfHaveAccessFalseIfNotErr
...
+		validPolicyID,
+		"users",
+		"documentID_XYZ",
+	)
+
+	require.ErrorIs(t, err, ErrInvalidActorID)
+
+	err = localACP.Close()
+	require.NoError(t, err)
+}
```
shahzadlone, line 699: todo: Same as above comment but add a counter-part test after this test.

acp/identity/identity.go
```diff
@@ -45,14 +46,20 @@ func FromPrivateKey(privateKey *secp256k1.PrivateKey) immutable.Option[Identity]
...
-// AddressFromPublicKey returns the unique address of the given public key.
-func AddressFromPublicKey(publicKey *secp256k1.PublicKey) string {
-	pub := cosmosSecp256k1.PubKey{Key: publicKey.SerializeCompressed()}
-	// conversion from well known types should never cause a panic
-	return types.MustBech32ifyAddressBytes("cosmos", pub.Address().Bytes())
+// DIDFromPublicKey returns the unique address of the given public key.
+func DIDFromPublicKey(publicKey *secp256k1.PublicKey) string {
+	bytes := publicKey.SerializeUncompressed()
+	did, err := key.CreateDIDKey(crypto.SECP256k1, bytes)
+	// because we are generating a did from a a supported and fixed key type,
```
shahzadlone, line 58:
```suggestion
	// because we are generating a `did` from a supported and fixed key type,
```

acp/identity/identity.go
```diff
@@ -45,14 +46,20 @@ func FromPrivateKey(privateKey *secp256k1.PrivateKey) immutable.Option[Identity]
...
+	// this operation is infallible
+	if err != nil {
+		panic(err)
+	}
```
shahzadlone, line 62: question: why do we have a panic here? instead of returning an error? also document that in a comment please.

acp/identity/identity_test.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func Test_DIDFromPublicKey_ProducesDIDForPublicKey(t *testing.T) {
+	pubKey := &secp256k1.PublicKey{}
+
+	did := DIDFromPublicKey(pubKey)
+
+	want := "did:key:z7r8ooUiNXK8TT8Xjg1EWStR2ZdfxbzVfvGWbA2FjmzcnmDxz71QkP1Er8PP3zyLZpBLVgaXbZPGJPS4ppXJDPRcqrx4F"
+	require.Equal(t, want, did)
+}
```
shahzadlone, line 27: praise: thanks for adding this test.

acp/identity/identity_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_DIDFromPublicKey_ProducesDIDForPublicKey(t *testing.T) {
+	pubKey := &secp256k1.PublicKey{}
```
shahzadlone, line 21: todo: Please add a test using a `519` key as well

acp/acp_local.go
```diff
@@ -36,73 +76,84 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
 }
 
 func (l *ACPLocal) Close() error {
-	return l.localACP.Close()
+	return l.manager.Terminate()
 }
 
 func (l *ACPLocal) AddPolicy(
 	ctx context.Context,
 	creatorID string,
```
shahzadlone, line 109: todo: I guess we should change all instances of `creatorID` to `actorID` now?

acp/identity/identity.go
Lodek: Initially I implemented it as an error, however when I started testing the function, I realized the error would never happen under those circumstances. It didn't seem to make sense to return an error if it would never happens. As for the comment, it's comented that the operation is infallible.

acp/identity/identity_test.go
Lodek: Currently the identity struct only seems to support secp256k1 keys. Adding other key types would be outside the scope of this PR

acp/identity/identity_test.go
Lodek: :saluting_face:

acp/identity/identity.go
shahzadlone: I have a preference to avoid panics in prod code. If this library `cyware/ssi-sdk/did/key` starts returning errors more frequently we should have proper error handling IMO over a panic.

However can hear from others too, @sourcenetwork/database-team happy to follow consensus. Also incase keeping panic should follow the naming of functions having `Must` key word in their name

acp/acp_local.go
```diff
@@ -106,13 +106,13 @@ func (l *ACPLocal) Close() error {
 
 func (l *ACPLocal) AddPolicy(
 	ctx context.Context,
-	creatorID string,
+	actorID string,
```
shahzadlone, line 109: todo: Don't forget to also update the creator in other places (like `acp/acp.go` and `source_hub_client.go` files)

acp/identity/identity.go
shahzadlone: https://github.com/sourcenetwork/defradb/issues/615

More context.

Discussed a bit in standup and we currently have only 3 panics, 2 of which we want to maintain.

Coming back to our discussion, I have a harder preference for it now.

acp/identity/identity.go
shahzadlone: Still todo pre-merge incase forgot.

acp/acp_local.go
Lodek: actually, I'm going to revert this change for now. It's not clear to me what "creator" even means in this context and what defra wants to do with it. ACP Core / SourceHub do not care about the actor creating the Policy. Maybe this could be removed entirely but that's not something I want to change in this PR. Once Andy resumes the integration with SourceHub we might have a better idea of what is desireable from the defra side. 
A thought regarding this is: if having the association between policy and the actor that created it is desireable from a Defra POV, maybe it could be added as part of the Policy attributes

acp/acp_local.go
Lodek: See comment on the other thread

acp/acp_local.go
shahzadlone: I am fine with that. We can change them once we have a more clearer picture.

acp/identity/identity_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func execTestWithMockedProducer(test func()) {
+	// pre: replace the producer function
+	didProducer = func(kt crypto.KeyType, publicKey []byte) (*key.DIDKey, error) {
+		return nil, fmt.Errorf("some did generation error")
+	}
+	test()
+	// post: restore producer function
+	didProducer = key.CreateDIDKey
+}
```
shahzadlone, line 77: todo: Would prefer to not use this fixture function and do these steps in all the test code then and there, if you take my other `todo` you can completely remove the `pre` and `post` steps and just pass the `did` producer as an arg and avoid the global mutating and reseting.

acp/source_hub_client.go
```diff
@@ -94,6 +90,7 @@ type sourceHubBridge struct {
 	client sourceHubClient
 }
 
+var _ ACP = (*sourceHubBridge)(nil)
```
shahzadlone, line 93: todo: Guessing you accidently duplicated this, please remove.

acp/identity/identity.go
```diff
@@ -17,6 +17,15 @@ import (
...
 )
 
+// didProducer is a concrete function which
+// produces a did from a given key type and pub key bytes
+//
+// Currently the did production operation is
+// infalliable, but in order to assure the correct error
+// is being returned, this pkg private variable
+// can be used in tests to locally mock the producer function
+var didProducer = key.CreateDIDKey
```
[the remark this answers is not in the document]
AndrewSisley: Very good catch, thanks Shahzad :)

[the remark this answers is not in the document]
Lodek: I really don't like this suggestion. I don't think callers of a package should be required to inject the concrete function implementation.

The package is succint enough such that there's no confusion about what the didProducer  means, furthermore it is simple to the point that we get the desired coverage / protection against changes breaking changes in the upstream libraries without adding abstraction for the sake of abstractions.

I'll change this construction but I *really* don't think it's worth it. Adding 1/2 interfaces to achieve essentially what will be an elaborate mock of 1 function (which doesn't have an alternative implementation) in a package with a single file screams unnecessary complexity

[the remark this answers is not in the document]
AndrewSisley: I had another look, @Lodek why is the test mutating this?  Why are you not just calling `key.CreateDIDKey` in `DIDFromPublicKey`?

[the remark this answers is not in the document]
Lodek: > Is it just to get code coverage over a dead-code line

Pretty much. Shahzad mentioned he had a preference to having coverage over that test scenario. I ended redesigning things a bit as to avoid the global mutation and still getting the coverage that was requested.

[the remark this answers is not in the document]
AndrewSisley: todo: You got rid of the global (I think), but the cure still looks way worse than the disease. @shahzadlone can we please not do this? (sorry Bruno for you being dragged back and forth on this)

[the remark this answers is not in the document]
shahzadlone: I don't necessarily care about testing for this scenario, if it will make us over abstract things.

I am not sure why the interface needed to be introduced, I agree with Andy the new changes aren't better than what we had previously.

My main concern was the previous glabal var.

If you don't want to let package inject / pass it. Then why not simply keep the previous implementation and just set that variable locally within that function.

The prev implementation was good, just needed to get rid of glabal var.

acp/acp_local.go
```diff
@@ -36,73 +76,84 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
 
-	l.localACP = &localACP
+	l.manager = manager
+	l.engine = engine.NewACPEngine(manager)
 	return nil
 }
 
 func (l *ACPLocal) Close() error {
-	return l.localACP.Close()
+	return l.manager.Terminate()
```
shahzadlone, line 104: question: `engine` doesn't need any cleanup?

acp/source_hub_client.go
shahzadlone: This was marked resolved but is still not done, so unresolving.

Lodek: i believe it should be done, unless i'm missing something

acp/acp_local.go
Lodek: No, the runtime manager is responsible for managing any resources allocated to the engine

acp/identity/identity.go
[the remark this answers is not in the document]
Lodek: I reverted the changes. Now there's a private pkg higher order function to produce the did, this allows testing the DIDFromPubKey func, however without the Identity provider there's no way to test the error handling in the IdentityFrom* methods, so those will be left untested.

acp/acp_local.go
```diff
@@ -36,73 +76,84 @@ func (l *ACPLocal) Init(ctx context.Context, path string) {
...
+	principal, err := auth.NewDIDPrincipal(creatorID)
+	if err != nil {
+		return "", newErrInvalidActorID(err, creatorID)
+	}
+	ctx = auth.InjectPrincipal(ctx, principal)
+
+	marshalType := types.PolicyMarshalingType_SHORT_YAML
+	if isJSON := fastjson.Validate(policy) == nil; isJSON { // Detect JSON format.
+		marshalType = types.PolicyMarshalingType_SHORT_JSON
+	}
```
shahzadlone, line 122: question: Why was the detection moved from `source_hub_client` into the `local acp` logic? Wouldn't this be nicer for @AndrewSisley's remote acp implementation as well?

shahzadlone (approved):
Just left one question, please resolve before merging. Giving an LGTM.

Thanks @Lodek for doing all the requested changes :)

WOOT WOOT your 1st DefraDB PR 🥳

acp/acp_local.go
Lodek: yeah, fair. I'll add a custom Defra type for it, as I did for the Registration Result

shahzadlone:
Also the related issue won't close upon merge, nor is it picked up by github at the moment, because you just have `#2694` in the PR description.

So please don't forget to add the resolving key word (`close`, `closes`, `closed`, `fix`, `fixes`, `fixed`, `resolve`, `resolves`, `resolved`) with it.

So you can put for example:

`Resolve #2694`

acp/acp_local.go
shahzadlone: I think just leaving as it was before should be fine? why do we need a custom type?

PR #2693, opened 2024-06-07
ci(i): Add check to keep table of contents up-to-date

### LINKED ISSUE #2692: Keep Readme table of content up to date
Keep Readme table of content up to date is one of the remaining steps to ensure all our documents upon release are up to date.

shahzadlone:
## Relevant issue(s)
Resolves #2692

## Description
- Combine previous documentation workflows in one file
- Add toc script tool
- Add the toc make rule
- Add toc job

## How has this been tested?
- Using `act` tool
- Manually in the commit which was reverted to ensure it fails if Readme is not updated.
Action-run: https://github.com/sourcenetwork/defradb/actions/runs/9417618174/job/25943201802?pr=2693

- WSL2


shahzadlone:
@fredcarle no more `code 12 + enter` hahaha

AndrewSisley:
> Combine previous documentation workflows in one file

question: What made you chose to do this?

tools/scripts/md-toc/gh-md-toc
```diff
@@ -0,0 +1,421 @@
+#!/usr/bin/env bash
+
```
AndrewSisley, line 2: thought: :sweat_smile: Please never ask me to maintain this file.  Are we really sure this is worth it?

shahzadlone: You/we don't have to. I been using this for a few years now and has 3.2k stars. Is a keep it and forget it kind of thing IMO.

Worst case scenario, if something does break, we copy the new version with the fix and replace this script.

AndrewSisley: It is copy-pasted as-is? No changes from the source?  Why have we copy pasted it?

shahzadlone:
> > Combine previous documentation workflows in one file
> 
> question: What made you chose to do this?

No hard pref or reason just some minor ones, seems nicer to have all documentation jobs in one workflow file (they still behave in parallel manner and can be controlled granularly to make required or not as that is based on the job name.)

One nitpicky thing I like is that the github check mark will show: `Check Documentation Workflow / Check http documentation` rather than `Check Http Documentation Workflow / Check http documentation`

Everything remains the same, except we loose the ability to configure the trigger (`push` and `pull`) in a granular manner per workfllow vs same for all jobs now, but we will always have same for all these jobs anyways so doesn't matter.

If there is a hard pref to keep them in separate workflow files lmk, as mentioned I am in different.

AndrewSisley:
> if there is a hard pref to keep them in separate workflow files lmk, as mentioned I am in different.

No real preference from me at all atm :)  I was just curious :)  Thanks for the explanation

tools/scripts/md-toc/gh-md-toc
AndrewSisley: The readme pasted in suggests that we wouldn't need devs to manually download it (unless maybe it requires sudo?)

It also looks like the script doesn't work on windows:
> If you want it on Windows, you better to use a golang based implementation

shahzadlone: > The readme pasted in suggests that we wouldn't need devs to manually download it (unless maybe it requires sudo?)
> 
> It also looks like the script doesn't work on windows:
> 
> > If you want it on Windows, you better to use a golang based implementation
> 
> todo: Why don't we just use the go version instead of the mac/linux-only bash script, then we wont need to copy-paste code about?
> 
> `go install "github.com/ekalinin/github-markdown-toc.go/cmd/gh-md-toc@latest"` etc...

I happy to adhere to that suggestion if there is preference :) I just didn't think we cared much about windows dev tooling.

shahzadlone: I haven't used the go version of the tool personally before, I wonder if formatting and everything else is the same.

shahzadlone: I tried switching to the go project, but sadly the go project is missing the `insert` option, and only generates the table of content (i.e. doesn't update it automatically). So will prefer the current approach.

shahzadlone: @sourcenetwork/database-team Please chime in if you have any other suggestions regarding the copy-paste and lack of window dev tool support (or another tool). Happy to follow the consensus.

README.md
```diff
@@ -15,29 +15,30 @@ DefraDB is a user-centric database that prioritizes data ownership, personal pri
...
-- [Securing the HTTP API with TLS](#securing-the-http-api-with-tls)
-- [Access Control System](#access-control-system)
-- [Supporting CORS](#supporting-cors)
-- [Backing up and restoring](#backing-up-and-restoring)
-- [Community](#community)
-- [Licensing](#licensing)
-- [Contributors](#contributors)
+
+Table of Contents
+=================
```
fredcarle, line 20: todo: Please keep the same syntax as our other headers :)

tools/scripts/md-toc/gh-md-toc
fredcarle: If we never have to maintain this, I don't really care that we just copy-pasted it. I also don't think it's a big deal that it doesn't work on windows. Devs on windows can just manually edit if they have to. If someone what's to contribute to the go version to include insert, we can switch to the go version when it's supported.

fredcarle (approved):
LGTM. Please fix the table of content header syntax before merging.

tools/scripts/md-toc/gh-md-toc
shahzadlone: > If we never have to maintain this, I don't really care that we just copy-pasted it. I also don't think it's a big deal that it doesn't work on windows. Devs on windows can just manually edit if they have to. If someone what's to contribute to the go version to include insert, we can switch to the go version when it's supported.

I am okay with this! I have subscribed to the issue, if it gets implemented happy to migrate to using go version of the tool.

README.md
shahzadlone: - [x] done

PR #2697, opened 2024-06-07
ci(i): Add action to ensure tidyness

shahzadlone:
## Relevant issue(s)
Resolves ##2695

### Backstory
- A commit (c42e7ee6) broke mod tidy command recently
- Which was fixed in the next commit in this PR: https://github.com/sourcenetwork/defradb/pull/2548
- Some head scratching was done to find out that mod tidy needed a minimum version, hence the current pinned mod tidy

There are 2 things being checked within the new added action:
1) That the backstory doesn't happen again (`make tidy` isn't broken).
2) `make tidy` is ran and up to date.

NOTE: `(2)` above I am not sure if we want to keep, or not I would be okay with just `(1)`. However here are some reasons why you would want `(2)`:
- always be in `tidy state`
- one less command to do at release stage.

## How has this been tested?
- Action that failed when `make tidy` was broken: https://github.com/sourcenetwork/defradb/actions/runs/9422665720/job/25959440810?pr=2697

- Action that failed when `make tidy` was not up to date (i.e. not tidy):
https://github.com/sourcenetwork/defradb/actions/runs/9422726089/job/25959622978?pr=2697

- WSL2

### DIFF
--- a/.github/workflows/check-tidy.yml
+++ b/.github/workflows/check-tidy.yml
@@ -1,0 +1,59 @@
+# Copyright 2024 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+# This workflow checks that go mod tidy command we have set for the specific
+# go version is not broken, for example `go mod tidy -go=1.21.3`. This
+# can cause some head scratching at times, so better catch this in the PR.
+#
+# Inaddition to that also checks that we are currently in a `tidy` state.
+name: Check Tidy Workflow
+
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+
+  push:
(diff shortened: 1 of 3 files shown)


.github/workflows/check-tidy.yml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+    branches:
+      - master
+      - develop
+
+jobs:
+  check-tidy:
+    name: Check mod tidy job
```
AndrewSisley, line 33: thought: I thought you just condensed all the `check changes` into a single workflow?  Would this not be better within that file?

AndrewSisley (approved):
LGTM, but please check my thought before deciding whether to merge or not

.github/workflows/check-tidy.yml
shahzadlone: It wasn't all checks that I moved in that file. Only the documentation related checks were moved there.

AndrewSisley: Ah okay nice, thanks for the clarification/reminder :)

PR #2698, opened 2024-06-07
refactor: Change counters to support encryption

### LINKED ISSUE #2696: Counter CRDT currently don't support encryption
The counter CRDTs are stored as concrete types (int64 or float64) instead of bytes. This makes them ill prepared for the upcoming document encryption feature.

fredcarle:
## Relevant issue(s)

Resolves #2696 

## Description
This PR changes the counter CRDTs to make them support the document encryption feature. They previously stored their values as concrete types (int64 and float64) instead of bytes. Storing them as bytes allow them to be stored plainly or encrypted.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/internal/core/crdt/counter.go
+++ b/internal/core/crdt/counter.go
@@ -33,7 +33,7 @@ type Incrementable interface {
 }
 
 // CounterDelta is a single delta operation for a Counter
-type CounterDelta[T Incrementable] struct {
+type CounterDelta struct {
 	DocID     []byte
 	FieldName string
 	Priority  uint64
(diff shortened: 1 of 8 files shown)


internal/core/crdt/counter.go
```diff
@@ -120,7 +111,7 @@ func (c Counter[T]) Value(ctx context.Context) ([]byte, error) {
 // WARNING: Incrementing an integer and causing it to overflow the int64 max value
 // will cause the value to roll over to the int64 min value. Incremeting a float and
 // causing it to overflow the float64 max value will act like a no-op.
-func (c Counter[T]) Increment(ctx context.Context, value T) (*CounterDelta[T], error) {
+func (c Counter) Increment(ctx context.Context, value []byte) (*CounterDelta, error) {
```
islamaliev, line 114: question: is it really necessary to change the API? I feel like changing `CounterDelta` to hold `[]byte` instead of a number is enough to enable encryption. I'd prefer having strong types till as late as possible.

islamaliev (approved):
Looks good. I just have 1 question/preference to clarify before you merge.

internal/core/crdt/counter.go
fredcarle: `FieldValue` is only practical on save and quite inconvenient on merge. Maybe I can overlook the duplication in this case to help with simplifying the merge.

shahzadlone (approved):
LGTM

internal/core/crdt/counter.go
fredcarle: There are 3 ways of handling this. 
1. Use different deltas (generics) for every type
2. Store the kind in the delta
3. Pass the kind to `NewCounter`
1 and 2 are similar in that they both store type-related info in the data structure. 1 stores a differently named data structure and 2 stores extra bytes. 
The disadvantage of 1 is that it requires maintaining more IPLD data structures and more types to support for `coreblock.Block`.
The disadvantage of 2 is that it store extra bytes and given that we need the field definition anyways to create the `MerkleCounter`, it doesn't make a lot of sense to store those extra bytes.
3 is equivalent to both 1 and 2 but depends on the field definition's kind that we pass along when we create the new counter. This is arguably the most type-safe way of doing things as it guarantees that we merge the proper type to the datastore on all code paths that call `Merge`. 
One could think that keeping the concrete type as long as possible in the `Increment` path would be beneficial for strong typing but this falls apart when we look at the `Merge` path from a P2P sync which doesn't have access to that strong type safety.

islamaliev: yeah, looks like these is no perfect solution here. I'm fine then with keeping it as is.

PR #2700, opened 2024-06-07
refactor: Use events to test network logic

### LINKED ISSUE #2687: Unsubscribing from db events causes race condition
Found in #2686

### LINKED ISSUE #2699: Use events system to test networking logic
Refactor the events system so that it can be used to test networking events. After these changes are done we should be able to mock the networking logic in tests.

nasdf:
## Relevant issue(s)

Resolves #2699
Resolves #2687

## Description

This PR refactors the events package and updates the networking tests to use events.

I think there is a bit more clean up to be done, but the majority should be ready for review.

## Tasks

## How has this been tested?

`make test`

- MacOS


tests/integration/events/utils.go
```diff
@@ -79,14 +80,19 @@ func ExecuteRequestTestCase(
...
 
 	indexOfNextExpectedUpdate := 0
 	go func() {
 		for {
 			select {
-			case update := <-eventsChan:
+			case value := <-eventsSub.Message():
+				update, ok := value.Data.(events.UpdateEvent)
+				if !ok {
+					continue // ignore invaid value
```
fredcarle, line 93: typo: invalid :)

events/bus.go
```diff
@@ -0,0 +1,138 @@
...
+// Events returns the names of all subscribed events.
+func (s *Subscription) Events() []string {
+	return s.events
+}
+
+// Bus is used to publish and subscribe to events.
+type Bus struct {
+	subId  int
+	subs   map[int]*Subscription
+	events map[string]map[int]any
```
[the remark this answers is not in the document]
fredcarle: suggestion: It seems like `any` could be changed to `struct{}`.

events/bus.go
```diff
@@ -0,0 +1,138 @@
...
+		events: make(map[string]map[int]any),
+	}
+}
+
+// Publish publishes the given event to all subscribers.
+func (b *Bus) Publish(msg Message) {
+	b.mutex.RLock()
+	defer b.mutex.RUnlock()
+
+	subscribers := make(map[int]any)
```
fredcarle, line 68: suggestion: Here too it seems like `any` could be changed to `struct{}`

fredcarle (commented):
I need to head out but first pass looks really good. I love that you removed 1000 lines of code 🤘. I'll finish a more in depth review on Monday.

tests/integration/p2p.go
```diff
@@ -398,12 +413,16 @@ func setupReplicatorWaitSync(
 		ready <- struct{}{}
 		for waitIndex := 0; waitIndex < len(sourceToTargetEvents); waitIndex++ {
 			for i := 0; i < targetToSourceEvents[waitIndex]; i++ {
-				err := sourceNode.WaitForPushLogByPeerEvent(targetPeerInfo.ID)
-				require.NoError(s.t, err)
+				msg, ok := <-sourceSub.Message()
```
AndrewSisley, line 416: todo: Please document why you are doing this, it is unclear reading the code (same goes for the other similar block).  

todo: Please also document what `sourceSub` and `targetSub` do in the `setupReplicatorWaitSync` definition.

tests/integration/utils2.go
```diff
@@ -656,6 +657,9 @@ func setStartingNodes(
 
 		s.nodes = append(s.nodes, c)
 		s.dbPaths = append(s.dbPaths, path)
+
+		sub := c.Events().Subscribe(100, events.MergeCompleteEventName)
```
AndrewSisley, line 661: todo: Please document why `100`

tests/integration/utils2.go
```diff
@@ -812,6 +828,9 @@ func configureNode(
 
 	s.nodes = append(s.nodes, c)
 	s.dbPaths = append(s.dbPaths, path)
+
+	sub := c.Events().Subscribe(100, events.MergeCompleteEventName)
```
AndrewSisley, line 832: todo: Please document why 100, and why you are subscribing

tests/integration/p2p.go
```diff
@@ -398,12 +413,16 @@ func setupReplicatorWaitSync(
...
+				if ok {
+					assert.Equal(s.t, targetPeerInfo.ID, msg.Data.(events.MergeEvent).ByPeer)
```
AndrewSisley, line 418: suggestion: I think `require` would be more appropriate here (and the other similar blocks in this file) - a failure here is a bug in the test, not a bug in the production code?  It would also probably be really confusing if the code continued past this point no?

todo: Please document why this assert/require is done

events/bus.go
```diff
@@ -0,0 +1,138 @@
...
+func (s *Subscription) Events() []string {
+	return s.events
+}
+
+// Bus is used to publish and subscribe to events.
+type Bus struct {
+	subId  int
+	subs   map[int]*Subscription
+	events map[string]map[int]any
+	mutex  sync.RWMutex
```
AndrewSisley, line 52: discussion: The scope of this mutex is quite large, and I don't think I agree with this change.  The previous solution was non-blocking and it didn't discard messages (the `default` block in `Bus`.`Publish` here).

I think it is very important that publishing events does not block 'core' database operations, particularly in user-space.  The rapid discarding of events if the buffer fills also far from ideal, and I worry that it will lead to some painful to figure out bug reports, particularly in the P2P space.

AndrewSisley (changes requested):
I think we will need to chat about this one - it might be worth separating the changes made outside of the `events` package from the rest - there are some very good changes in here that should not be delayed/confused with the events package implementation change.

Regarding the events package, it is now simpler, but it is also blocking (in user-space), and less reliable (it discards messages quite rapidly if the subscription buffer fills - the original buffer size was 100 + a 60 second timeout, it is now 10 items and no timeout) and as such I think I will require a bit of convincing in order to see this as a good change.

I do also disagree with the removal of the `Channel` interface, it was used to allow us to swap out implementations without having to change the calling code.  Whilst I think the original `simple` channel is better than the proposal for our use-cases, it still needs to be improved upon long-term, and different use-cases within the defra code base would likely benefit from different `Channel` implementations.

events/bus.go
fredcarle: I don't think you can have non-blocking and also not discarding messages at the same time. You can delay blocking by having a certain size of channel but when it's full we can either block or discard. Unless we have a store to keep track of unsent messages but that is a whole other subject which probably shouldn't apply here. 

The previous solution was blocking as far as I can tell.

events/events.go
```diff
@@ -8,48 +8,74 @@
...
+	ConnectEventName = "net:connect"
+)
 
-	// Unsubscribe unsubscribes from the Channel, closing the provided channel.
-	//
-	// Will do nothing if this object is already closed.
-	Unsubscribe(Subscription[T])
+// ConnectEvent is an event that is published when
+// a peer connection has changed status.
+type ConnectEvent = event.EvtPeerConnectednessChanged
```
fredcarle, line 38: suggestion: Since we are in the `events` package, the `Event` suffix isn't necessary. We could change `events.ConnectEvent`, `events.PubSubEvent`, `events.UpdateEvent`, `events.MergeEvent` to `events.Connect`, `events.PubSub`, `events.Update`, `events.Merge`

events/bus.go
fredcarle: We have 2 event bus in this PR. We have the internal event bus and the HTTP subscription event bus. The internal event bus will only ever block with the mutex on startup when it adds the internally defined and static subscribers. For the rest of its life, the mutex will never be blocking. Right now the `Publish` method is non-blocking on the channel and that is something we can debate. For internal message exchanges, blocking might actually be desirable.

The HTTP subscription bus, however, has a dynamic list of subscribers. This one will block more often to update the subscriber list. We do not want to block on the channel here that is certain. Since HTTP subscriptions touch user space, using timeouts in the `Publish` method would be appropriate.
```go
for id := range subscribers {
	go func(id int) {
		select {
		case b.subs[id].value <- msg:
			// published event
		case <-time.After(time.Second * 10):
			// channel full, timeout
		}
	}(id)
}
```

AndrewSisley: > The previous solution was blocking as far as I can tell.

It was much less blocking - it was lock free, and the (buffered) `write` channel was consumed by a dedicated Go routine that fed inputs into the (buffered) subscribing channel. And for the write channel to block a subscriber channel would need to be full as well as the write channel.  Atm the write channel is hardcoded to size 0, and the subscribers 100.  It won't throw away data until a subscriber times out after 60 seconds.  We planned on making all of these settings configurable but never got round to it

The proposed solution blocks on every concurrent call regardless of configuration thanks to the quite broad mutex, and there is no way to configure around that (unlike a buffered write channel).  It will discard data as soon as the (size 10) subscriber buffer is filled - this is really not very tolerant at all.

The current solution is far from perfect, but I do think it is better than the proposal.

AndrewSisley: > the mutex will never be blocking

The mutex will block the moment concurrent writers try to write to the bus.

events/bus.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Close closes the event bus by unsubscribing all subscribers.
+func (b *Bus) Close() {
+	var subs []*Subscription
+
+	b.mutex.RLock()
+	for _, sub := range b.subs {
+		subs = append(subs, sub)
+	}
```
islamaliev, line 132: todo: please either preallocate required space:
```go
subs := make([]*Subscription, len(b.subs))

b.mutex.RLock()
for i := range b.subs {
    subs[i] = b.subs[i]
}
```
or, even better, use `copy`:
```go
var subs []*Subscription

b.mutex.RLock()
copy(subs, b.subs)
```

events/bus_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+package events
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBusPublish(t *testing.T) {
```
islamaliev, line 19: suggestion: would be great to have more tests. Current one covers only the happy path and nothing else.
cases like these: subscribe twice to the same event, test unsubscribe, test if not subscribed-to event is not received, test close and others.

I'm sure it's mostly covered with integration tests, but would nice to have explicit unit tests for the event bus

events/events.go
islamaliev: that would be neat. But for this, following best go practices, would be better to make package name singular `event`.

tests/integration/events/utils.go
```diff
@@ -79,14 +80,19 @@ func ExecuteRequestTestCase(
 
 	testRoutineClosedChan := make(chan struct{})
 	closeTestRoutineChan := make(chan struct{})
-	eventsChan, err := db.Events().Updates.Value().Subscribe()
+	eventsSub := db.Events().Subscribe(5, events.UpdateEventName)
```
islamaliev, line 83: question: shouldn't this be unsubscribed-from at some point?

events/events.go
fredcarle: @nasdf you changed to `event` but didn't do my suggestion. Did you just forget or did you decide you didn't like it? :)

nasdf: @fredcarle I'm considering moving the event definitions to their respective packages.

tests/integration/events/utils.go
nasdf: node `Close` method takes care of unsubscribing.

tests/integration/state.go
```diff
@@ -52,6 +53,9 @@ type state struct {
 	// These synchronisation channels allow async actions to track their completion.
 	syncChans []chan struct{}
 
+	// eventSubs is a list of all event subscriptions
+	eventSubs []*events.Subscription
```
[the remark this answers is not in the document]
nasdf: confirmed they are closed on node `Close`

events/bus.go
nasdf: Implementation is now channel based like we discussed

tests/integration/utils2.go
nasdf: this is no longer relevant to the implementation

tests/integration/p2p.go
nasdf: moved to a helper func and added docs

events/events.go
fredcarle: Also a good option 👍

fredcarle: The downside of defining the events in their respective packages is that if might result in unwanted dependencies or even circular dependencies depending in which package it is defined. `MergeEvent` being defined in `db` or `net` for example.

net/process.go
```diff
@@ -64,7 +64,7 @@ func (bp *blockProcessor) processRemoteBlock(
 	// Initiate a sync of the block's children
 	bp.wg.Add(1)
 	bp.handleChildBlocks(ctx, block)
-
+	bp.wg.Wait()
```
fredcarle, line 67: praise: Make much more sense to have this here.

event/bus.go
```diff
@@ -0,0 +1,179 @@
...
+	return s.value
+}
+
+// Events returns the names of all subscribed events.
+func (s *Subscription) Events() []string {
+	return s.events
+}
+
+// Bus is used to broadcast events to subscribers.
+type Bus struct {
```
AndrewSisley, line 56: todo: I thought we agreed to keep this behind an interface?  Happy to resume that conversation if need be :)

http/client.go
```diff
@@ -451,7 +451,7 @@ func (c *Client) Headstore() ds.Read {
 	panic("client side database")
 }
 
-func (c *Client) Events() events.Events {
+func (c *Client) Events() *event.Bus {
 	panic("client side database")
```
AndrewSisley, line 455: thought: This can/should probably be implemented for http/cli at some point, half tempted to ask you to link this panic to a github issue.

internal/db/db.go
```diff
@@ -43,6 +43,13 @@ var (
 	_ client.Collection = (*collection)(nil)
 )
 
+const (
+	// commandBufferSize is the size of the channel buffer used to handle events.
+	commandBufferSize = 100
```
AndrewSisley, line 48: todo: I think the default should be a much larger value, maybe 100,000 for now. 100,000 would still only be 1.6MB I think, and stuff will start blocking if the capacity is exceeded.

event/bus.go
AndrewSisley, line 56: todo: Doesn't have to be in this PR, but the `OnFooAsync` funcs added in https://github.com/sourcenetwork/defradb/pull/2708 can now be removed IMO - they are dead code (unless we want to keep them to allow users to call them).

tests/integration/p2p.go
```diff
@@ -532,6 +488,47 @@ func waitForSync(
...
+
+	sourcePeerInfo := sourceNode.PeerInfo()
+	targetPeerInfo := targetNode.PeerInfo()
+
+	for waitIndex := 0; waitIndex < len(sourceToTargetEvents); waitIndex++ {
+		for i := 0; i < targetToSourceEvents[waitIndex]; i++ {
+			// wait for message or unsubscribe
+			msg, ok := <-sourceSub.Message()
+			if ok {
+				require.Equal(s.t, targetPeerInfo.ID, msg.Data.(event.MergeEvent).ByPeer)
```
AndrewSisley, line 518: todo: Please document why these requires exist

AndrewSisley (changes requested):
Looks really good Keenan, I really like these changes.  Just got a couple of small todos/suggestions for you before merge :)

event/bus.go
AndrewSisley: It already exists in the `events` package, this PR deleted it.  I think it would just need a rename, removal of the generic, and adding the `...string` param to `Subscribe`, otherwise it looks identical to the `*Bus` signature.

fredcarle: It doesn't make sense to define the interface in the `event` package has it doesn't need/use it. We could define it in  `client` though for the purpose of the `client.DB` interface method `Events() Events`.

AndrewSisley: If we are going to move it, I'd also suggest moving `Subscription` there too, and then the `event` package can become internal.

Otherwise, I do slightly disagree with you.  It is quite common for packages to define their public interfaces within themselves.  Especially given `event` is not really Defra specific, and could quite happily be moved outside of this repo for general use.

fredcarle: Packages defining public interfaces that are not used internally is bad practice though. We should try to avoid it. The goal of an interface is to make something more flexible. Like if we want `db` to be able to use different event bus implementation. I'd argue that we are building the `Bus` specifically for our needs so I don't think it flexibility makes sense here. I much prefer the concrete type as it is now.

events/events.go
nasdf: Good point. I've kept them in the package but renamed as you suggested.

events/events.go
```diff
@@ -8,48 +8,74 @@
...
+	UpdateEventName = "db:update"
+	// ResultsEventName is the name of the database results event.
+	ResultsEventName = "db:results"
+	// MergeRequestEventName is the name of the net merge request event.
+	MergeRequestEventName = "net:merge"
+	// PubSubEventName is the name of the network pubsub event.
+	PubSubEventName = "net:pubsub"
+	// ConnectEventName is the name of the network connect event.
+	ConnectEventName = "net:connect"
+)
```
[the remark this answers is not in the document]
nasdf: I'd like to keep the event names a human readable string for debugging purposes.

Having the events in one package does have the issue you mentioned, but it also makes it much easier to know what events can be subscribed to. There's also the issue of importing an entire package just for an event type.

http/client.go
nasdf: I'm planning on adding this functionality in a follow up PR. One of my design goals was to allow the CLI to subscribe to and print the events as they happen.

event/bus.go
nasdf: I'll try to remember to bring this up in the stand up

http/client.go
AndrewSisley: Sweet, sounds good :)

event/buffered_bus.go
```diff
@@ -0,0 +1,141 @@
...
+type subscribeCommand *Subscription
+
+type unsubscribeCommand *Subscription
+
+type publishCommand Message
+
+type closeCommand struct{}
+
+// bufferedBus is a bus that uses a buffered channel to manage subscribers and publish messages.
+type bufferedBus struct {
```
AndrewSisley, line 26: todo: Please add a compile time check to ensure (and document) that this type implements the `Bus` interface

event/bus.go
nasdf: I added the interface because I will be using it to create a JavaScript event bus for the WASM build.

event/buffered_bus.go
nasdf: The constructor method does this already. I usually choose one or the other not both.

AndrewSisley: Fair enough :)  Fred might ask you to change the constructor to return the concrete though :grin:

events/events.go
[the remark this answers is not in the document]
islamaliev: but the package that wants to listed to a specific event usually depends on the package that fires that event as it needs to receive the package-event-specific object.

> I'd like to keep the event names a human readable string for debugging purposes

You can declare string enum:
```go
type EventName string
```
and in `db` package:
```go
const (
    // MergeCompleteEventName is the name of the database merge complete event.
    MergeCompleteEventName = events.EventName("db:merge-complete")
    // UpdateEventName is the name of the database update event.
    UpdateEventName = events.EventName("db:update")
    // ResultsEventName is the name of the database results event.
    ResultsEventName = events.EventName("db:results")
)
```

events/events.go
islamaliev: this is being discussed in another thread https://github.com/sourcenetwork/defradb/pull/2700#discussion_r1633762423 :)

event/buffered_bus.go
```diff
@@ -0,0 +1,141 @@
...
+	// subID is incremented for each subscriber and used to set subscriber ids.
+	subID atomic.Uint64
+	// subs is a mapping of subscriber ids to subscriptions.
+	subs map[uint64]*Subscription
+	// events is a mapping of event names to subscriber ids.
+	events map[string]map[uint64]struct{}
+	// commandChannel manages all commands sent to this simpleChannel.
```
islamaliev, line 33: question: what are you referring to by "simpleChannel"?

event/buffered_bus.go
```diff
@@ -0,0 +1,141 @@
...
+	//
+	// It is important that all stuff gets sent through this single channel to ensure
+	// that the order of operations is preserved.
+	//
+	// WARNING: This does mean that non-event commands can block the database if the buffer
```
islamaliev, line 38: nitpick: sounds more like a NOTE, not a WARNING

event/buffered_bus.go
```diff
@@ -0,0 +1,141 @@
...
+		return
+	}
+	b.commandChannel <- unsubscribeCommand(sub)
+}
+
+func (b *bufferedBus) Close() {
+	if b.isClosed.Load() {
+		return
+	}
+	b.isClosed.Store(true)
```
islamaliev, line 93: todo: these 2 ops are not dome atomically and might close twice. Please use compare-and-swap:
```go
if !b.isClosed.CompareAndSwap(false, true) {
    return
}
...
```

Check please if you have similar occurrences.

islamaliev (changes requested):
Great job, but there is at least one critical issue related to atomics that needs to be addressed before merge.

events/bus.go
islamaliev: you marked the comment as "resolved" but I don't see how it was addressed. The same goes for https://github.com/sourcenetwork/defradb/pull/2700#discussion_r1633748528

Would be nice if for every comment you left a very short comment mentioning how you addressed it. That save reviewers' time. That's what we all tend to.

events/events.go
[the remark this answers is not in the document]
nasdf: > but the package that wants to listed to a specific event usually depends on the package that fires that event as it needs to receive the package-event-specific object.

This won't always be the case, and will cause problems since the `db` package is internal.

> You can declare string enum:

I've added an `event.Name` type.

event/buffered_bus.go
nasdf: good catch!

event/buffered_bus.go
nasdf: Fixed. It was a copy paste error.

events/bus.go
nasdf: My bad. I thought I left a message. This should no longer be an issue with the implementation.

islamaliev: > This should no longer be an issue with the implementation.

I don't quite understand what you mean. And I see no changes in the code.

nasdf: The code referenced in the above comment has been replaced.

event/buffered_bus.go
```diff
@@ -0,0 +1,143 @@
...
+	b.commandChannel <- unsubscribeCommand(sub)
+}
+
+func (b *bufferedBus) Close() {
+	if !b.isClosed.CompareAndSwap(false, true) {
+		return
+	}
+	b.commandChannel <- closeCommand{}
+	// Wait for the close command to be handled, in order, before returning
+	<-b.hasClosedChan
```
islamaliev, line 98: question: this is thread-safe only under the assumption that all the calls to `bufferedBus` are done by the same thread. I'm not sure this is the case. Is it?

Otherwise the following scenario is possible:
1. execution enters `Subscribe` method, `b.isClosed == false` so it proceeds but pauses just before `b.commandChannel <- subscribeCommand(sub)`
2. concurrently `Close` is being called, sets `b.isClosed` to `true`, pushes `b.commandChannel <- closeCommand{}`
3. `handleChannel` receives `closeCommand` and executes `close(b.commandChannel)`
4. now the thread from step 1 resumes, executes `b.commandChannel <- subscribeCommand(sub)` and causes a panic.

events/bus.go
islamaliev: sorry, my bad. Didn't look at the last changes.

event/buffered_bus.go
```diff
@@ -0,0 +1,143 @@
...
+				if _, ok := b.events[event]; !ok {
+					b.events[event] = make(map[uint64]struct{})
+				}
+				b.events[event][t.id] = struct{}{}
+			}
+			b.subs[t.id] = t
+
+		case unsubscribeCommand:
+			if _, ok := b.subs[t.id]; !ok {
+				continue // not subscribed
```
shahzadlone, line 123: question: Just pointing incase easy to test, that this path is not tested.

event/buffered_bus.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			for _, event := range t.events {
+				delete(b.events[event], t.id)
+			}
+			delete(b.subs, t.id)
+			close(t.value)
+
+		case publishCommand:
+			for id := range b.events[WildCardName] {
+				b.subs[id].value <- Message(t)
```
shahzadlone, line 133: question: Just pointing incase easy to test, that this path is not tested.

event/buffered_bus.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			for id := range b.events[t.Name] {
+				if _, ok := b.events[WildCardName][id]; ok {
+					continue
```
shahzadlone, line 137: question: Just pointing incase easy to test, that this path is not tested.

event/bus.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Message returns the next event value from the subscription.
+func (s *Subscription) Message() <-chan Message {
+	return s.value
+}
+
+// Events returns the names of all subscribed events.
+func (s *Subscription) Events() []Name {
+	return s.events
```
shahzadlone, line 61: nitpick: This path is not tested.

event/buffered_bus.go
AndrewSisley: That's quite a broad mutex, please don't merge yet

nasdf: It should only block when calling close.

AndrewSisley: Okay, I think I understand, and there is no nicer way of handling this that I can spot off the top of my head.

todo: Can you please rename `mutex` to `closeLock` or similar, we really really don't want this to be used for anything else.

todo: Can please also document the lock on its declaration in `bufferedBus`, explaining that it is only locked whilst closing.

fredcarle: @AndrewSisley can you explain what you mean by broad? We use the same technique in the memory store and I think it's really fine for this as well.

AndrewSisley: > @AndrewSisley can you explain what you mean by broad? We use the same technique in the memory store and I think it's really fine for this as well.

The lock spans every public call to this struct, it can't really get any broader :)  If it only locks whilst closing that's fine, but it would be very damaging if that were to change (hence the documentation and name change requested).

event/buffered_bus_test.go
```diff
@@ -0,0 +1,179 @@
...
+	output2Ch2 := <-sub2.Message()
+
+	assert.Equal(t, msg1, output1Ch1)
+	assert.Equal(t, msg1, output1Ch2)
+
+	assert.Equal(t, msg2, output2Ch1)
+	assert.Equal(t, msg2, output2Ch2)
+}
+
+func TestBufferedBusSubscribersDontRecieveItemsAfterUnsubscribing(t *testing.T) {
```
islamaliev, line 162: praise: tests overall look good. Thanks for taking your time.

event/buffered_bus_test.go
```diff
@@ -0,0 +1,179 @@
...
+	msg := NewMessage("test", 1)
+	bus.Publish(msg)
+
+	evt := <-sub.Message()
+	assert.Equal(t, evt, msg)
+
+	select {
+	case <-sub.Message():
+		t.Errorf("should not receive duplicate message")
+	case <-time.After(1 * time.Second):
```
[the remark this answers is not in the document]
nasdf: I've decreased the delay. It's required because the publish is asynchronous.

event/buffered_bus_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+func TestBufferedBusUnsubscribeTwice(t *testing.T) {
+	bus := NewBufferedBus(0, 0)
+	defer bus.Close()
+
+	sub, err := bus.Subscribe(WildCardName)
+	assert.NoError(t, err)
+
+	bus.Unsubscribe(sub)
+	bus.Unsubscribe(sub)
```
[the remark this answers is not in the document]
nasdf: This call itself won't panic because it is simply adding a message to the command channel which is processed asynchronously.

event/buffered_bus_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	bus.Close()
+
+	<-sub.Message()
+
+	// just assert that we reach this line, for the sake of having an assert
+	assert.True(t, true)
+}
+
+func TestBufferedBusUnsubscribeTwice(t *testing.T) {
```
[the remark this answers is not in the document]
nasdf: I've updated all the test names

event/buffered_bus_test.go
```diff
@@ -0,0 +1,179 @@
...
+	// just assert that we reach this line, for the sake of having an assert
+	assert.True(t, true)
+}
+
+func TestBufferedBusSubscribersAreNotBlockedAfterClose(t *testing.T) {
+	bus := NewBufferedBus(0, 0)
+	defer bus.Close()
+
+	sub, err := bus.Subscribe("test")
+	assert.Nil(t, err)
```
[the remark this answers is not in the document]
nasdf: It was copy pasted, but I've fixed them all.

event/buffered_bus_test.go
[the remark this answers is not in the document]
nasdf: Updated

islamaliev (approved):
LGTM! Thanks Keenan for the adjustments.

shahzadlone (approved):
Thanks for testing the requested paths.

LGTM

PR #2701, opened 2024-06-07
fix: Incorporate schema root into docID

### LINKED ISSUE #2688: SchemaRoot is not incorporated into the docID
This means that documents created against different schema may have the same document IDs.

AndrewSisley:
## Relevant issue(s)

Resolves #2688

## Description

Incorporates schema root into docID.

The production changes are all done in the first two commits, I broke out the test changes to a separate commit to avoid confusion.  There are no significant test changes - is just really really painful updating this as besides the docIDs and cids, the order in which results were returned changed.


tests/integration/mutation/delete/field_kinds/one_to_many/with_show_deleted_test.go
```diff
@@ -68,25 +38,36 @@ func TestDeletionOfADocumentUsingSingleDocIDWithShowDeletedDocumentQuery(t *test
...
+					"name":      "John has a chamber of secrets",
+					"rating":    9.9,
+					"author_id": testUtils.NewDocIndex(1, 0),
+				},
 			},
 			testUtils.Request{
-				Request: fmt.Sprintf(`mutation {
-					delete_Book(docID: "%s") {
+				Request: `mutation {
+					delete_Book(docID: "bae-b5c56d8f-b2f5-57f9-b371-4e9e04903e91") {
```
fredcarle, line 64: question: Is it necessary to hardcode the docID here? You're doing the effort to switch to `NewDocIndex` in a lot of places. It seems like these would also be good candidates for that.

fredcarle (approved):
LGTM!

shahzadlone (approved):
LGTM must have been a pain to update all these, cheers!

Glanced over / skipped reviewing the ID diffs but rest looks good.

PR #2719, opened 2024-06-13
fix: Merge retry logic

### LINKED ISSUE #2718: Merge retries can result in incorrect state
Found in #2700 

When a merge transaction conflict occurs and a merge is retried, it does not account for new DAG heads, which can result in an incorrect state.

### LINKED ISSUE #2721: Partial DAG sync will never merge
If a DAG sync partially fails it will never reach the DAG merge process. The DAG sync should not ignore existing blocks and instead walk the entire graph before merging.

nasdf:
## Relevant issue(s)

Resolves #2718 
Resolves #2721 

## Description

This PR fixes issues with merge retry and DAG sync processes.  It also moves the `docQueue` from the `net` package into the `db` package.

## Tasks

## How has this been tested?

`make test`

- MacOS


net/server.go
```diff
@@ -51,11 +51,6 @@ type server struct {
 	pubSubEmitter  event.Emitter
 	pushLogEmitter event.Emitter
```
fredcarle, line 53: suggestion: If you remove the `docQueue`, it opens the door for transaction conflicts on the dag sync process. One of the test matrix run failed for this exact reason. I suggest you keep it in :)

nasdf: It's still there but in the `db` package instead. I think it makes more sense to have the blocking occur there.

internal/db/merge.go
```diff
@@ -43,28 +45,37 @@ func (db *db) handleMerges(ctx context.Context, merges events.Subscription[event
...
 			}
 			go func() {
-				err := db.executeMerge(ctx, merge)
+				// ensure only one merge per docID
+				queue.add(merge.DocID)
+				defer queue.done(merge.DocID)
+
+				var err error
+				// retry merge up to max txn retries
+				for i := 0; i < db.MaxTxnRetries(); i++ {
```
fredcarle, line 54: thought: If you use the merge queue, I expect that a retry will never be needed.

nasdf: If a user updates a doc while a merge is in progress it could still happen.

fredcarle: 😅 very true

net/server.go
fredcarle: The problem is that if the conflict happens in exactly the wrong way, blocks will be missing from the blockstore. I wanted to remove it when I refactored the dag sync process and I found out that we have to keep it in. Alternatively we need to change the dags sync logic to deal with conflicts.

internal/db/merge.go
```diff
@@ -43,28 +45,37 @@ func (db *db) handleMerges(ctx context.Context, merges events.Subscription[event
 				return
 			}
 			go func() {
-				err := db.executeMerge(ctx, merge)
+				// ensure only one merge per docID
+				queue.add(merge.DocID)
```
fredcarle, line 49: suggestion: If the DocID is now part of the message, we can get rid of the `getDocIDFromBlock` function. The reason I hadn't included DocID before is that I wanted to guarantee that there was no mistake with it. Like the wrong DocID for the provided CID. That was probably a bad reason when I think about it.

net/server.go
nasdf: Oh right I remember now. Would a simple fix be to ignore transaction conflict errors?

fredcarle: `GetBlocks` doesn't handle errors very well. Even `GetBlock` doesn't return the block on `put` conflict so we would need to retry. It would be an efficien retry because the block would be found locally on the next try. I'll let you decide what to do from here :)

nasdf: The updated sync process should now handle conflicts.

fredcarle (approved):
LGTM. Too bad we are reintroducing ipld-format after trying to get rid of it. We can switch over to an ipld-prime solution in the future.

internal/db/merge.go
shahzadlone: > If a user updates a doc while a merge is in progress it could still happen.

I'd be super nice / reminder to document that here.

Issue #2401, opened 2024-03-11
Unify PatchSchema and PatchCollection


AndrewSisley:
I think it would be nicer if there is just one entry point for this:

```
			testUtils.SchemaPatch{
				Patch: `
					[
						{ "op": "add", "path": "/Schema/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} },
				                // Make the second collection the active one by moving its name from the first to the second
						{ "op": "move", "from": "/Collections/1/Name", "path": "/Collections/2/Name" }
					]
				`,
			},
```

This may be fairly fiddly to achieve however, for example in the above collection `2` does not exist yet.  Have a long thing about the dev-ex of this.

fredcarle:
Since we are trying to separate the Schema and Collection within the code base, wouldn't unify this go against what has been done already?

AndrewSisley:
> Since we are trying to separate the Schema and Collection within the code base, wouldn't unifying this go against what has been done already?

I dont think so, it is one end point, but they are very clearly separate paths (`Schema` vs `Collection`).  It just means users can submit many commands in a single patch (including test commands).

fredcarle:
You're right. It's the same endpoint but the path is the differentiator. Thanks for pointing that out.

Issue #2568, opened 2024-04-29
Return all errors from validators


nasdf:
Validator functions should return a list of all errors encountered using `errors.Join` or similar.

Example: https://github.com/sourcenetwork/defradb/blob/develop/internal/db/collection.go#L655

PR #2720, opened 2024-06-14
refactor: Rework definition validation

### LINKED ISSUE #2537: Refactor Collection and Schema validation
At the moment all the rules are defined in the very large `db/collection.go` file, they should be moved somewhere else.

Validation is also done in different ways depending on entry point. Creating new collections via an SDL, and `PatchCollection` validation is mostly (but not quite) done the same way, and `PatchSchema` is done another way.  Many of the rules should be run for multiple cases (e.g. new collection via an SDL should pass most `PatchSchema` and `PatchCollection` rules).

I suggest all entry points use something similar to the existing `PatchCollection` setup, it is more readable, and easier to reuse rules than the `PatchSchema` way.

AndrewSisley:
## Relevant issue(s)

Resolves #2537

## Description

Reworks definition validation, standardizing the rule signatures, allowing rule reuse across different contexts, and hopefully improving their readability.  Performance of the rules will have decreased slightly, but on col/schema update that is unimportant, performance of `createCollections` (called when creating via SDL docs) has probably improved slightly due to a reduction in datastore calls.

This is largely in preparation for https://github.com/sourcenetwork/defradb/issues/2401 and the removal/simplification of `SetActiveVersion`.

### DIFF
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -36,9 +36,9 @@ func (db *db) createCollections(
 		return nil, err
 	}
 
-	for i, def := range newDefinitions {
-		txn := mustGetContextTxn(ctx)
+	txn := mustGetContextTxn(ctx)
 
+	for i, def := range newDefinitions {
 		schemaByName := map[string]client.SchemaDescription{}
 		for _, existingDefinition := range existingDefinitions {
 			schemaByName[existingDefinition.Schema.Name] = existingDefinition.Schema
@@ -57,7 +57,6 @@ func (db *db) createCollections(
 
 	for _, def := range newDefinitions {
 		desc := def.Description
-		txn := mustGetContextTxn(ctx)
 
 		if desc.Name.HasValue() {
 			exists, err := description.HasCollectionByName(ctx, txn, desc.Name.Value())
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -56,24 +56,10 @@ func (db *db) createCollections(
 	}
 
 	for _, def := range newDefinitions {
-		desc := def.Description
-
-		if desc.Name.HasValue() {
-			exists, err := description.HasCollectionByName(ctx, txn, desc.Name.Value())
-			if err != nil {
-				return nil, err
-			}
-			if exists {
-				return nil, ErrCollectionAlreadyExists
-			}
-		}
-
-		definitionsByName := map[string]client.CollectionDefinition{}
-		for _, existingDefinition := range existingDefinitions {
-			definitionsByName[existingDefinition.GetName()] = existingDefinition
-		}
-		for _, newDefinition := range newDefinitions {
-			definitionsByName[newDefinition.GetName()] = newDefinition
+		if len(def.Description.Fields) == 0 {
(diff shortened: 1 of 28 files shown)


internal/db/definition_validation.go
```diff
@@ -36,23 +106,53 @@ var patchCollectionValidators = []func(
...
+	newCols := make([]client.CollectionDescription, 0, len(newColsByID))
+	for _, col := range newColsByID {
+		newCols = append(newCols, col)
+	}
+
+	newState := newDefinitionState(newCols, map[string]client.SchemaDescription{})
+	oldState := newDefinitionState(oldCols, map[string]client.SchemaDescription{})
+
+	for _, validator := range updateValidators {
+		err := validator(ctx, db, newState, oldState)
```
nasdf, line 155: question: would it be possible to return all validator errors? It would be helpful to know all of the problems with a schema / collection.

AndrewSisley: Was this what you meant in https://github.com/sourcenetwork/defradb/issues/2568?

It should be fine and easy to do, although the resultant errors might be quite noisy (some errors will likely result in many other errors being returned).

Can we do this in another PR? Then we can test it properly (I want to avoid changing the tests as much as possible in this refactor PR)

nasdf: > Was this what you meant in https://github.com/sourcenetwork/defradb/issues/2568 (I wasn't sure if that ticket was for this, or some other validation somewhere)?

Yeah I was referencing that issue. 

> Can we do this in another PR/issue if we want it? Then we can test it properly (I want to avoid changing the tests as much as possible in this refactor PR)

Yeah that's fine with me. I think its purely a user experience issue and not relevant to this PR.

internal/db/collection_define.go
```diff
@@ -25,113 +25,126 @@ import (
...
-	}
-	for _, newDefinition := range newDefinitions {
-		definitionsByName[newDefinition.GetName()] = newDefinition
-	}
-	err = db.validateNewCollection(def, definitionsByName)
-	if err != nil {
-		return nil, err
-	}
+	for i, def := range newDefinitions {
+		schema := def.Schema
```
fredcarle, line 40: todo: Since schema is not used before being reassigned, please use `def.Schema` directly in `CreateSchemaVersion` bellow and instanciate `schema` at that time.

internal/db/collection_define.go
```diff
@@ -25,113 +25,126 @@ import (
...
+		txn := mustGetContextTxn(ctx)
```
fredcarle, line 41: todo: `mustGetContextTxn` should be called just once, prior to `getAllActiveDefinitions` above.

fredcarle (approved):
LGTM. Just a couple minor todos prior to merging.

internal/db/collection_define.go
AndrewSisley: Nice spot, removing.

- [ ] sort out `schema` vars

internal/db/collection_define.go
AndrewSisley: Will move it to just after `getAllActiveDefinitions`

[ ] Move txn declaration(s)

PR #2731, opened 2024-06-17
feat: Doc encryption with symmetric key

### LINKED ISSUE #2711: Doc. Encryption: Enable doc encryption with symmetric keys
Upon creating a document it should be possible to add an encryption key that will be used to encrypt doc's field values.
They keys will be stored locally.

The task doesn't include key exchange between peers.

islamaliev:
## Relevant issue(s)

Resolves #2711 #2808

## Description

This change introduces doc encryption. Upon creation of a document the user can pass the encryption flag which will signal to db to create a new symmetric key using AES-GCM and will store it locally.
With the encryption flag all doc fields (deltas) being stored in the DAG encrypted. The datastore will still store data as plain text, as for it's encryption we can use encryption-at-rest.
Decryption is not used at the moment, as it is relevant only p2p environment and we don't have yet key exchange mechanism. All peers sync encrypted data.

This PR also adds 2 new parameters to GraphQL `create_` mutation:
1. `inputs`: a list of documents to be created
2. `encrypt`: flag that indicates if document(s) need to be encrypted.

### DIFF
--- a/cli/collection_create.go
+++ b/cli/collection_create.go
@@ -17,6 +17,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/internal/db"
 )
 
 func MakeCollectionCreateCommand() *cobra.Command {
@@ -79,9 +80,8 @@ Example: create from stdin:
 				return cmd.Usage()
 			}
 
-			if encryptionKey != "" {
-				setContextDocEncryptionKey(cmd, encryptionKey)
-			}
+			txn, _ := db.TryGetContextTxn(cmd.Context())
+			setContextDocEncryptionKey(cmd, encryptionKey, txn)
 
 			if client.IsJSONArray(docData) {
 				docs, err := client.NewDocsFromJSON(docData, col.Definition())
--- a/cli/collection_create.go
+++ b/cli/collection_create.go
@@ -81,7 +81,7 @@ Example: create from stdin:
 			}
 
 			txn, _ := db.TryGetContextTxn(cmd.Context())
-			setContextDocEncryptionKey(cmd, shouldEncrypt, txn)
+			setContextDocEncryption(cmd, shouldEncrypt, txn)
 
 			if client.IsJSONArray(docData) {
 				docs, err := client.NewDocsFromJSON(docData, col.Definition())
--- a/cli/collection_create.go
+++ b/cli/collection_create.go
@@ -34,7 +34,7 @@ Options:
 		and permissions are controlled by ACP (Access Control Policy).
 
 	-e, --encrypt
-		Encrypt flag specified if the document needs to be encrypted. If set DefraDB will generate a
+		Encrypt flag specified if the document needs to be encrypted. If set, DefraDB will generate a
 		symmetric key for encryption using AES-GCM.
 
 Example: create from string:
@@ -99,7 +99,7 @@ Example: create from stdin:
 		},
 	}
 	cmd.PersistentFlags().BoolVarP(&shouldEncrypt, "encrypt", "e", false,
-		"Encryption key used to encrypt/decrypt the document")
+		"Flag to enable encryption of the document")
 	cmd.Flags().StringVarP(&file, "file", "f", "", "File containing document(s)")
 	return cmd
 }
--- a/cli/collection_create.go
+++ b/cli/collection_create.go
@@ -21,10 +21,20 @@ import (
 
 func MakeCollectionCreateCommand() *cobra.Command {
 	var file string
+	var encryptionKey string
 	var cmd = &cobra.Command{
-		Use:   "create [-i --identity] <document>",
+		Use:   "create [-i --identity] [-e --encryption] <document>",
 		Short: "Create a new document.",
 		Long: `Create a new document.
+		
+Options:
+    -i, --identity 
+        Marks the document as private and set the identity as the owner. The access to the document
+		and permissions are controlled by ACP (Access Control Policy).
+
+	-e, --encryption
+		Encrypts the document with the encryption key. The encryption key is used to encrypt and decrypt 
+		the document using symmetric AES-GCM encryption algorithm.
 
 Example: create from string:
   defradb client collection create --name User '{ "name": "Bob" }'
@@ -69,6 +79,10 @@ Example: create from stdin:
 				return cmd.Usage()
 			}
 
+			if encryptionKey != "" {
+				setContextDocEncryptionKey(cmd, encryptionKey)
+			}
+
 			if client.IsJSONArray(docData) {
 				docs, err := client.NewDocsFromJSON(docData, col.Definition())
 				if err != nil {
@@ -84,6 +98,8 @@ Example: create from stdin:
 			return col.Create(cmd.Context(), doc)
 		},
 	}
+	cmd.PersistentFlags().StringVarP(&encryptionKey, "encryption", "e", "",
+		"Encryption key used to encrypt/decrypt the document")
 	cmd.Flags().StringVarP(&file, "file", "f", "", "File containing document(s)")
 	return cmd
 }
--- a/cli/collection_create.go
+++ b/cli/collection_create.go
@@ -22,9 +22,9 @@ import (
 
 func MakeCollectionCreateCommand() *cobra.Command {
 	var file string
-	var encryptionKey string
+	var shouldEncrypt bool
 	var cmd = &cobra.Command{
-		Use:   "create [-i --identity] [-e --encryption] <document>",
+		Use:   "create [-i --identity] [-e --encrypt] <document>",
 		Short: "Create a new document.",
 		Long: `Create a new document.
 		
@@ -33,9 +33,9 @@ Options:
         Marks the document as private and set the identity as the owner. The access to the document
 		and permissions are controlled by ACP (Access Control Policy).
 
-	-e, --encryption
-		Encrypts the document with the encryption key. The encryption key is used to encrypt and decrypt 
-		the document using symmetric AES-GCM encryption algorithm.
+	-e, --encrypt
+		Encrypt flag specified if the document needs to be encrypted. If set DefraDB will generate a
+		symmetric key for encryption using AES-GCM.
 
 Example: create from string:
   defradb client collection create --name User '{ "name": "Bob" }'
@@ -81,7 +81,7 @@ Example: create from stdin:
 			}
 
 			txn, _ := db.TryGetContextTxn(cmd.Context())
-			setContextDocEncryptionKey(cmd, encryptionKey, txn)
+			setContextDocEncryptionKey(cmd, shouldEncrypt, txn)
 
 			if client.IsJSONArray(docData) {
 				docs, err := client.NewDocsFromJSON(docData, col.Definition())
@@ -98,7 +98,7 @@ Example: create from stdin:
 			return col.Create(cmd.Context(), doc)
 		},
 	}
-	cmd.PersistentFlags().StringVarP(&encryptionKey, "encryption", "e", "",
+	cmd.PersistentFlags().BoolVarP(&shouldEncrypt, "encrypt", "e", false,
 		"Encryption key used to encrypt/decrypt the document")
 	cmd.Flags().StringVarP(&file, "file", "f", "", "File containing document(s)")
 	return cmd
--- a/client/document.go
+++ b/client/document.go
@@ -122,7 +122,9 @@ var jsonArrayPattern = regexp.MustCompile(`^\s*\[.*\]\s*$`)
 
 // IsJSONArray returns true if the given byte array is a JSON Array.
 func IsJSONArray(obj []byte) bool {
-	return jsonArrayPattern.Match(obj)
+	var js []interface{}
+	err := json.Unmarshal(obj, &js)
+	return err == nil
 }
 
 // NewFromJSON creates a new instance of a Document from a raw JSON object byte array.
--- a/client/document.go
+++ b/client/document.go
@@ -13,6 +13,7 @@ package client
 import (
 	"encoding/json"
 	"errors"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -117,11 +118,11 @@ func NewDocFromMap(data map[string]any, collectionDefinition CollectionDefinitio
 	return doc, nil
 }
 
+var jsonArrayPattern = regexp.MustCompile(`(?s)^\s*\[.*\]\s*$`)
+
 // IsJSONArray returns true if the given byte array is a JSON Array.
 func IsJSONArray(obj []byte) bool {
-	var js []any
-	err := json.Unmarshal(obj, &js)
-	return err == nil
+	return jsonArrayPattern.Match(obj)
 }
 
 // NewFromJSON creates a new instance of a Document from a raw JSON object byte array.
--- a/client/document.go
+++ b/client/document.go
@@ -13,7 +13,6 @@ package client
 import (
 	"encoding/json"
 	"errors"
-	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -118,11 +117,9 @@ func NewDocFromMap(data map[string]any, collectionDefinition CollectionDefinitio
 	return doc, nil
 }
 
-var jsonArrayPattern = regexp.MustCompile(`^\s*\[.*\]\s*$`)
-
 // IsJSONArray returns true if the given byte array is a JSON Array.
 func IsJSONArray(obj []byte) bool {
-	var js []interface{}
+	var js []any
 	err := json.Unmarshal(obj, &js)
 	return err == nil
 }
--- a/datastore/multi.go
+++ b/datastore/multi.go
@@ -23,11 +23,13 @@ var (
 	headStoreKey   = rootStoreKey.ChildString("heads")
 	blockStoreKey  = rootStoreKey.ChildString("blocks")
 	peerStoreKey   = rootStoreKey.ChildString("ps")
+	encStoreKey    = rootStoreKey.ChildString("enc")
 )
 
 type multistore struct {
 	root   DSReaderWriter
 	data   DSReaderWriter
+	enc    DSReaderWriter
 	head   DSReaderWriter
 	peer   DSBatching
 	system DSReaderWriter
@@ -43,6 +45,7 @@ func MultiStoreFrom(rootstore ds.Datastore) MultiStore {
 	ms := &multistore{
 		root:   rootRW,
 		data:   prefix(rootRW, dataStoreKey),
+		enc:    prefix(rootRW, encStoreKey),
 		head:   prefix(rootRW, headStoreKey),
 		peer:   namespace.Wrap(rootstore, peerStoreKey),
 		system: prefix(rootRW, systemStoreKey),
@@ -55,6 +58,11 @@ func MultiStoreFrom(rootstore ds.Datastore) MultiStore {
 // Datastore implements MultiStore.
 func (ms multistore) Datastore() DSReaderWriter {
 	return ms.data
+}
+
+// Encstore implements MultiStore.
+func (ms multistore) Encstore() DSReaderWriter {
+	return ms.enc
 }
 
 // Headstore implements MultiStore.
--- a/datastore/store.go
+++ b/datastore/store.go
@@ -38,6 +38,10 @@ type MultiStore interface {
 	// under the /data namespace
 	Datastore() DSReaderWriter
 
+	// Encstore is a wrapped root DSReaderWriter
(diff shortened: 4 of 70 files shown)


cli/collection_create.go
```diff
@@ -17,14 +17,25 @@ import (
...
 	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/internal/db"
 )
 
 func MakeCollectionCreateCommand() *cobra.Command {
 	var file string
+	var encryptionKey string
 	var cmd = &cobra.Command{
-		Use:   "create [-i --identity] <document>",
+		Use:   "create [-i --identity] [-e --encryption] <document>",
```
fredcarle, line 27: todo: The encryption key shouldn't be passed on doc creation. In a local first context, we can tell Defra to encrypt the document. The `-e`  flag can be that signal. Defra will then create the encryption key, store it like you're doing now and encrypt the fields. In a future multi-user context, we wouldn't want to trust the node to use the encryption key that we pass to it so the document would have to be encrypted before prior to sending it.

internal/db/db.go
```diff
@@ -59,6 +59,8 @@ type db struct {
 
 	// The maximum number of retries per transaction.
 	maxTxnRetries immutable.Option[int]
+	
+	isEncrypted bool
```
fredcarle, line 63: question: Did you have a plan for this? It doesn't seem to be used.

internal/encryption/context.go
```diff
@@ -0,0 +1,69 @@
...
+	return enc.Encrypt(docID, fieldID, plainText)
+}
+
+func DecryptDoc(ctx context.Context, docID string, fieldID uint32, cipherText []byte) ([]byte, error) {
+	enc, ok := TryGetContextEncryptor(ctx)
+	if !ok {
+		return cipherText, nil
+	}
+	return enc.Decrypt(docID, fieldID, cipherText)
+}
```
fredcarle, line 69: todo: `EncryptDoc` and `DecryptDoc` should be moved to somewhere more relevant than the `context.go` file.

islamaliev: done

cli/collection_create.go
islamaliev: made defra generate keys

datastore/multi.go
```diff
@@ -23,11 +23,13 @@ var (
 	headStoreKey   = rootStoreKey.ChildString("heads")
 	blockStoreKey  = rootStoreKey.ChildString("blocks")
 	peerStoreKey   = rootStoreKey.ChildString("ps")
+	encStoreKey    = rootStoreKey.ChildString("enc")
 )
 
 type multistore struct {
 	root   DSReaderWriter
 	data   DSReaderWriter
+	enc    DSReaderWriter
```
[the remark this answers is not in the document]
islamaliev: it sounds better, but less relevant and 3 bytes longer.
I not happy myself with "enc" as it might not be immediatly clear, but tend to like it slightly more than "crypto". If other also see "crypto" being a better much (or anything else), no problem.

http/client_collection.go
```diff
@@ -125,6 +130,13 @@ func (c *Collection) CreateMany(
 	return nil
 }
 
+func setDocEncryptionHeaderIfNeeded(ctx context.Context, req *http.Request) {
```
nasdf, line 133: todo: This should be merged into `setDefaultHeaders` func in `http_client.go`

http/handler_collection.go
```diff
@@ -43,6 +46,11 @@ func (s *collectionHandler) Create(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	ctx := req.Context()
+	if req.Header.Get(DocEncryptionHeader) == "1" {
```
nasdf, line 50: todo: Since this header can be applied to any request this should be moved to a middleware.

http/client.go
```diff
@@ -355,6 +356,12 @@ func (c *Client) ExecRequest(
 		return result
 	}
 	err = c.http.setDefaultHeaders(req)
+
+	encConf := encryption.GetContextConfig(ctx)
+	if encConf.HasValue() && encConf.Value().IsEncrypted {
```
nasdf, line 361: thought: if the encrypted param only applies to one or two functions it may make more sense as a query parameter instead.

cli/collection_create.go
```diff
@@ -17,14 +17,25 @@ import (
...
 	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/internal/db"
 )
 
 func MakeCollectionCreateCommand() *cobra.Command {
 	var file string
+	var shouldEncrypt bool
 	var cmd = &cobra.Command{
-		Use:   "create [-i --identity] <document>",
+		Use:   "create [-i --identity] [-e --encrypt] <document>",
```
nasdf, line 27: question: should the query cli command also have an encrypted flag?

nasdf (changes requested):
Looks good! Just a few small things to clear up.

http/client_collection.go
islamaliev: all right

http/handler_collection.go
islamaliev: what do you mean by "any"? Are there headers that can't be applied?
This header is relevant only when creating docs. What's the point of moving it to a middleware?

http/client.go
islamaliev: do you mean as a parameter in POST request body?

cli/collection_create.go
islamaliev: the flag is relevant only when creating a doc. Later if we add it, for example, to update docs, then we can move it somewhere else

http/client.go
nasdf: No, I think a query param like `?encrypt=true` could be a better option here. I typically use headers for things that apply to a broad set of functions like identity and transactions.

http/handler_collection.go
nasdf: See my above comment about the query param.

islamaliev: done

cli/collection_create.go
```diff
@@ -17,14 +17,25 @@ import (
...
 		Long: `Create a new document.
+		
+Options:
+    -i, --identity 
+        Marks the document as private and set the identity as the owner. The access to the document
+		and permissions are controlled by ACP (Access Control Policy).
+
+	-e, --encrypt
+		Encrypt flag specified if the document needs to be encrypted. If set, DefraDB will generate a
+		symmetric key for encryption using AES-GCM.
```
shahzadlone, line 38: todo: consistent indentation please, can convert identity one to use tabs instead of spaces.

client/document.go
```diff
@@ -118,11 +117,11 @@ func NewDocFromMap(data map[string]any, collectionDefinition CollectionDefinitio
...
 }
 
-var jsonArrayPattern = regexp.MustCompile(`^\s*\[.*\]\s*$`)
-
 // IsJSONArray returns true if the given byte array is a JSON Array.
 func IsJSONArray(obj []byte) bool {
-	return jsonArrayPattern.Match(obj)
+	var js []any
+	err := json.Unmarshal(obj, &js)
+	return err == nil
```
shahzadlone, line 124: question: What made you do this?

islamaliev: this test wouldn't pass with original implementation https://github.com/sourcenetwork/defradb/pull/2731/files#diff-8c5dc24d796c5772fcbb82cb58fdfd8e6046ca461d3fae69bf934f67d21dac25R252-R257

shahzadlone: Is it because the pattern approach didn't account for a JSON array with whitespace?

shahzadlone: Use of `json.Unmarshal` here makes me feel like it's not that efficient as it probably uses reflect.

How about something like: 

```suggestion
	trimmedObj := bytes.TrimLeft(obj, " \r\t\n")
	return len(trimmedObj) > 0 && trimmedObj[0] == '['
```

islamaliev: I investigated some more why regexp didn't work for this case and found a good fix.
Thanks for pointing that out.

tests/integration/encryption/utils.go
```diff
@@ -0,0 +1,29 @@
...
+package encryption
+
+import (
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+const userCollectionGQLSchema = (`
+	type Users {
+		name: String
+		age: Int @crdt(type: "lww")
```
AndrewSisley, line 20: question: Why the explicit `@crdt(type: "lww")` - is it required?

tests/integration/utils2.go
```diff
@@ -1204,90 +1207,136 @@ func createDoc(
...
 	action CreateDoc,
 	node client.DB,
 	collections []client.Collection,
-) (*client.Document, error) {
+) ([]*client.Document, error) {
 	collection := collections[action.CollectionID]
-	var err error
 	var input string
 
+	paramName := "input"
```
AndrewSisley, line 1305: suggestion: I suggest using `request.Input` instead of redefining the const

tests/integration/utils2.go
```diff
@@ -1204,90 +1207,136 @@ func createDoc(
...
+
+	var err error
 	if action.DocMap != nil {
 		input, err = valueToGQL(action.DocMap)
+	} else if client.IsJSONArray([]byte(action.Doc)) {
+		var docMaps []map[string]any
+		err = json.Unmarshal([]byte(action.Doc), &docMaps)
+		require.NoError(s.t, err)
+		paramName = "inputs"
```
AndrewSisley, line 1314: todo: This should really be defined in client/request/const.go IMO

tests/integration/utils2.go
```diff
@@ -1204,90 +1207,136 @@ func createDoc(
...
+		input, err = arrayToGQL(docMaps)
 	} else {
 		input, err = jsonToGQL(action.Doc)
 	}
 	require.NoError(s.t, err)
 
+	params := paramName + ": " + input
+
+	if action.IsEncrypted {
+		params = params + ", encrypt: true"
```
AndrewSisley, line 1324: todo: `"encrypt"` should really be defined in client/request/const.go IMO

tests/integration/utils2.go
```diff
@@ -1302,14 +1351,20 @@ func createDocViaGQL(
...
 
-	docIDString := resultantDocs[0]["_docID"].(string)
-	docID, err := client.NewDocIDFromString(docIDString)
-	require.NoError(s.t, err)
+	docs := make([]*client.Document, len(resultantDocs))
 
-	doc, err := collection.Get(ctx, docID, false)
-	require.NoError(s.t, err)
+	for i, docMap := range resultantDocs {
+		docIDString := docMap["_docID"].(string)
```
AndrewSisley, line 1357: suggestion: Sorry about the bother, as it was a string here before, but please use `request.DocIDFieldName` instead of redefining the string value

internal/planner/create.go
```diff
@@ -56,76 +56,65 @@ func (n *createNode) Kind() string { return "createNode" }
...
-	n.currentValue = currentValue
-
-	desc := n.collection.Description()
-	docID := base.MakeDataStoreKeyWithCollectionAndDocID(desc, currentValue.GetID())
-	n.results.Spans(core.NewSpans(core.NewSpan(docID, docID.PrefixEnd())))
-
-	err := n.results.Init()
-	if err != nil {
-		return false, err
+	if len(n.docs) == 1 {
```
AndrewSisley, line 88: question: Why do you think this check is worth it?  Why not always call `CreateMany`?

internal/planner/create.go
```diff
@@ -56,76 +56,65 @@ func (n *createNode) Kind() string { return "createNode" }
...
+	for i, doc := range docs {
+		docIDs[i] = doc.ID().String()
 	}
+	return docIDs
+}
 
-	currentValue := n.documentMapping.NewDoc()
+func (n *createNode) Start() (err error) {
+	n.docs = make([]*client.Document, len(n.input))
+	defer func() { n.err = err }()
```
AndrewSisley, line 78: todo: Please remove `n.err` completely, it is not having a positive impact on our codebase, and as far as I can tell, it has no purpose.

I assume it is a relic from before `node.Start` could return errors.

cli/collection_create.go
AndrewSisley, line 38: praise: Thanks for these Islam, they are very clear :)

tests/integration/encryption/query_test.go
```diff
@@ -0,0 +1,115 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestDocEncryption_WithEncryptionOnCounterCRDT_ShouldFetchDecrypted(t *testing.T) {
+	const docID = "bae-ab8ae7d9-6473-5101-ba02-66b217948d7a"
```
AndrewSisley, line 60: question: Do we care about the doc id in this test?  Can it not be removed?

tests/integration/encryption/peer_test.go
```diff
@@ -0,0 +1,147 @@
...
+							}
+						}
+					}
+				`,
+				Results: []map[string]any{
+					{
+						"cid":          "bafyreih7ry7ef26xn3lm2rhxusf2rbgyvl535tltrt6ehpwtvdnhlmptiu",
+						"collectionID": int64(1),
+						"delta":        encrypt(testUtils.CBORValue(21)),
+						"docID":        "bae-c9fb0fa4-1195-589c-aa54-e68333fb90b3",
```
AndrewSisley, line 106: suggestion: I don't think we care about the docID value in this test?  Can we perhaps use `testUtils.NewDocIndex(...)` instead?

tests/integration/encryption/commit_test.go
```diff
@@ -0,0 +1,359 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestDocEncryption_UponEncryptionSeveralDocs_ShouldStoreAllCommitsDeltaEncrypted(t *testing.T) {
+	const johnDocID = "bae-c9fb0fa4-1195-589c-aa54-e68333fb90b3"
```
AndrewSisley, line 295: suggestion: I don't think we care about the docID value in these commit tests?  Can we perhaps use `testUtils.NewDocIndex(...)` instead?

AndrewSisley (commented):
Part way through my review but need to get some food - looks good so far :)

internal/planner/create.go
```diff
@@ -56,76 +56,65 @@ func (n *createNode) Kind() string { return "createNode" }
...
-	docID := base.MakeDataStoreKeyWithCollectionAndDocID(desc, currentValue.GetID())
-	n.results.Spans(core.NewSpans(core.NewSpan(docID, docID.PrefixEnd())))
-
-	err := n.results.Init()
-	if err != nil {
-		return false, err
+	if len(n.docs) == 1 {
+		err = n.collection.Create(n.p.ctx, n.docs[0])
+	} else {
+		err = n.collection.CreateMany(n.p.ctx, n.docs)
```
AndrewSisley, line 91: todo: I do disagree with creating docs in `Start` - it feels like it is very much in the wrong place and may trip us up in the future (if not by being technically annoying for some future change, it seems likely to cause 'surprise').

Please either move this call into `Next`, or discuss in in discord/standup (for greater visibility).

internal/core/block/block.go
```diff
@@ -103,6 +103,8 @@ type Block struct {
 	Delta crdt.CRDT
 	// Links are the links to other blocks in the DAG.
 	Links []DAGLink
+	// IsEncrypted is a flag that indicates if the block's delta is encrypted.
+	IsEncrypted *bool
```
AndrewSisley, line 107: todo: Please don't use a pointer for a boolean, it feels very wasteful and confusing - suggesting that it's pointerness may be used for odd mutation reasons.  If you want it to be optional, please use `immutable.Option` instead, it is much cheaper, less confusing, and wont contribute to GC pressure.

If you have a very specific reason for using a pointer here, please document it.

datastore/store.go
```diff
@@ -38,6 +38,10 @@ type MultiStore interface {
 	// under the /data namespace
 	Datastore() DSReaderWriter
 
+	// Encstore is a wrapped root DSReaderWriter
+	// under the /enc namespace
+	Encstore() DSReaderWriter
```
AndrewSisley, line 43: todo: I know the documentation for the other members of this interface isn't great, but their names are a little more self explanatory.  Can you please document what this store is?  And perhaps rename it to something a little clearer - `Enc` is not really self documenting outside the context of this PR, and this is a public type.

AndrewSisley (changes requested):
Looks really good, just a handful of smaller comments for you before merge :)

nasdf (approved):

datastore/store.go
islamaliev: I'm not 100% happy myself with the name, but didn't find any better alternative. `EncryptionStore` seems to be too long. Fred [proposed](https://github.com/sourcenetwork/defradb/pull/2731#discussion_r1643153189) `CryptoStore`. Feel free to propose something better.

cli/collection_create.go
islamaliev: done

tests/integration/utils2.go
islamaliev: done

tests/integration/utils2.go
islamaliev: done

tests/integration/utils2.go
islamaliev: they are defined. Just aren't used here. Changed

tests/integration/utils2.go
islamaliev: done

internal/planner/create.go
islamaliev: no strong reason. Just didn't want to change much the previous behaviour.

datastore/store.go
AndrewSisley: I'm happy with the length of `EncryptionStore`, although I'm not really happy with that name anyway (is still vague and is probably more likely to be read as a store of encrypted data than a store of keys).

For me the most important is the documentation request - please see the name change as a suggestion.

tests/integration/encryption/utils.go
AndrewSisley: Ah okay makes sense - can you please add a comment explain that, else it is likely to be removed or cause head scratching at a later date :)

internal/planner/create.go
AndrewSisley: Ah, in that case can I ask you to just call `CreateMany`? Otherwise it is a bit weird and looks like an unnecessary code complication that will (at best) be removed next time someone has to change something in this function.

internal/planner/create.go
islamaliev: `n.err` is checked in `Next`. It was like this before (but less consistent). Not sure it there was a good reason for this.

tests/integration/encryption/query_test.go
islamaliev: not really. Removed

tests/integration/encryption/peer_test.go
islamaliev: done

tests/integration/encryption/commit_test.go
islamaliev: done

internal/core/block/block.go
islamaliev: documented

internal/planner/create.go
islamaliev: done

tests/integration/encryption/utils.go
islamaliev: done

internal/planner/create.go
islamaliev: removed

AndrewSisley: cheers :)

datastore/store.go
islamaliev: added documentation

internal/planner/create.go
islamaliev: I understand your concern. I had similar feelings but thought it was highly unlikely and having a nicer code was worth it. If you think otherwise, I will move it to `Next`. No problem.

AndrewSisley: Thanks Islam, sorry about the bother, and I understand it is a bit 'abstract' in this case, but I do think it is in the wrong place atm.  Ideally we would remove the Start/Init funcs anyway, and conceptually I very much think the 'creating' part of a 'create' iterator should be done in the iteration funcs (next/value).

AndrewSisley (approved):
LGTM Islam :)  And thanks for the random little cleanups you made along the way :)

PR #2732, opened 2024-06-17
ci: Cache dependencies to speed up test runs

### LINKED ISSUE #2735: If no dependencies have changes cache build state for faster action run
If no dependencies have changes cache build state for faster action run

shahzadlone:
## Relevant issue(s)
Resolves #2735 

## Description
In this PR we cache go and rust build/dep states, and restore them if an action run has same dependency graph. Go cache is around 877 MB and cargo/rust cache is about 189MB (99MB for macos) so we can roughly have room for 5 caches (x1 for linux and x1 for macos) before a cache is dropped / lost.

Notes:
- Github gives us 10GB of cache storage by default.
- Rather than going with the standard practice of using lockfiles to generate the caching key we will use the mod and toml files, in order to maximize cache hits. Incase troublesome we can go back to lockfiles.
- This change does not help the change detector. IMO change detector should move to a separate job, or a new workflow file. (opened https://github.com/sourcenetwork/defradb/issues/2736)

### Demo
- Without cache run (current state): https://github.com/sourcenetwork/defradb/actions/runs/9554784255?pr=2732

- With cache run (after this PR): https://github.com/sourcenetwork/defradb/actions/runs/9555174699

## How has this been tested?
- Manual
- `act` tool

### DIFF
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -110,6 +110,45 @@ jobs:
           go-version: "1.21"
           check-latest: true
 
+      - name: Set cache paths
+        id: cache-paths
+        shell: bash
+        run: |
+          echo "GO_CACHE=$(go env GOCACHE)" >> "${GITHUB_OUTPUT}"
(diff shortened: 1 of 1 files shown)


AndrewSisley (approved):
LGTM!  Thanks a whole load for this Shahzad :)

PR #2743, opened 2024-06-20
docs(i): Update contributors

nasdf:


README.md
```diff
@@ -479,5 +479,6 @@ DefraDB's code is released under the [Business Source License (BSL)](licenses/BS
 - Orpheus Lummis ([@orpheuslummis](https://github.com/orpheuslummis))
 - Fred Carle ([@fredcarle](https://github.com/fredcarle))
 - Islam Aliev ([@islamaliev](https://github.com/islamaliev))
+- Keenan Nemetz ([@nasdf](https://github.com/nasdf))
```
AndrewSisley, line 482: thought: Way overdue :laughing: - however, do we really need this section?  Contributors are listed in github, and overtime manually growing this list will not scale well.

AndrewSisley (approved):
LGTM :grin:

README.md
nasdf: could probably rename it to maintainers since there's a larger difference between contributors and active maintainers

AndrewSisley: Might be a good idea, although that would mean removing Orpheus from the list too

PR #2748, opened 2024-06-21
refactor: Network test sync logic

### LINKED ISSUE #2747: Improve network test sync logic
The network sync logic could be improved by keeping a list of document heads that should be merged into target nodes. This will allow network events to happen in any order and also allow republish / retry logic to be added to the network implementation.

nasdf:
## Relevant issue(s)

Resolves #2747

## Description

This PR improves the network sync test logic to allow for further improvements to the network implementation.

## Tasks

## How has this been tested?

`make test`

- MacOS


tests/integration/p2p.go
```diff
@@ -451,7 +309,61 @@ func getAllP2PCollections(
...
+		return // skip network testing
+	}
+
+	ctx, cancel := context.WithTimeout(s.ctx, subscriptionTimeout*10)
+	defer cancel()
+
+	for i := 0; i < len(s.nodes); i++ {
+		if nodeID.HasValue() && nodeID.Value() != i {
+			continue // node is not selected
+		}
```
fredcarle, line 330: suggestion: The documentation says `waits for all selected nodes` but the function only receives one. You can probably do away with the loop and just get the node directly.

tests/integration/p2p.go
```diff
@@ -451,7 +309,61 @@ func getAllP2PCollections(
...
+			if _, ok := s.actualDocHeads[id][evt.DocID]; ok {
+				s.expectedDocHeads[id][evt.DocID] = evt.Cid
+			}
+		}
+		// update the expected document heads of replicator sources
+		for id := range s.nodeReplicatorTargets[i] {
+			if _, ok := s.actualDocHeads[id][evt.DocID]; ok {
+				s.expectedDocHeads[id][evt.DocID] = evt.Cid
+			}
+		}
```
fredcarle, line 354: thougth: It's odd that we would be updating the expected heads only if already present in the actual heads. Shouldn't it be updated regardless?

tests/integration/p2p.go
nasdf: If `nodeID` has no value it will wait for all nodes.

tests/integration/p2p.go
nasdf: I believe the code is correct based on what we discussed regarding replicator target nodes sending updates of a document back to the replicator source node.

tests/integration/state.go
```diff
@@ -50,11 +51,41 @@ type state struct {
...
 
-	// These synchronisation channels allow async actions to track their completion.
-	syncChans []chan struct{}
+	// nodeMergeCompleteSubs is a list of all merge complete event subscriptions
+	nodeMergeCompleteSubs []*event.Subscription
 
-	// eventSubs is a list of all event subscriptions
-	eventSubs []*event.Subscription
+	// nodeUpdateSubs is a list of all update event subscriptions
+	nodeUpdateSubs []*event.Subscription
```
AndrewSisley, line 58: suggestion: Can we wrap these up in a dedicated type?  They are consuming a disproportionately large amount of attention on the state struct compared to their visibility needs.

tests/integration/utils2.go
```diff
@@ -1205,6 +1205,10 @@ func createDoc(
 		s.documents = append(s.documents, make([][]*client.Document, action.CollectionID-len(s.documents)+1)...)
 	}
 	s.documents[action.CollectionID] = append(s.documents[action.CollectionID], doc)
+
+	if action.ExpectedError == "" {
+		waitForUpdateEvents(s, action.NodeID, action.CollectionID)
```
AndrewSisley, line 1210: thought: This is a significant change in the tests - and one that looks like it removes our ability to test concurrent P2P stuff (something that we very much want/need to test)?

tests/integration/utils2.go
```diff
@@ -1355,6 +1359,10 @@ func deleteDoc(
 	}
 
 	assertExpectedErrorRaised(s.t, s.testCase.Description, action.ExpectedError, expectedErrorRaised)
+
+	if action.ExpectedError == "" {
+		waitForUpdateEvents(s, action.NodeID, action.CollectionID)
```
AndrewSisley, line 1364: question: Why is this only done using create-save, and not create-gql and the other create?  And why not on delete?

thought: This looks quite easy to mess up and forget, and it missing it might not create an obvious and immediate failure (just flaky tests).

AndrewSisley (changes requested):
I've not finished reviewing, but want to block the merge until my concerns around the apparent loss of concurrency have been resolved.  The code looks simpler, but I think we may have lost the ability to test some important things.

AndrewSisley:
todo: Please make sure you run the P2P tests a lot (IIRC Fred/I had to run them 100 or so times to reproduce the last solved issues) before merging this, we had flakiness problems with this before and they cost a lot of time.

tests/integration/utils2.go
nasdf: The function name is a bit misleading, but `Update` events are published immediately after a document is saved and shouldn't really change how we test things concurrently. 

We could also merge `waitForUpdateEvents` into `waitForSync` since the events are buffered if you would prefer that.

AndrewSisley: Ah I see what you mean - it is not waiting for the sync, just the event publishing to the current node's event queue?

Why are you calling it then?

nasdf: > Ah I see what you mean - it is not waiting for the sync, just the event publishing to the current node's event queue?

Yes

> Why are you calling it then?

Its a convenient way to track which doc heads are expected to be merged on remote nodes.

AndrewSisley: > Its a convenient way to track which doc heads are expected to be merged on remote nodes.

Can you please document that really well?  Our integration tests are already super magic and far more complex than most, and that complexity is not really tested.  The P2P space is particularly complicated and prone to annoying, late-to-discover, flaky bugs and it is well worth over-explaining things to save future hassle.

tests/integration/utils2.go
```diff
@@ -781,6 +766,10 @@ func refreshCollections(
 ) {
 	s.collections = make([][]client.Collection, len(s.nodes))
 
+	for i := len(s.nodePeerCollections); i < len(s.collectionNames); i++ {
+		s.nodePeerCollections = append(s.nodePeerCollections, make(map[int]struct{}))
```
AndrewSisley, line 770: todo: This looks like it will break when collections are removed/deactivated?  Users currently can do this, although we likely have no P2P tests for it atm.

There may be other test things that currently break when doing that, but we should avoid adding new ones.

tests/integration/utils2.go
```diff
@@ -739,33 +751,6 @@ func restartNodes(
...
-			waitGroupStartIndex = i + 1
-			break actionLoop
-		}
-	}
-
-	for _, tc := range s.testCase.Actions {
-		switch action := tc.(type) {
-		case ConnectPeers:
-			// Give the nodes a chance to connect to each other and learn about each other's subscribed topics.
-			time.Sleep(100 * time.Millisecond)
```
AndrewSisley, line 760: praise: I forgot we were currently relying on a sleep - thank you very much for removing this!

tests/integration/p2p.go
```diff
@@ -479,7 +337,61 @@ func getAllP2PCollections(
 	assert.Equal(s.t, expectedCollections, cols)
 }
 
-// waitForSync waits for all given wait channels to receive an item signaling completion.
+// waitForUpdateEvents waits for all selected nodes to publish an update event.
```
AndrewSisley, line 340: suggestion: Building on the other related documentation request, it might be worth highlighting that you mean publish to the local event queue, not to the pub/sub net or some other P2P related network thing.

tests/integration/utils2.go
shahzadlone: +1 similar thought, for here and other instances of this change.

tests/integration/state.go
nasdf: Done

tests/integration/utils2.go
nasdf: I added more documentation about how the document head are tracked. It should also be much easier to follow since its all moved to an `events.go` file. Let me know what you think.

tests/integration/utils2.go
nasdf: If I'm understanding this correctly the `createDoc` function calls the `create-save` and `create-gql` functions so they both should be covered.

tests/integration/state.go
```diff
@@ -50,11 +51,41 @@ type state struct {
...
+	//
+	// The index of the slice is the node id. The map key is the connected node id.
+	nodeConnections []map[int]struct{}
+
+	// nodeReplicatorSources contains all active replicators.
+	//
+	// The index of the slice is the source node id. The map key is the target node id.
+	nodeReplicatorSources []map[int]struct{}
+
+	// nodeReplicatorTargets contains all active replicators.
```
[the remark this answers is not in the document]
nasdf: Done

tests/integration/utils2.go
nasdf: This should be easy to handle. Just need to find which collections to deactivate from the p2p state.

- [ ] handle patchCollection

tests/integration/events.go
```diff
@@ -0,0 +1,241 @@
...
+// waitForNetworkSetupEvents waits for p2p topic completed and
+// replicator completed events to be published on the local node event bus.
+func waitForNetworkSetupEvents(s *state, nodeID int) {
+	cols, err := s.nodes[nodeID].GetAllP2PCollections(s.ctx)
+	require.NoError(s.t, err)
+
+	reps, err := s.nodes[nodeID].GetAllReplicators(s.ctx)
+	require.NoError(s.t, err)
+
+	p2pTopicEvents := 0
```
AndrewSisley, line 36: question: It looks like `p2pTopicEvents` is only ever 0 or 1, why is it not a boolean? Is this a bug/wip?

tests/integration/utils2.go
AndrewSisley: Changes look really good, thanks Keenan.  Has got me thinking about test Cid substitution too, in the future we might want to tweak `state.nodeP2P.DocHeads` to allow this.

tests/integration/utils2.go
AndrewSisley: :grin: My bad, the github presentation of the diff helped me miss-read where this line was...  Ignore this thread :)

tests/integration/state.go
```diff
@@ -24,6 +25,86 @@ import (
 	"github.com/sourcenetwork/defradb/tests/clients"
 )
 
+// p2pState contains all p2p related testing test.
```
AndrewSisley, line 28: nitpick: `related testing state.`?

AndrewSisley (approved):
Looks a really nice improvement, thanks Keenan :)

tests/integration/utils2.go
```diff
@@ -342,7 +341,7 @@ func performAction(
 		assertClientIntrospectionResults(s, action)
 
 	case WaitForSync:
-		waitForSync(s, action)
+		waitForMergeEvents(s, action)
```
islamaliev, line 344: praise: it's great that now waiting is done on the same thread. Earlier it used to call a goroutine for every `ConnectPeers` action and had subtle shortcomings.

tests/integration/utils2.go
```diff
@@ -1457,6 +1424,10 @@ func updateDoc(
 	}
 
 	assertExpectedErrorRaised(s.t, s.testCase.Description, action.ExpectedError, expectedErrorRaised)
+
+	if action.ExpectedError == "" {
+		waitForUpdateEvents(s, action.NodeID, action.CollectionID)
```
islamaliev, line 1429: questions: this looks like unsolicited sync. Why is it unconditional?

islamaliev (approved):
Looks good. Core looks now much more intuitive. I just a have a question to clarify.

tests/integration/test_case.go
```diff
@@ -349,8 +346,9 @@ type UpdateDoc struct {
 	// contains this string.
 	ExpectedError string
 
-	// Setting DontSync to true will prevent waiting for that update.
-	DontSync bool
+	// Skip waiting for an update event. This should only be used for
+	// tests that do not correctly publish an update event.
+	SkipUpdateEvent bool
```
islamaliev, line 351: suggestion: why is it synced by default now? That resembles sequentially consistent memory order instead of relaxed one that we used before.
I think, if we want to sync, we should request it explicitly. Unsynced actions are closer to realistic behaviour.

AndrewSisley: todo: I think this is a misleading name and comment, IIRC `waitForUpdateEvents` does not wait for the P2P sync, but only the update event to be published to the event bus.  If that is correct, @nasdf can you please rework the property name and documentation to make that clearer.

islamaliev: what is this then if not wait: `<-s.nodeEvents[i].update.Message():`?

AndrewSisley: That is just the event bus, it is written to synchronously by the database when mutating documents - the P2P code also subscribes to this bus - so 'waiting' to the bus should never really block, and the P2P stuff hasn't even been sent over the network at that point.

If I understand it correctly.  I made the same comment as you did earlier for `waitForUpdateEvents` - I think Keenan improved the documentation and code location, but given that it is still tripping up reviewers, that too should probably be improved further.

tests/integration/utils2.go
nasdf: There should always be an update event published to the local event bus if a document is updated or created.

tests/integration/state.go
nasdf: fixed

tests/integration/events.go
nasdf: Changed to bool

shahzadlone (approved):
Good Stuff 👍

PR #2768, opened 2024-06-25
refactor: Decouple client.DB from net

### LINKED ISSUE #2767: Decouple `client.DB` from `net`

fredcarle:
## Relevant issue(s)

Resolves #2767 

## Description

This PR make the `net` package no longer dependent on `client.DB`. Instead, on creation, the peer takes in a rootstore, a blockstore and an event bus.

The p2p collections and replicators have been moved to `db`. Most of the code is reused but it now makes use of the event bus to update the peer's related state (pubsub topics and peer connections).

## Tasks

## How has this been tested?

make test

- MacOS


datastore/store.go
```diff
@@ -24,8 +24,8 @@ var (
 	log = corelog.NewLogger("store")
 )
 
-// RootStore wraps Batching and TxnDatastore requiring datastore to support both batching and transactions.
-type RootStore interface {
+// Rootstore wraps Batching and TxnDatastore requiring datastore to support both batching and transactions.
+type Rootstore interface {
```
shahzadlone, line 28: question: What's wrong with `RootStore`?

datastore/mocks/utils.go
```diff
@@ -73,7 +73,7 @@ func NewTxnWithMultistore(t *testing.T) *MultiStoreTxn {
 	result := &MultiStoreTxn{
 		Txn:             txn,
 		t:               t,
-		MockRootstore:   prepareRootStore(t),
+		MockRootstore:   prepareRootstore(t),
 		MockDatastore:   prepareDataStore(t),
```
shahzadlone, line 77: todo: Shouldn't the other `perpareXYZ` names change too?

net/node_test.go
shahzadlone, line 1: todo: The contents of this test file should likely move to `peer_test.go` file now.

shahzadlone (commented):
suggestion: Also rename `func (c *Client) Root() datastore.Rootstore {` in `http/client.go` ?

internal/db/p2p_replicator.go
```diff
@@ -0,0 +1,341 @@
...
+		if err != nil {
+			return err
+		}
+		oldRep := client.Replicator{}
+		err = json.Unmarshal(repBytes, &oldRep)
+		if err != nil {
+			return err
+		}
+		for _, schema := range oldRep.Schemas {
+			oldSchemas[schema] = struct{}{}
```
shahzadlone, line 66: question: Is this new code or old/moved code?

internal/db/p2p_replicator.go
```diff
@@ -0,0 +1,341 @@
...
+		oldSchemas = make(map[string]struct{})
+	}
+
+	rep.Schemas = nil
+	for _, col := range collections {
+		delete(oldSchemas, col.SchemaRoot())
+	}
+	for schema := range oldSchemas {
+		rep.Schemas = append(rep.Schemas, schema)
+	}
```
shahzadlone, line 267: suggestion: If new code, one liner comment would be useful for what/why of this code.

internal/db/p2p_replicator.go
```diff
@@ -0,0 +1,341 @@
...
+func (db *db) loadReplicators(ctx context.Context) error {
+	replicators, err := db.GetAllReplicators(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, rep := range replicators {
+		schemaMap := make(map[string]struct{})
+		for _, schema := range rep.Schemas {
+			schemaMap[schema] = struct{}{}
```
shahzadlone, line 333: suggestion: One liner comment on the nested for-loops will be handy IMO

internal/db/p2p_replicator.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+		db.events.Publish(event.NewMessage(event.ReplicatorName, event.Replicator{
+			Info:    rep.Info,
+			Schemas: schemaMap,
+		}))
```
shahzadlone, line 338: question: Not tested?

internal/db/p2p_schema_root.go
```diff
@@ -205,3 +176,52 @@ func (p *Peer) GetAllP2PCollections(ctx context.Context) ([]string, error) {
...
+	if err != nil {
+		return err
+	}
+	db.events.Publish(event.NewMessage(event.P2PTopicName, event.P2PTopic{
+		ToAdd: schemaRoots,
+	}))
+
+	colMap := make(map[string]struct{})
+	for _, schemaRoot := range schemaRoots {
+		colMap[schemaRoot] = struct{}{}
```
shahzadlone, line 200: suggestion: One liner comment please.

shahzadlone:
I do like the changes but there seems to be a massive hit (-4%) to code coverage in this PR:

image

internal/db/p2p_schema_root.go
```diff
@@ -205,3 +176,52 @@ func (p *Peer) GetAllP2PCollections(ctx context.Context) ([]string, error) {
...
+func (db *db) PeerInfo() peer.AddrInfo {
+	db.peerMutex.RLock()
+	defer db.peerMutex.RUnlock()
+	if db.peerInfo.HasValue() {
+		return db.peerInfo.Value()
+	}
+	return peer.AddrInfo{}
+}
+
+func (db *db) loadP2PCollections(ctx context.Context) error {
```
islamaliev, line 189: todo: this method's name is misleading. It's not just loading, but fires a bunch of events.

I think it should be renamed to something more appropriate.

tests/integration/utils2.go
```diff
@@ -652,7 +652,7 @@ func setStartingNodes(
 		db, path, err := setupDatabase(s)
```
islamaliev, line 652: todo: rename please `db` to `node`

tests/integration/db.go
```diff
@@ -100,7 +100,7 @@ func NewBadgerFileDB(ctx context.Context, t testing.TB) (client.DB, error) {
 // setupDatabase returns the database implementation for the current
 // testing state. The database type on the test state is used to
 // select the datastore implementation to use.
-func setupDatabase(s *state) (client.DB, string, error) {
+func setupDatabase(s *state) (*node.Node, string, error) {
```
islamaliev, line 103: suggestion: maybe we should call it `setupNode`?

islamaliev (approved):
Looks great. Thanks for moving stuff around, this definitely improves the code quality. I just have few minor suggestions/concerns.

internal/db/p2p_schema_root.go
```diff
@@ -205,3 +176,52 @@ func (p *Peer) GetAllP2PCollections(ctx context.Context) ([]string, error) {
 
 	return collectionIDs, nil
 }
+
+func (db *db) PeerInfo() peer.AddrInfo {
+	db.peerMutex.RLock()
+	defer db.peerMutex.RUnlock()
```
AndrewSisley, line 182: todo: Why are you locking here?  Given that you are returning `db.peerInfo` outside of the `RUnlock` call, the lock appears pointless and creates a false sense of thread-safety.

Please either document why you are locking here, or fix the code so that it is obviously correct/expected (and probably documented on the `PeerInfo()` func signature)

internal/db/messages.go
```diff
@@ -0,0 +1,81 @@
...
+					if err != nil {
+						log.ErrorContextE(
+							ctx,
+							"Failed to execute merge",
+							err,
+							corelog.Any("Event", evt))
+					}
+				}()
+			case event.PeerInfo:
+				db.peerMutex.Lock()
```
AndrewSisley, line 63: todo: This mutex actually seems to be the wrong tool for the job, as this is the only instance I can spot with a write lock - can you please remove the mutex, and instead use an `atomic` type for `db.peerInfo`

internal/db/messages.go
```diff
@@ -0,0 +1,81 @@
...
+				db.peerInfo = immutable.Some(evt.Info)
+				db.peerMutex.Unlock()
+				go func() {
```
AndrewSisley, line 66: todo: Please document why these two go routines are spun up

fredcarle:
> I do like the changes but there seems to be a massive hit (-4%) to code coverage in this PR:

That can't be right. I've seen that in an other PR lately as well. This one for example: https://github.com/sourcenetwork/defradb/pull/2748. I think it's a codcov bug.

AndrewSisley:
> > I do like the changes but there seems to be a massive hit (-4%) to code coverage in this PR:
> 
> That can't be right. I've seen that in an other PR lately as well. This one for example: #2748. I think it's a codcov bug.

A lot of the CI in this PR is failing atm, it might be caused by that (failing tests hopefully dont contribute to code coverage)

datastore/mocks/utils.go
fredcarle: I'm changing the ones I touched. I'll work on converting the rest in a separate PR.

net/node_test.go
fredcarle: Good point:

-[ ] Move `node_test.go` to `peer_test.go`

datastore/store.go
shahzadlone: > libp2p and ipld use `Datastore`, `Peerstore`, `Blockstore`, etc. and I'm getting annoyed by the mix of `DataStore` vs `Datastore` so I'm starting to standardize it all to the libp2p and ipld way.

Ah I see, If that's the case makes sense to change :) thanks

internal/db/p2p_replicator_test.go
```diff
@@ -0,0 +1,224 @@
...
+	})
+	require.NoError(t, err)
+	for msg := range sub.Message() {
+		if msg.Name == event.ReplicatorName {
+			replicator := msg.Data.(event.Replicator)
+			require.Equal(t, peer.ID("other"), replicator.Info.ID)
+			require.Equal(t, map[string]struct{}{schema.Root: {}}, replicator.Schemas)
+			break
+		}
+	}
```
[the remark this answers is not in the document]
fredcarle: Good point.

- [ ] Verify we actually received the message

internal/db/p2p_replicator.go
```diff
@@ -0,0 +1,341 @@
...
+		key := core.NewReplicatorKey(rep.Info.ID.String())
+		_, err = txn.Systemstore().Get(ctx, key.ToDS())
+		if err != nil {
+			return err
+		}
+	} else {
+		oldSchemas = make(map[string]struct{})
+	}
+
+	rep.Schemas = nil
```
AndrewSisley, line 261: todo: Please document the funcs on the `client.P2P` interface noting that they mutate the `rep` input variable

internal/db/errors.go
```diff
@@ -127,7 +132,12 @@ var (
 	ErrSecondaryFieldOnSchema                   = errors.New(errSecondaryFieldOnSchema)
 	ErrRelationMissingField                     = errors.New(errRelationMissingField)
 	ErrMultipleRelationPrimaries                = errors.New("relation can only have a single field set as primary")
+	ErrP2PColHasPolicy                          = errors.New("p2p collection specified has a policy on it")
```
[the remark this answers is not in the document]
fredcarle: This error was just moved here. It's not really relevant to the current PR.

internal/db/p2p_replicator.go
fredcarle: The replicator event requires a map so we have to build it. Do we really need a comment for this?

AndrewSisley: What info are you looking for here Shahzad? I think a comment here would reduce the readability of the code

internal/db/p2p_replicator.go
fredcarle: is covered by integration test but now also covered by unit test

internal/db/p2p_replicator.go
fredcarle: is covered by integration test but now also covered by unit test

internal/db/p2p_replicator.go
fredcarle: now tested

internal/db/p2p_schema_root.go
```diff
@@ -205,3 +176,52 @@ func (p *Peer) GetAllP2PCollections(ctx context.Context) ([]string, error) {
...
+	// Get all DocIDs across all collections in the DB
+	cols, err := db.GetCollections(ctx, client.CollectionFetchOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, col := range cols {
+		// If we subscribed to the collection, we skip subscribing to the collection's docIDs.
+		if _, ok := colMap[col.SchemaRoot()]; ok {
+			continue
```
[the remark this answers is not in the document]
fredcarle: now tested

internal/db/p2p_schema_root.go
```diff
@@ -205,3 +176,52 @@ func (p *Peer) GetAllP2PCollections(ctx context.Context) ([]string, error) {
...
+		// TODO-ACP: Support ACP <> P2P - https://github.com/sourcenetwork/defradb/issues/2366
+		docIDChan, err := col.GetAllDocIDs(ctx)
+		if err != nil {
+			return err
+		}
+
+		for docID := range docIDChan {
+			db.events.Publish(event.NewMessage(event.P2PTopicName, event.P2PTopic{
+				ToAdd: []string{docID.ID.String()},
+			}))
```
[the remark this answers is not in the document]
fredcarle: tested

tests/integration/db.go
fredcarle: I agree. Will do.

- [ ] rename `setupDatabase` to `setupNode`

internal/db/p2p_schema_root.go
fredcarle: Change to an `atomic.Value`

internal/db/messages.go
fredcarle: changed to atomic.Value

internal/db/p2p_replicator.go
fredcarle: no longer mutating rep

node/node.go
```diff
@@ -138,27 +138,22 @@ func NewNode(ctx context.Context, opts ...Option) (*Node, error) {
 		return nil, err
 	}
 
-	var node *net.Node
+	var peer *net.Peer
 	if !options.disableP2P {
 		// setup net node
-		node, err = net.NewNode(ctx, db, netOpts...)
+		peer, err = net.NewPeer(ctx, db.Rootstore(), db.Blockstore(), db.Events(), netOpts...)
```
nasdf, line 144: thought: since we have access to both the `db` and `peer` here, would it be simpler to just set the `peerID` directory on the `db`?

Something like this:
```go
db.SetPeerID(peer.ID())
```

Another approach that takes a bit more work:
```go
mstore := datstore.NewMultistore(...)
events := event.NewBus()
peer, err := net.NewPeer(ctx, mstore.Rootstore(), mstore.Blockstore(), events)
db, err := db.NewDB(ctx, mstore, events, peer.ID())
```

nasdf: Maybe it makes more sense to do this in the `Start` method instead.

fredcarle: I had thought of the first approach and didn't like it. I do like the second approach though. I'll see what I can do.

event/event.go
```diff
@@ -110,3 +116,25 @@ type Subscription struct {
...
+// Replicator is an event that is published when a replicator is added or updated.
+type Replicator struct {
+	// The peer info for the replicator instance.
+	Info peer.AddrInfo
+	// The map of schema roots that the replicator will receive updates for.
+	Schemas map[string]struct{}
+	// Docs will receive Updates if new collections have been added to the replicator
+	// and those collections have documents to be replicated.
+	Docs <-chan Update
+}
```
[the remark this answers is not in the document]
islamaliev: this just means that we have a problem with design and code structure. Moving stuff here is just another work-around.
2 packages can not be dependent of each other, that means whoever is independent can host the event definition.
But it's your call. I believe this PR is a good step towards cleaner code structure overall and I'm fine with keep the events here as a temporary solution. Would be nice to ticket though to keep track of our technical dept.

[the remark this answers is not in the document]
fredcarle: I'm fine with having the event types in a different package than `event`. I just don't think it's a good approach to have them spread around in their "most appropriate" package for the reason I stated above and more.

`core` is internal and the `bus` is public. Maybe `client` would be better?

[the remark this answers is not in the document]
AndrewSisley: I'm happy with `client`, although a pub-internal split might occur with event types this way (I'm happy with that, but you might not be?)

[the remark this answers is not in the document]
fredcarle: There would be no split. Everything would be in `client` in this scenario. What I want to avoid is creating dependencies with packages that are one-of-possibly-many/interchangeable. The goal of this PR is to remove the dependencies. I don't want to recreate it with event messages.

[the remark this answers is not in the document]
fredcarle: Just to clarify, when I say everything I mean every event types. Not the bus implementation itself.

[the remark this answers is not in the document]
AndrewSisley: You don't picture us ever needing/wanting internal event types?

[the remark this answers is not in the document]
fredcarle: If you mean that the public ones can be in `client/event` and the internal ones in `internal/core/event` or something like that then yes I think that's totally fine and probably desired.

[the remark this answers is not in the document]
AndrewSisley: Nice, Islam are you happy with this?

And if so, are you happy with re-orging the events into `client`/`core` in another PR (if that's what Fred prefers)?  I do see moving them as out of scope here, and if I was the author would rather do it elsewhere.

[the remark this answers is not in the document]
islamaliev: yeah, I'm fine with, also with doing this in a separate PR.

internal/db/p2p_replicator.go
shahzadlone: perhaps this is more personal pref, I like comments like: "convert to map for replicator event". But no biggie

node/node.go
fredcarle: Decided to stay with the current implementation for now as it offers more safety/guarantee that the peer ID being stored on `db` is actually the right one.

internal/db/messages.go
```diff
@@ -0,0 +1,86 @@
...
+							"Failed to execute merge",
+							err,
+							corelog.Any("Event", evt))
+					}
+				}()
+			case event.PeerInfo:
+				db.peerInfo.Store(evt.Info)
+				// Load and publish P2P collections and replicators once per db instance start.
+				// Go routines are used to ensure the message handler is not blocked by these potentially
+				// long running operations.
```
AndrewSisley, line 70: todo: Sorry about the bother, part of my request for a comment here was to explain why there are two go routines here instead of just one - can you please either document that, or just spin up a single routine?

AndrewSisley (approved):
LGTM, just one small todo - thanks Fred :)

shahzadlone (approved):
Looks good! Great work

internal/db/messages.go
fredcarle: ✅

PR #2773, opened 2024-06-25
fix: Change new identity keys to hex format

### LINKED ISSUE #2770: `defradb identity new` returns keys in encoding not supported by defra calls
The private key returned from `defradb identity new` is binary encoded, whereas the defra only accepts hex-encoded identity private keys.

The encoding of the value returned from `defradb identity new` is also not documented.

shahzadlone:
## Relevant issue(s)
Resolves #2770 

## Description
- Return the priv/pub keys in hex

- WSL2

### DIFF
--- a/acp/identity/generate.go
+++ b/acp/identity/generate.go
@@ -10,17 +10,22 @@
 
 package identity
 
-import "github.com/sourcenetwork/defradb/crypto"
+import (
+	"encoding/hex"
+
+	"github.com/sourcenetwork/defradb/crypto"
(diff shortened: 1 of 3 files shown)


AndrewSisley (approved):
LGTM, thanks Shahzad (I'm assuming you update the CLI docs before merge :))

shahzadlone:
> LGTM, thanks Shahzad (I'm assuming you update the CLI docs before merge :))

Yup it's done.

PR #2777, opened 2024-06-25
fix: Remove shared mutable state between database instances

### LINKED ISSUE #2774: GQL types declared in `request/graphql/schema/types` pkg are mutatable and shared between defra instances
The GQL types declared in `request/graphql/schema/types` pkg are mutatable and shared between defra instances.

This means that when multiple defra instances are created using the embedded Go client, mutable state will be shared between the instances.

These objects are mutated by the `graphql-go` package when we call `NewSchema` on line 30 of `manager.go`.

The default types passed into the `NewSchema` call must be brand new instances, not global variables.

As well as being a production bug, this also prevents running tests in parallel within the same test package (using `t.Parallel()`).

AndrewSisley:
## Relevant issue(s)

Resolves #2774

## Description

Removes shared mutable GQL state between database instances.

These global static GQL type variables are actually mutated by the GQL library we use at different points during database init, and schema update.  This means that when multiple Defra instances are created in the same process they were sharing state.  As well as just being an inherently bad thing, this also prevented us from running tests within packages in parallel using `t.Parallel()` as the race detector would complain.

~~Our integration tests now can all run with `t.Parallel()` and the `--race` flag (tested locally).  I do not wish to make that change within this PR though for cleanliness/focus reasons.~~ The `--race` flag is still flaky in the CI, this PR improves things but does not let us Parellize the tests fully.  I believe that the shared state in https://github.com/graphql-go/graphql/blob/master/introspection.go is responsible for the remaining failures, I'd rather spend time moving away from the package entirely than to rework that, however I still think this PR represents an improvement and would like to merge.


fredcarle (approved):
LGTM

PR #2784, opened 2024-06-26
fix: Keyring output

### LINKED ISSUE #2757: Keyring `generate` command does not mention which keys will be overwritten/created by the command
Current documentation (`defradb keyring generate --help`):
```
Generate private keys.
Randomly generate and store private keys in the keyring.

WARNING: This will overwrite existing keys in the keyring.
```
It would be very useful for users to know the names of the keys that will be overwritten.

### LINKED ISSUE #2758: Keyring generate command does not output anything on command execution
There is no indication of what it did (if anything).  Executing `defradb keyring generate` prompts for password and that is it.  Entering different random password strings still does not produce any output to the terminal (password appears to be key specific, but that is not documented in user-space).

 Executing `defradb keyring generate --keyring-backend system` does not prompt for password, and does not output anything.

`defradb keyring import` also has the same problem, however given the explicitness of that command it is slight less important. 

Flagging as a bug, as I don't know what it is doing and whether it is correct or not. This issue is worsened quite a lot by https://github.com/sourcenetwork/defradb/issues/2757.

Tested on Linux Mint, GNOME Terminal.

### LINKED ISSUE #2759: Providing invalid `--keyring-backend` value not validated and defaults to `file` without notifying user
For example, executing `defradb keyring import bar efcc38b57ccb9ddba3cf85a856acaeb28bbe7360fc0fe37630b3e91ff0bca86c --keyring-backend fdggsfa` prompts for password and then completes without outputting anything (https://github.com/sourcenetwork/defradb/issues/2758).

The imported key is then accessible via `defradb keyring export bar` (using the `file` backend).

nasdf:
## Relevant issue(s)

Resolves #2759
Resolves #2758
Resolves #2757

## Description

This PR adds more logs to the keyring to improve the user experience.

## Tasks

## How has this been tested?

- unit test

- MacOS

### DIFF
--- a/cli/keyring_generate.go
+++ b/cli/keyring_generate.go
@@ -17,12 +17,14 @@ import (
 )
 
 func MakeKeyringGenerateCommand() *cobra.Command {
-	var noEncryption bool
+	var noEncryptionKey bool
+	var noPeerKey bool
 	var cmd = &cobra.Command{
 		Use:   "generate",
 		Short: "Generate private keys",
 		Long: `Generate private keys.
 Randomly generate and store private keys in the keyring.
+By default peer and encryption keys will be generated.
 
 WARNING: This will overwrite existing keys in the keyring.
 
@@ -32,6 +34,9 @@ Example:
 Example: with no encryption key
   defradb keyring generate --no-encryption-key
 
+Example: with no peer key
+  defradb keyring generate --no-peer-key
+
 Example: with system keyring
   defradb keyring generate --keyring-backend system`,
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -39,8 +44,7 @@ Example: with system keyring
 			if err != nil {
 				return err
 			}
-			if !noEncryption {
-				// generate optional encryption key
+			if !noEncryptionKey {
 				encryptionKey, err := crypto.GenerateAES256()
 				if err != nil {
 					return err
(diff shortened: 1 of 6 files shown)


cli/keyring_generate.go
```diff
@@ -17,12 +17,14 @@ import (
...
 func MakeKeyringGenerateCommand() *cobra.Command {
-	var noEncryption bool
+	var noEncryptionKey bool
+	var noPeerKey bool
 	var cmd = &cobra.Command{
 		Use:   "generate",
 		Short: "Generate private keys",
 		Long: `Generate private keys.
 Randomly generate and store private keys in the keyring.
+By default peer and encryption keys will be generated.
```
islamaliev, line 27: though: I wonder if we need to be more clear every time we mention encryption keys as there are different types of them for different context: encryption-at-rest, doc encryption, some ACP-related and so on. They grow like mushrooms these days.

islamaliev (approved):
LGTM!

cli/keyring_generate.go
AndrewSisley: I think it is well worth being very clear here - this is a destructive operation (information is overwritten), and the keys are accessible to users via the `export` command.  I have a preference for us to list each key with it's keyring name, and a short description as to what it does.

nasdf: There's a readme section dedicated to this already.

AndrewSisley: Can we link to it from the cli? It is very handy if the CLI is self-explanatory, I find it annoying when I have to search for CLI documentation instead of just using `--help`

nasdf: I've copied the key management section into the help for the base keyring command.

AndrewSisley (approved):
LGTM, thanks Keenan :)

PR #2786, opened 2024-06-27
test: Support asserting on doc index in test results

### LINKED ISSUE #2785: Support test.DocIndex usage in expected results

AndrewSisley:
## Relevant issue(s)

Resolves #2785

## Description

Support asserting on doc index in test results.

PR does not seek to propagate it's usage, only to permit it's usage.

## How has this been tested?

I also checked that it fails correctly if the results do not pair up.


fredcarle (approved):
LGTM

PR #2788, opened 2024-06-27
test: Allow assertion of AddSchema results

### LINKED ISSUE #2766: Allow integration tests to assert the return value of AddView and  AddSchema
At the moment we only check for errors on these actions.

It is not desirable to force all actions of these action types to declare the expected return values as almost all tests use `AddSchema` and including the results in the test declaration would be a waste of eye-space, but it should be possible to do so where appropriate.

AndrewSisley:
## Relevant issue(s)

Resolves #2766

## Description

Allows assertion of AddSchema results.

PR Does not attempt to spread it's usage (out of scope).

### DIFF
--- a/tests/integration/results.go
+++ b/tests/integration/results.go
@@ -220,5 +220,9 @@ func assertCollectionDescriptions(
 			// This is to say each test action from having to bother declaring an empty slice (if there are no sources)
 			require.Equal(s.t, expected.Sources, actual.Sources)
 		}
+
+		if expected.Fields != nil {
+			require.Equal(s.t, expected.Fields, actual.Fields)
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


nasdf (approved):
LGTM

tests/integration/results.go
```diff
@@ -184,3 +187,42 @@ func areResultArraysEqual[S any](expected []S, actual any) bool {
...
+		}
+		if expected.SchemaVersionID != "" {
+			require.Equal(s.t, expected.SchemaVersionID, actual.SchemaVersionID)
+		}
+
+		require.Equal(s.t, expected.Name, actual.Name)
+
+		if expected.Indexes != nil || len(actual.Indexes) != 0 {
+			// Dont bother asserting this if the expected is nil and the actual is nil/empty.
+			// This is to say each test action from having to bother declaring an empty slice (if there are no indexes)
```
fredcarle, line 214: typo: `This is to save`? Same thing bellow.

PR #2789, opened 2024-06-27
ci(i): Make change detector an independent workflow

### LINKED ISSUE #2736: split change detector action into a separate job, or a separate workflow completely
Split change detector action into a separate job, or a separate workflow completely main due to the following reasons:

- the coverage report job does not need to wait for change detector (which takes much longer, and doesn't utilize caching), currently it is blocked when change detector is running.
- extra `if` statements in the `test-and-upload-coverage.yml` workflow for a different flow (no code coverage steps).
- `test-and-upload-coverage.yml` is getting more complex and caching stuff doesn't apply to change detector either.
- optimize change detector when it is it's own job or workflow (perhaps using caching in a different way).

shahzadlone:
## Relevant issue(s)
Resolves #2736

## Description
- Move change detection to a separate workflow to not block code coverage upload to be dependent / blocked by it.
- More reasons in the issue (#2736) on why we want this.

## How has this been tested?
- Manually

- WSL2

### DIFF
--- a/.github/workflows/check-data-format-changes.yml
+++ b/.github/workflows/check-data-format-changes.yml
@@ -1,0 +1,50 @@
+# Copyright 2024 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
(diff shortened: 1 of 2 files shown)


nasdf (approved):
LGTM. This is a much better approach than including it in the test matrix.

PR #2790, opened 2024-06-28
ci(i): Remove redundant matrix rule that is repeated

shahzadlone:
## Description
- Remove redundant matrix rule that is repeated (1st commit)
- Move all non-ubuntu special rules at the bottom

## How has this been tested?
- `act` tool
- ci

- WSL2


AndrewSisley (approved):
Is weird, I wonder how that came to be :)  Thanks for cleaning up

PR #2796, opened 2024-06-28
fix: Allow primary field declarations on one-many

### LINKED ISSUE #961: @primary directive for 1-Many Fields throws odd error
Context:
```
type Todo{
    title: String!
    userName: String!
    isSeen: Boolean
    todoUser: [TodoUser]
}


type TodoUser{
    isVerified: Boolean
    name: String! 
    age: Int
    todoCount: Int
    todo: Todo @primary
}
```
Resulting error:
> duplicate field. Name: todo

AndrewSisley:
## Relevant issue(s)

Resolves #961

## Description

Allows primary field declarations on one side of one-many.

Declaring it as primary doesn't do anything, but there is no reason it shouldn't be permitted and it definitely shouldn't have errored the way it did.

### DIFF
--- a/internal/request/graphql/schema/collection.go
+++ b/internal/request/graphql/schema/collection.go
@@ -676,16 +676,19 @@ func finalizeRelations(
 			}
 
 			if !otherColFieldDescription.HasValue() || otherColFieldDescription.Value().Kind.Value().IsArray() {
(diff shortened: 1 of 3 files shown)


fredcarle (approved):
LGTM

Issue #2812, opened 2024-07-05
Change detector keeps failing due to doc encrypt test `TestDocEncryption_UponEncryptionSeveralDocs_ShouldStoreAllCommitsDeltaEncrypted`


shahzadlone:
Started noticing change detector fails after rebasing the latest commit on develop: https://github.com/sourcenetwork/defradb/commit/24fa14f895644621915e4bf65d41469c7dc52979

Failing test: `TestDocEncryption_UponEncryptionSeveralDocs_ShouldStoreAllCommitsDeltaEncrypted`

Failed runs:
- https://github.com/sourcenetwork/defradb/actions/runs/9813511994/job/27099625664?pr=2811
- https://github.com/sourcenetwork/defradb/actions/runs/9797928722/job/27055488520

islamaliev:
how ofter does it happen?

Maybe we can try to remove all `NewDocIndex` calls from the related tests and replaces them with real doc ids.

AndrewSisley:
> Maybe we can try to remove all `NewDocIndex` calls from the related tests and replaces them with real doc ids.

Doc IDs should very much be consistent across processes, if they are not something is very broken and we should fix that instead of removing `NewDocIndex` calls

AndrewSisley:
This is probably caused by the encrypted state not being deterministic: https://security.stackexchange.com/questions/42642/should-aes-always-give-the-same-output

PR #2817, opened 2024-07-08
feat: Doc field encryption

### LINKED ISSUE #2809: Doc. encryption: enable field-level encryption
At the moment doc encryption works on document level, meaning that it will encrypt all document's field with the same encryption key.
It should be possible to encrypt doc fields with dedicated encryption keys.

islamaliev:
## Relevant issue(s)

Resolves #2809 #2812

## Description

Adds field-level encryption which allows separate fields
to be encryption with a dedicated symmectic key.


cli/collection_create.go
```diff
@@ -17,25 +17,34 @@ import (
...
 
 	-e, --encrypt
 		Encrypt flag specified if the document needs to be encrypted. If set, DefraDB will generate a
 		symmetric key for encryption using AES-GCM.
+	
+	--encrypt-fields
+		Comma-separated list of fields to encrypt. If set, DefraDB will encrypt only the specified fields
+		and for every field in the list it will generate a symmetric key for encryption using AES-GCM.
+		If combined with '--encrypt' flag, all the fields in the document not listed in '--encrypt-fields' 
+		will be encrypted with the same key.
```
shahzadlone, line 47: question: Just need some clarification, when you say "will be encrypted with the same key" you mean that all the `not listed` fields will share a key that will be different from the per field key generated for the "listed fields specified by --encrypt-fields"?

cli/utils.go
```diff
@@ -163,15 +163,15 @@ func setContextIdentity(cmd *cobra.Command, privateKeyHex string) error {
...
+func setContextDocEncryption(cmd *cobra.Command, shouldEncrypt bool, encryptFields []string, txn datastore.Txn) {
+	if !shouldEncrypt && len(encryptFields) == 0 {
 		return
 	}
 	ctx := cmd.Context()
 	if txn != nil {
 		ctx = encryption.ContextWithStore(ctx, txn)
 	}
-	ctx = encryption.SetContextConfig(ctx, encryption.DocEncConfig{IsEncrypted: true})
+	ctx = encryption.SetContextConfigFromParams(ctx, shouldEncrypt, encryptFields)
```
shahzadlone: nitpick: Just reading the function signature it can seem odd why `shouldEncrypt` can be false and still have `encryptFields`, but I understand why that is the case due to the two types of options (with field and non-field enc).

shahzadlone:
Likely also want to add the `Resolve #2812` in your PR description

cli/collection_create.go
islamaliev: yes

cli/utils.go
islamaliev: I can rename it to `shouldEncryptDoc`. Will it make it better?
Otherwise I'm open to alternatives

internal/request/graphql/schema/descriptions.go
```diff
@@ -158,7 +158,15 @@ Returns the head commit for this document.
...
-symmetric key for encryption using AES-GCM.
+Encrypt flag specifies if the input document(s) needs to be encrypted. If set, 
+ DefraDB will generate a symmetric key for encryption using AES-GCM, and will 
+ use it to encrypt all fields' values.
+`
+
+	encryptFieldsArgDescription string = `
+An optional list of individual fields that should be encrypted. For every field
+ in the list DefraDB will generate a symmetric key for encryption using AES-GCM.
+ If 'encrypt' is set to true, it all fields not listed in 'encryptedFields' will
```
AndrewSisley, line 169: praise: This is a nice way of doing it, thanks Islam :)

tests/integration/utils2.go
```diff
@@ -1252,9 +1251,7 @@ func createDocViaColSave(
 
 func makeContextForDocCreate(ctx context.Context, action *CreateDoc) context.Context {
```
AndrewSisley, line 1252: todo: Sorry, this might be a problem with doc level encryption too that I missed, but by only testing via the context params it means that the GQL params are untested.  

I think it would be much better if we rely on context for the `collection-save` and `collection-named` mutation types, but the `gql` mutation type should really be using the new gql input params instead of the context (ideally it should test with various combinations, but that's probably not worth doing atm).

internal/request/graphql/schema/generate.go
```diff
@@ -1051,9 +1051,10 @@ func (g *Generator) GenerateMutationInputForGQLType(obj *gql.Object) ([]*gql.Fie
 		Description: createDocumentDescription,
 		Type:        obj,
 		Args: gql.FieldConfigArgument{
-			"input":   schemaTypes.NewArgConfig(mutationInput, "Create a "+obj.Name()+" document"),
-			"inputs":  schemaTypes.NewArgConfig(gql.NewList(mutationInput), "Create "+obj.Name()+" documents"),
-			"encrypt": schemaTypes.NewArgConfig(gql.Boolean, encryptArgDescription),
+			"input":         schemaTypes.NewArgConfig(mutationInput, "Create a "+obj.Name()+" document"),
+			"inputs":        schemaTypes.NewArgConfig(gql.NewList(mutationInput), "Create "+obj.Name()+" documents"),
+			"encrypt":       schemaTypes.NewArgConfig(gql.Boolean, encryptArgDescription),
+			"encryptFields": schemaTypes.NewArgConfig(gql.NewList(gql.String), encryptFieldsArgDescription),
```
AndrewSisley, line 1057: todo: The set of valid params is known at 'compile time' here, so instead of a `[String]` the type of the `encryptFields` input should really be `[Enum]` - this will reduce the ability for users to mess it up, and give them nicer intelisense options with GQL clients and ORMs.

tests/integration/test_case.go
```diff
@@ -233,6 +233,9 @@ type CreateDoc struct {
 	// Specifies whether the document should be encrypted.
 	IsEncrypted bool
 
+	// Individual fields of the document to encrypt.
+	EncryptedFields []string
```
AndrewSisley, line 237: todo: Please add tests where non-existing fields are added to this set.  If you are able to make the GQL input type a `[Enum]` you will unfortunately need to split the tests by mutation types as the errors should be different (unless the GQL error is suitable for re-use via the collection API).

cli/utils.go
shahzadlone: I am okay with that rename :)

AndrewSisley (changes requested):
LGTM Islam, thanks :)  Just got a couple of todo's for your RE GQL and testing.

internal/request/graphql/schema/generate.go
islamaliev: brilliant! That's a great idea. Thanks

cli/utils.go
islamaliev: done

tests/integration/utils2.go
islamaliev: done

tests/integration/test_case.go
islamaliev: done

tests/integration/schema/type_explicit_fields_test.go
```diff
@@ -0,0 +1,60 @@
...
+		Actions: []any{
+			testUtils.SchemaUpdate{
+				Schema: `
+					type User {
+						name: String
+						age:  Int
+					}
+				`,
+			},
+			testUtils.IntrospectionRequest{
```
AndrewSisley, line 33: praise: Thanks for adding an introspection test!

tests/integration/utils2.go
AndrewSisley: Quick sanity check, as I can't spot where this is done in the PR - did you push this change? How did you do this?

internal/request/graphql/schema/generate.go
```diff
@@ -1085,11 +1097,27 @@ func (g *Generator) GenerateMutationInputForGQLType(obj *gql.Object) ([]*gql.Fie
...
+	}
+
+	for f, field := range obj.Fields() {
+		enumFieldsCfg.Values[field.Name] = &gql.EnumValueConfig{Value: f}
+	}
+
+	return gql.NewEnum(enumFieldsCfg)
+}
+
+func (g *Generator) genUserExplicitTypeFieldsEnum(obj *gql.Object) *gql.Enum {
```
AndrewSisley, line 1111: praise: I like how you managed to do this, will hopefully be re-used in the future :)

AndrewSisley (approved):
LGTM - thanks Islam :)

tests/integration/utils2.go
islamaliev: it's right here https://github.com/sourcenetwork/defradb/pull/2817/commits/c38bca4cb18a19d48ef42be5ec81180ca1e8aaef#diff-13c3c742bdb7dd9a68f413104c129a63c831dab4e2b4d484be82115ab20812fcR1341

AndrewSisley: I see my mistake - I actually misread originally - you were already using the GQL params (I didn't read closely enough and thought you were constructing a context key).  Thanks for the clarification :)

nasdf (approved):
LGTM. Great work!

One thought I had after reviewing: it might be beneficial to merge all of the db context related stuff into a single package at some point in the future.

AndrewSisley:
> One thought I had after reviewing: it might be beneficial to merge all of the db context related stuff into a single package at some point in the future.

I like that thought :)

islamaliev:
> One thought I had after reviewing: it might be beneficial to merge all of the db context related stuff into a single package at some point in the future.

I'm not a big fan to be honest. I'd much prefer to split and group packages by their domain, not based on their functional types. Just because they have something to do with `context` doesn't mean they are cohesive enough to put them in one package.

AndrewSisley:
> I'd much prefer to split and group packages by their domain, not based on their functional types. Just because they have something to do with context doesn't mean they are cohesive enough to put them in one package.

I understand what you mean, but for me (thinking as a user) I see them all in the 'Defra' domain, just like we lump config, and  the `client` package into singular clumps.

PR #2822, opened 2024-07-09
test: Remove hardcoded test identities

### LINKED ISSUE #2821: Remove hardcoded test identities
Broken out of #2366

AndrewSisley:
## Relevant issue(s)

Resolves #2821

## Description

Removes the hardcoded test identities from our integration tests.

The current (develop) setup becomes impossible with source-hub ACP as the identity needs to include a signed bearer token containing the sourcehub address (not accessible during package init).

I also think this is a bit more maintainable.

This commit has been broken out of https://github.com/sourcenetwork/defradb/pull/2657 to minimize the risk of conflicts and to keep the PRs focused. It might be useful for reviewers to have a quick look at the full (WIP) `getIdentity` func in the original PR.


tests/integration/acp/add_policy/basic_test.go
```diff
@@ -23,7 +25,7 @@ func TestACP_AddPolicy_BasicYAML_ValidPolicyID(t *testing.T) {
 
 		Actions: []any{
 			testUtils.AddPolicy{
-				Identity: actor1Identity,
+				Identity: immutable.Some(1),
```
nasdf, line 28: question: why use index 1 if we never use index 0?

nasdf: Okay that makes sense. I thought there might be something else I'm missing.

nasdf (approved):
LGTM.

PR #2828, opened 2024-07-10
fix(i): Make badger dependency optional

### LINKED ISSUE #2827: Make badger dependency optional
Move badger dependencies to files with build flags to allow JS build

nasdf:
## Relevant issue(s)

Resolves #2827

## Description

This PR moves some things around to make the badger dependency optional to allow for JS builds.

## Tasks

## How has this been tested?

`make test`

- MacOS


datastore/errors/errors_badger.go
```diff
@@ -0,0 +1,21 @@
...
+// licenses/APL.txt.
+
+//go:build !js
+
+package errors
+
+import "github.com/sourcenetwork/badger/v4"
+
+func init() {
+	ErrTxnConflict = badger.ErrConflict
```
AndrewSisley, line 18: suggestion: I can picture this causing problems - not building for JS does not mean that the errors should be badger ones!  Other stores may be used (also or as well).  It would be far better IMO to make sure that we return consistent errors instead of doing this.

In my opinion errors are part of the interface (contract), not the implementation - yes, a bunch of annoying libs break this, and occasionally it is necessary if the interface does not declare enough useful errors (not a problem here, as we have control - even if it means a thin wrapper), but this is the point of view I have when making the above suggestion.

node/store.go
```diff
@@ -14,65 +14,75 @@ import (
...
 	}
 }
 
-// WithDefraStore sets the defra store flag.
-//
-// Setting this to true will result in the defra node being created with
-// the a custom defra implementation of the rootstore instead of badger.
-func WithDefraStore(defraStore bool) StoreOpt {
+// WithBadgerInMemory sets the badger in memory option.
+func WithBadgerInMemory(enable bool) StoreOpt {
```
AndrewSisley, line 62: todo: I think if the badger store is protected by build flags, then all of the production badger stuff should also be, including these functions.  Please move them into a build-tag specific file, or explain why that is not practical :)

node/store.go
```diff
@@ -14,65 +14,75 @@ import (
...
 	}
 }
 
-// WithPath sets the datastore path.
-func WithPath(path string) StoreOpt {
+// WithBadgerPath sets the badger datastore path.
+func WithBadgerPath(path string) StoreOpt {
 	return func(o *StoreOptions) {
-		o.path = path
+		o.badgerPath = path
```
AndrewSisley, line 71: question: Why have you made these (props and public funcs) store specific?

datastore/errors/errors.go
```diff
@@ -0,0 +1,39 @@
...
+
+var (
+	ErrClosed = errors.New("datastore closed")
+	// ErrConflict is returned when a transaction conflicts with another transaction. This can
+	// happen if the read rows had been updated concurrently by another transaction.
+	//
+	// This error is copied from badger because it is used in a few places outside of the datastore
+	// package. This copy ensures that we can build in environments that do not support badger.
+	//
+	//nolint:revive
```
AndrewSisley, line 23: question: What is the `nolint` for?  What error do you get without it?

AndrewSisley (changes requested):
LGTM, looking forward to getting the full JS build :)  Got a handful of small stuff for you first :)

datastore/errors/errors.go
nasdf: Capital letter at the beginning of the message.

AndrewSisley: This reason nudges me even more towards changing the errors (either in the badger fork, or with a wrapper)

datastore/errors/errors_badger.go
nasdf: I've added a function to convert the errors within the badger datastore. It was much simpler than I first thought.

node/store.go
nasdf: The trade-off to make here is: if we move the options to the build specific file, then we will need to add platform specific files to the test harness as well. I was hoping the node package could serve as the entry point for all platforms so that we can avoid adding platform specific files elsewhere. I'm fine with either option.

nasdf: I moved the options to the badger specific file like we discussed.

node/store.go
nasdf: I renamed `badgerPath` to `path`.

datastore/errors/errors.go
AndrewSisley: Are these code-comments now out of date? These can now be our own errors due to the wrapping?  Can we remove the lint and start them with lower case?

nasdf: Done

shahzadlone (approved):
LGTM

PR #2841, opened 2024-07-16
fix: Enable filtering doc by fields of JSON and Blob types

### LINKED ISSUE #2840: Filtering by JSON field returns "Unknown field" error
Schema looks like this
```graphql
type PlayerQuizzes {
    playerId: String
    data: JSON
    createdAt: DateTime 
    updatedAt: DateTime
}
```

Query look like this
```graphql
{
  PlayerQuizzes(filter: {
    data: { _like: "%tweetId%" }  
  }) {
    id: _docID
  }
}
```

And a result of query
```json
{
  "data": null,
  "errors": [
    "Argument \"filter\" has invalid value {data: {_like: \"tweetId\"}}.\nIn field \"data\": Unknown field."
  ]
}
```

islamaliev:
## Relevant issue(s)

Resolves #2840

## Description

Fixes the issue not being able to filter by JSON and Blob fields


internal/request/graphql/schema/descriptions.go
```diff
@@ -33,8 +33,8 @@ var (
 		client.FieldKind_NILLABLE_STRING:       gql.String,
 		client.FieldKind_STRING_ARRAY:          gql.NewList(gql.NewNonNull(gql.String)),
 		client.FieldKind_NILLABLE_STRING_ARRAY: gql.NewList(gql.String),
-		client.FieldKind_NILLABLE_BLOB:         schemaTypes.BlobScalarType(),
```
AndrewSisley, line 36: question: Why have you made this change?  This was done quite recently as these are mutated by the gql library and making them singletons results in shared mutated state between database instances.

AndrewSisley (changes requested):
Requesting changes, as I think you have introduced a regression.

internal/request/graphql/schema/descriptions.go
islamaliev: this is done to follow the same style that is used by `github.com/sourcenetwork/graphql-go`: to have 1 scalar definition instance instead of creating always a new.

islamaliev:
> Requesting changes, as I think you have introduced a regression.

where? how?

internal/request/graphql/schema/descriptions.go
AndrewSisley: This is wrong (and `graphql-go`), and we should not do it.  As well as being a production bug, it prevents us from using `t.Parallel()` in our tests.

tests/integration/query/simple/with_filter/with_in_json_test.go
```diff
@@ -0,0 +1,55 @@
...
+				},
+			},
+			testUtils.CreateDoc{
+				DocMap: map[string]any{
+					"name":   "Andy",
+					"custom": "{\"tree\": \"oak\", \"age\": 450}",
+				},
+			},
+			testUtils.Request{
+				// the filtered-by JSON has no spaces, because this is now it's stored.
```
nasdf, line 43: thought: if we normalize JSON when parsing we could avoid the space issue and also match out of order properties.

nasdf (approved):
LGTM. Nice work!

internal/request/graphql/schema/descriptions.go
islamaliev: If we have issues with `t.Paralel()` we should change how we use `gql.String` and others. I just made it consistent.

I tried to roll this part back and see if it works. It doesn't because there is a pointer comparison https://github.com/sourcenetwork/graphql-go/blob/master/schema.go#L295

Seems like the change is required for the fix.

AndrewSisley: It is not just about `t.Parallel()`.   `graphql-go` is very singleton heavy, and it mutates the singleton types we provide it.  This means multiple embedded Defra instances leak GQL types between them, instance1 will have bits of instance2's gql types in its GQL schema.

> It doesn't because there is a pointer comparison 

You need to only call `schemaTypes.BlobScalarType()` once per database instance, and fetch it from the type map where-ever else you require it.  All our other types do this.

AndrewSisley (approved):
LGTM, thanks for fixing this Islam :)

PR #2848, opened 2024-07-18
feat: Handle P2P with SourceHub ACP

### LINKED ISSUE #2366: Add support for P2P and ACP

AndrewSisley:
## Relevant issue(s)

Resolves #2366

## Description

Handles P2P with SourceHub ACP.

Local ACP remains blocked off, as documents synced would essentially become public unless encrypted.

The testing assumes that a distributed SourceHub chain would be tested in the SourceHub repo, and so it doesn't bother spinning up a chain per node and dealing with SourceHub sync stuff.  Long term we'll want some Defra-based testing around this, but I think we can get away without it.  Feel very free to argue against this :)

### DIFF
--- a/acp/README.md
+++ b/acp/README.md
@@ -458,9 +458,10 @@ If authentication fails for any reason a `403` forbidden response will be return
 ## _FAC Usage: (coming soon)_
 
 ## Warning / Caveats
+- If using Local ACP, P2P will only work with collections that do not have a policy assigned.  If you wish to use ACP
+on collections connected to a multi-node network, please use SourceHub ACP.
+
 The following features currently don't work with ACP, they are being actively worked on.
-- [P2P: Adding a replicator with permissioned collection](https://github.com/sourcenetwork/defradb/issues/2366)
-- [P2P: Subscription to a permissioned collection](https://github.com/sourcenetwork/defradb/issues/2366)
 - [Adding Secondary Indexes](https://github.com/sourcenetwork/defradb/issues/2365)
 - [Backing/Restoring Private Documents](https://github.com/sourcenetwork/defradb/issues/2430)
 
--- a/internal/db/p2p_replicator.go
+++ b/internal/db/p2p_replicator.go
@@ -45,8 +45,6 @@ func (db *db) SetReplicator(ctx context.Context, rep client.Replicator) error {
 		return ErrSelfTargetForReplicator
 	}
 
-	// TODO-ACP: Support ACP <> P2P - https://github.com/sourcenetwork/defradb/issues/2366
-	// ctx = db.SetContextIdentity(ctx, identity)
 	ctx = SetContextTxn(ctx, txn)
 
 	storedRep := client.Replicator{}
@@ -90,20 +88,19 @@ func (db *db) SetReplicator(ctx context.Context, rep client.Replicator) error {
 		}
 
 	default:
-		// default to all collections (unless a collection contains a policy).
-		// TODO-ACP: default to all collections after resolving https://github.com/sourcenetwork/defradb/issues/2366
 		allCollections, err := db.GetCollections(ctx, client.CollectionFetchOptions{})
 		if err != nil {
 			return NewErrReplicatorCollections(err)
 		}
 
-		for _, col := range allCollections {
-			// Can not default to all collections if any collection has a policy.
-			// TODO-ACP: remove this check/loop after https://github.com/sourcenetwork/defradb/issues/2366
-			if col.Description().Policy.HasValue() {
-				return ErrReplicatorSomeColsHavePolicy
+		if db.acp.HasValue() && !db.acp.Value().SupportsP2P() {
+			for _, col := range allCollections {
+				if col.Description().Policy.HasValue() {
+					return ErrReplicatorSomeColsHavePolicy
+				}
 			}
 		}
+
 		collections = allCollections
 	}
 
--- a/internal/db/p2p_replicator.go
+++ b/internal/db/p2p_replicator.go
@@ -80,28 +80,22 @@ func (db *db) SetReplicator(ctx context.Context, rep client.Replicator) error {
 				return NewErrReplicatorCollections(err)
 			}
 
-			if col.Description().Policy.HasValue() {
-				return ErrReplicatorColHasPolicy
-			}
-
 			collections = append(collections, col)
 		}
 
 	default:
-		allCollections, err := db.GetCollections(ctx, client.CollectionFetchOptions{})
+		collections, err = db.GetCollections(ctx, client.CollectionFetchOptions{})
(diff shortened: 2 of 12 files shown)


tests/integration/acp/p2p/replicator_test.go
```diff
@@ -89,3 +81,113 @@ func TestACP_P2POneToOneReplicatorWithPermissionedCollection_Error(t *testing.T)
...
+				testUtils.SourceHubACPType,
+			},
+		),
+		Actions: []any{
+			testUtils.RandomNetworkingConfig(),
+			testUtils.RandomNetworkingConfig(),
+			testUtils.AddPolicy{
+				Identity: immutable.Some(1),
+				Policy: `
+                    name: test
```
islamaliev, line 98: suggestion: this seems like a regular policy and doesn't really improve testing readability. Would be nice to extract it to a const.

islamaliev (approved):
LGTM! Left a few comments

tests/integration/acp.go
```diff
@@ -98,7 +99,17 @@ func addPolicyACP(
...
+	addedToSourceHub := false
+
 	for i, node := range getNodes(action.NodeID, s.nodes) {
+		// The policy should only be added to the SourceHub chain once.
+		if addedToSourceHub {
+			break
+		}
+		if acpType == SourceHubACPType {
+			addedToSourceHub = true
+		}
```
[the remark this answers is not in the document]
AndrewSisley: >  why not to put in the end of the loop body this

That would be much nicer lol, thanks :grin:

- [ ] Simplify addedToSourceHub stuff

tests/integration/acp/p2p/replicator_test.go
AndrewSisley: Going to shrug this one off as personal preference, as I prefer it inline - partly so that I can see that it is just a regular policy.  I could store it in a var called `totallyNormalAndUnimportantUserPolicy`, but that would make me even more suspicious and would cause me to read it even more carefully than if it was inline (and I'd have to bounce around out of the test to do so) :)

internal/db/p2p_replicator.go
```diff
@@ -90,20 +88,19 @@ func (db *db) SetReplicator(ctx context.Context, rep client.Replicator) error {
...
 		if err != nil {
 			return NewErrReplicatorCollections(err)
 		}
 
-		for _, col := range allCollections {
-			// Can not default to all collections if any collection has a policy.
-			// TODO-ACP: remove this check/loop after https://github.com/sourcenetwork/defradb/issues/2366
-			if col.Description().Policy.HasValue() {
-				return ErrReplicatorSomeColsHavePolicy
+		if db.acp.HasValue() && !db.acp.Value().SupportsP2P() {
```
nasdf, line 96: todo: the switch case above this also checks for a policy (unsure why there is not a todo there). It might be cleaner to combine both checks outside of the switch case

nasdf (changes requested):
Changes look good! One potential issue to resolve.

internal/db/p2p_schema_root_test.go
```diff
@@ -249,9 +249,6 @@ func TestGetAllP2PCollections_WithMultipleValidCollections_ShouldSucceed(t *test
 	require.Equal(t, []string{schema2.Root, schema1.Root}, cols)
 }
 
-// This test documents that we don't allow adding p2p collections that have a policy
-// until the following is implemented:
```
shahzadlone, line 253: todo: Still makes sense to have the comment that this test documents that we don't allow adding p2p collections that have a policy for local acp.

tests/integration/acp.go
```diff
@@ -110,6 +111,12 @@ func addPolicyACP(
 
 		expectedErrorRaised := AssertError(s.t, s.testCase.Description, err, action.ExpectedError)
 		assertExpectedErrorRaised(s.t, s.testCase.Description, action.ExpectedError, expectedErrorRaised)
+
+		// The policy should only be added to a SourceHub chain once - there is no need to loop through
```
shahzadlone, line 115: question: What happens if a node tries to re-add the same exact policy ? Guessing it will output a new policyID and still add the policy?

tests/integration/acp/p2p/replicator_test.go
```diff
@@ -18,23 +18,18 @@ import (
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
 )
 
-// This test documents that we don't allow setting replicator with a collections that has a policy
-// until the following is implemented:
-// TODO-ACP: ACP <> P2P https://github.com/sourcenetwork/defradb/issues/2366
-func TestACP_P2POneToOneReplicatorWithPermissionedCollection_Error(t *testing.T) {
+func TestACP_P2POneToOneReplicatorWithPermissionedCollection_LocalACP(t *testing.T) {
```
shahzadlone, line 21: todo: I like the documentation on what the test does (can just add that its for local acp case)

tests/integration/acp/p2p/replicator_test.go
shahzadlone: Happy as is

shahzadlone (approved):
Lovely work, finally ties a lot of pieces together.

Just left some todos and questions. Please resolve before merge.

internal/db/p2p_replicator.go
AndrewSisley: Nice spot!  Yeah I'll pull them both out of the switch and remove one of the error types.  Thanks Keenan :)

- [ ] replicator error stuff

internal/db/p2p_schema_root_test.go
AndrewSisley: I will delete the test

- [ ] Delete unit test

tests/integration/acp.go
AndrewSisley: That has not changed, the test framework will still attempt to re-add the policy.

tests/integration/acp/p2p/replicator_test.go
AndrewSisley: I dont think it is worth the eye/brain-space here. This one-liner should have been in the description, and I have a large dislike for the description prop as they are very often misleading.  

It will save us time in the long run to let the test definition define itself.

acp/README.md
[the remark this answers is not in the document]
AndrewSisley: I think that is too lengthy to include in this document and it will help drown out the information currently contained within it.

nasdf (approved):

PR #2850, opened 2024-07-19
fix(i): Tests with primary doc patch update events

### LINKED ISSUE #2849: Integration tests don't account for primary document patches
The integration test event logic does not account for updates from primary document patching.

nasdf:
## Relevant issue(s)

Resolves #2849

## Description

This PR fixes an issue with primary document patch updates. We also now assert that the document updates match the expected updates. We can most likely remove the `events` test package now that every test asserts the expected events.

## Tasks

## How has this been tested?

`make test`

- MacOS


tests/integration/events.go
```diff
@@ -148,52 +150,76 @@ func waitForUnsubscribeToCollectionEvent(s *state, action UnsubscribeToCollectio
...
+			require.True(s.t, ok, "unexpected document update")
+			delete(expect, evt.DocID)
 
-		// update the expected document heads of replicator targets
-		for id := range s.nodeP2P[i].replicators {
-			// replicator target nodes push updates to source nodes
-			s.nodeP2P[id].expectedDocHeads[evt.DocID] = evt.Cid
-		}
+			if i >= len(s.nodeConfigs) {
+				continue // not testing network state
```
AndrewSisley, line 187: todo: This comment caused me a fair bit of confusion, and I initially thought it was skipping P2P tests (via some strange mechanic that I didn't yet understand).  Can you please rework the comment a bit?

tests/integration/events.go
```diff
@@ -148,52 +150,76 @@ func waitForUnsubscribeToCollectionEvent(s *state, action UnsubscribeToCollectio
...
-		select {
-		case msg, ok := <-s.nodeEvents[i].update.Message():
-			if !ok {
-				require.Fail(s.t, "subscription closed waiting for update event")
-			}
-			evt = msg.Data.(event.Update)
-
-		case <-time.After(eventTimeout):
-			require.Fail(s.t, "timeout waiting for update event")
+		expect := make(map[string]struct{})
```
AndrewSisley, line 163: todo: Please set the length in the `make` call instead of resizing the map for each item.

tests/integration/events.go
```diff
@@ -240,3 +266,84 @@ func waitForMergeEvents(s *state) {
...
+		}
+	}
+
+	return expect
+}
+
+// getEventsForDelete returns a map of expected update events for a document delete action.
+func getEventsForDelete(docID client.DocID) map[string]struct{} {
+	expect := make(map[string]struct{})
+	expect[docID.String()] = struct{}{}
```
AndrewSisley, line 347: todo: Please dont grow the map like this, please either set the length on `make`, or (IMO) better replace this with:
```
return map[string]struct{}{
   docID.String(): {},
}
```
I would probably delete the `getEventsForDelete` entirely and just do this inline where `getEventsForDelete`, it saves readers from having to bounce around the file when reading (I won't push for that as a todo though :)).

AndrewSisley (approved):
Thank you very much for removing that test package Keenan, just a couple of small todo's for you before merge :)

tests/integration/events.go
```diff
@@ -148,52 +150,76 @@ func waitForUnsubscribeToCollectionEvent(s *state, action UnsubscribeToCollectio
 // update event to the local event bus.
 //
 // Expected document heads will be updated for any connected nodes.
-func waitForUpdateEvents(s *state, nodeID immutable.Option[int], collectionID int) {
+func waitForUpdateEvents(
+	s *state,
+	nodeID immutable.Option[int],
+	updates map[string]struct{},
```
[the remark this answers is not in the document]
nasdf: renamed to `docIDs`

tests/integration/events.go
nasdf: I updated the docs and broke out the rest into a more dedicated function. Hopefully its more clear now.

tests/integration/events.go
nasdf: I dropped the function and created the map like you suggested.

Issue #2619, opened 2024-05-14
Add `Self` field kind


AndrewSisley:
Required for (and broken out of) https://github.com/sourcenetwork/defradb/issues/2493 in order to prevent the schema-DAG from becoming circular.

AndrewSisley:
I have started this work on branch `2619-self-field-kind`, switching to https://github.com/sourcenetwork/defradb/issues/2625 as it will make doing this ticket easier.

Issue #2493, opened 2024-04-04
FieldKind points to schema and collection by name but they are not always the same


AndrewSisley:
And collection name is mutable, and thus a poor way of linking them.

We likely have a few references in the codebase that still assume they are the same and use the wrong means of getting a definition (relations).

Can only be done after https://github.com/sourcenetwork/defradb/issues/2451 as schemaField.Kind must point to the schema, and collectionField.Kind must point to the collection - at the moment they are the same property and thus cannot be detangled.

I suggest that collectionField.Kind references the linked collection by RootID, not by name, so that name mutations do not require a horrible cross-object update.

AndrewSisley:
Note, I have a (early) WIP branch of name 2493-field-kind-schema-vs-col, breaking https://github.com/sourcenetwork/defradb/issues/2619 out of this task and working on that first.

PR #2859, opened 2024-07-23
feat: Remove IsObjectArray

### LINKED ISSUE #2356: IsObject shouldn't be true for an array of objects
The `IsObject` method on `FieldKind` returns true for an array of objects. This should be fixed.

AndrewSisley:
## Relevant issue(s)

Resolves #2356

## Description

Removes IsObjectArray.

As well as causing confused code, and an overly-large interface, where their are multiple overlapping ways to check the same thing, this func also dirties the solution for #2493 and #2619.

PR is marked as a feature as it is a public change, could be called 'fix', I think it is a bit vague here give me a shout if you have a strong preference.

### DIFF
--- a/internal/request/graphql/schema/generate.go
+++ b/internal/request/graphql/schema/generate.go
@@ -471,18 +471,15 @@ func (g *Generator) buildTypes(
 				}
 
 				var ttype gql.Type
-				if field.Kind.IsObject() && !field.Kind.IsArray() {
+				if field.Kind.IsObject() {
 					var ok bool
 					ttype, ok = g.manager.schema.TypeMap()[field.Kind.Underlying()]
 					if !ok {
 						return nil, NewErrTypeNotFound(field.Kind.Underlying())
 					}
-				} else if field.Kind.IsObjectArray() {
-					t, ok := g.manager.schema.TypeMap()[field.Kind.Underlying()]
-					if !ok {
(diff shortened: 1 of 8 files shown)


internal/request/graphql/schema/generate.go
```diff
@@ -471,18 +471,20 @@ func (g *Generator) buildTypes(
...
+						if !ok {
+							return nil, NewErrTypeNotFound(field.Kind.Underlying())
+						}
+						ttype = gql.NewList(t)
+					} else {
+						var ok bool
+						ttype, ok = g.manager.schema.TypeMap()[field.Kind.Underlying()]
+						if !ok {
+							return nil, NewErrTypeNotFound(field.Kind.Underlying())
+						}
```
nasdf, line 486: nitpick: minor simplification

```go
if field.Kind.IsObject() {
	var ok bool
	ttype, ok = g.manager.schema.TypeMap()[field.Kind.Underlying()]
	if !ok {
		return nil, NewErrTypeNotFound(field.Kind.Underlying())
	}
	if field.Kind.IsArray() {
		ttype = gql.NewList(t)
	}
}
```

nasdf (approved):
Looks good! Code is much easier to read now.

internal/request/graphql/schema/generate.go
AndrewSisley: Good spot, thanks :)

- [ ] generate.go simplification

Issue #1395, opened 2023-04-23
Only first GraphQL request in a HTTP request is applied


orpheuslummis:
(not sure this title is clear, but anyway)

I suggest that either it could be giving an error or be documented behavior instead of silently only applying the first query.

e.g. with the HTTP Request sending two GraphQL requests,

```
query A {
  User {
    _key
  }
}

mutation B {
    create_User(data: "{\"name\": \"Two\", \"age\": 7}") {
      _key
    }
}
```

only `A`, the first one, is applied.

jsimnz:
This is actually basically defined as per the spec: https://spec.graphql.org/June2018/#sec-Executing-Requests

GQL will only execute a single operation. If multiple are given, the need to be named, and you need to specify the name of the operation to run.

Theres an open issue to change this: https://github.com/graphql/graphql-spec/issues/375

orpheuslummis:
Perhaps a future issue could be made for this, or this issue could be renamed and implemented as "Support GraphQL batching" once a GraphQL spec has it.

PR #2872, opened 2024-07-26
refactor: GQL responses

### LINKED ISSUE #2869: GQL response types are incorrect
Query and mutation responses should be objects of their respective types.

See https://spec.graphql.org/October2021/#sel-GAPHPDCB3CB4Ih0a

nasdf:
## Relevant issue(s)

Resolves #2869

## Description

This PR makes our GQL response types spec compliant. It also enables running multiple selections within `queries` and `mutations`.

- [x] Update readme examples

## Tasks

## How has this been tested?

`make test`

- MacOS


internal/planner/mapper/mapper.go
```diff
@@ -41,7 +41,75 @@ const (
...
+			renderKey := core.RenderKey{Index: i}
+			if t.Alias.HasValue() {
+				renderKey.Key = t.Alias.Value()
+			} else {
+				renderKey.Key = t.Name
+			}
+			operation.Mutations = append(operation.Mutations, m)
+			operation.DocumentMapping.Add(i, m.Name)
+			operation.DocumentMapping.SetChildAt(i, m.DocumentMapping)
+			operation.DocumentMapping.RenderKeys = append(operation.DocumentMapping.RenderKeys, renderKey)
```
islamaliev, line 105: nitpick: the bodies of all 3 cases are very similar. Is there a way to extract common functionality?

internal/planner/operation.go
```diff
@@ -0,0 +1,178 @@
...
+	return p.children
+}
+
+func (n *operationNode) Explain(explainType request.ExplainType) (map[string]any, error) {
+	switch explainType {
+	case request.SimpleExplain:
+		return map[string]any{}, nil
+
+	case request.ExecuteExplain:
+		return map[string]any{}, nil
```
islamaliev, line 86: question: is it supposed to be empty?

Did you add explain tests that assert this?

internal/planner/operation.go
```diff
@@ -0,0 +1,178 @@
...
+				if !hasChild {
+					break
+				}
+				docs = append(docs, child.Value())
+			}
+			n.currentValue.Fields[i] = docs
+		}
+	}
+
+	n.isDone = true
```
islamaliev, line 127: question: do you really need this flag? It's being set only in `Init()`. `Next()` is not supposed to be called anyway before `Init()` and it's not multi-threaded.

internal/planner/planner.go
```diff
@@ -627,6 +567,26 @@ func (p *Planner) RunSubscriptionRequest(
 // resources of the returned plan.
 //
 // @TODO {defradb/issues/368}: Test this exported function.
-func (p *Planner) MakePlan(request *request.Request) (planNode, error) {
-	return p.makePlan(request)
+func (p *Planner) MakePlan(req *request.Request) (planNode, error) {
+	var operation *request.OperationDefinition
+	if len(req.Mutations) > 0 {
+		operation = req.Mutations[0] // @todo: handle multiple mutation operation statements
```
islamaliev, line 573: todo: is this todo going to be resolved within this PR?

Otherwise please create a ticket and link it here.

tests/integration/acp/index/query_with_relation_test.go
```diff
@@ -111,12 +111,14 @@ func TestACPWithIndex_UponQueryingPrivateOneToManyRelatedDocWithoutIdentity_Shou
...
 						}
 					}`,
-				Results: []map[string]any{
-					{
-						"name": "John Grisham",
-						"published": []map[string]any{
-							{
-								"name": "Painted House",
+				Results: map[string]any{
+					"Author": []map[string]any{
```
islamaliev, line 115: praise: you did it all by hand? Wow. That must have been tedious.

tests/integration/utils2.go
```diff
@@ -1906,6 +1944,22 @@ func assertRequestResults(
 	return false
 }
 
+func convertToArrayOfMaps(t testing.TB, value any) []map[string]any {
```
islamaliev, line 1947: question: these (and others) are changes from my PR that was merged. How did it end up here?

islamaliev (commented):
Briefly looked over. Looks good so far.

tests/integration/utils2.go
nasdf: I think I copied these on accident when I was mid-refactor and didn't realize my local branch was behind. When I merge the develop branch these should appear as unchanged.

tests/integration/acp/index/query_with_relation_test.go
nasdf: I'm not gonna lie, It was a lot of work, but it had to be done.

internal/planner/planner.go
nasdf: I copied these over from another function. I'll double check if there is an existing issue or create one if needed.

nasdf: Found it: https://github.com/sourcenetwork/defradb/issues/1395

internal/planner/operation.go
nasdf: I'm not super familiar with the planner package. Is this supposed to return the operations that this node executes? It does appear in the existing explain tests correctly.

internal/planner/mapper/mapper.go
nasdf: I moved some of this to a function and reduced most of the duplicate lines

internal/planner/operation.go
nasdf: I tried removing it, but the return value was always incorrect. I believe we need it so that the operation is able to run at least once.

internal/planner/operation.go
jsimnz: I believe you can omit the `Explain` method entirely, since its an optional interface. I don't know if there is any context that this planNode can add to the existing plan, so simply not including it in the explain system is an option.

tests/integration/explain.go
```diff
@@ -57,6 +57,7 @@ var (
 		"valuesNode":    {},
 		"viewNode":      {},
 		"lensNode":      {},
+		"operationNode": {},
```
shahzadlone, line 60: praise: Thanks for not forgetting this :P (easy to miss if the `operationNode` was not explainable node).

tests/integration/explain/debug/fixture.go
```diff
@@ -14,9 +14,13 @@ type dataMap = map[string]any
 
 var basicPattern = dataMap{
 	"explain": dataMap{
-		"selectTopNode": dataMap{
-			"selectNode": dataMap{
-				"scanNode": dataMap{},
+		"operationNode": []dataMap{
```
shahzadlone, line 17: question: Just to confirm even the bare bones / basic most query will use an `operationNode` from now on?

internal/planner/operation.go
```diff
@@ -0,0 +1,185 @@
...
+import (
+	"github.com/sourcenetwork/defradb/client/request"
+	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/planner/mapper"
+)
+
+const operationNodeKind string = "operationNode"
+
+// operationNode is the top level node for operations with
+// one or more child selections, such as queries or mutations.
```
shahzadlone, line 22: nitpick: I understand what you mean by `operation` but wonder if there is a better name for this node (I don't have any good ideas atm), the reason I say this is because every node is essentially an "instruction/operation step" so it seems a bit vague.

internal/planner/operation.go
shahzadlone: As @jsimnz pointed out, making a node explainable is optional. Moreover, `MultiNode` types are not / don't need to be explainable but we use them to wrap the children under them so the explain graph builder I believe handles that case. But when/if you make `operationNode` unexplainable don't forget to leave the added line in the `tests/integration/explain.go` file because the debug explain graph should still need it to assert it properly in the testing framework.

An example of this is `parallelNode` has no `Explain` method, yet still shows in the explain graph to wrap the children under it.

tests/integration/explain/debug/fixture.go
nasdf: Yes that's correct

internal/planner/operation.go
nasdf: I agree the name could be improved. It should be fairly easy to rename if we find a better name.

internal/planner/operation.go
AndrewSisley: nitpick/todo: The planNode interface and it's consuption dictates that `Init` can be called multiple times, and that it's state should be reset by the first call (before doing it's thing again).  I can't think why this node would have an instance Init'ed multiple times, but it is kind of incorrect atm, and just last week we had a production issue caused by this.

nasdf: Should be fixed now. I'm assuming you were referencing the `currentValue` not being reset.

internal/db/request.go
```diff
@@ -53,10 +53,10 @@ func (db *db) execRequest(ctx context.Context, request string) *client.RequestRe
 	results, err := planner.RunRequest(ctx, parsedRequest)
 	if err != nil {
 		res.GQL.Errors = []error{err}
-		return res
 	}
-
-	res.GQL.Data = results
+	if len(results) > 0 {
+		res.GQL.Data = results[0]
```
AndrewSisley, line 58: question/suggestion: Why is results still an array?  Can we return just a `map[string]any` now instead?

Same question/suggestion for `RunSelection`.

internal/planner/mapper/mapper.go
```diff
@@ -41,7 +41,51 @@ const (
...
+	ctx context.Context,
+	store client.Store,
+	operationRequest *request.OperationDefinition,
+) (*Operation, error) {
+	operation := &Operation{
+		DocumentMapping: core.NewDocumentMapping(),
+	}
+
+	for i, s := range operationRequest.Selections {
+		switch t := s.(type) {
```
AndrewSisley, line 58: suggestion: The default case is not handled, it might be better to return an error instead of quietly ignoring it - I believe it could only happen if we mess up in the future, but it could save some head-scratching even if it means a line or two of dead code.

internal/planner/mapper/operation.go
```diff
@@ -0,0 +1,49 @@
...
+	CommitSelects []*CommitSelect
+}
+
+// addSelection adds a new selection to the operation's document mapping.
+// The request.Field is used as the key for the core.RenderKey and the Select
+// document mapping is added as a child field.
+func (o *Operation) addSelection(i int, f request.Field, s Select) {
+	renderKey := core.RenderKey{Index: s.Index}
+	if f.Alias.HasValue() {
+		renderKey.Key = f.Alias.Value()
```
AndrewSisley, line 42: todo: Please add an operation alias test

internal/planner/mapper/operation.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// CommitSelects is the list of commit selections in the operation.
+	CommitSelects []*CommitSelect
+}
+
+// addSelection adds a new selection to the operation's document mapping.
+// The request.Field is used as the key for the core.RenderKey and the Select
+// document mapping is added as a child field.
+func (o *Operation) addSelection(i int, f request.Field, s Select) {
+	renderKey := core.RenderKey{Index: s.Index}
```
AndrewSisley, line 40: todo: There is a function `getRenderKey` that handles this, I would prefer that we don't duplicate that logic as it should remain consistent across all places.

internal/planner/operation.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	n.isDone = true
+	return true, nil
+}
+
+// Operation creates a new operationNode using the given Selects.
+func (p *Planner) Operation(operation *mapper.Operation) (*operationNode, error) {
+	var children []planNode
+	var childIndexes []int
```
AndrewSisley, line 124: suggestion: I think a `map[int]planNode` would be safer and a little easier to read.  Including on the `operationNode` type.

internal/planner/operation.go
```diff
@@ -0,0 +1,174 @@
...
+		child, err := p.CommitSelect(s)
+		if err != nil {
+			return nil, err
+		}
+		children = append(children, child)
+		childIndexes = append(childIndexes, s.Index)
+	}
+
+	if len(children) == 0 {
+		return nil, ErrOperationDefinitionMissingSelection
```
AndrewSisley, line 166: question: Is this spec-compliant? A selection set must always be provided?

internal/planner/operations.go
```diff
@@ -34,9 +34,11 @@ var (
 	_ planNode = (*valuesNode)(nil)
 	_ planNode = (*viewNode)(nil)
 	_ planNode = (*lensNode)(nil)
+	_ planNode = (*operationNode)(nil)
```
AndrewSisley, line 37: nitpick: `MultiNode` implements/extends `planNode`, so this line is not required.

internal/planner/planner.go
```diff
@@ -593,32 +539,26 @@ func (p *Planner) RunRequest(
...
-	ctx context.Context,
-	request *request.Select,
-) (result []map[string]any, err error) {
-	planNode, err := p.makePlan(request)
+// MakeSelectionPlan makes a plan for a single selection.
+//
+// Note: Caller is responsible to call the `Close()` method to free the allocated
+// resources of the returned plan.
+//
+// @TODO {defradb/issues/368}: Test this exported function.
```
AndrewSisley, line 547: todo: This todo is incorrect - this is not a public function and does not require direct testing - if it is not used by us it can be deleted.

AndrewSisley (approved):
Looks good to me! Just a handful of small localized comments from me before merge.

internal/db/request.go
nasdf: Good catch! I've updated both to return maps.

internal/planner/mapper/mapper.go
nasdf: I've added a default case with an error.

internal/planner/mapper/operation.go
nasdf: Added `tests/integration/query/simple/with_operation_alias_test.go`

internal/planner/mapper/operation.go
nasdf: Done

internal/planner/operation.go
nasdf: Done

internal/planner/operation.go
nasdf: I didn't see anything in the spec about empty selections. I've removed this error since it should be fine to return empty results.

internal/planner/operations.go
nasdf: Done

internal/planner/planner.go
nasdf: Removed

PR #2882, opened 2024-07-30
feat: Operation selection in gql requests

### LINKED ISSUE #1395: Only first GraphQL request in a HTTP request is applied

nasdf:
## Relevant issue(s)

Resolves #1395

## Description

This PR allows multiple operations to be defined within a request and selected using an `operationName` param.

## Tasks

## How has this been tested?

added tests

- MacOS

### DIFF
--- a/cli/request.go
+++ b/cli/request.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/errors"
 )
 
@@ -26,8 +27,9 @@ const (
 
 func MakeRequestCommand() *cobra.Command {
 	var filePath string
+	var operationName string
 	var cmd = &cobra.Command{
-		Use:   "query [-i --identity] [request]",
+		Use:   "query [-i --identity] [request] [--operation <name>]",
 		Short: "Send a DefraDB GraphQL query request",
 		Long: `Send a DefraDB GraphQL query request to the database.
 
@@ -48,27 +50,28 @@ with the database more conveniently.
 
 To learn more about the DefraDB GraphQL Query Language, refer to https://docs.source.network.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			var request string
+			var request client.GQLRequest
 			switch {
 			case filePath != "":
 				data, err := os.ReadFile(filePath)
 				if err != nil {
 					return err
 				}
-				request = string(data)
+				request.Query = string(data)
 			case len(args) > 0 && args[0] == "-":
 				data, err := io.ReadAll(cmd.InOrStdin())
 				if err != nil {
 					return err
 				}
-				request = string(data)
+				request.Query = string(data)
 			case len(args) > 0:
-				request = string(args[0])
+				request.Query = string(args[0])
 			}
 
-			if request == "" {
-				return errors.New("request cannot be empty")
+			if request.Query == "" {
+				return errors.New("query cannot be empty")
 			}
+			request.OperationName = operationName
 
 			store := mustGetContextStore(cmd)
 			result := store.ExecRequest(cmd.Context(), request)
@@ -88,7 +91,7 @@ To learn more about the DefraDB GraphQL Query Language, refer to https://docs.so
 			return nil
 		},
 	}
-
 	cmd.Flags().StringVarP(&filePath, "file", "f", "", "File containing the query request")
+	cmd.Flags().StringVarP(&operationName, "operation", "o", "", "Name of the operation to execute")
 	return cmd
 }
--- a/client/db.go
+++ b/client/db.go
@@ -246,7 +246,16 @@ type Store interface {
 	GetAllIndexes(context.Context) (map[CollectionName][]IndexDescription, error)
 
 	// ExecRequest executes the given GQL request against the [Store].
-	ExecRequest(ctx context.Context, request string) *RequestResult
+	ExecRequest(ctx context.Context, request GQLRequest) *RequestResult
+}
+
+// GQLRequest represents a GraphQL request to execute.
+type GQLRequest struct {
+	// Query contains the GraphQL document that will be executed.
+	Query string `json:"query"`
+
+	// OperationName is the name of the operation in the parsed document to execute.
+	OperationName string `json:"operationName,omitempty"`
 }
 
 // GQLResult represents the immediate results of a GQL request.
--- a/client/request/request.go
+++ b/client/request/request.go
@@ -20,6 +20,41 @@ type Request struct {
 	Subscription []*OperationDefinition
 }
 
+// Operation returns the operation that should be executed from the request.
+//
+// The operationName is used to select an operation when multiple are defined.
+func (r *Request) Operation(operationName string) (*OperationDefinition, error) {
+	switch {
+	case len(r.Queries) == 1 && len(r.Mutations) == 0 && len(r.Subscription) == 0:
+		return r.Queries[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 1 && len(r.Subscription) == 0:
+		return r.Mutations[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 1:
+		return r.Subscription[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 0:
+		return nil, ErrMissingQueryOrMutation
+	}
+	for _, op := range r.Queries {
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	for _, op := range r.Mutations {
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	for _, op := range r.Subscription {
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	return nil, ErrMissingOperationName
+}
+
 type Selection any
 
 // Directives contains all the optional and non-optional
@@ -32,6 +67,7 @@ type Directives struct {
 }
 
 type OperationDefinition struct {
+	Name       string
 	Selections []Selection
 	Directives Directives
 }
--- a/internal/db/request.go
+++ b/internal/db/request.go
@@ -18,15 +18,15 @@ import (
 )
 
 // execRequest executes a request against the database.
-func (db *db) execRequest(ctx context.Context, request string) *client.RequestResult {
+func (db *db) execRequest(ctx context.Context, request client.GQLRequest) *client.RequestResult {
 	res := &client.RequestResult{}
-	ast, err := db.parser.BuildRequestAST(request)
+	ast, err := db.parser.BuildRequestAST(request.Query)
 	if err != nil {
 		res.GQL.Errors = []error{err}
 		return res
 	}
 	if db.parser.IsIntrospection(ast) {
-		return db.parser.ExecuteIntrospection(request)
+		return db.parser.ExecuteIntrospection(request.Query)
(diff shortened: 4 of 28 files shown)


client/db.go
```diff
@@ -246,7 +246,16 @@ type Store interface {
...
+	ExecRequest(ctx context.Context, request GQLRequest) *RequestResult
+}
+
+// GQLRequest represents a GraphQL request to execute.
+type GQLRequest struct {
+	// Query contains the GraphQL document that will be executed.
+	Query string `json:"query"`
+
+	// OperationName is the name of the operation in the parsed document to execute.
+	OperationName string `json:"operationName,omitempty"`
```
AndrewSisley, line 258: question: Why have you broken out the operation like this?

cli/request.go
```diff
@@ -26,8 +27,9 @@ const (
 
 func MakeRequestCommand() *cobra.Command {
 	var filePath string
+	var operationName string
 	var cmd = &cobra.Command{
-		Use:   "query [-i --identity] [request]",
+		Use:   "query [-i --identity] [request] [--operation <name>]",
```
AndrewSisley, line 32: question: Why are you limiting the CLI to a single operation at a time?

client/db.go
```diff
@@ -246,7 +246,16 @@ type Store interface {
 	GetAllIndexes(context.Context) (map[CollectionName][]IndexDescription, error)
 
 	// ExecRequest executes the given GQL request against the [Store].
-	ExecRequest(ctx context.Context, request string) *RequestResult
+	ExecRequest(ctx context.Context, request GQLRequest) *RequestResult
```
AndrewSisley, line 249: thought: I am very reluctant to support coupling this function so strongly to a single query language.

tests/integration/query/simple/with_operation_name_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+package simple
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestQuerySimpleWithMultipleOperationsAndOperationName_ShouldSucceed(t *testing.T) {
```
AndrewSisley, line 19: todo: Please do not create new tests using the old `RequestTestCase` - please use `TestCase`

tests/integration/query/simple/with_operation_name_test.go
```diff
@@ -0,0 +1,83 @@
...
+				`{
+					"Name": "John",
+					"Age": 21
+				}`,
+			},
+		},
+		Results: map[string]any{
+			"Users": []map[string]any{
+				{
+					"_docID": "bae-d4303725-7db9-53d2-b324-f3ee44020e52",
```
AndrewSisley, line 46: todo: The exact characters used in the docID do not appear to be important to this test, please use `NewDocIndex()` instead

cli/request.go
AndrewSisley: :) This is as per the gql spec, resolving thread

client/db.go
AndrewSisley: :) This is as per the gql spec, resolving thread

tests/integration/utils.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 type RequestTestCase struct {
```
AndrewSisley, line 17: todo: Please revert this file - we should not be extending this old legacy type that we are trying to remove.

tests/integration/test_case.go
```diff
@@ -535,6 +535,9 @@ type Request struct {
 	// The request to execute.
 	Request string
 
+	// Name of the operation within the request to execute.
```
AndrewSisley, line 538: todo: Please document that this property is optional.

suggestion: I do typically prefer `Option[string]` for optional string properties, I find it is clearer to consumers of the interface.  More than happy for you to shrug this off as personal preference though (please still do the todo though)

client/db.go
AndrewSisley: suggestion: Reworking this to use the option pattern would remove this coupling:

```
func WithOperation(operationName string) RequestOption {
  return func(o *gqlOptions) {
     o.OperationName = operationName
  }
}

ExecRequest(ctx context.Context, request string, opts ...RequestOption) *RequestResult
```

client/request/request.go
```diff
@@ -20,6 +20,41 @@ type Request struct {
...
+	case len(r.Queries) == 0 && len(r.Mutations) == 1 && len(r.Subscription) == 0:
+		return r.Mutations[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 1:
+		return r.Subscription[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 0:
+		return nil, ErrMissingQueryOrMutation
+	}
+	for _, op := range r.Queries {
```
AndrewSisley, line 40: todo: Please add an integration test documenting what happens when multiple operations of the same name are defined, especially when the operations of shared names are of different operation types (e.g. one is a query and one is a mutation).

todo: Please make sure that that behaviour is gql spec compliant.

todo: Please document whatever that behaviour is in a location very easy for users to find.

client/request/request.go
```diff
@@ -20,6 +20,41 @@ type Request struct {
...
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	for _, op := range r.Subscription {
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	return nil, ErrMissingOperationName
```
AndrewSisley, line 55: todo: This error is incorrect, as the operationName may have been provided, only it does not match any defined operations.  Please change this as otherwise it will likely confuse users at some point.

client/request/request.go
```diff
@@ -32,6 +67,7 @@ type Directives struct {
 }
 
 type OperationDefinition struct {
+	Name       string
```
AndrewSisley, line 70: todo: Sorry for the bother, as the other existing props are not documented, but this is a public member and as such should have some documentation on it.

internal/db/request.go
```diff
@@ -50,7 +50,7 @@ func (db *db) execRequest(ctx context.Context, request string) *client.RequestRe
 	identity := GetContextIdentity(ctx)
 	planner := planner.New(ctx, identity, db.acp, db, txn)
 
-	results, err := planner.RunRequest(ctx, parsedRequest)
+	results, err := planner.RunRequest(ctx, parsedRequest, request.OperationName)
```
AndrewSisley, line 53: todo: This appears to be a less than optimal code flow, as the operation is known by this point, and we can discard the operation name and all irrelevant operations.

Unless I have missed something, and the 'extra' ops are not necessarily irrelevant, I don't think the planner should be burdened with this unwanted information (and the hassle of selecting the correct operation).

AndrewSisley (changes requested):
I'm really glad you are adding this, but I think it needs a little more work before merge

client/db.go
nasdf: I've pictured the abstraction to look like this:

```go
ExecRequest(ctx context.Context, request *request.Request) *RequestResult
```

The language frontend would handle converting the query into the `request.Request` format and the result would be converted by the frontend into the expected result format.

This approach should result in the removal of any GQL specific things from the DB interface, but will also require reformatting all of the tests to use `request.Request` instead of gql queries.

AndrewSisley: I like this quite a lot and it would save embedded clients from a lot of parsing work.

I don't think the tests would need changing right away, initially you could just parse the gql strings into `request.Request` in the test framework (instead of internally as it is now).

The framework will always need to handle such a conversion anyway (to auto-test all query langs including the embedded), and although long term we may prefer writing most tests using the embedded, I think tests should be free to be written in any lang (and the test framework handle the conversion) - this would allow proper language specific testing.

Meaning we can have this in this PR without subjecting you to another really tedious reformat of all our tests.

client/request/request.go
```diff
@@ -20,6 +20,41 @@ type Request struct {
...
+		return r.Queries[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 1 && len(r.Subscription) == 0:
+		return r.Mutations[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 1:
+		return r.Subscription[0], nil
+
+	case len(r.Queries) == 0 && len(r.Mutations) == 0 && len(r.Subscription) == 0:
+		return nil, ErrMissingQueryOrMutation
```
shahzadlone, line 38: suggestion: In this case it is also missing subscription, perhaps rename the error type name and string?

shahzadlone, line 38: todo: This case is untested, add a test please if easy to do so.

client/request/request.go
```diff
@@ -20,6 +20,41 @@ type Request struct {
...
+	}
+	for _, op := range r.Queries {
+		if op.Name == operationName {
+			return op, nil
+		}
+	}
+	for _, op := range r.Mutations {
+		if op.Name == operationName {
+			return op, nil
```
shahzadlone, line 47: todo: I noticed none of the tests cover the `mutation` case. Please add tests for a mutation, for multiple mutations, and also for a mix of `query` and `mutation` in one request.

client/request/request.go
```diff
@@ -20,6 +20,41 @@ type Request struct {
...
+		}
+	}
+	for _, op := range r.Subscription {
+		if op.Name == operationName {
+			return op, nil
```
shahzadlone, line 52: suggestion: Similar to above mutation comment test in combination with `mutation` and `query`. However if this is too much trouble don't bother.

internal/request/graphql/parser/mutation.go
```diff
@@ -38,6 +38,9 @@ func parseMutationOperationDefinition(
 	qdef := &request.OperationDefinition{
 		Selections: make([]request.Selection, len(def.SelectionSet.Selections)),
 	}
+	if def.Name != nil {
+		qdef.Name = def.Name.Value
```
shahzadlone, line 42: question: What is this for?

nitpick: No coverage.

internal/request/graphql/parser/subscription.go
```diff
@@ -26,6 +26,9 @@ func parseSubscriptionOperationDefinition(
 	sdef := &request.OperationDefinition{
 		Selections: make([]request.Selection, len(def.SelectionSet.Selections)),
 	}
+	if def.Name != nil {
+		sdef.Name = def.Name.Value
```
shahzadlone, line 30: nitpick: Same as above.

tests/integration/query/simple/with_operation_name_test.go
```diff
@@ -0,0 +1,83 @@
...
+						_docID
+						Age
+					}
+				}
+				query usersWithName {
+					Users {
+						_docID
+						Name
+					}
+				}`,
```
shahzadlone, line 70: nitpick: Not sure if it's just github preview issue, but the indentation seems off.

tests/integration/test_case.go
shahzadlone: I prefer string over `Option[string]` here.

shahzadlone (changes requested):
Awesome PR, excited to see this come in.

Most of my comments/requests are for testing.

nasdf:
Included in #2993

PR #2891, opened 2024-08-07
feat: Secure document encryption key exchange

### LINKED ISSUE #2856: Doc Encryption: implement key exchange between nodes.
A node that has a permission to access an encrypted document should be able to retrieve an encryption key from peer.

islamaliev:
## Relevant issue(s)

Resolves #2856 #2909

## Description

This PR introduces a secure mechanism for exchanging document encryption keys in our decentralized network. It supports both whole-document and field-level encryption, enhancing data privacy without compromising the system's distributed nature.

### Data Flow and Key Exchange Process

1. Document Creation:
   * When a user creates a document with encryption enabled (whole document or specific fields), the system generates a new DAG block containing the encrypted delta.
   * A new `Encryption` IPLD block is created, containing the docID, optional fieldName (for field-level encryption), and the encryption key itself.
   * The DAG block references the `Encryption` block by storing its CID.

2. Encryption Detection:
   * Upon receiving an update event for a document, a node checks if the block is encrypted by examining the reference to the `Encryption` block.
   * If encrypted and the node lacks the corresponding key, it initiates a key fetch request.

3. Key Request and Retrieval:
   * The node sends an "enc-keys-request" event on the event bus, which is picked up by the Key Management Service (KMS).
   * The KMS handles the key exchange process over the "encryption" pubsub channel.

4. Key Reception and Storage:
   * Upon receiving the encryption keys, the KMS stores them in a dedicated IPLD storage.
   * It then notifies the merging process that is waiting for the results.

5. Decryption and Merging:
   * During the `executeMerge` process, the node collects CIDs of all `Encryption` blocks that are not available locally.
   * After the initial merge attempt, if any encryption keys are missing, the node requests these keys via the KMS.
   * Once the keys are received, the merge process is restarted to decrypt and apply the previously encrypted deltas.

### Technical Details

* Encryption: ECIES with X25519, HKDF (SHA-256), and AES-256-GCM
* Key Storage: Encryption keys are stored in separate IPLD blocks, referenced by the main DAG blocks
* Key Management: Introduced a new KMS abstraction for handling key exchange and storage
* Pubsub: All peers listen on the "encryption" topic for key exchange messages
* Event Bus: New "enc-keys-request" event for initiating key requests

### Current Limitations and Future Work

* This version assumes all documents are public. Future iterations will integrate with access control mechanisms.
* The system currently trusts the first received response. Future enhancements will implement more robust validation.
* Further work is needed to optimize the key exchange process for large-scale deployments and to handle network partitions or failed exchanges gracefully.
* The current KMS implementation can be replaced in the future, e.g., with an Orbis KMS, for more advanced key management features.

### Testing Improvements

Introduced an "assert stack" for integration tests, providing detailed failure contexts (e.g., 'path: commits[2].links[1].cid' instead of 'doc: 1').

These changes significantly enhance our system's security, enabling confidential data storage and transmission in a decentralized environment, while laying the groundwork for more advanced encryption and access control features. The introduction of the KMS abstraction provides flexibility for future improvements in key management.

## Tasks

## How has this been tested?

With integration tests.

- MacOS


net/server.go
```diff
@@ -164,6 +207,108 @@ func (s *server) PushLog(ctx context.Context, req *pb.PushLogRequest) (*pb.PushL
...
+	}
+
+	res := &pb.FetchEncryptionKeyReply{
+		SchemaRoot:            req.SchemaRoot,
+		ReqEphemeralPublicKey: req.EphemeralPublicKey,
+		Targets:               targets,
+		EncryptedKeys:         encryptedKey,
+	}
+
+	res.Signature, err = s.signResponse(res)
```
nasdf, line 280: thought: if these signatures are only used to verify the sender is authenticate the pubsub implementation can handle this automatically https://github.com/libp2p/go-libp2p-pubsub/blob/master/sign.go

islamaliev: that's a good point. @jsimnz what do you think?

tests/integration/assert_stack.go
```diff
@@ -0,0 +1,53 @@
...
+// licenses/APL.txt.
+
+package tests
+
+import (
+	"strconv"
+	"strings"
+)
+
+type assertStack struct {
```
AndrewSisley, line 18: question: Why does this exist?

thought: It looks like a notable complication to our already complicated (and somewhat fragile yet really really important) assertion logic, and it appears that it is only used for a couple of log lines.

suggestion: Remove this complication (if I'm right and it is just used for a log line).

todo: If keeping, please document it very well.

tests/integration/events.go
```diff
@@ -349,6 +351,40 @@ func getEventsForCreateDoc(s *state, action CreateDoc) map[string]struct{} {
...
+			}
+
+		case <-time.After(eventTimeout):
+			require.Fail(s.t, "timeout waiting for key retrieved event")
+		}
+	}
+}
+
+func waitForSync(s *state, action WaitForSync) {
+	count := int(action.Count)
```
AndrewSisley, line 373: todo: This looks like it is wrong, and would result on serial waits and timeouts for concurrent operations.  In particular, this will make some failures very slow - e.g. if `count == 10`, then the effective timeout would be `10 * eventTimeout`.

Please rework this, I suggest using a go routine + waitGroup.

AndrewSisley (commented):
praise: Thank you very much for the excellent PR description, it was very useful.

The code/feat looks good, but I lack the confidence to approve the P2P/encryption stuff (to make sure we aren't leaking anything), so I'll leave you my small todo's and wait for someone else to approve.

jsimnz:
Islam and I agreed on a notable refactor for how keys are referenced in blocks. So will properly review after the changes land.

tests/integration/assert_stack.go
islamaliev: the time I spent on implementing it already payed off in double. If you have a large result to assert on (like a commit query) it show exactly where the mismatch happened.
For example, it shows `path: commits[2].links[1].cid` instead of previous `doc: 1`
So, I would want to keep it.
I will add some documentation.

internal/core/block/block.go
```diff
@@ -97,27 +99,71 @@ func NewDAGLink(name string, link cidlink.Link) DAGLink {
...
-	// It needs to be a pointer so that it can be translated from and to `optional Bool` in the IPLD schema.
-	IsEncrypted *bool
+	// Encryption contains the encryption information for the block's delta.
+	// It needs to be a pointer so that it can be translated from and to `optional` in the IPLD schema.
+	Encryption *Encryption
+}
+
+// IsEncrypted returns true if the block is encrypted.
+func (b *Block) IsEncrypted() bool {
+	return b.Encryption != nil && (*b.Encryption).Type != NotEncrypted
```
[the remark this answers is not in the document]
islamaliev: good catch. Fixed

tests/integration/assert_stack.go
AndrewSisley: Okay nice :)  Thanks Islam.

internal/db/messages.go
```diff
@@ -78,6 +79,38 @@ func (db *db) handleMessages(ctx context.Context, sub *event.Subscription) {
...
+							encStoreKey.DocID,
+							encStoreKey.FieldName,
+							encStoreKey.BlockHeight,
+							encKey,
+						)
+
+						if err != nil {
+							log.ErrorContextE(
+								ctx,
+								"Failed to save doc encryption key",
```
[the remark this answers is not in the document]
islamaliev: extracted to const strings

internal/db/messages.go
```diff
@@ -78,6 +79,38 @@ func (db *db) handleMessages(ctx context.Context, sub *event.Subscription) {
 						log.ErrorContextE(ctx, "Failed to load replicators", err)
 					}
 				})
+
+			case encryption.KeyRetrievedEvent:
```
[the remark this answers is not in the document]
islamaliev: extracted and documented

net/server.go
```diff
@@ -82,6 +98,24 @@ func newServer(p *Peer, opts ...grpc.DialOption) (*server, error) {
 	return s, nil
 }
 
+func (s *server) extractSessionAndRemoveOldOnes(id string) *session {
```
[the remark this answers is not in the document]
islamaliev: renamed and documented

tests/integration/encryption/peer_share_test.go
```diff
@@ -0,0 +1,407 @@
...
+			},
+			testUtils.SubscribeToCollection{
+				NodeID:        1,
+				CollectionIDs: []int{0},
+			},
+			testUtils.UpdateDoc{
+				NodeID: immutable.Some(0),
+				Doc:    `{"age": 2}`,
+			},
+			testUtils.WaitForSync{
```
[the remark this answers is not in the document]
islamaliev: what do you mean "why"? The reason is the same as with all others peers-related tests.
No, if I remove it, it will not fail. But if you run other integration tests and comment out `WaitForSync` it will also pass.
It is there to make sure the test is not flaky.
Is your question specifically about `encryption.KeysRetrievedEventName` or sync in general?
And I'm not actually sure which new information documentation can add. It is pretty explicit: wait on node 1 for keys-retrieved event.

tests/integration/events.go
```diff
@@ -349,6 +351,40 @@ func getEventsForCreateDoc(s *state, action CreateDoc) map[string]struct{} {
 	return expect
 }
 
+func waitForKeyRetrievedEvent(s *state, nodeIDs []int) {
+	for nodeID := 0; nodeID < len(s.nodes); nodeID++ {
+		if len(nodeIDs) > 0 && !slices.Contains(nodeIDs, nodeID) {
```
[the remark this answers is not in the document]
islamaliev: > len(nodeIDs) > 0 is impossible given the loop constraints
which constraints? `nodeIDs` is defined in test case

I renamed it and added some comments

tests/integration/events.go
islamaliev: done

tests/integration/p2p.go
```diff
@@ -134,7 +137,16 @@ type GetAllP2PCollections struct {
...
 // node 1 to see if it has been replicated.
-type WaitForSync struct{}
+type WaitForSync struct {
+	// Event is the name of the event to wait for.
+	// If is not provided then the action will wait for MergeComplete event.
+	Event immutable.Option[event.Name]
+	// NodeIDs are the node IDs (indexes) of the nodes to wait for sync on.
+	// If not provided then the action will wait for sync on all nodes.
+	NodeIDs []int
+	// Count is the number of times the event should be received.
```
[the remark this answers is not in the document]
islamaliev: done

tests/integration/p2p.go
```diff
@@ -134,7 +137,16 @@ type GetAllP2PCollections struct {
...
+	Count uint
```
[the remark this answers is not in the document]
islamaliev: done

net/server.go
islamaliev: as discussed on a stand up, it's fine to leave it like this

tests/integration/encryption/peer_share_test.go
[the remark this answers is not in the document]
AndrewSisley: > It is pretty explicit: wait on node 1 for keys-retrieved event.

Why does it need to wait for the keys-retrieved event?  What happens if the request is processed before the event?

tests/integration/events.go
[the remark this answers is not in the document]
AndrewSisley: Ah got it, thanks Islam - I see where I misread, the new name and comment helped a fair bit thanks :)

tests/integration/encryption/peer_share_test.go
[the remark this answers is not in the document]
islamaliev: we wait for not 1 to received the encryption key. If we do it, node 1 won't be able to fetch decrypted doc in the request (following action)

I don't see why it needs to be commented.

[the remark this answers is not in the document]
AndrewSisley: > If we don't do it, node 1 won't be able to fetch decrypted doc in the request (following action)

What happens if it tries to do so?

> I don't see why it needs to be commented.

It is far from obvious what the public behaviour is (I still dont know), and it could be a fairly common thing for users to come across.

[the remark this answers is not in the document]
islamaliev: > What happens if it tries to do so? Does it error, panic, return an empty set, wait for the key-sync during the request?

If node 1 doesn't receive the enc key and makes gql request, the result will be empty. I actually had a test that shows this in my previous PR that I remove in this PR 
https://github.com/sourcenetwork/defradb/pull/2891/files#diff-6fd6bae6910d2da0509cd14c385e002a3c3969665086213ed36bc37d28b2927cL21-L56

In the previous PR there was no key exchange, so all keys were treated as private, but the DAG anyway was synced.
With this PR all docs are assumed to be public because a node gives keys to whoever requests.

Once we are done with permission granting they can be married together.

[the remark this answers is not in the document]
islamaliev: `AnyOf` is supported only for doc's field values, not for the entire doc. 

But the idea with test is nice. Anyway we will have bunch of tests like this once permission sharing is in place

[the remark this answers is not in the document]
islamaliev: brought back the test with your `AnyOf` suggestion. Had to tweak our testing framework a bit

internal/db/merge.go
```diff
@@ -386,6 +587,62 @@ func getHeadsAsMergeTarget(ctx context.Context, txn datastore.Txn, dsKey core.Da
...
+	for len(cids) > 0 {
+		cid := cids[0]
+		block, err := loadBlockFromBlockStore(ctx, txn, cid)
+		if err != nil {
+			return nil, err
+		}
+
+		if block.Encryption != nil && bytes.Equal(block.Encryption.KeyID, []byte(keyID)) {
+			blocks = append(blocks, block)
+			prevCid := block.GetPrevBlockCid()
```
jsimnz, line 636: todo: Same issue here regarding `GetPrevBlockCid()`

internal/encryption/encryptor.go
```diff
@@ -42,21 +45,23 @@ func generateEncryptionKey(_, _ string) ([]byte, error) {
...
 // It acts based on the configuration [DocEncConfig] provided and data stored in the provided store.
 // It uses [core.EncStoreDocKey] to store and retrieve encryption keys.
 type DocEncryptor struct {
-	conf  immutable.Option[DocEncConfig]
-	ctx   context.Context
-	store datastore.DSReaderWriter
+	conf          immutable.Option[DocEncConfig]
+	ctx           context.Context
+	store         datastore.DSReaderWriter
+	cache         map[core.EncStoreDocKey][]byte
```
jsimnz, line 59: todo: This is an infinite cache?

internal/encryption/encryptor.go
```diff
@@ -42,21 +45,23 @@ func generateEncryptionKey(_, _ string) ([]byte, error) {
...
+	generatedKeys []core.EncStoreDocKey
```
jsimnz, line 60: question: Was this helping you solve the "chicken and egg" problem you described? 

The problem here is that similar to the cache, this is an infinite growing list. I also don't think its necessary. I believe your issue is how you are trying to resolve the key for a given document. However, you have already stored the key in the `EncStore` with a DocID and FieldID prefix, you should be able to query that store for the generated keys.

internal/encryption/encryptor.go
```diff
@@ -69,136 +74,266 @@ func (d *DocEncryptor) SetStore(store datastore.DSReaderWriter) {
...
+	}
+	return encryptionKey, nil
+}
+
+// getGeneratedKeyFor returns the generated key for the given docID and fieldName.
+func (d *DocEncryptor) getGeneratedKeyFor(
+	docID string,
+	fieldName immutable.Option[string],
+) (immutable.Option[core.EncStoreDocKey], []byte) {
+	for _, key := range d.generatedKeys {
```
jsimnz, line 188: todo: You don't need to keep this list of generated keys to be able to resolve this function. Unless specifically told otherwise, the DB should continue to use the key from the last block within the document. If this is the first block (create document) then we need to generate a new key.

You can get the last generated key from the current head of the document in the headstore.

internal/merkle/clock/clock.go
```diff
@@ -86,20 +87,20 @@ func (mc *MerkleClock) AddDelta(
 	delta.SetPriority(height)
 	block := coreblock.New(delta, links, heads...)
 
-	isEncrypted, err := mc.checkIfBlockEncryptionEnabled(ctx, block.Delta.GetFieldName(), heads)
+	fieldName := immutable.None[string]()
+	if block.Delta.GetFieldName() != "" {
+		fieldName = immutable.Some(block.Delta.GetFieldName())
+	}
+	blockEnc, err := mc.determineBlockEncryptionData(ctx, string(block.Delta.GetDocID()), fieldName, heads)
```
jsimnz, line 94: todo: This is the issue referenced above regarding the `generatedKeys` slice in the `encryptor` struct.

The `generatedKeys` is used to keep track of the locally generated keys to quickly get them for later use. However, this isn't necessary because we already have the `heads` here at this call site, which means we can get the last keyID of the head (or deterministically choose a key from multiple heads if they exist, eg: lexigraphical CID compare).

jsimnz (changes requested):
Couple of notable issues raised. Some smaller thoughts/concerns.


- Not correctly seeking backwards through a DAG because you only grab the first head GetPrevBlockForCid() func
- The encStore cache and generatedKeys slice in the encryptor aren't necessary. And cause a huge (infinite) amount of overhead that is already accessible elsewhere

internal/encryption/encryptor.go
jsimnz: Update: Should still just delete the cache though ;)

internal/encryption/encryptor.go
```diff
@@ -42,21 +45,23 @@ func generateEncryptionKey(_, _ string) ([]byte, error) {
...
 
 // DocEncryptor is a document encryptor that encrypts and decrypts individual document fields.
 // It acts based on the configuration [DocEncConfig] provided and data stored in the provided store.
 // It uses [core.EncStoreDocKey] to store and retrieve encryption keys.
 type DocEncryptor struct {
-	conf  immutable.Option[DocEncConfig]
-	ctx   context.Context
-	store datastore.DSReaderWriter
+	conf          immutable.Option[DocEncConfig]
+	ctx           context.Context
```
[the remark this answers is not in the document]
islamaliev: on the call we discussed that an instance of `encryptor` is context-(or session-)bound. So keeping this in mind I think your comment is resolved, right?

internal/core/block/block.go
```diff
@@ -97,27 +99,71 @@ func NewDAGLink(name string, link cidlink.Link) DAGLink {
...
+	Encryption *Encryption
+}
+
+// IsEncrypted returns true if the block is encrypted.
+func (b *Block) IsEncrypted() bool {
+	return b.Encryption != nil && b.Encryption.Type != NotEncrypted
+}
+
+// GetPrevBlockCid returns the CID of the previous block.
+func (b *Block) GetPrevBlockCid() immutable.Option[cid.Cid] {
```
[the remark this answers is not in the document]
islamaliev: changed to return all heads

internal/encryption/encryptor.go
islamaliev: deleted cache

internal/db/merge.go
islamaliev: fixed by returning all heads

internal/db/merge.go
```diff
@@ -190,13 +314,12 @@ func (mp *mergeProcessor) loadComposites(
...
 	if block.Delta.GetPriority() >= mt.headHeight {
-		mp.composites.PushFront(block)
-		for _, link := range block.Links {
-			if link.Name == core.HEAD {
-				err := mp.loadComposites(ctx, link.Cid, mt)
-				if err != nil {
-					return err
-				}
+		mp.blocks.PushFront(block)
+		prevCid := block.GetPrevBlockCid()
```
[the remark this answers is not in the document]
islamaliev: added a test task https://github.com/sourcenetwork/defradb/issues/2910

net/pb/net.proto
```diff
@@ -42,10 +34,51 @@ message PushLogRequest {
...
+    bytes keyID = 3;
+}
+
+// FetchEncryptionKeyRequest is a request to receive a doc encryption key
+// from a peer that holds it.
+message FetchEncryptionKeyRequest {
+    // targets is the list of docs/fields for which encryption keys are being requested.
+    repeated EncryptionKeyTarget targets = 1;
+    // schemaRoot is the SchemaRoot of the collection that the document resides in.
+    bytes schemaRoot = 2;
```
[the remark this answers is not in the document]
islamaliev: it's included in the request so that we can receive it back in the reply. After we received enc key we send an event that included schema root and the enc key. The merger handling the event can assosiate the doc with the collection based on the schema root. Otherwise it doesn't know which collection it belongs to.
But I think we can include schema root in a session the same way we do with ephemeral keys. Will this be fine? Or you want to propose another way of linking a doc to a collection?

internal/db/merge.go
```diff
@@ -129,12 +245,14 @@ func (m *mergeQueue) done(docID string) {
...
-	mCRDTs     map[string]merklecrdt.MerkleCRDT
-	col        *collection
-	dsKey      core.DataStoreKey
-	composites *list.List
+	txn                          datastore.Txn
+	lsys                         linking.LinkSystem
+	mCRDTs                       map[string]merklecrdt.MerkleCRDT
+	col                          *collection
+	dsKey                        core.DataStoreKey
+	blocks                       *list.List
```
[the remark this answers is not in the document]
islamaliev: yes, field-level blocks with field-level encryption can/shoud be merged separately.

If a node needs an enc key only for a field and receives it why would it touch anything else? CIDs are the same, blockstore and headstore are not changed by decryption-pass merging. It's only datastore.

net/pb/net.proto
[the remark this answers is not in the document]
jsimnz: We can leave it as is. 👍

internal/core/block/block.go
```diff
@@ -238,3 +287,20 @@ func GetLinkPrototype() cidlink.LinkPrototype {
...
+
+// Validate checks if the block is valid.
+func (b *Block) Validate() error {
+	if b.Encryption != nil {
+		switch (*b.Encryption).Type {
+		case NotEncrypted, DocumentEncrypted, FieldEncrypted:
+			// Valid values
+		default:
+			return ErrInvalidBlockEncryptionType
+		}
```
shahzadlone, line 299: suggestion: Why do you need to explicitly dereference `*b.Encryption` why not just `b.Encryption` (missing C++ ? haha)

nitpick: I prefer just an `if` block over a `switch` here:

```suggestion
		eType := (b.Encryption).Type
		if eType != NotEncrypted && eType != DocumentEncrypted && eType != FieldEncrypted {
			return ErrInvalidBlockEncryptionType☆
		}
```

internal/core/block/block_test.go
shahzadlone, line 1: praise: thanks for these tests.

internal/db/merge.go
```diff
@@ -58,24 +60,28 @@ func (db *db) executeMerge(ctx context.Context, dagMerge event.Merge) error {
 		return err
 	}
 
-	mt, err := getHeadsAsMergeTarget(ctx, txn, dsKey)
+	mt, err := getHeadsAsMergeTarget(ctx, txn, dsKey.WithFieldId(core.COMPOSITE_NAMESPACE))
```
shahzadlone, line 63: nitpick: I see you updated some `fieldId` to `fieldID`.

Missed this one from `func (k DataStoreKey) WithFieldId(fieldId string) DataStoreKey {`

internal/db/merge.go
```diff
@@ -88,6 +94,116 @@ func (db *db) executeMerge(ctx context.Context, dagMerge event.Merge) error {
...
+
+		mp, err := db.newMergeProcessor(txn, ls, col, dsKey)
+		if err != nil {
+			return err
+		}
+
+		var blocks []*coreblock.Block
+		// if merge ground includes composite, we process only the composite block and skip the field
+		// blocks, as they will be processed as part of the composite block anyway.
+		// Otherwise, we load the blocks for each field
```
shahzadlone, line 150: praise: Thanks for this comment!

internal/db/merge.go
```diff
@@ -58,24 +60,28 @@ func (db *db) executeMerge(ctx context.Context, dagMerge event.Merge) error {
...
 	}
 
-	mt, err := getHeadsAsMergeTarget(ctx, txn, dsKey)
+	mt, err := getHeadsAsMergeTarget(ctx, txn, dsKey.WithFieldId(core.COMPOSITE_NAMESPACE))
 	if err != nil {
 		return err
 	}
 
-	err = mp.loadComposites(ctx, dagMerge.Cid, mt)
+	err = mp.loadBlocks(ctx, dagMerge.Cid, mt, false)
```
shahzadlone, line 68: question: why the name change from composites to blocks now?

internal/db/merge.go
```diff
@@ -129,12 +245,14 @@ func (m *mergeQueue) done(docID string) {
...
-	dsKey      core.DataStoreKey
-	composites *list.List
+	txn                          datastore.Txn
+	lsys                         linking.LinkSystem
+	mCRDTs                       map[string]merklecrdt.MerkleCRDT
+	col                          *collection
+	dsKey                        core.DataStoreKey
+	blocks                       *list.List
+	pendingEncryptionKeyRequests map[core.EncStoreDocKey]struct{}
+	hasPendingCompositeBlock     bool
```
shahzadlone, line 255: suggestion: I know this is a private struct, but would be helpful to have a quick one-liner documentation on the last three.

internal/db/merge.go
```diff
@@ -406,10 +663,15 @@ func syncIndexedDoc(
 		return err
 	}
 
-	if isDeletedDoc {
-		return col.deleteIndexedDoc(ctx, oldDoc)
-	} else if isNewDoc {
+	// this can happen we received an encrypted document that we haven't decrypted yet
```
shahzadlone, line 666: typo:

```suggestion
	// this can happen if we received an encrypted document that we haven't decrypted yet
```

net/pb/Makefile
```diff
@@ -10,14 +14,15 @@ deps:
...
-	$<
+	$< # This line specifies the input file
 
 clean:
 	rm -f *.pb.go
 	rm -f *pb_test.go
 
-.PHONY: clean
\ No newline at end of file
+.PHONY: clean deps
```
shahzadlone, line 28: question: Why did you add `deps` here?

net/pb/Makefile
shahzadlone, line 1: question(out-of-scope): Wonder if we need a ci action in future to ensure generated proto code is up to date?

net/server_encryption_key.go
```diff
@@ -0,0 +1,351 @@
...
+			encryption.ContextWithStore(ctx, s.peer.encstore),
+			core.NewEncStoreDocKey(docID.String(), optFieldName, string(target.KeyID)),
+		)
+		if err != nil {
+			return nil, nil, err
+		}
+		// TODO: we should test it somehow. For this this one peer should have some keys and
+		// another one should have the others. https://github.com/sourcenetwork/defradb/issues/2895
+		if len(encKey) == 0 {
+			continue
```
shahzadlone, line 68: praise: Thanks for leaving this comment here and the issue is quite descriptive as well.

shahzadlone (commented):
There is a lot to like in this PR (lots of praises), the use of encryption event was cool and the modification of the testing framework stuff like `WaitForSync` is nice additions. I didn't spot anything major , mostly are questions. Will leave the approval to John.

internal/db/merge.go
islamaliev: https://github.com/sourcenetwork/defradb/pull/2891#discussion_r1716463753

internal/db/merge.go
islamaliev: added some docs

net/pb/Makefile
islamaliev: I thought it was the the best practice to make sure the right target is always called. Even if it's not strictly necessary, it doesn't hurt AFAIK.
Do you have a different opinion?

net/pb/Makefile
islamaliev: I think it would be very beneficial, otherwise everybody does it his own way

crypto/nonce.go
```diff
@@ -18,12 +18,12 @@ import (
...
 )
 
-const nonceLength = 12
+const AESNonceSize = 12
 
 var generateNonceFunc = generateNonce
 
 func generateNonce() ([]byte, error) {
-	nonce := make([]byte, nonceLength)
+	nonce := make([]byte, AESNonceSize)
```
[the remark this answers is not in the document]
islamaliev: This function is executed only in production, not for tests.
Look at `init` function on the bottom to see why.

I know it's not ideal and it's better to use production encryption in tests as well.
But the idea to make it deterministic for tests came up after a disscussion with John where he mentioned that users might choose to use a different encryption algorithm for docs. And in order to test if it really works it should be deterministic. Andy had some minor work on his own branch to tackle this.

net/pb/Makefile
shahzadlone: Don't we want deps to run every time? Even if we don't, shouldn't it just be added above with the deps target as a separate phony target with only deps.

More reference on phony
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

islamaliev: added PHONY above each target

net/pb/Makefile
shahzadlone: might be worthwhile to make a ticket then :)

crypto/nonce.go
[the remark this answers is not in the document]
shahzadlone: But some lines show as covered, shouldn't the whole function be uncovered in that case?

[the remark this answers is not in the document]
islamaliev: that's strange

net/pb/Makefile
islamaliev: https://github.com/sourcenetwork/defradb/issues/2917

crypto/ecies.go
```diff
@@ -0,0 +1,159 @@
...
+//   - publicKey: The recipient's X25519 public key
+//   - associatedData: Optional associated data for additional authentication
+//
+// Returns:
+//   - Byte slice containing the encrypted message and necessary metadata for decryption
+//   - Error if any step of the encryption process fails
+func EncryptECIES(plainText []byte, publicKey *ecdh.PublicKey, associatedData []byte) ([]byte, error) {
+	ephemeralPrivate, err := GenerateX25519()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ephemeral key: %w", err)
```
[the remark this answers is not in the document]
islamaliev: done

internal/merkle/clock/clock.go
```diff
@@ -127,57 +140,96 @@ func (mc *MerkleClock) AddDelta(
 	return link, b, err
 }
 
-func (mc *MerkleClock) checkIfBlockEncryptionEnabled(
+// TODO: rename this to reflect storage
+func (mc *MerkleClock) determineBlockEncryptionData(
```
fredcarle, line 144: todo: Remove todo and maybe rename to just `determineBlockEncryption`

fredcarle (commented):
Initial batch of comment bellow. There are a few parts I want to go over again but overall it looks really nice.

cli/collection_create.go
```diff
@@ -121,9 +121,6 @@ func setContextDocEncryption(cmd *cobra.Command, shouldEncryptDoc bool, encryptF
 		return
 	}
 	ctx := cmd.Context()
-	if txn != nil {
```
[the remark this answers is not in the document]
islamaliev: done

internal/core/block/block.go
```diff
@@ -97,27 +105,78 @@ func NewDAGLink(name string, link cidlink.Link) DAGLink {
...
+func (block *Block) Clone() *Block {
+	return &Block{
+		Delta:      block.Delta.Clone(),
+		Links:      block.Links,
+		Encryption: block.Encryption,
+	}
+}
+
+// GetPrevBlockCids returns the CIDs of the previous blocks. It can be more than 1 with multiple heads.
+func (block *Block) GetPrevBlockCids() []cid.Cid {
```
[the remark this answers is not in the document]
islamaliev: done

internal/db/config.go
```diff
@@ -19,12 +19,16 @@ const (
...
 // Option is a funtion that sets a config value on the db.
-type Option func(*db)
+type Option func(*dbOptions)
 
 // WithMaxRetries sets the maximum number of retries per transaction.
 func WithMaxRetries(num int) Option {
-	return func(db *db) {
-		db.maxTxnRetries = immutable.Some(num)
+	return func(opts *dbOptions) {
+		opts.maxTxnRetries = immutable.Some(num)
```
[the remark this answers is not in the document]
islamaliev: I considered adding KMS options as part of db options and noticed that it's not quite flexible, as it didn't allow post-processing of options, since it directly manipulates the values on `db` object.
So I changed it. But then I decided to move KMS elsewhere.
I can revert this change to options, but this style also seems to align with the way we do it in other places.

internal/db/merge.go
```diff
@@ -227,47 +234,162 @@ func (mp *mergeProcessor) loadComposites(
...
+) (*coreblock.Encryption, error) {
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.SetReadStorage(mp.txn.Encstore().AsIPLDStorage())
+
+	_, blockCid, err := cid.CidFromBytes(encLink.Cid.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	nd, err := lsys.Load(linking.LinkContext{Ctx: ctx}, cidlink.Link{Cid: blockCid},
```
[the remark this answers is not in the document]
islamaliev: done

internal/merkle/clock/clock.go
islamaliev: done

internal/db/merge.go
```diff
@@ -227,47 +234,162 @@ func (mp *mergeProcessor) loadComposites(
...
+
+	mp.availableEncryptionBlocks[encLink] = encBlock
+
+	return encBlock, nil
+}
+
+// processEncryptedBlock decrypts the block if it is encrypted and returns the decrypted block.
+// If the block is encrypted and we were not able to decrypt it, it returns true as the second return value
+// which indicates that the we should skip merging of the block.
+// If we were able to decrypt the block, we return the decrypted block and false as the second return value.
```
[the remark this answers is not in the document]
islamaliev: yeah, it described the old behavior. Thanks for spotting this!

internal/db/merge.go
```diff
@@ -227,47 +234,162 @@ func (mp *mergeProcessor) loadComposites(
...
+	if _, ok := mp.missingEncryptionBlocks[encLink]; ok {
+		return nil, nil
+	}
+
+	encBlock, err := mp.loadEncryptionBlock(ctx, encLink)
+	if err != nil {
+		return nil, err
+	}
+
+	mp.availableEncryptionBlocks[encLink] = encBlock
```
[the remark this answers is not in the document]
islamaliev: nice catch!

internal/db/merge.go
```diff
@@ -227,47 +234,162 @@ func (mp *mergeProcessor) loadComposites(
...
 		}
 	}
 	return nil
 }
 
+func (mp *mergeProcessor) loadEncryptionBlock(
+	ctx context.Context,
+	encLink cidlink.Link,
+) (*coreblock.Encryption, error) {
+	lsys := cidlink.DefaultLinkSystem()
```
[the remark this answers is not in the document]
islamaliev: done

internal/db/merge.go
```diff
@@ -282,17 +404,39 @@ func (mp *mergeProcessor) processBlock(
...
+		return block, nil
+	}
+
+	bytes, err := encryptor.Decrypt(block.Delta.GetData(), encBlock.Key)
+	if err != nil {
+		return nil, err
+	}
+	if len(bytes) == 0 {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
islamaliev: we encryptd CBOR-encoded values which has byte representation for empty strings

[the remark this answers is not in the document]
islamaliev: added to more tests that ensure encryption works with empty strings and nil values

internal/kms/pubsub.go
```diff
@@ -0,0 +1,325 @@
...
+
+// handleEncryptionMessage handles incoming FetchEncryptionKeyRequest messages from the pubsub network.
+func (s *pubSubService) handleRequestFromPeer(peerID libpeer.ID, topic string, msg []byte) ([]byte, error) {
+	req := new(pb.FetchEncryptionKeyRequest)
+	if err := proto.Unmarshal(msg, req); err != nil {
+		log.ErrorContextE(s.ctx, "Failed to unmarshal pubsub message %s", err)
+		return nil, err
+	}
+
+	ctx := grpcpeer.NewContext(s.ctx, net.NewGRPCPeer(peerID))
```
fredcarle, line 139: todo: It doesn't make sense to create a dependency to the net package only for this. You can even create the `grpcpeer.Peer` struct inline if you want.

net/server.go
```diff
@@ -409,3 +409,47 @@ func (s *server) updateReplicators(evt event.Replicator) {
 	}
 	s.peer.bus.Publish(event.NewMessage(event.ReplicatorCompletedName, nil))
 }
+
+func (s *server) AddPubSubTopic(topicName string, handler rpc.MessageHandler) error {
```
fredcarle, line 413: todo: `addPubSubTopic` already exists. You can simply call it from this function if you need it to be public.

fredcarle (approved):
LGTM. Just a couple todos to resolve before merge. Thanks Islam!

internal/kms/pubsub.go
islamaliev: done

Issue #2906, opened 2024-08-12
Revoke or unshare a document previously created a relation for


shahzadlone:
Revoke or unshare a document previously created a relation for

Issue #2905, opened 2024-08-12
Split ACP `write` perm into `delete` and `update`


shahzadlone:
Split ACP `write` perm into `delete` and `update`

PR #2907, opened 2024-08-12
feat: Ability to relate private documents to actors

### LINKED ISSUE #2762: Ability to share a resource of  DAC (document access control)
Ability to share document access through acp

shahzadlone:
## Relevant issue(s)
Resolves #2762

## Description
This PR introduces the ability to make use of the `relation`s defined within a policy to create relationships between an actor and a document within a collection. For users sake, I have made the clients (http, and cli) not consume the `policyID` and `resource` name but instead a `docID` and `collection name`, since the collection will have the policy and resource information available we can fetch that and make lives easier for the users.

This PR also makes use of the `manages` feature we have had in our policy. The manages essentially defines who can make the relationship manipulation requests.

There are a lot of tests in this PR due to a lot of edge cases I wanted to have tested specific to `manger`, and ensuring `write` and `read` permissions don't leak (i.e. are accidently granted).

## CLI Demo
The following lets the target actor be able to now read the private document: 

```bash
defradb client acp relationship add \
--collection Users \
--docID bae-ff3ceb1c-b5c0-5e86-a024-dd1b16a4261c \
--relation reader \
--actor did:key:z7r8os2G88XXBNBTLj3kFR5rzUJ4VAesbX7PgsA68ak9B5RYcXF5EZEmjRzzinZndPSSwujXb4XKHG6vmKEFG6ZfsfcQn \
--identity e3b722906ee4e56368f581cd8b18ab0f48af1ea53e635e3f7b8acd076676f6ac
```

Result:
```json
{
  "ExistedAlready": false // <-------------- Indicates a new relationship was formed
}
```

### Future (out-of-scope of this PR):
- Most of write tests will split into `delete` and `update` in #2905 
- Ability to revoke or delete relation coming in #2906 
- Decide on the `can't write if no read permission` in #2992 
- Move acp logic to a shared repo: https://github.com/sourcenetwork/defradb/issues/2980

## How has this been tested?
- Integration tests

- Manjaro WSL2


tests/integration/acp.go
```diff
@@ -120,6 +120,119 @@ func addPolicyACP(
...
+		} else {
+			collection := s.collections[i][action.CollectionID]
+			if !collection.Description().Name.HasValue() {
+				require.Fail(s.t, "Expected non-empty collection name, but it was empty.", s.testCase.Description)
+			}
+			collectionName = collection.Description().Name.Value()
+		}
+
+		var docID string
+		if action.DocID == -1 {
```
shahzadlone, line 184: discussion: I understand if people prefer having `DocIDIndex` and `DocID` to test these edge cases rather than `-1` to test empty/invalid DocID input. I don't have any strong preferences here, happy to follow consensus.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_only_write_test.go
```diff
@@ -0,0 +1,359 @@
...
+                    description: A Policy
+
+                    actor:
+                      name: actor
+
+                    resources:
+                      users:
+                        permissions:
+                          read:
+                            expr: owner + reader
```
shahzadlone, line 51: discussion: Since in these tests writer won't get `read` `permission`, even if a relation is made to make an actor a writer, they won't be able to write because the policy doesn't model ability to read. This behavior we can try manipulate on defradb level to change (however not in this PR), but opening this here to start a discussion.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_gql_test.go
```diff
@@ -0,0 +1,604 @@
...
+			testUtils.AddDocActorRelationship{ // Admin tries to make another actor a writer
+				RequestorIdentity: 2,
+
+				TargetIdentity: 3,
+
+				CollectionID: 0,
+
+				DocID: 0,
+
+				Relation: "writer",
```
islamaliev, line 552: question: I don't see how this triggers the error. The title of the test says "CanNotGivesAccessNotDefinedInPolicy", but "writer" IS defined in the policy.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
```diff
@@ -0,0 +1,1286 @@
...
+
+				ExpectedError: "document not found or not authorized to access",
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestACP_ManagerGivesReadAccessToAnotherActor_OtherActorCanWrite(t *testing.T) {
```
islamaliev, line 209: type: ManagerGivesWriteAccess

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
```diff
@@ -0,0 +1,1286 @@
...
+					"Users": []map[string]any{},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestACP_OwnerMakesAManagerThatGivesItSelfReadAccess_ManagerCanRead(t *testing.T) {
```
islamaliev, line 410: question: why not to make manager automatically have this permissions? It seems to be pointless to do it separately.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
```diff
@@ -0,0 +1,1286 @@
...
+					"Users": []map[string]any{},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestACP_ManagerCanNotGivesAccessNotDefinedInPolicyToAnotherActor_Error(t *testing.T) {
```
islamaliev, line 845: suggestion: I think it should be "TestACP_ManagerGivesAccessNotDefinedInPolicyToAnotherActor_Error"

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
```diff
@@ -0,0 +1,1286 @@
...
+				CollectionID: 0,
+
+				Identity: immutable.Some(2), // Manager can not delete.
+
+				DocID: 0,
+
+				ExpectedError: "document not found or not authorized to access",
+			},
+
+			testUtils.AddDocActorRelationship{ // Manager can manger only.
```
islamaliev, line 1162: typo: manger?

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_reader_test.go
```diff
@@ -0,0 +1,808 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestACP_OwnerGivesReadAccessToAnotherActor_OtherActorCanReadSoCanTheOwner(t *testing.T) {
```
islamaliev, line 306: question: why does this need to be tested?

islamaliev (commented):
Good work so far. I'm just not sure why it's still in Draft state.
There are a lot of tests here, but I'm not sure if all them are necessary.
A lot of repetitive code in tests, which makes it hard to read every detail.

tests/integration/acp.go
shahzadlone: `immutable.Option` is for optional fields. This is not an optional field, it's more to test invalid input of that field. Is why I was wondering if others prefer non-index input as well (i.e. `DocIDIndex` for normal cases and `DocID` for actual invalid testing).

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_delete_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+			testUtils.DeleteDoc{
+				CollectionID: 0,
+
+				Identity: immutable.Some(2), // This identity can not delete yet.
+
+				DocID: 0,
+
+				ExpectedError: "document not found or not authorized to access",
+			},
```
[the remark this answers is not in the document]
shahzadlone: - [ ] Revisit and remove some earlier steps that are tested already in other tests

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_gql_test.go
shahzadlone: That is because a manager can not give access to another actor with a relation that it does not manage.

Good shout on the test name, will change to something better.

- [ ] Better test name

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
shahzadlone: This is a sourcehub level decision not a defradb one. I am guessing it was for more fine grain metrics or flexibility in future.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_reader_test.go
shahzadlone: There is an operation on sourcehub called `transfer ownership` which does exactly that. So here I wanted to make sure no bugs slip through in the future. i.e. if the `sharing` ever `transfers` the document ownership it should definitely be caught.

tests/integration/acp.go
```diff
@@ -120,6 +120,119 @@ func addPolicyACP(
 	}
 }
 
+// AddDocActorRelationship will attempt to add the given policy using DefraDB's ACP system.
```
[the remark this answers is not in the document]
shahzadlone: - [x] done

tests/integration/acp.go
```diff
@@ -120,6 +120,119 @@ func addPolicyACP(
...
+type AddDocActorRelationship struct {
+	// NodeID may hold the ID (index) of the node we want to add doc actor relationship on.
+	//
+	// If a value is not provided the relationship will be added in all nodes, unless testing with
+	// sourcehub ACP, in which case the relationship will only be defined once.
+	NodeID immutable.Option[int]
+
+	// The collection in which this document we want to add a relationship for exists.
+	//
+	// This is a required field, to test the invalid usage of not having this arg, use -1 index.
```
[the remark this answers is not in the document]
shahzadlone: - [x] Done for this and other fields

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
shahzadlone: - [x] Fix typo

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
shahzadlone: - [x] Done

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
islamaliev: the questions is then why are we testing sourcehub functionality so vigorously on defra side? After all, this is just one of potential remote ACP implementations. What if other ACPs don't have such behavior? Or what if it changes on the sourcehub side?

shahzadlone: > the questions is then why are we testing sourcehub functionality so vigorously on defra side?
I don't consider this as `vigorous` at all. It's just that the manager functionality existed before since acp was implemented but it doesn't become functional or even testable until we have relationship feature. So now that we have the relationship sharing implemented this is important to cover in my opinion as it's something users can now use.

> After all, this is just one of potential remote ACP implementations.
Not sure I understand, but every test runs against both `local` and `sourcehub` implementation. 

> What if other ACPs don't have such behavior?
What is this other ACP you are referring to? Not sure I follow.

> Or what if it changes on the sourcehub side?
Then that is a change that DefraDB must adjust with like with all other third party dependency changes (however I doubt this behavior will change).

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_reader_test.go
shahzadlone: I can add a comment here sure, but I don't see how this makes hard for other non-sourcehub implementations. They should all have the same behavior (in this case making sure that after new actor can access a document the previous owner doesn't loose access, i.e. is still the owner). I understand is an edge case, but this will be true for all acp implementations I can think of

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
shahzadlone: - [x] Changed to a better name: `TestACP_ManagerAddsRelationshipWithRelationItDoesNotManageAccordingToPolicy_Error`

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_delete_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+			testUtils.DeleteDoc{
+				CollectionID: 0,
+
+				Identity: immutable.Some(2), // This identity can not delete yet.
+
+				DocID: 0,
+
+				ExpectedError: "document not found or not authorized to access",
```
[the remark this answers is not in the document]
shahzadlone: While I like the suggestion, this seems like a personal preference unless we reach consensus as a team to follow this rule. So leaving as is.

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_reader_test.go
shahzadlone: - [x] Added comment

tests/integration/acp/p2p/replicator_with_doc_actor_relationship_test.go
```diff
@@ -0,0 +1,219 @@
...
+				`,
+
+				Results: map[string]any{
+					"Users": []map[string]any{},
+				},
+			},
+
+			testUtils.AddDocActorRelationship{
+				NodeID: immutable.Some(0),
+
```
AndrewSisley, line 144: question: I've seen you do this quite a bit, why do you add whitespace between every property in an instance declaration?

shahzadlone: No good reason, I am claustrophobic perhaps and like spaced lines lol. I often wonder why people don't add whitespaces.

acp/source_hub_client.go
```diff
@@ -335,6 +356,83 @@ func (a *sourceHubBridge) CheckDocAccess(
...
+			policyID,
+			resourceName,
+			docID,
+			relation,
+			requestActor.DID,
+			targetActor,
+		)
+	}
+
+	if exists {
```
AndrewSisley, line 409: todo: Duplicating the log between the if-else here really reduces the information density, and increases the amount of stuff maintainers have to read. Please simplify this to something similar to the below:
```
log.InfoContext(
			ctx,
			"Document and actor relationship set",
			corelog.Any("PolicyID", policyID),
			corelog.Any("ResourceName", resourceName),
			corelog.Any("DocID", docID),
			corelog.Any("Relation", relation),
			corelog.Any("RequestActor", requestActor.DID),
			corelog.Any("TargetActor", targetActor),
			corelog.Any("Existed", exists),
		)
		return exists, nil
```

shahzadlone: Good suggestion, will do.

- [ ] logs

tests/integration/acp/p2p/replicator_with_doc_actor_relationship_test.go
shahzadlone: TBH it's just a habit while writing code, I don't care if some formatter formats these style choices to one format that is more consistent (idk if there is one, or a linter rule?)

tests/integration/acp/relationship/add_doc_actor_test/add_doc_actor_with_manager_test.go
AndrewSisley: >  why are we testing sourcehub functionality so vigorously on defra side?

I think it is worth having a proper team conversation about this - the ACP interface is kind of a black-box to Defra, and it should be up to the implementations to ensure they properly satisfy the interface.  Defra just needs to worry about calling correctly.

The maintenance cost for all these tests is not free.

shahzadlone: > > why are we testing sourcehub functionality so vigorously on defra side?
> 
> I think it is worth having a proper team conversation about this - the ACP interface is kind of a black-box to Defra, and it should be up to the implementations to ensure they properly satisfy the interface. Defra just needs to worry about calling it correctly.
> 
> The maintenance cost for all these tests is not free.

I agree, I do kind of look at this situation as along the same line as `Unit test` vs `Integration test` specially because the "sourcehub unit" at this current stage lacks the level of testing I would feel comfortable with specially when using it to deliver some key privacy/security feature of a database.

AndrewSisley: > the "sourcehub unit" at this current stage lacks the level of testing

Yes, this is a problem, but I very strongly believe that any Source member should be free to correct that instead of over-compensating in what could be seen as the wrong repo.

shahzadlone: > Source member should be free to correct that instead of feeling like they need to over-compensate in what could be seen as the wrong repo.

This is actually a very good idea, this way the testing can belong in more appropriate areas. For example the DPI stuff `write` and `read` specific permission testing can happy on defra and the general invalid and edge cases can belong on sourcehub.

islamaliev: I know it feels more reassuring to have many tests covered on Defra side, but as Andy pointed out, it comes at a cost. Defra should test only defra-related edge cases and some general happy-paths to make sure modules work well together.

islamaliev: > What is this other ACP you are referring to? Not sure I follow.

We are thinking for example about implementing UCAN-based remote ACP which relies purely on decentralized permission resolution between peers, i.e. without some entity (like sourcehub) in the middle.

shahzadlone: I am happy to follow consensus, can point to individual tests that are seemingly out of defra-scope and I will push back only if I do deem it "defra" specific, otherwise will try to move them to sourcehub :)

shahzadlone: Tracking here: https://github.com/sourcenetwork/defradb/issues/2980

islamaliev (approved):
Approving assuming tests will be moved as discussed

internal/db/db.go
```diff
@@ -196,6 +198,50 @@ func (db *db) AddPolicy(
...
+	docID string,
+	relation string,
+	targetActor string,
+) (bool, error) {
+	if !db.acp.HasValue() {
+		return false, client.ErrACPOperationButACPNotAvailable
+	}
+
+	if collectionName == "" {
+		return false, client.ErrOperationNotPermittedOnNamelessCols
```
AndrewSisley, line 213: todo: This is already handled by `GetCollectionByName` which will error if no collection is found.  Please remove the new error and this block.

client/db.go
```diff
@@ -99,6 +99,23 @@ type DB interface {
 	//
 	// Note: A policy can not be added without the creatorID (identity).
 	AddPolicy(ctx context.Context, policy string) (AddPolicyResult, error)
+
+	// AddDocActorRelationship creates a relationship between document and the target actor.
+	//
+	// Moreover, since collection has the policy and resource information we don't need to
+	// explicitly have the user provide them as arguments.
```
AndrewSisley, line 106: > // Moreover, since collection has the policy and resource information we don't need to
	// explicitly have the user provide them as arguments

todo: I'm not sure I understand this, can you please reword it?  Are some of these params optional?

acp/source_hub_client.go
```diff
@@ -335,6 +356,71 @@ func (a *sourceHubBridge) CheckDocAccess(
...
+func (a *sourceHubBridge) AddDocActorRelationship(
+	ctx context.Context,
+	policyID string,
+	resourceName string,
+	docID string,
+	relation string,
+	requestActor identity.Identity,
+	targetActor string,
+) (bool, error) {
+	// Must have all required args.
```
AndrewSisley, line 368: suggestion: This comment is not very useful, please either expand it to cover *why* all of the are required (and move it to the func-level), or remove it - the code already states that they are required, without a why this comment is not useful.

shahzadlone: - [ ] remove redundant comment

internal/db/db.go
shahzadlone: good catch

- [x] removed

client/db.go
shahzadlone: params are not optional was trying to explain that the magic of having policy and resource info on collection is nice because the user doesn't have to input that but just the collection name. going to remove it as it causes more confusion that clarification

- [x] extra fluff makes things more confusing to remove

AndrewSisley (approved):
LGTM, thanks Shahzad!

PR #2931, opened 2024-08-20
chore: Change from ipld traversal to direct link access

### LINKED ISSUE #2930: Switch from ipld schema traversal to direct link access.
We currently use the ipld schema traversal approach to extract links from blocks to then sync them across the network. The schema traversal approach is great for network exchange of non-deterministic data structures. However, in our case, the data structure is known in advance and we would benefit from the efficiency of direct link access. Also, it will enable using separate  blockservices for data blocks and encrypted blocks.

fredcarle:
## Relevant issue(s)

Resolves #2930 

## Description

This PR removes the use of the ipld traversal function in favor of direct link access. This makes the sync process a bit more efficient and will make it easier to initiate encryption key exchange.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/net/sync_dag.go
+++ b/net/sync_dag.go
@@ -12,20 +12,13 @@ package net
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/ipfs/boxo/blockservice"
-	"github.com/ipld/go-ipld-prime"
-	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/linking"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
-	"github.com/ipld/go-ipld-prime/linking/preload"
-	basicnode "github.com/ipld/go-ipld-prime/node/basic"
-	"github.com/ipld/go-ipld-prime/schema"
 	"github.com/ipld/go-ipld-prime/storage/bsrvadapter"
-	"github.com/ipld/go-ipld-prime/traversal"
-	"github.com/ipld/go-ipld-prime/traversal/selector"
-	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
 
 	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
 )
@@ -40,9 +33,6 @@ var syncDAGTimeout = 60 * time.Second
 // This process walks the entire DAG until the issue below is resolved.
 // https://github.com/sourcenetwork/defradb/issues/2722
 func syncDAG(ctx context.Context, bserv blockservice.BlockService, block *coreblock.Block) error {
-	ctx, cancel := context.WithTimeout(ctx, syncDAGTimeout)
-	defer cancel()
-
 	// use a session to make remote fetches more efficient
 	ctx = blockservice.ContextWithSession(ctx, bserv)
 	store := &bsrvadapter.Adapter{Wrapped: bserv}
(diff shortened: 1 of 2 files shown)


AndrewSisley (approved):
LGTM :)

net/sync_dag.go
```diff
@@ -58,37 +48,56 @@ func syncDAG(ctx context.Context, bserv blockservice.BlockService, block *corebl
...
-		return nil
+
+	for _, lnk := range block.Links {
+		wg.Add(1)
+		go func(lnk coreblock.DAGLink) {
+			defer wg.Done()
+			if ctx.Err() != nil {
+				return
+			}
+			nd, err := lsys.Load(linking.LinkContext{Ctx: ctx}, lnk, coreblock.SchemaPrototype)
```
islamaliev, line 82: question: with this change doesn't it mean we can read directly from `datastore.Blockstore` in `mergeProcessor.loadComposites` instead of pulling it from block service?

net/sync_dag.go
```diff
@@ -58,37 +48,56 @@ func syncDAG(ctx context.Context, bserv blockservice.BlockService, block *corebl
...
+			if err != nil {
+				asyncErrOnce.Do(func() { setAsyncErr(err) })
+				return
+			}
+			linkBlock, err := coreblock.GetFromNode(nd)
+			if err != nil {
+				asyncErrOnce.Do(func() { setAsyncErr(err) })
+				return
+			}
+			err = loadBlockLinks(ctx, lsys, linkBlock)
```
islamaliev, line 92: questions: are we supposed to traverse the whole tree every time an update happens? Shouldn't we stop once we read an existing block?

islamaliev (approved):
Looks great! Now it's straightforward to comprehend. I don't understand why we used that convoluted selector/builder/walker approach.

net/sync_dag.go
fredcarle: `loadComposites` doesn't use a block service. It already reads directly from the blockstore.

net/sync_dag.go
fredcarle: The benefit for the moment is that partial syncs can be completed at a later time. If we stop when we have a known block, then completing partial syncs won't be possible until we implement some tracking logic either for the completely synced or partially synced branches. The implementation before this PR did a complete traversal as well.

net/sync_dag.go
islamaliev: alright, I see

net/sync_dag.go
islamaliev: okay. Thanks for clarification.

net/sync_dag.go
```diff
@@ -58,37 +48,48 @@ func syncDAG(ctx context.Context, bserv blockservice.BlockService, block *corebl
...
+				})
+				cancel()
+				return
+			}
+			linkBlock, err := coreblock.GetFromNode(nd)
+			if err != nil {
+				asyncErrOnce.Do(func() {
+					asyncErr = err
+				})
+				cancel()
```
shahzadlone: question: `cancel()` is only for err cases and not to be run using `defer`?

shahzadlone (approved):
LGTM, just a question

net/sync_dag.go
fredcarle: It is used with `defer` but also explicitly called on error to help stop the goroutines.

shahzadlone: Ah makes sense. I missed the defer above.

PR #2944, opened 2024-08-22
ci(i): Docker build separate from goreleaser

### LINKED ISSUE #2941: Build and push docker image outide of goreleaser
Docker flags our image as having a high-severity vulnerability because gorelease using the  pre-built binary in the build process. We should make the docker build portion of the release separate from goreleaser.

### LINKED ISSUE #2943: Revert eslint upgrade
eslint 9 is not compatible with eslint-plugin-react-hooks. We need to revert it back to 8.57

fredcarle:
## Relevant issue(s)

Resolves #2941 

## Description

This PR brings our docker build and push outside the goreleaser process and instead defines it as a separate ci job.

Since I was working on the release and testing the build, I noticed problem with the JS dependencies of the playground and I added the fix here.

## Tasks

## How has this been tested?

Ran the release workflow on my fork

- MacOS


AndrewSisley (approved):
LGTM :)

playground/package.json
```diff
@@ -23,7 +23,7 @@
     "@typescript-eslint/eslint-plugin": "^8.2.0",
     "@typescript-eslint/parser": "^8.2.0",
     "@vitejs/plugin-react-swc": "^3.7.0",
-    "eslint": "^9.9.0",
+    "eslint": "^8.57.0",
```
shahzadlone, line 26: Question: why did you have to revert back?

fredcarle: As stated in the PR description, another dependency doesn't support version 9 of eslint and the build would fail as a result.

fredcarle: Sorry my brain thought I had written about the eslint issue in the PR description but I did that in the linked issue.

shahzadlone: I had read the linked issue as well. No worries.

But the added flag to `npm i` fixes the dep issue no?

Where in the build does it fail with the new eslint version?

shahzadlone: You are referring to the react hooks plugin, right?

fredcarle: Discussed on Discord. Added `--legacy-peer-dep` to `make deps:playground` keeping `eslint` at version 9.9.

.github/workflows/release.yml
```diff
@@ -179,17 +154,58 @@ jobs:
...
+
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: tools/defradb.containerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: |
+            ${{ github.actor }}/defradb:latest
```
shahzadlone, line 193: question: To confirm the github.actor is `sourcenetwork`? not the `source-dev` account?

fredcarle: good catch. It worked when I tested on my fork because the repo account is my user.

fredcarle: changed to `repository_owner`

shahzadlone (approved):
LGTM, thanks for:

- Sorting out the `make deps:playground`
- Removing the other container file
- Lower casing the build targets

.github/workflows/release.yml
shahzadlone: cool

.github/workflows/release.yml
```diff
@@ -179,17 +154,58 @@ jobs:
           GITHUB_REPOSITORY: ${{ github.repository }}
           GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
 
+  docker-build-push:
```
shahzadlone, line 157: question: This job doesn't need the release or prepare at all? It just pushes?

fredcarle: The docker container has everything it needs to run without being dependent on release or prepare 👍

shahzadlone: If the release or prepare fail then it will still push docker container?

fredcarle: If the docker image can build, then that's fine with me. There will always be the possibility of one or the other failing regardless of the order or lack thereof.

PR #2947, opened 2024-08-22
fix: Panic with different composite-indexed child objects

### LINKED ISSUE #2924: Fetching indexed field in child document panics
The following setup panics.  It does not panic when `reward_id` is removed from the query.  It does not panic when the index is removed from `ActionRequirement`.

Schema:
```
type ActionRequirement @index(unique: true, fields: ["action_id", "reward_id"], name: "actionId_rewardId")  {
  count: Int
  isPreReq: Boolean # @default(false)
  action: Action @relation(name: "actionActionRequirements")
  reward: Reward @relation(name: "rewardActionRequirements")
  createdAt: DateTime # @default(now())
  oldId: String
}

type Reward {
  name: String @index(unique: true) # @unique
  description: String
  value: Int
  limit: Int
  tiers: [RewardTier] @relation(name: "rewardRewardTiers")
  priority: Int
  enabled: Boolean # @default(true)
  actionRequirements: [ActionRequirement] @relation(name: "rewardActionRequirements")
  playerRewards: [PlayerReward] @relation(name: "rewardPlayerRewards")
  rewardGroup: [String] # [ENUM] RewardGroup
  createdAt: DateTime # @default(now())
  oldId: String
}
```
Query:
```
query {
  Reward(filter: {name: {_eq: "FOLLOW_TWITTER"}}) {
    _deleted
    _docID
    id: _docID
    createdAt
    description
    enabled
    limit
    name
    priority
    rewardGroup
    value
    oldId
    actionRequirements {
      id: _docID
      _docID
      action_id
      count
      createdAt
      isPreReq
      oldId
      reward_id
    }
  }}
```
Panic (was reported via http client, so original panic is lost):
> panic: Unclosed iterator at time of Txn.Discard.
 github.com/sourcenetwork/badger/v4.(Txn).Discard
  github.com/sourcenetwork/badger/v4@v4.2.1-0.20231113215945-a63444ca5276/txn.go:524
    github.com/sourcenetwork/badger/v4.(Txn).Commit
      github.com/sourcenetwork/badger/v4@v4.2.1-0.20231113215945-a63444ca5276/txn.go:665
    github.com/sourcenetwork/defradb/datastore/badger/v4.(txn).commit
      github.com/sourcenetwork/defradb/datastore/badger/v4/datastore.go:765
    github.com/sourcenetwork/defradb/datastore/badger/v4.(txn).Commit
      github.com/sourcenetwork/defradb/datastore/badger/v4/datastore.go:761
    github.com/sourcenetwork/defradb/datastore.(txn).Commit
      github.com/sourcenetwork/defradb/datastore/txn.go:106
    github.com/sourcenetwork/defradb/internal/db.(db).ExecRequest
      github.com/sourcenetwork/defradb/internal/db/store.go:38
    github.com/sourcenetwork/defradb/http.(*storeHandler).ExecRequest
      github.com/sourcenetwork/defradb/http/handler_store.go:302
    net/http.HandlerFunc.ServeHTTP
      net/http/server.go:2171

islamaliev:
## Relevant issue(s)

Resolves #2924

## Description

Make multiScanNode keep track of all calls.

The issue was that multiScanNode that keeps track of how many calls to Next() should be made, would call Init() and Start() every time without tracking, which would result in child nodes being initialized and started multiple times, which in turn created index iterators without closing them.

### DIFF
--- a/internal/db/fetcher/indexer_iterators.go
+++ b/internal/db/fetcher/indexer_iterators.go
@@ -81,6 +81,11 @@ var _ indexIterator = (*indexPrefixIterator)(nil)
 func (iter *indexPrefixIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
 	iter.ctx = ctx
 	iter.store = store
+	if iter.resultIter != nil {
+		if err := iter.resultIter.Close(); err != nil {
+			return err
+		}
+	}
 	iter.resultIter = nil
 	return nil
 }
--- a/internal/db/fetcher/indexer_iterators.go
+++ b/internal/db/fetcher/indexer_iterators.go
@@ -64,13 +64,41 @@ type indexIterResult struct {
 	value    []byte
 }
 
-type queryResultIterator struct {
-	resultIter    query.Results
+// indexPrefixIterator is an iterator over index keys with a specific prefix.
+type indexPrefixIterator struct {
 	indexDesc     client.IndexDescription
 	indexedFields []client.FieldDefinition
+	indexKey      core.IndexDataStoreKey
+	matchers      []valueMatcher
+	execInfo      *ExecInfo
+	resultIter    query.Results
+	ctx           context.Context
+	store         datastore.DSReaderWriter
 }
 
-func (iter *queryResultIterator) Next() (indexIterResult, error) {
+var _ indexIterator = (*indexPrefixIterator)(nil)
+
+func (iter *indexPrefixIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
+	iter.ctx = ctx
+	iter.store = store
+	iter.resultIter = nil
+	return nil
+}
+
+func (iter *indexPrefixIterator) checkResultIterator() error {
+	if iter.resultIter == nil {
+		resultIter, err := iter.store.Query(iter.ctx, query.Query{
+			Prefix: iter.indexKey.ToString(),
+		})
+		if err != nil {
+			return err
+		}
+		iter.resultIter = resultIter
+	}
+	return nil
+}
+
+func (iter *indexPrefixIterator) nextResult() (indexIterResult, error) {
 	res, hasVal := iter.resultIter.NextSync()
 	if res.Error != nil {
 		return indexIterResult{}, res.Error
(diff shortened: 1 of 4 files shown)


internal/planner/scan.go
```diff
@@ -325,35 +325,53 @@ func (p *Planner) Scan(
...
-	lastBool bool
-	lastErr  error
+	nextCount  int
+	initCount  int
+	startCount int
+	closeCount int
+
+	nextResult bool
+	nextErr    error
+	initErr    error
```
AndrewSisley, line 335: question: Why are you storing these as state?  And why are they stored separately (`nextErr`, `initErr`, etc instead of just `err`)?

internal/planner/scan.go
```diff
@@ -325,35 +325,53 @@ func (p *Planner) Scan(
...
+	startErr   error
+	closeErr   error
 }
 
 func (n *multiScanNode) Init() error {
-	return n.scanNode.Init()
+	n.countAndCall(&n.initCount, func() {
```
AndrewSisley, line 341: todo: Please document why this is done - it is not obvious within this PR, and it was already missed previously.  And please do the same for the other funcs changed here too.

AndrewSisley (changes requested):
I trust it fixes it, but it is not clear why - please add some documentation :)

AndrewSisley:
> which in turn created index iterators without closing them.

To me this makes it sound like the previous index iterator should be closed on `Init` of whatever created it.  `Init` is designed to be called multiple times and it should reset the node state on each call.  I think that the changes in this PR actually breaks this in multinode, in order to manage the symptoms of the index iterator not disposing of it's prior state correctly.

internal/planner/scan.go
islamaliev: left only 1 err

internal/planner/scan.go
islamaliev: documented

internal/db/fetcher/indexer_iterators.go
```diff
@@ -64,13 +64,41 @@ type indexIterResult struct {
...
+	ctx           context.Context
+	store         datastore.DSReaderWriter
+}
+
+var _ indexIterator = (*indexPrefixIterator)(nil)
+
+func (iter *indexPrefixIterator) Init(ctx context.Context, store datastore.DSReaderWriter) error {
+	iter.ctx = ctx
+	iter.store = store
+	iter.resultIter = nil
```
AndrewSisley, line 84: suggestion: If `iter.resultIter` is not nil here, it looks like it should probably be closed before clearing the ref?

I have a vague memory of the normal fetcher failing because of this some time ago, but I might be wrong.

internal/planner/multi.go
```diff
@@ -214,22 +214,22 @@ func (s *selectNode) addSubPlan(fieldIndex int, newPlan planNode) error {
 		if err := s.planner.walkAndReplacePlan(s.source, origScan, multiscan); err != nil {
 			return err
 		}
-		// create multinode
-		multinode := &parallelNode{
+		// create parallelNode
+		parallelNode := &parallelNode{
```
AndrewSisley, line 218: praise: I forgot to add to my earlier review, thanks for renaming this.

internal/planner/scan.go
```diff
@@ -322,38 +322,70 @@ func (p *Planner) Scan(
 // If we have two readers on our multiScanNode, then
 // we call Next() on the underlying scanNode only
 // once every 2 Next() calls on the multiScan
+//
+// NOTE: calling Init() on multiScanNode is subject to counting as well and as such
+// doesn't not provide idempotency guarantees. Counting is purely for performance
+// reasons and removing it should be safe.
```
AndrewSisley, line 328: praise: This is very clear, thanks Islam :)

tests/integration/index/query_with_composite_inxed_on_relation_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+// This test was added during https://github.com/sourcenetwork/defradb/issues/2924
+// The issue was that [multiScanNode] that keeps track of how many calls to [Next()] should
+// be made, would call [Init()] and [Start()] every time without tracking, which would result
+// in child nodes being initialized and started multiple times, which in turn created
+// index iterators without closing them.
```
AndrewSisley, line 23: praise: Thanks for taking the time to link this to the ticket, it highlights the importance of the test very nicely.

AndrewSisley (approved):
Looks good to me, thanks Islam :)

Please have a quick look at the final suggestion before merge.

PR #2963, opened 2024-08-26
refactor: Make SourceHub dep internal-only

### LINKED ISSUE #2962: Remove SourceHub sdk dep from test utils
Mostly so the vuln checker goes green.  Is quick to do, but the dep will probably come back medium term via a future sourcehub testing lib.

AndrewSisley:
## Relevant issue(s)

Resolves #2962

## Description

I was mostly curious if this will make the vuln checker go green (it doesn't, it just reports it from our internal acp package now), but it is a good change anyway IMO to avoid embedded Go client users from having to directly import the dependency.

### DIFF
--- a/cli/start.go
+++ b/cli/start.go
@@ -16,7 +16,6 @@ import (
 	"syscall"
 
 	"github.com/sourcenetwork/immutable"
-	"github.com/sourcenetwork/sourcehub/sdk"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


shahzadlone (approved):
Change LGTM even if doesn't fix vul checker. I wish a silencing feature was there https://github.com/golang/go/issues/61211

PR #2967, opened 2024-08-28
fix: Ignore badger path if in-memory

### LINKED ISSUE #2964: `defradb start ... --store memory` errors
Selecting the in memory store using the `--store memory` param on a clean/nil db errors with the below.

> Error: Cannot use badger in Disk-less mode with Dir or ValueDir set

It seems strange that our integration tests did not catch this, please find out why and try and fix that too within this ticket.

Full commands tested:
```
defradb start --no-keyring --no-p2p --url :9181 --allowed-origins "*" --store memory
defradb start --store memory
```

Clearing the value from `datastore: badger: path:` in the generated config file allows the command to succeed.

ONLYUSEmePHONE:
## Relevant issue(s)

Resolves #2964 

## Description

Small change that now only sets the badger options `Dir` and `ValueDir` to the  `path`  parameter if the badger option `InMemory` is not `true`.

The crash was happening because badger expects the two Dir values to be nil if InMemory is true.

I chose this fix because it's very simple and requires the least changes. The bulk of the problem is that the `datastore.badger.path` config value gets set to `data` by default regardless of whether  `datastore.store` is set to `memory` and this value gets passed as the `path`  parameter. I didn't think it would be nice to change any of the upstream logic or the signature of the NewDatastore function.

## Tasks

## How has this been tested?

Running standard test suite and using both commands provided in #2964 .

- Debian Linux

### DIFF
--- a/datastore/badger/v4/datastore.go
+++ b/datastore/badger/v4/datastore.go
@@ -142,8 +142,11 @@ func NewDatastore(path string, options *Options) (*Datastore, error) {
 		gcSleep = gcInterval
 	}
 
-	opt.Dir = path
-	opt.ValueDir = path
+	if !opt.InMemory {
+	    opt.Dir = path
+	    opt.ValueDir = path
+    }
+
 	opt.Logger = &compatLogger{
 		SugaredLogger: *log.Desugar().WithOptions(zap.AddCallerSkip(1)).Sugar(),
 		skipLogger:    *log.Desugar().WithOptions(zap.AddCallerSkip(2)).Sugar(),
--- a/datastore/badger/v4/datastore.go
+++ b/datastore/badger/v4/datastore.go
@@ -143,9 +143,9 @@ func NewDatastore(path string, options *Options) (*Datastore, error) {
 	}
 
 	if !opt.InMemory {
-	    opt.Dir = path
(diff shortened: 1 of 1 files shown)


AndrewSisley (approved):
@ONLYUSEmePHONE Hey, and welcome to our codebase!

Thank you very much for this fix, it solves the bigger problem, and is very minimal as you noted in the description :)

AndrewSisley:
FYI: I'm going to update the PR title, hope you don't mind - we have a 60 character limit (I also rebased onto the latest, also a requirement for merge)

datastore/badger/v4/datastore.go
```diff
@@ -142,8 +142,11 @@ func NewDatastore(path string, options *Options) (*Datastore, error) {
 		gcSleep = gcInterval
 	}
 
-	opt.Dir = path
-	opt.ValueDir = path
+	if !opt.InMemory {
+	    opt.Dir = path
```
AndrewSisley, line 146: todo: The linter is complaining, looks like you have tried to sneak spaces, not tabs into our codebase - sorry about the bother, but if you fix this I'll merge your PR immediately after :) :tada:

PR #2994, opened 2024-09-10
refactor: Index field directive

### LINKED ISSUE #2926: Index directives defined twice
**Describe the problem**
The `@index` directive is defined twice, once for objects, and once for fields.  The spec references that each directive "must have a unique name", therefore defining two schemas with the same name `@index` is invalid.

I believe this is currently used because the two locations we can define a schema (object and field) have different input parameters.

**To Reproduce**
Introspect any schema (even default)

**Expected behavior**

**Additional context**
We can fixed this either A) have two different named schemas for each use case (field and object) B) merge the arguments from the two types into one.

nasdf:
## Relevant issue(s)

Resolves #2926

## Description

This PR fixes an issue where the index directive was defined twice. ~The `@index` field level directive has been renamed to `@indexField`.~

The `fields` arg has been renamed to `includes` and is now a list of objects of type:

```
type IndexFieldInput {
  name: String
  direction: Ordering
}
```

The `direction` argument now sets the default direction of all fields in the `includes` list.

When the index is used on a field definition and the field is not in the `includes` list it will be implicitly added.

## Tasks

## How has this been tested?

Covered by existing tests.

- MacOS

### DIFF
--- a/internal/request/graphql/schema/types/types.go
+++ b/internal/request/graphql/schema/types/types.go
@@ -33,16 +33,14 @@ const (
 	PolicySchemaDirectivePropID       = "id"
 	PolicySchemaDirectivePropResource = "resource"
 
-	IndexDirectiveLabel          = "index"
-	IndexDirectivePropName       = "name"
-	IndexDirectivePropUnique     = "unique"
-	IndexDirectivePropFields     = "fields"
-	IndexDirectivePropDirections = "directions"
+	IndexDirectiveLabel         = "index"
+	IndexDirectivePropName      = "name"
+	IndexDirectivePropUnique    = "unique"
+	IndexDirectivePropDirection = "direction"
+	IndexDirectivePropIncludes  = "includes"
 
-	IndexFieldDirectiveLabel         = "indexField"
-	IndexFieldDirectivePropName      = "name"
-	IndexFieldDirectivePropUnique    = "unique"
-	IndexFieldDirectivePropDirection = "direction"
+	IndexFieldInputName      = "name"
+	IndexFieldInputDirection = "direction"
 
 	FieldOrderASC  = "ASC"
 	FieldOrderDESC = "DESC"
@@ -125,10 +123,24 @@ func PolicyDirective() *gql.Directive {
 	})
 }
 
-func IndexDirective(orderingEnum *gql.Enum) *gql.Directive {
+func IndexFieldInputObject(orderingEnum *gql.Enum) *gql.InputObject {
+	return gql.NewInputObject(gql.InputObjectConfig{
+		Name: "IndexField",
+		Fields: gql.InputObjectConfigFieldMap{
+			IndexFieldInputName: &gql.InputObjectFieldConfig{
+				Type: gql.String,
+			},
+			IndexFieldInputDirection: &gql.InputObjectFieldConfig{
+				Type: orderingEnum,
+			},
+		},
+	})
+}
+
+func IndexDirective(orderingEnum *gql.Enum, indexFieldInputObject *gql.InputObject) *gql.Directive {
 	return gql.NewDirective(gql.DirectiveConfig{
 		Name:        IndexDirectiveLabel,
-		Description: "@index is a directive that can be used to create an index on a type.",
+		Description: "@index is a directive that can be used to create an index on a type or a field.",
 		Args: gql.FieldConfigArgument{
 			IndexDirectivePropName: &gql.ArgumentConfig{
 				Type: gql.String,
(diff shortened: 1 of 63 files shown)


AndrewSisley (changes requested):
question: Why can you not just use the same directive in both places?  Do they have different arguments?

EDIT: They look like the have the same args, apart from `fields` which is missing at the field level - can we please add support for field selection indexes defined at the field level and just have one directive?

nasdf:
> question: Why can you not just use the same directive in both places? Do they have different arguments?
> 
> EDIT: They look like the have the same args, apart from `fields` which is missing at the field level - can we please add support for field selection indexes defined at the field level and just have one directive?

The arguments are of different types. The `@index` directive creates an index from one or more fields whereas `@indexField` is for a singular field.

AndrewSisley:
> The arguments are of different types. The @index directive creates an index from one or more fields whereas @indexField is for a singular field.

Yes, but can we not make them the same type?  `fields` on a field level directive would implicitly include the field that it is on.  This way users only have to worry about a single, consistent directive.

nasdf:
@AndrewSisley and I have come up with some potential improvements to the index directive.

The `directions` arg will be replaced with a new arg `includes`. The `includes` arg is a list of type:

```gql
input IndexIncludeArg{
  field: String
  direction: Ordering
}
```

The `direction` arg will set the default ordering for all fields. When the directive is used on a field definition, that field will be implicitly added to the `includes` list with the default `direction` value.

Example of default direction:

`@index(name: "idx", direction: ASC, unique: true, includes: [{field: name}, {field: email, direction: DESC}])`

would be equivelent to:

`@index(name: "idx", unique: true, includes: [{field: name, direction: ASC}, {field: email, direction: DESC}])`

nasdf:
> @AndrewSisley and I have come up with some potential improvements to the index directive.
> 
> The `directions` and `fields` args will be merged into a new`includes` arg. The `includes` arg is a list of type:
> 
> ```graphql
> input IndexIncludeArg{
>   field: String
>   direction: Ordering
> }
> ```
> 
> The `direction` arg will set the default ordering for all fields. When the directive is used on a field definition, that field will be implicitly added to the `includes` list with the default `direction` value.
> 
> Example of default direction:
> 
> `@index(name: "idx", direction: ASC, unique: true, includes: [{field: name}, {field: email, direction: DESC}])`
> 
> would be equivelent to:
> 
> `@index(name: "idx", unique: true, includes: [{field: name, direction: ASC}, {field: email, direction: DESC}])`

Collective decision was to go forward with this design but also allow defining the field orderings in `include` explicitly as well as implicitly.

internal/request/graphql/schema/types/types.go
```diff
@@ -121,31 +123,24 @@ func PolicyDirective() *gql.Directive {
...
-func IndexDirective(orderingEnum *gql.Enum) *gql.Directive {
-	return gql.NewDirective(gql.DirectiveConfig{
-		Name:        IndexDirectiveLabel,
-		Description: "@index is a directive that can be used to create an index on a type.",
-		Args: gql.FieldConfigArgument{
-			IndexDirectivePropName: &gql.ArgumentConfig{
+func IndexFieldInputObject(orderingEnum *gql.Enum) *gql.InputObject {
+	return gql.NewInputObject(gql.InputObjectConfig{
+		Name: "IndexField",
+		Fields: gql.InputObjectConfigFieldMap{
```
AndrewSisley, line 129: todo: Please add a Description, and note the implicit/explicit behaviour on it when attached to a field.

These descriptions are important and get rendered in the docs section and intellisense of GQL clients, and I guess that ORMs will also use them to document bound types.

tests/integration/index/create_get_test.go
```diff
@@ -23,7 +23,7 @@ func TestIndexGet_ShouldReturnListOfExistingIndexes(t *testing.T) {
 		Actions: []any{
```
AndrewSisley, line 23: todo: (not necessarily in this file, this is just the first test file) Please add some integration tests testing implicit field level directives, the 'default' behaviour of the `direction` property, the overriding of a default `direction` by explicitly stating it in a field in `includes`, the overriding of the field location by explicitly stating a field when the directive is hosted on a field, and composite indexes declared on a field.

AndrewSisley (changes requested):
Production code looks good, but please test the new behaviour before merging.

tests/integration/index/create_get_test.go
nasdf: done

internal/request/graphql/schema/types/types.go
nasdf: done

fredcarle (approved):
LGTM! Thanks for taking care of this. I like the improvement to defining composite indexes and their components direction.

AndrewSisley (approved):
LGTM, thanks Keenan!

PR #2998, opened 2024-09-12
feat: CLI purge command

### LINKED ISSUE #2953: Add a CLI command to reset the database
While doing development, it is often needed to reset the database. Instead of having to delete the data directory, we could add a command that cleans up the database.

nasdf:
## Relevant issue(s)

Resolves #2953

## Description

~This PR adds a purge command that clears the data directory of the defradb instance on the local machine.~

This PR adds a purge command that clears the configured datastore and restarts the defradb instance.

The purge command requires a `--force` flag to be run, and the defradb instance must have the `development` flag or config value enabled.

- [x] Add `dev` flag and config parameter that is required for destructive operations

## Tasks

## How has this been tested?

Added unit tests and tested manually

- MacOS


cli/purge.go
```diff
@@ -0,0 +1,42 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !force {
+				return ErrPurgeForceFlagRequired
+			}
+			cfg := mustGetContextConfig(cmd)
+			return os.RemoveAll(cfg.GetString("datastore.badger.path"))
+		},
+	}
+	cmd.Flags().BoolVarP(&force, "force", "f", false, "Must be set for the operation to run")
```
AndrewSisley, line 40: todo: I don't think  requiring `-f` is sufficient protection against users nuking a production database, I can't remember the details but I thought we agreed on something slightly safer when we discussed this as a team?

cli/purge.go
```diff
@@ -0,0 +1,42 @@
...
+				return err
+			}
+			return setContextConfig(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !force {
+				return ErrPurgeForceFlagRequired
+			}
+			cfg := mustGetContextConfig(cmd)
+			return os.RemoveAll(cfg.GetString("datastore.badger.path"))
```
AndrewSisley, line 37: thought: This will only clear the badger directory, if the user is using another datastore this command will not work (unclear if will panic, or quietly do nothing), and the added tests do not automatically cover other dbs.  I cannot remember how hard-coded other CLI commands are to badger, this might be a drop in the ocean when it comes to adding support for other datastores in the CLI.

cli/purge_test.go
```diff
@@ -0,0 +1,45 @@
...
+	dataDir := filepath.Join(rootDir, "data")
+	err := os.MkdirAll(dataDir, 0755)
+	require.NoError(t, err)
+
+	cmd := NewDefraCommand()
+	cmd.SetArgs([]string{"purge", "--force", "--rootdir", rootDir})
+
+	err = cmd.Execute()
+	require.NoError(t, err)
+	assert.NoDirExists(t, dataDir)
```
AndrewSisley, line 44: todo: This test does not test that the badger data directory is cleared, but the hardcoded path set in the test (that currently corresponds with the default badger path).  Please either change this test (or add another one) so that it will test that the correct directory is cleared, no matter what value is set within `datastore.badger.path`.

AndrewSisley (changes requested):
The code looks good, but I have a couple of todos for you

cli/purge.go
nasdf: When we add more datastores we can have them all use the same `datastore.path` as their root path and have sub directories for each datastore type. Then we just need to delete the root path to remove them all.

cli/purge.go
nasdf: I only remember the force flag from the discussion. As long as its non interactive I'm open to changing it to anything else.

cli/purge_test.go
nasdf: done

shahzadlone (approved):
I understand `-f` might not be sufficient. Perhaps if an additional WARNING message pops up and required an explicit [yes|Yes] input to continue it might be even better? However I am happy as is also and can improve it easily. Leaving LGTM but please do resolve Andy's concerns

cli/purge.go
fredcarle: I think we had discussed the possibility of having a `dev` flag on startup with only `dev` instances capable of calling `purge`.  If we do this, I think we don't even need `force`. Also, having a `dev` flag will allow for other commands to be added to `dev` only in the future. `dev` could by default require no password for the keyring for example. This could all be part of a strategy to get new devs to the "fun part" asap.

AndrewSisley: > I think we had discussed the possibility of having a `dev` flag on startup with only `dev` instances capable of calling `purge`. If we do this, I think we don't even need `force`. Also, having a `dev` flag will allow for other commands to be added to `dev` only in the future. `dev` could by default require no password for the keyring for example. This could all be part of a strategy to get new devs to the "fun part" asap.

I don't remember this, but I like it - is kind of like a build flag without the hassle of multiple builds, and like you said it scales to other situations very nicely too

cli/purge_test.go
AndrewSisley: Why did you need to change the prod code to do this?

nasdf: This was reverted after the team discussion

cli/start.go
```diff
@@ -72,70 +76,98 @@ func MakeStartCommand() *cobra.Command {
...
+			}
 
-			signalCh := make(chan os.Signal, 1)
-			signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
-
+		SELECT:
 			select {
+			case <-purgeSub.Message():
+				if !cfg.GetBool("development") {
+					goto SELECT
```
fredcarle, line 142: question: If the DB receives a purge command while not in development, this seems to give no feedback to the caller as to what happened. Shouldn't we log a `cannot purge production database` error/warning or something like that>

cli/start.go
```diff
@@ -72,70 +76,98 @@ func MakeStartCommand() *cobra.Command {
...
+			signalCh := make(chan os.Signal, 1)
+			signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
+
+			for _, o := range storeOpts {
+				nodeOpts = append(nodeOpts, any(o))
 			}
 
-			n, err := node.NewNode(cmd.Context(), opts...)
+		START:
+			n, err := node.NewNode(cmd.Context(), nodeOpts...)
```
fredcarle, line 125: todo: Please add a top level message, if defra is in development mode, warning the user to not use this configuration in production.

http/handler_extras_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/sourcenetwork/defradb/event"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPurge(t *testing.T) {
```
fredcarle, line 23: todo: Please also add integration tests that show that purging when in production mode will not work and also that purging while in development mode will in fact purge the database.

fredcarle: Discussed on Discord. The `start` command cannot currently be tested with the integration test suite and `purge` is handled within the `start` command. We will find a different way to test the effectiveness of the `purge` command.

AndrewSisley: > The start command cannot currently be tested with the integration test suite and purge is handled within the start command

I'm curious - why can it not be tested with integration tests?

nasdf: > > The start command cannot currently be tested with the integration test suite and purge is handled within the start command
> 
> I'm curious - why can it not be tested with integration tests?

It's possible but there are a few things that make it really difficult and not worth the effort in my opinion. No easy way for random port assignments and the timing of the command (waiting for start up / restart) will cause the tests to be flaky.

nasdf: > Why would any of these tests need to care about port assignment?

To test the purge command you need the http api running.

AndrewSisley: Oh you changed it to go via the http client?  Is that not exposing a highly destructive command a little too much? I wonder if that was an oversight in our last conversation.

nasdf: > Oh you changed it to go via the http client? Is that not exposing a highly destructive command a little too much? I wonder if that was an oversight in our last conversation.

That's why I've added the development mode config value that must be enabled to run purge.

AndrewSisley: Yeah, I assumed it would still only be available locally though, not over http too.

People will still mess up and accidentally set the dev mode to on in production deployments, and at the moment (prior to admin ACP, and correct config of it...) it would allow world plus dog to truncate the db.

fredcarle: We can't put guardrails against devs that shoot themselves in the foot to that extent. Devs can also just expose a production DB without any ACP policies. It would be just as bad. At least here we have to be explicit with putting the database in dev mode to allow the purge command.

fredcarle (approved):
LGTM. I like this approach. 

Don't forget the typo I flagged earlier :)

cli/start.go
nasdf: done. this returns and logs an error now

cli/start.go
nasdf: done

http/handler_extras_test.go
```diff
@@ -0,0 +1,42 @@
...
+	require.NoError(t, err)
+
+	handler, err := NewHandler(cdb)
+	require.NoError(t, err)
+	handler.ServeHTTP(rec, req)
+
+	res := rec.Result()
+	require.Equal(t, 200, res.StatusCode)
+
+	// test will timeout if purge never recieved
```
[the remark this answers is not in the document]
nasdf: done

islamaliev:
bag bush result:
All is working, but I'd like to share some thoughts:
In normal mode:
```shell
$ defradb client purge
Error: run this command again with --force if you really want to purge all data
```
```shell
$ defradb client purge --force
```
There is no output as if it was successful.
Node's output:
```shell
Oct 15 13:57:19.878 INF cli Received purge event; restarting...
Oct 15 13:57:19.878 ERR cli failed to purge $err="cannot purge database when development mode is disabled"
```
I think it's would be better to output that `purge` can not be performed instead of letting it through and silently failing.

islamaliev:
created #3140

Issue #2999, opened 2024-09-13
Views with transforms to not correctly handle related records 


AndrewSisley:
The Lens node does not fully convert the output of a Lens transform to a `core.Doc` - it only converts the top level document, this means that inner documents cannot properly have post-view stuff done to them (e.g. aggregates, filters, etc).

This has mostly been fixed in #2951 /`2951-cached-views` (not yet merged), however some problems remain caused by stuff not expecting schema only definitions (e.g. the sum node).

E.g. the below test will fail on the last action (failure depends on branch):
```
func TestView_OneToManyWithTransformOnOuterAndSum(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type Author {
						name: String
						books: [Book]
					}
					type Book {
						name: String
						author: Author
						pages: Int
					}
				`,
			},
			testUtils.CreateView{
				Query: `
					Author {
						name
						books {
							name
						}
					}
				`,
				SDL: `
					type AuthorView @materialized(if: false) {
						fullName: String
						books: [BookView]
					}
					interface BookView {
						name: String
						pages: Int
					}
				`,
				Transform: immutable.Some(model.Lens{
					// This transform will copy the value from `name` into the `fullName` field,
					// like an overly-complicated alias
					Lenses: []model.LensModule{
						{
							Path: lenses.CopyModulePath,
							Arguments: map[string]any{
								"src": "name",
								"dst": "fullName",
							},
						},
					},
				}),
			},
			testUtils.CreateDoc{
				CollectionID: 0,
				Doc: `{
					"name":	"Ferdowsi"
				}`,
			},
			testUtils.CreateDoc{
				CollectionID: 1,
				DocMap: map[string]any{
					"name":   "Shahnameh",
					"author": testUtils.NewDocIndex(0, 0),
					"pages":  10,
				},
			},
			testUtils.Request{
				Request: `
					query {
						AuthorView {
							fullName
							_sum(books: {field: pages})
						}
					}
				`,
				Results: map[string]any{
					"AuthorView": []map[string]any{
						{
							"fullName": "Ferdowsi",
							"_sum":     10,
						},
					},
				},
			},
		},
	}

	testUtils.ExecuteTestCase(t, test)
}
```
Please do not pick this up until after #2951 is merged, as a lot of the problem has been solved and the code rewritten.

Issue #2640, opened 2024-05-21
Admin Operation Access Control (AAC)


shahzadlone:

shahzadlone:
Some discussion, leaving here for future when we are sorting dependency trees of permissions: https://discord.com/channels/1374835078781468722/1422306930348523560/1423423824019325118

Issue #2018, opened 2023-11-05
ACP - Permissioned Views


shahzadlone:
ACP - Permissioned Views

todo:
- link initial acp doc
- link planning acp doc

PR #3000, opened 2024-09-13
feat: Add materialized views

### LINKED ISSUE #2951: Support for materialized views
We currently have support for non-cached views. We should look into supporting materialized views (views that are static).

AndrewSisley:
## Relevant issue(s)

Resolves #2951

## Description

Adds materialized views.  Also makes materialized views the default (see discord discussion).

The caching behaviour of views in tests is now selected via an environment variable, meaning (with the exception of a few specific examples) a test with a view will test both cacheless and materialized variants - in the CI this adds a new dimension to the matrix, although materialized views are only executed using the simple settings (in-mem store, go client, etc) for now.

https://github.com/sourcenetwork/defradb/issues/2999 has been mostly fixed in this PR, but not completely - this is why some logic in the lens node has changed.

### DIFF
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -37,6 +37,7 @@ jobs:
         lens-type: [wasm-time]
         acp-type: [local]
         database-encryption: [false]
+        view-type: [cacheless]
         include:
           - os: ubuntu-latest
             client-type: go
@@ -45,6 +46,7 @@ jobs:
             lens-type: wasm-time
             acp-type: local
             database-encryption: true
+            view-type: cacheless
           - os: ubuntu-latest
             client-type: go
             database-type: badger-memory
@@ -52,6 +54,7 @@ jobs:
             lens-type: wazero
             acp-type: local
             database-encryption: false
+            view-type: cacheless
           - os: ubuntu-latest
             client-type: go
             database-type: badger-memory
@@ -59,6 +62,7 @@ jobs:
             lens-type: wasmer
             acp-type: local
             database-encryption: false
+            view-type: cacheless
           - os: ubuntu-latest
             client-type: go
             database-type: badger-memory
@@ -66,6 +70,7 @@ jobs:
             lens-type: wasm-time
             acp-type: source-hub
             database-encryption: false
+            view-type: cacheless
           - os: ubuntu-latest
             client-type: http
             database-type: badger-memory
@@ -73,6 +78,7 @@ jobs:
             lens-type: wasm-time
             acp-type: source-hub
             database-encryption: false
+            view-type: cacheless
           - os: ubuntu-latest
             client-type: cli
             database-type: badger-memory
@@ -80,6 +86,15 @@ jobs:
             lens-type: wasm-time
             acp-type: source-hub
             database-encryption: false
+            view-type: cacheless
+          - os: ubuntu-latest
+            client-type: go
+            database-type: badger-memory
+            mutation-type: collection-save
+            lens-type: wasm-time
+            acp-type: local
+            database-encryption: false
+            view-type: materialized
           - os: macos-latest
             client-type: go
             database-type: badger-memory
@@ -87,6 +102,7 @@ jobs:
             lens-type: wasm-time
             acp-type: local
             database-encryption: false
+            view-type: cacheless
 ## TODO: https://github.com/sourcenetwork/defradb/issues/2080
 ## Uncomment the lines below to Re-enable the windows build once this todo is resolved.
 ##        - os: windows-latest
@@ -96,6 +112,7 @@ jobs:
 ##          lens-type: wasm-time
 ##          acp-type: local
 ##          database-encryption: false
+##          view-type: cacheless
 
     runs-on: ${{ matrix.os }}
 
@@ -115,6 +132,7 @@ jobs:
       DEFRA_MUTATION_TYPE: ${{ matrix.mutation-type }}
       DEFRA_LENS_TYPE: ${{ matrix.lens-type }}
       DEFRA_ACP_TYPE: ${{ matrix.acp-type }}
+      DEFRA_VIEW_TYPE: ${{ matrix.view-type }}
 
     steps:
       - name: Checkout code into the directory
--- a/.github/workflows/test-and-upload-coverage.yml
+++ b/.github/workflows/test-and-upload-coverage.yml
@@ -222,6 +222,7 @@ jobs:
             _${{ matrix.mutation-type }}\
             _${{ matrix.lens-type }}\
             _${{ matrix.acp-type }}\
+            _${{ matrix.matrix.view-type }}\
             _${{ matrix.database-encryption }}\
           "
           path: coverage.txt
--- a/cli/view_refresh.go
+++ b/cli/view_refresh.go
@@ -29,7 +29,7 @@ func MakeViewRefreshCommand() *cobra.Command {
 persisting the results.
 
 View is refreshed as the current user, meaning results returned for all subsequent query requests
-to the view will recieve items accessible to the user refreshing the view's permissions.
+to the view will receive items generated using the user refreshing the view's permissions.
 
 Example: refresh all views
   defradb client view refresh
--- a/cli/view_refresh.go
+++ b/cli/view_refresh.go
@@ -1,0 +1,74 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"github.com/sourcenetwork/immutable"
+	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+func MakeViewRefreshCommand() *cobra.Command {
+	var name string
+	var schemaRoot string
+	var versionID string
+	var getInactive bool
+	var cmd = &cobra.Command{
+		Use:   "refresh",
+		Short: "Refresh views.",
+		Long: `Refresh views, executing the underlying query and LensVm transforms and
+persisting the results.
+
+View is refreshed as the current user, meaning results returned for all subsequent query requests
+to the view will recieve items accessible to the user refreshing the view's permissions.
+
+Example: refresh all views
+  defradb view refresh
+
+Example: refresh views by name
+  defradb view refresh --name UserView
+
+Example: refresh views by schema root id
+  defradb view refresh --schema bae123
+
+Example: refresh views by version id. This will also return inactive views
+  defradb view refresh --version bae123
+		`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			store := mustGetContextStore(cmd)
+
+			options := client.CollectionFetchOptions{}
+			if versionID != "" {
+				options.SchemaVersionID = immutable.Some(versionID)
+			}
+			if schemaRoot != "" {
+				options.SchemaRoot = immutable.Some(schemaRoot)
+			}
+			if name != "" {
+				options.Name = immutable.Some(name)
+			}
+			if getInactive {
+				options.IncludeInactive = immutable.Some(getInactive)
+			}
+
+			return store.RefreshViews(
+				cmd.Context(),
+				options,
+			)
+		},
+	}
+	cmd.Flags().StringVar(&name, "name", "", "View name")
+	cmd.Flags().StringVar(&schemaRoot, "schema", "", "View schema Root")
+	cmd.Flags().StringVar(&versionID, "version", "", "View version ID")
+	cmd.Flags().BoolVar(&getInactive, "get-inactive", false, "Get inactive views as well as active")
+	return cmd
+}
--- a/cli/view_refresh.go
+++ b/cli/view_refresh.go
@@ -28,8 +28,9 @@ func MakeViewRefreshCommand() *cobra.Command {
 		Long: `Refresh views, executing the underlying query and LensVm transforms and
 persisting the results.
 
-View is refreshed as the current user, meaning results returned for all subsequent query requests
-to the view will receive items generated using the user refreshing the view's permissions.
+View is refreshed as the current user, meaning the cached items will reflect that user's
+permissions. Subsequent query requests to the view, regardless of user, will receive
+items from that cache.
 
 Example: refresh all views
   defradb client view refresh
--- a/cli/view_refresh.go
+++ b/cli/view_refresh.go
@@ -32,16 +32,16 @@ View is refreshed as the current user, meaning results returned for all subseque
 to the view will recieve items accessible to the user refreshing the view's permissions.
 
 Example: refresh all views
-  defradb view refresh
+  defradb client view refresh
 
 Example: refresh views by name
-  defradb view refresh --name UserView
+  defradb client view refresh --name UserView
 
 Example: refresh views by schema root id
-  defradb view refresh --schema bae123
+  defradb client view refresh --schema bae123
 
 Example: refresh views by version id. This will also return inactive views
-  defradb view refresh --version bae123
+  defradb client view refresh --version bae123
 		`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			store := mustGetContextStore(cmd)
--- a/client/collection_description.go
+++ b/client/collection_description.go
@@ -79,6 +79,15 @@ type CollectionDescription struct {
 	// parsing is done, to avoid storing an invalid policyID or policy resource
 	// that may not even exist on acp.
 	Policy immutable.Option[PolicyDescription]
+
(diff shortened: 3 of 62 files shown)


cli/view_refresh.go
```diff
@@ -0,0 +1,74 @@
...
+		Use:   "refresh",
+		Short: "Refresh views.",
+		Long: `Refresh views, executing the underlying query and LensVm transforms and
+persisting the results.
+
+View is refreshed as the current user, meaning results returned for all subsequent query requests
+to the view will recieve items accessible to the user refreshing the view's permissions.
+
+Example: refresh all views
+  defradb view refresh
```
nasdf, line 35: todo: `defradb client view refresh`

tests/integration/view/one_to_many/simple_test.go
```diff
@@ -402,25 +402,26 @@ func TestView_OneToManyWithDoubleSidedRelation_Errors(t *testing.T) {
 					}
 				`,
 			},
-			testUtils.CreateView{
-				Query: `
-					AuthorView {
-						name
-						books {
+			/*
```
nasdf, line 405: question: did you forget to uncomment or remove this?

nasdf (changes requested):

tests/integration/view/one_to_many/simple_test.go
AndrewSisley: I did! Thanks Keenan

- [ ]  Uncomment test, and revert the second half of the test

cli/view_refresh.go
```diff
@@ -0,0 +1,74 @@
...
+	var versionID string
+	var getInactive bool
+	var cmd = &cobra.Command{
+		Use:   "refresh",
+		Short: "Refresh views.",
+		Long: `Refresh views, executing the underlying query and LensVm transforms and
+persisting the results.
+
+View is refreshed as the current user, meaning results returned for all subsequent query requests
+to the view will recieve items accessible to the user refreshing the view's permissions.
```
fredcarle, line 32: praise: Thanks for documenting this. 

thought: It makes me think though that this is most likely not the way we want to handle access long term. Access should be reflecting the rights of the user doing the request, not that of the user who refreshed the view. Curious to have your opinion on this.

client/collection_description.go
```diff
@@ -79,6 +79,15 @@ type CollectionDescription struct {
 	// parsing is done, to avoid storing an invalid policyID or policy resource
 	// that may not even exist on acp.
 	Policy immutable.Option[PolicyDescription]
+
+	// IsMaterialized defines whether the items in this collection are cached or not.
+	//
+	// If it is true, they will be, if false, the data returned on query will be calculated
+	// at query-time from source.
+	//
+	// At the moment this can only be set to `false` if this collection sources it's data from
```
fredcarle, line 88: typo: `its data`

client/db.go
```diff
@@ -196,6 +196,14 @@ type Store interface {
 		transform immutable.Option[model.Lens],
 	) ([]CollectionDefinition, error)
 
+	// RefreshViews refreshes the caches of all views matching the given options.  If no options are set all views
```
fredcarle, line 199: suggestion: `If no options are set,`

client/db.go
```diff
@@ -196,6 +196,14 @@ type Store interface {
...
+	// will be refreshed.
+	//
+	// The cached result is dependent on the ACP settings of the source data and the permissions of the user making
```
fredcarle, line 202: thought: This makes me think that there might be need for setting who is allowed to cache views on a given node. Otherwise this could be abused.

internal/core/key.go
```diff
@@ -77,6 +79,24 @@ type DataStoreKey struct {
...
+// ViewCacheKey is a trimmed down [DataStoreKey] used for caching the results
+// of View items.
+//
+// It is stored in the format `/collection/vi/[CollectionRootID]/[ItemID]`. It points to the
+// full serialized View item.
+type ViewCacheKey struct {
+	// CollectionRootID is the Root of the Collection that this item belongs to.
+	CollectionRootID uint32
+
+	// ItemID is the unique (to this CollectionRootID) of the View item.
```
fredcarle, line 91: typo: is the unique ID?

tests/integration/collection_description/updates/replace/materialized_test.go
```diff
@@ -0,0 +1,121 @@
...
+					}
+				`,
+			},
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{ "op": "replace", "path": "/1/IsMaterialized", "value": false }
+					]
+				`,
+				ExpectedError: "non-materialized collections (only views) are not supported. Collection: User",
```
fredcarle, line 38: suggestion: This reads a little weird. I would suggest to move the parenthesis at the end `non-materialized collections are not supported. (only views)`

islamaliev (commented):
Just half way through. Looks good so far.

cli/view_refresh.go
AndrewSisley: I like the current behaviour long term, and see very few alternatives (we could have the node-user refresh the view).

Worth remembering is that medium-long term view items will be shareable across P2P without needing to share the underlying docs.  And we have Lens stuff at play too (not all view items are sourced from any document, Lens can create their own).

The view has its own set of permissions that should be respected, although this was not handled when ACP was introduced to Defra https://github.com/sourcenetwork/defradb/issues/2018.

IMO a view is a copy of the transformed data, a user has created a view and is sharing it with who ever they see fit, as a convenient alternative to printing the results out and mailing it to people.

client/collection_description.go
AndrewSisley: :) Cheers

- [ ] Typo

client/db.go
AndrewSisley: Leaving this one as is, I don't think the comma helps, and I think I'm usually guilty of over-using them

client/db.go
AndrewSisley: Yes of course, we'll either want it covered by admin ACP or allow users to own views.

internal/core/key.go
AndrewSisley: Cheers, took me a while to figure out what you meant on re-read :)

- [ ] Add `ID` after closing of brackets

internal/db/view.go
```diff
@@ -88,3 +93,218 @@ func (db *db) addView(
...
+			err = defErr
+		}
+	}()
+
+	err = source.Start()
+	if err != nil {
+		return err
+	}
+
+	hasNext, err := source.Next()
```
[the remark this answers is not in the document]
AndrewSisley: Agreed, although I think `hasNext` is used more frequently in the codebase.

- [ ] Rename to `hasValue`

cli/view_refresh.go
AndrewSisley: Cheers Keenan, nice spot :)

- [ ] Fix CLI doc

cli/view_refresh.go
fredcarle: > Worth remembering is that medium-long term view items will be shareable across P2P without needing to share the underlying docs.

This is specially why we can't have this behaviour long term. We can't allow some random user to access the contents of a view because the user that created the view had the rights to the documents.

client/db.go
fredcarle: Just adding that without the comma it reads `If no options are set-all-views...`

cli/view_refresh.go
AndrewSisley: The user that created the view would need to give that random user access to it.

There is nothing we (or anyone else, really) can do to stop users sharing data they already have read access too - that is not a technical problem (see print and mail example).

AndrewSisley: Requiring view-viewers to also have read access to the underlying data would totally undermine what I see as the primary use case - exposing data aggregates without exposing the data.

fredcarle: > The user that created the view would need to give that random user access to it.

Right. Hence why I'm saying that the description is not what we want long term.

> There is nothing we (or anyone else, really) can do to stop users sharing data they already have read access to - that is not a technical problem (see print and mail example).

This tells me we have a misunderstanding. We can clarify things over a call if you'd like.

client/db.go
AndrewSisley: Discussed on discord, adding comma

- [ ] add comma

internal/db/definition_validation.go
```diff
@@ -978,3 +980,39 @@ func validateSchemaNameNotEmpty(
...
+		}
+	}
+
+	return nil
+}
+
+// validateCollectionMaterialized verifies that a materialized view has no ACP policy.
+//
+// Long term we wish to support this, however for now we block it off.
+func validateMaterializedHasNoACP(
```
shahzadlone, line 1005: Suggestion: perhaps instead of `NoACP` in the name it can be `NoPolicy` or `NoACPPolicy` as no ACP can mean that acp is turned off completely.

internal/db/view.go
```diff
@@ -88,3 +93,218 @@ func (db *db) addView(
...
+	return q.Close()
+}
+
+func (db *db) generateMaximalSelectFromCollection(
+	ctx context.Context,
+	col client.CollectionDefinition,
+	fieldName immutable.Option[string],
+	typesHit map[string]struct{},
+) (*request.Select, error) {
+	// `__-` is an impossible field name prefix, so we can safely concat using it as a separator without risk
```
shahzadlone, line 260: Question: curious were this seperater comes from is a one char or a more symmetric separator not possible?

cli/view_refresh.go
```diff
@@ -0,0 +1,74 @@
...
+	var versionID string
+	var getInactive bool
+	var cmd = &cobra.Command{
+		Use:   "refresh",
+		Short: "Refresh views.",
+		Long: `Refresh views, executing the underlying query and LensVm transforms and
+persisting the results.
+
+View is refreshed as the current user, meaning results returned for all subsequent query requests
+to the view will receive items generated using the user refreshing the view's permissions.
```
fredcarle, line 32: thought: This is better but still not very clear. I assume a new user looking at this would still get confused.

suggestion: I said this at some point previously but I can't remember where exactly and John just reminded me of it: We should probably materialize views from the view point of the node and not that of a given user. Queries on the view can then be filtered through ACP for the user doing the request. This will alleviate the need to have multiple materialized versions of the view.

AndrewSisley: > This will alleviate the need to have multiple materialized versions of the view.

This is an alternative, and long term we probably want both anyway (e.g. here's my personal quick view, go and have a look).

I just picked the easiest, as there is time pressure on this PR and getting the node user is a bit more involved.

> This is better but still not very clear. I assume a new user looking at this would still get confused

Suggestions appreciated, it is a moderately complex thing to state in a line or two.

internal/db/view.go
AndrewSisley: a double underscore prefix is reserved for special gql fields like `__type`, and no special field starts with `-`, thus `__-` is safe(and a bit more obvious than `__y`)

internal/db/definition_validation.go
AndrewSisley: Not a problem.

- [ ] Rename `validateMaterializedHasNoPolicy`

cli/view_refresh.go
AndrewSisley: > Lets clear up the question of what user is used to generate the view and then we can improve the documentation. 

That sounds like unwanted scope creep forced into a time-sensitive and deliberately lean initial PR, but sure.

client/db.go
shahzadlone: Might be worth documenting this in: https://github.com/sourcenetwork/defradb/issues/2640

tests/integration/collection_description/updates/replace/view_policy_test.go
```diff
@@ -0,0 +1,95 @@
...
+					}
+				`,
+			},
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{ "op": "replace", "path": "/2/IsMaterialized", "value": true }
+					]
+				`,
+				ExpectedError: "materialized views do not support ACP. Collection: UserView",
```
shahzadlone, line 89: praise: Thanks for not forgetting this :)

cli/view_refresh.go
AndrewSisley: Discussed in standup, user-identity is fine for now, long term we may want something else

tests/integration/view/simple/materialized_test.go
```diff
@@ -0,0 +1,129 @@
...
+				`,
+			},
+			testUtils.Request{
+				Request: `query {
+							UserView {
+								name
+							}
+						}`,
+				Results: map[string]any{
+					// Even though UserView was created after the document was created, the results are
```
shahzadlone, line 61: question: It says: `Even though UserView was created after the document was created` but correct me if I am wrong weather `UserView` is created after or even before the results will always be empty unless the views are refreshed manually.

shahzadlone (approved):
LGTM

tests/integration/view/simple/materialized_test.go
AndrewSisley: You are correct, I want the createDoc before the view though, as it is more likely to change/break in the future (e.g. auto-refreshing on View create) - this test will catch that as a behaviour change, whereas creating the doc after defining the view would not.

cli/view_refresh.go
fredcarle: Discussed in the standup. This implementation is working for the use case that triggered this PR and as such we can leave it as is.

As for the wording of the documentation, here is an possible alternative:
```suggestion
View is refreshed as the current user, meaning the cached items will reflect that user's 
permissions. Subsequent query requests to the view, regardless of user, will receive 
items from that cache.
```

AndrewSisley: Will update, thanks Fred

- [ ] Update doc

fredcarle (approved):
LGTM!

internal/planner/view.go
```diff
@@ -117,3 +107,142 @@ func (n *viewNode) Close() error {
 
 	return nil
 }
+
+func convertBetweenMaps(srcMap *core.DocumentMapping, dstMap *core.DocumentMapping, src core.Doc) core.Doc {
```
islamaliev, line 111: suggestion: would be nice to have a comment for this function. From the name it's not clear why it's called "convert" although both source and destination have the same type.

internal/planner/view.go
```diff
@@ -117,3 +107,142 @@ func (n *viewNode) Close() error {
...
+
+	var err error
+	n.queryResults, err = n.p.txn.Datastore().Query(n.p.ctx, query.Query{
+		Prefix: prefix.ToString(),
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
islamaliev, line 212: nitpick: `return err`

internal/planner/view.go
AndrewSisley: I mean, the first two params to this function are `srcMap` and `dstMap`, and the place it is called is heavily documented with the reason for calling it.

internal/planner/view.go
```diff
@@ -117,3 +107,142 @@ func (n *viewNode) Close() error {
...
+}
+
+func (n *cachedViewFetcher) Next() (bool, error) {
+	result, hasNext := n.queryResults.NextSync()
+	if !hasNext || result.Error != nil {
+		return false, result.Error
+	}
+
+	var err error
+	n.currentValue, err = core.UnmarshalViewItem(n.documentMapping, result.Value)
```
islamaliev, line 230: question: is this view stuff really part of the `core`?

tests/integration/test_case.go
```diff
@@ -220,6 +227,22 @@ type CreateView struct {
 	ExpectedError string
 }
 
+type RefreshViews struct {
```
islamaliev, line 230: suggestion: would be nice to have the struct documented

.github/workflows/test-and-upload-coverage.yml
shahzadlone, line 1: todo: The coverage file names will clash currently as you forget to add the new type after line `225`

Just append after this block:

```YAML
coverage\
            _${{ matrix.os }}\
            _${{ matrix.client-type }}\
            _${{ matrix.database-type }}\
            _${{ matrix.mutation-type }}\
            _${{ matrix.lens-type }}\
            _${{ matrix.acp-type }}\
            _${{ matrix.database-encryption }}\
            
```

islamaliev (approved):
LGTM! Good job Andy!

.github/workflows/test-and-upload-coverage.yml
AndrewSisley: Cheers Shahzad, sorted.

internal/planner/view.go
AndrewSisley: I wanted it in encoding but that created a circular dependency, so it had to go here.

tests/integration/test_case.go
AndrewSisley: sorted

internal/planner/view.go
AndrewSisley: For some reason I dislike that if it is not with the immediate result of another func call (e.g. `return foo()`), otherwise it looks like it is actually returning a non-nil error.  Leaving as is.

PR #3013, opened 2024-09-16
fix: GraphQL null argument parsing

### LINKED ISSUE #3012: Explicitly null GraphQL input values cause panic
We do not correctly handle explicitly null values in GQL requests.

nasdf:
## Relevant issue(s)

Resolves #3012

## Description

This PR fixes an issue where explicitly null GQL inputs would cause a panic.

## Tasks

## How has this been tested?

Added tests

- MacOS


internal/planner/mapper/targetable.go
```diff
@@ -110,7 +110,7 @@ func filterObjectToMap(mapping *core.DocumentMapping, obj map[connor.FilterKey]a
 	for k, v := range obj {
 		switch keyType := k.(type) {
 		case *PropertyIndex:
-			subObj := v.(map[connor.FilterKey]any)
+			subObj, _ := v.(map[connor.FilterKey]any)
```
AndrewSisley, line 113: todo: This looks like it was accidentally left in after debugging - can it be reverted?

tests/integration/mutation/create/with_null_input_test.go
```diff
@@ -0,0 +1,110 @@
...
+			testUtils.SchemaUpdate{
+				Schema: `
+					type Users {
+						name: String
+					}
+				`,
+			},
+			testUtils.Request{
+				Request: `mutation {
+					create_Users(encrypt: null, input: {name: "Bob"}) {
```
AndrewSisley, line 32: question: `input` is not nillable? (I don't see any tests for that)

tests/integration/mutation/create/with_null_input_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2024 Democratized Data Foundation
```
AndrewSisley, line 1: priase: Thanks for all theses tests Keenan, looks good :)

AndrewSisley (approved):
Looks good to me Keenan, thanks for sorting this out so quickly.  Approving now, but have a tiny todo and a question for you before merge please :)

internal/planner/mapper/targetable.go
nasdf: This line was causing a panic when the property value is nil. Is there a different way I should be handling the nil value here?

AndrewSisley: Ah I see.  I would prefer a `if v != nil` wrapper then - it is a much stronger declaration of intent.

nasdf: done

tests/integration/mutation/create/with_null_input_test.go
nasdf: nice catch! added a test

PR #3021, opened 2024-09-17
feat: Add support for one sided relations

### LINKED ISSUE #2830: One sided primary relations are not queryable
Found by David, e.g.:
######### Schema
```gql
type TypeA {
  test: String
}

type TypeB {
  test: String
  ref: TypeA @primary
}
```

######### Create Types
```gql
mutation create_typeA {
  create_TypeA(input: { test: "1" }) {
    _docID
  } # bae-31fc529a-4edb-54e3-bf44-439a6570daf4
}
mutation create_typeB {
  create_TypeB(
    input: { test: "2", ref_id: "bae-31fc529a-4edb-54e3-bf44-439a6570daf4" }
  ) {
    _docID
  } # bae-b7a8d5ee-375d-5d90-90a6-f518fed888b2
}
```

######### Query
```gql
query { TypeB { test ref { test } } }
```

######### Result
```json
{
  "data": null,
  "errors": [
    "The given field does not exist. Name: ref"
  ]
}
```

The schema is valid (although I'm not sure we test with the primary directive), as one-sided relations are fairly new it might be that we never actually test querying them and have relied on introspection/collection-description tests.

AndrewSisley:
## Relevant issue(s)

Resolves #2830

## Description

Adds support for one sided relations.

I thought we had this a while back, but David discovered otherwise.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -286,6 +286,11 @@ func findFilteredByRelationFields(
 }
 
 func (p *Planner) tryOptimizeJoinDirection(node *invertibleTypeJoin, parentPlan *selectTopNode) error {
+	if !node.childSide.relFieldDef.HasValue() {
+		// If the relation is one sided we cannot invert the join, so return early
+		return nil
+	}
+
 	filteredSubFields := findFilteredByRelationFields(
 		parentPlan.selectNode.filter.Conditions,
 		node.documentMapping,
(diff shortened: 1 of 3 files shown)


internal/planner/type_join.go
```diff
@@ -377,8 +382,12 @@ func (p *Planner) newInvertableTypeJoin(
 }
 
 type joinSide struct {
-	plan               planNode
-	relFieldDef        client.FieldDefinition
+	plan planNode
+	// The field definition of the relation-object field on this side of the relation.
+	//
+	// This will always have a value on the primary side, but it may not have a value on
+	// the secondary side, as the secondary half of the relation is optional.
```
fredcarle, line 389: thought: I look forward to the secondary side being handle automatically with the proposed handling of relationship from the primary side.

fredcarle (approved):
LGTM. Thanks for fixing this.

internal/planner/planner.go
```diff
@@ -286,6 +286,11 @@ func findFilteredByRelationFields(
 }
 
 func (p *Planner) tryOptimizeJoinDirection(node *invertibleTypeJoin, parentPlan *selectTopNode) error {
+	if !node.childSide.relFieldDef.HasValue() {
+		// If the relation is one sided we cannot invert the join, so return early
+		return nil
```
shahzadlone, line 291: suggestion: If not too much trouble would be nice to hit this case as it shows not covered right now.

shahzadlone (approved):
LGTM, just one suggestion

islamaliev (approved):
LGTM!

internal/planner/planner.go
AndrewSisley: It might be dead code, but I'm not sure - the invertable join stuff is quite complex and if it isn't hit by the new test I don't know what will hit it.  And I don't want to spend much time on this right now, especially given the pending secondary relation rework Fred mentioned.

PR #3023, opened 2024-09-17
fix: Handle missing type in an SDL

### LINKED ISSUE #3022: Panic when omitting field type from SDL
Spotted by David, e.g.
```
type User {
  name @default(string: "foo")
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #3022

## Description

Handles missing type in an SDL instead of panicing.


nasdf (approved):
LGTM. I'm surprised that this isn't considered invalid GQL and fails parsing instead.

AndrewSisley:
> LGTM. I'm surprised that this isn't considered invalid GQL and fails parsing instead.

Might just be the library we are using leaks it, made me smile when Dave messaged about it lol

PR #3026, opened 2024-09-17
chore: Make keyring non-interactive

### LINKED ISSUE #2995: Support for non-interactive use of the keyring
Currently the password his passed using the interactive shell. Ideally, for automated deployments, we would support non-interactive usage.

fredcarle:
## Relevant issue(s)

Resolves #2995 

## Description

This PR makes the keyring interactions non-interactive by requiring the keyring password to be set as an environment variable secret. It also adds support for that secret to be stored in a `.env` file in the working directory or in a file at a path defined by the `--secret-file` flag.

Making the keyring non-interactive is necessary to support automated deployments. 

## Tasks

## How has this been tested?

make test

Also tested manually starting Defra with a `.env` file to make sure that the secret was picked up.

- MacOS

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -43,30 +43,31 @@ var configPaths = []string{
 
 // configFlags is a mapping of cli flag names to config keys to bind.
 var configFlags = map[string]string{
-	"log-level":          "log.level",
-	"log-output":         "log.output",
-	"log-format":         "log.format",
-	"log-stacktrace":     "log.stacktrace",
-	"log-source":         "log.source",
-	"log-overrides":      "log.overrides",
-	"no-log-color":       "log.colordisabled",
-	"url":                "api.address",
-	"max-txn-retries":    "datastore.maxtxnretries",
-	"store":              "datastore.store",
-	"valuelogfilesize":   "datastore.badger.valuelogfilesize",
-	"peers":              "net.peers",
-	"p2paddr":            "net.p2paddresses",
-	"no-p2p":             "net.p2pdisabled",
-	"allowed-origins":    "api.allowed-origins",
-	"pubkeypath":         "api.pubkeypath",
-	"privkeypath":        "api.privkeypath",
-	"keyring-namespace":  "keyring.namespace",
-	"keyring-backend":    "keyring.backend",
-	"keyring-path":       "keyring.path",
-	"no-keyring":         "keyring.disabled",
-	"no-encryption-key":  "keyring.noencryptionkey",
-	"source-hub-address": "acp.sourceHub.address",
-	"development":        "development",
+	"log-level":           "log.level",
+	"log-output":          "log.output",
+	"log-format":          "log.format",
+	"log-stacktrace":      "log.stacktrace",
+	"log-source":          "log.source",
+	"log-overrides":       "log.overrides",
+	"no-log-color":        "log.colordisabled",
+	"url":                 "api.address",
+	"max-txn-retries":     "datastore.maxtxnretries",
+	"store":               "datastore.store",
+	"valuelogfilesize":    "datastore.badger.valuelogfilesize",
+	"peers":               "net.peers",
+	"p2paddr":             "net.p2paddresses",
+	"no-p2p":              "net.p2pdisabled",
+	"allowed-origins":     "api.allowed-origins",
+	"pubkeypath":          "api.pubkeypath",
+	"privkeypath":         "api.privkeypath",
+	"keyring-namespace":   "keyring.namespace",
+	"keyring-backend":     "keyring.backend",
+	"keyring-path":        "keyring.path",
+	"no-keyring":          "keyring.disabled",
+	"no-encryption-key":   "keyring.noencryptionkey",
+	"keyring-secret-file": "keyring.secretfile",
+	"source-hub-address":  "acp.sourceHub.address",
+	"development":         "development",
 }
 
 // configDefaults contains default values for config entries.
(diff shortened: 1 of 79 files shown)


cli/config.go
```diff
@@ -94,6 +96,9 @@ var configDefaults = map[string]any{
 
 // defaultConfig returns a new config with default values.
 func defaultConfig() *viper.Viper {
+	// load environment variables from .env file if one exists
+	_ = godotenv.Load()
```
nasdf, line 100: todo: viper has built in support for envfiles. We should be able to remove this dependency

fredcarle: I think viper can read from only one config file at a time. We already have the defra config file that is used with viper.

shahzadlone (commented):
might want to also ignore the `.env` file in the `.gitignore`

shahzadlone (commented):
I like the approach, just had some questions:

question: Where is the loading of the environment variables from '.env' file being done?

question: When you say `root` directory what do you mean i.e. what are the paths that this `.env` file will be picked from (and the order of priority if many paths)?

question: Have you tested this works when `.env` flow?

cli/keyring_export_test.go
```diff
@@ -13,21 +13,20 @@ package cli
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestKeyringExport(t *testing.T) {
 	rootdir := t.TempDir()
-	readPassword = func(_ *cobra.Command, _ string) ([]byte, error) {
-		return []byte("secret"), nil
-	}
+	err := os.Setenv("DEFRA_KEYRING_SECRET", "password")
```
islamaliev, line 28: question: don't we want to keep at least one test that uses the old prompt?

fredcarle: The prompt has been removed entirely.

fredcarle:
> question: Where is the loading of the environment variables from '.env' file being done?

@shahzadlone, it's done from where Defra is started.

> question: When you say root directory what do you mean i.e. what are the paths that this .env file will be picked from (and the order of priority if many paths)?

Currently there is only one path and it's `./.env`. Eventually I would like to add a flag to let users set the path.

> question: Have you tested this works when .env flow?

Yes I've tested this manually.

cli/config.go
nasdf: ahh okay that seems like an odd choice on their part

fredcarle:
> > question: When you say root directory what do you mean i.e. what are the paths that this .env file will be picked from (and the order of priority if many paths)?
> 
> Currently there is only one path and it's `./.env`. Eventually I would like to add a flag to let users set the path.

@shahzadlone I've added a flag to let the user define the path and updated the documentation to say `working directory` instead of `root directory`.

cli/keyring_generate.go
nasdf, line 1: thought: I'm not sure how useful this command is now that we generate keys on start

cli/config.go
```diff
@@ -42,29 +43,31 @@ var configPaths = []string{
...
+	"no-p2p":              "net.p2pdisabled",
+	"allowed-origins":     "api.allowed-origins",
+	"pubkeypath":          "api.pubkeypath",
+	"privkeypath":         "api.privkeypath",
+	"keyring-namespace":   "keyring.namespace",
+	"keyring-backend":     "keyring.backend",
+	"keyring-path":        "keyring.path",
+	"no-keyring":          "keyring.disabled",
+	"no-encryption-key":   "keyring.noencryptionkey",
+	"keyring-secret-file": "keyring.secretfile",
```
nasdf, line 68: question: What do you think about making this more general purpose and renaming to `secretfile`? If we decide to add more secrets in the future it might be confusing having the keyring name.

nasdf (approved):
A few thoughts and questions but nothing blocking.

shahzadlone (approved):
LGTM

cli/config.go
```diff
@@ -42,29 +43,31 @@ var configPaths = []string{
...
+	"p2paddr":             "net.p2paddresses",
+	"no-p2p":              "net.p2pdisabled",
+	"allowed-origins":     "api.allowed-origins",
+	"pubkeypath":          "api.pubkeypath",
+	"privkeypath":         "api.privkeypath",
+	"keyring-namespace":   "keyring.namespace",
+	"keyring-backend":     "keyring.backend",
+	"keyring-path":        "keyring.path",
+	"no-keyring":          "keyring.disabled",
+	"no-encryption-key":   "keyring.noencryptionkey",
```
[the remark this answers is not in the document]
fredcarle: That's a sensible thought. I'll look into it.

cli/keyring_generate.go
fredcarle: It sensible if someone wants to generate the keys first for some reason. I had initially removed it then decided to keep it as a convenience.

cli/config.go
fredcarle: I did this initially and then reverted. I think I'm more and more in favour of making it general purpose. I'll change it.

cli/config.go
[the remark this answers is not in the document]
fredcarle: changed to `datastore.noencryption` for now. We can add a keyname configuration later.

cli/config.go
fredcarle: Moved to `secretfile`.

cli/config.go
```diff
@@ -159,6 +163,9 @@ func loadConfig(rootdir string, flags *pflag.FlagSet) (*viper.Viper, error) {
 		}
 	}
 
+	// load environment variables from .env file if one exists
+	_ = godotenv.Load(cfg.GetString("keyring.secretfile"))
```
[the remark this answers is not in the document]
fredcarle: Added error handling when it's not an file not found error.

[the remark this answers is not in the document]
fredcarle: The `.env` file is optional so I don't want to pollute the logs with statements of a missing optional file.

PR #3028, opened 2024-09-18
feat: Inline array filters

### LINKED ISSUE #2857: Enable documents filtering by inline array fields
currently we allow filtering of documents by related objects:
```graphql
query {
  Author(filter: {
    books: {rating: {_gt: 3}}
  }) {
    name
    books {
      name
    }
  }
}
```
But there is no way we can do it for inline arrays like `[String]`.
One of the ideas is to introduce new operators `_all`, `_any` and `_none` that should count number of filter matches for every element in the array:
```graphql
query {
  Author(filter: {
    genres: {_any: {_eq: "Horror"}}
  }) {
    name
  }
}
```
```graphql
query {
  Author(filter: {
    publishedYears: {_all: {_gt: 2010}}
  }) {
    name
  }
}
```

Alternatively (or additionally) we can integrate aggregates into filters with their own filters within:
```graphql
query {
  Author(filter: {
    _count(books: { filter: {
      rating: {_gt: 3}
    }}): {_eq: 2}
  }) {
    name
  }
}
```
This would allow us to specify exactly how many matching elements should be in the array.

Tangentially related idea:
it might be also useful to request indexes of matching elements into the array and potentially afterwards use them to manipulate the array.

nasdf:
## Relevant issue(s)

Resolves #2857

## Description

This PR adds three new filter types for inline array types: `_all`, `_any`, and `_none`.

## Tasks

## How has this been tested?

Added integration tests

- MacOS


fredcarle (approved):
LGTM! Very nice and thanks for all the tests.

internal/connor/all.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/sourcenetwork/immutable"
+)
+
+// all is an operator which allows the evaluation of
+// a number of conditions over a list of values
+// matching if all of them match.
+func all(condition, data any) (bool, error) {
+	switch t := data.(type) {
+	case []string:
```
islamaliev, line 14: todo: I think it's better to use reflection here to avoid repetitive code:
```go
func all(condition, data any) {
    val := reflect.ValueOf(v)
    if val.Kind() != reflect.Slice {
        return false, client.NewErrUnhandledType("data", data)
    }

    for i := 0; i < val.Len(); i++ {
        m, err := eq(condition, val.Index(i).Interface())
            if err != nil {
                return false, err
            } else if !m {
                return false, nil
            }
        }
    }
    return true, nil
}
```
Or at least use generic:
```go
func allInArr[T any](slice []T, condition any) (bool, error) {
	for _, item := range slice {
		m, err := eq(condition, item)
		if err != nil {
			return false, err
		}
		if !m {
			return false, nil
		}
	}
	return true, nil
}

func all(condition, data any) (bool, error) {
	switch t := data.(type) {
	case []string:
		return allInArr(t, condition)
	case []immutable.Option[string]:
		return allInArr(t, condition)
	case []int64:
		return allInArr(t, condition)
// ...
```
This will also improve test coverage.

internal/connor/any.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/sourcenetwork/immutable"
+)
+
+// anyOp is an operator which allows the evaluation of
+// a number of conditions over a list of values
+// matching if any of them match.
+func anyOp(condition, data any) (bool, error) {
+	switch t := data.(type) {
+	case []string:
```
islamaliev, line 14: todo: please use the same approach as with `all`

internal/connor/eq.go
```diff
@@ -29,28 +29,32 @@ func eq(condition, data any) (bool, error) {
...
 		}
-		data = arr.Value()
 
 	case immutable.Option[string]:
-		if !arr.HasValue() {
-			return condition == nil, nil
+		if arr.HasValue() {
+			data = arr.Value()
+		} else {
+			data = nil
```
islamaliev, line 56: suggestion: this could also be turned into generic code

internal/request/graphql/schema/types/base.go
```diff
@@ -40,6 +40,28 @@ func BooleanOperatorBlock() *gql.InputObject {
...
+			},
+			"_all": &gql.InputObjectFieldConfig{
+				Description: allOperatorDescription,
+				Type:        op,
+			},
+			"_none": &gql.InputObjectFieldConfig{
+				Description: noneOperatorDescription,
+				Type:        op,
+			},
+		},
```
islamaliev, line 61: suggestion: this can be extracted into a function and used everywhere

tests/integration/query/inline_array/with_filter_all_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+package inline_array
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestQueryInlineStringArrayWithAllFilter(t *testing.T) {
```
islamaliev, line 19: suggestion: we try to stick to the following test naming convention:
`<Category>_<Condition|Prerequisite>_<Result|Expectation>`
In this case it could be 
`TestQueryInlineStringArray_WithAllFilter_Succeeds`

tests/integration/query/inline_array/with_filter_all_test.go
```diff
@@ -0,0 +1,305 @@
...
+				Doc: `{
+					"name": "Fred",
+					"testScores": [0, 60]
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					Users(filter: {testScores: {_all: {_lt: 70}}}) {
+						name
+					}
```
islamaliev, line 147: suggestion: would be also great to see a reassurance (in for of another test) that this new op works with compound conditions. For example:
`filter: {testScores: {_all: {_and: [{_lt: 70}, {_gte: 0}]}}}`
And to do the same with other new ops.

islamaliev (changes requested):
Look good. I just requested few changes.

internal/connor/all.go
fredcarle: reflect is a bit slow and not fully supported in wasm so lets try to avoid it for now.

AndrewSisley: 100% agree with Fred, we should avoid reflect, especially in a performance sensitive area like this.  Extracting the shared logic to a generic func would be good though.

internal/request/graphql/schema/manager.go
```diff
@@ -188,6 +188,22 @@ func defaultTypes(
...
+	jsonOpBlock := schemaTypes.JSONOperatorBlock(jsonScalarType)
+	blobOpBlock := schemaTypes.BlobOperatorBlock(blobScalarType)
+	dateTimeOpBlock := schemaTypes.DateTimeOperatorBlock()
+
+	notNullIntOpBlock := schemaTypes.NotNullIntOperatorBlock()
+	notNullFloatOpBlock := schemaTypes.NotNullFloatOperatorBlock()
+	notNullBooleanOpBlock := schemaTypes.NotNullBooleanOperatorBlock()
+	notNullStringOpBlock := schemaTypes.NotNullStringOperatorBlock()
+	notNullJSONOpBlock := schemaTypes.NotNullJSONOperatorBlock(jsonScalarType)
+	notNullBlobOpBlock := schemaTypes.NotNullBlobOperatorBlock(blobScalarType)
```
AndrewSisley, line 205: question: I'm curious - why did you pull the existing stuff out of the return and into temp vars? And then add new func calls within the return?

internal/request/graphql/schema/types/base.go
AndrewSisley: I have mixed feelings about the extraction of the full `InputObjectFieldConfig` blocks to a shared func (it couples them together, and hides what each InputObject contains, yet the enforcement of consistency is nice).

internal/request/graphql/schema/types/base.go
```diff
@@ -40,6 +40,28 @@ func BooleanOperatorBlock() *gql.InputObject {
 	})
 }
 
+// BooleanListOperatorBlock filter block for [Boolean] types.
+func BooleanListOperatorBlock(op *gql.InputObject) *gql.InputObject {
+	return gql.NewInputObject(gql.InputObjectConfig{
+		Name:        "BooleanListOperatorBlock",
+		Description: "These are the set of filter operators available for use when filtering on [Boolean] values.",
+		Fields: gql.InputObjectConfigFieldMap{
+			"_any": &gql.InputObjectFieldConfig{
```
AndrewSisley, line 49: suggestion: Independent of Islam's suggestion, it might be nice to extract the names to consts in the client/request package and then also reference them in the connor package as well as here.

tests/integration/query/inline_array/with_filter_all_test.go
AndrewSisley: nitpick: I did think we agreed to not force that suggestion upon developers though, especially the last block.

islamaliev: I agree that that last part is optional

internal/request/graphql/schema/types/base.go
nasdf: My preference is to keep them as is because it makes modifying each individual type easier.

tests/integration/query/inline_array/with_filter_all_test.go
nasdf: I've added two compound tests.

tests/integration/query/inline_array/with_filter_all_test.go
nasdf: done

internal/connor/eq.go
nasdf: done

internal/connor/any.go
nasdf: done

internal/connor/all.go
nasdf: done

internal/request/graphql/schema/manager.go
nasdf: It made sense to me organizationally. Also the temp vars are used as inputs to the calls in the return.

internal/request/graphql/schema/types/base.go
nasdf: This is out of scope for this PR imo, but I will create an issue for this.

islamaliev (approved):
LGTM!

shahzadlone (approved):
LGTM

islamaliev:
Bug bash results: all good

Issue #2928, opened 2024-08-19
Unique index on relation filter does not return all records with that child


AndrewSisley:
Schema:
```gql
type PlayerAction @index(unique: true, fields: ["action_id", "player_id", "createdAt"], name: "action_unqiue") {
  metadata: JSON
  deleted: Boolean
  action: Action @relation(name: "actionPlayerActions")
  player: Player @relation(name: "playerPlayerActions")
  playerRewards: [PlayerActionToPlayerReward] @relation(name: "playerActionPlayerRewards")
  createdAt: DateTime
  oldId: String
}
type Action {
  key: String @index(unique: true) # @unique @db.VarChar(30)
  name: String # @db.VarChar(100)
  description: String # @db.VarChar(250)
  enabled: Boolean # @default(true)
  levelRequired: Int # @default(0)
  actionRequirements: [ActionRequirement] @relation(name: "actionActionRequirements")
  playerActions: [PlayerAction] @relation(name: "actionPlayerActions")
  createdAt: DateTime # @default(now())
  oldId: String
}
```
Query:
```gql
query query_PlayerAction {
  byId: PlayerAction(
    filter: {
      player_id: {_eq: "bae-74922412-523d-556a-897b-3af6fc47701c"}, 
      action_id: {_in: ["bae-d6e62c28-c064-5e96-8404-de8facc1f401"]}
      deleted: {_ne: true}}
  ) {
    action {
      key
      _docID
    }
  }
  byKey: PlayerAction(
    filter: {
      player_id: {_eq: "bae-74922412-523d-556a-897b-3af6fc47701c"}, 
      action: {key: {_in: ["GITHUB_STAR"]}}, 
      deleted: {_ne: true}}
  ) {
    action {
      key
      _docID
    }
  }
}
```
Results:
```json
{
  "data": {
    "byId": [
      {
        "action": {
          "_docID": "bae-d6e62c28-c064-5e96-8404-de8facc1f401",
          "key": "GITHUB_STAR"
        }
      },
      {
        "action": {
          "_docID": "bae-d6e62c28-c064-5e96-8404-de8facc1f401",
          "key": "GITHUB_STAR"
        }
      },
      {
        "action": {
          "_docID": "bae-d6e62c28-c064-5e96-8404-de8facc1f401",
          "key": "GITHUB_STAR"
        }
      },
      {
        "action": {
          "_docID": "bae-d6e62c28-c064-5e96-8404-de8facc1f401",
[...]

PR #3035, opened 2024-09-19
fix: Queries with filter on 2 rel fields of composite index

### LINKED ISSUE #3032: Secondary indexes on relations cause different documents to be returned depending on joins included
Schema:
```gql
type ActionRequirement @index(unique: true, includes: [{ name: "action_id" }, { name: "reward_id" }], name: "actionId_rewardId") {
  count: Int
  isPreReq: Boolean 
  action: Action @relation(name: "actionActionRequirements")
  reward: Reward @relation(name: "rewardActionRequirements")
  createdAt: DateTime 
  oldId: String
}

type Reward {
  name: String 
  description: String
  value: Int
  limit: Int
  tiers: [RewardTier] @relation(name: "rewardRewardTiers")
  priority: Int
  enabled: Boolean 
  actionRequirements: [ActionRequirement] @relation(name: "rewardActionRequirements")
  playerRewards: [PlayerReward] @relation(name: "rewardPlayerRewards")
  rewardGroup: [String] 
  createdAt: DateTime
  oldId: String
}

type RewardTier @index(unique: true, includes: [{ name: "level" }, { name: "reward_id" }], name: "level_rewardId") {
  level: Int
  name: String
  description: String
  value: Int
  reward: Reward @index @primary @relation(name: "rewardRewardTiers") 
  playerReward: [PlayerReward] @relation(name: "playerRewardRewardTiers")
  oldId: String
}

type Action {
  key: String 
  name: String 
  description: String 
  enabled: Boolean 
  levelRequired: Int 
  actionRequirements: [ActionRequirement] @relation(name: "actionActionRequirements")
  playerActions: [PlayerAction] @relation(name: "actionPlayerActions")
  createdAt: DateTime 
  oldId: String
}
```
Queries:
```gql
query queryRewardForLogPlayerAction {
  withTier: Reward(
    filter: {actionRequirements: {action_id: {_in: ["bae-4a72e698-51f8-5f83-ac96-7dee17e7f1a6"]}}}
  ) {
    id: _docID
    actionRequirements {
      id: _docID
    }
    tiers {
      _docID
    }
  }
  withoutTier: Reward(
    filter: {actionRequirements: {action_id: {_in: ["bae-4a72e698-51f8-5f83-ac96-7dee17e7f1a6"]}}}
  ) {
    actionRequirements {
      id: _docID
    }
  }
  withTierNoActionReq: Reward(
    filter: {actionRequirements: {action_id: {_in: ["bae-4a72e698-51f8-5f83-ac96-7dee17e7f1a6"]}}}
  ) {
    id: _docID
    tiers {
      _docID
    }
  }
}
```
All the queries return different sets of data!  Removing the all the indexes from the schema results in all queries returning the correct (and same) results.

Preliminary testing suggests that the problem is with `type ActionRequirement @index(unique: true, includes: [{ name: "action_id" }, { name: "reward_id" }], name: "actionId_rewardId") {` and that removing all but that index results in the same (incorrect) 3 sets of differing results.

islamaliev:
## Relevant issue(s)

Resolves #3032 #2928

## Description

Make parallel node perform a real merge instead of copying fields of a fetched doc, which resulted every next doc overwrite previous fields.


internal/planner/multi.go
```diff
@@ -136,12 +136,22 @@ func (p *parallelNode) nextMerge(_ int, plan planNode) (bool, error) {
...
 
 	return true, nil
 }
 
+func (p *parallelNode) mergeDoc(doc core.Doc, newFields core.DocFields) core.Doc {
+	for i := range newFields {
+		if doc.Fields[i] == nil {
+			doc.Fields[i] = newFields[i]
+		} else if newSubDoc, ok := newFields[i].(core.Doc); ok {
+			doc.Fields[i] = p.mergeDoc(doc.Fields[i].(core.Doc), newSubDoc.Fields)
```
AndrewSisley, line 149: todo: Please document why this is needed, it is not terribly clear why the node should be mutating child relations.

tests/integration/index/query_with_unique_index_on_relation_filter_test.go
```diff
@@ -70,3 +70,106 @@ func TestQueryWithUniqueCompositeIndex_WithFilterOnIndexedRelation_ShouldFilter(
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQueryWithUniqueCompositeIndex_WithIndexComprising2RelationsAndFilterOnIt_ShouldFilter(t *testing.T) {
```
AndrewSisley, line 74: suggestion: It might be highlighting with a comment why this test is important to have, for example compared to just for composite scalar tests.

Please also document the significance of including two id fields, and one object field in the index.  Perhaps highlighting why a test with two object fields in the index (e.g. `includes: [{name: "owner"}, {name: "manufacturer"}` is not required.

AndrewSisley (changes requested):
I trust it works, but please explain why the the new code does what it does before merge.

internal/planner/multi.go
islamaliev: well, it now justifies the name of method `nextMerge`.

I'll add a comment.

tests/integration/index/query_with_unique_index_on_relation_filter_test.go
islamaliev: The test case is actually a bit simpler. I wanted to update it, but forgot. Thanks for reminding.

I will add some comments.

internal/planner/multi.go
AndrewSisley: I'm not sure the new comment answers what I'm most uncertain about - why does this function recursively mutate child documents, instead of just the top doc?  Why and under which circumstances is the mutation of the children required?  Do please message/call me on discord if you unsure as to what I'm asking.

tests/integration/index/query_with_unique_index_on_relation_filter_test.go
AndrewSisley: Thanks Islam, looks good!

AndrewSisley (approved):
LGTM, thanks Islam - please try and tweak the documentation for `mergeDoc` a little bit before merge though.

internal/planner/multi.go
islamaliev: there is no real reason for recursion here. I just though it would cover some future perverted edge-cases.
I made it merge just top fields to keep it simple.

AndrewSisley: Ah got it, thanks Islam!

PR #3040, opened 2024-09-19
fix: Make GraphQL errors spec compliant

### LINKED ISSUE #3039: GraphQL response errors are not spec compliant
See https://spec.graphql.org/October2021/#sec-Errors

### LINKED ISSUE #3041: client GQLResult is not spec compliant
The `client.GQLResult` type does not define errors in a spec compliant way.

See https://spec.graphql.org/October2021/#sec-Errors

nasdf:
## Relevant issue(s)

Resolves #3039
Resolves #3041

## Description

This PR fixes an issue with response errors not being spec compliant.

It also adds a `GQLError` type to the client package.

## Tasks

## How has this been tested?

Added tests here https://github.com/sourcenetwork/defradb-third-party-test

- MacOS

### DIFF
--- a/client/db.go
+++ b/client/db.go
@@ -11,6 +11,7 @@
 package client
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 
@@ -318,8 +319,10 @@ type gqlResult struct {
 }
 
 func (res *GQLResult) UnmarshalJSON(data []byte) error {
+	dec := json.NewDecoder(bytes.NewBuffer(data))
+	dec.UseNumber()
 	var out gqlResult
-	if err := json.Unmarshal(data, &out); err != nil {
+	if err := dec.Decode(&out); err != nil {
 		return err
 	}
 	res.Data = out.Data
--- a/client/db.go
+++ b/client/db.go
@@ -12,6 +12,7 @@ package client
 
 import (
 	"context"
+	"encoding/json"
 
 	ds "github.com/ipfs/go-datastore"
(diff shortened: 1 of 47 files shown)


fredcarle (approved):
LGTM

client/db.go
```diff
@@ -290,14 +296,33 @@ type GQLResult struct {
...
 	// If there are values in this slice the request will likely not have run to completion
 	// and [Data] will be nil.
-	Errors []error `json:"errors,omitempty"`
+	Errors []GQLError `json:"errors,omitempty"`
 
 	// Data contains the resultant data produced by the GQL request.
 	//
 	// It will be nil if any errors were raised during execution.
 	Data any `json:"data"`
 }
```
AndrewSisley, line 305: suggestion:  If you add a Marshal/Unmarshal func for `GQLResult` you could probably drop the new types and functions.  It might also be nicer for users of the Go client.

fredcarle: I would just add a custom Marshal/Unmarshal func for `GQLError` instead .

nasdf: > I would just add a custom Marshal/Unmarshal func for GQLError instead .

I think the custom marshalling only works if its on the GQLResult and we want to keep errors as the generic `error` type.

I've refactored it to use a custom marshaller and added some tests to make sure our http server returns proper responses.

fredcarle (approved):
LGTM. The Marshal/Unmarshal change is nice.

PR #3044, opened 2024-09-20
refactor: GraphQL order input

### LINKED ISSUE #3043: GraphQL order argument does not work with variables

nasdf:
## Relevant issue(s)

Resolves #3043

## Description

This PR fixes an issue where GQL variables could not be used with ordering input args. The order input type on collections and aggregates has been refactored to a list so that ordering is preserved in all environments.

## Tasks

## How has this been tested?

Added integration tests

- MacOS


internal/request/graphql/parser/order.go
```diff
@@ -0,0 +1,90 @@
...
+			return nil, err
+		}
+		if cond == nil {
+			return nil, nil
+		}
+		// prepend the current field name, to the parsed condition from the slice
+		// Eg. order: {author: {name: ASC, birthday: DESC}}
+		// This results in an array of [name, birthday] converted to
+		// [author.name, author.birthday].
+		// etc.
```
fredcarle, line 69: thought: The order object in this comment seems invalid as it's an unordered map intending for name and birthday to be in an explicit order.

fredcarle: I think it would have to be `order: [{author: {name: ASC}}, {author: {birthday: DESC}}]`

internal/request/graphql/parser/order.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	return conditions, nil
+}
+
+func parseOrderCondition(arg map[string]any) (*request.OrderCondition, error) {
+	if len(arg) == 0 {
+		return nil, nil
+	}
+	if len(arg) != 1 {
```
fredcarle, line 39: todo: Please add a test for the case where we have more than one argument

internal/request/graphql/parser/order.go
```diff
@@ -0,0 +1,90 @@
...
+
+func parseOrderDirection(v int) (request.OrderDirection, error) {
+	switch v {
+	case 0:
+		return request.ASC, nil
+
+	case 1:
+		return request.DESC, nil
+
+	default:
```
fredcarle, line 87: todo: please add a test that satisfies this.

fredcarle (approved):
LGTM. Just a couple test todos and a comment to fix before merge.

internal/request/graphql/parser/order.go
nasdf: done

internal/request/graphql/parser/order.go
nasdf: This case is technically unreachable because the GQL parsing will fail. I've added a test to demonstrate this. I felt it was better to keep the default case instead of having it undefined, but I can remove it if you have a different preference.

shahzadlone (approved):
LGTM :)

PR #3061, opened 2024-09-24
refactor: Change from protobuf to cbor for gRPC

### LINKED ISSUE #3057: Change net package data protocol from protobuf to cbor.
This is to simplify dependencies for browser deployments and others that are compiling to wasm.

fredcarle:
## Relevant issue(s)

Resolves #3057

## Description

This PR makes the net gRPC server use cbor instead of protobuf. By doing this we remove the protobuf overhead and the related dependencies which will make compilation to wasm simpler and simplify browser support.

## Tasks

## How has this been tested?

make test

- MacOS


internal/kms/pubsub.go
```diff
@@ -194,17 +198,23 @@ func (s *pubSubService) requestEncryptionKeyFromPeers(
 	return nil
 }
 
+type fetchEncryptionKeyReply struct {
+	Links              [][]byte
+	Blocks             [][]byte
+	EphemeralPublicKey []byte
+}
```
shahzadlone, line 205: nitpick/question: Why `fetchEncryptionKeyReply` and not `fetchEncryptionKeyResponse`?

net/grpc.go
```diff
@@ -0,0 +1,103 @@
...
+	interceptor grpc.UnaryServerInterceptor,
+) (any, error) {
+	in := new(pushLogRequest)
+	if err := dec(in); err != nil {
+		return nil, err
+	}
+	if interceptor == nil {
+		return srv.(serviceServer).PushLog(ctx, in)
+	}
+	info := &grpc.UnaryServerInfo{
```
shahzadlone, line 79: question: Everything below and including this line is never hit with any tests? i.e. the `info`, the `handler` and the `interceptor` return stuff.

net/grpc.go
```diff
@@ -0,0 +1,103 @@
...
+package net
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+)
+
+const (
+	serviceGetDocGraphName  = "/defradb.net.Service/GetDocGraph"
```
shahzadlone, line 20: suggestion: Maybe define a string const for `"defradb.net.Service"` and use it here and below.

net/grpc.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	handler := func(ctx context.Context, req any) (any, error) {
+		return srv.(serviceServer).PushLog(ctx, req.(*pushLogRequest))
+	}
+	return interceptor(ctx, in, info, handler)
+}
+
+func registerServiceServer(s grpc.ServiceRegistrar, srv serviceServer) {
+	desc := &grpc.ServiceDesc{
+		ServiceName: "defradb.net.Service",
```
shahzadlone, line 91: suggestion: If accepted the above suggestion can use the string const var here as well.

internal/kms/pubsub.go
fredcarle: This is just copy pasted from what we had before. It's also how google names the response parameter (reply).

net/grpc.go
fredcarle: This is to satisfy some gRPC logic. I've based it off what was used with protobufs. I'm not sure how to test this at the moment and since it's copy pasted from the previously generated code, I would rather not spend too much time on it for this PR. I'm adding the head exchange functionality next so I'll look into testing that part at the same time.

internal/kms/pubsub.go
```diff
@@ -127,10 +126,15 @@ func (s *pubSubService) handleKeyRequestedEvent() {
...
+type fetchEncryptionKeyRequest struct {
+	Links              [][]byte
+	EphemeralPublicKey []byte
+}
+
 // handleEncryptionMessage handles incoming FetchEncryptionKeyRequest messages from the pubsub network.
 func (s *pubSubService) handleRequestFromPeer(peerID libpeer.ID, topic string, msg []byte) ([]byte, error) {
-	req := new(pb.FetchEncryptionKeyRequest)
-	if err := proto.Unmarshal(msg, req); err != nil {
+	req := new(fetchEncryptionKeyRequest)
```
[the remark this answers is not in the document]
fredcarle: You know that is your code right 😉

nasdf (approved):
LGTM

net/grpc.go
fredcarle: done

net/grpc.go
fredcarle: done

Issue #3067, opened 2024-09-25
Backup import and export for all nodes or just first


shahzadlone:
The backup import and export tests have a copy pasted comment line:

`// If a value is not provided the indexes will be retrieved from the first nodes.`

Which is confusing to know if it supports backup import and export for all nodes per-command or only the target node.

The implementation in the testing logic assumes it works on all nodes, but I am not certain it does, so marking a todo linking this issue.

Test framework test cases:

```go
// BackupExport will attempt to export data from the datastore using the db api.
type BackupExport struct {
	// NodeID may hold the ID (index) of a node to generate the backup from.
	//
	// If a value is not provided the indexes will be retrieved from the first nodes.
	NodeID immutable.Option[int]

	// The backup configuration.
	Config client.BackupConfig

	// Content expected to be found in the backup file.
	ExpectedContent string

	// Any error expected from the action. Optional.
	//
	// String can be a partial, and the test will pass if an error is returned that
	// contains this string.
	ExpectedError string
}

// BackupExport will attempt to export data from the datastore using the db api.
type BackupImport struct {
	// NodeID may hold the ID (index) of a node to generate the backup from.
	//
	// If a value is not provided the indexes will be retrieved from the first nodes.
	NodeID immutable.Option[int]

	// The backup file path.
	Filepath string

	// The backup file content.
	ImportContent string

	// Any error expected from the action. Optional.
	//
	// String can be a partial, and the test will pass if an error is returned that
	// contains this string.
	ExpectedError string
}

```

PR #3068, opened 2024-09-25
fix(i): Sort out invalid testing framework node indexing

### LINKED ISSUE #3065: bug: consecutive source-hub acp operations fail for http client due to an identity auth error
While implementing #2762 in #2907 I came across a bug where calling a sourcehub operation on two different nodes failed i.e. attempting to share a document by the same owner with a 2nd owner on each 2 different nodes causes the first node's operation to succeed and the second nodes operation to fail.

My hunch was that this is not specific to the relationship sharing feature, as this is only an issue for http client with sourcehub.

I tested a basic example of creating 2 private documents using the same identity (which uses acp identity operations underneath ofc.), one on node 1 and second on node 2 using sourcehub with http client. As suspected the tests fail, hence this is an existing bug.

Here is a test that documents the failure:

```go
func TestDEBUG(t *testing.T) {
	expectedPolicyID := "fc56b7509c20ac8ce682b3b9b4fdaad868a9c70dda6ec16720298be64f16e9a4"

	test := testUtils.TestCase{

		Description: "Test source-hub acp with http, consecutive operations on different nodes, identity bug",

		SupportedACPTypes: immutable.Some(
			[]testUtils.ACPType{
				testUtils.SourceHubACPType,
			},
		),

		Actions: []any{

			testUtils.AddPolicy{

				Identity: immutable.Some(1),

				Policy: `
                    name: Test Policy

                    description: A Policy

                    actor:
                      name: actor

                    resources:
                      users:
                        permissions:
                          read:
                            expr: owner + reader + writer

                          write:
                            expr: owner + writer

                          nothing:
                            expr: dummy

                        relations:
                          owner:
                            types:
                              - actor

                          reader:
                            types:
                              - actor

                          writer:
                            types:
                              - actor

                          admin:
                            manages:
                              - reader
                            types:
                              - actor

                          dummy:
                            types:
                              - actor
                `,

				ExpectedPolicyID: expectedPolicyID,
			},

			testUtils.SchemaUpdate{
				Schema: fmt.Sprintf(`
						type Users @policy(
							id: "%s",
							resource: "users"
						) {
							name: String
							age: Int
						}
					`,
					expectedPolicyID,
				),
			},

			testUtils.CreateDoc{
				Identity: immutable.Some(1),

				NodeID: immutable.Some(0),

				CollectionID: 0,

				DocMap: map[string]any{
					"name": "Shahzad",
				},
			},

			testUtils.CreateDoc{
				Identity: immutable.Some(1),

				NodeID: immutable.Some(1),

				CollectionID: 0,

				DocMap: map[string]any{
					"name": "Shahzad Lone",
				},
			},
		},
	}

}

```

Output:

```sh
Error: Received unexpected error: 403: forbidden
```

shahzadlone:
## Relevant issue(s)
Resolves #3065 

## Description
The main bug was only visible on sourcehub acp using http, due to the identity being copied with the audience value of another node's host failing authentication (the bearer tokens should be unique using correct node's audience). The biggest issue was the way we use `getNodes` and `getNodeCollections`. I would be in favor of completely removing them as they are more troublesome than the utility they provide.

- First commit documents the bug
- Some utility functions were overwriting and producing the wrong node index
- Forbidden bug happening on sourcehub<>http test run: https://github.com/sourcenetwork/defradb/actions/runs/10930293192/job/30342883535?pr=2907

### Future:
- Resolve import/export documentation and implementation if different (#3067)
- Should likely clean this test utils up and make helper methods to avoid code duplication (https://github.com/sourcenetwork/defradb/issues/3069)
- Likely should remove all usages of `getNodeCollections`  (https://github.com/sourcenetwork/defradb/issues/3069)
- Likely should remove all usages of `getNodes` (https://github.com/sourcenetwork/defradb/issues/3069)

## How has this been tested?
- Very painfully haha, had to install Linux bare-metal to investigate the first bug (sourcehub doesn't build on wsl for me) that was only occurring on sourcehub acp using http, due to the identity being copied with the audience value of another node the way we use `getNodes` and `getNodeCollections`

### DIFF
--- a/tests/integration/acp/p2p/create_test.go
+++ b/tests/integration/acp/p2p/create_test.go
@@ -123,8 +123,6 @@ func TestACP_P2PCreatePrivateDocumentsOnDifferentNodes_SourceHubACP(t *testing.T
 				DocMap: map[string]any{
 					"name": "Shahzad Lone",
 				},
-
-				ExpectedError: "403: forbidden", // TODO: FIX THIS BUG
 			},
 		},
 	}
--- a/tests/integration/acp/p2p/create_test.go
+++ b/tests/integration/acp/p2p/create_test.go
@@ -1,0 +1,133 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package test_acp_p2p
+
+import (
+	"fmt"
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/immutable"
+)
+
+func TestACP_P2PCreatePrivateDocumentsOnDifferentNodes_SourceHubACP(t *testing.T) {
+	expectedPolicyID := "fc56b7509c20ac8ce682b3b9b4fdaad868a9c70dda6ec16720298be64f16e9a4"
+
+	test := testUtils.TestCase{
+
+		Description: "Test acp, p2p create private documents on different nodes, with source-hub",
+
+		SupportedACPTypes: immutable.Some(
+			[]testUtils.ACPType{
+				testUtils.SourceHubACPType,
+			},
+		),
+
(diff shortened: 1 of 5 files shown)


nasdf (approved):
LGTM. Thanks for documenting everything so well. I'm also in favor of removing the `getNodes` utilities.

PR #3077, opened 2024-09-26
fix(i): Validate nearby relation fields in SchemaPatch

### LINKED ISSUE #3074: Relation fields on different objects not in patch are mutating with PatchSchema (and erroring)
e.g.
```
type Contact {
  email: String
  player: [Player] @relation(name: "playerContact")
} 

type Player {
  authId: String
  characterName: String
  email: String
  refCode: String
  contact: Contact @primary @relation(name: "playerContact")
}

http://localhost:9181/api/v0/schema
{"Patch":"[{\"op\":\"add\",\"path\":\"/Player/Fields/-\",\"value\":{\"Name\":\"test1\",\"Kind\":11}}]","SetAsDefaultVersion":true}
Error: mutating an existing field is not supported. ProposedName: contact
```

AndrewSisley:
## Relevant issue(s)

Resolves #3074

## Description

Correctly handles validation of nearby relation fields in schema patch.

Previously the equality check failed to account for `Kind` being a pointer and thus always flagged relation fields as having mutated.  This was likely introduced in https://github.com/sourcenetwork/defradb/pull/2961 when `Kind` became a pointer.

Collections referenced by relation fields were also not included in the validation set, causing the rules to think that the related object did not exist.

### DIFF
--- a/internal/db/definition_validation.go
+++ b/internal/db/definition_validation.go
@@ -818,7 +818,9 @@ func validateFieldNotMutated(
 
 		for _, newField := range newSchema.Fields {
 			oldField, exists := oldFieldsByName[newField.Name]
-			if exists && oldField != newField {
+
+			// DeepEqual is temporary, as this validation is temporary
+			if exists && !reflect.DeepEqual(oldField, newField) {
 				return NewErrCannotMutateField(newField.Name)
 			}
 		}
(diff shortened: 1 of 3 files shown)


nasdf (approved):
LGTM

shahzadlone (approved):
LGTM

PR #3078, opened 2024-09-27
feat: Min and max numerical aggregates

### LINKED ISSUE #2978: Support for `_min` and `_max` aggregates
We should add an aggregate for min and max values from an array of numerical values.

nasdf:
## Relevant issue(s)

Resolves #2978

## Description

This PR adds support for two new query aggregates `min` and `max`.

## Tasks

## How has this been tested?

Added integration tests

- MacOS


tests/integration/query/simple/with_group_average_max_test.go
```diff
@@ -0,0 +1,168 @@
...
+				},
+			},
+		},
+	}
+
+	executeTestCase(t, test)
+}
+
+// Note: this test should follow a different code path to `_avg` on it's own
+// utilising the existing `_min` node instead of adding a new one.  This test cannot
```
fredcarle, line 115: typo: `_max`

AndrewSisley (changes requested):
I want too review this but will wait until the tests are passing, please re-request my review when they are and I'll get right on this :)

tests/integration/query/simple/with_group_average_max_test.go
nasdf: fixed

nasdf:
> I want too review this but will wait until the tests are passing, please re-request my review when they are and I'll get right on this :)

Should be all fixed now. There was an issue with casting max value floats to ints which I think may still exist in the sum aggregate where I copied the code from.

AndrewSisley (changes requested):
Looks good to me, just some minor stuff to sort out first IMO, and the below question.

question: It does not lok like you have added max/min to non-numeric types such as `DateTime`, `String`, `Boolean`, etc, and the gql related code does not look like it will facilitate the introduction of that support - was this an oversight or do you have a plan on how to introduce it later?  I am concerned that doing so will be a breaking change to the GQL syntax due to the return type.

fredcarle (approved):
Nice work! I like what you did with the aggregate generics and thanks for adding all those tests. Just need to resole Andy's comments before merge.

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+		aggregateMapping:  field.AggregateTargets,
+		virtualFieldIndex: field.Index,
+		docMapper:         docMapper{field.DocumentMapping},
+	}, nil
+}
+
+func (n *maxNode) Kind() string           { return "maxNode" }
+func (n *maxNode) Init() error            { return n.plan.Init() }
+func (n *maxNode) Start() error           { return n.plan.Start() }
+func (n *maxNode) Spans(spans core.Spans) { n.plan.Spans(spans) }
```
shahzadlone, line 71: suggestion: Would really like this to be hit with the tests

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+			// get the target aggregate document mapping. Since the filters
+			// are relative to the target aggregate collection (and doc mapper).
+			var targetMap *core.DocumentMapping
+			if source.Index < len(n.documentMapping.ChildMappings) &&
+				n.documentMapping.ChildMappings[source.Index] != nil {
+				targetMap = n.documentMapping.ChildMappings[source.Index]
+			} else {
+				targetMap = n.documentMapping
+			}
+			simpleExplainMap[filterLabel] = source.Filter.ToMap(targetMap)
```
shahzadlone, line 95: todo: These two lines are not hit in any tests, if easy to do so, pelase add

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+		// Add the main field name.
+		simpleExplainMap[fieldNameLabel] = source.Field.Name
+
+		// Add the child field name if it exists.
+		if source.ChildTarget.HasValue {
+			simpleExplainMap[childFieldNameLabel] = source.ChildTarget.Name
```
shahzadlone, line 103: suggestion: same comment as above.

internal/planner/min.go
```diff
@@ -0,0 +1,233 @@
...
+			// get the target aggregate document mapping. Since the filters
+			// are relative to the target aggregate collection (and doc mapper).
+			var targetMap *core.DocumentMapping
+			if source.Index < len(n.documentMapping.ChildMappings) &&
+				n.documentMapping.ChildMappings[source.Index] != nil {
+				targetMap = n.documentMapping.ChildMappings[source.Index]
+			} else {
+				targetMap = n.documentMapping
+			}
+			simpleExplainMap[filterLabel] = source.Filter.ToMap(targetMap)
```
shahzadlone, line 95: suggestion: Similar comment as made on `max` explain.

internal/planner/min.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+		// Add the main field name.
+		simpleExplainMap[fieldNameLabel] = source.Field.Name
+
+		// Add the child field name if it exists.
+		if source.ChildTarget.HasValue {
+			simpleExplainMap[childFieldNameLabel] = source.ChildTarget.Name
```
shahzadlone, line 103: suggestion: similar comment as `max` explain

tests/integration/explain.go
```diff
@@ -41,6 +41,8 @@ var (
 		"deleteNode":    {},
 		"groupNode":     {},
 		"limitNode":     {},
+		"maxNode":       {},
+		"minNode":       {},
```
shahzadlone, line 45: praise: :)

tests/integration/explain/debug/top_with_max_test.go
```diff
@@ -0,0 +1,96 @@
...
+					{
+						"selectTopNode": dataMap{
+							"selectNode": dataMap{
+								"scanNode": dataMap{},
+							},
+						},
+					},
+					{
+						"maxNode": dataMap{},
+					},
```
shahzadlone, line 34: question: This is not under the `selectTopNode`? Is a sibling to `selectTopNode` might be worth documenting that at planner level if not already (maybe I missed it).

tests/integration/explain/debug/with_max_join_test.go
```diff
@@ -0,0 +1,151 @@
...
+		"operationNode": []dataMap{
+			{
+				"selectTopNode": dataMap{
+					"maxNode": dataMap{
+						"selectNode": dataMap{
+							"typeIndexJoin": dataMap{
+								"typeJoinMany": normalTypeJoinPattern,
+							},
+						},
+					},
```
shahzadlone, line 31: question: here `maxNode` is under `selectTopNode` in the `debug` explain but not in the example above ? Why is that?

internal/planner/aggregate.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if !b.HasValue() {
+		return false
+	}
+
+	return a.Value() < b.Value()
+}
+
+func reverse[T any](original func(T, T) bool) func(T, T) bool {
```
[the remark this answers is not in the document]
nasdf: done

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+	field *mapper.Aggregate,
+	parent *mapper.Select,
+) (*maxNode, error) {
+	isFloat := false
+	for _, target := range field.AggregateTargets {
+		isTargetFloat, err := p.isValueFloat(parent, &target)
+		if err != nil {
+			return nil, err
+		}
+		// If one source property is a float, the result will be a float - no need to check the rest
```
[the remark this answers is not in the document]
nasdf: I'm not seeing which part is incorrect.

internal/planner/max.go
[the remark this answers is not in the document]
nasdf: added a test that should cover it

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+	hasNext, err := n.plan.Next()
+	if err != nil || !hasNext {
+		return hasNext, err
+	}
+	n.currentValue = n.plan.Value()
+
+	max := -math.MaxFloat64
+	for _, source := range n.aggregateMapping {
+		child := n.currentValue.Fields[source.Index]
+		collectionMax := -math.MaxFloat64
```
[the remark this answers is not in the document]
nasdf: refactored to return nil on empty values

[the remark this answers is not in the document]
nasdf: refactored, this is no longer applicable

internal/planner/max.go
```diff
@@ -0,0 +1,233 @@
...
+
+type maxNode struct {
+	documentIterator
+	docMapper
+
+	p    *Planner
+	plan planNode
+
+	isFloat           bool
+	virtualFieldIndex int
```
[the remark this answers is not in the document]
nasdf: done

internal/planner/max.go
nasdf: should be covered now. I noticed that every other aggregate is also missing this coverage

tests/integration/explain/debug/with_max_join_test.go
nasdf: They are testing two different cases. This one is testing an aggregate from within a select and the other is testing an aggregate from the top level.

internal/planner/max.go
```diff
@@ -0,0 +1,261 @@
...
+	hasNext, err := n.plan.Next()
+	if err != nil || !hasNext {
+		return hasNext, err
+	}
+	n.currentValue = n.plan.Value()
+
+	var max *float64
+	for _, source := range n.aggregateMapping {
+		child := n.currentValue.Fields[source.Index]
+		var collectionMax *float64
```
AndrewSisley, line 147: todo: This is incorrect for large integers, you cannot convert a very big int to a float and then back again and expect it to have the same value.

suggestion: Add failing tests before fixing this (if you can, you might struggle to set an Int to Int64MaxVal atm via our public clients, as GQL is technically only 32bit Ints)

tests/integration/explain/debug/top_with_max_test.go
nasdf: In this test case it's used from the top level.

internal/planner/max.go
AndrewSisley: e.g. the below will result in a negative number:
```
a := math.MaxInt64 - 1
		println(fmt.Sprint(a))
		b := float64(a)
		c := int64(b)
		panic(fmt.Sprint(c))
```

nasdf: refactored to use `big.Float`

internal/planner/max.go
nasdf: should be covered now

internal/planner/max.go
nasdf: should also be covered now

internal/planner/min.go
nasdf: should also be covered now

internal/planner/min.go
nasdf: should also be covered now

nasdf:
> Which aggregate types? The new types added in this PR? Or some existing stuff to?

I haven't thought through all of it yet, but because the aggregates are expected to return a float and can then be used as input into other aggregates, there will be issues with an aggregate that can return a union type of String, DateTime, etc.
 
> I am concerned that you will just end up re-writing a lot the new stuff added in this PR to handle (IMO) a pretty core sub-feature (DateTime).

The plan node code is very generic. I'm not too worried about this.

shahzadlone (approved):
LGTM thanks for doing the suggestions

internal/planner/max.go
AndrewSisley: did you try to test this?

AndrewSisley:
> there will be issues with an aggregate that can return a union type of String, DateTime, etc.

Yes, that is why I am concerned you are painting the syntax into a corner and we will need to make an avoidable breaking change in the (possibly near) future in order to correct this problem.  The breaking change will require us to change our tests, and users to change their code.

I do not think this is a good thing to do, but will not block the PR over it.  Please resolve the cast problem/test request before merge though.

internal/planner/max.go
nasdf: I've added tests for int max value

internal/planner/max.go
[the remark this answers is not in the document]
nasdf: Okay I see what you mean. I've refactored the logic so it actually returns the correct type now. 

Correct me if I'm wrong but I don't think we have any operations that would allow us to test mixed type aggregates.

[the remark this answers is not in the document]
AndrewSisley: > Correct me if I'm wrong but I don't think we have any operations that would allow us to test mixed type aggregates.

No you are correct, I'm only commenting on the comment, the code/tests are all good IMO :)

> I've refactored the logic so it actually returns the correct type now.

Nice, thanks!

nasdf:
> > there will be issues with an aggregate that can return a union type of String, DateTime, etc.
> 
> Yes, that is why I am concerned you are painting the syntax into a corner and we will need to make an avoidable breaking change in the (possibly near) future in order to correct this problem. The breaking change will require us to change our tests, and users to change their code.
> 
> I do not think this is a good thing to do, but will not block the PR over it. Please resolve the cast problem/test request before merge though.

The GQL type system won't allow for an aggregate that returns a union to be used as input in another aggregate.

I think the best approach will be to add new aggregates: `_minDateTime`, `_maxString`, etc.

This shouldn't result in any breaking changes to the existing aggregates.

AndrewSisley:
> I think the best approach will be to add new aggregates: _minDateTime, _maxString, etc.
>
> This shouldn't result in any breaking changes to the existing aggregates.

Only if we dont consider it weird that we have `_min` and not `_minFloat`, if that is the long term syntax we want I would prefer that all the types be consistent.

AndrewSisley (approved):
Talked over the long term plan in discord, is fully compatible with this PR - thanks for all of this Keenan, and for the discussion :)

islamaliev:
bug bash result: all looks good.

PR #3080, opened 2024-09-27
fix(i): Null compound filter panic

### LINKED ISSUE #3079: Null compound filter causes panic
Null compound filters and most likely null elements cause a panic.

Example: 

```json
{
  "filter": {
    "_or": null
  }
}
```

Results:

```
panic: interface conversion: interface {} is nil, not []interface {}
 
 -> github.com/sourcenetwork/defradb/internal/planner/mapper.resolveInnerFilterDependencies
 ->   github.com/sourcenetwork/defradb/internal/planner/mapper/mapper.go:969
```

nasdf:
## Relevant issue(s)

Resolves #3079

## Description

This PR fixes an issue where null values within compound filters would panic.

## Tasks

## How has this been tested?

Added integration tests

- MacOS

### DIFF
--- a/internal/planner/mapper/mapper.go
+++ b/internal/planner/mapper/mapper.go
@@ -966,7 +966,10 @@ func resolveInnerFilterDependencies(
 
 	for key := range source {
 		if key == request.FilterOpAnd || key == request.FilterOpOr {
-			compoundFilter := source[key].([]any)
+			compoundFilter, ok := source[key].([]any)
+			if !ok {
+				continue // filter is nil
+			}
 			for _, innerFilter := range compoundFilter {
 				innerFields, err := resolveInnerFilterDependencies(
 					ctx,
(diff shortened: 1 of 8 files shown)


internal/planner/mapper/mapper.go
```diff
@@ -966,7 +966,10 @@ func resolveInnerFilterDependencies(
 
 	for key := range source {
 		if key == request.FilterOpAnd || key == request.FilterOpOr {
-			compoundFilter := source[key].([]any)
+			compoundFilter, ok := source[key].([]any)
+			if !ok {
+				continue // filter is nil
```
AndrewSisley, line 971: suggestion: I do think `if foo == nil` conveys intent much better than a cast+comment

AndrewSisley (approved):
LGTM, but I really think you should be checking for nil with an 'if nil' check instead of a failed cast.  Might be worth raising with the team, as I've mentioned this before to you and I don't want to be pushing you away from it if it is only me who thinks this is a bad way of handling it.

internal/planner/mapper/mapper.go
fredcarle: I don't mind the type assertion here as it does 2 things. It ensures it's an array and not nil at the same time.

AndrewSisley: I see that as part of the problem, it shouldn't be checking the type here, and it is not really handling it if it is somehow the wrong type - it is just silently ignoring what could be a serious coding error.

PR #3091, opened 2024-09-30
ci(i): Fix vulnerabilities scan by ignoring x/crises

### LINKED ISSUE #3090: Fix ci vulncheck failures due to x/crisis
Fix ci vulncheck failures due to x/crisis (will not be fixed by maintainers)

We started having vulncheck scanner only report `x/crisis` failures here: https://github.com/sourcenetwork/defradb/actions/runs/9965792152/job/27536721672?pr=2842

Note: Failures started happening before too, but there were other fixable vulnerabilities then too which we did want to see (so the failures before were desirable).

A rough fix was done here:

branch: https://github.com/shahzadlone/defradb/tree/lone/fix-check-vul
commit: https://github.com/shahzadlone/defradb/commit/d57da7091258ef726e62bd18c5b5ebdd6a1bfe6e

Which just asserts that "Your code is affected by 2 vulnerabilities from 1 module." is what the scan reports (it's remained consistent for the last two months). If any new vulnerability is introduced the assert will fail and we will be notified.

shahzadlone:
## Relevant issue(s)
Resolves #3090

## Description
- Hacky fix for the vulnerability scanner until they improve the tool (or we solve the vulnerability).
- Make the vulnerability scan fail if new vulnerabilities are introduced and pass if previous known ones remain.

## How has this been tested?
- Tried introducing a vul and it works and ignores the current `x/crisis` one
- Run: https://github.com/sourcenetwork/defradb/actions/runs/11110357130/job/30867868823?pr=3091

### DIFF
--- a/.github/workflows/check-vulnerabilities.yml
+++ b/.github/workflows/check-vulnerabilities.yml
@@ -30,10 +30,32 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Run govulncheck
-        uses: golang/govulncheck-action@v1
+      - name: Checkout code into the directory
+        uses: actions/checkout@v4
+
+      - name: Setup Go environment explicitly
+        uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
           check-latest: true
           cache: false
-          go-package: ./...
+
+      - name: Install govulncheck
+        run: make deps:vulncheck
+
+      - name: Run govulncheck scan
+        run: govulncheck -C . -format text ./... | tee govulncheck.txt
+
+      - name: Check if only known vulnerabilities were found (there are new vulnerabilities if this fails)
+        run: cat govulncheck.txt | grep "Your code is affected by 2 vulnerabilities from 1 module."
+
+    # Use the steps below once the x/crisis (crisis.init) bug is fixed or if the
+    # ability to silence is implemented: https://github.com/golang/go/issues/61211
+    #steps:
+    # - name: Run govulncheck
+    #   uses: golang/govulncheck-action@v1
+    #   with:
+    #     go-version-file: 'go.mod'
+    #     check-latest: true
+    #     cache: false
+    #     go-package: ./...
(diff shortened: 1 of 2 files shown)


AndrewSisley:
> Make the vulnerability scan fail if new vulnerabilities are introduced and pass if previous known ones remain

thought: I dislike this a fair amount and would much prefer the job fails until we fix them (IMO this is kind of like commenting out tests once they start failing).

Might be worth chatting about this one in the standup.

shahzadlone:
> > Make the vulnerability scan fail if new vulnerabilities are introduced and pass if previous known ones remain
> 
> thought: I dislike this a fair amount and would much prefer the job fails until we fix them (IMO this is kind of like commenting out tests once they start failing).
> 
> Might be worth chatting about this one in the standup.

I agree it's not clean. But currently we miss any new vulnerabilities that might be introduced as we just assume it's failing due to crisis vulnerability.

Also note that this action isn't required and we said to use it as a notifier of sorts.

Other than it being hacky, what downsides do you think merging this has? If it's that we will sleep on this vulnerability and forget about it, do check out the action job it will still show the vulnerability failures in red when you click on it.

AndrewSisley:
> But currently we miss any new vulnerabilities that might be introduced as we just assume it's failing due to crisis vulnerability.

I think we'll miss a lot of new ones anyway, as they'll show up once and then disappear no?

shahzadlone:
> > But currently we miss any new vulnerabilities that might be introduced as we just assume it's failing due to crisis vulnerability.
> 
> I think we'll miss a lot of new ones anyway, as they'll show up once and then disappear no?

Why will that be? They should always show up, until they are actually resolve. If any new ones are not resolved then the action will continue to fail. Only the `crises` one will be silenced from this change

AndrewSisley:
> * Make the vulnerability scan fail if new vulnerabilities are introduced and pass if previous known ones remain.

Ah okay, I read the PR description differently, I thought you'd introduced a rolling window of sorts.

shahzadlone:
> Okay, I think I agree with you now that I understand the change, thanks for the explanation! It is a nuisance, but like you said better than missing new vulnerabilities.

Sorry about any confusion! Thanks

PR #3092, opened 2024-10-02
feat: Enable Indexing of array fields

### LINKED ISSUE #2279: Sec. Index: enable indexing of elements of array fields
See how it's done in CocroachDB https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20171020_inverted_indexes.md

islamaliev:
## Relevant issue(s)

Resolves #2279

## Description

Enable indexing of array fields.

This change also adds `IsEqual` method to `NormalValue`.


tests/integration/index/array_composite_test.go
```diff
@@ -0,0 +1,597 @@
...
+							"_docID":  testUtils.NewDocIndex(0, 1),
+							"numbers": []int64{30, 40, 50, 30},
+						},
+					},
+				},
+			},
+			testUtils.Request{
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(1),
+			},
```
shahzadlone, line 78: nitpick: I do feel like the abstraction of `makeExplainQuery` and the request in a variable `req` makes it a bit less readable at first glance, I would prefer to avoid jumping into util functions to know what is happened. However I understand might be just my subjective opinion, so feel free to ignore.

client/normal_array_of_nillables.go
```diff
@@ -140,3 +172,41 @@ func normalizeNillableCharsArr[R string | []byte, T string | []byte](val []immut
...
+	if len(a) != len(b) {
+		return false
+	}
+	for i, v := range a {
+		if v.HasValue() && b[i].HasValue() {
+			if !bytes.Equal(v.Value(), b[i].Value()) {
+				return false
+			}
+		} else if v.HasValue() || b[i].HasValue() {
+			return false
```
shahzadlone, line 208: suggestion: Might be worthwhile to hit this line in test.

client/normal_nillable_array.go
```diff
@@ -84,6 +100,16 @@ func (v normalBytesNillableArray) BytesNillableArray() (immutable.Option[[][]byt
...
 }
 
+func (v normalBytesNillableArray) IsEqual(other NormalValue) bool {
+	if otherVal, ok := other.BytesNillableArray(); ok {
+		if v.val.HasValue() && otherVal.HasValue() {
+			return are2DArraysEqual(v.val.Value(), otherVal.Value())
+		}
+		return !v.val.HasValue() && !otherVal.HasValue()
+	}
+	return false
```
shahzadlone, line 110: suggestion: Perhaps also worthwhile to hit with test if not too much hassle.

client/normal_nillable_array_of_nillables.go
```diff
@@ -158,3 +192,20 @@ func normalizeNillableCharsNillableArr[R string | []byte, T string | []byte](
...
 }
+
+func areNormalNillableArraysOfNillablesEqual[T comparable](
+	val immutable.Option[[]immutable.Option[T]],
+	f func() (immutable.Option[[]immutable.Option[T]], bool),
+) bool {
+	if otherVal, ok := f(); ok {
+		return areNillableArraysOfNillablesEqual(val, otherVal)
+	}
+	return true
```
shahzadlone, line 203: suggestion: not covered in test

internal/core/key.go
```diff
@@ -626,6 +626,25 @@ func (k *IndexDataStoreKey) ToString() string {
...
 }
 
+// IsEqual returns true if the two keys are equal
+func (k *IndexDataStoreKey) IsEqual(other IndexDataStoreKey) bool {
+	if k.CollectionID != other.CollectionID || k.IndexID != other.IndexID {
+		return false
+	}
+
+	if len(k.Fields) != len(other.Fields) {
+		return false
```
shahzadlone, line 636: suggestion: Might be worth hitting if not too much hassle.

tests/integration/index/array_test.go
```diff
@@ -0,0 +1,1103 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package index
```
AndrewSisley, line 11: praise: These tests are very readable and it made understanding the feature and reviewing the PR much easier,  thanks Islam :)

client/normal_array.go
```diff
@@ -41,6 +41,10 @@ func (v normalBoolArray) BoolArray() ([]bool, bool) {
 	return v.val, true
 }
 
+func (v normalBoolArray) IsEqual(other NormalValue) bool {
```
AndrewSisley, line 44: nitpick: Whilst I prefer `IsEqual`, `Equal` seems to be the norm in Go (e.g. `bytes.Equal`), I suggest renaming these.

client/normal_array_of_nillables.go
```diff
@@ -140,3 +172,41 @@ func normalizeNillableCharsArr[R string | []byte, T string | []byte](val []immut
...
+		}
+	}
+	return true
+}
+
+func areArrayOfNillableBytesEqual(a, b []immutable.Option[[]byte]) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i, v := range a {
```
AndrewSisley, line 202: todo: It looks like you may have missed a case in this loop - if only one of `v` and `b[i]` has a value it looks like this will return true.

If I'm correct, please also add a test(s) for this before fixing.

internal/core/key.go
```diff
@@ -626,6 +626,25 @@ func (k *IndexDataStoreKey) ToString() string {
 	return string(k.Bytes())
 }
 
+// IsEqual returns true if the two keys are equal
+func (k *IndexDataStoreKey) IsEqual(other IndexDataStoreKey) bool {
```
AndrewSisley, line 630: nitpick: It looks a little odd to see an equality function taking a pointer and a value as it's two arguments, if feeling generous it might be nicer to change this so that `k` and `other` are of the same type :)

AndrewSisley, line 630: nitpick: See nitpick in `client` package about `IsEqual` vs `Equal`, if you change those, please change this.

internal/db/collection_index.go
```diff
@@ -181,6 +181,22 @@ func (c *collection) deleteIndexedDoc(
 	return nil
 }
 
+func (c *collection) deleteIndexedDocWithID(
+	ctx context.Context,
+	docID client.DocID,
+) error {
+	doc, err := c.get(
```
AndrewSisley, line 188: suggestion: It felt a little odd to see the code fetching the doc in order to delete it - I had to chase down the implementation of `deleteIndexedDoc` to realise why - whilst it is kind of obvious once you are in the sec. index mindset, it felt strange briefly stepping into the codespace and it might be worth a comment here to save future people chasing down into the depths of the index code.

internal/db/fetcher/indexer_iterators.go
```diff
@@ -689,22 +798,40 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
...
+					cond.val, err = client.NewNormalNil(cond.kind)
+				} else if !f.indexedFields[i].Kind.IsArray() {
+					cond.val, err = client.NewNormalValue(filterVal)
 				} else {
-					normalVal, err = client.NewNormalValue(filterVal)
+					subCondMap := filterVal.(map[connor.FilterKey]any)
+					for subKey, subVal := range subCondMap {
+						arrKind, ok := cond.kind.(client.ScalarArrayKind)
+						if !ok {
+							return nil, NewErrNotSupportedKindByIndex(cond.kind)
```
AndrewSisley, line 816: suggestion: This line looks impossible to hit, I suggest deleting it and the error type and just do `arrKind := cond.kind.(client.ScalarArrayKind)`.  By including the check you suggest to readers that it is possible.  No worries if you feel strongly about keeping it though.

AndrewSisley (changes requested):
Part way through reviewing, looks really great so far, only very localised comments atm :)

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+			Fields:       make([]core.IndexedField, len(key.Fields)),
+		}
+		copy(resultKey.Fields, key.Fields)
+
+		for i, counter := range arrFieldCounter {
+			field := &resultKey.Fields[index.arrFieldsIndexes[i]]
+			field.Value = normValsArr[i][counter]
+		}
+
+		for i := len(arrFieldCounter) - 1; i >= 0; i-- {
```
AndrewSisley, line 502: todo: Please document why you are iterating in reverse

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+			arrFieldCounter[i]++
+			if arrFieldCounter[i] < len(normValsArr[i]) {
+				break
+			}
+			arrFieldCounter[i] = 0
+			if i == 0 {
+				done = true
```
AndrewSisley, line 509: todo: Please explain why this is code does this

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+		return nil, err
+	}
+	newKeys, err := index.getAllKeys(newDoc, appendDocID)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, oldKey := range oldKeys {
+		isFound := false
+		for i := len(newKeys) - 1; i >= 0; i-- {
```
AndrewSisley, line 554: todo: Please document why you are iterating in reverse

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+}
+
+var _ CollectionIndex = (*collectionArrayIndex)(nil)
+
+func newCollectionArrayIndex(base collectionBaseIndex) *collectionArrayIndex {
+	return &collectionArrayIndex{collectionArrayBaseIndex: newCollectionArrayBaseIndex(base)}
+}
+
+// Save indexes a document by storing the indexed field value.
+func (index *collectionArrayIndex) Save(
```
AndrewSisley, line 584: question: Why does `Save` not retire the old keys? If it does, please include this in the function level documentation

AndrewSisley (changes requested):
Looks really good Islam, just a couple more todos from me before merge

internal/db/index.go
islamaliev: I've updated the documentation for the interface.

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+) error {
+	getNextKey, err := index.newIndexKeyGenerator(doc, true)
+	if err != nil {
+		return err
+	}
+
+	for {
+		key, ok := getNextKey()
+		if !ok {
+			break
```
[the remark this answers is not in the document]
islamaliev: renamed

internal/db/index.go
islamaliev: updated

internal/db/index.go
```diff
@@ -430,3 +425,323 @@ func isUpdatingIndexedFields(index CollectionIndex, oldDoc, newDoc *client.Docum
...
+	// Collect unique values to use as source for generating keys
+	normValsArr := make([][]client.NormalValue, 0, len(index.arrFieldsIndexes))
+	for _, arrFieldIndex := range index.arrFieldsIndexes {
+		arrVal := key.Fields[arrFieldIndex].Value
+		normVals, err := client.ToArrayOfNormalValues(arrVal)
+		if err != nil {
+			return nil, err
+		}
+		sets := make(map[client.NormalValue]struct{})
+		for i := len(normVals) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
islamaliev: updated

internal/db/index.go
islamaliev: done

internal/db/index.go
islamaliev: done

internal/db/index.go
```diff
@@ -210,13 +196,7 @@ var _ CollectionIndex = (*collectionSimpleIndex)(nil)
...
 	doc *client.Document,
 ) (core.IndexDataStoreKey, error) {
-	key, err := index.collectionBaseIndex.getDocumentsIndexKey(doc)
-	if err != nil {
-		return core.IndexDataStoreKey{}, err
-	}
-
-	key.Fields = append(key.Fields, core.IndexedField{Value: client.NewNormalString(doc.ID().String())})
-	return key, nil
+	return index.collectionBaseIndex.getDocumentsIndexKey(doc, true)
```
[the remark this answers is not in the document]
islamaliev: done

internal/db/fetcher/indexer_iterators.go
islamaliev: I'm fine with deleting it. I even put into this place a TODO before, something like this: `TODO: can this be tests?` :)

internal/db/collection_index.go
islamaliev: done

internal/core/key.go
islamaliev: done

internal/core/key.go
islamaliev: done

internal/core/key.go
islamaliev: I see what you mean, but it also feels weird to have all other receivers declared as pointers and only this one as value (if I change to all values). But if I change both to pointers, it won't be clear from the API, if we check deep equality of just pointers.
I don't have a strong opinion here either, but would tend more toward keeping it as is, as it has slightly better performance (because of the pointer), and has clear API.

client/normal_array_of_nillables.go
islamaliev: I don't think I missed any case:
```go
if v.HasValue() && b[i].HasValue() { // if both have values, check them
    ...
} else if v.HasValue() || b[i].HasValue() { // if just one has value, return false
```

client/normal_array_of_nillables.go
islamaliev: done

client/normal_array.go
islamaliev: done

client/normal_nillable_array_of_nillables.go
islamaliev: nice catch, akhi! It wasn't just not covered, it misbehaved.

client/normal_nillable_array.go
islamaliev: should be covered now

internal/db/index.go
AndrewSisley: okay got it - thanks Islam, makes sense to me now :)

internal/core/key.go
AndrewSisley: > but would tend more toward keeping it as is

Not a problem, is only a nitpick, and will vary from person to person which weirdness they tolerate more :)

client/normal_array_of_nillables.go
AndrewSisley: Ah! I think I misread that as if-both-have-values, and if-neither-have-values - thanks for correcting me Islam :)

AndrewSisley (approved):
Looks really good Islam, thanks for this :)

Issue #3089, opened 2024-09-30
FieldName is held on composite and field level blocks


AndrewSisley:
It only needs to be on the child. No idea why this is on the composite at all.

AndrewSisley:
Discussed in standup, this can be removed from the composite.

Issue #3087, opened 2024-09-30
DocID is held on composite and field level blocks


AndrewSisley:
It only needs to be on the child.

AndrewSisley:
Long term we may want to remove this from field commits, leaving it only on the composite, with the exception of the `height=1` field level commits in order to scope them to a particular document.

We could then have quite a lean secondary index using a (future) shortened local docID to a potentially shortened cid in order to not make the commit queries really horrible (finding the composite from a field level commit requires a scan). 

This is however a lot more effort than we want to put in atm.  A quick win in removing docID from the composite would be possible, but does not align with the long-term best-outcome.

AndrewSisley:
Documented in codebase

Issue #3086, opened 2024-09-30
SchemaVersionID is held on composite and field level blocks


AndrewSisley:
It only needs to be on the child.

AndrewSisley:
Removing this from the composite would make the P2P code a fair bit more complicated, it can be done in the future if we really want.

Removing this from the child would require a (space consuming) index in order for the commit queries to not be horrible, exploring exact savings/losses is not desired at this point in time (without a short local representation the index may take up more space than the block).

AndrewSisley:
Documented in codebase

Issue #3056, opened 2024-09-23
Remove Name from block links 


AndrewSisley:
It is not required and just consumes CPU/network/storage resources and complicates the introduction of new stuff like https://github.com/sourcenetwork/defradb/issues/3038

jsimnz:
Can you expand on this? Are you referring to the name inside a block that is linked.

Or the name of a link within a block? 

If it's the latter, I don't see how we can remove it, since we need to (quickly) differentiate between types of links (eg "head" vs field link).

Named links are also necessary for ipld selectors

AndrewSisley:
Discussed on standup and rejected.  Name field deemed to be valuable in the future.

WIP branch name is `3056-rm-block-link-names` (my in my fork).

I tested the space savings for non-branchable collections. With a database size of 10 documents with 3 fields (age, name, email), and a single update to each age field, the change reduced the size of the blockstore by 3.79%, and the total size of the database by 2.57%.   The percentage savings will increase with field name length, but decrease with field value size.

The impact on branchable collections will be much higher as a much higher percentage of the block is links/link-names.

AndrewSisley:
We want this to allow us to use GraphSync https://github.com/ipfs/go-graphsync in the future, it *might* be used for some ACP related client-side optimisation (debated), it *will* be exposed to users to allow them to sync only x,y,z fields to a node.

AndrewSisley:
>  We want this to allow us to use GraphSync https://github.com/ipfs/go-graphsync in the future, it might be used for some ACP related client-side optimisation (debated), it will be exposed to users to allow them to sync only x,y,z fields to a node.

Documented in codebase

PR #3096, opened 2024-10-02
fix: Remove duplication of block heads on delete

### LINKED ISSUE #3085: Block store duplicates lots of data
There is a lot of duplicate data held in the blockstore, and we can considerably shrink the size if we want to.

Note: The block store accounts for the vast majority of Defra's storage requirements.  Bigger blocks also means slower writes, reads, and more network traffic.

These issues also permit structural data-divergance, where the duplication of data creates a structure where the data on the parent may diverge with the data on the child.  It also prevents parents from containing more than one unique value (e.g. a composite of multiple fields, or multiple documents).

Testing also finds that property name length does impact the storage size given our current encoding, we may wish to reduce this too (or, less likely, change the encoding so field name has no impact on size).
```[tasklist]
### Tasks
- [ ] https://github.com/sourcenetwork/defradb/issues/3056
- [ ] https://github.com/sourcenetwork/defradb/issues/3086
- [ ] https://github.com/sourcenetwork/defradb/issues/3087
- [ ] https://github.com/sourcenetwork/defradb/issues/3089
```

AndrewSisley:
## Relevant issue(s)

Resolves #3085 #3089

Documents #3056 #3086 #3087

## Description

Removes the duplication of head links from delete blocks.

PR also includes the following to save the hassle of multiple test-cid updates:
- Removes `fieldName` from composite block deltas
- Removes the magic `_head` link name, and extracts head links to a new, optional prop
- Documents the reasons for duplicating various bits of data in the blockstore blocks as discussed in standup

With the actions defined in `TestQueryCommitsWithFieldIDFieldWithUpdate`, create block size has been reduced by 4%, and update block size by 7% - this will vary a lot depending on what fields are being updated though, the test chosen to calc was just the first test I found that created one small doc, and updated a single field.

I recommend reviewing commit by commit.  The test-cid changes have been pulled out to their own commit.


tests/integration/query/commits/with_delete_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+package commits
+
+import (
+	"testing"
+
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestQueryCommits_WithDelete(t *testing.T) {
```
islamaliev, line 19: nitpick: the test name is not quite descriptive. Had to read the test's body in order to understand what it tests.
Wouldn't something like `TestQueryCommits_AfterDocDeletion_ShouldStillFetch` fit better?

islamaliev (approved):
Looks good, Andy. A huge thanks for good comments you left for block fields. Very helpful.

shahzadlone (approved):
Thanks for all the documentation and the removal of `C` haha LGTM

internal/core/block/block.go
```diff
@@ -82,6 +82,11 @@ type DAGLink struct {
 	// Name is the name of the link.
 	//
 	// This will be either the field name of the CRDT delta or "_head" for the head link.
+	//
+	// This field currently serves no purpose and is duplicating data already held on the target
+	// block.  However we want to have this long term to enable some fancy P2P magic to allow users
+	// to configure the collection to only sync particular fields using
+	// [GraphSync](https://github.com/ipfs/go-graphsync) which will need to make use of this property.
```
fredcarle, line 89: suggestion: You could just say that this field is needed for graphSync and sync optimisation instead of (not sure this was you intention but this is what I get from this) showing through the comment that you are annoyed that the field is still there.

AndrewSisley: No intended annoyance in the wording, I do feel that it is important to note that the current lack of purpose is known and somewhat deliberate though otherwise people (including future me) will assume the lack of purpose is accidental and try and remove it,

PR #3098, opened 2024-10-03
feat: JSON type coercion

### LINKED ISSUE #3097: Improve JSON type coercion
The JSON input type should be able to handle any valid GQL type and not just strings.

This is how mutations with JSON fields currently work:

```graphql
mutation {
  create_Users(input: {custom: "{\"tree\": \"maple\", \"age\": 250}"}) {
    custom
  }
}
```

This is how they should work:

```graphql
mutation {
  create_Users(input: {custom: {tree: "maple", age: 250}}) {
    custom
  }
}
```

nasdf:
## Relevant issue(s)

Resolves #3097

## Description

This PR fixes JSON type coercion and makes it much simpler to filter, create, and update JSON fields.

## Tasks

## How has this been tested?

Added / updated unit and integration tests.

- MacOS


tests/integration/mutation/create/field_kinds/field_kind_json_test.go
```diff
@@ -66,24 +141,31 @@ func TestMutationCreate_WithJSONFieldGivenInvalidJSON_Error(t *testing.T) {
...
 			testUtils.Request{
 				Request: `mutation {
-					create_Users(input: {name: "John", custom: "{\"tree\": \"maple, \"age\": 250}"}) {
+					create_Users(input: {name: "John", custom: "hello"}) {
 						_docID
 						name
 						custom
 					}
 				}`,
-				ExpectedError: `Argument "input" has invalid value {name: "John", custom: "{\"tree\": \"maple, \"age\": 250}"}.
```
AndrewSisley, line 75: todo: Please adjust the test name to match the new test. `TestMutationCreate_WithJSONFieldGivenSimpleString_Error` has the same problem

internal/connor/eq.go
```diff
@@ -43,15 +43,36 @@ func eq(condition, data any) (bool, error) {
...
+	case map[string]any:
+		d, ok := data.(map[string]any)
+		if !ok {
+			return false, nil
+		}
+		for k, v := range d {
+			m, err := eq(cn[k], v)
+			if err != nil {
+				return false, err
+			} else if !m {
```
AndrewSisley, line 55: nitpick: The `else` here is not necessary :)

internal/connor/eq.go
```diff
@@ -43,15 +43,36 @@ func eq(condition, data any) (bool, error) {
...
+		}
+		return true, nil
+
 	case string:
 		if d, ok := data.(string); ok {
 			return d == cn, nil
 		}
 		return false, nil
 
+	case uint64:
```
AndrewSisley, line 67: todo: `case uint64` and `case uint32` appear to be untested, and I expect they are dead code as json typically doesnt get parsed into unsigned types.  Please either delete them, or add integration tests covering them if you can.

internal/connor/eq.go
```diff
@@ -43,15 +43,36 @@ func eq(condition, data any) (bool, error) {
 		}
 		return true, nil
 
+	case map[string]any:
+		d, ok := data.(map[string]any)
+		if !ok {
```
AndrewSisley, line 48: todo: if `ok` is false here that suggests a programming error in which case we should not quietly ignore the value.  Please either remove this check, or convert it to a nil check and test it if that is what you mean to check here.

internal/core/encoding.go
```diff
@@ -191,6 +193,26 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
 	}
 }
 
+func convertToJSON(propertyName string, untypedValue any) (any, error) {
+	untypedMap, ok := untypedValue.(map[any]any)
+	if !ok {
+		return untypedValue, nil
```
AndrewSisley, line 199: todo: If this line is hit it is a programming error, please either return an error, or let it panic.  Or explicitly test for nil if that is what is really wanted here.

client/document.go
```diff
@@ -401,16 +401,66 @@ func getDateTime(v any) (time.Time, error) {
...
 }
 
-func getJSON(v any) (string, error) {
-	s, err := getString(v)
-	if err != nil {
-		return "", err
+func getJSON(v any) (any, error) {
+	val, ok := v.(*fastjson.Value)
+	if !ok {
+		return v, nil
```
AndrewSisley, line 407: todo: if `ok` is false here that suggests a programming error in which case we should not quietly ignore the value. Please either remove this check, or convert it to a nil check and test it if that is what you mean to check here.

AndrewSisley (changes requested):
The behaviour change looks really good Keenan, it is much nicer thanks :)  I have a handful of todos for you before I can approve though.

client/normal_scalar.go
```diff
@@ -86,6 +86,14 @@ func (v normalDocument) Document() (*Document, bool) {
 	return v.val, true
 }
 
+type normalJSON struct {
+	baseNormalValue[any]
+}
+
+func (v normalJSON) JSON() (any, bool) {
```
islamaliev, line 93: todo: this doesn't quite conform to the current style of normal values.

If you want to make it work as a new separate type available as normal value, you should add it to the `NormalValue` interface. And please make sure you follow the convention for regular (required) fields and optional (nillable) fields. You'd also need to include this into current tests for normal values and add new ones if needed.

All current normal scalar type can also be array elements, but I'm not sure if we want or need to enable array of jsons. An array of json can also be presented as json array after all.

It is also a bit unusual to have `any` type as underlying type for `normalJSON` which is supposed to be a strong type like all others.

islamaliev (changes requested):
Looks nice so far. I added some change requests that I think are necessary to address before the merge.

tests/integration/mutation/create/field_kinds/field_kind_json_test.go
nasdf: done

internal/core/encoding.go
nasdf: This is required because when maps are decoded from cbor they are of type `map[any]any` and must be converted to `map[string]any`. Any other value is also valid JSON so we return early if they are not maps.

client/document.go
nasdf: This func can be called with any value that's used to set document field values. Since JSON can be any of our supported scalar types we need this check.

internal/core/encoding.go
AndrewSisley: Ah I see what you mean, can you please document this with a comment?  It looks like defensive coding otherwise

client/document.go
AndrewSisley: Same as other comment in encoding.go - please document this :)

client/normal_scalar.go
nasdf: > It is also a bit unusual to have any type as underlying type for normalJSON which is supposed to be a strong type like all others.

It has to be of type `any` because the JSON value can hold any valid scalar or list of scalars. Is there an alternative to adding the `NormalJSON` type that I have missed?

AndrewSisley: Another option could be to define a json type, this would allow us to differentiate between a known json value and an unknown `any` type:
```
type Json any // no =, not just an alias
```

internal/connor/eq.go
```diff
@@ -43,15 +43,36 @@ func eq(condition, data any) (bool, error) {
...
+			return false, nil
+		}
+		for k, v := range d {
```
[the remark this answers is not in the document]
nasdf: good catch! added length check

internal/core/encoding.go
```diff
@@ -191,6 +193,26 @@ func convertToInt(propertyName string, untypedValue any) (int64, error) {
...
+	}
+	resultValue := make(map[string]any)
+	for k, v := range untypedMap {
+		key, ok := k.(string)
+		if !ok {
+			return nil, client.NewErrUnexpectedType[any](propertyName, k)
+		}
+		val, err := convertToJSON(propertyName, v)
```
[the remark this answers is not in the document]
nasdf: done

internal/planner/mapper/mapper.go
```diff
@@ -1309,6 +1309,12 @@ func toFilterMap(
 		key := &Operator{
 			Operation: sourceKey,
 		}
+		// if the operator is a simple comparison it may be
+		// comparing values of an embedded or JSON object
+		// that are not explicitly mapped in the document
+		if key.Operation == "_eq" || key.Operation == "_in" {
+			return key, sourceClause
+		}
```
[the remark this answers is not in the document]
nasdf: I've refactored this a bit. Let me know if it still needs some changes.

internal/planner/mapper/targetable.go
```diff
@@ -33,8 +33,11 @@ func (k *PropertyIndex) GetProp(data any) any {
 	if data == nil {
 		return nil
 	}
-
-	return data.(core.Doc).Fields[k.Index]
+	doc, ok := data.(core.Doc)
+	if ok {
+		return doc.Fields[k.Index]
+	}
+	return data
```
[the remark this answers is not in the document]
nasdf: I've reverted this

tests/integration/query/simple/with_filter/with_eq_json_test.go
```diff
@@ -28,21 +28,26 @@ func TestQuerySimple_WithEqOpOnJSONField_ShouldFilter(t *testing.T) {
...
+						"tree": "oak",
+						"age": 450
+					}
+				}`,
 			},
 			testUtils.Request{
-				// the filtered-by JSON has no spaces, because this is now it's stored.
 				Request: `query {
-					Users(filter: {custom: {_eq: "{\"tree\":\"oak\",\"age\":450}"}}) {
+					Users(filter: {custom: {_eq: {tree:"oak",age:450}}}) {
```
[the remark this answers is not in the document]
nasdf: added a new test for this

internal/connor/eq.go
nasdf: I removed them

internal/connor/eq.go
nasdf: good catch! I've changed it as suggested

internal/core/encoding.go
nasdf: done

client/document.go
nasdf: done

client/normal_scalar.go
nasdf: > Another option could be to define a json type, this would allow us to differentiate between a known json value and an unknown `any` type:
> 
> ```
> type Json any // no =, not just an alias
> ```

I tried this but it doesn't quite work because `Json` would also match against `any` in a type switch. I ended up having to wrap the value inside of a struct instead and implement the `Unrwap` method of the normal value.

internal/connor/eq.go
```diff
@@ -87,6 +69,27 @@ func eq(condition, data any) (bool, error) {
 	}
 }
 
+// objectsEqual returns true if the given condition and data
+// contain equal key value pairs.
+func objectsEqual(condition map[string]any, data any) (bool, error) {
+	if data == nil {
```
islamaliev, line 75: nitpick: what if `condition` is also `nil`? I think it should return true.

client/normal_scalar.go
```diff
@@ -87,13 +94,17 @@ func (v normalDocument) Document() (*Document, bool) {
 }
 
 type normalJSON struct {
```
islamaliev, line 96: question: is there a reason why we have normal json (non-optional) and don't have normal nillable json?

islamaliev (approved):
I'm happy with the changes. Left some clarifying comments.
Good job, Keenan!

client/normal_scalar.go
nasdf: Since JSON has a `null` value I didn't think it made sense to also represent that within our type system. It would also cause a bit of confusion if `null` was valid for both `NillableJSON` as well as `JSON`. This also applies to arrays of JSON.

internal/connor/eq.go
nasdf: good catch! should be able to handle either being nil now

internal/connor/eq.go
```diff
@@ -66,6 +69,27 @@ func eq(condition, data any) (bool, error) {
 	}
 }
 
+// objectsEqual returns true if the given condition and data
+// contain equal key value pairs.
+func objectsEqual(condition map[string]any, data any) (bool, error) {
+	if data == nil || condition == nil {
+		return reflect.DeepEqual(condition, data), nil
```
AndrewSisley, line 76: todo: This looks very strange, and seems like an unusual place to justify a reflect call, can you please justify this, and if we are happy to keep as-is can you please document why you are doing this, as it doesn't seem to make any sense to me.

AndrewSisley (changes requested):
Looks good, but I have a todo for you RE the reflect call in the connor package before merge.

internal/connor/eq.go
nasdf: I replaced it with a simpler nil comparison and I've added a test for this case

jsimnz:
Small nitpick. The title of the PR should prob be `feat` not `fix` since this is expanded functionality. There was nothing "broken" about the old coercion. Moreover, even the linked issue refers to this as an "improvement"

tests/integration/query/simple/with_filter/with_eq_json_test.go
```diff
@@ -28,27 +28,126 @@ func TestQuerySimple_WithEqOpOnJSONField_ShouldFilter(t *testing.T) {
...
+					"name": "Andy",
+					"custom": {
+						"tree": "oak",
+						"age": 450
+					}
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					Users(filter: {custom: {_eq: {tree:"oak",age:450}}}) {
```
islamaliev, line 50: question: what about other filter ops?
```graphql
query {
	Users(filter: {custom: {_gt: {company: {salary: 100}}}}) {
		name
	}
}
```
Any plans for something like this?

islamaliev:
another thing I'd like to clarify before the merge: these filters on json values will ever match only if a whole json object is equal to what's in the filter, right? So, with this change it still won't be possible to filter by a particular field on json objects.

I presume it's not within the scope of this task.

I see the change makes much nicer to deal with JSON within GraphQL, but I'm not sure how much of filter-related code will stay or have be rewritten to enable by-field filtering.

nasdf:
> another thing I'd like to clarify before the merge: these filters on json values will ever match only if a whole json object is equal to what's in the filter, right? So, with this change it still won't be possible to filter by a particular field on json objects.
> 
> I presume it's not within the scope of this task.
> 
> I see the change makes much nicer to deal with JSON within GraphQL, but I'm not sure how much of filter-related code will stay or have be rewritten to enable by-field filtering.

The plan is to implement untyped filters using the JSON type. This will allow users to filter on any field without us knowing their types. The same approach can be used for alias targeting as well.

internal/connor/connor.go
```diff
@@ -16,41 +53,41 @@ func Match(conditions map[FilterKey]any, data any) (bool, error) {
...
+	case CaseInsensitiveLikeOp:
 		return ilike(conditions, data)
-	case "_nilike":
+	case CaseInsensitiveNotLikeOp:
 		return nilike(conditions, data)
-	case "_none":
+	case NoneOp:
 		return none(conditions, data)
-	case "_not":
+	case NotOp:
```
shahzadlone, line 90: praise: Thanks for making `const` for these!

internal/connor/eq.go
```diff
@@ -66,6 +69,27 @@ func eq(condition, data any) (bool, error) {
 	}
 }
 
+// objectsEqual returns true if the given condition and data
+// contain equal key value pairs.
+func objectsEqual(condition map[string]any, data any) (bool, error) {
+	if data == nil {
+		return condition == nil, nil
```
shahzadlone, line 76: suggestion: Might be worth hitting this case in test.

shahzadlone (approved):
Cool stuff! LGTM

AndrewSisley (approved):
LGTM, thanks Keenan! :)

internal/connor/eq.go
nasdf: I did add a test for this but I can't tell if its working because codecov seems broken for me.

shahzadlone: Codecov takes some time for the report to uplpad. Try the local coverage make rule?

nasdf: I updated the test and made sure its covered now

islamaliev:
Bug bash result: #3141

Issue #3076, opened 2024-09-26
Improve `getNodes` in test utils to return the correct indexes with the relevant nodes


shahzadlone:
We can improve the following function in `test/integration/utils.go`:
```go

func getNodes(nodeID immutable.Option[int], nodes []clients.Client) []clients.Client {
	if !nodeID.HasValue() {
		return nodes
	}

	return []clients.Client{nodes[nodeID.Value()]}
}

```

To return a list of correct node indexes and remove the `getNodeCollections` as suggested in #3069 

So:

`getNodes` would return -> [ListOfIntsCorrespondingToTheNodeIndex] and [ListOfCorrespondingNodes]

This would solve the extra duplicate code logic introduced in #3068 while fixing #3065

PR #3099, opened 2024-10-03
feat: Ability to unrelate private documents from actors

### LINKED ISSUE #2906: Revoke or unshare a document previously created a relation for

shahzadlone:
## Relevant issue(s)
Resolves #2906 

## Description
Follow-up to https://github.com/sourcenetwork/defradb/pull/2907, the ability to delete a relationship (in order to revoke access from an identity).

## CLI Demo
The following revokes the target actors 'relational' access to the private doc (can't read anymore): 

```bash
defradb client acp relationship delete \
--collection Users \
--docID bae-ff3ceb1c-b5c0-5e86-a024-dd1b16a4261c \
--relation reader \
--actor did:key:z7r8os2G88XXBNBTLj3kFR5rzUJ4VAesbX7PgsA68ak9B5RYcXF5EZEmjRzzinZndPSSwujXb4XKHG6vmKEFG6ZfsfcQn \
--identity e3b722906ee4e56368f581cd8b18ab0f48af1ea53e635e3f7b8acd076676f6ac
```

Result:
```json
{
  "RecordFound": true // <-------------- Indicates a relationship was found and deleted
}
```

## How has this been tested?
 CI + Integration Tests + Unit Tests


http/handler_acp.go
```diff
@@ -113,6 +142,22 @@ func (h *acpHandler) bindRoutes(router *Router) {
 		Value: acpAddDocActorRelationshipRequest,
 	}
 
+	acpDeleteDocActorRelationshipRequest := openapi3.NewRequestBody().
+		WithRequired(true).
+		WithContent(openapi3.NewContentWithSchema(openapi3.NewStringSchema(), []string{"text/plain"}))
```
AndrewSisley, line 147: todo: This looks incorrect, you should be strongly typing the schema so that the openAPI magic can document it properly.  Have a look at `openApiSchemas` in http/openAPI.go

Please test this manually after correcting.

tests/integration/acp.go
```diff
@@ -310,11 +310,192 @@ func addDocActorRelationshipACP(
...
+
+		deleteDocActorRelationshipResult, err := node.DeleteDocActorRelationship(
+			ctx,
+			collectionName,
+			docID,
+			action.Relation,
+			targetIdentity,
+		)
+
+		if err == nil {
```
AndrewSisley, line 417: question: Why are these requires wrapped in an if, and not just done after the calls to `AssertError` and `assertExpectedErrorRaised`?

tests/integration/acp.go
```diff
@@ -310,11 +310,192 @@ func addDocActorRelationshipACP(
...
+	// String can be a partial, and the test will pass if an error is returned that
+	// contains this string.
+	ExpectedError string
+}
+
+func deleteDocActorRelationshipACP(
+	s *state,
+	action DeleteDocActorRelationship,
+) {
+	if action.NodeID.HasValue() {
```
AndrewSisley, line 364: todo: This seems like an unmanagably large if-else block and it looks like there is a fair amount of duplication.  Please rework this function.

AndrewSisley (changes requested):
It looks good, but there are a couple of important todos for you before I can approve.  My brain is a little foggy atm too, so it might be wise to wait for someone else to review too.

http/handler_acp.go
shahzadlone: Resolved in commit: `PR(ADHERE-REVIEW): Fix the open api schemas`

tests/integration/acp.go
shahzadlone: This is an issue since the previous node index bug was fixed, I will be taking care of this in https://github.com/sourcenetwork/defradb/issues/3076

Won't do in this PR.

tests/integration/acp.go
shahzadlone: Is copy pasted from `addPolicy` and the `addDocActorRelationshipACP`, I don't remember of the top of my head.

AndrewSisley: Can you change both then please?

tests/integration/acp.go
AndrewSisley: If you really don't want to tidy it up a bit in this function can you please add a todo comment linked to that ticket and make sure this function is mentioned in the issue?

shahzadlone: > If you really don't want to tidy it up a bit in this function can you please add a todo comment linked to that ticket and make sure this function is mentioned in the issue?

I can if you really like, but I literally have the fix I am working on in a separate branch so will be opening the clean up PR soon

AndrewSisley: Ah okay, if it is all coded up already somewhere nearly ready to go dont bother here :)

shahzadlone: Cheers

tests/integration/acp.go
shahzadlone: > Can you change both then please?

Sure, I don't mind :)

shahzadlone: - [ ] Remove `err == nil` if check from test

shahzadlone: Update: I did remove the asserts in the end and the tests started failing it's because it does the checks weather an error was raised or not. I am undoing the change. If I can remember I did do this change because It would upon failure sometimes show a false positive error not raised. I am too lazy to understand why haha but the current framework works, so keeping as is

AndrewSisley: I would guess that you forgot to return/continue early if `expectedErrorRaised == true` and continued to the value assertion after receiving the expected error

AndrewSisley (approved):
Looks good to me, but I would prefer you got a second reviewer to look over this one before merge :) (don't wait more than a couple of working days for that though)

cli/acp_relationship_delete.go
```diff
@@ -0,0 +1,122 @@
...
+		targetActorFlagShort,
+		"",
+		"Actor to delete relationship for",
+	)
+	_ = cmd.MarkFlagRequired(targetActorFlagLong)
+
+	cmd.Flags().StringVarP(
+		&docIDArg,
+		docIDFlag,
+		"",
```
nasdf, line 115: todo: short flag is empty

http/handler_acp.go
```diff
@@ -75,44 +75,110 @@ func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.R
 	responseJSON(rw, http.StatusOK, addDocActorRelResult)
 }
 
-func (h *acpHandler) bindRoutes(router *Router) {
-	successResponse := &openapi3.ResponseRef{
-		Ref: "#/components/responses/success",
+func (s *acpHandler) DeleteDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
+	db, ok := req.Context().Value(dbContextKey).(client.DB)
+	if !ok {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
```
nasdf, line 81: suggestion: panic or internal server error (500) would be better imo. Bad request means the user did something wrong.

nasdf (approved):
Two minor things to consider before merging. The documentation and tests are awesome as always!

cli/acp_relationship_delete.go
shahzadlone: IIRC I don't think we ever had a short flag for --docID

What short flag are you suggesting?

tests/integration/acp.go
shahzadlone: But then that's still another if check anyways. I am happy with the current setup. Is fairly easy to change later for all the test actions if we want.

shahzadlone: But then that's still another if check anyways. I am happy with the current setup. Is fairly easy to change later for all the test actions if we want.

AndrewSisley: Another `if` yes, but you are no longer spoofing the main error assertion logic.

cli/acp_relationship_delete.go
nasdf: I would replace it with `StringVar` if there's no shorthand.

shahzadlone: Ah, I see that is what you meant. Will do.

- [ ] Use `StringVar` for `--docID`

shahzadlone: Resolved in commit: ```PR(ADHERE-REVIEW): Use `StringVar` for only long flag arg```

http/handler_acp.go
shahzadlone: Discussed on discord will resolve in:
https://github.com/sourcenetwork/defradb/issues/3103

tests/integration/acp.go
shahzadlone: Resolved in commit: ``` PR(ADHERE-REVIEW): Check `require.Equal` after the `assertExpectedErrorRaised` ```

PR #3101, opened 2024-10-04
fix: Treat explicitly set nil values like omitted values

### LINKED ISSUE #3100: Setting relation fields to nil breaks one-one creation
Found by David.

If you set the primary side to nil on multiple objects it prevents linking from the secondary side as it finds multiple ojects already linked.

If you set the secondary side to nil on creation it panics.

AndrewSisley:
## Relevant issue(s)

Resolves #3100

## Description

Treat explicitly set nil values like omitted values.

There were a couple of bugs here, I suggest reviewing commit by commit as the old behaviour was documented before being corrected.

### DIFF
--- a/client/document.go
+++ b/client/document.go
@@ -27,6 +27,18 @@ import (
 	ccid "github.com/sourcenetwork/defradb/internal/core/cid"
 )
 
+func init() {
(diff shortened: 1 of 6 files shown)


nasdf (approved):
LGTM

PR #3107, opened 2024-10-07
feat: Add replicator retry

### LINKED ISSUE #3070: Unstable policyIDs in next few sourcehub version bumps
Unstable policyIDs in next few sourcehub version bumps so rework `policyIDs` in tests to be stored in state and not be explicitly required.

It's not as straight forward as `DocIndex` and `CollectionIndex` because policy id is to be embedded in the schema that gets loaded.

Like likely work on this in the cool down week.

### LINKED ISSUE #3072: Replicator basic retry

fredcarle:
## Relevant issue(s)

Resolves #3072 

## Description

This PR adds a replication retry functionality to the database. It uses an exponential backoff until 32 minutes is reached and then it will continuously retry every 32 minutes until the inactive peer is removed from the list of replicators.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/internal/db/config.go
+++ b/internal/db/config.go
@@ -11,6 +11,8 @@
 package db
 
 import (
+	"time"
+
 	"github.com/sourcenetwork/immutable"
 )
 
@@ -20,7 +22,24 @@ const (
 )
 
 type dbOptions struct {
-	maxTxnRetries immutable.Option[int]
+	maxTxnRetries  immutable.Option[int]
+	RetryIntervals []time.Duration
+}
+
+// defaultOptions returns the default db options.
+func defaultOptions() *dbOptions {
+	return &dbOptions{
+		RetryIntervals: []time.Duration{
+			// exponential backoff retry intervals
+			time.Second * 30,
+			time.Minute,
+			time.Minute * 2,
+			time.Minute * 4,
+			time.Minute * 8,
+			time.Minute * 16,
+			time.Minute * 32,
+		},
+	}
 }
 
 // Option is a funtion that sets a config value on the db.
@@ -30,5 +49,13 @@ type Option func(*dbOptions)
 func WithMaxRetries(num int) Option {
 	return func(opts *dbOptions) {
 		opts.maxTxnRetries = immutable.Some(num)
+	}
+}
+
+func WithRetryInterval(interval []time.Duration) Option {
+	return func(opt *dbOptions) {
+		if len(interval) > 0 {
+			opt.RetryIntervals = interval
+		}
 	}
 }
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	ds "github.com/ipfs/go-datastore"
 	dsq "github.com/ipfs/go-datastore/query"
@@ -80,6 +81,15 @@ type db struct {
 	// The peer ID and network address information for the current node
 	// if network is enabled. The `atomic.Value` should hold a `peer.AddrInfo` struct.
 	peerInfo atomic.Value
+
+	// To be able to close the context passed to NewDB on DB close,
+	// we need to keep a reference to the cancel function. Otherwise,
+	// some goroutines might leak.
+	ctxCancel context.CancelFunc
+
+	retryIntervals []time.Duration
+	retryChan      chan event.ReplicatorFailure
+	retryDone      chan retryStatus
 }
 
 // NewDB creates a new instance of the DB using the given options.
@@ -107,20 +117,25 @@ func newDB(
 		return nil, err
 	}
 
-	db := &db{
-		rootstore:    rootstore,
-		multistore:   multistore,
-		acp:          acp,
-		lensRegistry: lens,
-		parser:       parser,
-		options:      options,
-		events:       event.NewBus(commandBufferSize, eventBufferSize),
+	opts := defaultOptions()
+	for _, opt := range options {
+		opt(opts)
 	}
 
-	// apply options
-	var opts dbOptions
-	for _, opt := range options {
-		opt(&opts)
+	ctx, cancel := context.WithCancel(ctx)
+
+	db := &db{
+		rootstore:      rootstore,
+		multistore:     multistore,
+		acp:            acp,
+		lensRegistry:   lens,
+		parser:         parser,
+		options:        options,
+		events:         event.NewBus(commandBufferSize, eventBufferSize),
+		retryChan:      make(chan event.ReplicatorFailure, 100),
+		retryDone:      make(chan retryStatus),
+		retryIntervals: opts.RetryIntervals,
+		ctxCancel:      cancel,
 	}
 
 	if opts.maxTxnRetries.HasValue() {
@@ -136,11 +151,12 @@ func newDB(
 		return nil, err
 	}
 
-	sub, err := db.events.Subscribe(event.MergeName, event.PeerInfoName)
+	sub, err := db.events.Subscribe(event.MergeName, event.PeerInfoName, event.ReplicatorFailureName)
 	if err != nil {
 		return nil, err
 	}
 	go db.handleMessages(ctx, sub)
+	go db.handleReplicatorRetries(ctx)
 
 	return db, nil
 }
@@ -173,7 +189,7 @@ func (db *db) Encstore() datastore.Blockstore {
 }
 
 // Peerstore returns the internal DAG store which contains IPLD blocks.
-func (db *db) Peerstore() datastore.DSBatching {
+func (db *db) Peerstore() datastore.DSReaderWriter {
 	return db.multistore.Peerstore()
 }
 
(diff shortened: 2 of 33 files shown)


internal/db/config.go
```diff
@@ -20,7 +22,24 @@ const (
...
+	return &dbOptions{
+		RetryIntervals: []time.Duration{
+			// exponential backoff retry intervals
+			time.Second * 30,
+			time.Minute,
+			time.Minute * 2,
+			time.Minute * 4,
+			time.Minute * 8,
+			time.Minute * 16,
+			time.Minute * 32,
```
shahzadlone, line 40: question: Interesting strat! Do we want the retry intervals to be configurable?

tests/integration/utils.go
```diff
@@ -723,55 +731,53 @@ func setStartingNodes(
...
-
+func startNodes(s *state, action Start) {
+	nodes := getNodes(action.NodeID, s.nodes)
 	// We need to restart the nodes in reverse order, to avoid dial backoff issues.
-	for i := len(s.nodes) - 1; i >= 0; i-- {
+	for i := len(nodes) - 1; i >= 0; i-- {
+		nodeIndex := i
+		if action.NodeID.HasValue() {
+			nodeIndex = action.NodeID.Value()
+		}
```
shahzadlone, line 741: info: When I fix the `getNodes` here https://github.com/sourcenetwork/defradb/issues/3076 won't have to do this hack anymore, I am guessing this is because the nodeIndex is wrong with their is a nodeID specified correct?

internal/db/db.go
```diff
@@ -80,6 +81,15 @@ type db struct {
...
 	peerInfo atomic.Value
+
+	// To be able to close the context passed to NewDB on DB close,
+	// we need to keep a reference to the cancel function. Otherwise,
+	// some goroutines might leak.
+	ctxCancel context.CancelFunc
+
+	retryIntervals []time.Duration
+	retryChan      chan event.ReplicatorFailure
+	retryDone      chan retryStatus
```
AndrewSisley, line 92: suggestion: `db` is quite a busy type, it might be worth rolling these 3 props up into a new type.

internal/db/p2p_replicator.go
```diff
@@ -23,9 +25,15 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/event"
 	"github.com/sourcenetwork/defradb/internal/core"
+	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
 	"github.com/sourcenetwork/defradb/internal/merkle/clock"
 )
 
+const (
+	retryLoopInterval = 2 * time.Second
+	retryTimeout      = 10 * time.Second
```
AndrewSisley, line 34: todo: Without chasing down the usage of these properties it is quite hard to guess their difference.  Please add some documentation to them.

internal/db/p2p_replicator.go
```diff
@@ -335,3 +344,377 @@ func (db *db) loadAndPublishReplicators(ctx context.Context) error {
...
+				log.ErrorContextE(ctx, "Failed to handle completed replicator retry", err)
+			}
+
+		case <-time.After(retryLoopInterval):
+			db.retryReplicators(ctx)
+		}
+	}
+}
+
+func (db *db) handleReplicatorFailure(ctx context.Context, r event.ReplicatorFailure) error {
```
AndrewSisley, line 378: todo: It looks like the stuff done in this function (and child calls) should be protected by an internal transaction, otherwise we will have partial successes.

suggestion: When introducing the txn, I suggest not hosting the child functions on `db` as it makes it quite easy to accidentally bypass the txn, especially in the short term when we have no tests protecting against this.

Same comments apply to `retryReplicators`, and the `r.Success` half of `handleCompletedReplicatorRetry`.

internal/db/p2p_replicator.go
```diff
@@ -335,3 +344,377 @@ func (db *db) loadAndPublishReplicators(ctx context.Context) error {
...
+	Retrying   bool
+}
+
+func (db *db) createIfNotExistsReplicatorRetry(ctx context.Context, peerID string) error {
+	key := core.NewReplicatorRetryIDKey(peerID)
+	exists, err := db.Peerstore().Has(ctx, key.ToDS())
+	if err != nil {
+		return err
+	}
+	if !exists {
```
AndrewSisley, line 459: suggestion: Inverting this if would remove a level of indentation (and complexity) from the bulk of this function:
```go
if exists {
  return
}
r := ...
```

internal/db/p2p_replicator.go
```diff
@@ -335,3 +344,377 @@ func (db *db) loadAndPublishReplicators(ctx context.Context) error {
...
+		log.ErrorContextE(ctx, "Failed to query replicator retries", err)
+		return
+	}
+	defer closeQueryResults(results)
+	now := time.Now()
+	for result := range results.Next() {
+		rInfo := retryInfo{}
+		err = cbor.Unmarshal(result.Value, &rInfo)
+		if err != nil {
+			log.ErrorContextE(ctx, "Failed to unmarshal replicator retry info", err)
```
AndrewSisley, line 492: thought: If this error is ever hit, it seems likely that it will not be the only record (programming error).  I worry that if we do not delete the record in this block we will have a very rapidly growing database and log file.

AndrewSisley (changes requested):
Sorry about the bother, overall the PR looks good but I think the retry logic needs a little bit of work/documentation before it can be merged.

net/server.go
```diff
@@ -266,24 +266,28 @@ func (s *server) removeAllPubsubTopics() error {
...
-
 	if s.peer.ps == nil { // skip if we aren't running with a pubsub net
 		return nil
 	}
 	s.mu.Lock()
 	t, ok := s.topics[topic]
 	s.mu.Unlock()
 	if !ok {
-		err := s.addPubSubTopic(topic, false, nil)
+		subscribe := false
```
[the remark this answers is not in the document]
fredcarle: I agree. It's like this from a change I had made and didn't clean it up. Thanks for pointing it out.

internal/db/db.go
fredcarle: ✅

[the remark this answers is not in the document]
fredcarle: ✅

internal/db/p2p_replicator.go
fredcarle: ✅

internal/db/p2p_replicator.go
fredcarle: ✅

internal/db/p2p_replicator.go
fredcarle: ✅

net/client.go
```diff
@@ -30,7 +30,20 @@ var (
 
 // pushLog creates a pushLog request and sends it to another node
 // over libp2p grpc connection
-func (s *server) pushLog(evt event.Update, pid peer.ID) error {
+func (s *server) pushLog(evt event.Update, pid peer.ID) (err error) {
+	defer func() {
+		if err != nil && !evt.IsRetry {
```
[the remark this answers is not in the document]
nasdf: thought: the failure event could be used to queue another retry instead of using a channel on the event

nasdf (changes requested):

net/client.go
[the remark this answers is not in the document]
fredcarle: > thought: the failure event could be used to queue another retry instead of using a channel on the event

The retry interval is set per peer and not per update. This is why the channel on the event is used to say if the retry was successful or not.

internal/db/p2p_replicator.go
```diff
@@ -335,3 +344,377 @@ func (db *db) loadAndPublishReplicators(ctx context.Context) error {
...
+	}
+	rep := client.Replicator{}
+	err = json.Unmarshal(repBytes, &rep)
+	if err != nil {
+		return err
+	}
+	switch active {
+	case true:
+		rep.Status = client.ReplicatorStatusActive
+		if rep.Status == client.ReplicatorStatusInactive {
```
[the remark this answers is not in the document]
fredcarle: Good catch. 

✅

internal/db/p2p_replicator.go
```diff
@@ -335,3 +344,377 @@ func (db *db) loadAndPublishReplicators(ctx context.Context) error {
...
+			rep.LastStatusChange = time.Time{}
+		}
+	case false:
+		rep.Status = client.ReplicatorStatusInactive
+		if rep.Status == client.ReplicatorStatusActive {
```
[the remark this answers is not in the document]
fredcarle: ✅

shahzadlone:
todo: you need to change the marked issue number in the description, it's linked to #3070 which is not accurate.

tests/integration/utils.go
fredcarle: yes that's pretty much it.

nasdf (approved):

AndrewSisley (approved):
praise: I found this much easier to read, the retry stuff seems more linear, and the documentation is very clear.

It also looks like it will make it fairly straightforward to adjust if we want to change the way new retries and retry processing are handled/queued in the future if we chose to. 

Thanks Fred :)

PR #3108, opened 2024-10-07
feat: Inherit `read` permission if has only `write` access

### LINKED ISSUE #2992: An actor granted a write permission still can't write unless also given read permission
An actor granted a write permission still can't write unless also given read permission

Example Policy where reader can strictly only read and writer can strictly only write:
```yaml
name: Test Policy

description: A Policy

actor:
  name: actor

resources:
  users:
    permissions:
      read:
        expr: owner + reader

      write:
        expr: owner + writer

    relations:
      owner:
        types:
          - actor

      reader:
        types:
          - actor

      writer:
        types:
          - actor

```

Then the policy above (assume `XYZ` is resulting `policyID`) is linked in a schema that is loaded:

```gql
type Users @policy(id: XYZ, resource: "users") {
	name: String
	age: Int
}
```

Now if the `owner` (index `1`) makes a relationship giving `write` access to the `second` actor (index `2`) in our testing frame work like syntax:

```go
testUtils.AddDocActorRelationship{
	DocID: 0,
	RequestorIdentity: 1,
	TargetIdentity: 2,
	Relation: "writer",
}
```

The identity `2` still can not mutate due to lack of read permission.

```go
testUtils.UpdateDoc{
	Identity: immutable.Some(2), // This identity can still not update.
	DocID: 0,
	Doc: `
		{
			"name": "Shahzad Lone"
		}
	`,
	ExpectedError: "document not found or not authorized to access",
}
```

Some existing tests that document this: 
- `TestACP_OwnerGivesUpdateWriteAccessToAnotherActorWithoutExplicitReadPerm_OtherActorCantUpdate`
- `TestACP_OwnerGivesUpdateWriteAccessToAnotherActorWithoutExplicitReadPerm_OtherActorCantDelete`

shahzadlone:
## Relevant issue(s)
Resolves #2992 

## Description
- Issue has more details

## How has this been tested?
- CI & Fixed the asserted test that documented this behavior


acp/acp.go
```diff
@@ -89,10 +89,11 @@ type ACP interface {
 	// Otherwise if check failed then an error is returned (and the boolean result should not be used).
 	//
 	// Note(s):
-	// - permission here is a valid DPI permission we are checking for ("read" or "write").
+	// - permissions here is a valid DPI permission we are checking for ("read" or "write").
+	// - returns true if has access to ANY of the given permissions.
```
islamaliev, line 93: todo: I think some more documentation of why exactly it behaves like this is important.
Say I call `CheckDocAccess(["delete", "read"])` and it says `true` just because I have "read" access. For a certain action users should have specific permissions (all of them). Not sure how useful it is.

acp/source_hub_client.go
```diff
@@ -336,45 +337,52 @@ func (a *sourceHubBridge) IsDocRegistered(
...
+				permission.String(),
+				policyID,
+				actorID,
+				resourceName,
+				docID,
+			)
+		}
+
+		if hasAccess {
+			break
```
islamaliev, line 372: todo: commented on this in `acp/acp.go`, will repeat:
I don't think it's very useful to return `true` if any permission is present.
There should be a good reason for this I believe.

internal/db/collection.go
```diff
@@ -327,7 +327,7 @@ func (c *collection) getAllDocIDsChan(
 
 			canRead, err := c.checkAccessOfDocWithACP(
 				ctx,
-				acp.ReadPermission,
+				acp.PermissionsThatImplyRead,
```
islamaliev, line 330: note: if this is the reason `CheckDocAccess` checks if ANY permission is given, I strongly disagree with it.
In this particular case you know exactly that you need "read" permission and it's the one you need to pass. I think it's the job of a particular ACP implementation to handle permission inheritance in the best manner. So SourceHub ACP upon handling `CheckDocAccess("read")` should check both "read" and "write" without making other worry about it.

islamaliev (changes requested):
Good improvement, but I disagree with some decisions.

acp/acp.go
```diff
@@ -89,10 +89,11 @@ type ACP interface {
...
 	CheckDocAccess(
 		ctx context.Context,
-		permission DPIPermission,
+		permissions []DPIPermission,
```
AndrewSisley, line 96: suggestion: If you move this to the end you could allow simpler calling of it for those who call it with a single item:
```go
CheckDocAccess(
  ctx context.Context,
  docID string,
  permissions ...DPIPermission,
)
```

AndrewSisley (changes requested):
I agree with Islam's comments, and added a minor suggestion of my own.

acp/source_hub_client.go
```diff
@@ -342,39 +343,86 @@ func (a *sourceHubBridge) CheckDocAccess(
 	resourceName string,
 	docID string,
 ) (bool, error) {
-	isValid, err := a.client.VerifyAccessRequest(
+	// We grant "read" access even if the identity does not explicitly have the "read" permission,
+	// as long as they have any of the permissions that imply read access.
+	if permission == ReadPermission {
```
islamaliev, line 348: nitpick: I'd personally prefer to extract it into a separate method:
```go
if permission == ReadPermission {
    return a.checkDocReadAccess(ctx, ...)
}
...
```
This way you could clearly see that the branch returns. It wasn't that obvious at first when I reviewed it.

islamaliev (approved):
LGTM! Good job, Shahzad!

internal/db/collection.go
shahzadlone: - [x] done

acp/source_hub_client.go
shahzadlone: - [x] done

acp/acp.go
shahzadlone: - [x] done

acp/acp.go
shahzadlone: No longer needed, moved the implementation to source hub client

acp/source_hub_client.go
shahzadlone: I like that I can see the full flow in one function tbh without jumping to another. leaving as is

AndrewSisley (approved):
LGTM, thanks Shahzad :)

PR #3124, opened 2024-10-10
fix: Prevent mutations from secondary side of relation

### LINKED ISSUE #3102: Updating a relation from secondary side does not work
Providing a new docID fails with an error as documented by `TestMutationUpdateOneToOne_RelationIDToLinkFromSecondarySide`.  Setting the id field to `""` returns `malformed document ID, missing either version or cid`.  Setting the id field to `null` panics.

UPDATE: We will resolve this by removing the ability to mutate the relationship from the secondary side.

AndrewSisley:
## Relevant issue(s)

Resolves #3102

## Description

Prevents mutations from secondary side of relation.

Also validates that values given to relation fields are actually valid docIDs - previously it only validated this on the secondary side, and it was as easy to introduce it to the primary side as it was to correct the tests expecting the failure, so I've added this here.

As IDs set via secondary affected the secondary docID the order of results in some tests have changed.  Also changed are tests that accidentally created docs via the secondary side that didn't test the secondary mutation (e.g. a lot of tests testing queries from the secondary side also created the test docs from this direction).


tests/integration/mutation/create/field_kinds/one_to_many/with_alias_test.go
```diff
@@ -94,35 +94,6 @@ func TestMutationCreateOneToMany_AliasedRelationNameNonExistingRelationManySide_
...
-						{
-							"name": "Painted House",
-						},
-					},
-				},
-			},
-		},
-	}
-	executeTestCase(t, test)
-}
```
shahzadlone, line 125: question: Why not assert the current behavior rather than removing?

tests/integration/mutation/create/field_kinds/one_to_one/with_alias_test.go
```diff
@@ -73,23 +73,6 @@ func TestMutationCreateOneToOne_UseAliasWithNonExistingRelationPrimarySide_Creat
...
-				Doc: `{
-					"name": "Painted House",
-					"author": "bae-be6d8024-4953-5a92-84b4-f042d25230c6"
-				}`,
-				ExpectedError: "document not found or not authorized to access",
-			},
-		},
-	}
-	executeTestCase(t, test)
-}
```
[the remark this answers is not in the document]
AndrewSisley: Same answer :)

internal/request/graphql/schema/generate.go
```diff
@@ -566,6 +566,20 @@ func (g *Generator) buildMutationInputTypes(collections []client.CollectionDefin
 					continue
 				}
 
+				if field.Kind == client.FieldKind_DocID && strings.HasSuffix(field.Name, request.RelatedObjectID) {
```
nasdf, line 569: thought: the logic here and in `client.Document` is the same. Would it make sense to have a `field.IsSecondaryRelation` helper function instead?

nasdf (approved):
LGTM. Thanks for also fixing the doc ID validation logic.

internal/request/graphql/schema/generate.go
AndrewSisley: I thought about it, but decided it wasn't worth coupling them together and boost the (public) client package surface area.  Especially given that the gql/schema package is begging for a re-write.

tests/integration/mutation/create/field_kinds/one_to_many/with_alias_test.go
shahzadlone: Ah. Cheers thanks!

shahzadlone (approved):
LGTM

PR #3125, opened 2024-10-10
feat: Add node identity

### LINKED ISSUE #2908: Node identity
At the moment identity is only client-side feature. We should extend it to nodes as well so that they can also participate in document permission sharing.

A node identity should be generated from a private key and can be stored in the keyring on the node start (prompting password).

islamaliev:
## Relevant issue(s)

Resolves #2908

## Description

Assign an identity to the node upon upon startup.


cli/start.go
```diff
@@ -147,6 +134,13 @@ func MakeStartCommand() *cobra.Command {
 			isDevMode := cfg.GetBool("development")
 			if isDevMode {
 				cmd.Printf(devModeBanner)
+				if cfg.GetBool("keyring.disabled") {
```
islamaliev, line 137: note: If keyring is available we use it for the identity. Otherwise we generate a temporary one only if `--development` flat is passed. I thought temporary identities make sense only if you explicitly in non-production mode. Not 100% sure we want it like this.

acp/identity/identity.go
```diff
@@ -158,3 +158,12 @@ func didFromPublicKey(publicKey *secp256k1.PublicKey, producer didProducer) (str
...
 }
+
+// IntoRawIdentity converts an `Identity` into a `RawIdentity`.
+func (identity Identity) IntoRawIdentity() RawIdentity {
+	return RawIdentity{
+		PrivateKey: hex.EncodeToString(identity.PrivateKey.Serialize()),
+		PublicKey:  hex.EncodeToString(identity.PublicKey.SerializeCompressed()),
+		DID:        identity.DID,
+	}
+}
```
shahzadlone, line 169: todo: Since you have made a function for this, why not call this at the end of `Generate()` in the file: `acp/identity/generate.go` right at the `return` stage.

suggestion: Also perhaps can move this function and the `RawIdentity` struct to a new file like `raw_identity.go` perhaps?

cli/start.go
```diff
@@ -252,3 +246,97 @@ func MakeStartCommand() *cobra.Command {
...
+		time.Duration(0),
+		immutable.None[string](),
+		immutable.None[string](),
+		false,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info(info, corelog.Any("Identity", nodeIdentity.IntoRawIdentity()))
```
shahzadlone, line 317: question: Are you sure we want to dump/log the full identity information of the node? i.e. including the private key.

cli/start.go
```diff
@@ -252,3 +246,97 @@ func MakeStartCommand() *cobra.Command {
...
+		time.Duration(0),
+		immutable.None[string](),
+		immutable.None[string](),
+		false,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info("Generated ephemeral identity", corelog.Any("Identity", nodeIdentity.IntoRawIdentity()))
```
shahzadlone, line 339: Same as above:
question: Are you sure we want to dump/log the full identity information of the node? i.e. including the private key.

cli/start.go
shahzadlone: This should definitely be documented in the `README`  specially near the `keyring` section

cli/start.go
islamaliev: good catch. I didn't think about it.
I just wanted to output it to the console. I can just print it out directly to the console. Or you mean it's better not to output private key at all?

cli/start.go
shahzadlone, line 1: todo: Can you please document under the `README` start section about (some might me out of scope of this pr):
1) What is required to have setup in order to `start` the node (essentially from the `peer` `enc` keys and `node identity`) first time. 
2) Things that aren't required if we skip them, for e.g. `--no-keyring` what happens (i.e. adds ephemeral keys/identity)
3)  Config options and info about what node identity is.

acp/identity/identity.go
islamaliev: extracted common code.

`IntoRawIdentity` belongs to `Identity`, so I think it should stay there. 

I think we will need in the near future `IntoIdentity` for `RawIdentity`. This one will to into `raw_identity.go`

cli/start.go
shahzadlone: Does it print only on the first `make start`? Tbh I am not 100% sure if we should print at all. But might be good to think if others have good reasons for printing it (specially the private key).

shahzadlone:
Some questions that might be worth it to have in the PR description for reviewers:

question: are users expected to save the node key after it's generated? How exactly is it to be used by users?

question: can a custom private key be given instead by the user initiating the first start. Can node identity be changed after a start?

question: How is the node identity different from the `requesting identity` (`--identity` input)?

cli/node.go
```diff
@@ -0,0 +1,34 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func MakeNodeCommand() *cobra.Command {
```
nasdf, line 17: todo: I think it's confusing to have both `node` and `client` subcommands that do the same thing. If the naming of `node` is preferred over `client` we should just rename it and merge them.

cli/start.go
nasdf: +1 please add to the `Key Management` section of the readme

cli/start.go
```diff
@@ -252,3 +246,97 @@ func MakeStartCommand() *cobra.Command {
 		"Skip generating an encryption key. Encryption at rest will be disabled. WARNING: This cannot be undone.")
 	return cmd
 }
+
+func addEncryptionKey(kr keyring.Keyring, cfg *viper.Viper, opts []node.Option) ([]node.Option, error) {
```
nasdf, line 250: suggestion: `getOrCreateEncryptionKey` better reflects this function

cli/start.go
```diff
@@ -252,3 +246,97 @@ func MakeStartCommand() *cobra.Command {
...
+		}
+		log.Info("generated encryption key")
+	} else if err != nil && !errors.Is(err, keyring.ErrNotFound) {
+		return nil, err
+	}
+	opts = append(opts, node.WithBadgerEncryptionKey(encryptionKey))
+	return opts, nil
+}
+
+func addPeerKey(kr keyring.Keyring, opts []node.Option) ([]node.Option, error) {
```
nasdf, line 269: suggestion: same as above

cli/start.go
```diff
@@ -252,3 +246,97 @@ func MakeStartCommand() *cobra.Command {
...
+			return nil, err
+		}
+		log.Info("generated peer key")
+	} else if err != nil {
+		return nil, err
+	}
+	return append(opts, net.WithPrivateKey(peerKey)), nil
+}
+
+func addIdentity(kr keyring.Keyring, opts []node.Option) ([]node.Option, error) {
```
nasdf, line 287: suggestion: same as above

nasdf (changes requested):
Looks good overall, just a few minor things to discuss.

cli/start.go
islamaliev: done

cli/start.go
jsimnz: Absolutely do not print the private key 🤦

http/handler_store.go
```diff
@@ -342,6 +342,17 @@ func (s *storeHandler) ExecRequest(rw http.ResponseWriter, req *http.Request) {
...
 
+func (s *storeHandler) GetNodeIdentity(rw http.ResponseWriter, req *http.Request) {
+	store := req.Context().Value(dbContextKey).(client.DB)
+
+	identity, err := store.GetNodeIdentity(req.Context())
+	if err != nil {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+		return
+	}
+	responseJSON(rw, http.StatusOK, identity)
```
jsimnz, line 353: Todo: scrub the private key from this response

cli/start.go
islamaliev: done

jsimnz (changes requested):
Quick Todo. Also commented on another Todo regarding logging of identity. In general, you should be extremely careful when touching the private key for any I/O related actions. Should only be touched for actual cryptographic ops

cli/start.go
islamaliev: removed

cli/start.go
islamaliev: done

cli/start.go
islamaliev: done

cli/node.go
islamaliev: I'm also not a big fan of having a dedicated `node` command just for one subcommand. But I think moving it under the `client` might send a wrong message to the users as there might be multiple clients dealing with the same node and command `client identity` is even more confusing.
Maybe we should discuss it on the standup.

http/handler_store.go
islamaliev: done

docs/website/references/cli/defradb_start.md
```diff
@@ -14,7 +14,12 @@ defradb start [flags]
 
 ```
       --allowed-origins stringArray   List of origins to allow for CORS requests
-      --development                   Enables a set of features that make development easier but should not be enabled in production
+      --development                   Enables a set of features that make development easier but should not be enabled
```
shahzadlone, line 17: Suggestion: I think this reads a bit odd now "should not be enabled" and under production some bullets. Perhaps leave as was and still point out the differences that happen with flag is enabled

tests/integration/acp.go
```diff
@@ -357,134 +247,92 @@ type DeleteDocActorRelationship struct {
...
+	}
 
-		if !expectedErrorRaised {
-			require.Equal(s.t, action.ExpectedError, "")
-			require.Equal(s.t, action.ExpectedRecordFound, deleteDocActorRelationshipResult.RecordFound)
-		}
+	executeACPAction(action.NodeID, processNode, s)
+}
+
+func executeACPAction(nodeID immutable.Option[int], processNode func(nodeID int), s *state) {
```
shahzadlone, line 271: question: why did this helper and utility function needed to be introduced?

tests/integration/acp.go
```diff
@@ -357,134 +247,92 @@ type DeleteDocActorRelationship struct {
...
 
-func deleteDocActorRelationshipACP(
-	s *state,
-	action DeleteDocActorRelationship,
-) {
-	if action.NodeID.HasValue() {
-		nodeID := action.NodeID.Value()
-		collections := s.collections[nodeID]
+func deleteDocActorRelationshipACP(s *state, action DeleteDocActorRelationship) {
+	processNode := func(nodeID int) {
```
shahzadlone, line 251: Question: why did this action needed to be refactored?

README.md
```diff
@@ -67,10 +67,11 @@ The following keys are loaded from the keyring on start:
 
 - `peer-key` Ed25519 private key (required)
 - `encryption-key` AES-128, AES-192, or AES-256 key (optional)
+- `identity-key` Secp256k1 private key (optional). This key is used for node's identity.
```
shahzadlone, line 70: Todo: I have mixed feelings about the name identity key, It can be very easily confusing and can be mistaken for requestor's identity. I propose adding the word node in it perhaps.

README.md
```diff
@@ -67,10 +67,11 @@ The following keys are loaded from the keyring on start:
...
 
 A secret to unlock the keyring is required on start and must be provided via the `DEFRA_KEYRING_SECRET` environment variable. If a `.env` file is available in the working directory, the secret can be stored there or via a file at a path defined by the `--secret-file` flag.
```
shahzadlone, line 72: Just like this paragraph it will be nice to explain here what a node identity is.

tests/integration/acp.go
islamaliev: so that we don't have somewhat sensitive logic of of running acp-related actions in 2 different places

docs/website/references/cli/defradb_start.md
shahzadlone: On re-read, I think its just the odd line break before it says "in production:" on a new line?

internal/db/db.go
```diff
@@ -302,6 +308,13 @@ func (db *db) DeleteDocActorRelationship(
 	return client.DeleteDocActorRelationshipResult{RecordFound: recordFound}, nil
 }
 
+func (db *db) GetNodeIdentity(context.Context) (immutable.Option[identity.PublicRawIdentity], error) {
```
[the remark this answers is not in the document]
islamaliev: we won't return anything but public identity. I don't see a point of making the method name unnecessarily longer. And I don't understand why it's 'Todo'

tests/integration/acp.go
islamaliev: there was code duplication in 3-4 places. It is tiresome to keep them in sync. That's why I refactored. This file is shrunk by decent amount of lines now.

README.md
islamaliev: ok, will change to `node-identity-key`

README.md
islamaliev: will do

tests/integration/acp.go
shahzadlone: I am very strongly against this change for 2 reasons:

1) My work resolving: https://github.com/sourcenetwork/defradb/issues/3076 already fixes the duplication issue and the `getNodes` bug. So in addition to being a directly conflict it also doesn't fix the underlying bug.

2) This over abstraction complicates things for no reason, and only applied to acp actions. All other actions now have a different flow and acp actions have a totally different flow. Either do it for all of them or none of them. I do prefer the explicitness of each action the way it is, as it keeps them flexible. I now have to jump to just to `executeACPAction` to know if there are multiple nodes specified does it skip for `SourceHubACPType` or not.

tests/integration/acp.go
shahzadlone: Similar to my other comment, I don't like this over abstraction and jumping around into helpers only for acp actions.

docs/website/references/cli/defradb_start.md
shahzadlone: Ah i see.

I think something ugly but this should fix it?:

```go
const developmentDescription = `Enables a set of features that make development easier but should not be enabled` +
	` in production:` + `

- allows purging of all persisted data 
- generates temporary node identity if keyring is disabled
`

````

File: tests/integration/acp.go (hunk 11 above)
Comment (islamaliev) [reply tests/integration/acp.go:271]: call them "helpers" or not, they fix code duplication problem. And I though this is something that should be appreciated. Otherwise I'm absolutely confused especially when seeing tasks like this #3069

File: tests/integration/acp.go (hunk 12 above)
Comment (AndrewSisley) [reply tests/integration/acp.go:251]: > There in no abstraction here. I don't know why you use this word.

A function is an abstraction, I think Shahzad's use of the word is totally fair here.

My (and possibly Shahzad's) primary objection is that you are reworking the `getNodes` stuff in a very ACP specific way that does not solve the general wider problem noted in #3076 - at a time when (I believe) Shahzad is actively working on and is very nearly ready to open up a PR to solve that wider issue in a consistent way for all actions (not just ACP).

You are both solving the same problem at the same time, only Shahzad was working on it first and has solved it across all of our actions, not just ACP.

At the same time your changes have coupled various other, independent ACP action logic into the same helper function (abstraction).  I also disagree with this choice - IMO it is nicer to keep them independent and individually as simple as they can be, especially given that this is essentially untested test code in a very complicated test framework.

Comment (islamaliev) [reply tests/integration/acp.go:251]: > you are reworking the getNodes

> You are both solving the same problem

I did not rework anything. Can't you guys see that it's just refactoring? The execution flow is the same. Maybe I'm blind here.

Comment (AndrewSisley) [reply tests/integration/acp.go:251]: You have reworked the code that manages `getNodes` for the ACP actions, something that is actively being worked on by another team member who is looking to solve this awkwardness for all our actions.

We all recognise that there is a problem, I am mostly asking you to be a little patient, tolerate it within the ACP actions for a little longer, and not step on the toes of the developer who has been trying to sort out the wider problem.

By solving the problem for all actions, we should have simpler and more consistent action code, allowing us to maintain test actions a little more easily without worrying about action/action-sub-set specific oddities.

Comment (AndrewSisley) [reply tests/integration/acp.go:251]: This thread is getting a little long, perhaps @islamaliev, @shahzadlone , myself and anyone else interested can have a call in dev-hangout?

File: tests/integration/acp.go (hunk 11 above)
Comment (shahzadlone) [reply tests/integration/acp.go:271]: I am not sure why you are confused about #3069 in this PR. I do agree it can be more updated with more specific info as I found the #3076 bug after that.

File: tests/integration/acp.go (hunk 12 above)
Comment (shahzadlone) [reply tests/integration/acp.go:251]: @AndrewSisley has described all the concerns I had. In summary #3076 resolves the duplication problem and the underlying bug for all actions, and doesn't use another abstraction/function. Was going to open PR over the weekend but was traveling to Dubai, I will have the PR open soon.

Regardless of the decision I don't see how/why this has to be done in this PR, this refactor/abstraction is independent of the Node Identity feature.

Comment (islamaliev) [reply tests/integration/acp.go]: removed executeAPCAction

File: internal/db/db.go (hunk 15 above)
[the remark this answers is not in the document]
Comment (shahzadlone) [reply internal/db/db.go:311]: This is because to me `Identity` makes up of multiple components which includes the public,priv, etc. You also have a `PublicRawIndentity` as return type. I think it's quite fair to indicate it's just the public part of the identity, not the full identity.

File: tests/integration/utils.go [hunk 16]
```diff
@@ -2565,3 +2571,26 @@ func parseCreateDocs(action CreateDoc, collection client.Collection) ([]*client.
...
+
+		expectedIdent := getIdentity(s, i, immutable.Some(0))
+		expectedRawIdent := immutable.None[identity.PublicRawIdentity]()
+		if expectedIdent.HasValue() {
+			expectedRawIdent = immutable.Some(expectedIdent.Value().IntoRawIdentity().Public())
+		}
+
+		require.Equal(s.t, expectedRawIdent, actualIdent, "identities mismatch")
+	}
+}
```
shahzadlone, line 2596: question: I don't quite understand why this is needed / what this actually does. I also feel like we are "mixing" identities? Isn't a node's `Node identity` different from a nodes `Requesting Identities` (which we have set up from before). I don't get the correlation.

tests/integration/acp.go
shahzadlone: Seems like you are deciding the keep the `-1` helpers. In that case I would suggest fixing them and making every single action follow that pattern so we can have that `-1` feature to test invalid docID's, collectionIDs etc.

tests/integration/utils.go
shahzadlone: Ah cheers for explaining!

Here is my suggestion regarding resolving the `mixing` of identities as you have right now:
- Rename `getIdentity` to `getRequestingIdentity` and then make a similar function called `getNodeIdentity`.
- Similarly rename the `identities [][]identity.Identity` on `state` struct (i.e. `s.idenities`) to `requestingIdentities` and add a new variable to the `state` struct called `nodeIdentities [][]identity.Identity` which does similar stuff for every node's identity.

Then you can do what you are doing above and it will be somewhat easier to understand, would still document this in the `checkNodesIndentity` function.

question: Could all this checking not just happen in the assertion function that tests the `GetNodeIdentity` behavior?

internal/db/db.go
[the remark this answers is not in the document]
islamaliev: > You also have a PublicRawIndentity as return type.

This is exactly the reason why I don't see a point in stating it twice.

With this logic we should also change CLI to `defradb client public_node_identity`

tests/integration/acp.go
islamaliev: > Seems like you are deciding the keep the -1 helpers. 

I'm now really confused. We discussed it yesterday and agreed to keep helpers as is and remove only `executeACPAction`. I didn't decide to keep anything. I just extracted 4 occurrences of this:
```go
var targetIdentity string
if action.TargetIdentity == -1 {
	targetIdentity = ""
} else {
	optionalTargetIdentity := getIdentity(s, i, immutable.Some(action.TargetIdentity))
	if !optionalTargetIdentity.HasValue() {
		require.Fail(s.t, "Expected non-empty target identity, but it was empty.", s.testCase.Description)
	}
	targetIdentity = optionalTargetIdentity.Value().DID
}
```
and other parts into separate functions.
And please leave your comments precisely on a line that you are referring to.

tests/integration/utils.go
islamaliev: I already have the identities separated, but it's on another branch. I call them `nodeIdentities` and `userIdentities`. You think I should cherry-pick it here?

> question: Could all this checking not just happen in the assertion function that tests the GetNodeIdentity behavior?

you mean something like this?
```go
testUtils.AssertNodeIdentity {
    NodeID: 0,
    ExpectedIdentity: 0,
}
```
I'm not sure it's worth the effort. And I don't see how it will improve the test.

internal/db/db.go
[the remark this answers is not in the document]
shahzadlone: if for example a command like `defradb client node-identity get`

I would definitely prefer it to print:

```json
{
  "PublicIdentity": "XYZ"
}
```

Rather than:

```json
{
  "Identity": "XYZ"
}
```

Note: PrivateKey contains all the components of an identity to me as you can derive public key and `did` from it. So only saying `Identity` can be confusing if all it returns is the public part of the set identity.

However I am happy to concede this point if consensus disagrees with this nitpick of mine. Might be worth seeing what the others think as well.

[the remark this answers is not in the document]
islamaliev: it prints:
```json
{
    "PublicKey": "XYZ",
    "DID": "did"
}
```

tests/integration/acp.go
islamaliev: This conversation leads nowhere. You merge first your stuff. I will see what I can do later.

AndrewSisley: > We discussed it yesterday and agreed to keep helpers as is and remove only executeACPAction

I also thought we agreed to keep the helpers.  I have a minor preference against the partial refactor, but it is not nearly important enough to drag out the review over, and it shouldn't impact the work Shahzad is currently working on.

acp/identity/raw_identity.go
```diff
@@ -0,0 +1,54 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package identity
```
AndrewSisley, line 11: praise: Thanks for breaking this out to a new file

internal/db/db.go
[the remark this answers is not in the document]
AndrewSisley: Just noting that I prefer the current than the suggestion.  IMO there is no such thing as a public/private identity - only public and private components of the same identity, and as such I agree with the function name `GetNodeIdentity`, and the current print structure (essentially printing the public properties of the identity).

cli/node_identity.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func MakeNodeIdentityCommand() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "node_identity",
+		Short: "Get information about the node's identity",
+		Long: `Get information about the node's identity.
```
AndrewSisley, line 21: suggestion: I think it is worth noting in the `Short` and first line of `Long` that this command only returns the public components of the node's identity.  Unless perhaps you have immediate plans to also yield the private key too in a follow-up PR.

Maybe just change it to:
> Get the public information about the node's identity

Otherwise this info is buried (and even then only by omission) within the body of `Long`.

cli/start.go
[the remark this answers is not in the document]
AndrewSisley: I think that is out of scope as it requires a quite broad look at what needs to be done here, I also *think* Keenan has plans but I'm not sure.

cli/start.go
```diff
@@ -244,11 +244,99 @@ func MakeStartCommand() *cobra.Command {
...
+		false,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(opts, db.WithNodeIdentity(nodeIdentity)), nil
+}
+
+func addEphemeralIdentity(opts []node.Option) ([]node.Option, error) {
```
AndrewSisley, line 324: todo: 'ephemeral' suggests that this is temporary, however I cannot spot any documentation as to why it is so, and it seems a little odd if it is only kept in memory for a dev mode database (potentially file-based) - surely things will break if it is restarted?

I'm happy with a ticket being opened and linked with a code-comment-todo if the restart problem is real, as this feels quite niche.  But please make sure that this function is documented.

Might be a bit of a hassle now, but long term we could perhaps just store this in plain text on disk (it is in dev-mode).

AndrewSisley (approved):
Looks good to me, just couple of small suggestions from me :)

nasdf (approved):
Looks good!

internal/db/db.go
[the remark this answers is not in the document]
shahzadlone: > it prints:
> 
> ```json
> {
>     "PublicKey": "XYZ",
>     "DID": "did"
> }
> ```

Ah okay cheers! Will then just call this a `nitpick` and Andy agrees with the way it is so I concede this `nitpick`.

tests/integration/acp.go
shahzadlone: We did agree to keep the helpers, however all I am saying is to mark that this needs to be done for all actions (as this is a partial refactor). In other words at least have an issue marking we need to do this for all actions (or whatever way we decide to handle in future the invalid/empty cases).

shahzadlone: Made this: https://github.com/sourcenetwork/defradb/issues/3147

Resolving this thread

cli/start.go
islamaliev: well, it IS temporary just to have some identity to play around with during development. I didn't think it was important to persist it.

Created a task https://github.com/sourcenetwork/defradb/issues/3148

tests/integration/utils.go
shahzadlone: > I already have the identities separated, but it's on another branch. I call them `nodeIdentities` and `userIdentities`. You think I should cherry-pick it here?

Interesting, why is it in a separate PR? I would really like us to not mix both these different kinds of identities. I do slightly prefer the name `requestingIdentities` or `requestorIdentities` over `userIdentities` because incase of relationship management we take an other argument that is the `target actor's identity` which can be confused as a `userIdentity`.

> ```go
> testUtils.AssertNodeIdentity {
>     NodeID: 0,
>     ExpectedIdentity: 0,
> }
> ```
> 
> I'm not sure it's worth the effort. And I don't see how it will improve the test.


As for the test it's an interesting one for sure I am not sure off the top of my head what the best approach might be, but here are 4 test cases that I would like to see definitely see tested:

- If no `node-identity-key` is not given on start then
	- keys are randomly generated (can just assert it is non-empty)
	- the keys stay the same even after database restart (can assert that the key from the above case is same as after restart)
	
- If `node-identity-key` is given on start then
	- keys are actually set based on that given key
	- the keys stay the same even after database restarts to that same key.
	
If you feel that setting integration tests might be a bit useless and not worth the effort I can also vibe with just integration tests asserting the above 4 test cases (perhaps might be easier). However maybe others might have other thoughts on this. @AndrewSisley ?

cli/node_identity.go
shahzadlone: > plans to also yield the private key too in a follow-up PR.

We should not be dumping priv key IMO, I raised that previously on this PR as well.

tests/integration/acp.go
```diff
@@ -357,137 +266,101 @@ type DeleteDocActorRelationship struct {
...
+	boolDesc string,
+) {
+	expectedErrorRaised := AssertError(s.t, s.testCase.Description, actualErr, expectedError)
+	assertExpectedErrorRaised(s.t, s.testCase.Description, expectedError, expectedErrorRaised)
+
+	if !expectedErrorRaised {
+		require.Equal(s.t, expectedError, "")
+		require.Equal(s.t, expectedBool, actualBool, boolDesc)
+	}
+}
```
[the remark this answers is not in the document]
islamaliev: done

tests/integration/utils.go
islamaliev: > Interesting, why is it in a separate PR? I would really like us to not mix both these different kinds of identities. I do slightly prefer the name requestingIdentities or requestorIdentities over userIdentities because incase of relationship management we take an other argument that is the target actor's identity which can be confused as a userIdentity.

It's not 100% ready yet and I wasn't sure it's something we really want. I think it makes sense to create a draft after we merge this one so that you can take a quick look before I go far with it

cli/start.go
AndrewSisley: Dev/test environments can persist for quite a while, it is not uncommon to have permanent environments for that.  At one org I worked for each dev had their own permanent, always running, dev deployment (web server, plus infrastructure).

tests/integration/utils.go
shahzadlone: > It's not 100% ready yet and I wasn't sure it's something we really want. I think it makes sense to create a draft after we merge this one so that you can take a quick look before I go far with it

Ah cool. Happy to have it in separate PR. Is definitely best IMO to keep request's identities used for acp separate from the node's identity. Looking forward for that draft PR.

shahzadlone: Wondering what about the test cases I mentioned, would those be after in that PR?

islamaliev: as discussed, created a task to cover `start` command with tests.

shahzadlone (approved):
Thanks for doing the requested changes and bearing with the review suggestions :)

Just one nitpick, perhaps rename the PR title with first word as an action like `feat: Add node identity`

cli/assign_node_identity.go
```diff
@@ -0,0 +1,64 @@
...
+	var cmd = &cobra.Command{
+		Use:   "assign-node-identity [identity]",
+		Short: "Assign an identity to the node",
+		Long: `Assign an identity to the node.
+
+Identity is hex-formatted private key.
+Node uses the identity to be able to exchange encryption keys with other nodes.
+		
+Example to assign an identity to the node:
+  defradb client assign-node-identity 028d53f37a19afb9a0dbc5b4be30c65731479ee8cfa0c9bc8f8bf198cc3c075f
```
shahzadlone, line 35: suggestion:

Why not, for this do: `defradb client node-identity assign 028d53f37a`

And for the other one do: ``defradb client node-identity get``

cli/utils.go
```diff
@@ -168,7 +169,7 @@ func setContextIdentity(cmd *cobra.Command, privateKeyHex string) error {
 		return err
 	}
 
-	ctx := db.SetContextIdentity(cmd.Context(), immutable.Some(identity))
+	ctx := identity.WithContext(cmd.Context(), immutable.Some(ident))
```
shahzadlone, line 172: question: Sorry reviewed a bit fast, lost context, what happened to `db.SetContext...`?

internal/db/collection_acp.go
```diff
@@ -36,10 +37,9 @@ func (c *collection) registerDocWithACP(
 	if !c.db.acp.HasValue() {
 		return nil
 	}
-	identity := GetContextIdentity(ctx)
 	return permission.RegisterDocOnCollectionWithACP(
 		ctx,
-		identity,
+		identity.FromContext(ctx),
```
shahzadlone, line 42: question: Why was this needed?

internal/db/collection_get.go
```diff
@@ -59,11 +60,10 @@ func (c *collection) get(
 	showDeleted bool,
 ) (*client.Document, error) {
 	txn := mustGetContextTxn(ctx)
-	identity := GetContextIdentity(ctx)
 	// create a new document fetcher
 	df := c.newFetcher()
 	// initialize it with the primary index
-	err := df.Init(ctx, identity, txn, c.db.acp, c, fields, nil, nil, false, showDeleted)
+	err := df.Init(ctx, identity.FromContext(ctx), txn, c.db.acp, c, fields, nil, nil, false, showDeleted)
```
shahzadlone, line 66: question: What made you do this?

tests/clients/http/wrapper.go
```diff
@@ -284,5 +284,9 @@ func (w *Wrapper) Host() string {
 }
 
 func (w *Wrapper) GetNodeIdentity(ctx context.Context) (immutable.Option[identity.PublicRawIdentity], error) {
-	return w.node.DB.GetNodeIdentity(ctx)
+	return w.client.GetNodeIdentity(ctx)
+}
+
+func (w *Wrapper) AssignNodeIdentity(ctx context.Context, ident identity.Identity) error {
+	return w.client.AssignNodeIdentity(ctx, ident)
 }
```
shahzadlone: question: We want to be able to assign a node identity through Http?

tests/integration/state.go
```diff
@@ -132,10 +132,17 @@ type state struct {
 	// Identities by node index, by identity index.
 	identities [][]identity.Identity
 
-	// Will recieve an item once all actions have finished processing.
+	// identitiesByName
```
shahzadlone, line 135: todo: Missing documentation?

tests/integration/state.go
```diff
@@ -132,10 +132,17 @@ type state struct {
...
+	identitiesByName map[string]identity.Identity
+
+	// The seed for the next node identity generation. It starts at 0x7fffffff to avoid
+	// collisions with the default identities.
+	nextNodeIdentityGenSeed int
```
shahzadlone, line 140: todo: can you describe the purpose? I don't quite understand why this is needed?

tests/integration/test_case.go
```diff
@@ -779,7 +779,26 @@ type BackupImport struct {
...
-// It asserts that the `GetNodeIdentity` method returns an identity of the node that matches
-// the first generated identity (with index 0 in the nodes array of identities) for the node.
-type CheckNodesIdentities struct{}
+// GetNodeIdentity is an action that calls the [DB.GetNodeIdentity] method and asserts the result.
+// It checks if a node at the given index has an identity matching another identity under the same index.
+type GetNodeIdentity struct {
+	// NodeID holds the ID (index) of a node to get the identity from.
+	NodeID int
+
+	// ExpectedIdentityName is the name associated with the expected identity of the node.
```
shahzadlone, line 788: suggestion: Perhaps worsened by the previous missing documentation for this attribute, but can you describe what `IdentityName` is? I am specially wondering about the `name` part as opposed to what we have before in tests as `identity Index`

shahzadlone (changes requested):
I do like some of the new changes haha, but I hope you are okay with the delay in this PR as it has now got some new features. Would suggest asking the previous reviewer to give a quick re-review as well.

question: Just to confirm the separation of nodes' and users' identities will be done later and not in this PR correct?

cli/assign_node_identity.go
```diff
@@ -0,0 +1,64 @@
...
+
+			db := mustGetContextDB(cmd)
+			data, err := hex.DecodeString(args[0])
+			if err != nil {
+				return err
+			}
+			privKey := secp256k1.PrivKeyFromBytes(data)
+			identity, err := acpIdentity.FromPrivateKey(
+				privKey,
+				time.Hour*24*365*10, // 10 years
```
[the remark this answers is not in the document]
islamaliev: I have a note to discuss it on standup :) Tell me if you know what's best in this case.

cli/assign_node_identity.go
islamaliev: I like that

cli/utils.go
islamaliev: I don't see a reason to depend on `db` just to get an identity from a `Context`

tests/clients/http/wrapper.go
islamaliev: that's what I also wanted to discuss on stand up

internal/db/collection_acp.go
islamaliev: nothing changed here, akhi. `identity` used to be a variable, now it's package.

internal/db/collection_get.go
islamaliev: if you asked why I move it to `identity` package? So that other package don't depend on `db` just to get an identity from a `Context`

cli/assign_node_identity.go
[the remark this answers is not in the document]
AndrewSisley: I guess this doesn't need to be too accurate lol, could maybe make it `time.Hour*24*365.256*10`? :rofl:

tests/clients/http/wrapper.go
shahzadlone: okie dokes can leave it for then

cli/utils.go
shahzadlone: Ah I see! thanks for clarification.

internal/db/collection_get.go
shahzadlone: gotchu, thanks

tests/integration/state.go
islamaliev: I added some more docs. Not sure what else to add.

tests/integration/state.go
islamaliev: done

AndrewSisley (changes requested):
todo: This still feels out of step with how I thought it should behave, especially given the 10 year expiry on the jws token.

I think the `identity.Identity` should be constructed within `db.GetNodeIdentity`, with a 15min (default) expiry, the only thing held in state on the `db` object should be the keyring/name-of-key-in-keyring (could be const for now).

Very much worth checking in with the rest of the team on this in case I'm talking nonsense :)

tests/integration/acp.go
shahzadlone, line 1: question: What made you convert identities to 1d?

tests/integration/acp.go
```diff
@@ -161,13 +159,13 @@ type AddDocActorRelationship struct {
 	// The target public identity, i.e. the identity of the actor to tie the document's relation with.
 	//
 	// This is a required field. To test the invalid usage of not having this arg, use -1 index.
-	TargetIdentity int
+	TargetIdentity identRef
```
shahzadlone, line 162: question: why do these need to be custom types now, instead of just int indexes? Doesn't `optional` struct work if you keep them both as separate identities instead of making them 1D?

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
```diff
@@ -25,7 +23,7 @@ func TestACP_AddPolicy_ExtraPermissionsAndExtraRelations_ValidPolicyID(t *testin
 
 		Actions: []any{
 			testUtils.AddPolicy{
-				Identity: immutable.Some(1),
+				Identity: testUtils.UserIdentity(1),
```
shahzadlone, line 26: suggestion: I would like to avoid this if possible.

tests/integration/acp.go
```diff
@@ -112,7 +112,7 @@ func addPolicyACP(
 
 	nodeIDs, nodes := getNodesWithIDs(action.NodeID, s.nodes)
 	for index, node := range nodes {
-		ctx := getContextWithIdentity(s, action.Identity, nodeIDs[index])
+		ctx := getContextWithIdentity(s.ctx, s, action.Identity, nodeIDs[index])
```
shahzadlone, line 115: question: Curious why this was done?

http/client.go
```diff
@@ -525,7 +525,7 @@ func (c *Client) GetNodeIdentity(ctx context.Context) (immutable.Option[identity
 }
 
 func (c *Client) AssignNodeIdentity(ctx context.Context, ident identity.Identity) error {
-	methodURL := c.http.baseURL.JoinPath("node", "assign-identity")
+	methodURL := c.http.baseURL.JoinPath("node", "identity")
```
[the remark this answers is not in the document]
islamaliev: well, it should be actually removed

tests/clients/http/wrapper.go
```diff
@@ -288,5 +288,5 @@ func (w *Wrapper) GetNodeIdentity(ctx context.Context) (immutable.Option[identit
...
 
 func (w *Wrapper) AssignNodeIdentity(ctx context.Context, ident identity.Identity) error {
-	return w.client.AssignNodeIdentity(ctx, ident)
+	return w.node.DB.AssignNodeIdentity(ctx, ident)
```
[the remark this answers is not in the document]
islamaliev: beacuse we don't allow assigning node identities over http

shahzadlone:
question: Why did we want to have the `set`/`assign` command as a separate command and not just a function of start?

tests/integration/acp.go
islamaliev: with just an index you can not differenciate between a node identity and a user identity (e.g. an identity provided externally).

tests/integration/acp.go
islamaliev: first of all, they are not quite straightforward to understand. It look me awhile to figure out that the 2D array is needed only to accomodate the same node identities for different targets, i.e. with different audience encoded in token.
And if token don't need to be generated (e.g. with go client) then all the identities are the same. That was confusing.
Now with this change we generate the identity itself and before sending a request, if there is a need to generate a token, we do it and cache it.

tests/integration/acp.go
islamaliev: what exactly? We need a context with an identity in it for sending a request.
Please clarify your question. I also documented some of my decisions. Let me know, if they are not clear enough.

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
islamaliev: But we want to differenciate between node identities and externally provided udentities. How else do you want to do it?
Yes, this make it more verbose, but explicit at the same time. You like "explicit", don't you? :)

islamaliev:
> todo: This still feels out of step with how I thought it should behave, especially given the 10 year expiry on the jws token.
> 
> I think the `identity.Identity` should be constructed within `db.GetNodeIdentity`, with a 15min (default) expiry, the only thing held in state on the `db` object should be the keyring/name-of-key-in-keyring (could be const for now).
> 
> Very much worth checking in with the rest of the team on this in case I'm talking nonsense :)

I removed the duration. And actually I removed `node-identity assign` command altogether.

Posting [my quote](https://discord.com/channels/427944769851752448/887814166331662396/1298596665090904136) from discord with explanation
> initially I created it, because I stumbled upon a problem in our testing framework. It was not possible to provide an identity on startup, because it required audience first which could be retrieved only after a node is created. Now thinking about it after I split identity creation and token generation, it might be possible. I will check it again. Maybe we don't need it anymore.

http/auth.go
```diff
@@ -18,8 +18,8 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/sourcenetwork/immutable"
 
+	"github.com/sourcenetwork/defradb/acp/identity"
 	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
```
fredcarle, line 22: todo: You're importing the same package twice under a different name. Please choose only one. (My preference is to just use "identity" instead of "acpIdentity")

tests/integration/identity.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"context"
+	"math/rand"
+
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/sourcenetwork/immutable"
+	"github.com/stretchr/testify/require"
+
+	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
+	identity "github.com/sourcenetwork/defradb/acp/identity"
```
fredcarle, line 22: todo: Importing the same package twice. Please choose one.

tests/integration/acp.go
shahzadlone: question: does it effect the test run times?

I did prefer the 2D approach because the token (generation stuff) would be done once and stored. Rather than be done every time an identity token is needed and done on the fly.

I guess I was a bit confused because I thought we discussed here: https://github.com/sourcenetwork/defradb/pull/3125#discussion_r1803180817

That we will keep the node identity implementation separate. Why can't we keep the old identity stuff the same as it was and just introduce a new variable on `state` like discussed before that keeps identities the way we need them for this implementation, I am really not a big fan of how this new and separate feature makes us have to change the way we use identity for every other identity use case. I am happy if you want to keep that new separate identities even flattened (I see the use case for node identities). 

That way you also won't need the custom struct `identRef` that was introduced, which essentially does what `Option` does with one extra variable. I would much prefer sticking with `Option`.

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
shahzadlone: > But we want to differenciate between node identities and externally provided identities. How else do you want to do it?

I thought I described my prefer way to do it here already and that you were okay with it? https://github.com/sourcenetwork/defradb/pull/3125#discussion_r1803180817

In short, leave previous identities the way they are and make a new one for node Identities rather than trying to have to both fit the same mold.

tests/integration/acp.go
shahzadlone: Answered here: https://github.com/sourcenetwork/defradb/pull/3125#discussion_r1814847264

http/auth.go
shahzadlone: nitpick: I prefer `acpIdentity` haha and use `identity` for var names rather than `ident`

tests/integration/acp.go
islamaliev: because in some parts of the code the context is not stored in `s.ctx`. I don't know if there is a good reason for this.
I can actually change it to something like this:
```go
s.ctx = newCtx
updateContextWithIdentity(s, action.Identity, nodeIDs[index]) // will store in s.ctx
policyResult, err := node.AddPolicy(s.ctx, action.Policy)
```
Not sure if that makes it better, but I don't have a strong opinion here.

tests/integration/acp.go
islamaliev: looks like you missed some documentation:
```go
// identityHolder holds an identity and the generated tokens for each target node.
// This is used to cache the generated tokens for each node.
```
A short answer is the run time didn't change. It might have been even slightly improved, as now to tonek is generated before a request (not together with identity).

> Why can't we keep the old identity stuff the same as it was and just introduce a new variable on state like discussed before that keeps identities the way we need them for this implementation

yes, we discussed it and I've done it this way, but I turned out to have 2 almost identical code paths for user and node identities. That's why I decided to change it. Now it feels a bit simpler, because you don't need to bother with keeping arrays consistent.
Another performance note: with old vertion if you requested identity with index 10, you would generate all 10 identities and tokens at once event if the might not need them at all. So there is definitely a performance benefit here.

> That way you also won't need the custom struct identRef that was introduced, which essentially does what Option does with one extra variable. I would much prefer sticking with Option.

But how would you specify if you want to use user or node identity? You can't just provide an index. I'm not claiming that my solution is perfect, but I would love to see some alternatives.

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
islamaliev: You mean to add to every test action an additional field? Like:
```go
Identity immutable.Option[int]
NodeIdentity immutable.Option[int]
```
I don't think it's good design. First of all, you can't have both of them set, you will have to document every action and check on executtion of this action if not more then 1 is set. Another point is, what if in the future we add another type of identity? That's going to hurt.

And anyway all identities we used so far ARE node identities, beucase they are attached to a node. So it should be:
```go
Identity immutable.Option[int]
UserIdentity immutable.Option[int]
```

tests/integration/identity.go
islamaliev: good catch

http/auth.go
islamaliev: done

tests/integration/acp.go
shahzadlone: > because in some parts of the code the context is not stored in `s.ctx`. I don't know if there is a good reason for this. I can actually change it to something like this:
> 
> ```go
> s.ctx = newCtx
> updateContextWithIdentity(s, action.Identity, nodeIDs[index]) // will store in s.ctx
> policyResult, err := node.AddPolicy(s.ctx, action.Policy)
> ```
> 
> Not sure if that makes it better, but I don't have a strong opinion here.

Oh wow, perhaps worth investigating later where and why that happens. I am happy with the suggested solution you gave for places that have missing ctx

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
shahzadlone: That is not what I am proposing.

I am saying leave all actions that existed before to use the identity implementation that existed before and not touch them, all tests should remain the way they were. Just rename the `identities [][]identity.Identity` on `state` struct (i.e. `s.idenities`) to `requestingIdentities` (or `userIdentities`) and add a new variable to the `state` struct called `nodeIdentities ` (or whatever you would like to call it, and make the implementation you have introduced in this PR for example 1D array, do all of that for that identity).

Then only the test.cases that use node identity stuff can just use that new variables identity logic for example: `testUtils.GetNodeIdentity` action can operate over just the `nodeIdentities` and doesn't need to worry about the other identities logic the way we had it before.

This was you don't need the new custom `IdentRef` and know for sure each action case is operating on it's domain of identities just like you said you can't have both of them set at once, so I am very much agains't mixing these.

tests/integration/acp.go
shahzadlone: Is somewhat an overlapping convo here: https://github.com/sourcenetwork/defradb/pull/3125#discussion_r1815300220

tests/integration/acp/add_policy/with_extra_perms_and_relations_test.go
fredcarle: There is a problem with keeping things as is because when we start needing to set node identities to docs and/or policies, things will need to be reworked anyways.

One thing I like about the test changes in this PR is the decoupling of identity generation from token generation. It has the potential to keep things cleaner.

I think the `identRef` approach could be improved though and I have a few ideas to simplify things and make both Shahzad and Islam happy with the outcome. We can discuss during standup.

shahzadlone: > There is a problem with keeping things as is because when we start needing to set node identities to docs and/or policies, things will need to be reworked anyways.

What do you mean by setting node identity to docs/policy? I thought node identity is an identity that will be set once for a node on start (if an admin user chooses to ofc).


> One thing I like about the test changes in this PR is the decoupling of identity generation from token generation. It has the potential to keep things cleaner.

I like the decoupling as well (specially for the type of identity that needs it).

> I think the `identRef` approach could be improved though and I have a few ideas to simplify things and make both Shahzad and Islam happy with the outcome. We can discuss during standup.

I think a middle approach can be to make an optional type that takes a custom type containing the index and the type of identity (i.e. remove the `hasValue` and value type from the custom type). However I just still fail to quite understand why we have to couple both types of identities together

AndrewSisley (approved):
Looks really good Islam - sorry the review stage was a bit of a drag - I think we as a team failed here a little bit and that we all could have saved a fair amount of effort by paying more attention earlier on and discussing things before code was written.

Please make sure the conversation RE testutils internal stuff with Fred is resolved before merge.  My personal opinion on it is that the difference is pretty small and delaying the merge is probably more costly than achieving consensus on what the the perfect solution is there.  Fred's suggestion looks pretty easy to do post-merge too (no test changes).

shahzadlone (approved):
Thanks for bearing with the review.

I do prefer fred's approach as it is very close to what I was proposing: https://github.com/sourcenetwork/defradb/commit/62154fadfe3ee4ff968914ddc03dff1f38727ed7#diff-b4cf51ec3aeceed911432510677147fd0d93a5c02680a4efce39566bfa8ebeb0

It's not worth blocking the PR over though so I am leaving an LGTM :)

nitpick: bit more descriptive PR title, perhaps: `feat: Add Node Identity`

PR #3134, opened 2024-10-15
refactor(i): Lint casts & move panics under `must` funcs

### LINKED ISSUE #3103: Fix StatusBadRequest -> StatusInternalServerError for internal errors in http handler
Suggestion from @nasdf:
https://github.com/sourcenetwork/defradb/pull/3099#discussion_r1787947741

Change: 
``` go
	db, ok := req.Context().Value(dbContextKey).(client.DB)
	if !ok {
		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
		return
	}

```

To return error `500` so roughly this:
```go
	db, ok := req.Context().Value(dbContextKey).(client.DB)
	if !ok {
		responseJSON(rw, http.StatusInternalServerError, errorResponse{NewErrFailedToGetContext("db")})
		return
	}

```

shahzadlone:
## Relevant issue(s)
Resolves #3103 

## Description
- Fix the BadRequest in the acp handlers to now be panics
- Make the panics documented under `must` functions
- Add the linter to avoid increasing of undocumented panics
- Ignore linter in existing files outside cli and http (this should either be resolved gradually over time or in a single PR, regardless would prefer to not block this PR because of them)

### DIFF
--- a/http/handler_acp.go
+++ b/http/handler_acp.go
@@ -22,11 +22,7 @@ import (
 type acpHandler struct{}
 
 func (s *acpHandler) AddPolicy(rw http.ResponseWriter, req *http.Request) {
-	db, ok := req.Context().Value(dbContextKey).(client.DB)
-	if !ok {
-		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
-		return
-	}
+	db := req.Context().Value(dbContextKey).(client.DB)
 
 	policyBytes, err := io.ReadAll(req.Body)
 	if err != nil {
@@ -47,11 +43,7 @@ func (s *acpHandler) AddPolicy(rw http.ResponseWriter, req *http.Request) {
 }
 
 func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
-	db, ok := req.Context().Value(dbContextKey).(client.DB)
-	if !ok {
-		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
-		return
-	}
+	db := req.Context().Value(dbContextKey).(client.DB)
 
 	var message addDocActorRelationshipRequest
 	err := requestJSON(req, &message)
@@ -76,11 +68,7 @@ func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.R
 }
 
 func (s *acpHandler) DeleteDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
-	db, ok := req.Context().Value(dbContextKey).(client.DB)
-	if !ok {
-		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
-		return
-	}
+	db := req.Context().Value(dbContextKey).(client.DB)
 
 	var message deleteDocActorRelationshipRequest
 	err := requestJSON(req, &message)
--- a/http/handler_acp.go
+++ b/http/handler_acp.go
@@ -15,14 +15,12 @@ import (
 	"net/http"
 
 	"github.com/getkin/kin-openapi/openapi3"
-
-	"github.com/sourcenetwork/defradb/client"
 )
 
 type acpHandler struct{}
 
 func (s *acpHandler) AddPolicy(rw http.ResponseWriter, req *http.Request) {
-	db := req.Context().Value(dbContextKey).(client.DB)
+	db := mustGetContextClientDB(req)
 
 	policyBytes, err := io.ReadAll(req.Body)
 	if err != nil {
@@ -43,7 +41,7 @@ func (s *acpHandler) AddPolicy(rw http.ResponseWriter, req *http.Request) {
 }
 
 func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
-	db := req.Context().Value(dbContextKey).(client.DB)
+	db := mustGetContextClientDB(req)
 
 	var message addDocActorRelationshipRequest
 	err := requestJSON(req, &message)
@@ -68,7 +66,7 @@ func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.R
 }
 
 func (s *acpHandler) DeleteDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
-	db := req.Context().Value(dbContextKey).(client.DB)
+	db := mustGetContextClientDB(req)
 
 	var message deleteDocActorRelationshipRequest
 	err := requestJSON(req, &message)
--- a/http/handler_tx.go
+++ b/http/handler_tx.go
@@ -13,12 +13,10 @@ package http
 import (
 	"net/http"
 	"strconv"
-	"sync"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/go-chi/chi/v5"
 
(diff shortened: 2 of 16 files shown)


tools/configs/golangci.yaml
```diff
@@ -151,6 +152,67 @@ issues:
       linters:
         - goheader
 
+    # Exclude running force type assert check in these paths (should likely fix them all)
+    # Note: The last item must not have a `|` at the end otherwise linter ignores everyfile.
+    - path: "(\
```
AndrewSisley, line 157: question: Why would we want this?

AndrewSisley (changes requested):
todo: This appears to be entirely untested, given the lack of any changes made to test files, and no note of any manual testing in the PR description.

I also have a question RE the linter excluded file list.

tools/configs/golangci.yaml
shahzadlone: Is kind of the last PR point "discussion" I wanted to hear opinions about. It's because there are many linter complaints in those files I didn't want to resolve in this PR. These should absolutely be all resolved so this list is empty.

AndrewSisley: Ah my bad, I didn't read the description carefully enough - I'm totally okay with this then, so long as this is documented in the comment, so it is very obvious to anyone reading that they should work towards reducing the list, not expanding it.

AndrewSisley:
> HTTP handler functions changed to follow the same practice as all other existing handlers to panic instead of returning 400.

It was this I was mostly after - might not be worth the hassle of writing tests for though as I guess it is mostly just the http status code that changed, and that's not something the integration test framework can handle atm I think.  Have you manually tested this?

http/handler_acp.go
```diff
@@ -47,11 +41,7 @@ func (s *acpHandler) AddPolicy(rw http.ResponseWriter, req *http.Request) {
 }
 
 func (s *acpHandler) AddDocActorRelationship(rw http.ResponseWriter, req *http.Request) {
-	db, ok := req.Context().Value(dbContextKey).(client.DB)
-	if !ok {
-		responseJSON(rw, http.StatusBadRequest, errorResponse{NewErrFailedToGetContext("db")})
-		return
-	}
+	db := mustGetContextClientDB(req)
```
fredcarle, line 44: question: Why is this turned into a `must` instead of a `try` like what you did is some other handles?

http/middleware.go
```diff
@@ -46,6 +46,43 @@ var (
 	colContextKey = contextKey("col")
 )
 
+// mustGetContextClientCollection returns the client collection from the http request context or panics.
+//
+// This should only be called from functions within the http package.
+func mustGetContextClientCollection(req *http.Request) client.Collection {
+	return req.Context().Value(colContextKey).(client.Collection) //nolint:forcetypeassert
+}
```
fredcarle, line 54: thought: It feels wrong to have these in the `middleware` file. They should probably be moved to the `utils` file instead including the context variables above.

http/handler_tx.go
```diff
@@ -57,41 +55,55 @@ func (h *txHandler) NewConcurrentTxn(rw http.ResponseWriter, req *http.Request)
...
+	txVal, ok := txs.Load(txID)
 	if !ok {
 		responseJSON(rw, http.StatusBadRequest, errorResponse{ErrInvalidTransactionId})
 		return
 	}
-	err = txVal.(datastore.Txn).Commit(req.Context())
+
+	dsTxn, ok := txVal.(datastore.Txn)
+	if !ok {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{ErrInvalidDataStoreTransaction})
```
fredcarle, line 73: todo: This isn't a `BadRequest` but rather an `InternalServerError` (500). 

Out of scope of this PR but a lot of our returned statuses are wrong like this. I create #3139 to track this. If a panic within the handlers automatically returns a 500 status then it might be better to just do a `must` typecast and let the server panic for cases where a 500 is appropriate.

tools/configs/golangci.yaml
shahzadlone: Agreed will document that in the comment :)

- [ ] Document more clearly we need to reduce the list not expand it.

shahzadlone:
> > HTTP handler functions changed to follow the same practice as all other existing handlers to panic instead of returning 400.
> 
> It was this I was mostly after - might not be worth the hassle of writing tests for though as I guess it is mostly just the http status code that changed, and that's not something the integration test framework can handle atm I think. Have you manually tested this?

Manually some yes, but not all. I am a bit unsure how to test certain situations.

http/handler_acp.go
shahzadlone: Only one that remained a `try` (not panic) is the P2P one because I think in that case it makes sense because P2P might be turned off and I see that specific error being returned by original author of the p2p code as useful for that situation as useful.

Rest all context casts are converted to a `must`.

http/middleware.go
shahzadlone: I agree with this thought.

(1) I was tempted to perhaps have a shared `utils` files some how for both `http` and `cli` as cli implements some of these methods too (however will have to make `http` and `cli` under a common parent folder perhaps that has the common `utils` file but we already have a `client` folder which I feel a bit uneasy moving both these into, perhaps another folder?). 

(2) Another approach might be to do what you suggested in a `utils` file similar to how it's done in `cli` package and call it a day.

Can follow consensus, does anyone else have any suggestions/prefs? @islamaliev @nasdf @AndrewSisley

http/handler_tx.go
shahzadlone: Actually #3139 seems similar / duplicate of what I wanted #3103 to be originally (if i understand correctly) . So I would say that is very much in the scope of this PR. I can change all the required `BadRequests` that need changing if you can guide me to which additional ones should be changed (@nasdf suggested only for the `context` gets+casts originally), I can take care of them in this PR.

http/middleware.go
AndrewSisley: I prefer (2) :)

shahzadlone: - [ ] Do (2)

tools/configs/golangci.yaml
shahzadlone: Resolved in commit: `PR(ADHERE): Adhere to code review - better documentation`

http/middleware.go
shahzadlone: Resolved in commit: `PR(ADHERE): Adhere to code review - move to utils`

http/handler_tx.go
nasdf: I agree, allowing a panic or changing to a 500 status would be better.

AndrewSisley (approved):
LGTM, but please make sure Keenan in particular is happy before merge :)

nasdf (approved):
LGTM assuming you fix Fred's remaining todo.

http/handler_tx.go
shahzadlone: > I agree, allowing a panic or changing to a 500 status would be better.

Just so I understand this, which one are you suggesting:
1) all the `ErrInvalidDataStoreTransaction` to panic?
2) all the `http.StatusBadRequest` to panic?
3) only this line the suggestion is on?

nasdf: Anywhere we do this type cast `txVal.(datastore.Txn)`.

shahzadlone: Done in commit: `PR(ADHERE): Adhere to code review - mustify datastore.Txn`

PR #3150, opened 2024-10-16
fix(i): Aggregate json filter

### LINKED ISSUE #3149: Aggregate with JSON object filter does not work
This schema:

```
type Users {
  custom: JSON
}
```

with this query:

```gql
query {
  _count(Users: {filter: {custom: {tree: {_eq: "oak"}}}})
}
```

returns an error:

```
type not found. Type: custom
```

nasdf:
## Relevant issue(s)

Resolves #3149

## Description

This PR fixes an issue with aggregate JSON filtering.

## Tasks

## How has this been tested?

Added integration test.

- MacOS

### DIFF
--- a/internal/planner/mapper/mapper.go
+++ b/internal/planner/mapper/mapper.go
@@ -439,6 +439,7 @@ func resolveAggregates(
 				if err != nil {
 					return nil, err
 				}
+				fixAggregateNestedTargets(childMapping, hostSelectRequest)
 
 				childFields, _, err := getRequestables(
 					ctx,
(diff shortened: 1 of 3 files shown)


internal/planner/mapper/mapper.go
```diff
@@ -569,6 +570,32 @@ func resolveAggregates(
 	return fields, nil
 }
 
+// fixAggregateNestedTargets ensures that selections of
+// JSON objects are not interpreted as joins.
+//
+// This can happen when an aggregate contains a filter
+// on a JSON object, but we can't tell if it is a relation
+// until the child mapping is created.
+func fixAggregateNestedTargets(
```
fredcarle, line 579: suggestion: I got stuck on this function name a little too long for my liking lol. I think the `fix` prefix threw me off. This fonction does something that is very specific to JSON objects so my suggestion is to make that clearer in the name. What we are really doing here is filtering out/removing JSON sub-fields from the selection set.

Also, this is used at the moment specifically for aggregates but there is nothing in the function itself that handles aggregate specific objects.

As a result of the above, I would suggest renaming to `removeJSONSubFields`.

fredcarle (approved):
Thanks for fixing. I just have a suggestion regarding the function name.

tests/integration/query/json/with_aggregate_test.go
```diff
@@ -0,0 +1,55 @@
+package json
```
AndrewSisley, line 1: todo: Please add some more tests.  At least one to cover the `default` case in the switch, and another where `isMapped` is false.  Also consider tests for invalid user filters, although that could arguably be done in the other JSON filter PR.

shahzadlone: +1

nasdf: I've added an additional test to cover the `isMapped` case. The `default` case is technically dead code but removing it could cause a hard to find bug in the future.

internal/planner/mapper/mapper.go
nasdf: Renamed to `removeJSONSubFields` as suggested

tests/integration/query/json/with_aggregate_test.go
AndrewSisley: Okay no worries, I have a very minor preference for an error to be returned from `default` then, but happy as is, so only change it if you like the suggestion.

Thanks for the `isMapped` test :)

AndrewSisley (approved):
LGTM, thanks Keenan!

PR #3162, opened 2024-10-20
test(i): Fix the nodes indexing bug in tests

### LINKED ISSUE #3076: Improve `getNodes` in test utils to return the correct indexes with the relevant nodes

shahzadlone:
## Relevant issue(s)
Resolves #3076 

## Description
- Fix the `getNodes` bug
- Remove the `getNodesCollection`


tests/integration/utils.go
```diff
@@ -566,38 +570,32 @@ func closeNodes(
...
-// The caller must not assume the returned collections are in order of the node index if the specified
-// index is greater than 0. For example if requesting collections with nodeID=2 then the resulting output
-// will contain only one element (at index 0) that will be the collections of the respective node, the
-// caller might accidentally assume that these collections belong to node 0.
-func getNodeCollections(nodeID immutable.Option[int], collections [][]client.Collection) [][]client.Collection {
+// The caller must not assume the returned node's ID is in order of the node's index if the specified nodeID is
+// greater than 0. For example if requesting a node with nodeID=2 then the resulting output will contain only
+// one element (at index 0) caller might accidentally assume that this node belongs to node 0. Therefore, the
+// caller should always use the returned IDs, instead of guessing the IDs based on node indexes.
+func getNodesWithIDs(nodeID immutable.Option[int], nodes []clients.Client) ([]int, []clients.Client) {
```
nasdf, line 589: thought: this could be a custom iterator https://pkg.go.dev/iter that returns nodeID and node.

nasdf (approved):
LGTM

tests/integration/utils.go
shahzadlone: Can do so in future if others feel strongly for it.

PR #3171, opened 2024-10-22
test(i): Refactor tests to avoid asserting policyIDs

### LINKED ISSUE #3070: Unstable policyIDs in next few sourcehub version bumps

shahzadlone:
## Relevant issue(s)
Resolves #3070 

### Problem:
PolicyIDs will be unstable for the next little while and we want to avoid depending on them and having to change and maintain them overtime.

### Solution in this PR:
1) Made it so that `AddPolicy`a action doesn't need to assert the exact `policyID` anymore, it is optional we can still do so for some edge cases if we want but don't have to.

2) Introduced a `PolicyIDs` attribute on `SchemaUpdate` action as an optional list arg which makes it easier to substitute/embed PolicyIDs without depending on the exact policyIDs, the action will sub the policyIDs into the schema string where place holder labels ("%policyID%") are at, without us using explicit policy identifiers.
	
## Note:
	- The number of placeholder labels (`%policyID%`) must exactly match the number of policyID
	  indexes in this list.
	- The list must contain the policyID Indexes in the same order they are to be substituted in.
	- Can substitute the same policyID at an index, by specifying that index multiple times.
	- The indexes must be valid (correspond to the number of policies added so far).
	
### Consider we have two policies that are added resulting in the following policyIDs:

```go	
	PolicyID1="xyz1"
	PolicyID2="xyz2"
	i.e. -> s.PolicyIDs[0] = {PolicyID1, PolicyID2}
	
	Then using this attribute like:
	PolicyIDs: immutable.Some([]int{0, 0, 1}),
```
	
### On a Schema string like (before transformation):
```go
	 type Users1 @policy(id: "%policyID%", resource: "users") {
	 	name: String
	 	age: Int
	 }
	
	 type Users2 @policy(id: "%policyID%", resource: "users") {
	 	name: String
	 	age: Int
	 }
	 type Users3 @policy(id: "%policyID%", resource: "users") {
	 	name: String
	 	age: Int
	 }
```

#### The Schema that will be loaded will be this modified one:
```go
	 type Users1 @policy(id: "xyz1", resource: "users") {
	 	name: String
	 	age: Int
	 }
	
	 type Users2 @policy(id: "xyz1", resource: "users") {
	 	name: String
	 	age: Int
	 }

	 type Users3 @policy(id: "xyz2", resource: "users") {
	 	name: String
	 	age: Int
	 }
```

## For reviewers:
-The above was changed to the following after the suggestions:
```go
testUtils.SchemaUpdate{
    Schema: `
        type User1 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User2 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User3 @policy(
            id: "{{.Policy1}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }
    `,
    Replace: map[string]any{
        "Policy0": NewPolicyIndex(0),
        "Policy1": NewPolicyIndex(1),
    },
}
```

## How has this been tested?
- I have manually tested the testing framework refactor is accurate, and errors out on the sanity checks if used incorrectly.


tests/integration/acp/demo_test.go
```diff
@@ -0,0 +1,137 @@
...
+					type User3 @policy(
+						id: "%policyID%",
+						resource: "users"
+					) {
+						name: String
+						age: Int
+					}
+				`,
+
+				PolicyIDs: immutable.Some([]int{0, 0, 1}),
```
fredcarle, line 131: suggestion: Having an index of the policyIDs is an nice addition. However, I find that this pattern is not very intuitive. Maybe if we made this a bit more generic it could even be useful for other dynamic assignments. My suggestion would be to change `PolicyIDs` in `SchemaUpdate` to a `Replace` field that holds a `map[string]any`. The outcome would give us something like this:
```go
testUtils.SchemaUpdate{
	Schema: `
		type Users1 @policy(
			id: "policy_0",
			resource: "users"
		) {
			name: String
			age: Int
		}

		type User2 @policy(
			id: "policy_0",
			resource: "users"
		) {
			name: String
			age: Int
		}

		type User3 @policy(
			id: "policy_1",
			resource: "users"
		) {
			name: String
			age: Int
		}
	`,

	Replace: immutable.Some(map[string]any{
        "policy_0": NewPolicyIndex(0),
        "policy_1": NewPolicyIndex(1),
    }),
}
```

shahzadlone: > ```go
> NewPolicyIndex(0)
> ```

While I do like the flexibility of the high-level suggestion, might more interesting to see if the usage with the util function `NewPolicyIndex(0)` is nicer (i am unsure atm).

Brainstorming a bit here, so the `Replace` command will do all the replacing and handle "all substitution" types (like  fetching policyIDs at Index x y z, or anything else in the future), for which we would need to define all these new index types? like one for policy and one for each other substitution types, etc?

But I guess I see the benefits are more than the cons above.

 If others are happy I can try to move to this proposed approach

@sourcenetwork/database-team Would be nice to hear prefs others have

fredcarle: > for which we would need to define all these new index types

It's just a switch statement in the replace logic. We already have the `NewFooIndex` functions for docIDs and soon CIDs. One of the switch items could even be `string` to be able to substitute a manually defined string.

tests/integration/test_case.go
```diff
@@ -129,6 +129,51 @@ type SchemaUpdate struct {
...
+	//		name: String
+	//		age: Int
+	//	}
+	//
+	//	type Users2 @policy(id: "94eb195c0e459aa79e02a1986c7e731c5015721c18a373f2b2a0ed140a04b454", resource: "users") {
+	//		name: String
+	//		age: Int
+	//	}
+	// ```
+	PolicyIDs immutable.Option[[]int]
```
AndrewSisley, line 175: suggestion: I don't think this being `immutable.Option` adds anything besides complexity - there appears to be no behavioural differences between `immutable.None` and an empty slice.

Same goes for Fred's suggestion about decoupling this from `policyID`.

tests/integration/acp.go
```diff
@@ -107,12 +112,23 @@ func addPolicyACP(
...
+	if action.ExpectedError != "" && !action.ExpectedPolicyID.HasValue() {
 		require.Fail(s.t, "Expected error should not have an expected policyID with it.", s.testCase.Description)
 	}
 
 	nodeIDs, nodes := getNodesWithIDs(action.NodeID, s.nodes)
+	maxNodeID := slices.Max(nodeIDs)
+	policyIDsAddedSoFar := len(s.policyIDs)
+	// Expand the policyIDs slice once, so we can minimize how many times we need to expaind it.
+	// We use the maximum nodeID provided to make sure policyIDs slice can accomodate upto that nodeID.
+	if policyIDsAddedSoFar <= maxNodeID {
```
islamaliev, line 125: nitpick: would prefer `if len(s.policyIDs) <= maxNodeID` as the variable doesn't add any clarity.

tests/integration/acp/demo_test.go
islamaliev: I'd like to propose using Go templates as an alternative approach. It would make our schema definitions more readable and maintainable. 
Many developers are already familiar with native Go template syntax. We can easily extend it with custom template functions for our specific needs.
Here's a simple example:
```go
testUtils.SchemaUpdate{
    Schema: `
        type User1 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User2 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User3 @policy(
            id: "{{.Policy1}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }
    `,
    TemplateData: map[string]string{
        "Policy0": "policy_0",
        "Policy1": "policy_1",
    },
}
```

We can even do some perversions with it (I'm not advocating):
```go
testUtils.SchemaUpdate{
    Schema: `
        {{- define "userType" -}}
        type {{.Name}} @policy(
            id: "{{.PolicyID}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }
        {{- end -}}

        {{template "userType" dict "Name" "User1" "PolicyID" (policyIndex 0)}}
        {{template "userType" dict "Name" "User2" "PolicyID" (policyIndex 0)}}
        {{template "userType" dict "Name" "User3" "PolicyID" (policyIndex 1)}}
    `,
    TemplateFuncs: template.FuncMap{
        "policyIndex": func(i int) string {
            return fmt.Sprintf("policy_%d", i)
        },
        "dict": func(values ...any) (map[string]any, error) {
            if len(values)%2 != 0 {
                return nil, errors.New("dict requires pairs")
            }
            dict := make(map[string]any, len(values)/2)
            for i := 0; i < len(values); i += 2 {
                key, ok := values[i].(string)
                if !ok {
                    return nil, errors.New("dict keys must be strings")
                }
                dict[key] = values[i+1]
            }
            return dict, nil
        },
    },
}
```
It doesn't look very straightforward, at least not to the level we want to stick to in out tests. 
It will just give us a lot more room for creativity in the future.

islamaliev: well, it can be `NewPolicyIndex(0)` or whatever else we decide.
I will update my proposal to make it complete.

shahzadlone:
Accepting @fredcarle's suggestion to make it a more generic `Replace` argument using `NewPolicyIndex(...)`
Accepting @AndrewSisley's suggestion to remove the `immutable.Some` wrap and make a `map[string]any`
Accepting @islamaliev's suggestion to use go templates approach `{{.XYZ}}`

So we have:

```go
testUtils.SchemaUpdate{
    Schema: `
        type User1 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User2 @policy(
            id: "{{.Policy0}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }

        type User3 @policy(
            id: "{{.Policy1}}",
            resource: "users"
        ) {
            name: String
            age: Int
        }
    `,
    Replace: map[string]any{
        "Policy0": NewPolicyIndex(0),
        "Policy1": NewPolicyIndex(1),
    },
}
```

tests/integration/acp.go
shahzadlone: - done

tests/integration/acp/add_policy/with_multi_policies_test.go
shahzadlone, line 1: Note: This is the only file now that will have `policyID` asserted manually in integration tests.

tests/integration/acp/index/create_test.go
```diff
@@ -22,21 +22,24 @@ func TestACP_IndexCreateWithSeparateRequest_OnCollectionWithPolicy_NoError(t *te
...
-				ExpectedPolicyID: "94eb195c0e459aa79e02a1986c7e731c5015721c18a373f2b2a0ed140a04b454",
+				Identity: testUtils.ClientIdentity(1),
+				Policy:   userPolicy,
 			},
 
 			testUtils.SchemaUpdate{
 				Schema: `
 					type Users @policy(
-						id: "94eb195c0e459aa79e02a1986c7e731c5015721c18a373f2b2a0ed140a04b454",
+						id: "{{.Policy0}}",
```
AndrewSisley, line 32: suggestion: I think we can make this easier to use and read by removing `Replace` and selecting the value to insert based on the inner string value.

For example:
```
id: "{{.PolicyID.0}}",
``` 
Would be transformed to use `testUtils.NewPolicyIndex(0)`, without the tests having to spend an extra 4 lines to do so.

The same pattern would be fairly easy to extend to other actions, and could even support them automatically if you follow a similar wrapping pattern to that used by `corekv.TxnAction` - this way the `SchemaUpdate` action never needs to complicate itself with substitution logic, and we as maintainers are somewhat guaranteed a consistent result.

AndrewSisley: You leave me wanting to request changes here Shahzad :)

I understand you may be feeling some time pressure, but in some sense #3521 is requesting that this PR gets redone, because the new issue encompasses almost everything within this PR.
 
I really don't think we should be making this a habit, as it calls in to question what the point of the reviewing is if authors start creating new issues to redo the vast majority of the work that they are trying to commit instead of improving within the PR.

shahzadlone: The current change was done, after a consensus was reached in the standup, people preferred Islam's `template suggestion`. Also was discussed here (you have given a thumbs up on the `Replace` suggestion): https://github.com/sourcenetwork/defradb/pull/3171#discussion_r1814171193

Now that all of that is done, a completely new request is being made, which undoes that entire initial consensus.

AndrewSisley: I understand that it can be frustrating, but the PR has been open for 3 and a half months and new ideas are going to pop out.

shahzadlone: > I understand that it can be frustrating, but the PR has been open for 3 and a half months and new ideas are going to pop out.

Can you expand on your solution/idea again? It seems that you are wanting me to undo/remove the template approach (is that correct? or you have a suggestion to also keep the template approach and then made that work?)

AndrewSisley: I think it is still a template approach, but instead of passing stuff to the action, the template targets `state`.

So if state is defined as:
```
type state struct {
   ...
   PolicyIDs []string
}
```
The template would be defined as:
```
id: "{{.PolicyIDs.0}}",
```

tests/integration/utils.go
```diff
@@ -1011,9 +1013,65 @@ func updateSchema(
...
+		var modifiedSchema = action.Schema
+
+		// We need to substitute the policyIDs into the `%policyID% place holders.
+		if len(action.Replace) > 0 {
+			nodeID := nodeIDs[index]
+			nodesPolicyIDs := s.policyIDs[nodeID]
+			templateData := map[string]string{}
+			for substituteLabel, replaceWith := range action.Replace {
+				switch repType := replaceWith.(type) {
+				case replacePolicyIndex:
```
AndrewSisley, line 1046: suggestion: I think the `Replace` interface should be tweaked along with the `replacePolicyIndex` implementation of that interface so that this block is handled by `replacePolicyIndex.Replace`.

By doing it with a switch here we are missing out on a lot of the benefits of using an interface.

shahzadlone: This is a nice suggestion. I will do.

- [x] remove switch and use the new interface method. Done in last commit.

AndrewSisley: Looks good, thanks Shahzad!  I have a quick follow up suggestion though - `Replacer.Value()` is no longer called and can be removed I think.

tests/integration/acp/index/create_test.go
shahzadlone: I am leaving this one out of scope of this PR.

tests/integration/utils.go
shahzadlone: it is called here:  `policyReplaceIndex, ok := r.Value().(int)`

However can be removed and instead can use `r.value`.

- [x] will do

AndrewSisley: It is calling the concrete `Value()` function there, you can leave `replacePolicyIndex.Value()` there if you like, but there is no reason to keep `Replacer.Value()` atm.

shahzadlone: Oh yea you were referring about the interface one. Yea I removed it completely, is not needed. Using `r.value` instead

PR #3173, opened 2024-10-22
test: Allow referencing of Cids by index in test framework

### LINKED ISSUE #3172: Add `CidIndex` magic to test framework to avoid hardcoding of test CIDs

AndrewSisley:
## Relevant issue(s)

Resolves #3172

## Description

Allows referencing of Cids by index in test framework and migrates a handful of tests to the new system.

Migrating all the tests to this system is out of scope.  New tests can now be written using this system where appropriate.

Does not appear to have a significant effect on test execution time, there are a few things we can do to reduce this if we want to in the future.

## How has this been tested?

I verified that specifying the incorrect index(es) results in test failure.

### DIFF
--- a/tests/integration/state.go
+++ b/tests/integration/state.go
@@ -175,6 +175,9 @@ type state struct {
 	// nodes.
 	docIDs [][]client.DocID
 
+	// CIDs by index, by field name, by doc index, by collection index.
+	cids [][]map[string][]cid.Cid
+
 	// Indexes, by index, by collection index, by node index.
 	indexes [][][]client.IndexDescription
 
--- a/tests/integration/test_case.go
+++ b/tests/integration/test_case.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/internal/core"
 	"github.com/sourcenetwork/defradb/net"
 	"github.com/sourcenetwork/defradb/tests/gen"
 	"github.com/sourcenetwork/defradb/tests/predefined"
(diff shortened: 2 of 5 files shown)


tests/integration/utils.go
```diff
@@ -1271,8 +1274,59 @@ func createDoc(
...
+			docIndex := startDocIndex + relativeDocIndex
+
+			if s.cids[action.CollectionID][docIndex] == nil {
+				s.cids[action.CollectionID][docIndex] = map[string][]cid.Cid{}
+			}
+
+			// This is a new doc, so we can blindly overwrite the array of cids
+			s.cids[action.CollectionID][docIndex][key.FieldID] = []cid.Cid{
+				key.Cid,
+			}
```
fredcarle, line 1326: thought: I assume that since this is a create function, this will never be false. You could then just do this:
```go
s.cids[action.CollectionID][docIndex] = map[string][]cid.Cid{
	key.FieldID: {key.Cid},
}
```

tests/integration/state.go
```diff
@@ -175,6 +175,9 @@ type state struct {
 	// nodes.
 	docIDs [][]client.DocID
 
+	// CIDs by index, by field name, by doc index, by collection index.
```
fredcarle, line 178: todo: `by field name` should say `by field id`. I was wondering why you where asserting `field name` with value `"1"` and `"2"` until I realized it's the `field id` that is used.

tests/integration/test_case.go
```diff
@@ -348,6 +349,50 @@ func NewDocIndex(collectionIndex int, index int) DocIndex {
...
+	// CollectionIndex is the index of the collection used to create the document that the Cid belongs to.
+	CollectionIndex int
+
+	// DocIndex is the index of the document that the Cid belongs to.
+	DocIndex int
+
+	// FieldName is the name of the field that this cid is for.
+	//
+	// Composite commits can be targetted with "C".
+	FieldName string
```
fredcarle, line 364: todo: Similar to the previous comment, please use `FieldID`.

fredcarle (commented):
I'm happy with this new test feature. The headsotre lookup, like you said, isn't expensive so it shouldn't have a important impact on test run time.

I'll approve once it's out of draft.

Do you understand what's causing the change detector issue?

AndrewSisley:
> Do you understand what's causing the change detector issue?

Yes, but there are no 'good' solutions, I plan on building the least bad option today and see what people think.

AndrewSisley:
I have a better idea, and it will not look anything like this.  Closing PR, sorry for the bother.

PR #3198, opened 2024-10-29
refactor: Breakup core/keys.go file

### LINKED ISSUE #3197: Breakup core/keys.go

AndrewSisley:
## Relevant issue(s)

Resolves #3197

## Description

Extracts core/keys.go to multiple files.

Original file had grown way to large and was getting quite painful to find stuff in it, and I need to add new types to it shortly.


internal/keys/datastore.go
```diff
@@ -0,0 +1,15 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package keys
```
fredcarle, line 11: todo: Please use the singular for package name. `key`

internal/keys/datastore_doc.go
```diff
@@ -0,0 +1,292 @@
...
+	// PriorityKey is a type that represents a priority instance.
+	PriorityKey = InstanceType("p")
+	// DeletedKey is a type that represents a deleted document.
+	DeletedKey = InstanceType("d")
+
+	DATASTORE_DOC_VERSION_FIELD_ID = "v"
+)
+
+// DataStoreKey is a type that represents a key in the database.
+type DataStoreKey struct {
```
fredcarle, line 37: suggestion: Since the package name is `keys` (should be `key`), the `Key` suffix is less relevant. I would suggest removing it to get `key.DataStore`. This applies for all keys.

fredcarle (changes requested):
I was about to reoganise keys within `core` but I think putting them in their own package makes a lot of sense. Just a todo and a suggestion before I approve.

internal/keys/datastore.go
AndrewSisley: why?  Go has lots of packages with plural names (e.g. `strings`) and there is nothing in https://go.dev/blog/package-names against it.  `keys` contains many different key-types, and according to the norms I have experienced, the name should be plural.

internal/keys/datastore_doc.go
AndrewSisley: Yes, I agree, but I didn't want to bother dealing with the many niggles like this already present in the code I moved.  I just want to break it up so I can add new stuff in a clean way.

internal/keys/datastore.go
fredcarle: The packages in Go that use plural names are usually plural because the singular form is a reserved word or keyword (strings, types, errors, bytes).

It is generally preferred in the Go community to use singular names for packages. We've discussed this plenty of times before.

internal/keys/datastore_doc.go
fredcarle: We've been doing this often lately — refactoring but not cleaning things up fully. It feels like a bad habit that's setting in that might hurt us in the future. This one is minor but still part of the habit so I'm pointing it out.

AndrewSisley: Code is never cleaned up fully.  Leaving as is.

shahzadlone: > We've been doing this often lately — refactoring but not cleaning things up fully. It feels like a bad habit that's setting in that might hurt us in the future.

I am curious to know where we have done this lately? We should def have issues to track any pending cleanups if there were any.

internal/keys/datastore.go
AndrewSisley: Discussed over discord, and noted here for future reference:

When attempting to rename to `key`, I found this caused a large amount (+100) of variable shadowing (most places using this package want variables with the name `key`).  Even if consensus is reached RE avoiding plural package names, we probably still want this to be called `keys` anyway (quite possibly the same reason why many std lib packages have plural names).

PR #3200, opened 2024-10-31
refactor: Simplify merkle/crdt code

### LINKED ISSUE #3199: Further simplify merkle/crdt package

AndrewSisley:
## Relevant issue(s)

Resolves #3199

## Description

Simplifies merkle/crdt code via a handful of small changes.

It should be possible to remove the package and it's types entirely, but I want to get a little deeper into branchable collections before I commit to doing that.


shahzadlone (approved):
LGTM

internal/db/fetcher/versioned.go
```diff
@@ -377,32 +387,25 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 }
 
 func (vf *VersionedFetcher) processBlock(
```
fredcarle, line 389: suggestion: To make the function name more representative of what it has become, it would make sense to rename it to `processFieldBlock`. This way there will be no confusion as to what it's for.

AndrewSisley: Do you mind if I remove this function entirely?

fredcarle: I don't.

AndrewSisley: Thanks :)

internal/db/fetcher/versioned.go
```diff
@@ -182,7 +181,7 @@ func (vf *VersionedFetcher) Start(ctx context.Context, spans core.Spans) error {
 	}
 
 	vf.ctx = ctx
-	vf.dsKey = dk
+	vf.dsKey = dk.WithCollectionRoot(vf.col.Description().RootID)
```
fredcarle, line 184: question: This change is unexpected and unexplained. Can you describe why this was needed?

fredcarle (approved):
This is a nice change 👍

internal/db/fetcher/versioned.go
AndrewSisley: Ah sorry - I see this was fairly inexplicable from the PR - previously this was done (in quite a complex way) by the `base.MakePrimaryIndexKeyForCRDT` (within the iteration loop).  Pulling it out here aided the simplification of `processBlock`.

PR #3201, opened 2024-11-01
feat: Filter alias target

### LINKED ISSUE #3194: Filter alias targeting
Add support for filtering by aliased fields. This does not include aggregates.

nasdf:
## Relevant issue(s)

Resolves #3194

## Description

This PR adds alias targeting in filters. 

**Aggregate targets are not included in this PR as they require more changes.**

## Tasks

## How has this been tested?

Added and updated integration tests.

- MacOS


internal/planner/mapper/targetable.go
```diff
@@ -85,6 +85,9 @@ func (k *ObjectProperty) GetProp(data any) any {
 	if data == nil {
 		return nil
 	}
+	if _, ok := data.(core.Doc); ok {
```
AndrewSisley, line 88: question: Why are you checking for this particular type?

suggestion: Depending on the answer to the above question, it might be better to instead check the success of the cast below:
```
object, ok := data.(map[string]any)
if !ok {
  return nil // this can happen when an alias target does not exist
}
```

internal/planner/mapper/mapper.go
```diff
@@ -992,6 +992,12 @@ func resolveInnerFilterDependencies(
 	newFields := []Requestable{}
 
 	for key, value := range source {
+		// alias fields are guarenteed to be resolved
+		// because they refer to existing fields
```
AndrewSisley, line 996: praise: Thanks for this explanation :)

internal/request/graphql/parser/filter.go
```diff
@@ -93,20 +93,20 @@ func parseFilterFieldsForDescriptionMap(
 	fields := make([]client.FieldDefinition, 0)
 	for k, v := range conditions {
 		switch k {
-		case "_or", "_and":
+		case request.FilterOpOr, request.FilterOpAnd:
```
AndrewSisley, line 96: praise: Thanks for these cleanups :)

tests/integration/query/simple/with_filter/with_alias_test.go
```diff
@@ -0,0 +1,322 @@
...
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"Name": "Bob",
+					"Age": 32
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					Users(filter: {_alias: {}}) {
```
AndrewSisley, line 75: question: Is this consistent with the behaviour of other 'empty' filter properties?

tests/integration/query/simple/with_filter/with_alias_test.go
```diff
@@ -0,0 +1,322 @@
...
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"Name": "Bob",
+					"Age": 32
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					Users(filter: {_alias: null}) {
```
AndrewSisley, line 117: question: I thought we could make input params not-null without making them required?

suggestion: If so, please make the `alias` param not-null, and change this test so that it expects an error :)

tests/integration/query/simple/with_filter/with_alias_test.go
```diff
@@ -0,0 +1,322 @@
...
+			},
+		},
+	}
+
+	executeTestCase(t, test)
+}
+
+func TestQuerySimple_WithNonAliasedField_Succeeds(t *testing.T) {
+	test := testUtils.TestCase{
+		Description: "Simple query with non aliased filter",
```
AndrewSisley, line 200: praise: Thank you for adding this test, and the ones for the compound filters!

tests/integration/query/simple/with_filter/with_alias_test.go
```diff
@@ -0,0 +1,322 @@
...
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"Name": "Bob",
+					"Age": 32
+				}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					Users(filter: {_alias: {UserAge: {_eq: 21}}}) {
```
AndrewSisley, line 183: suggestion: Would it not be better to error in this case? This looks like it would mainly come about through a user-programming error, in which case they would likely much prefer an error over a silent failure.

tests/integration/query/one_to_many/with_count_test.go
```diff
@@ -118,3 +118,66 @@ func TestQueryOneToManyWithCount(t *testing.T) {
...
+func TestQueryOneToMany_WithCountAliasFilter_Succeeds(t *testing.T) {
+	test := testUtils.TestCase{
+		Description: "One-to-many relation query from many side with count",
+		Actions: []any{
+			testUtils.CreateDoc{
+				CollectionID: 0,
+				Doc: `{
+					"name": "Painted House",
+					"rating": 4.9,
+					"author_id": "bae-e1ea288f-09fa-55fa-b0b5-0ac8941ea35b"
```
AndrewSisley, line 133: todo: Please use `DocIndex` for these IDs instead of strings, it reduces the maintenance and readability overhead quite a bit IMO, it will also force the test to create the docs in a 'correct' order :)

AndrewSisley (changes requested):
> Aggregate targets are not included in this PR as they require more changes.

question: I thought the aggregate filters were of the same type as normal filters, in which case they will now have a (presumably broken) `alias` property on them.  Have you checked what happens when someone tries to use it, and when do you plan on adding support for it?

praise: Looks really good Keenan, just a couple of localised suggestions for you - will be really good to have this in - we've been talking about it for a long time now lol and the JSON-type solution is really nice IMO :)

tests/integration/query/simple/with_filter/with_alias_test.go
nasdf: Not-null would mean it is required. It's possible to have not-null variables with default values, maybe that is what you are referring to?

internal/connor/connor.go
```diff
@@ -8,6 +8,10 @@
 // such as JSON can be filtered as expected.
 package connor
 
+import (
+	"github.com/sourcenetwork/defradb/client/request"
```
islamaliev, line 12: todo: connor should not depend on defra

internal/planner/filter/copy_field.go
```diff
@@ -80,6 +80,14 @@ func traverseFilterByProperty(
 				} else if shouldDelete {
 					delete(result, opKey)
 				}
+
+			case map[connor.FilterKey]any:
```
islamaliev, line 84: suggestion: would be great to have this one tested in `copy_field_test.go`

tests/integration/query/simple/with_filter/with_alias_test.go
```diff
@@ -0,0 +1,322 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	executeTestCase(t, test)
+}
+
+func TestQuerySimple_WithNullAlias_Succeeds(t *testing.T) {
```
islamaliev, line 99: suggestion: you have a lot of expected result described as "Succeeds". This is not very descriptive. Would be more readable to explain more specifically the expected behaviour. For example in this case you don't expect filter to match any element, then you can write:
`TestQuerySimple_WithNullAlias_FilterShouldNotMatch`.
The same applied to other occurrences.

islamaliev (commented):
It's a nice feature. Good job Keenan.
I left some minor suggestions/concerns.

internal/connor/connor.go
AndrewSisley: I instinctively wanted to agree and give this comment a thumbs up, but then I wondered why it actually needs to be so now.

Why do you think it should not be dependent on Defra?

islamaliev: AFAIK it was originally third-party dependency.

And there was a time when we played around with transitioning to Rust and we [started with connor package ](https://github.com/sourcenetwork/defradb/pull/2555). This dependency it trivial and should no cause any trouble, as it's just a constant. But other devs might start adding more stuff to it.

internal/connor/connor.go
```diff
@@ -62,7 +66,7 @@ func matchWith(op string, conditions, data any) (bool, error) {
 		return anyOp(conditions, data)
 	case AllOp:
 		return all(conditions, data)
-	case EqualOp:
+	case EqualOp, request.FilterOpAlias:
```
fredcarle, line 69: suggestion: I think it's fine to duplicate the `const` for `_alias` here and avoid the added dependency.

tests/integration/query/simple/with_filter/with_alias_test.go
nasdf: Yes, there's no existing tests but I double checked

AndrewSisley: Nice, thanks Keenan :)

internal/planner/mapper/targetable.go
nasdf: In past reviews you had a preference for specific type checks instead of a single type assertion. I thought I could anticipate this one before review.

tests/integration/query/one_to_many/with_count_test.go
nasdf: done

tests/integration/query/simple/with_filter/with_alias_test.go
nasdf: I have a preference to not return errors from otherwise validly typed queries. I also don't think the added complexity of checking aliases would be worth the benefits.

internal/connor/connor.go
nasdf: I removed the dependency by copying the alias constant.

internal/planner/filter/copy_field.go
nasdf: I've added a test. There was potentially a bug with `_not` filters before I modified it.

tests/integration/query/simple/with_filter/with_alias_test.go
nasdf: I've updated all of the added tests

internal/connor/connor.go
nasdf: done

internal/planner/mapper/targetable.go
AndrewSisley: My preferences probably vary by context.  Here you are casting to the type you actually want the line below, it doesn't seem to make sense to care about some other type as well.

tests/integration/query/simple/with_filter/with_alias_test.go
AndrewSisley: The added complexity for us should pretty minimal no (might even be just be converting a `continue` to a `return NewErrorFoo`)?  And the potential time/money savings for our users could be very large.

islamaliev (approved):
LGTM! Good work, Keenan.

tests/integration/query/simple/with_filter/with_alias_test.go
nasdf: If you see a solution that simple I'd definitely consider it. 

It would be really difficult to tell whether a filter property is targeting a relationship field or a nested JSON property without adding a good amount of new validation logic.

nasdf: I found a kind of simple solution for this. Let me know what you think.

internal/planner/mapper/targetable.go
```diff
@@ -81,16 +72,15 @@ type ObjectProperty struct {
...
-		return nil
+		return nil, defaultOp, nil
 	}
-	object := data.(map[string]any)
-	return object[k.Name]
-}
-
-func (k *ObjectProperty) GetOperatorOrDefault(defaultOp string) string {
-	return defaultOp
+	_, ok := data.(core.Doc)
```
AndrewSisley, line 79: suggestion:  Made me smile a little to see you do this again in this PR - and I'm curious as to why/where you think I prefer this :grin: (are you thinking back to me preferring nil checks vs failed casting?)

My suggestion is the same as the last, code is simpler IMO, and avoids having to care about an extra, somewhat irrelevant type:
```go
docMap, ok := data.(map[string]any)
if !ok {
  return nil, defaultOp, NewErrFieldOrAliasNotFound(k.Name)
}
return docMap[k.Name], defaultOp, nil
```

AndrewSisley (approved):
LGTM - thanks Keenan :)

Remaining suggestion from me is very optional.

internal/planner/mapper/targetable.go
nasdf: I think I burned that one into my brain. Should be fixed now.

islamaliev:
bug bash result: ran all tests from integration tests and many other related tests. Everything is good.

PR #3206, opened 2024-11-04
feat: Add ACP to pubsub KMS

### LINKED ISSUE #2893: Doc Encryption: Bind doc encryption to ACP
Integrate document encryption with ACP to ensure secure and controlled access to encrypted documents in our decentralized network.

This task depends on #2762

islamaliev:
## Relevant issue(s)

Resolves #2893

This change adds ACP to pubsub KMS.

A new event `MergeComplete` was introduced in order to make it correspond to `MergeCompleteName`. This is necessary to notify listeners that the merge was executed on decrypted blocks.

Upon granting access to a document via `AddDocActorRelationship` we publish not `Update` event so that the actors that being given access to can now request a document anew.

In testing framework `WaitForSync` is extended to allow specifying documents that should be received in decrypted state.


shahzadlone:
@islamaliev should I wait for you to fix the failing tests, or is it okay to review?

internal/kms/pubsub.go
```diff
@@ -327,6 +357,43 @@ func (s *pubSubService) getEncryptionKeysLocally(
...
+	ctx context.Context,
+	actorIdentity string,
+	entBlock *coreblock.Encryption,
+) (bool, error) {
+	if !s.acp.HasValue() {
+		return true, nil
+	}
+
+	docID := string(entBlock.DocID)
+	collection, err := s.colRetriever.RetrieveCollectionFromDocID(ctx, docID)
```
AndrewSisley, line 370: todo: This looks like we have modelled something incorrectly.

Medium-long term documents can belong to multiple collections, and the way this is coded means multiple policies can apply.  However, the document can (atm), only have a single set of encryption keys.

To me, the best option looks like we should decouple the doc encryption keys from the collection-level policies.  IMO this would provide the most powerful system.

Alternatively, we could remove policies from the collection level, however that seems extremely limiting in a bunch of ways that we've already gone over as a team (happy to expand).

Either way I think something needs to change here, and it might be quite expensive to do so.

AndrewSisley (changes requested):
I think we have modelled something incorrectly, and the discussion around that needs to be resolved before people spend time reviewing this IMO (ideally before you'd have spent your time coding this too, but such things happen, sorry).

internal/kms/pubsub.go
AndrewSisley: Discussed on discord, changing the model can wait, this PR cannot.  This comment can be ignored :)

AndrewSisley, line 370: todo: It looks like the rest of this function from this line down is just a duplication of `permission.CheckAccessOfDocOnCollectionWithACP` - can we use that function instead?

tests/integration/acp.go
```diff
@@ -356,7 +364,9 @@ func setupSourceHub(s *state) ([]node.ACPOpt, error) {
 		return nil, err
 	}
 
-	out, err := exec.Command("sourcehubd", "init", moniker, "--chain-id", chainID, "--home", directory).CombinedOutput()
+	args := []string{"init", moniker, "--chain-id", chainID, "--home", directory}
```
AndrewSisley, line 367: question: Just a quick check - you intend to commit these logs to develop?  They weren't accidentally included in the PR?

tests/integration/acp.go
```diff
@@ -566,23 +585,32 @@ func getFreePort() (int, func(), error) {
...
 func crossLock(port uint16) (func(), error) {
-	l, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%v", port))
-	if err != nil {
-		if strings.Contains(err.Error(), "address already in use") {
-			time.Sleep(5 * time.Millisecond)
-			return crossLock(port)
+	timeout := time.After(20 * time.Second)
+	for {
+		select {
+		case <-timeout:
```
AndrewSisley, line 595: praise: This is a nice addition, thanks :)

internal/kms/pubsub.go
```diff
@@ -36,13 +40,20 @@ type PubSubServer interface {
 	SendPubSubMessage(context.Context, string, []byte) (<-chan rpc.Response, error)
 }
 
+type CollectionRetriever interface {
```
AndrewSisley, line 43: question: Does this need to be interfaced?  There is only one type that implements it.

nitpick: The concrete type is a pointer to a pointer, I understand that is driven by Go norms, but it still feels a little unnecessary when I see it :)

AndrewSisley (changes requested):
Ignoring long term concerns regarding the model you are building against - it looks good, and was nice and easy/simple to read - just a couple of small todos from me.

Thanks Islam :)

internal/db/collection_retriever.go
```diff
@@ -0,0 +1,72 @@
...
+
+// CollectionRetriever is a helper struct that retrieves a collection from a document ID.
+type CollectionRetriever struct {
+	db *db
+}
+
+// NewCollectionRetriever creates a new CollectionRetriever.
+func NewCollectionRetriever(database client.DB) *CollectionRetriever {
+	internalDB, ok := database.(*db)
+	if !ok {
```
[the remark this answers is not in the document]
shahzadlone: guessing the linter will complain as all panics need to now be documented by putting under a linter suppressed `must` function, or just a linter ignore comment.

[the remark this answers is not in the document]
AndrewSisley: Can we remove that rule?  That sounds like a bad rule to have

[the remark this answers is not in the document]
AndrewSisley: When was it added?

internal/db/db.go
```diff
@@ -262,9 +262,41 @@ func (db *db) AddDocActorRelationship(
 		return client.AddDocActorRelationshipResult{}, err
 	}
 
+	err = db.publishDocUpdateEvent(ctx, docID, collection)
+	if err != nil {
+		return client.AddDocActorRelationshipResult{}, err
+	}
```
shahzadlone, line 269: todo: Current implementation as it stands will `publishDocUpdateEvent` even if the `AddDocActorRelation` call was a no-op, i.e. the relationship already existed and someone calls the command again and again.

IMO can likely check if `ExistedAlready` == `false` and publish the event only then, otherwise if the relationship existed already then it's a no-op, hence no need to publish the event.

internal/db/db.go
```diff
@@ -262,9 +262,41 @@ func (db *db) AddDocActorRelationship(
...
+			Cid:        headsIterator.CurrentCid(),
+			SchemaRoot: collection.Schema().Root,
+			Block:      headsIterator.CurrentRawBlock(),
+		}
+		db.events.Publish(event.NewMessage(event.UpdateName, updateEvent))
+	}
+	return nil
+}
+
 func (db *db) DeleteDocActorRelationship(
```
shahzadlone, line 300: todo: Don't we need to do the exact same thing for when we delete the relationship. i.e. `DeleteDocActorRelationship`

shahzadlone (commented):
Did a quick pass, mostly looks good. Wondering why we aren't doing the same thing for when the relation is deleted. And another optimization request to only publish when the relationship adding request is not a no-op.

internal/db/collection_retriever.go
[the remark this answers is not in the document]
shahzadlone: Was added here: https://github.com/sourcenetwork/defradb/pull/3134

Somewhat related discussion: https://github.com/sourcenetwork/defradb/pull/3134#discussion_r1802084140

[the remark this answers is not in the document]
shahzadlone: > Can we remove that rule? That sounds like a bad rule to have

I am happy with the consensus if people think it's not useful then so be it.

Just to understand your position you are saying implicit panics should be allowed?:

`internalDB := database.(*db)`

Which means we shouldn't enforce the users to either check the `ok` or suppress linter with a comment explaining why we are okay if it panics?:

`internalDB, ok := database.(*db)`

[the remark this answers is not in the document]
AndrewSisley: Ah okay - cheers - maybe we should discuss this rule some more - here it appears to have encouraged bad behaviour.  Although the code itself (constructor param loss of type safety) is fairly un-ideal anyway.

@islamaliev suggestion: I think we can actually remove the awkwardness of this func anyway, by just using `client.Db` instead of casting - the only func on `db.db` that is used seems to be `getCollectionFromRootSchema`, and that could be replaced by a call to the public `GetCollections` function.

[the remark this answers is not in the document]
AndrewSisley: > Just to understand your position you are saying implicit panics should be allowed?

I'm not sure on my position, but here the linter arguably damaged the codebase.  That damage was dependent on the somewhat misbehaving function signature though - so you could argue that the linter flagged the smell and it was incorrectly handled by the author.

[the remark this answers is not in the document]
shahzadlone: > > Just to understand your position you are saying implicit panics should be allowed?
> 
> I'm not sure on my position, but here the linter arguably damaged the codebase. That damage was dependent on the somewhat misbehaving function signature though - so you could argue that the linter correctly flagged the smell and it was incorrectly handled by the author.

I am in a similar boat regard my position. However I am just guessing that the linter is why islam did this, maybe he had other reasons.

internal/kms/pubsub.go
shahzadlone: +1

internal/kms/pubsub.go
```diff
@@ -260,9 +279,12 @@ func (s *pubSubService) tryGenEncryptionKeyLocally(
 	req *fetchEncryptionKeyRequest,
 ) (*fetchEncryptionKeyReply, error) {
 	blocks, err := s.getEncryptionKeysLocally(ctx, req)
-	if err != nil || len(blocks) == 0 {
+	if err != nil {
 		return nil, err
 	}
+	if len(blocks) == 0 {
+		return &fetchEncryptionKeyReply{}, nil
+	}
```
shahzadlone, line 287: suggestion: Maybe hit with a test if possible.

internal/kms/pubsub.go
islamaliev: done

islamaliev: `permission.CheckAccessOfDocOnCollectionWithACP` takes `immutable.Option[Identity]`. What I have in this functoin is just DID string. So I passed `immutable.Some(identity.Identity{DID: actorIdentity})` to the function. Not sure if it's 100% correct.

internal/kms/pubsub.go
islamaliev: This interface allows KMS to depend on only `client` type instead of `db`

the concrete type is `*db.CollectionRetriever`. I don't see 2 pointes there.

internal/db/db.go
```diff
@@ -262,9 +263,40 @@ func (db *db) AddDocActorRelationship(
...
 
+	err = db.publishDocUpdateEvent(ctx, docID, collection)
+	if err != nil {
+		return client.AddDocActorRelationshipResult{}, err
+	}
+
 	return client.AddDocActorRelationshipResult{ExistedAlready: exists}, nil
 }
 
+func (db *db) publishDocUpdateEvent(ctx context.Context, docID string, collection client.Collection) error {
```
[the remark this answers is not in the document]
islamaliev: donw

internal/db/db.go
islamaliev: good catch. Adjusted.

internal/db/db.go
islamaliev: the update event upon granting access is just to notify other nodes that they can try fetching encryption keys again.
I'm not sure why you would do it here. I can't even think of a test case.

tests/integration/acp.go
islamaliev: yeah, plan to keep it.
I didn't see much use of a command's output to console without the comman itself.

internal/db/db.go
fredcarle: > Maybe add a test, that after a document is shared its access is revoked, and then again it's access is reshared again. In this case keys already exist and it should still be accessible again.

I just want to point out that revoking access is only possible for future versions of a document. Not for versions that have already been shared.

internal/kms/pubsub.go
islamaliev: I added an additional tests. But strangely that line was already hit by another test

internal/db/db.go
```diff
@@ -30,7 +31,7 @@ import (
 	"github.com/sourcenetwork/defradb/acp/identity"
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/datastore"
-	"github.com/sourcenetwork/defradb/errors"
+	defraErrors "github.com/sourcenetwork/defradb/errors"
```
fredcarle, line 34: todo: Please add the `Join` function to the defra `errors` package like we did for `Is` and only import the defra `errors` package.

internal/db/db.go
shahzadlone: Yes indeee

internal/kms/pubsub.go
AndrewSisley: If it was unexpectedly hit it would be worth finding out why.

internal/db/collection_retriever.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/internal/db/description"
+)
+
+// CollectionRetriever is a helper struct that retrieves a collection from a document ID.
+type CollectionRetriever struct {
```
fredcarle, line 23: suggestion: This type doesn't need to be public. It's the internal db implementation of `kms.CollectionRetriever` interface. It's not meant to be used directly.

internal/db/iterator.go
```diff
@@ -0,0 +1,112 @@
...
+	ctx        context.Context
+	blockstore datastore.Blockstore
+	cids       []cid.Cid
+
+	currentCid      cid.Cid
+	currentBlock    *coreblock.Block
+	currentRawBlock []byte
+}
+
+var _ io.Closer = (*DocHeadBlocksIterator)(nil)
```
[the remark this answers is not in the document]
islamaliev: Just to be consistent with other iterators. No real reason. I can remove it.

internal/db/collection_retriever.go
islamaliev: done

internal/kms/pubsub.go
islamaliev: it is expected: upon doc creation the peer that receive an update (not creator) requests encryption key as it can't merge blocks. But the creator peer rejects (by responding with empty response), because the access was not granted yet.

AndrewSisley: Thanks for the explanation Islam :)

internal/kms/pubsub.go
AndrewSisley: The only thing `*db.CollectionRetriever` contains is a pointer to `db` - thus all instances of `kms.CollectionRetriever` are pointers to a pointer.

internal/db/db.go
```diff
@@ -227,6 +227,32 @@ func (db *db) AddPolicy(
 	return client.AddPolicyResult{PolicyID: policyID}, nil
 }
 
+func (db *db) publishDocUpdateEvent(ctx context.Context, docID string, collection client.Collection) error {
```
AndrewSisley, line 230: todo: This function is scoped to a large and important type, does very similar things to a couple of existing places, and ignores the context txn.  It looks very easy to erroneously call this, and in an area where testing is relatively weak (transactions).

Please document this function to highlight that it ignores the transaction, and does a whole load of work to find the heads (something the two existing locations do not need to do).

Please do not refactor the existing locations as I have changed them in another open PR, with this 3rd addition we can look at refactoring the 3 after this (and maybe the other PR, depending on timing) has merged.

AndrewSisley (approved):
Looks good to me, the code is very easy to read - thanks Islam :)

Just one documentation request from me (I think) before I'm happy with merging :)

shahzadlone (approved):
LGTM Thanks very much for this change and explaining why that line was being hit.

Just one suggestion remaining incase you missed.

suggestion: Add a test where after sharing a document, it is revoked and then shared again. i.e. using `DeleteDocActorRelationship`

islamaliev:
> suggestion: Add a test where after sharing a document, it is revoked and then shared again. i.e. using `DeleteDocActorRelationship`

I don't think it makes much sense, as this would mostly test acp functionality which we already have plenty tests for. Revoking of encryption keys is planned for later anyway. This is where we will add a bunch of tests.

PR #3210, opened 2024-11-05
refactor: Rework core.Spans

### LINKED ISSUE #3209: Rework core.Spans
Broken out of #3038

AndrewSisley:
## Relevant issue(s)

Resolves #3209

## Description

Simplifies core.Spans a little bit, before removing some old hacky code allowing a nicer way of handling headstore keys within the planner/fetcher system.

Is probably easier to review this commit by commit, the changes are documented by the commit messages.

### DIFF
--- a/internal/db/fetcher/fetcher.go
+++ b/internal/db/fetcher/fetcher.go
@@ -278,11 +278,20 @@ func (df *DocumentFetcher) start(ctx context.Context, spans []core.Span, withDel
 	} else {
 		valueSpans := make([]core.Span, len(spans))
 		for i, span := range spans {
-			// We can only handle value keys, so here we ensure we only read value keys
 			if withDeleted {
-				valueSpans[i] = core.NewSpan(span.Start.WithDeletedFlag(), span.End.WithDeletedFlag())
+				// DocumentFetcher only ever recieves document keys
+				//nolint:forcetypeassert
+				valueSpans[i] = core.NewSpan(
+					span.Start.(keys.DataStoreKey).WithDeletedFlag(),
+					span.End.(keys.DataStoreKey).WithDeletedFlag(),
+				)
 			} else {
-				valueSpans[i] = core.NewSpan(span.Start.WithValueFlag(), span.End.WithValueFlag())
+				// DocumentFetcher only ever recieves document keys
+				//nolint:forcetypeassert
+				valueSpans[i] = core.NewSpan(
+					span.Start.(keys.DataStoreKey).WithValueFlag(),
+					span.End.(keys.DataStoreKey).WithValueFlag(),
+				)
 			}
 		}
 
--- a/internal/db/fetcher/fetcher.go
+++ b/internal/db/fetcher/fetcher.go
@@ -267,7 +267,7 @@ func (df *DocumentFetcher) start(ctx context.Context, spans core.Spans, withDele
 
 	df.deletedDocs = withDeleted
 
-	if !spans.HasValue { // no specified spans so create a prefix scan key for the entire collection
+	if len(spans.Value) == 0 { // no specified spans so create a prefix scan key for the entire collection
 		start := base.MakeDataStoreKeyWithCollectionDescription(df.col.Description())
 		if withDeleted {
 			start = start.WithDeletedFlag()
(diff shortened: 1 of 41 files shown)


tests/integration/explain/default/delete_test.go
```diff
@@ -292,7 +292,12 @@ func TestDefaultExplainMutationRequestWithDeleteUsingNoIds(t *testing.T) {
 							"collectionID":   "3",
 							"collectionName": "Author",
 							"filter":         nil,
-							"spans":          []dataMap{},
+							"spans": []dataMap{
+								{
+									"end":   "/4",
+									"start": "/3",
+								},
+							},
```
shahzadlone, line 300: question: So this was the bug that was caused due to the way `HasValue` was being used?

internal/db/fetcher/fetcher.go
```diff
@@ -278,11 +278,20 @@ func (df *DocumentFetcher) start(ctx context.Context, spans []core.Span, withDel
...
+				// DocumentFetcher only ever recieves document keys
+				//nolint:forcetypeassert
+				valueSpans[i] = core.NewSpan(
+					span.Start.(keys.DataStoreKey).WithDeletedFlag(),
+					span.End.(keys.DataStoreKey).WithDeletedFlag(),
+				)
 			} else {
-				valueSpans[i] = core.NewSpan(span.Start.WithValueFlag(), span.End.WithValueFlag())
+				// DocumentFetcher only ever recieves document keys
+				//nolint:forcetypeassert
```
shahzadlone, line 290: suggestion: Have a must DataStoreKey function for the casting? and do this lint ignore there once for all these 4 casts?

internal/keys/datastore_doc.go
```diff
@@ -41,7 +41,7 @@ type DataStoreKey struct {
 	FieldID          string
 }
 
-var _ Key = (*DataStoreKey)(nil)
+var _ Walkable = (*DataStoreKey)(nil)
```
shahzadlone, line 44: question: What inspired the name change, and the context for the name "walkable"? I am guessing it's the key that can be accessed.

EDIT: Was answered by documentation on `Walkable` ignore this question.

internal/keys/key.go
```diff
@@ -20,3 +20,20 @@ type Key interface {
 	Bytes() []byte
 	ToDS() ds.Key
 }
+
+// Walkable represents a key in the database that can be 'walked along'
+// by prefixing the end of the key.
+type Walkable interface {
+	Key
+	PrefixEnd() Walkable
+}
```
shahzadlone, line 29: praise: Answered my question in another comment, ignore the previous question please.

internal/planner/commit.go
```diff
@@ -142,13 +144,14 @@ func (n *dagScanNode) simpleExplain() (map[string]any, error) {
 
 	// Build the explanation of the spans attribute.
 	spansExplainer := []map[string]any{}
+	undefinedHsKey := keys.HeadStoreKey{}
```
shahzadlone, line 147: nitpick:
```suggestion
	undefinedHSKey := keys.HeadStoreKey{}
```

shahzadlone (approved):
Praise: I like the change, specially the de-abstraction of the `interface` into struct and the removal of `HasValue`

internal/db/fetcher/fetcher.go
AndrewSisley: I do not think moving a simple cast into a function would improve the readability of the code here.

internal/planner/commit.go
AndrewSisley: Headstore is one word, the current capitalization of `HeadStoreKey` is incorrect and has been fixed in the branchable collections branch (is bundled with some other branchable-specific changes, which is why it isnt in this PR).

internal/db/fetcher/fetcher.go
shahzadlone: Can also move `span.Start.(keys.DataStoreKey)` and `span.End.(keys.DataStoreKey)` out of if-else blocks into a variable, and then use that in their respective occurrences. Feel free to ignore as this seems like subjective pref

tests/integration/explain/default/delete_test.go
AndrewSisley: Yeah, there was something fishy going on here with HasValue.  The newly expected result is correct, and the old property/type has died so I didn't bother chasing down exactly where the bug was

Issue #3214, opened 2024-11-06
Querying docs with just a `cid` (no docID) panics


AndrewSisley:
Documented by `TestQuerySimpleWithCid`

E.g.:
```
Request: `query {
	Users (
			cid: "bafyreib7afkd5hepl45wdtwwpai433bhnbd3ps5m2rv3masctda7b6mmxe"
		) {
		name
	}
}`,
```
When fixing, also test what happens when a collection-level cid is provided, as that is currently untested.

> panic: runtime error: index out of range [0] with length 0 [recovered]
>	panic: runtime error: index out of range [0] with length 0
>
> goroutine 15 [running]:
testing.tRunner.func1.2({0x493f480, 0xc002480150})
	/home/andy/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.7.linux-amd64/src/testing/testing.go:1631 +0x24a
testing.tRunner.func1()
	/home/andy/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.7.linux-amd64/src/testing/testing.go:1634 +0x377
panic({0x493f480?, 0xc002480150?})
	/home/andy/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.7.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/sourcenetwork/defradb/internal/planner.(*selectNode).initSource(0xc0008b5ad0)
	/home/andy/Projects/Source/defradb/internal/planner/select.go:270 +0x496
github.com/sourcenetwork/defradb/internal/planner.(*Planner).Select(0xc001736900, 0xc001f904d0)
	/home/andy/Projects/Source/defradb/internal/planner/select.go:508 +0x105
github.com/sourcenetwork/defradb/internal/planner.(*Planner).Operation(0xc001736900, 0xc000f37360)
	/home/andy/Projects/Source/defradb/internal/planner/operation.go:139 +0x145
github.com/sourcenetwork/defradb/internal/planner.(*Planner).MakePlan(0xc001736900, 0x0?)
	/home/andy/Projects/Source/defradb/internal/planner/planner.go:603 +0x65
github.com/sourcenetwork/defradb/internal/planner.(*Planner).RunRequest(0xc001736900, {0x582fbb0, 0xc001ab1020}, 0xc000f37310)
	/home/andy/Projects/Source/defradb/internal/planner/planner.go:520 +0x65
[...]

PR #3216, opened 2024-11-07
feat: Add support for branchable collections

### LINKED ISSUE #3038: Add support for branchable collections
There is some high level info in an almanac doc here: https://source.almanac.io/folders/database-team-v1-q4qgzK/branchable-collections-Z0SmQmyPcgaYimjMlHuLZPEGM1s38Ewg?mention_id=4192&docView=Commenting 

WIP thoughts:
- We do not need a composite for each document change, instead we only need one for each transaction commit.  This would reduce the cost (performance, size on disk) of a branchable collection considerably in a lot of use cases.  Some event based code may need to change for this.  The composite commit would contain the cids of all the changes to the collection within the transaction, instead of essentially one cid per composite.
  - It might be best to submit a txn commited event, and include txn.id on update events instead of aggregate update events into a set before publishing them.
  - Make sure the branchable collection is modified within the commit, not after it is commited - may need before/after on-success operations?.
  - collection-commit could be queried mid-transaction, so if doing that now, handle it (needs to mutate with every doc-update).
  - ~~Remember that the order of operations within a transaction is important, so make sure that the contents of the composite commit handles that.  And remember that a document can be mutated multiple times within a single transaction (although the head cids will always differ).~~ doc-commit height handles this.
- We can allow users to query branchable collections at a specific version like we do with documents (the `cid` query param).  This does not however need to be implemented in the initial PR, is just something to bear in mind for now.
- The existing CRDT related code is still overly complex and abstracted.  The important file is `internal/block/block.go` (the `Block` struct is also used in P2P), strongly consider bypassing all the other noise, you may get complaints at review time. ~~Do not use `DAGLink`, use the ipld `Link` directly in the payload.~~
  - You can store the head cid in the headstore, but use a sensible Key, do not use the HeadStoreKey.
  - ~~Decoupling `Block` from `DAGLink` might be a pain due to the existing code structure.~~  
  - ~~`DAGLink`.`Name` is only ever used for the `"_head"` magic value, and commit queries. `"_head"` links are the commit cids that this commit is based off of, there can be multiple if multiple blocks share the same height. Non-head values are duplicated and waste space as they are already stored within the other commit.~~
  - `DAGLink` removal was rejected by team.
- Commit queries:
  - There is already a `collectionID` prop, use it.  This prop can be changed to use the collection cid.  
  - Do not store the collection cid in the block for field/doc level commits as it is local only and will take up space, you can rely on the one-one schema-collection relationship internally whilst presenting the cid at a field/doc commit level. 
  - We can add a collection name prop later (#850)
  - Check existing gql documentation for props like `fieldName` and adjust them if needed
- Testing this will increase the number of test cids, strongly consider removing the need to hardcode them first (`CidIndex`?)
- ACP question - long term these need to be permissioned, hopefully can skip this for now (a document create is an update to the collection, so this could probably be handled by collection-level write permissions?).
  - Make sure adding a policy to a branchable collection errors, and make sure making a collection that has a policy  branchable errors too.
- The `cid` param in normal queries can return an error for now if given a collectionCID (document on gql description), long term this needs to work.
- I can't think of any schema-migration problems that could come up with the current implementations, however there has been talk of making Lens migrations DAG-aware, which could introduce new problems.  It might be worth adding a basic test or two in case they manage to catch anything silly in the future.
- RE block design (size *is* important here)
  - `optional` removes overhead for block storage size on empty fields but is limited to certain types (not string)
  - Property name affects overhead! Longer names = bigger storage costs

```[tasklist]
### Tasks
- [ ] https://github.com/sourcenetwork/defradb/issues/3056
- [ ] https://github.com/sourcenetwork/defradb/issues/3085
- [ ] https://github.com/sourcenetwork/defradb/issues/3172
- [ ] https://github.com/sourcenetwork/defradb/issues/3191
- [ ] https://github.com/sourcenetwork/defradb/issues/3209
- [ ] https://github.com/sourcenetwork/defradb/issues/3212
- [ ] https://github.com/sourcenetwork/defradb/issues/3213
```

AndrewSisley:
## Relevant issue(s)

Resolves #3038

## Description

Adds support for branchable collections.

Does not add support for syncing the collection level commits via the P2P system (broken out, lots of people working in the space, significant changes required, I'm not so familiar with that part of the code so will take longer).  This does mean that the somewhat surprising (to me at least) implementation of `Collection.Merge` is currently untested.

Commits are queriable via the `commits` GQL queries.

Time travel queries do not work due to an existing bug: https://github.com/sourcenetwork/defradb/issues/3214  - once that bug is fixed I expect there to be some more work (definitely testing) in order to get it to work with branchable collections.

This is a breaking change due to the moving/namespacing of the existing document headstore keys.

### DIFF
--- a/client/collection_description.go
+++ b/client/collection_description.go
@@ -89,7 +89,7 @@ type CollectionDescription struct {
 	// another collection/query (is a View).
 	IsMaterialized bool
 
-	// IsMaterialized defines whether the history of this collection is tracked as a single,
+	// IsBranchable defines whether the history of this collection is tracked as a single,
 	// verifiable entity.
 	//
 	// If set to `true` any change to the contents of this set will be linked to a collection
--- a/client/collection_description.go
+++ b/client/collection_description.go
@@ -88,6 +88,21 @@ type CollectionDescription struct {
 	// At the moment this can only be set to `false` if this collection sources its data from
 	// another collection/query (is a View).
 	IsMaterialized bool
+
+	// IsMaterialized defines whether the history of this collection is tracked as a single,
+	// verifiable entity.
+	//
+	// If set to `true` any change to the contents of this set will be linked to a collection
+	// level commit via the document(s) composite commit.
+	//
+	// This enables multiple nodes to verify that they have the same state/history.
+	//
+	// The history may be queried like a document history can be queried, for example via 'commits'
+	// GQL queries.
+	//
+	// Currently this property is immutable and can only be set on collection creation, however
+	// that will change in the future.
+	IsBranchable bool
 }
 
 // QuerySource represents a collection data source from a query.
@@ -189,6 +204,7 @@ type collectionDescription struct {
 	RootID          uint32
 	SchemaVersionID string
 	IsMaterialized  bool
+	IsBranchable    bool
 	Policy          immutable.Option[PolicyDescription]
 	Indexes         []IndexDescription
 	Fields          []CollectionFieldDescription
@@ -209,6 +225,7 @@ func (c *CollectionDescription) UnmarshalJSON(bytes []byte) error {
 	c.RootID = descMap.RootID
 	c.SchemaVersionID = descMap.SchemaVersionID
 	c.IsMaterialized = descMap.IsMaterialized
+	c.IsBranchable = descMap.IsBranchable
 	c.Indexes = descMap.Indexes
 	c.Fields = descMap.Fields
 	c.Sources = make([]any, len(descMap.Sources))
--- a/internal/core/crdt/collection.go
+++ b/internal/core/crdt/collection.go
@@ -13,7 +13,6 @@ package crdt
 import (
 	"context"
 
-	"github.com/sourcenetwork/defradb/datastore"
 	"github.com/sourcenetwork/defradb/internal/core"
 	"github.com/sourcenetwork/defradb/internal/keys"
 )
@@ -22,8 +21,6 @@ import (
 // collection in a similar way to a document composite commit, only simpler,
 // without the need to track status and a simpler [Merge] function.
 type Collection struct {
-	store datastore.DSReaderWriter
-
 	// schemaVersionKey is the schema version datastore key at the time of commit.
 	//
 	// It can be used to identify the collection datastructure state at the time of commit.
@@ -32,9 +29,8 @@ type Collection struct {
 
 var _ core.ReplicatedData = (*Collection)(nil)
 
-func NewCollection(store datastore.DSReaderWriter, schemaVersionKey keys.CollectionSchemaVersionKey) *Collection {
+func NewCollection(schemaVersionKey keys.CollectionSchemaVersionKey) *Collection {
 	return &Collection{
-		store:            store,
 		schemaVersionKey: schemaVersionKey,
 	}
 }
--- a/internal/core/crdt/collection.go
+++ b/internal/core/crdt/collection.go
@@ -1,0 +1,78 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package crdt
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/defradb/datastore"
+	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/keys"
+)
+
+// Collection is a simple CRDT type that tracks changes to the contents of a
+// collection in a similar way to a document composite commit.
+type Collection struct {
+	store datastore.DSReaderWriter
+
+	// schemaVersionKey is the schema version datastore key at the time of commit.
+	//
+	// It can be used to identify the collection datastructure state at the time of commit.
+	schemaVersionKey keys.CollectionSchemaVersionKey
+}
+
+var _ core.ReplicatedData = (*Collection)(nil)
+
+func NewCollection(store datastore.DSReaderWriter, schemaVersionKey keys.CollectionSchemaVersionKey) *Collection {
+	return &Collection{
+		store:            store,
+		schemaVersionKey: schemaVersionKey,
+	}
+}
+
+func (c *Collection) Merge(ctx context.Context, other core.Delta) error {
+	// Collection merges don't actually need to do anything, as the delta is empty,
+	// and doc-level merges are handled by the document commits.
+	return nil
+}
+
+func (c *Collection) Append() *CollectionDelta {
+	return &CollectionDelta{
+		SchemaVersionID: c.schemaVersionKey.SchemaVersionID,
+	}
+}
+
+type CollectionDelta struct {
+	Priority uint64
+
+	// As we do not yet have a global collection id we temporarily rely on the schema
+	// version id for tracking which collection this belongs to.  See:
+	// https://github.com/sourcenetwork/defradb/issues/3215
+	SchemaVersionID string
+}
+
+var _ core.Delta = (*CollectionDelta)(nil)
+
+func (delta *CollectionDelta) IPLDSchemaBytes() []byte {
+	return []byte(`
+	type CollectionDelta struct {
+		priority  		Int
+		schemaVersionID String
+	}`)
+}
+
+func (d *CollectionDelta) GetPriority() uint64 {
+	return d.Priority
+}
+
+func (d *CollectionDelta) SetPriority(priority uint64) {
+	d.Priority = priority
+}
--- a/internal/core/crdt/collection.go
+++ b/internal/core/crdt/collection.go
@@ -19,7 +19,8 @@ import (
 )
 
 // Collection is a simple CRDT type that tracks changes to the contents of a
-// collection in a similar way to a document composite commit.
+// collection in a similar way to a document composite commit, only simpler,
+// without the need to track status and a simpler [Merge] function.
 type Collection struct {
 	store datastore.DSReaderWriter
 
--- a/internal/core/crdt/collection.go
+++ b/internal/core/crdt/collection.go
@@ -41,7 +41,7 @@ func (c *Collection) Merge(ctx context.Context, other core.Delta) error {
 	return nil
 }
 
-func (c *Collection) Append() *CollectionDelta {
+func (c *Collection) NewDelta() *CollectionDelta {
 	return &CollectionDelta{
 		SchemaVersionID: c.schemaVersionKey.SchemaVersionID,
 	}
--- a/internal/db/collection.go
+++ b/internal/db/collection.go
@@ -711,6 +711,29 @@ func (c *collection) save(
 		doc.SetHead(link.Cid)
 	})
 
+	if c.def.Description.IsBranchable {
+		collectionCRDT := merklecrdt.NewMerkleCollection(
+			txn,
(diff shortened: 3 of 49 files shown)


client/collection_description.go
```diff
@@ -88,6 +88,21 @@ type CollectionDescription struct {
 	// At the moment this can only be set to `false` if this collection sources its data from
 	// another collection/query (is a View).
 	IsMaterialized bool
+
+	// IsMaterialized defines whether the history of this collection is tracked as a single,
```
shahzadlone, line 92: typo: says IsMaterialized

AndrewSisley: "typo" :grin: Cheers, fixing now

- [ ] fix typo

islamaliev (approved):
LGTM! Good work, Andy.

internal/core/crdt/collection.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+
+	"github.com/sourcenetwork/defradb/datastore"
+	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/keys"
+)
+
+// Collection is a simple CRDT type that tracks changes to the contents of a
+// collection in a similar way to a document composite commit.
```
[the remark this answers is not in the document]
AndrewSisley: Okay - cheers, I'll see what I can do :)

- [ ] Expand/clarify doc

internal/db/fetcher/dag.go
```diff
@@ -31,13 +31,18 @@ type HeadFetcher struct {
 func (hf *HeadFetcher) Start(
 	ctx context.Context,
 	txn datastore.Txn,
-	prefix keys.HeadStoreKey,
+	prefix immutable.Option[keys.HeadstoreKey],
```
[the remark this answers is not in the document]
AndrewSisley: None of the implementations of `HeadstoreKey` would result in an empty prefix that allows scanning across the entire headstore (required in order to return collection and doc commits).

We could pass `nil` and check for that, or we could implement a Nil/All HeadstoreKey type (very weird), but both of those seem a lot less pleasant than using option.

It does sound like why this param is optional should be documented on `Start` though, so I'll add something there.

- [ ] Doc prefix param

internal/db/collection.go
```diff
@@ -711,6 +711,29 @@ func (c *collection) save(
...
+
+		updateEvent := event.Update{
+			Cid:        link.Cid,
+			SchemaRoot: c.Schema().Root,
+			Block:      headNode,
+		}
+
+		txn.OnSuccess(func() {
+			c.db.events.Publish(event.NewMessage(event.UpdateName, updateEvent))
+		})
```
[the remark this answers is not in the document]
AndrewSisley: > Can't the receiver of the even simpli check on this side if the collection is branchable and handle correspondingly?

I hadn't considered that, I'll think about that and get back to you :)

> if we need to send another event, then I would say it serves a different purpose and in this case we should introduce another event type.

I disagree with this, it is an update event, an update of a collection.  It needs to be handled for exactly the same reasons that an update (doc) event is currently handled, and in pretty much the same way (some rework required as P2P is very focused on docIDs etc atm as noted in PR description).

tests/integration/query/commits/branchables/cid_test.go
```diff
@@ -0,0 +1,63 @@
...
+							cid
+							collectionID
+							docID
+							fieldName
+						}
+					}`,
+				Results: map[string]any{
+					"commits": []map[string]any{
+						{
+							"cid": testUtils.NewUniqueCid("collection"),
```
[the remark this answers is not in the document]
AndrewSisley: Yeah, I understand that - I think when I next use `NewUniqueCid` I should force myself to use another type (probs int) to see how that feels.

internal/db/collection.go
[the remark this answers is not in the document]
AndrewSisley: > I hadn't considered that, I'll think about that and get back to you

I think I prefer the extra events.  It feels more correct to send independent updates for the doc and the collection, especially so as the event bus is public and can be subscribed to by users.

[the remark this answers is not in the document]
fredcarle: I also prefer the extra event but the reason for why I prefer it is missing from this implementation.

I would expect a `CreateMany` to generate only one collection block linking to all the docs created (same for updates or mix of creates and updates). This means that we would have one update event for every doc and one update event at the end for the collection.

internal/db/fetcher/dag.go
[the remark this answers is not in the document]
fredcarle: >  If None, the entire headstore will be scanned - for example, in order to fetch document and collection heads.

Is that something we even want to allow?

internal/request/graphql/schema/types/types.go
```diff
@@ -237,6 +240,31 @@ func MaterializedDirective() *gql.Directive {
...
+ as a single, verifiable entity or not. It will default to false if ommited.
+
+ If multiple @branchable directives are provided, they will aggregated with OR logic (if any are true, the
+ collection history will be tracked).
+
+ The history may be queried like a document history can be queried, for example via 'commits'
+ GQL queries.
+
+ Currently this property is immutable and can only be set on collection creation, however
+ that will change in the future.`,
```
fredcarle, line 256: suggestion: We should probably add the todo comment with the linked issue above so that we don't forget to edit this once we add the feature.

internal/core/crdt/collection.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/sourcenetwork/defradb/datastore"
+	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/keys"
+)
+
+// Collection is a simple CRDT type that tracks changes to the contents of a
+// collection in a similar way to a document composite commit, only simpler,
+// without the need to track status and a simpler [Merge] function.
+type Collection struct {
+	store datastore.DSReaderWriter
```
fredcarle, line 25: todo: This field can be removed as it's never used.

internal/core/crdt/collection.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// schemaVersionKey is the schema version datastore key at the time of commit.
+	//
+	// It can be used to identify the collection datastructure state at the time of commit.
+	schemaVersionKey keys.CollectionSchemaVersionKey
```
fredcarle, line 30: suggestion: I don't think there is any reason to have to create this schema version key only to extract the `schemaVersionID` later on. I think going directly with the `schemaVersionID` would be preferable.

internal/core/crdt/collection.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+func (c *Collection) Merge(ctx context.Context, other core.Delta) error {
+	// Collection merges don't actually need to do anything, as the delta is empty,
+	// and doc-level merges are handled by the document commits.
+	return nil
+}
+
+func (c *Collection) Append() *CollectionDelta {
```
fredcarle, line 48: suggestion: `Set` might be a better name for this. There is no reason for it to be different than the composite CRDT.

fredcarle (changes requested):
Overall this is well done. I like the tests that document the unimplemented features as well.

Just a few comments to resolve before approval.

internal/db/collection.go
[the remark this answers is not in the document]
AndrewSisley: > I would expect a CreateMany to generate only one collection block linking to all the docs created (same for updates or mix of creates and updates). This means that we would have one update event for every doc and one update event at the end for the collection.

I was planning on building that here, but then I realised it is not specific to collection-level commits.   The same system can (should) be applied to all crdt/blocks.

internal/db/fetcher/dag.go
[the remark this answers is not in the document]
AndrewSisley: Yes, most of the tests in this PR rely on that, and it is the default behaviour of `commits` queries.

internal/request/graphql/schema/types/types.go
AndrewSisley: Not a problem, although that issue does not exist yet, so I'll be adding another item to sit in our long backlog.

- [ ] Create and link issue for toggling branchableness

internal/core/crdt/collection.go
AndrewSisley: Will have a look, not a lot of thought went into this, I copy paasted from another crdt and dropped the unused props.

- [ ] rm col.schemaversionKey?

internal/core/crdt/collection.go
AndrewSisley: I renamed composite.Set to `Append`

internal/db/collection.go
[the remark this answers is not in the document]
fredcarle: > The same system can (should) be applied to all crdt/blocks.

Can you expand on this? Why would that apply to composites or field level CRDTs?

internal/request/graphql/schema/types/types.go
fredcarle: I thought it did already exist 🤷‍♂️

internal/core/crdt/collection.go
fredcarle: Append is equally a bad name for something that doesn't append anything though. Set does set fields in the delta object.

internal/db/collection.go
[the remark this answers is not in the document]
AndrewSisley: Just as multiple collection blocks can be created in a single transaction, so too can multiple composite and field level blocks.  Grouping any of these together into a single block should work consistently regardless of type.

E.g. if in a single transaction `Age` on `Document1` in `branchable collection 1` is updated twice, the same mechanic for rolling the would-be-two collection level commits into a single collection commit should also apply to the field and composite level commits.  

There is nothing special about collection-level commits here, although the need for it is possibly higher.

internal/core/crdt/collection.go
AndrewSisley: A new delta object, is sets nothing on the host.  Although `Append` doesnt append to the host really either.  Both names are bad, but they are now consistent at least.

The two functions are constructors, nothing more.  They could be renamed `NewDelta` perhaps.

Note: I don't see this as something terribly important, and am not very keen on thinking much about it.

internal/db/collection.go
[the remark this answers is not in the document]
fredcarle: I would say this is definitely not the case for field level CRDTs. They only point to their previous head (they don't have links to other items) plus they don't cause network updates on their own. It *may* apply to composites for the multiple updates to the same document in one transaction but I'm not sure there is much benefit to it as it would create multiple heads for the updated fields in question only to save a composite block.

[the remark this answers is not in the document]
AndrewSisley: > I would say this is definitely not the case for field level CRDTs.

We disagree, and it makes me very glad I did not try and do that work within this PR :)

internal/core/crdt/collection.go
fredcarle: Can you leave it as `Set` for now. It doesn't make sense to me to chose the objectively less accurate name `Append` when `Set` was there previously. I have a vague memory of John deciding to use `Set` because it would have been the correct naming in CRDT lingo. If that's not the case and you really don't want to use `Set`, `NewDelta` is definitely the better choice.

AndrewSisley: done :)

fredcarle (approved):
LGTM. I think the CRDT code is starting to be much more approchable 👍

fredcarle:
bug bash: Manually tested running 2 nodes, creating docs on a branchable collection, running standard and time travelling queries on collection commits and inspected commit queries. All works as expected.

PR #3220, opened 2024-11-08
refactor: Identity type in tests

fredcarle:
## Description

I had this branch locally from a suggestion that I had made on the testing identity change that Islam made. I'm bringing it up just to see if we prefer this slight modification of the merged change.

### DIFF
--- a/tests/integration/acp.go
+++ b/tests/integration/acp.go
@@ -88,7 +88,7 @@ type AddPolicy struct {
 	Policy string
 
 	// The policy creator identity, i.e. actor creating the policy.
-	Identity immutable.Option[identityRef]
+	Identity identity
 
 	// The expected policyID generated based on the Policy loaded in to the ACP system.
 	ExpectedPolicyID string
@@ -159,13 +159,13 @@ type AddDocActorRelationship struct {
 	// The target public identity, i.e. the identity of the actor to tie the document's relation with.
 	//
 	// This is a required field. To test the invalid usage of not having this arg, use NoIdentity() or leave default.
-	TargetIdentity immutable.Option[identityRef]
+	TargetIdentity identity
 
 	// The requestor identity, i.e. identity of the actor creating the relationship.
 	// Note: This identity must either own or have managing access defined in the policy.
 	//
 	// This is a required field. To test the invalid usage of not having this arg, use NoIdentity() or leave default.
-	RequestorIdentity immutable.Option[identityRef]
+	RequestorIdentity identity
 
 	// Result returns true if it was a no-op due to existing before, and false if a new relationship was made.
 	ExpectedExistence bool
@@ -195,7 +195,7 @@ func addDocActorRelationshipACP(
 			collectionName,
 			docID,
 			action.Relation,
-			getIdentityDID(s, action.TargetIdentity),
+			action.TargetIdentity.get(s).DID,
 		)
 
 		expectedErrorRaised := AssertError(s.t, s.testCase.Description, err, action.ExpectedError)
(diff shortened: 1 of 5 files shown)


AndrewSisley (commented):
question: What do you think is better here Fred?

My quick thoughts given that I have not used the existing anyway yet are below. 

1. IMO regarding what I see as the most important factor (test readability), the use of `Option` is nicer than the enum you propose regarding some/none values. 
2. Both old and new are odd in that they are using private types for public members on public types (`identityRef` and `identity`)
3. The internal code of the proposal looks basically the same as the original to me - the new version's it's handling of none is less nice (enum), but it has pulled some stuff out into new functions (kind of independent from the type change by the looks of things) (perhaps prettier at review time, but adds an extra layer to debug/read/jump-around-to when actually maintaining, as well as the extra clutter to an already busy package scope).

fredcarle:
> question: What do you think is better here Fred?

The only thing I like better is using an enum for the identity type as it's more flexible and descriptive than just a boolean in the struct. 

I remember people in the standup saying that they liked that version and since I still had it available, I though I'd ask if it was still the case.

AndrewSisley:
> The only thing I like better is using an enum for the identity type as it's more flexible and descriptive than just a boolean in the struct.

Ah fair enough :) IMO currently there should only be two values in that enum (none being covered by `Option`), and I have no real preference RE a bool vs 2-member-enum (maybe here the enum is slightly nicer given the nature of the values).

tests/integration/identity.go
```diff
@@ -123,25 +128,27 @@ func generateIdentity(s *state) acpIdentity.Identity {
...
 // If the identity does not exist, it will be generated.
 // The identity added to the context is prepared for a request, i.e. its [Identity.BearerToken] is set.
-func getContextWithIdentity(
-	ctx context.Context,
-	s *state,
-	ref immutable.Option[identityRef],
-	nodeIndex int,
-) context.Context {
-	if !ref.HasValue() {
+func getContextWithIdentity(ctx context.Context, s *state, ref identity, nodeIndex int) context.Context {
```
shahzadlone, line 148: todo: Don't need to call the arg `ref` anymore specially with the `identityRef` rename.

tests/integration/identity.go
```diff
@@ -21,31 +21,59 @@ import (
...
+	switch i.iType {
+	case clientType:
+		identities = s.clientIdentities
+	case nodeType:
+		identities = s.nodeIdentities
+	default:
+		return acpIdentity.Identity{}
+	}
+	return getIdentityHolder(s, i.index, identities).Identity
+}
```
shahzadlone, line 64: question: Did you consider an identity interface that has a `Get` method perhaps? Maybe over abstraction perhaps but would remove the need for knowing the `iType` and the respective implementation would return/get their specific field.

shahzadlone (commented):
I like the rename to `identity` from `identityRef`.

I do prefer a 2 enum with an empty `Option` for the identity of `noType`.

I do also like the splitting of identities into client and node identities

tests/integration/acp.go
```diff
@@ -88,7 +88,7 @@ type AddPolicy struct {
 	Policy string
 
 	// The policy creator identity, i.e. actor creating the policy.
-	Identity immutable.Option[identityRef]
```
islamaliev, line 91: though: this is a bit strange after being persuaded during the code review or standup discussion into making it `Option` seeing it gone again here. I'm curious to know how it is different now @shahzadlone

tests/integration/identity.go
```diff
@@ -21,31 +21,59 @@ import (
...
-// identityRef is a type that refers to a specific identity of a certain type.
-type identityRef struct {
-	isClient bool
-	index    int
+type identityType int
+
+const (
+	noType identityType = iota
+	clientType
+	nodeType
```
islamaliev, line 29: todo: please add some scope prefix, because `integration` package is somewhat big and from other context these consts might be confusing.

tests/integration/identity.go
islamaliev: I believe adding an interface is overkill for this purpose. And the type will be known anyway, it will just take different shape.

islamaliev (commented):
honestly I don't see much benefit of having this. Some things got a bit better, other not, but overall feels like just shuffling stuff around.

fredcarle:
> honestly I don't see much benefit of having this. Some things got a bit better, other not, but overall feels like just shuffling stuff around.

There doesn't seem to be enough consensus in saying that this in an improvement to warrant merging. Closing for now and if we want to use some of this in the future, the proposed changes will be available as reference.

PR #3232, opened 2024-11-11
refactor: Consolidate node-related fields into a struct

### LINKED ISSUE #3208: Refactor test state to make working with nodes more convenient
At the moment we have test state with a number of slices that hold elements for nodes with corresponding indexes.
With this setup we always need to make sure every slice is slice is properly handled. Instead of using struct of arrays we should switch to array of structs.

We should consolidate all nodes-related fields into on struct `nodeState` and make `state` have a slice of those. This way we would need to worry about arrays only in one place: on node creation.

The main reason why would one give preference to struct or array is in performance-critical applications (like game engines) when you want to to reach high CPU-cache efficiency. This is clearly not our case.

islamaliev:
## Relevant issue(s)

Resolves #3208

## Description

All node-related fields are moved into a separate `nodeState` struct so now we don't need to maintain all slices and node indexes and it's also easier to reason about a node's state.

### DIFF
--- a/tests/integration/db.go
+++ b/tests/integration/db.go
@@ -22,6 +22,7 @@ import (
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/crypto"
 	"github.com/sourcenetwork/defradb/internal/kms"
+	"github.com/sourcenetwork/defradb/net"
 	"github.com/sourcenetwork/defradb/node"
 	changeDetector "github.com/sourcenetwork/defradb/tests/change_detector"
 )
@@ -140,7 +141,7 @@ func getDefaultNodeOpts() []node.Option {
 // setupNode returns the database implementation for the current
 // testing state. The database type on the test state is used to
 // select the datastore implementation to use.
-func setupNode(s *state, opts ...node.Option) (*node.Node, string, error) {
+func setupNode(s *state, opts ...node.Option) (*nodeState, error) {
 	opts = append(getDefaultNodeOpts(), opts...)
 
 	switch acpType {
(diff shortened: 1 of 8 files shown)


tests/integration/state.go
```diff
@@ -114,6 +113,30 @@ func newEventState(bus *event.Bus) (*eventState, error) {
 	}, nil
 }
 
+// nodeState contains all testing state for a node.
+type nodeState struct {
+	// The node active in this test.
+	node *node.Node
+	// The node's client active in this test.
+	client clients.Client
```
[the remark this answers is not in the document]
islamaliev: > suggestion: node.Node and clients.Client seem to basically be the same thing, it looks like it would be minimal hassle to get rid of clients.Client and just call nodeState.node.Peer.Connect(...).

You mean to get rid of `nodeState`'s `client` and instead of for example:
```go
node.client.GetCollections(s.ctx, client.CollectionFetchOptions{})
```
call
```go
node.DB.GetCollections(s.ctx, client.CollectionFetchOptions{})
```
(assuming your first suggestion is applied).

nasdf (approved):
This is a nice improvement to the test framework!

tests/integration/state.go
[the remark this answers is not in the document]
AndrewSisley: Ah - my bad, when I made that suggestion I though `node.Node` implemented `client.DB` and you could have done `node.GetCollections`.  Don't bother with that suggestion :)

shahzadlone (approved):
Love it.

tests/integration/utils.go
```diff
@@ -711,83 +711,35 @@ ActionLoop:
...
-		var addresses []string
-		for _, addr := range s.nodeAddresses[nodeIndex].Addrs {
-			addresses = append(addresses, addr.String())
-		}
-
-		nodeOpts := s.nodeConfigs[nodeIndex]
-		nodeOpts = append(nodeOpts, net.WithListenAddresses(addresses...))
-
-		node.Peer, err = net.NewPeer(s.ctx, node.DB.Blockstore(), node.DB.Encstore(), node.DB.Events(), nodeOpts...)
-		require.NoError(s.t, err)
```
shahzadlone, line 778: question: What happened to this logic?

islamaliev: it's all inside `setupNode`

tests/integration/db.go
```diff
@@ -189,20 +190,52 @@ func setupNode(s *state, opts ...node.Option) (*node.Node, string, error) {
...
+			netOpts = append(netOpts, opt)
+		}
+	}
+
+	if s.isNetworkEnabled {
+		var addresses []string
+		for _, node := range s.nodes {
+			addresses = append(addresses, node.Peer.PeerInfo().String())
+		}
+		netOpts = append(netOpts, net.WithListenAddresses(addresses...))
```
fredcarle, line 212: todo: `WithListenAddresses` is an option to define the node address. What you are doing here is assigning all the nodes known addresses to the new node. Using this option in test is only relevant when restarting a node and the previously known address of that specific node should be used.

That may require adding the address info to the `nodeState` for future usage.

tests/integration/db.go
```diff
@@ -189,20 +190,52 @@ func setupNode(s *state, opts ...node.Option) (*node.Node, string, error) {
...
-	return node, path, nil
+
+	c, err := setupClient(s, node)
+	require.Nil(s.t, err)
+
+	eventState, err := newEventState(c.Events())
+	require.NoError(s.t, err)
+
+	st := &nodeState{
+		Node:    node,
```
fredcarle, line 233: todo: Using node here is breaking the test harnest because it bypasses the use of the selected client by always using the go client. You should change `*node.Node` for `clients.Client`.

tests/integration/db.go
```diff
@@ -189,20 +190,52 @@ func setupNode(s *state, opts ...node.Option) (*node.Node, string, error) {
...
+		event:   eventState,
+		p2p:     newP2PState(),
+		dbPath:  path,
+		netOpts: netOpts,
```
fredcarle, line 237: todo: `netOpts` is store on the struct but never reused. These options should be passed to `setupNode` on restart.

fredcarle (changes requested):
Overall this is a nice change Islam. There are a few issues that I've spotted that caused the tests you pointed out to me to fail. You'll find `todo`s in the relavent areas.

One thing I want to highlight is that this PR managed to make all ci test jobs use the go client regardless of the client selected by the matrix. It would be nice if we could add some kind of validation to prevent this form happening again (out of scope)

While troubleshooting, I've applied changes that fix the issues so if you want to use any or all of it feel free to do so: https://github.com/sourcenetwork/defradb/commit/b3f4fbec0ba36b449e155345be32e08805c2be43#diff-31fde4ee4ab77f7265d4f25d4941c99e626d8b1969c9ff3667e1c6f0c7c4878f

I also noticed that even after fixing, some of the test were a little flaky due to `Dial backoff`. I'm not sure yet what to do about this but if it turns out to be a problem in the CI, I'll look deeper into it.

tests/integration/db.go
islamaliev: applied your proposal

tests/integration/db.go
islamaliev: applied your proposal

PR #3237, opened 2024-11-12
feat: Support for descending fields CLI index creation

### LINKED ISSUE #2460: CLI and HTTP parameter mismatch for creating indexes.
When creating a new index via HTTP, the direction can be set via the `Descending` boolean field. However, it's not possible to set the direction via the CLI.

ChrisBQu:
## Relevant issue(s)

Resolves #2460

## Description

Using the HTTP endpoint, it was possible to create a collection index, with each field being either descending or ascending. This was possible because one of the parameters was a boolean field called "Descending." However, this functionality was not present in the CLI equivalent.

This feature adds support for creating descending fields.

For example, the following sorts name in ascending order, because ascension is the default

`defradb client index create -c User --fields name`

The following sorts it in descending order:

`defradb client index create -c User --fields name:DESC`

And the following sorts multiple fields, in different combinations of descending and ascending order:

`defradb client index create -c User --fields name:ASC,score:DESC`

## Tasks

- [x] I made sure the pull request title adheres to the conventional commit style

## How has this been tested?

- Windows


ChrisBQu:
todo: I will be adding integration tests for this.

islamaliev (commented):
Looks good. Tests would be great. Actually I don't know why there is no tests for this yet. I might have missed this case.

You can look into `tests/integration/index/query_with_composite_index_field_order_test.go` and `tests/integration/index/query_with_index_only_field_order_test.go`. Just copy&paste one of the tests and instead of including an index into `testUtils.SchemaUpdate` create another `testUtils.CreateIndex` with whatever params you deem appropriate.

Our integration tests will run for all 3 clients: go client, HTTP and CLI. So whatever action you write in integration tests will be translated to the "language" of these 3 clients. So you will have to change `CreateIndex` method in `tests/clients/cli/wrapper_collection.go`.

islamaliev (changes requested):
One thing that needs to be added is documentation. Please make sure you include the new change. Couldn't leave an inline comment as you didn't change that part. It's the description in `cli/index_create.go`.

After this you can run `make docs` to sync docs with other places.

tests/clients/cli/wrapper_collection.go
```diff
@@ -348,10 +348,24 @@ func (c *Collection) CreateIndex(
...
+		orders[i] = indexDesc.Fields[i].Descending
+	}
+
+	orderedFields := make([]string, len(fields))
+
+	for i := range fields {
+		if orders[i] {
+			orderedFields[i] = fields[i] + ":DESC"
+		} else {
+			orderedFields[i] = fields[i] + ":ASC"
```
islamaliev, line 364: thought: I wonder if we should skip ":ASC" altogether, otherwise it looks like the default value won't be tested.

cli/index_create.go
```diff
@@ -39,9 +41,33 @@ Example: create a named index for 'Users' collection on 'name' field:
...
+
+			for _, field := range fieldsArg {
+				var fieldName string
+				var order string
+				// For each field, parse it into a field name and ascension order, separated by a colon
+				// If there is no colon, assume the ascension order is ASC by default
+				const asc = "ASC"
+				const desc = "DESC"
+				parts := strings.Split(field, ":")
+				if len(parts) == 1 {
```
AndrewSisley, line 57: suggestion: It looks like the way you have modified `wrapper_collection.go` makes it impossible to test this with our integration tests.  Please have a quick think and see if you can cover this without too much hassle.

Same goes for the `len(parts) > 2` block, but I see that as slightly less important.

AndrewSisley (approved):
Looks really good Chris, solution is nice and simple, and easy to read. Am curious as to how nice/miserable you found writing that first integration test? Congrats on your first PR :)

I have a couple of minor suggestions for you, please respond to them before merging (your response can very much be 'no I think that is a bad idea because of xyz'), as well as making sure Islam and any other reviewers are of course happy :)

cli/index_create.go
islamaliev: I proposed one [here](https://github.com/sourcenetwork/defradb/pull/3237/files#r1841233334), but I'm not sure if it's good enough

tests/clients/cli/wrapper_collection.go
ChrisBQu: It's an interesting thought... By including ":ASC" we test two out of three paths (":DESC" and ":ASC") But if we don't include it, we still only test two out of three paths (":DESC" and unspecified.)

Should we change something?

AndrewSisley: If you change `testultils.IndexedField.Descending` to `immutable.Option[bool]` you should be able to handle the default case, but not the error/len>2 case.

cli/index_create.go
```diff
@@ -22,7 +24,7 @@ func MakeIndexCreateCommand() *cobra.Command {
 	var fieldsArg []string
 	var uniqueArg bool
 	var cmd = &cobra.Command{
-		Use:   "create -c --collection <collection> --fields <fields> [-n --name <name>] [--unique]",
+		Use:   "create -c --collection <collection> --fields <fields[:ASC|:DESC]> [-n --name <name>] [--unique]",
```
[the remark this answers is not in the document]
ChrisBQu: I have now made this change.

cli/index_create.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/sourcenetwork/defradb/client"
+
+	"strings"
 )
```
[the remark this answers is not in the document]
ChrisBQu: I have now made this change.

ChrisBQu:
I discussed the way the CreateIndex function is written with Andy. Currently, it takes in an IndexDescription object. Because of this, the function has access to a boolean indicating whether or not the order should be descending. However, the function does not have access to any context indicating whether a false value means ":ASC" was explicitly passed in, or because nothing was passed in explicitly, and the default value (ascending order) was used.

cli/index_create.go
ChrisBQu: thought: I defined a new error type, NewErrInvalidAscensionOrder which is returned if len(parts) > 2... because if that check is true, the fields parameter must be malformed. I thought this was a reasonable approach. Do we not think so?

Relevant comment also: [https://github.com/sourcenetwork/defradb/pull/3237#issuecomment-2474973523](url)

AndrewSisley: 100% agree with the prod code approach, my comment was only regarding the testing of it :)  It'll hit the same wrapper-problem in the discussion linked by Islam though, so not really worth worrying about in this PR IMO.

fredcarle: suggestion(non-blocking): To remove one of the code paths, you could do something like this:
```go
const asc = "ASC"
const desc = "DESC"
parts := strings.Split(field, ":")
fieldName := parts[0]
order := asc
if len(parts) == 2 {
	order = strings.ToUpper(parts[1])
	if order != asc && order != desc {
		return NewErrInvalidAscensionOrder(field)
	}
} else if len(parts) > 2 {
	return NewErrInvalidAscensionOrder(field)
}
```

islamaliev: you can't do `parts[0]` before checking the `len`

fredcarle: > you can't do parts[0] before checking the len

Why not? `len(parts)` will never be zero.

islamaliev: if you pass an empty string to `strings.Split` it will. This can happen for inputs "name,", ",name", "name,,age", ",", ",," ...

fredcarle: The only way `strings.Split` will return a zero length slice is if both the string and the delimiter are empty `strings.Split("", "")`. This will never be true in this case. The delimiter is always `:` which means the length of the slice will always be at least 1.

https://go.dev/play/p/BQlVFauqCqn

cli/errors.go
```diff
@@ -20,6 +20,7 @@ const (
 	errInvalidLensConfig        string = "invalid lens configuration"
 	errSchemaVersionNotOfSchema string = "the given schema version is from a different schema"
 	errRequiredFlag             string = "the required flag [--%s|-%s] is %s"
+	errInvalidAscensionOrder    string = "invalid order for field %s: expected ASC or DESC"
```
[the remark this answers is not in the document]
ChrisBQu: I have now made this change.

Thanks for the heads up -- you are right, that I did it that way originally because of how I saw errRequiredFlag done. But I see this is cleaner indeed.

cli/index_create.go
ChrisBQu: That makes sense to me. And I did run it to confirm, that yes even if you call it with a blank field, it won't cause a problem. Because instead you will get an error message returned:

Error: flag needs an argument: --fields

Although I think I am on board with Islam's idea of a `NewErrInvalidInxedFieldDescription` error for the case of one with a length > 2.

ChrisBQu: I have added the new error type.

islamaliev (approved):
LGTM! Good job, Chris.

islamaliev:
bug bash result: run different combinations of composite indexes with different directions. Everything work fine.

Issue #3246, opened 2024-11-15
Parent commits only link to one prior head, regardless as to how many there were


AndrewSisley:
Affects composite and collection commits.

For example if a document has two heads, and is then updated, the new head will only link to a single prior head.  This appears to be deterministically selected, so strong eventual consistency is still maintained.

This likely has the same root cause as #3245 in `headset.List`, but I haven't confirmed.

e.g.:
```
testUtils.SchemaUpdate{
	Schema: `
		type Users {
			name: String
		}
	`,
},
testUtils.CreateDoc{
	Doc: `{
		"name":	"John"
	}`,
},
testUtils.UpdateDoc{
	NodeID: immutable.Some(0),
	Doc: `{
		"name":	"Fred"
	}`,
},
testUtils.UpdateDoc{
	NodeID: immutable.Some(1),
	Doc: `{
		"name":	"Shahzad"
	}`,
},
testUtils.ConnectPeers{
	SourceNodeID: 1,
	TargetNodeID: 0,
},
testUtils.WaitForSync{},
testUtils.UpdateDoc{
	NodeID: immutable.Some(1),
	Doc: `{
		"name":	"Chris"
	}`,
},
testUtils.WaitForSync{},
```
Results:
```
map[commits:[map[cid:bafyreiak4w4gmuxs3fotzc5ya5ncwhh5og3a2wxbsoob75yjthu5awmmba docID:bae-6845cfdf-cb0f-56a3-be3a-b5a67be5fbdc fieldId:1 height:2 links:[map[cid:bafyreigqewgxaedp7rghx3yjtj5zmqpfjz2qz2xxlqvehjjrk33uxtjjwi]]] 

map[cid:bafyreigqewgxaedp7rghx3yjtj5zmqpfjz2qz2xxlqvehjjrk33uxtjjwi docID:bae-6845cfdf-cb0f-56a3-be3a-b5a67be5fbdc fieldId:1 height:1 links:[]]

 map[cid:bafyreibtvtndtkdxdffz4bhdgt7igo4vwkmw257vwy5etqwwnajex2ewbm docID:bae-6845cfdf-cb0f-56a3-be3a-b5a67be5fbdc fieldId:1 height:3 links:[map[cid:bafyreie5an34ju3l5ayqq3neyyllw3w26ylvxgswi7c2a7emjpcqdpfglu]]]

[...]

Issue #3245, opened 2024-11-15
Multiple commits at the same height are not returned by `commits` queries


AndrewSisley:
Likely because the code traverses down through the DAG from the head, and because of #3246 multiple heads are not linked to their parent commit.

Both bugs are likely caused by the same problem in `headset.List`, but I haven't confirmed.

I am pretty sure the commits all exist in the blockstore and that this is just a read problem, but I haven't actually verified that.

AndrewSisley:
This was incorrectly opened - I was expecting the commits to sync before they should.  Behaviour is correct.

PR #3247, opened 2024-11-15
feat(i): Handle collection commits over P2P

### LINKED ISSUE #3212: Handle collection commits across P2P system
The P2P system cant currently handle collection-level commits, and needs some changes in order to handle them, so this ticket has been broken out of #3038.

There is a test that documents this, `TestQueryCommitsBranchables_SyncsAcrossPeerConnection`.

AndrewSisley:
## Relevant issue(s)

Resolves #3212 

## Description

Handles the syncing of collection commits over P2P.

~~Note: The behaviour in `TestQueryCommitsBranchables_HandlesConcurrentUpdatesAcrossPeerConnection` is currently under discussion and may be changed or linked to #3245 and #3246 before this PR is merged.~~ Discussed, and test updated.

## Todo

- [x] I forgot to update the wait for sync test action and it looks like some of the new tests are flaking in the CI because of this

### DIFF
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -28,6 +28,7 @@ import (
 	"github.com/sourcenetwork/defradb/event"
 	"github.com/sourcenetwork/defradb/internal/core"
 	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
+	"github.com/sourcenetwork/defradb/internal/core/crdt"
 	"github.com/sourcenetwork/defradb/internal/db/base"
 	"github.com/sourcenetwork/defradb/internal/encryption"
 	"github.com/sourcenetwork/defradb/internal/keys"
@@ -35,30 +36,30 @@ import (
 	merklecrdt "github.com/sourcenetwork/defradb/internal/merkle/crdt"
 )
 
-func (db *db) executeMerge(ctx context.Context, dagMerge event.Merge) error {
+func (db *db) executeMerge(ctx context.Context, col *collection, dagMerge event.Merge) error {
 	ctx, txn, err := ensureContextTxn(ctx, db, false)
 	if err != nil {
 		return err
 	}
 	defer txn.Discard(ctx)
 
-	col, err := getCollectionFromRootSchema(ctx, db, dagMerge.SchemaRoot)
-	if err != nil {
-		return err
+	var mt mergeTarget
+	if dagMerge.DocID != "" {
+		mt, err = getHeadsAsMergeTarget(ctx, txn, keys.HeadstoreDocKey{
+			DocID:   dagMerge.DocID,
+			FieldID: core.COMPOSITE_NAMESPACE,
+		})
+		if err != nil {
+			return err
+		}
+	} else {
+		mt, err = getHeadsAsMergeTarget(ctx, txn, keys.NewHeadstoreColKey(col.Description().RootID))
+		if err != nil {
+			return err
+		}
 	}
 
-	docID, err := client.NewDocIDFromString(dagMerge.DocID)
-	if err != nil {
-		return err
-	}
-	dsKey := base.MakeDataStoreKeyWithCollectionAndDocID(col.Description(), docID.String())
-
-	mp, err := db.newMergeProcessor(txn, col, dsKey)
-	if err != nil {
-		return err
-	}
-
-	mt, err := getHeadsAsMergeTarget(ctx, txn, dsKey.WithFieldID(core.COMPOSITE_NAMESPACE))
+	mp, err := db.newMergeProcessor(txn, col)
 	if err != nil {
 		return err
 	}
@@ -73,9 +74,15 @@ func (db *db) executeMerge(ctx context.Context, dagMerge event.Merge) error {
 		return err
 	}
 
-	err = syncIndexedDoc(ctx, docID, col)
-	if err != nil {
-		return err
+	for docID := range mp.docIDs {
+		docID, err := client.NewDocIDFromString(docID)
+		if err != nil {
+			return err
+		}
+		err = syncIndexedDoc(ctx, docID, col)
+		if err != nil {
+			return err
+		}
 	}
 
 	err = txn.Commit(ctx)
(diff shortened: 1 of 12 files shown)


internal/db/merge.go
```diff
@@ -28,37 +28,38 @@ import (
...
-	docID, err := client.NewDocIDFromString(dagMerge.DocID)
-	if err != nil {
-		return err
-	}
-	dsKey := base.MakeDataStoreKeyWithCollectionAndDocID(col.Description(), docID.String())
-
-	mp, err := db.newMergeProcessor(txn, col, dsKey)
-	if err != nil {
-		return err
+	var mt mergeTarget
```
nasdf, line 46: nitpick: saves a few lines

```go
var key keys.HeadstoreKey
if dagMerge.DocID != "" {
  key = keys.HeadstoreDocKey{...}
} else {
  key = keys.NewHeadstoreColKey(...)
}
```

nasdf (approved):
Nice work! The implementation is simple considering the complexity of the feature.

internal/db/merge.go
AndrewSisley: lol, my bad - thanks for the suggestion Keenan :grin:

- [ ] Simplify executeMerge key block

tests/integration/query/commits/branchables/peer_test.go
```diff
@@ -50,76 +48,153 @@ func TestQueryCommitsBranchables_SyncsAcrossPeerConnection(t *testing.T) {
...
+							}
 						}
-					}
-				}`,
+					}`,
 				Results: map[string]any{
 					"commits": []map[string]any{
 						{
-							"cid": testUtils.NewUniqueCid("collection"),
+							"cid": testUtils.NewUniqueCid(0),
```
AndrewSisley, line 62: note: I was playing around here using ints as cid ids and forgot it was an existing test - I still prefer strings, but let me know your thoughts :)

nasdf: Is the input param needed or could the function generate a random ID in the `NewUniqueCid` func?

AndrewSisley: The input param is purely descriptive, it has no relation to the value of the cid that it represents

AndrewSisley: Although, if I misunderstood you - yes there needs to be a value as otherwise there is no way to describe which values should/shouldn't be the same

nasdf: I ask because if generated internally the pattern for reuse could be:

```go
cidA := testUtils.NewUniqueCid()
cidB := testUtils.NewUniqueCid()
```

Just a thought I had, no need to change the implementation.

AndrewSisley: Ah, I guess we could just make it a pointer and check for equality of the reference - would be a better non-descriptive version than using `int`, but a fair bit more wordy as you'd need to declare the variables if used in multiple places.  (I still prefer strings atm though)

internal/db/messages.go
```diff
@@ -22,7 +22,9 @@ import (
 )
 
 func (db *db) handleMessages(ctx context.Context, sub *event.Subscription) {
-	queue := newMergeQueue()
+	docIdQueue := newMergeQueue()
```
fredcarle, line 25: suggestion: `docIDQueue` would be more idomatic.

AndrewSisley: 2 years later and I'm still messing that up :grin:

- [ ] docIDQueue

tests/integration/acp.go
```diff
@@ -215,7 +215,12 @@ func addDocActorRelationshipACP(
 	}
 
 	if action.ExpectedError == "" && !action.ExpectedExistence {
-		waitForUpdateEvents(s, actionNodeID, map[string]struct{}{docID: {}})
+		expect := make([]map[string]struct{}, action.CollectionID+1)
```
fredcarle, line 218: question: You prefer creating a `CollectionID+1` lenght array that will have just one item instead of passing the collection ID to the `waitForUpdateEvents` function?

AndrewSisley: I started working from the `waitForUpdateEvents` (and late, as a PR fixup).  `waitForUpdateEvents` was multi-collection friendly and so I preserved that without thinking too hard about it, although I think you are right and that all the callers are dealing with a single collection.

Your suggestion makes a lot of sense :)  Although I think I might be able to pass a collection instead of just the index.

- [ ] Simplify `waitForUpdateEvents`

tests/integration/query/commits/branchables/peer_test.go
fredcarle: The strings definitely make this more comprehensible. I much prefer how it was before.

AndrewSisley: I'll change this test back to strings in a sec, I can revert if anyone comes back very soon before merge in favour of the ints @islamaliev  

- [ ] Use strings in test

internal/db/merge.go
```diff
@@ -435,50 +441,59 @@ func decryptBlock(
...
-
+func (mp *mergeProcessor) initCRDTForType(crdt crdt.CRDT) (merklecrdt.MerkleCRDT, error) {
 	schemaVersionKey := keys.CollectionSchemaVersionKey{
 		SchemaVersionID: mp.col.Schema().VersionID,
 		CollectionID:    mp.col.ID(),
 	}
 
-	if field == "" {
-		mcrdt = merklecrdt.NewMerkleCompositeDAG(
+	switch {
```
fredcarle, line 450: praise: Nice use of the block's `IsFoo` methods.

tests/integration/query/commits/branchables/peer_update_test.go
```diff
@@ -0,0 +1,260 @@
...
+							}
+						}
+					}`,
+				Results: map[string]any{
+					"commits": []map[string]any{
+						{
+							"cid": testUtils.NewUniqueCid("collection, update3"),
+							"links": []map[string]any{
+								{
+									"cid": testUtils.NewUniqueCid("collection, update2"),
```
fredcarle, line 94: suggestion: If you keep the node in the string (`collection, node0 update2`) it would help to know where this update comes from.

fredcarle (approved):
LGMT. Good job Andy. Just. one suggestion to improve readability of the expected result.

tests/integration/query/commits/branchables/peer_update_test.go
AndrewSisley: Oh true, I dropped that forgetting that the update was actually applied to a single node (at first).  Will do, thanks Fred.

- [ ] note update node in cid description

PR #3252, opened 2024-11-18
feat: Aggregate filter alias targeting

### LINKED ISSUE #3195: Aggregate filter alias targeting
Add support for targeting aggregates in filters.

nasdf:
## Relevant issue(s)

Resolves #3195

## Description

This PR enables aggregate alias targeting in filters.

Blocked by #3253

## Tasks

## How has this been tested?

Added and updated integration tests.

- MacOS


internal/planner/average.go
```diff
@@ -102,7 +106,7 @@ func (n *averageNode) Next() (bool, error) {
 		return false, client.NewErrUnhandledType("sum", sumProp)
 	}
 
-	return true, nil
+	return mapper.RunFilter(n.currentValue, n.aggregateFilter)
```
AndrewSisley, line 109: question: Why are you running the filter in the aggregate nodes, and not the select nodes they sit on top of (that also run the aggregate filters)?

AndrewSisley: Or, is this more like a scan level filter?  And it is filtering the value produced by the aggregate node?  In which case would it not be better to run the filter in the select node that sits on top of the aggregates, allowing the aggregate nodes to remain focused on aggregating, and not adding an additional code-location in which filters may execute?

Or perhaps we could argue that the select node is doing too much, and we should in fact remove filtering from the Select node's remit, and instead introduce a filter node, removing filtering from select and aggregate nodes.

AndrewSisley (changes requested):
I disagree with the solution and see it as introducing/propagating architectural tech. debt that should be dealt with instead of worsened.

Please either rework this, or fight very hard back against my todo :)

internal/planner/average.go
nasdf: I'm on board with moving it into a new `Filter` node but not within this same PR.

AndrewSisley: What is the rush on this feature?  IMO refactors like this should be done before the feature, not after.  We need to think about the long term implications of how we build features before we build them.

nasdf: There's no rush. I'm trying to keep my PRs more focused, so I'll be doing the refactor in a different PR.

islamaliev:
bug bash result:
played with lots of different scenarios. Found some [quirk behavior](https://github.com/sourcenetwork/defradb/issues/3313). The rest work fine.

islamaliev:
[this one](https://github.com/sourcenetwork/defradb/issues/3315) is probably doesn't not pertain to the task, but found during the bug bash

PR #3258, opened 2024-11-19
fix: Prevent over span

### LINKED ISSUE #3242: After doc deletion the doc can still be fetched from another collection
This test does not pass:
```go
func TestMutationDeletion_WithMultipleCollections_(t *testing.T) {
	test := testUtils.TestCase{
		Description: "Delete a document from one collection and check if the other collection is unaffected.",
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String
					}
					type Pet {
						name: String
						breed: [String]
					}
				`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Shahzad"
				}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "John"
				}`,
			},
			testUtils.CreateDoc{
				CollectionID: 1,
				Doc: `{
					"name": "Scruffy",
					"breed": ["Yorkie"]
				}`,
			},
			testUtils.DeleteDoc{
				CollectionID: 1,
				DocID:        0,
			},
			testUtils.Request{
				Request: `query {
					Pet {
						_docID
					}
				}`,
				Results: map[string]any{
					"Pet": []map[string]any{},
				},
			},
			testUtils.Request{
				Request: `query {
					User {
						_docID
					}
				}`,
				Results: map[string]any{
					"User": []map[string]any{
						{
							"_docID": testUtils.NewDocIndex(0, 1),
						},
						{
							"_docID": testUtils.NewDocIndex(0, 0),
						},
					},
				},
			},
		},
	}

}
```
The last query returns 3 documents.

fredcarle:
## Relevant issue(s)

Resolves #3242 

## Description

This PR solves the situation where deleted documents in the immediate next collection by ID were returned on a full collection query. The reason for this behaviour was due to the fetcher start method redefining the spans based on wanting deleted docs or not and was defining an end key that might have been "prefix-ended" based on a shorter prefix (i.e. collection instead of instance type). ~~The solution was to always redefine the end key as the prefix end of the start key.~~

To fix this we removed the concept of spans and replaced it with a list of prefixes. This results in the fetcher being asked, for example, for all docs in collection 1 with a prefix of `/data/1` instead of a span from `/data/1` to `/data/2`. Furthermore, when the fetcher checks if it need to get deleted docs or non-deleted docs, the resulting prefix becomes `/data/1/< v or d >` instead of the span from `/data/1/< v or d >` to `/data/2/< v or d >` (the span is wrong).

The first commit documents the bug with an integration test.

## Tasks

## How has this been tested?

make test

- MacOS


shahzadlone (commented):
Please add a test or two for the fix.

fredcarle:
> Please add a test or two for the fix.

Right... I forgot to commit them...

internal/db/fetcher/fetcher.go
```diff
@@ -281,17 +281,17 @@ func (df *DocumentFetcher) start(ctx context.Context, spans []core.Span, withDel
...
 			} else {
 				// DocumentFetcher only ever recieves document keys
 				//nolint:forcetypeassert
-				valueSpans[i] = core.NewSpan(
-					span.Start.(keys.DataStoreKey).WithValueFlag(),
-					span.End.(keys.DataStoreKey).WithValueFlag(),
-				)
+				span.Start = span.Start.(keys.DataStoreKey).WithValueFlag()
+				// The end key should always be the prefix end of the start key
+				span.End = span.Start.PrefixEnd()
```
AndrewSisley, line 293: thought: I think this is incorrect if the user provides two docIDs that are next to each other, there is a function that is called earlier that merges spans that sit next to each other - the change here will mean that the second doc would not be fetched.

e.g.:
```
User (docIDs: [baefooooooa, baefoooooob]{
...
}
```
We currently have no integration tests for this as it is quite painful to find docs with docIds next to eachother, but there are unit tests for the func responsible IIRC.

AndrewSisley (changes requested):
It solves an urgent and bad miss from us, but introduces another (less urgent bug) and well as kind of abuses spans.

I think we should find another fix, probably one that handles whatever is causing the span.End to be invalid.

internal/db/fetcher/fetcher.go
islamaliev: we should create a test action `testUtils.MineDocWithID` :D

tests/integration/query/simple/simple_test.go
```diff
@@ -214,3 +214,81 @@ func TestQuerySimpleWithDefaultValue(t *testing.T) {
...
+                        _docID
+                    }
+                }`,
+				Results: map[string]any{
+					"User": []map[string]any{
+						{
+							"_docID": "bae-1ef746f8-821e-586f-99b2-4cb1fb9b782f",
+						},
+						{
+							"_docID": "bae-22dacd35-4560-583a-9a80-8edbf28aa85c",
```
islamaliev, line 264: suggestion: use `testUtils.NewDocIndex` (or whatever the type was)

islamaliev (commented):
Looks good to me, but I will leave the approval to Andy

fredcarle:
> I think we should find another fix, probably one that handles whatever is causing the span.End to be invalid.

@AndrewSisley I don't think. there is anything other than what was there before the change that is making the span.End invalid. I initially thought that the problem was somewhere else but the span coming in is never invalid per say. It's not precise but not invalid. What made it invalid was that the function param span.End was already prefix-ended, as it should be, and prefix-ended once more after the instance type being assigned. This means that different path parameters where being prefix-ended and thus causing an over span.

AndrewSisley:
> as it should be, and prefix-ended once more after the instance type being assigned. 

We can fix this then, and prefix-end the correct key property, it shouldn't be terribly difficult, and then we wont be introducing a new bug/ignoring the span.end

internal/db/fetcher/fetcher.go
fredcarle: If you're talking about `core.MergeAscending`, it's called within the same function after this change to the span and should have no impact when querying two adjacent docs.

fredcarle:
> We can fix this then, and prefix-end the correct key property, it shouldn't be terribly difficult, and then we wont be introducing a new bug/ignoring the span.end

It is fixed with this change.  As stated above, there is no new bug I believe.

AndrewSisley:
@fredcarle I had a bit of a think, and chased down the references to `Spans`.

1.  I think tech. debt wise the change in this PR is not great.  Fetchers take a Spans object with a start and an end, and this PR essentially causes the end value to be ignored.  This PR does not document this in any space that would make that clear to stuff (like plan nodes) dealing with Spans that the normal case (this fetcher) ignores half of the properties it is assigning.  If the solution is left as-is I think it will cost us plenty in the future.

However:

2. `Spans` has always been pretty ugly, both in fetcher and in planner.  I consider it to notably confuse the flow of things in the otherwise moderately confused `planNode` interface in particular.
3. Most of `Spans` is already ignored anyway in the other, more niche fetchers.
4. `core.MergeAscending` is a well meaning but highly ineffective optimisation that almost certainly costs more computing resources than it will ever save.  And it far from cheap maintenance-wise.
5. You as team lead picked this bug up from another developer, who picked this up from a partner.  It seems like people are not terribly keen to work on this right now, however the bug is not great and needs a quick fix.
6. Open PRs aside, I am between tasks right now.

Technically I think the best thing to do is to remove the `Spans` concept entirely, and I think it will be pretty quick to do (a day or so tops).  This includes the `Spans` type, and the `planNode.Spans` functions (as a bonus).  We only ever need to care about an array of prefixes, not an array of ranges.  Even if we want to care about ranges in the future we can do so in a much nicer, localised way that doesn't involve spreading the mess across every implementation of the `fetcher` and `planNode` interfaces.

I suggest you merge the fix as-is, and then I carry out the proposed refactor.

If you disagree with the proposed refactor, I remain against the proposed solution for the reasons in (1).  Let me know :grin:  Sorry if this feels a little hostage-like :sweat_smile:

fredcarle:
> 1. I think tech. debt wise the change in this PR is not great

I don't disagree.

> Technically I think the best thing to do is to remove the Spans concept entirely.

This is what I thought of doing at some point but I wanted a quick fix first because I wasn't sure how deep of a rabbit hole I would be getting into. I should have documented that in the PR description. 

I'm keen on trying to do the "spans to array of prefix" change in this PR. Unless you prefer taking that on.

AndrewSisley (approved):
Looks fantastic, thank you very much for this Fred! And thanks for renaming the various span related variables/docs along the way :)

PR #3256, opened 2024-11-19
feat: Add support for cid-only time travel queries

AndrewSisley:
## Relevant issue(s)

Resolves #3214

## Description

Adds support for cid-only time travel queries. 

Collection-level commit don't work yet, but I'm going to sort that out in another branch.

Also removes some dead code.

PR #3260, opened 2024-11-20
feat: Add support for branchable collection time-traveling

### LINKED ISSUE #3257: Add support for time travel with collection-level cids

AndrewSisley:
## Relevant issue(s)

Resolves #3257

## Description

Add support for branchable collection time-traveling.

Also fixes the docID param which was misbehaving (see commit `Create document with actual docID not user provided value`).

Suggest reviewing commit by commit, it should be easier to follow.

Do not review the first commit here - this branch is based of off #3256 - please review that PR first.

### DIFF
--- a/internal/db/fetcher/versioned.go
+++ b/internal/db/fetcher/versioned.go
@@ -312,6 +312,13 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 
 	var mcrdt merklecrdt.MerkleCRDT
 	switch {
+	case block.Delta.IsCollection():
+		mcrdt = merklecrdt.NewMerkleCollection(
+			vf.store,
+			keys.NewCollectionSchemaVersionKey(vf.col.Description().SchemaVersionID, vf.col.Description().ID),
+			keys.NewHeadstoreColKey(vf.col.Description().RootID),
+		)
+
 	case block.Delta.IsComposite():
 		mcrdt = merklecrdt.NewMerkleCompositeDAG(
 			vf.store,
--- a/internal/db/fetcher/versioned.go
+++ b/internal/db/fetcher/versioned.go
@@ -88,8 +88,6 @@ type VersionedFetcher struct {
 	root  datastore.Rootstore
 	store datastore.Txn
 
-	dsKey keys.DataStoreKey
-
 	queuedCids *list.List
 
 	acp immutable.Option[acp.ACP]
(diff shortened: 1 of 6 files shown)


internal/planner/select.go
```diff
@@ -255,19 +255,24 @@ func (n *selectNode) initSource() ([]aggregateNode, error) {
...
+		// query, which means we need to propagate the values to the underlying VersionedFetcher
 		if n.selectReq.Cid.HasValue() {
 			c, err := cid.Decode(n.selectReq.Cid.Value())
 			if err != nil {
 				return nil, err
 			}
+
+			var docID string
+			if len(n.selectReq.DocIDs.Value()) > 0 {
+				docID = n.selectReq.DocIDs.Value()[0]
```
fredcarle, line 268: thought: Passing the docID for a timetravel query is unnecessary as it will never be used. Even worst, it's confusing the reader by pretending to be useful for the query. One can pass an invalid docID and still get a valid response. We should probably remove references to docID in this block.

tought: It might be more appropriate to allow setting CID or docID but not both in the query params.

internal/db/fetcher/versioned.go
```diff
@@ -339,56 +310,63 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 		return err
 	}
 
-	link, err := block.GenerateLink()
-	if err != nil {
-		return err
-	}
+	var mcrdt merklecrdt.MerkleCRDT
```
fredcarle, line 313: praise: I really like the changes in this method 🙂

fredcarle (approved):
Approving with two thoughts that aren't todos because the underlying issue wasn't introduced in this PR.

internal/planner/select.go
AndrewSisley: Oh I forgot to test that, sorry.

There is a small chance it is still used within the planner nodes, and I actually want it to be so, so that a docID can be provided alongside a collection cid - otherwise there is a gap in the query syntax like you said.  IIRC John really likes being able to set the `docID` and `cid` in the same query, even though the `docID` has always been somewhat pointless from an external perspective.

That said, even then the docID in `Spans` is almost certainly irrelevant (I didn't notice I'd removed it completely from the versionedFetcher) - and so this block deserves a revisit for being highly misleading like you said.

I'll have a play, if it looks like it'll take a bit more effort to get collection-cids with docIDs working I might add a temp error and we can add support whenever it becomes high enough priority.

- [ ] col-cid/docID interplay stuff

AndrewSisley: It worked with the prod code unchanged, test is `TestQuerySimpleWithCidOfBranchableCollectionAndDocID`.

I think I remember actually thinking this through, and saw that the span is passed on to the backing document fetcher.  But looking through the code I don't see that atm.  It might be that the split logic refuses to split the filter as the first fetcher is not a doc-fetcher, leaving the span/select node to handle the docID filter.

Where this happens is not really important now though, I'll see if I can remove the block here without stuff breaking :)

- [ ] Remove docID from prefix (it isn't actually a prefix either)

AndrewSisley: Removing it was problem free. I've documented the oddness here, and will merge this PR once the CI is done.

PR #3263, opened 2024-11-20
fix: Resolve CORS errors in OpenAPI tab of Playground

### LINKED ISSUE #2458: CORS error when using Playground's OpenAPI
The default URL when we start DefraDB is shown at startup to be:
```
$msg="Providing HTTP API at http://127.0.0.1:9181."
```
If we open a browser tab at that URL, we get the playground and we can send GraphQL queries without issue. However, when using the OpenAPI tool, the application sends request to `localhost` instead of `127.0.0.1` and that is causing CORS errors.

Suggested fix is to ensure that the URL used by the OpenAPI tool is the same as the displayed URL on start.

ChrisBQu:
## Relevant issue(s)

Resolves #2458

## Description

There was a bug in the Playground's OpenAPI tool. This tool would list the URL of the endpoints as localhost instead of 127.0.0.1 (which is what the startup Defradb message shows.) The result was that attempting to execute one of the queries returned a CORS error.

This was a one-liner fix. I adjusted openapi.go to use `http://127.0.0.1:9181/api/v0` as a string literal for the URL, instead of `http://localhost:9181/api/v0`

## Tasks

- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in nts, 

## How has this been tested?

- Windows


http/openapi.go
```diff
@@ -111,7 +111,7 @@ func NewOpenAPISpec() (*openapi3.T, error) {
 		Servers: openapi3.Servers{
 			&openapi3.Server{
 				Description: "Local DefraDB instance",
-				URL:         "http://localhost:9181/api/v0",
+				URL:         "http://127.0.0.1:9181/api/v0",
```
fredcarle, line 114: todo: This indeed solves the problem in the case of using the playground locally but in the case where someone wants to expose this endpoint using a different port or a public address, it will be incorrect. I do think that this can be used as a placeholder or documentation for the openAPI server within the codebase but ideally the URL would be assigned based on the configured DefraDB server address.

Unfortunately the openAPI object definition is created within the handler instantiation which doesn't have access to the server options that contain the server address. What can be done though is to pass the server address to the handler function for the openAPI endpoint using the request context. I've created a rough draft of what that might look like here: https://github.com/sourcenetwork/defradb/commit/fd640e2968a92f78796d3b81d9882b151392a393

I'm happy to get on a call to discuss this further if needed 🙂

fredcarle (changes requested):

http/openapi.go
ChrisBQu: I have incorporated these changes, but I have addressed something that Fred and I discussed.

The proposed changes he drafted up contained one bug, in that the URL was treated as a relative path instead of an absolute path. This resulted in URLs of the form: `http://127.0.0.1/127.0.0.1/api/v0`

My proposed fix is to prepend the URL with either HTTP or HTTPS, pending an `isTLS` flag which is set by examining the `options` parameter that is passed to the `NewServer` function of `server.go`

nasdf: Just wanted to add a thought. If you use the http `Host` header from the request it would allow it to work even when behind a reverse proxy, load balancer, or different DNS name.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

http/handler.go
```diff
@@ -86,7 +86,15 @@ func NewHandler(db client.DB) (*Handler, error) {
...
-		responseJSON(rw, http.StatusOK, router.OpenAPI())
+		oas := router.OpenAPI()
+		addr, _ := req.Context().Value(serverAddrKey).(string)
+		isTLS, _ := req.Context().Value(serverTLSKey).(bool)
+		if isTLS {
+			oas.Servers[0].URL = "https://" + addr + "/api/" + Version
+		} else {
+			oas.Servers[0].URL = "http://" + addr + "/api/" + Version
+		}
+		responseJSON(rw, http.StatusOK, oas)
```
fredcarle, line 97: thought: I'm very sorry Chris. After Keenan suggested to use the `req.Host` field I spent a bit of time thinking about this and I'm realizing that what we are doing is futile. There are a couple reasons for this but the most important one it that there are good chances that Defra will be deployed behind a proxy. This means that `isTLS` being false has no relationship to the client using a TLS connection or not. Also, there are no guarantees that `req.Host` will be populated with the correct address. So regardless of how smart we try to be about it, we will never be able to accurately predict the correct URL for all deployment scenarios.  The only thing that is predictable here is the path of the API. As such, this is my suggestion:

suggestion: Leave everything as it was (remove all new code) except for the URL of the OpenAPI server in `http/openapi.go` (see bellow).

http/openapi.go
```diff
@@ -111,7 +111,8 @@ func NewOpenAPISpec() (*openapi3.T, error) {
 		Servers: openapi3.Servers{
 			&openapi3.Server{
 				Description: "Local DefraDB instance",
-				URL:         "http://localhost:9181/api/v0",
+				// This will be overriden by the ServerAddressMiddleware
+				URL: "http://127.0.0.1:9181/api/v0",
```
fredcarle, line 115: suggestion: As stated above, we can change this to `URL: "/api/v0"`. The client will treat it as a relative path to the playground and it will work just fine.

http/server.go
```diff
@@ -56,7 +56,7 @@ type ServerOptions struct {
 // DefaultOpts returns the default options for the server.
 func DefaultServerOptions() *ServerOptions {
 	return &ServerOptions{
-		Address: "127.0.0.1:9181",
+		Address: "http://127.0.0.1:9181",
```
fredcarle, line 59: thought: I'm surprised that this doesn't cause problems. It's probably because the listener ignores the scheme. The scheme will actually be determined by whether TLS has been requested or not. I would prefer not adding it here.

fredcarle (changes requested):
Again, sorry for this back peddling.. Your initial simple fix was the closest to the actual change we should have done.

http/openapi.go
ChrisBQu: I just tested and confirmed that this works, so I think it is also the cleanest way to do it.

http/handler.go
ChrisBQu: I understand, and this makes sense. I have now removed these changes.

http/server.go
ChrisBQu: Understood. I have made this change as requested.

http/openapi.go
ChrisBQu: In light of this discussion, I have rolled back the changes. The PR is now two line changes, one in `openapi.go` and one in the docs.

fredcarle (approved):
LGTM! Thanks for dealing with the flip flop.

On the bright side you got exposed to the middleware system 🙂

PR #3268, opened 2024-11-22
fix: Adjust OpenAPI index POST example request body

### LINKED ISSUE #2459: OpenAPI index POST example request body show's ID as a valid field
Currently, the OpenAPI index POST tab has the example request body indicating that one can set the ID field when creating the index.
```json
{
  "Fields": [
    {
      "Descending": true,
      "Name": "string"
    }
  ],
  "ID": 4294967295,
  "Name": "string",
  "Unique": true
}
```

However, this field is set automatically by the system and will return a `non-zero index ID provided.` error if we try to set it.

Let ensure that the example request body shows only valid fields.

ChrisBQu:
## Relevant issue(s)

Resolves #2459

## Description

On the OpenAPI tool of the playground, the Index POST tab contained an example body that was incorrect. It listed ID as a field which could be set, which was untrue (this value is automatically generated, and trying to pass one in throws an error.) To adjust this, I have created a new description structure called `IndexCreateRequestDescription` which is meant to be used for this type of HTTP request. To go alongside this change, a new schema type, `indexCreateRequestSchema` was also created.

## Tasks

## How has this been tested?

- Windows


client/index.go
```diff
@@ -36,6 +36,18 @@ type IndexDescription struct {
 	Unique bool
 }
 
+// IndexCreateRequestDescription describes an index creation request.
+// It does not contain the ID, as it is not a valid field for the request body.
+// Instead it should be automatically generated.
+type IndexCreateRequestDescription struct {
```
nasdf, line 42: todo: The client interface should be updated to use this new struct.

I would suggest moving it to the file below and changing the `IndexDescription` parameter.

https://github.com/sourcenetwork/defradb/blob/develop/client/collection.go#L118

nasdf (changes requested):
This does fix the OpenAPI issue, but I think we should also update the client interface.

client/index.go
fredcarle: nitpick: `IndexDescriptionCreateRequest` might be more accurate.

ChrisBQu: I have addressed the variable name, that was straightforward.

I have also adjusted the code as Keenan has asked. This resulted in a lot of cascading (small) changes to functions, and function signatures, and in many changes to the integration tests to accommodate those changes. I would like to know if this is what was desired. I will mark for re-review, but I do know that Keenan is out of office this week.

nasdf (approved):
Looks really good! Nice work!

PR #3292, opened 2024-12-04
fix: Add support for operationName and variables in HTTP GET

### LINKED ISSUE #3153: GraphQL HTTP GET is missing operationName and variables
Add support for query params:
- `operationName` string
- `variables` JSON string

ChrisBQu:
## Relevant issue(s)

Resolves #3153

## Description

I have made it so that operationName and variables are supported by HTTP GET requests. I did this by modifying the `ExecRequest` function inside `http/handler_store.go` such that `operationName` and `variables` parameters are extracted if they are present.

I added two new tests to `http/handler_store_test.go` which test this functionality. See:

`TestExecRequest_WithValidQuery_HttpGet_WithOperationName_OmitsErrors`
and
`TestExecRequest_HttpGet_WithVariables_OmitsErrors`

## Tasks

## How has this been tested?

- Windows


http/handler_store_test.go
```diff
@@ -93,3 +94,79 @@ func TestExecRequest_WithInvalidQuery_HasSpecCompliantErrors(t *testing.T) {
...
 }
+
+func TestExecRequest_WithValidQuery_HttpGet_WithOperationName_OmitsErrors(t *testing.T) {
+	cdb := setupDatabase(t)
+
+	query := `query {
+		User {
+			name
+		}
+	}`
```
fredcarle, line 105: todo: Although the test does pass along the operation name in the query string, it doesn't test that it's actually passed to the GraphQL Request. To do so, the query might look like this
```go
query := `query UserQuery {
		User {
			name
		}
	}
        query UserQueryWithDocID {
		User {
                         _docID
			name
		}
	}`
```
And you can ensure that only the name field is returned to know that the operation name was correctly received. Even one operation might be enough to do so.

At the moment, this test will probably fail due to the operation not existing.

playground/package-lock.json
fredcarle, line 1: todo: This shouldn't be part of this PR.

fredcarle (changes requested):
Some initial comments :)

http/handler_store_test.go
fredcarle: Seems like you fixed that already so I'll resolve the comment.

playground/package-lock.json
ChrisBQu: I think I accidentally did that while trying to troubleshoot why defradb wasn't running on my machine anymore (restarting is actually all I needed to do.) In any case, I rolled it back. No harm, no foul. :)

fredcarle (approved):
LGTM! Thanks for taking care of this, Chris :)

PR #3295, opened 2024-12-06
feat: Error if purge request made with dev mode disabled

### LINKED ISSUE #3140: purge command should give feedback if can not be executed
In normal mode:
```shell
$ defradb client purge
Error: run this command again with --force if you really want to purge all data
```
```shell
$ defradb client purge --force
```
There is no output as if it was successful.
Node's logs:
```shell
Oct 15 13:57:19.878 INF cli Received purge event; restarting...
Oct 15 13:57:19.878 ERR cli failed to purge $err="cannot purge database when development mode is disabled"
```
Client's CLI should output that `purge` can not be performed instead of letting it through and silently failing.

ChrisBQu:
## Relevant issue(s)

Resolves #3140

## Description

The issue was that if a purge request was made when the database was not in development mode, it would fail, and an error would be output on the node side of things. However, on the client side nothing indicated that the process had failed. Whether the purge was successful, or it failed, there would be no output.

I have created a new variable in the http package called `IsDevMode`, which is checked in the `http/handler_extras.go/Purge` function. If dev mode is enabled then `htttp.StatusOK` is written to the response header. If it is not enabled, then `http.StatusBadRequest` is written to the header instead, with a message indicating what happened.

## Tasks

## How has this been tested?

- Windows


AndrewSisley (changes requested):
praise: The code looks good and is easy to follow :)  Good job on implementing the chosen design.

todo: I do think the design is faulty here, this will improve things for the `http` and `cli` clients, but not the embedded clients.  And it worsens an existing discrepancy between the embedded Go client and the `http`/`cli` clients.

I expect push back from some other team members on my todo, so maybe chat with them before making changes.

IMO this is adding more tech debt on top of existing tech debt, and it would be much more sensible/cost-effective to fix the existing `purge` problem first instead of continuing to build upon it.

fredcarle (approved):
LGMT. I don't see the tech debt issue raised by Andy as the purge command is a development utility that I don't see a need for on the embedded Go client. Let just wait to resolve the "purge problem" conversation before merging.

AndrewSisley (approved):

Issue #1813, opened 2023-08-24
Replace ds.Query interface with lower-level Iterator interface


jsimnz:
Currently we use the `ds.Query` interface in several places, when really, we only need the general Iterator interfaces provided by most KV stores. 

This will likely require a much more foundational refactor to Defra to replace the general ds.Datastore interfaces, with native abstractions that are designed to expose Iterators instead of Queries.

The reason we need to replace the ds.Query system, is that its basically just a *very bad* implementation of our query planner engine, as it supports complex ordering, filtering, and more. But, if we look at the call sites of where we use the `Query` system, all we need is an iterator. Whats worse, is that those call sites, like in the fetcher, need to access common Iterator methods, like `Seek` which `Query` doesn't provide.

## Update 2024-12-16

A new repo `CoreKV` has been created by John, this contains a new stuff to replace most of the Defra `datastore` package, it exposes `Seek` functionality.

This ticket will now integrate that library into Defra.

jsimnz:
Closed #1600 due to it being a duplicate. Context from that thread:
> Follow up refactor for https://github.com/sourcenetwork/defradb/issues/490 and https://github.com/sourcenetwork/defradb/pull/1500.

AndrewSisley:
Blocked by https://github.com/sourcenetwork/defradb/issues/3354

PR #3298, opened 2024-12-08
refactor: Datastore interfaces replaced with corekv

jsimnz:
## Relevant issue(s)

Related to #3275 #1813

## Description
This PR is ONLY as a reference for #3275 on going by Andy. This refactor is based on a rather old (back in February ~v0.9) version of Defra. There hasn't been many changes to the `datastore` package since then, but there has been a few changes to the `fetcher` refactor, beyond the notable #3277 open PR.

This Draft PR should be able to clearly demonstrate the design approach for how to integrate corekv, how we can delete almost all of the `datastore` package.

What isn't shown in this PR that is a win for replacing `go-datatsore` with `corekv` is just how much more simple the internals of the `corekv.Iterator` is in comparison to the `ds.Query`, not just in API, but internals of the how the Query system actual is evaluated.

Initial refactor to integrate corekv into the codebase, focusing on two goals
1) Simplifying the datastore package design
2) Setup the fetcher to use the `Seek` API in the `fetcher.DocumentFetcher` design to enable early filter optimization

Note: The 2nd goal isn't actually implemented, work was started recently on converting the fetcher seek functionality into this refactor branch, but considering the current open PR for refactoring the fetcher #3277, and the delay to integrate the filter-seek optimization until it is formally verified to be more efficient, i'll leave that work out of this branch.

Additionally, this refactor should be reviewed as a "plain text" refactor. Which means we're not aiming to optimize or redesign, soley to swap datastore interfaces and implementations with the least number of semantic differences.

## How has this been tested?

IT HASN'T, it isn't finished to actually be compilable and therefore testable.

## Notes for Reviewers (Andy ;) )
The commits are in a farily organized and chronological order of operations for the refactor. Some of the commits have some extra context that might help understand reasons.


jsimnz:
Closing now that #3508 is open :)

PR #3317, opened 2024-12-11
feat: P2P ACP

### LINKED ISSUE #3179: Blockservice ACP
~~Wrap the blockstore into an ACP wrapper to handle access control through the blockservice~~

The initial strategy wouldn't support passing the identity to be sent over network calls. Instead, we'll explore modifying the blockservice itself.

fredcarle:
## Relevant issue(s)

Resolves #3179 

## Description

This PR brings ACP functionality to the P2P system. This ensures that permissioned collections do not share blocks of registered documents without ensuring the requesting node has access. This is bypassed on collection replication because the node sending the blocks is sending to peers explicitly.

## Tasks

## How has this been tested?

Some new and updated integration tests.

- MacOS


tests/integration/acp/p2p/create_test.go
```diff
@@ -130,3 +130,147 @@ func TestACP_P2PCreatePrivateDocumentsOnDifferentNodes_SourceHubACP(t *testing.T
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestACP_P2PCreatePrivateDocumentsOnDifferentNodes_SourceHubACP1(t *testing.T) {
```
islamaliev, line 134: todo: adjust the name of the test. 

It doesn't create private documents on different nodes. Just 1 private doc on one node.

tests/integration/acp/p2p/create_test.go
```diff
@@ -130,3 +130,147 @@ func TestACP_P2PCreatePrivateDocumentsOnDifferentNodes_SourceHubACP(t *testing.T
...
+	expectedPolicyID := "fc56b7509c20ac8ce682b3b9b4fdaad868a9c70dda6ec16720298be64f16e9a4"
+
+	test := testUtils.TestCase{
+		Description: "Test acp, p2p create private documents on different nodes, with source-hub",
+		SupportedACPTypes: immutable.Some(
+			[]testUtils.ACPType{
+				testUtils.SourceHubACPType,
+			},
+		),
+		Actions: []any{
```
islamaliev, line 144: question: so the only thing that makes this test different from the KMS ones is that you don't activate KMS here, right?

net/server.go
```diff
@@ -447,3 +493,153 @@ func (s *server) SendPubSubMessage(
 	}
 	return t.Publish(ctx, data)
 }
+
+// hasAccess checks if the requesting peer has access to the given cid.
+func (s *server) hasAccess(p peer.ID, c cid.Cid) bool {
```
islamaliev, line 498: suggestion: took me awhile to figure out that this is used by bitswap. Maybe add a comment?

net/peer.go
```diff
@@ -149,6 +150,10 @@ func NewPeer(
 		return nil, err
 	}
 
+	bswapnet := network.NewFromIpfsHost(h, ddht)
+	bswap := bitswap.New(ctx, bswapnet, blockstore, bitswap.WithPeerBlockRequestFilter(p.server.hasAccess))
```
islamaliev, line 154: question: don't we want to have on option to turn it off? Maybe uses would want just purely pubsub KMS or Orbis. Might also be usefull for testing to be able to turn things on and off.

islamaliev (commented):
Look good overall. Good job, Fred!

I left some comments and suggestions.

In general I think it would be also nice to have it written down somewhere how the new bitswap-level ACP works together with the pubsub KMS. Otherwise it's not clear when to use what.

internal/db/permission/check.go
```diff
@@ -40,6 +40,81 @@ func CheckAccessOfDocOnCollectionWithACP(
...
+// access when acp is not available (acp turned off).
+//
+// Since we know acp is enabled we have these components to check in this function:
+// (1) the request is permissioned (has an identity),
+// (2) the collection is permissioned (has a policy),
+//
+// Unrestricted Access to document if:
+// - (2) is false.
+// - Document is public (unregistered), whether signatured request or not doesn't matter.
+func CheckDocAccessWithIdentityFunc(
```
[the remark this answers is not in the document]
fredcarle: This fonction is used in the net package so it can't be made private.

net/acp.go
```diff
@@ -0,0 +1,31 @@
...
+
+type ACP interface {
+	acp.ACP
+	// GetCollections returns the list of collections according to the given options.
+	GetCollections(ctx context.Context, opts client.CollectionFetchOptions) ([]client.Collection, error)
+	// GetIndentityToken returns an identity token for the given audience.
+	GetIdentityToken(ctx context.Context, audience immutable.Option[string]) ([]byte, error)
+	// GetNodeIdentity returns the node's public raw identity.
+	GetNodeIdentity(ctx context.Context) (immutable.Option[identity.PublicRawIdentity], error)
+}
```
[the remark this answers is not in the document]
fredcarle: I'll try to improve this.

tests/integration/acp.go
```diff
@@ -471,7 +471,7 @@ func setupSourceHub(s *state) ([]node.ACPOpt, error) {
 	//
 	// We need to lock before getting the ports, otherwise they may try and use the port we use for locking.
 	// We can only unlock after the source hub node has started and begun listening on the assigned ports.
-	unlock, err := crossLock(55555)
+	unlock, err := crossLock(44444)
```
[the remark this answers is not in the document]
fredcarle: 55555 is a reserved port on macos. That's why you had a problem with it. I changed it to avoid having other devs on mac have the same issue.

tests/integration/acp/p2p/create_test.go
fredcarle: Thanks for flagging this. I totally forgot to rename after copying.

net/peer.go
fredcarle: I don't think we want to turn this off no. It's not a replacement for KMS or Orbis. If users want to turn it off they can turn off ACP. Having ACP on and this off would be like locking the front door of your house but having a garden level window wide open where anyone can get in. It doesn't really make sense.

net/server.go
```diff
@@ -172,6 +202,22 @@ func (s *server) GetHeadLog(
 	return nil, nil
 }
 
+// GetIdentity receives a get identity request and returns the identity token
+// with the requesting peer as the audience.
+func (s *server) GetIdentity(
```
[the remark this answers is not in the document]
fredcarle: You're probably right that it doesn't need to be public. I just followed the old pattern for GRPC methods. I'll change it.

net/server.go
```diff
@@ -447,3 +493,153 @@ func (s *server) SendPubSubMessage(
...
+		log.ErrorE("Failed to check access", err)
+		return true
+	}
+
+	return peerHasAccess
+}
+
+// verifyAuthToken verifies that the jwt auth token is valid and that the signature
+// matches the identity of the subject.
+func verifyAuthToken(ident identity.Identity, audience string) error {
```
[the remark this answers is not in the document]
fredcarle: it could be. I'll explore it.

net/server.go
fredcarle: will do.

net/acp.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/client"
+)
+
+type ACP interface {
+	acp.ACP
+	// GetCollections returns the list of collections according to the given options.
+	GetCollections(ctx context.Context, opts client.CollectionFetchOptions) ([]client.Collection, error)
+	// GetIndentityToken returns an identity token for the given audience.
+	GetIdentityToken(ctx context.Context, audience immutable.Option[string]) ([]byte, error)
```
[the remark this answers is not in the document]
fredcarle: I agree. Will change.

internal/db/permission/check.go
```diff
@@ -40,6 +40,81 @@ func CheckAccessOfDocOnCollectionWithACP(
...
+// access when acp is not available (acp turned off).
+//
+// Since we know acp is enabled we have these components to check in this function:
+// (1) the request is permissioned (has an identity),
+// (2) the collection is permissioned (has a policy),
+//
+// Unrestricted Access to document if:
+// - (2) is false.
+// - Document is public (unregistered), whether signatured request or not doesn't matter.
+func CheckAccessDocAccessWithDID(
```
shahzadlone, line 72: Todo: Rename to `CheckDocAccessWithDID`. The extra Access is odd

fredcarle: lol. Not sure why I did that.

tests/integration/acp/p2p/create_test.go
fredcarle: That seems to be the most important difference yes.

net/acp.go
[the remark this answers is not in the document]
fredcarle: ✅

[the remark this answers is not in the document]
fredcarle: Let me know what you think of the change.

net/server.go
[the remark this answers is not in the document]
fredcarle: ✅

net/server.go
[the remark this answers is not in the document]
fredcarle: ✅

net/acp.go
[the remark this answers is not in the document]
fredcarle: ✅

tests/integration/acp/p2p/create_test.go
fredcarle: ✅

net/server.go
fredcarle: ✅

internal/db/permission/check.go
fredcarle: ✅

node/node.go
```diff
@@ -160,7 +160,8 @@ func (n *Node) Start(ctx context.Context) error {
 			coreDB.Blockstore(),
 			coreDB.Encstore(),
 			coreDB.Events(),
-			NewNetACP(acp, coreDB),
+			acp,
+			coreDB,
```
shahzadlone, line 164: suggestion: At this point do we just want to remove the `coreDB.XYZ` args above and just consume the one `DB` arg (`coreDB` here)

internal/db/permission/check.go
```diff
@@ -40,6 +40,81 @@ func CheckAccessOfDocOnCollectionWithACP(
...
+	did string,
+	acpSystem acp.ACP,
+	collection client.Collection,
+	permission acp.DPIPermission,
+	docID string,
+) (bool, error) {
+	identityFunc := func() immutable.Option[acpIdentity.Identity] {
+		if did == "" {
+			return immutable.None[acpIdentity.Identity]()
+		}
```
shahzadlone, line 83: question: Do we really want this? Specially with the new changes that were introduced with "*" pattern. Also does any test hit this block?

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+			s.piMux.Unlock()
+		}
+		return immutable.Some(ident)
+	}
+
+	peerHasAccess, err := permission.CheckDocAccessWithIdentityFunc(
+		s.peer.ctx,
+		identFunc,
+		s.peer.acp.Value(),
+		cols[0], // For now we assume there is only one collection.
```
shahzadlone, line 538: todo: Might be worth marking this with a issue.

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+			s.piMux.Lock()
+			s.peerIdentities[p] = ident
+			s.piMux.Unlock()
+		}
+		return immutable.Some(ident)
+	}
+
+	peerHasAccess, err := permission.CheckDocAccessWithIdentityFunc(
+		s.peer.ctx,
+		identFunc,
```
shahzadlone, line 536: question: Does this have to be a function? Why not resolve the identity stuff here and then pass the identity to the check call?

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+		log.Info("No collections found", corelog.Any("Schema Version ID", block.Delta.GetSchemaVersionID()))
+		return true
+	}
+	ident, err := s.peer.db.GetNodeIdentity(s.peer.ctx)
+	if err != nil {
+		log.ErrorE("Failed to get node identity", err)
+		return true
+	}
+	if !ident.HasValue() {
+		log.Info("No node identity found")
```
shahzadlone, line 578: suggestion: Perhaps also indicate that therefore "self has access"?

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+		return true
+	}
+
+	peerHasAccess, err := permission.CheckDocAccessWithDID(
+		s.peer.ctx,
+		ident.Value().DID,
+		s.peer.acp.Value(),
+		cols[0], // For now we assume there is only one collection.
```
shahzadlone, line 586: suggestion: Same as above

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+		acp.ReadPermission,
+		string(block.Delta.GetDocID()),
+	)
+	if err != nil {
+		log.ErrorE("Failed to check access", err)
+		return true
```
shahzadlone, line 592: question: If access checking fails we will assume they have access?

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+	if len(cols) == 0 {
+		log.Info("No collections found", corelog.Any("Schema Version ID", block.Delta.GetSchemaVersionID()))
+		return true
+	}
+	ident, err := s.peer.db.GetNodeIdentity(s.peer.ctx)
+	if err != nil {
+		log.ErrorE("Failed to get node identity", err)
+		return true
+	}
+	if !ident.HasValue() {
```
shahzadlone, line 577: question: Is `ident` the identity of the node that may or may not be set? i.e. not in any way the client's requesting identity correct?

internal/db/db.go
```diff
@@ -104,7 +104,7 @@ func NewDB(
 	acp immutable.Option[acp.ACP],
 	lens client.LensRegistry,
 	options ...Option,
-) (client.DB, error) {
+) (*db, error) {
```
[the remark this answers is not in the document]
fredcarle: There are a few reasons. One of them is because otherwise we have to make the client.DB API support everything even though some methods may just be useful internally. The maintenance overhead of that is large since we have to update the CLI and HTTP clients and test wrappers every time we update the interface. On top of that, there is no good reason to return an interface if the returned value is always of the same type. It's an unnecessary abstraction.

internal/db/db.go
```diff
@@ -339,13 +339,20 @@ func (db *db) DeleteDocActorRelationship(
 	return client.DeleteDocActorRelationshipResult{RecordFound: recordFound}, nil
 }
 
-func (db *db) GetNodeIdentity(context.Context) (immutable.Option[identity.PublicRawIdentity], error) {
+func (db *db) GetNodeIdentity(_ context.Context) (immutable.Option[identity.PublicRawIdentity], error) {
```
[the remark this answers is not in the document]
fredcarle: Because just `context.Context` seems like a mistake. The underscore says that the function parameters are this way because we're following some interface pattern or something like that but we aren't using this parameter within the function.

internal/db/db.go
```diff
@@ -339,13 +339,20 @@ func (db *db) DeleteDocActorRelationship(
...
 	}
 	return immutable.None[identity.PublicRawIdentity](), nil
 }
 
+func (db *db) GetIdentityToken(_ context.Context, audience immutable.Option[string]) ([]byte, error) {
+	if db.nodeIdentity.HasValue() {
+		return db.nodeIdentity.Value().NewToken(time.Hour*24, audience, immutable.None[string]())
+	}
+	return nil, nil
+}
```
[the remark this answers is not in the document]
fredcarle: It's a pattern we decided to keep after a discussion we had several months back about removing context. Happy to bring that up for discussion but for now leaving as is.

internal/db/permission/check.go
```diff
@@ -40,6 +40,81 @@ func CheckAccessOfDocOnCollectionWithACP(
...
+// Since we know acp is enabled we have these components to check in this function:
+// (1) the request is permissioned (has an identity),
+// (2) the collection is permissioned (has a policy),
+//
+// Unrestricted Access to document if:
+// - (2) is false.
+// - Document is public (unregistered), whether signatured request or not doesn't matter.
+func CheckDocAccessWithIdentityFunc(
+	ctx context.Context,
+	identityFunc func() immutable.Option[acpIdentity.Identity],
```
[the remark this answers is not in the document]
fredcarle: Because in some cases in doesn't make sense to do an extra network call for the identity before knowing if the document is even registered.

net/grpc.go
```diff
@@ -52,6 +53,17 @@ type pushLogRequest struct {
 
 type pushLogReply struct{}
 
+type getIdentityRequest struct {
```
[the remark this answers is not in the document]
fredcarle: What else would it be? Do you foresee more types of identity within the `net` package?

net/grpc.go
```diff
@@ -63,6 +75,30 @@ type serviceServer interface {
...
+	GetIdentity(context.Context, *getIdentityRequest) (*getIdentityReply, error)
+}
+
+func getIdentityHandler(
+	srv any,
+	ctx context.Context,
+	dec func(any) error,
+	interceptor grpc.UnaryServerInterceptor,
+) (any, error) {
+	in := new(getIdentityRequest)
```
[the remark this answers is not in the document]
fredcarle: In probably stands for input.

net/grpc.go
```diff
@@ -63,6 +75,30 @@ type serviceServer interface {
 	PushLog(context.Context, *pushLogRequest) (*pushLogReply, error)
 	// GetHeadLog from this peer
 	GetHeadLog(context.Context, *getHeadLogRequest) (*getHeadLogReply, error)
+	GetIdentity(context.Context, *getIdentityRequest) (*getIdentityReply, error)
+}
+
+func getIdentityHandler(
+	srv any,
+	ctx context.Context,
+	dec func(any) error,
```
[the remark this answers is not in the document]
fredcarle: These handler functions are copy pasted from the auto generated code that we use to have with protobufs. The GRPC server expects to have this function signature for handlers and I'm simply reusing this here for simplicity.

net/grpc.go
```diff
@@ -63,6 +75,30 @@ type serviceServer interface {
...
+		return srv.(serviceServer).GetIdentity(ctx, in)
+	}
+	info := &grpc.UnaryServerInfo{
+		Server:     srv,
+		FullMethod: serviceGetIdentityName,
+	}
+	handler := func(ctx context.Context, req any) (any, error) {
+		return srv.(serviceServer).GetIdentity(ctx, req.(*getIdentityRequest))
+	}
+	return interceptor(ctx, in, info, handler)
```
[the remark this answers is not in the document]
fredcarle: Reaching this line depends on the GRPC server behaviour. Since it's reused from generated code, I'm not interested in creating mocks just to get full coverage.

node/node.go
fredcarle: Fair point. I'll look into it.

internal/db/permission/check.go
fredcarle: I could add a unit test to test this but the reason it's there is so that all the checks have the same behaviour regarding passing no identity. This is not new behaviour and seems to be aligned with the "*" pattern.

net/server.go
fredcarle: I don't think it's worth it no as we might never have more than one. It's only sharing information as to why we select index 0.

net/server.go
fredcarle: Reason stated above

net/server.go
fredcarle: The function doc states that already. Not sure it's worth adding everywhere we return

net/server.go
fredcarle: it's not they. It's self. It's a best effort check. If the check fails, the node receiving the request can make the decision.

net/server.go
fredcarle: Five lines above :)
```
ident, err := s.peer.db.GetNodeIdentity(s.peer.ctx)
```

internal/db/db.go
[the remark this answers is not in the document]
AndrewSisley: > net.DB doesn't implement client.DB as GetNodeIdentityToken method is missing

This feels regressive, and to me it looks like we are forgetting the lessons learnt before `client.DB` was implemented by all packages - allowing the 3 clients to diverge (they now diverge in both directions, it is no longer a case of CLI contains embedded+some, CLI is now missing embedded stuff).

Returning a private type here, with public functions not declared in the `client` package also smashes the reasoning behind the `client` package in the first place - `GetNodeIdentityToken` is essentially a hidden public function that lives in an unexpected location, a location that we are very likely to forget about when making breaking changes (think semver).  And will unlikely be documented to the same standards as `client` members.

The multiple clients act as complexity multipliers.  The greater the divergence between clients, the greater that multiplier is.  This increases our maintenance costs (including documentation) and increases the learning required by users, in way that is far greater than the immediate X lines of code added.

IMO we need to be very very careful doing stuff like this, and seeing this so soon after another divergence (`purge` CLI command), makes it feel like we are falling back towards the time-sink that was our 3 un-interfaced, separate clients without making a conscious decision to do so.

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+	}
+
+	cols, err := s.peer.db.GetCollections(
+		s.peer.ctx,
+		client.CollectionFetchOptions{
+			SchemaVersionID: immutable.Some(block.Delta.GetSchemaVersionID()),
+		},
+	)
+	if err != nil {
+		log.ErrorE("Failed to get collections", err)
```
AndrewSisley, line 565: suggestion: By not returning the error here, and in a few other similar places you are obfuscating where the 'best effort' is actually required and makes sense.  As far as I can see, it looks like there is no good reason for `GetCollections` to fail and we should be pushing the error up the call stack even though the check is somewhat optional.

net/server.go
```diff
@@ -447,3 +461,136 @@ func (s *server) SendPubSubMessage(
...
+		log.ErrorE("Failed to check access", err)
+		return false
+	}
+
+	return peerHasAccess
+}
+
+// trySelfHasAccess checks if the local node has access to the given block.
+//
+// This is a best-effort check and returns true unless we explicitly find that we don't have access.
```
AndrewSisley, line 552: todo: Please document why we are even bothering if a 'best effort' is correct enough.

If the reason is that a proper check will be carried out somewhere else (e.g. another node), please also document why false-negatives are either impossible, or are tolerable, as it looks like we could refuse to process a block that should actually be processed.

internal/db/db.go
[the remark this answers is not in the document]
fredcarle: I actually did the change because I believe returning an interface where only one concrete type is ever returned is a needlessly restrictive abstraction.

The goal of an interface is to ensure that a given type implements a set of methods. **Not to limit** that type to the set of methods. There is a set of functionalities that we decided had to be covered by all clients/types (Go, CLI, HTTP) and that's where the interface is useful. If anything changes with those functionalities, it needs to change for all clients. 

There are some functionalities that will only ever make sense for one client. Purge for example only really makes sense for a server instance so it would be unnecessary overhead to require it for all clients through the client.DB interface.

What I think we can do is have a discussion on whether we want a given functionality to be part of the interface or not. But I really do think that some divergence between server and embedded is quite ok when it makes sense.

net/server.go
fredcarle: > as it looks like we could refuse to process a block that should actually be processed.

I think you probably meant the opposite?

This method is for **self** check of access so it will only block processing if it can confirm no granted access. In any other case it will continue with the sync process.

net/server.go
shahzadlone: I still am not sure I follow, if the local node receiving (self) fails to check weather it has access then the check will assume it does have access and the sender node can still decide if they want to share or not?

net/grpc.go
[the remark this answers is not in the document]
shahzadlone: I didn't know it's reused from generated code. thanks for explaination.

net/grpc.go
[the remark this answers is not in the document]
shahzadlone: Just didn't want to confuse with client identity.

internal/db/permission/check.go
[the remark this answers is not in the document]
shahzadlone: Can you explain a bit more? Which document is registered check are we talking about? and which network call for identity?

internal/db/db.go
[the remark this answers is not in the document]
shahzadlone: This is a known go idiom?

net/grpc.go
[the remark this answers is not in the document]
fredcarle: There is no discrimination of node and client in this package so just Identity is fine in my opinion.

internal/db/db.go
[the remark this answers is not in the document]
fredcarle: It is the same as the other areas we use `_` to ignore a returned argument. https://go.dev/ref/spec#Blank_identifier

The only reason why just using `context.Context` worked is because it was the only parameter in the method. If there was more than one, it would be a syntax error.

Unused function parameters are allowed in Go and that's why we often have `ctx context.Context` as parameter with `ctx` never being used. I feel that's misleading. Doing `_ context.Context` at least tells the reader that this parameter needed to be there but we aren't using it.

net/server.go
fredcarle: Yes. The initial block received is just a composite so there is no protected info within it. The node receiving checks if it has access to the document to see if it's worth requesting the other block. Ultimately, the node sending is responsible for the verification.

internal/db/permission/check.go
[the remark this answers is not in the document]
fredcarle: In the net package, when we want to check if a peer has access to a doc we have to request the identity from the node requesting. That's a network call. If we request it before checking that the document is registered, then we might have done that network call for nothing. So the order of priority should be checking that the doc is registered and then requesting the identity to check if it has access.

node/node.go
fredcarle: Made the change 👍

net/server.go
fredcarle: Changed to return an error

nasdf (changes requested):
Implementation looks solid. One blocking issue to resolve.

internal/db/db.go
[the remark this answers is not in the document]
AndrewSisley: > Not to limit that type to the set of methods. 

I quite strongly disagree with this, limiting a type to a sub-set of methods is a major reason for interface usage and is known as the interface segregation principle (the `I` in `SOLID`).

It reduces the surface area of the codebase and the number of pathways through it.  This reduces user-complexity, and the number of support tickets we will receive (as they wont be sending us tickets for funcs that are hidden by the interface).

net/server.go
AndrewSisley: > I think you probably meant the opposite?

Nope.

This function looks like it can falsely claim that we dont have access (e.g. local node doesn't have access (yet), but the other remote node does), in which case the block will not be processed.

fredcarle: > This function looks like it can falsely claim that we dont have access (e.g. local node doesn't have access (yet), but the other remote node does), in which case the block will not be processed.

`Doesn't have access yet` is a correct claim that we don't have access. It doesn't matter what the remote node has access to. Trying to process a block that we know we don't have access will just result in the remote node not giving us access anyways.

AndrewSisley: Ah okay - can you add that to the func docs, along with a bit detailing as to why a 'best-effort' is relevant/okay here?

internal/db/db.go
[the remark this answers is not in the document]
fredcarle: This is the `I` in `SOLID`
> A client should never be forced to implement an interface that it doesn’t use, or clients shouldn’t be forced to depend on methods they do not use.

In my interpretation, limiting a type to a sub-set of methods is not what the above means, at all.

net/server.go
fredcarle: It is documented already no?

AndrewSisley: Ah yes it has been since I opened this thread, thanks!  LGTM :)

internal/db/db.go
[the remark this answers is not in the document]
AndrewSisley: I see the signature of concrete type as an interface, by exposing the maximalist concrete interface you expose consuming code to problems in swapping it out for something else (essentially reducing the usefulness of defining the actual interface-interface).

That aside, IMO interface segregation goes beyond that one liner.  By segregating via interfaces you control the number pathways through the codebase and reduce complexity.

We don't want our users to be able to call all our internal funcs, and they probably don't want to see them either.

We end up with a public function that returns the public interface.  The user-code that consumes that type only ever directly depends-on/has-access-to the client/interface funcs, instead of a maximal concrete implementation.

The contract/interface of the public `NewDB` function then also remains identical to all other `NewDB` functions, allowing them to be used interchangeably without the swap being a breaking change.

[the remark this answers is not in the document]
fredcarle: > I see the signature of concrete type as an interface, by exposing the maximalist concrete interface you expose consuming code to problems in swapping it out for something else (essentially reducing the usefulness of defining the actual interface-interface).

Why would `internal/DB` be swapped for something else? And if it is, what prevents it from supporting the `client.DB` interface?

> That aside, IMO interface segregation goes beyond that one liner. By segregating via interfaces you control the number pathways through the codebase and reduce complexity.

That one maybe. But I find that overly constraining to the point where one has to create more complex patterns to bypass the restrictions.

> We end up with a public function that returns the public interface. The user-code that consumes that type only ever directly depends-on/has-access-to the client/interface funcs, instead of a maximal concrete implementation.

We don't. `NewDB` is an internal function. Users don't have access to it.

> The contract/interface of the public NewDB function then also remains identical to all other NewDB functions, allowing them to be used interchangeably without the swap being a breaking change.

There is only one `NewDB` function. But even if we created many other DB alternatives, the important part is that those DB types cover the `client.DB` interface so that our http and cli clients along with the other internal types that need `client.DB` can use the `client.DB` methods. It doesn't matter that those other DB types provide more features.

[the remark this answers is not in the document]
shahzadlone: While I agree with the points raised by Andy, I don't think it's worth blocking the PR over. Specially in light of us doing the "moving fast experiment" I happy for this discussion to even resolve outside this PR. Just noting that the helper private function that was being called `newDB` was already using this signature before (irrelevant to the discussion I know).

Side question: I noticed that since the discussion thread was started the `*db` was made public (`*DB) now, curious for reasons behind that?

internal/db/permission/check.go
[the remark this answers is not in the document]
shahzadlone: Is the identity not included from the node requesting from the get go? Maybe I am missing something, why does it need to be requested again?

net/server.go
shahzadlone: Ah I got it now so the best effort check is there for the receiver to not make unnecessary requests (for it's own sake to not waste effort if it doesn't have access to other blocks). I forget if this is documented somewhere already, but might be worth it to have a one liner comment explaining this near here somewhere.

shahzadlone (approved):
LGTM, thanks for answering all the questions, just left some minor suggestions. Please do resolve the on going discussions before merge from others.

todo: Please make the PR title more descriptive right now it only says `P2P ACP` which is quite vague for example it can be confused with the initial P2P acp work that has been done for https://github.com/sourcenetwork/defradb/issues/2366

internal/db/db.go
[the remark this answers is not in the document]
AndrewSisley: > We don't. NewDB is an internal function. Users don't have access to it.

FFS lol - I forgot we moved it, sorry. 90% of me bothering about this is because I thought it was still a public func and you were exposing internals to users.

> There is only one NewDB function

There are ones for the other 2 clients, and one from the node package no? They may have different names atm, but they are all basically the same func (and should probs have the same name).

> Why would internal/DB be swapped for something else?

The callers to the `NewDB` func(s) may wish to swap them out - this becomes much harder if the type changes and extra/non-interface stuff on the return value is used.

But anyway, this is just internal stuff, the extra exposure does make stuff easier, and we (in theory :)) should really be able to tell what the newly exposed stuff does anyway.

Happy for this convo to be resolved lol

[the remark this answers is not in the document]
jsimnz: Just adding for historical sake since I think Andy has dropped this (and as shahzad pointed out, we're aiming not to get bogged down during this cycle) - I think adding compile time checks to the `db` package that make sure the `DB` type implements the `client.DB` is all that is really necessary. That along with the downstream callers (internal to the repo) having the option to assign/cast to `client.DB` for their own safety if they want should cover us for most of our cases.

Fred and I briefly discussed this whole return type a few weeks ago and one thing it raised is we currently don't have a proper "embedded" client flow ever since we moved the `db` package to `internal`. The way the integration tests access the embedded flow is to instanciate a `node.Node` instance, then getting the `client.DB` instance from it. But we should have a dedicated and public flow for instantiating the `client.DB` without the complexitity of the `node.New` constructor as follow up work to this

internal/db/permission/check.go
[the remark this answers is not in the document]
jsimnz: I'd also like clarification here :)

net/peer.go
```diff
@@ -61,15 +75,18 @@ type Peer struct {
...
 	bootCloser io.Closer
 }
 
 // NewPeer creates a new instance of the DefraDB server as a peer-to-peer node.
 func NewPeer(
 	ctx context.Context,
-	blockstore datastore.Blockstore,
-	encstore datastore.Blockstore,
 	bus *event.Bus,
+	acp immutable.Option[acp.ACP],
```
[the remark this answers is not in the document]
fredcarle: I'm not sure in implies more safety and I'm not sure it matters it the ACP implementation is the same or not. I could see a case for net ACP being different than db ACP.

net/server.go
```diff
@@ -447,3 +460,135 @@ func (s *server) SendPubSubMessage(
 	}
 	return t.Publish(ctx, data)
 }
+
+// hasAccess checks if the requesting peer has access to the given ci.
```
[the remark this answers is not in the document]
fredcarle: fixed!

net/server.go
```diff
@@ -447,3 +460,135 @@ func (s *server) SendPubSubMessage(
...
+		log.ErrorE("Failed to get collections", err)
+		return false
+	}
+	if len(cols) == 0 {
+		log.Info("No collections found", corelog.Any("Schema Version ID", block.Delta.GetSchemaVersionID()))
+		return false
+	}
+
+	// If the requesting peer is in the replicators list for that collection, then they have access.
+	if peerList, ok := s.replicators[cols[0].SchemaRoot()]; ok {
```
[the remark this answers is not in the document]
fredcarle: thanks for spotting this!

Fixed

nasdf (approved):

AndrewSisley:
Bug bash result: No bugs found.

PR #3331, opened 2024-12-16
feat: Add 'keyring list' command to CLI

### LINKED ISSUE #2756: No way to list keys using `keyring` command
Users can generate, import and export individual keys via the CLI, but they have no way of view which keys are held by the keyring.  I suggest adding a `list` command or something so that they can find this out.

ChrisBQu:
## Relevant issue(s)

Resolves #2756

## Description

I am adding support for a new feature in the CLI. By running `defradb keyring list`, the names of the keys in the fileKeyring will be listed out.

## Tasks

## How has this been tested?

- Windows


fredcarle:
Please add the `feat: ` prefix to the title 🙂

keyring/system.go
```diff
@@ -53,3 +53,9 @@ func (s *systemKeyring) Get(name string) ([]byte, error) {
 func (s *systemKeyring) Delete(user string) error {
 	return keyring.Delete(s.service, user)
 }
+
+func (s *systemKeyring) List() ([]string, error) {
+	// List does not need to be returned here
+	// This function is a stub
```
fredcarle, line 59: suggestion: It would be nice to have a bit more details as to why list does not need to be returned for the system keyring.

cli/keyring_list.go
```diff
@@ -0,0 +1,44 @@
+package cli
```
fredcarle, line 1: todo: Please add a test file for this command like we did for the other keyring commands. This won't be tested in the integration tests and it's simple to do so within the `cli` package.

ChrisBQu: Absolutely. This one was always on my to-do list. I now have it added.

keyring/system.go
ChrisBQu: This seems like a fair suggestion. I agree with it. I've added a bit more to it.

I'm not sure how much detail you (or other team members(?)) think this comment should go into. I'm open to discussion on this point if anyone feels strongly.

keyring/system.go
```diff
@@ -53,3 +53,10 @@ func (s *systemKeyring) Get(name string) ([]byte, error) {
 func (s *systemKeyring) Delete(user string) error {
 	return keyring.Delete(s.service, user)
 }
+
+func (s *systemKeyring) List() ([]string, error) {
+	// The OS keyring does not support listing keys
+	// This function is a stub for now because the Keyring interface requires it
+	// Currently, the 'defradb keyring list' command uses only fileKeyring
+	return nil, nil
```
AndrewSisley, line 61: todo: I was worried you would leave this func like this :) Sorry I should have clarified on discord.  I agree that this is the best choice, however I think it should error, not silently return - otherwise the users will go bald from the large amount of head scratching that they'll do when trying to figure out why their added keys never appear in the keyring! (it looks like it is empty when the call list on an os-keyring no?)

AndrewSisley (commented):
I would request changes, but am on holiday and don't want to block the PR.  Please address/push-back my one todo before merge :)

keyring/system.go
ChrisBQu: I see the logic behind it throwing an error. My original thoughts were that this function should never actually get called so there was no need to have it do anything. But as I think over it some more, that's all the more reason to make it throw an error. I think you are right.

ChrisBQu: I've made the changes. I'm going to tag you for re-review, so let me know what you think!

fredcarle (approved):
LGTM. Just a couple comments that you can resolve if you want to.

keyring/system_keyring_list_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func TestSystemKeyringListThrowsError(t *testing.T) {
+	service := "test-service"
+	systemKeyring := OpenSystemKeyring(service)
+
+	keys, err := systemKeyring.List()
+
+	assert.Nil(t, keys, "keys should be nil when List is called")
+	assert.Error(t, err, "an error should be returned when List is called")
+	assert.Equal(t, ErrSystemKeyringListInvoked, err, "the error should be ErrSystemKeyringListInvoked")
```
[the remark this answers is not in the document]
ChrisBQu: Still a bit of growing pains getting used to writing Go! But I agree with this suggestion (both parts.) 

Making this change.

islamaliev:
Bug bash result:
When running a defradb node with keyring = "system" I get this:
```
$ defradb keyring list
Error: listing keys is not supported by OS keyring
```
Although another keyring command seems to work fine:
```
$ defradb keyring export peer-key
6de50e8f154ea884fed826a118da3c95a4067029313cbde5dc5c005332be55f49403ca1798a8aa6dea29579f53b089285c30682830f2ec9de47e0792fb69fb91
```

PR #3352, opened 2025-01-02
feat: Multiple schema file support in schema add

### LINKED ISSUE #1248: `schema add` command to support multiple files
Status quo: only the last file added via `-f` is considered

ChrisBQu:
## Relevant issue(s)

Resolves #1248 

## Description

Previously, a single schema file could be added by calling `defradb schema add -f filename`. Now we can add multiple new schemas from different files in the following way: `defradb schema add -f file_one -f file_two -f file_three`

This PR also adjusts some import statements that were using the incorrect `error` import.

## Tasks

## How has this been tested?

- Windows


cli/config.go
```diff
@@ -20,6 +19,8 @@ import (
 	"github.com/sourcenetwork/corelog"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+
+	"github.com/sourcenetwork/defradb/errors"
```
shahzadlone, line 23: suggestion: Might be worth adding a linter rule for this check.

ChrisBQu: I think that does make sense. But should it be a separate issue? I discussed with Fred how its a little weird that the error imports are a part of this PR at all, because they realistically were meant to be part of a previous (already merged) issue. Because its such a small thing, he agreed that leaving it here is probably fine.

But this to say: I don't believe a new linter rule would fall into the domain of this particular PR (?)

jsimnz: I think the 3 instances of changing `errors` to `defradb/errors` in this PR should *all* be in a separate PR along with the linter rule change (or technically that can be two follow up PRs).

Theres an argument to be made (happy to concede) that this specific change in the `cli` package is OK since this PR is making changes to the `cli` package so its all fair game.but the `internal/db` and `crypto` packages shouldnt be included in this PR

jsimnz: its up to you Chris what you want to do within this PR (with the exception of the linter rule, that def should be in a dedicated PR)

ChrisBQu: I have considered all of the options, and reverted the import statements. I think it is reasonable to approach it as a separate PR, with the linter rule change as you both suggest.

cli/schema_add.go
```diff
@@ -36,40 +36,56 @@ Example: add from an argument string:
...
-			case schemaFile != "":
-				data, err := os.ReadFile(schemaFile)
-				if err != nil {
-					return err
+			case len(schemaFiles) > 0:
+				// Read schemas from files and concatenate them
+				for _, schemaFile := range schemaFiles {
+					data, err := os.ReadFile(schemaFile)
+					if err != nil {
+						return fmt.Errorf("failed to read file %s: %w", schemaFile, err)
```
shahzadlone, line 56: todo: please move the errors to the `cli/errors.go` file.

cli/schema_add.go
```diff
@@ -36,40 +36,56 @@ Example: add from an argument string:
...
+					combinedSchema += string(data) + "\n"
 				}
-				schema = string(data)
+
 			case len(args) > 0 && args[0] == "-":
+				// Read schema from stdin
 				data, err := io.ReadAll(cmd.InOrStdin())
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to read schema from stdin: %w", err)
```
shahzadlone, line 65: todo: please move the errors to the `cli/errors.go` file.

cli/schema_add.go
```diff
@@ -36,40 +36,56 @@ Example: add from an argument string:
...
+
 			default:
 				return fmt.Errorf("schema cannot be empty")
 			}
 
-			cols, err := store.AddSchema(cmd.Context(), schema)
+			// Process the combined schema
+			cols, err := store.AddSchema(cmd.Context(), combinedSchema)
 			if err != nil {
+				return fmt.Errorf("failed to add schema: %w", err)
```
shahzadlone, line 80: todo: please move the errors to the `cli/errors.go` file.

docs/website/references/cli/defradb_client_schema_add.md
```diff
@@ -29,8 +32,8 @@ defradb client schema add [schema] [flags]
 ### Options
 
 ```
-  -f, --file string   File to load a schema from
-  -h, --help          help for add
+  -f, --file strings   File(s) to load schema from
```
shahzadlone, line 35: thought: saying `strings` over `string` kind of signals to me that this is valid:

`defradb client schema add -f "schema1.graphql" "schema2.graphql"`

rather than:

 `defradb client schema add -f "schema1.graphql" -f "schema2.graphql"`
 
 non-blocking: because maybe it might just be me

AndrewSisley: I prefer your suggestion, but I have a vague memory of the team having this discussion and deciding on the multiple `-f`s. I hope I am wrong though :)

ChrisBQu: I have an opinion on this, but not a strong one. Happy to go with the team consensus (or majority opinion.)

ChrisBQu: Oh, I see where you are coming from now. Hmm... I think I agree. I will be changing that to the singular as you suggest.

AndrewSisley (approved):
LGTM, thanks Chris!

Approving now, on condition that:
1. Shahzad is happy with the `f` flag docs
2. You have manually tested this with multiple SDLs

cli/errors.go
```diff
@@ -65,3 +67,15 @@ func NewErrInvalidAscensionOrder(fieldName string) error {
...
+	return errors.Wrap(fmt.Sprintf("failed to read file %s", schemaFile), inner)
+}
+
+func NewErrFailedToReadSchemaFromStdin(inner error) error {
+	return errors.Wrap("failed to read schema from stdin", inner)
+}
+
+func NewErrFailedToAddSchema(inner error) error {
+	return errors.Wrap("failed to add schema", inner)
+}
```
shahzadlone, line 81: praise: Thanks for doing this suggestion :)

shahzadlone (approved):
Thanks for taking care of this, don't forget to change to `string` as I still see it as `strings` still. 

Also please do make an issue to do a similar change for other file arguments in other cli commands (like Andy suggested) if they won't be done in this PR.

shahzadlone:
thought/question:

@ChrisBQu 

Looking at the way`StringSliceVarP` is used, and finding out that the change from `string` to `strings` was an autogenerated thing.

I have a hunch that both the `-f` with list of strings and multiple `-f`s will both work?

I haven't tested, but I suspect that:

`defradb client schema add -f="schema1.graphql,schema2.graphql" -f="schema3.graphql"`

Will result in all 3 files being loaded?

Meaning that this:

```sh
defradb client schema add -f="schema1.graphql,schema2.graphql"
```

Also works? Have you tested this? (so `1` and `2` both work in a way from my previous comment?)

Sorry for the bother if this assumption of mine is incorrect.

Got the hunch after reading this: https://github.com/spf13/pflag/blob/v1.0.5/string_slice.go#L120

ChrisBQu:
> thought/question:
> 
> @ChrisBQu
> 
> Looking at the way`StringSliceVarP` is used, and finding out that the change from `string` to `strings` was an autogenerated thing.
> 
> I have a hunch that both the `-f` with list of strings and multiple `-f`s will both work?
> 
> I haven't tested, but I suspect that:
> 
> ```shell
> defradb client schema add -f="schema1.graphql,schema2.graphql" -f="schema3.graphql"
> ```
> 
> Will result in all 3 files being loaded?
> 
> Meaning that this:
> 
> ```shell
> defradb client schema add -f="schema1.graphql,schema2.graphql"
> ```
> 
> Also works? Have you tested this? (so `1` and `2` both work in a way from my previous comment?)
> 
> Sorry for the bother if this assumption of mine is incorrect.
> 
> Got the hunch after reading this: https://github.com/spf13/pflag/blob/v1.0.5/string_slice.go#L120

This is correct. I did test it. In other words: separating by a comma on a single -f flag does work.

And if both work, maybe the plural of strings is desired(?)

shahzadlone:
> This is correct. I did test it. In other words: separating by a comma on a single -f flag does work.
> 
> And if both work, maybe the plural of strings is desired(?)

Ok perfect. Then in that case please add an example of this case in the cli command documentation demonstrating both these cases. And `strings` makes sense then.

shahzadlone:
Bug bash result: Performs as expected

PR #3357, opened 2025-01-07
chore: Review HTTP support codes

### LINKED ISSUE #3139: Review http status codes
We often use `BadRequest` (400) status codes in our HTTP handlers when the status should in fact be `InternalServerError` (500).

`BadRequest` is for errors that are due to user input.
`InternalServerError` is for unexpected server behaviour.

```go
// This is a good example of `BadRequest`
txID, err := strconv.ParseUint(chi.URLParam(req, "id"), 10, 64)
if err != nil {
	return
}
```
```go
// This is a bad example of `BadRequest`. Should in fact be `InternalServerError`
dsTxn, ok := txVal.(datastore.Txn)
if !ok {
	responseJSON(rw, http.StatusBadRequest, errorResponse{ErrInvalidDataStoreTransaction})
	return
}
```

ChrisBQu:
## Relevant issue(s)

Resolves #3139

## Description

Across the code base we sometimes use `http.StatusBadRequest` (400) where we should be using `http.StatusInternalServerError` (500). I have gone through the code and made some of the changes that made logical sense to me.

## Tasks

## How has this been tested?

Testing is in progress.

- Windows

### DIFF
--- a/http/handler_collection.go
+++ b/http/handler_collection.go
@@ -70,7 +70,7 @@ func (s *collectionHandler) Create(rw http.ResponseWriter, req *http.Request) {
 		}
 
 		if err := col.CreateMany(ctx, docList); err != nil {
-			responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+			responseJSON(rw, http.StatusInternalServerError, errorResponse{err})
 			return
 		}
 		rw.WriteHeader(http.StatusOK)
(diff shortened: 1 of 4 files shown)


AndrewSisley:
question: Why is this a draft PR?

AndrewSisley:
> I forgot to change it! I try to always create my PRs as drafts, then adjust them when I'm ready. I just forgot here.

Ah okay :grin:

AndrewSisley:
thought: I'm not sure these are all correct - many of these errors returned could be user input errors, in which case `BadRequest` would be appropriate no?

PR #3363, opened 2025-01-08
feat: Purge ACP state on DB purge

### LINKED ISSUE #3358: ACP purge state
Clear the existing ACP state when running the `client purge` command

jsimnz:
## Relevant issue(s)

Resolves #3358 

## Description
***Crude*** implementation to purge the ACP state for LocalACP implementation (SourceHub ACP returns error at the moment). Basically it deletes the entire underlying KVStore and reinitializes it.

## Tasks

## How has this been tested?
Manually, also added unit tests. No current integration purge tests (afaik)

- Arch Linux
- Debian Linux
- MacOS
- Windows

### DIFF
--- a/acp/acp.go
+++ b/acp/acp.go
@@ -43,6 +43,9 @@ type ACP interface {
 	// Close closes the resources in use by acp.
 	Close() error
 
+	// DropAll purges the entire ACP state.
+	DropAll(ctx context.Context) error
+
 	// AddPolicy attempts to add the given policy. Detects the format of the policy automatically
 	// by assuming YAML format if JSON validation fails. Upon success a policyID is returned,
 	// otherwise returns error.
--- a/acp/acp.go
+++ b/acp/acp.go
@@ -44,6 +44,7 @@ type ACP interface {
 	Close() error
 
 	// ResetState purges the entire ACP state.
+	// Resetting will close the ACP engine, purge the state, then restart it
 	ResetState(ctx context.Context) error
 
 	// AddPolicy attempts to add the given policy. Detects the format of the policy automatically
--- a/acp/acp.go
+++ b/acp/acp.go
@@ -43,8 +43,8 @@ type ACP interface {
 	// Close closes the resources in use by acp.
 	Close() error
 
-	// DropAll purges the entire ACP state.
-	DropAll(ctx context.Context) error
+	// ResetState purges the entire ACP state.
+	ResetState(ctx context.Context) error
 
 	// AddPolicy attempts to add the given policy. Detects the format of the policy automatically
 	// by assuming YAML format if JSON validation fails. Upon success a policyID is returned,
--- a/acp/acp_local.go
+++ b/acp/acp_local.go
@@ -12,8 +12,7 @@ package acp
 
 import (
 	"context"
-	"errors"
-	"fmt"
+
 	"os"
 
 	protoTypes "github.com/cosmos/gogoproto/types"
@@ -24,6 +23,7 @@ import (
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/errors"
 )
 
 const localACPStoreName = "local_acp"
@@ -33,6 +33,7 @@ type ACPLocal struct {
 	pathToStore immutable.Option[string]
 	engine      types.ACPEngineServer
 	manager     runtime.RuntimeManager
+	closed      bool
 }
 
 var _ sourceHubClient = (*ACPLocal)(nil)
@@ -84,6 +85,7 @@ func (l *ACPLocal) Start(ctx context.Context) error {
 	var opts []runtime.Opt
 	var storeLocation string
 
+	l.closed = false
 	if !l.pathToStore.HasValue() { // Use a non-persistent, i.e. in memory store.
 		storeLocation = "in-memory"
 		opts = append(opts, runtime.WithMemKV())
@@ -104,10 +106,14 @@ func (l *ACPLocal) Start(ctx context.Context) error {
 }
 
 func (l *ACPLocal) Close() error {
-	return l.manager.Terminate()
+	if !l.closed {
+		l.closed = true
+		return l.manager.Terminate()
+	}
+	return nil
 }
 
-func (l *ACPLocal) DropAll(ctx context.Context) error {
+func (l *ACPLocal) ResetState(ctx context.Context) error {
 	err := l.Close()
 	if err != nil {
 		return err
@@ -119,20 +125,20 @@ func (l *ACPLocal) DropAll(ctx context.Context) error {
 		path := storeLocation + "/" + localACPStoreName
 		info, err := os.Stat(path)
 		if os.IsNotExist(err) {
-			return fmt.Errorf("drop all: path to store does not exist")
+			return errors.Join(ErrACPResetState, err)
 		} else if err != nil {
-			return fmt.Errorf("drop all: os stat: %w", err)
+			return errors.Join(ErrACPResetState, err)
 		}
 
 		if info.IsDir() {
 			// remove dir
 			if err := os.RemoveAll(path); err != nil {
-				return fmt.Errorf("drop all: delete directory: %w", err)
+				return errors.Join(ErrACPResetState, err)
 			}
 		} else {
 			// remove file
 			if err := os.Remove(path); err != nil {
-				return fmt.Errorf("drop all: delete file: %w", err)
+				return errors.Join(ErrACPResetState, err)
 			}
 		}
 	}
--- a/acp/acp_local.go
+++ b/acp/acp_local.go
@@ -13,6 +13,8 @@ package acp
 import (
 	"context"
 	"errors"
+	"fmt"
+	"os"
 
 	protoTypes "github.com/cosmos/gogoproto/types"
 	"github.com/sourcenetwork/acp_core/pkg/auth"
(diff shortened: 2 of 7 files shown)


acp/acp.go
```diff
@@ -43,6 +43,9 @@ type ACP interface {
 	// Close closes the resources in use by acp.
 	Close() error
 
+	// DropAll purges the entire ACP state.
+	DropAll(ctx context.Context) error
```
shahzadlone, line 47: suggestion: As mentioned on discord :P prefer `ResetState` over `DropAll`

```suggestion
	// ResetState resets the entire ACP state.
	ResetState(ctx context.Context) error
```

jsimnz: Would `Reset` work, or do you prefer the full `ResetState`?

acp/acp_local.go
```diff
@@ -105,6 +107,40 @@ func (l *ACPLocal) Close() error {
 	return l.manager.Terminate()
 }
 
+func (l *ACPLocal) DropAll(ctx context.Context) error {
+	err := l.Close()
```
shahzadlone, line 111: question: `l.manager.Terminate()` is safe to call close on even if it was already closed before?

acp/acp_local.go
```diff
@@ -105,6 +107,40 @@ func (l *ACPLocal) Close() error {
...
+		return err
+	}
+
+	// delete state (applicable to persistent store)
+	if l.pathToStore.HasValue() {
+		storeLocation := l.pathToStore.Value()
+		path := storeLocation + "/" + localACPStoreName
+		info, err := os.Stat(path)
+		if os.IsNotExist(err) {
+			return fmt.Errorf("drop all: path to store does not exist")
```
fredcarle, line 122: suggestion: Errors should be added to `errors.go` instead of defined inline.

acp/acp.go
shahzadlone: Reset is fine too

acp/acp_local.go
jsimnz: Thanks, forgot to do that. Meant to come back to it 😔

shahzadlone: Also rather that `drop all: xyz` error would prefer something that indicates it's an acp reset error.

acp/acp_local_test.go
```diff
@@ -128,6 +128,91 @@ func Test_LocalACP_InMemory_AddPolicy_CreatingSamePolicyAfterWipeReturnsSameID(t
...
+	require.Nil(t, errAddPolicy)
+	require.Equal(
+		t,
+		validPolicyID,
+		policyID,
+	)
+
+	errClose := localACP.Close()
+	require.Nil(t, errClose)
+}
```
shahzadlone, line 214: praise: Thanks for adding these!

fredcarle (approved):
LGTM.

node/node.go
```diff
@@ -240,5 +242,21 @@ func (n *Node) PurgeAndRestart(ctx context.Context) error {
...
+			// implement the DropAll.
+			log.ErrorE("Failed to drop ACP state", err)
+		}
+		// follow up close call on ACP is required since the node.Start function starts
+		// ACP again anyways so we need to gracefully close before starting again
+		err = acp.Close()
+		if err != nil {
+			return err
+		}
+	}
```
shahzadlone, line 259: question: Is this why you needed to introduce the `acp immutable.Option[acp.ACP]` variable on the `Node struct`?

todo: If so, why not just operate on the acp instance that is already on `n.DB` (just introduce a method on the `client.DB` interface like the existing other acp methods. OR just add this logic to `n.DB.Close()` which is called by `Node.Close()` which is called in the function above anyway.

acp/acp_local_test.go
```diff
@@ -128,6 +128,91 @@ func Test_LocalACP_InMemory_AddPolicy_CreatingSamePolicyAfterWipeReturnsSameID(t
...
+	)
+	require.Nil(t, errAddPolicy)
+
+	require.Equal(
+		t,
+		validPolicyID,
+		policyID,
+	)
+
+	errDrop := localACP.DropAll(ctx)
```
shahzadlone, line 195: nitpick: Might be interesting to also test if calling this when acp is already closed (not started) is safe or errors/panics.

acp/acp_local.go
jsimnz: Yea there might be an issue here, thanks for flagging

node/node.go
shahzadlone: Happy with reasoning will change my todo to non-blocking. Giving lgtm

shahzadlone (approved):
While I still don't fancy the acp var on the node :P it is a minimal change.
Other nitpick highlighted is the renaming to `Reset`  everywhere that "drop all" is used.

Rest all good :) cheers for sorting quickly

acp/acp_local_test.go
jsimnz: Thinking this through, should we only allow dropping if it isn't closed yet?

shahzadlone: Since we will soon have option to turn acp off (which can be done only by authorized admin), as long as we also plan to (which I believe we do) to gate the purge command also with authorized admin (AAC will use it's own separate local acp logic not mixed with this document access control logic), so it makes sense that the purge resets even if it's closed.

However happy to discuss further. We could also have a --force flag with purge that does the behavior of purging on close and otherwise a "safer" once that will error if acp is off (or is closed for whatever reason).

fredcarle: Let's not forget that the purge command is a dev only feature so protecting it with Admin ACP doesn't really make sense.

shahzadlone: > Let's not forget that the purge command is a dev only feature so protecting it with Admin ACP doesn't really make sense.

Kind of forgot about that. Do we still want dev only features when aac is implemented?

fredcarle: I think so yes. I would never want purge to be accessible by anyone on a production DB.

jsimnz:
@shahzadlone 
>While I still don't fancy the acp var on the node :P it is a minimal change.

Agreed, in the spirit of the velocity experiment, Im going to power through and mark this as a follow up issue, that we can get some more feedback on how we might want to make this change more permanent, but for now its a minor change that can easily be fixed.

acp/acp_local.go
jsimnz: Resolved this, can you do another quick review on the last 2 commits. Added a flag to track if we're already closed so we don't duplicate the termination stuff. @shahzadlone

acp/acp_local.go
```diff
@@ -104,10 +106,14 @@ func (l *ACPLocal) Start(ctx context.Context) error {
 }
 
 func (l *ACPLocal) Close() error {
-	return l.manager.Terminate()
+	if !l.closed {
+		l.closed = true
+		return l.manager.Terminate()
```
shahzadlone, line 111: suggestion: Should probably `l.manager.Terminate()` first. Then if no error occurs do `l.closed = true`?

acp/acp_local.go
shahzadlone: Re-reviewed. Just left one suggestion

node/node.go
```diff
@@ -240,5 +242,21 @@ func (n *Node) PurgeAndRestart(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	if n.acp.HasValue() {
+		acp := n.acp.Value()
+		err := acp.ResetState(ctx)
+		if err != nil {
+			// for now we will just log this error, since SourceHub ACP doesn't yet
+			// implement the DropAll.
```
shahzadlone, line 250: todo: Remove mention of `DropAll`

acp/acp_local_test.go
```diff
@@ -213,6 +213,39 @@ func Test_LocalACP_Persistent_AddPolicy_CreatingSamePolicyAfterResetStateReturns
...
+	errStart := localACP.Start(ctx)
+	require.Nil(t, errStart)
+
+	errClose := localACP.Close()
+	require.NoError(t, errClose)
+
+	errReset := localACP.ResetState(ctx)
+	require.NoError(t, errReset)
+}
+
```
shahzadlone, line 248: prase: thanks

acp/acp_local_test.go
```diff
@@ -128,6 +128,124 @@ func Test_LocalACP_InMemory_AddPolicy_CreatingSamePolicyAfterWipeReturnsSameID(t
...
+	)
+	require.Nil(t, errAddPolicy)
+
+	require.Equal(
+		t,
+		validPolicyID,
+		policyID,
+	)
+
+	errDrop := localACP.ResetState(ctx)
```
shahzadlone, line 152: todo: Rename the `drop` terminology

```suggestion
	errReset := localACP.ResetState(ctx)
```

acp/acp_local_test.go
```diff
@@ -128,6 +128,124 @@ func Test_LocalACP_InMemory_AddPolicy_CreatingSamePolicyAfterWipeReturnsSameID(t
...
```
shahzadlone, line 195: todo: Rename the `drop` terminology

```suggestion
	errReset := localACP.ResetState(ctx)
```

node/node.go
```diff
@@ -240,5 +242,21 @@ func (n *Node) PurgeAndRestart(ctx context.Context) error {
...
+			log.ErrorE("Failed to drop ACP state", err)
```
shahzadlone, line 251: todo:

```suggestion
			log.ErrorE("Failed to reset ACP state", err)
```

shahzadlone:
Bug bash: works

PR #3367, opened 2025-01-10
fix: Maintain indexes across schema versions

### LINKED ISSUE #3365: Indexes disappear after patching schema

AndrewSisley:
## Relevant issue(s)

Resolves #3365

## Description

Completes the fix started in https://github.com/sourcenetwork/defradb/pull/3366 - I missed a test case there and only tested with docs created after the patch.  This adds a new test, testing with docs created before the patch, and then updates all index related references from col.ID to col.RootID.

Sorry about the bother.

### DIFF
--- a/internal/db/collection_index.go
+++ b/internal/db/collection_index.go
@@ -414,7 +414,7 @@ func (c *collection) dropIndex(ctx context.Context, indexName string) error {
 			break
 		}
 	}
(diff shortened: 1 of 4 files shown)


shahzadlone (approved):
Good catch! LGTM

nasdf (approved):
LGTM

PR #3368, opened 2025-01-13
feat: Add special encoding of json array elements

### LINKED ISSUE #3329: Sec. index: make json array elements be encoded with depth information
Depth information for array elements allow to be more specific about which items you want to deal with.

At the moment we store all array element values as if they were just normal scalar fields on a json doc.

Lets say we have this schema:
```graphql
type User {
  name: String 
  custom: JSON @index
}
```
And we create this documents:
```json
{
  "name": "Andy",
  "custom": {
    "numbers": 4,
  }
}
{
  "name": "John",
  "custom": {
    "numbers": [3, 4],
  }
}
```
And then query:
```graphql
query {
  User(filter: {custom: {numbers: {_none: {_eq: 4}}}}) {
    name
  }
}
```
This will result at the moment in an empty result, but correct behaviour would be to return "Andy" document as its `number` is not an array and shouldn't be checked against `_none`.

The indexes are stored roughly as follows:
```
/1/1/custom.number/3/john-doc-id
/1/1/custom.number/4/andy-doc-id
/1/1/custom.number/4/john-doc-id
```
CocroachDB uses `arr` suffix. With it the indexes might look like:
```
/1/1/custom.number/arr/3/john-doc-id
/1/1/custom.number/arr/4/john-doc-id
/1/1/custom.number/4/andy-doc-id
```

islamaliev:
## Relevant issue(s)

Resolves #3329

## Description

Mark array elements in a special way so that it's possible to scope the search only to arrays.

Introduces `JSONPath` that helps differenciate parts of json path.

Also changes the behavior of `_none` filter operator so that it's exlude resuls if they are not arrays.


AndrewSisley:
> Mark array elements in a special way 

question: What made you go with approach instead of just checking the type of the value as/when needed?  Why did you feel like the encoding needed to change?

client/json.go
```diff
@@ -14,11 +14,102 @@ import (
...
+// Json path can be either a property of an object or an index of an element in an array.
+type JSONPathPart interface {
+	// Property returns the property name if the part is a property, and a boolean indicating if the part is a property.
+	Property() (string, bool)
+	// Index returns the index if the part is an index, and a boolean indicating if the part is an index.
+	Index() (uint64, bool)
+}
+
+type propPathPart string
+type indexPathPart uint64
```
AndrewSisley, line 33: question: Why are these different types?  Why would we want them to behave differently?

islamaliev:
> question: What made you go with approach instead of just checking the type of the value as/when needed? Why did you feel like the encoding needed to change?

when index is reading encoded keys it has how idea what's inside. It has either a prefix or an exact value. We use to encode like this (taking the example from the issue):
For these docs:
```json
{
  "name": "Andy",
  "custom": {
    "numbers": 4,
  }
}
{
  "name": "John",
  "custom": {
    "numbers": [3, 4],
  }
}
```
Old way to store indexes:
```
/1/1/custom.number/3/john-doc-id
/1/1/custom.number/4/andy-doc-id
/1/1/custom.number/4/john-doc-id
```
So when one of these is fetched there is no info if it's array or not.
New way (byte code is a bit different than present here):
```
/1/1/custom.number.0/3/john-doc-id
/1/1/custom.number.0/4/john-doc-id
/1/1/custom.number/4/andy-doc-id
```

islamaliev:
> What is the significance of 0 here?

`0` is just a random marker I chose for now. We can easily turn this into a real array index:
```
/1/1/custom.number.0/3/john-doc-id
/1/1/custom.number.1/4/john-doc-id
```
but it's not necessary for our use case for now. Later we can employ it when we need to process more sophisticated array queries.

> "numbers": {
>       0: "foo"
>     }

this is illegal json. This would be correct
```
{
"numbers": {
       "0": "foo"
   }
}
```

And `0` and `"0"` are encoded differently, so there is no collision there

client/json.go
AndrewSisley: Makes sense, thanks :)

todo: Please document this reason somewhere prominent so that readers understand why this difference exists.

shahzadlone: +1 for documenting this please.

client/json.go
```diff
@@ -14,11 +14,102 @@ import (
...
+	case int:
+		return indexPathPart(uint64(val))
+	case uint64:
+		return indexPathPart(val)
+	}
+	return nil
+}
+
+// Creates a path from mixed string/integer values
+func MakeJSONPath[T string | int | uint64](parts ...T) JSONPath {
```
AndrewSisley, line 105: question: I cannot spot where this is called, besides in unit tests, is this dead code?

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,68 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
...
+	if isArrayCondition(cond.op) {
+		subCondMap := filterVal.(map[connor.FilterKey]any)
+		for subKey, subVal := range subCondMap {
+			cond.arrOp = cond.op
+			cond.op = subKey.(*mapper.Operator).Operation
+			jsonVal, err = client.NewJSONWithPath(subVal, jsonPath)
+			if err == nil {
+				cond.val = client.NewNormalJSON(jsonVal)
+			}
+			// the sub condition is supposed to have only 1 record
```
AndrewSisley, line 608: question: Why?

internal/db/index.go
```diff
@@ -129,7 +129,11 @@ func (g *JSONFieldGenerator) Generate(value client.NormalValue, f func(client.No
 			return err
 		}
 		return f(val)
-	}, client.TraverseJSONOnlyLeaves(), client.TraverseJSONVisitArrayElements(false))
+	},
+		client.TraverseJSONOnlyLeaves(),
+		client.TraverseJSONWithArrayIndexInPath(),
+		client.TraverseJSONVisitArrayElements(false),
```
AndrewSisley, line 135: todo: Please document why these are here.

client/json.go
islamaliev: interesting... it was used before, but now it seems like it wasn't necessary.
Anyway it's a nice util method and we might make use of it later.
You think I should remove it (or move to tests)?

internal/db/fetcher/indexer_iterators.go
islamaliev: done

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,68 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
 	return result, nil
 }
 
+func determineJSONFilterCondition(cond *fieldFilterCond, filterVal any, jsonPath client.JSONPath) error {
```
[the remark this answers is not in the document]
islamaliev: done

client/json.go
```diff
@@ -14,11 +14,102 @@ import (
...
+func (p indexPathPart) Property() (string, bool) {
+	return "", false
+}
+
+func (p indexPathPart) Index() (uint64, bool) {
+	return uint64(p), true
+}
+
+// JSONPath represents a path to a JSON value in a JSON tree.
+type JSONPath []JSONPathPart
```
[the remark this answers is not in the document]
islamaliev: what do you mean "against the index?"

> This type suggests to me that the value will only exist either as part of an index, or as 'normal', but never both.

I don't quite understand how you classify/distinguish "part of an index" and "normal".

Sorry, I'm a bit confused by your questions

client/json.go
AndrewSisley: I think we should remove it (and `toJSONPathPart`), especially given as they are public here. 

Why would you want it kept in the test code?

client/json.go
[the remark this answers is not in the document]
AndrewSisley: Ah sorry Islam, I'll try and expand:

With our other indexed properties, the value is still stored in the datastore like 'normal', it is then also indexed, so for a single `User`.`Name` value it would exist in two places (key structure is *very* approximate :)):
```
/data/CollectionID/FieldID/DocID => "John"
/whereverIndexesAre/IndexID/"John" => DocID  
```

To me `type JSONPath []JSONPathPart` suggests that there can only be one path for a field-value, it will exist in the index, or it will exist 'normally' in the datastore, it will never exist in both.  Is this correct?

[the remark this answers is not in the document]
islamaliev: oh, now I see what you mean.

Yes, at the moment `JSONPath` is used only by the indexer.

For 'normal' storage records we don't process the content of json anyway and just store it as bytes blob.
```
/data/CollectionID/FieldID/DocID => '{ "custom": { "name": "John", "age": 33 } }'
/index/CollectionID/IndexID/custom.name/John/DocID => ""
/index/CollectionID/IndexID/custom.age/33/DocID => ""
```

client/json.go
islamaliev: remove from production code

client/json.go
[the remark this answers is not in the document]
AndrewSisley: > Yes, at the moment JSONPath is used only by the indexer.

I see you are correct, but the code structure very strongly suggests otherwise (`JSONPath` is returned from `client`.`Json`.`GetPath()`, which is called by funcs in shared packages like `encoding` and `keys`, before eventually reaching obviously index-specific code.

question: Why does the code structure (including public `client` stuff) suggest this is not index specific?  Is there a long-term plan to store JSON values differently, and if so, what is it?

client/json.go
```diff
@@ -14,11 +14,107 @@ import (
...
+		} else if index, ok := part.Index(); ok {
+			sb.WriteByte('[')
+			sb.WriteString(strconv.FormatUint(index, 10))
+			sb.WriteByte(']')
+		}
+	}
+	return sb.String()
+}
+
+func toJSONPathPart[T string | int | uint64](v T) JSONPathPart {
```
AndrewSisley, line 106: todo: This func is only called in tests, please move it to test code.

internal/db/index.go
AndrewSisley: Sorry, I have to reopen this.  You have documented what these do, but please document *why* we need to do these here.

client/json.go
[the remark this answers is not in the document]
islamaliev: I didn't say that `GetPath()` is index-specific. It's used only by index at the moment, but can be used by whatever else.

> Is there a long-term plan to store JSON values differently, and if so, what is it?

Not that I know of.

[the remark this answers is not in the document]
AndrewSisley: If we have no plans for it to be consumed by non-index code, why are we suggesting to readers, internal and external, that it is relevant to non-index code?

For users this will potentially lead to miss-use, and/or avoidable confusion.

For maintainers this will lead to effort being spent on preserving it's (unused) genericness, probably leading to more complex than required code in the future (if not now).  Both in this code, and the code around it.

internal/db/index.go
islamaliev: I though is was clear. I added some more sentences. Not sure what else to add.

client/json.go
islamaliev: done

client/json.go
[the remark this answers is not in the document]
islamaliev: well, where do we draw the line between what's needed and what's not? 

The `JSON` interface wasn't originated from external users' needs either. And it's probably the same for `NormalValue` interface. They were all mostly driven by the indexer and encoder needs. 

It seems like we non-verbally agreed on having doc-field-value-related stuff to be in `client` package.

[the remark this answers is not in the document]
AndrewSisley: > where do we draw the line between what's needed and what's not?

The most knowledgeable member of Source regarding the index code and author of the original code noted in this conversation that they know of no plans to use this for non-index stuff.

`NormalValue` is the type returned by `Document`, regardless as to whether the property was indexed or not.  It is the primary way in which embedded users will receive field values.

`Json` is the type of value returned when a `NormalValue` is unwrapped (e.g. via `Json()`), regardless of indexing.

`Json`.`GetPath` and `JSONPath` exist within the space, and appear to be general purpose to users and maintainers, yet have no use outside of the index system.  Maintaining it as such, despite having no plans to ever user it outside of the index system is a poor use of our time IMO.  As mentioned, I also don't think it benefits our users.

internal/db/index.go
```diff
@@ -129,7 +129,17 @@ func (g *JSONFieldGenerator) Generate(value client.NormalValue, f func(client.No
 			return err
 		}
 		return f(val)
-	}, client.TraverseJSONOnlyLeaves(), client.TraverseJSONVisitArrayElements(false))
+	},
+		// we don't want to traverse intermediate nodes, because we encode only values that can be filtered on
+		client.TraverseJSONOnlyLeaves(),
```
AndrewSisley, line 134: question: Why are these functions declared in the `client` package?  Do you think users will want to use them, and if so, for what?

internal/db/index.go
AndrewSisley: Thanks Islam!

tests/integration/index/json_array_test.go
```diff
@@ -140,7 +140,7 @@ func TestJSONArrayIndex_WithNestedArrays_ShouldNotConsiderThem(t *testing.T) {
 				DocMap: map[string]any{
 					"name": "Islam",
 					"custom": map[string]any{
-						"numbers": []any{0, []int{2, 6}, 9},
+						"numbers": []any{0, []int{2}, 4},
```
AndrewSisley, line 143: praise: I was looking for this test, thanks Islam :)

tests/integration/query/json/with_none_test.go
```diff
@@ -107,9 +107,6 @@ func TestQueryJSON_WithNoneFilterAndNestedArray_ShouldFilter(t *testing.T) {
 				Results: map[string]any{
 					"Users": []map[string]any{
 						{"name": "Shahzad"},
-						{"name": "John"},
-						{"name": "Islam"},
-						{"name": "Bruno"},
```
AndrewSisley, line 112: thought:  This looks rather odd to me, although I was on the minority side in the standup discussion around this and got a little lost along the way.

@sourcenetwork/database-team Is this the behaviour we want?

internal/db/index.go
islamaliev: it doesn't have to be in `client` package. It can be easily moved elsewhere? Do you have an idea where it would fit best?

AndrewSisley: They are only ever called in `internal/db/index.go`, so my preference would be to make them private funcs in that file.  But the most important thing for me is to move them out of the users' scope and to somewhere internal.

client/json.go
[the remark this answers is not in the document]
islamaliev: if we want to move `GetPath` elsewhere we will have to do some type juggling. For example we will have `JSON` interface in `client` package but without `GetPath` method. Have in some `internal` package:
```go
JSONPath []JSONPartPath

type JSONWithPath interface {
    client.JSON
    GetPath() JSONPath
}
```
in `client`:
```go
type jsonNumber struct {
   // ...
}

// var _ internal.JSONWithPath = (*jsonNumber)(nil)
// we can not use such interface enforcement here 

func (j jsonNumber)GetPath() JSONPath { ... }
```
Then somewhere in encodered:
```
func EncodeFieldValue(b []byte, val client.NormalValue, descending bool) []byte {
	// ...
	if v, ok := val.JSON(); ok {
		if descending {
			return EncodeJSONDescending(b, v)
		}
		return EncodeJSONAscending(b, v)
	}

	return b
}

func EncodeJSONAscending(b []byte, v json) {
    if j, ok := v.(internal.JSONWithPath); ok {
        encodeJSONPath(j.GetPath())
    }
}
```
Is this what you would prefer?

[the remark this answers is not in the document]
AndrewSisley: Roughly, although it doesn't look like `GetPath` needs to be hosted on an object at all, and could just be a global/package func in the encoding package?
```
func getPath(j client.JSON) jsonPath {
  ...
}
```
Or maybe deleted entirely, and just do it directly in `encodeJSONPath` instead of worrying about a fancy type system.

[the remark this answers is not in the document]
islamaliev: how do you propose to extract the path from `client.JSON` if it doesn't have `GetPath` public method?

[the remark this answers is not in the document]
AndrewSisley: I think I misunderstood your suggestion :)

And I think you might need to a fair bit of code restructuring to remove all the bits I consider undesirable from the `client` package.

In the short term you could probably revert `GetPath` to return a string array, and the encode it in the encoding package - removing the new stuff from the `client` package, but yes I think without reworking some other stuff it would be hard to remove `GetPath` (or a variant of it).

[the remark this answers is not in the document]
islamaliev: `client.JSON` represents not only a complete json tree, but also a json subtree. What makes a subtree a subtree is presence of a path. You can not pass around a subtree without having it carry the path.
```json
{
  "custom": {
    "name": "John"
  }
}
```
So `{ "name": "John" }` in this doc is subtree with path "custom".
You can not have some arbitrary `getPath` method that receives a subtree that doesn't carry path information and magically extract the path from it.

tests/integration/query/json/with_none_test.go
AndrewSisley: Discussed in standup, this is what the team wants :)

client/json.go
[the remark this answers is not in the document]
AndrewSisley: I disagree that hosting this code in user-space is the only way to solve this, but lets park this conversation and mark this thread as resolved.   Whilst removing these types from the client package is a breaking change, it can be done later if we want to.

Thank you for your answers, I understand the changes better :)

client/json.go
```diff
@@ -14,11 +14,95 @@ import (
...
+//		"val": 1
+//	  }
+//	}
+//	[
+//	  {
+//	    "val": 1
+//	  }
+//	]
+//
+// It can be described as "0.val" but they are different.
```
AndrewSisley, line 38: suggestion: I find the wording of `It can be described as "0.val" but they are different.`  a little confusing - could you please tweak it?  Maybe note how they are different - the `0` magic value is not described here and is not obvious to readers as to why `0.val` is relevant to the second example - they might guess incorrectly that `0` is the array index of the first value, yet IIRC you noted that all elements in an array would be `0` in the path.

EDIT: Some of the code further down the PR looks like it is now the element index, and that the hard coding has been removed, is this the case?  If so, please still tweak the wording, as it seems a little vague atm.

client/json.go
```diff
@@ -59,17 +143,17 @@ type JSON interface {
 	Marshal(w io.Writer) error
 
 	// GetPath returns the path of the JSON value in the JSON tree.
-	GetPath() []string
+	GetPath() JSONPath
```
AndrewSisley, line 146: todo: Please note that this is the index path only, and does not represent the path of the base (datastore-doc) data.

internal/db/fetcher/indexer_iterators.go
```diff
@@ -515,16 +519,26 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
...
 				for {
 					for key, filterVal := range condMap {
 						prop, ok := key.(*mapper.ObjectProperty)
 						if !ok {
+							// if filter contains an array condition, we need to append index 0 to the json path
+							// to limit the search only to array elements
+							op, ok := key.(*mapper.Operator)
+							if ok && isArrayCondition(op.Operation) {
+								if op.Operation == compOpNone {
+									// if the array condition is _none it doesn't make sense to use index
```
AndrewSisley, line 534: todo: Please document *why* it does not make sense to use an index for `_none`.

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,72 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
 	return result, nil
 }
 
+// determineJSONFilterCondition determines the condition and its corresponding operation for a
+// JSON filter condition.
+// It mutates the given condition to make it match the filter value and JSON path so that
+// it can be used to fetch the indexed data.
+func determineJSONFilterCondition(cond *fieldFilterCond, filterVal any, jsonPath client.JSONPath) error {
```
AndrewSisley, line 600: suggestion: I think the name of this function contributed to my confusion as to what it did - `determine` to me suggests a read-only operation, but this function is really setting the filter condition no? Would `setJSONFilterCondition` be a better name perhaps?

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,72 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
...
+	var jsonVal client.JSON
+	var err error
+	if isArrayCondition(cond.op) {
+		subCondMap := filterVal.(map[connor.FilterKey]any)
+		for subKey, subVal := range subCondMap {
+			cond.arrOp = cond.op
+			cond.op = subKey.(*mapper.Operator).Operation
+			jsonVal, err = client.NewJSONWithPath(subVal, jsonPath)
```
AndrewSisley, line 608: todo: This flow is a little confusing and looks like you are silently discarding the error.  Perhaps the following would be a little clearer?
```
jsonVal, err = client.NewJSONWithPath(subVal, jsonPath)
if err != nil {
  return err
}
cond.val = client.NewNormalJSON(jsonVal)
```

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,72 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
 	return result, nil
 }
 
+// determineJSONFilterCondition determines the condition and its corresponding operation for a
+// JSON filter condition.
+// It mutates the given condition to make it match the filter value and JSON path so that
+// it can be used to fetch the indexed data.
+func determineJSONFilterCondition(cond *fieldFilterCond, filterVal any, jsonPath client.JSONPath) error {
+	var jsonVal client.JSON
+	var err error
```
AndrewSisley, line 602: suggestion: neither `jsonVal` nor `err` are used in the scope in which they are declared, and it looks like this has caused an err to silently be ignored.  I think removing these would improve readability, and make it more obvious in the child blocks that they are really local to their own scopes.

AndrewSisley (changes requested):
Thanks for your earlier answers Islam, I have now reviewed the PR - looks good, but I think you are dropping an error, and a little more documentation would be handy.

internal/encoding/json.go
```diff
@@ -130,21 +130,35 @@ func decodeJSONPath(b []byte) ([]byte, []string, error) {
...
 func encodeJSONPath(b []byte, v client.JSON) []byte {
 	b = append(b, jsonMarker)
 	for _, part := range v.GetPath() {
-		pathBytes := unsafeConvertStringToBytes(part)
-		b = EncodeBytesAscending(b, pathBytes)
+		if prop, ok := part.Property(); ok {
+			pathBytes := unsafeConvertStringToBytes(prop)
+			b = EncodeBytesAscending(b, pathBytes)
+		} else if _, ok := part.Index(); ok {
+			b = EncodeUvarintAscending(b, 0)
```
AndrewSisley, line 160: todo: Please document the `0` here.

internal/encoding/json.go
```diff
@@ -130,21 +130,35 @@ func decodeJSONPath(b []byte) ([]byte, []string, error) {
 			b = b[1:]
 			break
 		}
-		rem, part, err := DecodeBytesAscending(b)
-		if err != nil {
-			return b, nil, NewErrInvalidJSONPath(b, err)
+
+		if PeekType(b) == Bytes {
+			rem, part, err := DecodeBytesAscending(b)
```
AndrewSisley, line 135: suggetion: I do not know what `rem` is here, please consider using another name.

internal/encoding/json.go
```diff
@@ -130,21 +130,35 @@ func decodeJSONPath(b []byte) ([]byte, []string, error) {
...
+			if err != nil {
+				return b, nil, NewErrInvalidJSONPath(b, err)
+			}
+			path = path.AppendProperty(string(part))
+			b = rem
+		} else {
```
AndrewSisley, line 141: todo: Please document why this `else` can safely assume the value is a Uvarint

internal/db/fetcher/indexer_iterators.go
```diff
@@ -594,6 +593,72 @@ func (f *IndexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
...
+			cond.arrOp = cond.op
+			cond.op = subKey.(*mapper.Operator).Operation
+			jsonVal, err = client.NewJSONWithPath(subVal, jsonPath)
+			if err == nil {
+				cond.val = client.NewNormalJSON(jsonVal)
+			}
+			// the array sub condition (_any, _all or _none) is supposed to have only 1 record
+			break
+		}
+	} else if cond.op == opIn {
```
AndrewSisley, line 615: suggestion: It is not obvious why we only need to check for `opIn` here, and not the other array ops too - I assume because the index code resolves them all down into 'in', but I am not confident and would not want to rely on that assumption when making changes in the future.  Please consider documenting the actual reason here.

internal/encoding/json.go
shahzadlone: +1

islamaliev: done

internal/encoding/json.go
islamaliev: done

internal/encoding/json.go
islamaliev: done

client/json.go
islamaliev: as I said earlier, although at the moment it's used only by the indexer, it is not index-specific, hence doesn't need to mention it.
And there no point in saying anything about storage, as it's irrelevant. It says "in the JSON tree".

I thought we agreed that TODOs are going to be used only if a reviewer considers the code (or absence of it) a blocker. I don't think any request for comment or documentation can ever be considered a blocker. To me it feels like we failed the experiment at it's very start.

client/json.go
islamaliev: change documentation

internal/db/fetcher/indexer_iterators.go
islamaliev: done

internal/db/fetcher/indexer_iterators.go
islamaliev: done

internal/db/fetcher/indexer_iterators.go
islamaliev: dont

internal/db/fetcher/indexer_iterators.go
islamaliev: done

internal/db/fetcher/indexer_iterators.go
islamaliev: I thought it was obvious. It feel like I'm writing information that has no value.

I have to ask you @AndrewSisley again why you mark this as TODO. Event without the experiment I would disagree with it's TODO status.
It feels like you are not onboard with the experiment and going the opposite direction. I'm a bit frustrated.

internal/db/index.go
islamaliev: turns out it's not that easy to move it out.

I created a [ticket](https://github.com/sourcenetwork/defradb/issues/3388).

internal/db/fetcher/indexer_iterators.go
AndrewSisley: It is obvious to you because you wrote it, and have spent considerably more time than most thinking about it.

It is a blocker, because I want code that is easy to read merged into to the codebase.  This review took far longer than it needed to because it was not easy to read, it felt like you dumped 600 lines of undocumented code on us with a 3 line PR description.  We are both frustrated.  If you want quicker reviews, document your work so that other people can read it.

The experiment is not a 'merge whatever you like whether others can work with it or not' sprint.

AndrewSisley: I spoke with Fred, he agreed with you that this was not a blocker.  He suggested I make it a suggestion, and successfully convinced me that it was suggestion worthy, and that my original comment was information-sparse :)

suggestion: This comment feels incomplete and raises questions in the reader's mind.  It would speed up reading if it described why it does not make sense to use an index for `_none`.

Perhaps you could could change it to something along the lines of:
```
// if the array condition is _none it doesn't make sense to use index as we need to do a full table scan anyway
```

fredcarle: As discussed in person, if we write documentation to make the code more readable, we should make an effort to describe `why` the code has this behaviour. In this case, the `why` is that the full table scan that would be required on the index would perform worst than the datastore full table scan so we don't use the index. It makes sense to update it in my opinion. 

A todo was maybe a little much in this case but I don't think the comment goes against the experiment as we also discussed in person.

client/json.go
fredcarle: In this case I think the current documentation is relevant and precise enough. This is not index specific and not related to datastore keys so I don't think its relevant to state that it does not represent a datastore doc path.

jsimnz: > I thought we agreed that TODOs are going to be used only if a reviewer considers the code (or absence of it) a blocker. I don't think any request for comment or documentation can ever be considered a blocker. To me it feels like we failed the experiment at it's very start.

Reviewers are free to put `todo` on whatever they *feel* warrants it, reviewers *should* be more strict with what they mark as `todo` currently. As I said, its on the author to be able to declare the `todo` out of scope (importantly, this doesnt mean its wrong or shouldnt be done, just out of scope of this PR).

For this particular instance, a `todo` requesting a comment/documentation is a very small request and would take a total of 30 second to do. Unless you feel strongly that adding this comment is blatantly wrong or counter productive (which I don't personally see how it would be in this case).

AndrewSisley (approved):
LGTM, thanks Islam, and thanks for the extra docs!

PR #3387, opened 2025-01-16
fix: Ensure no ACP doc check when action is node specific

### LINKED ISSUE #3386: `checkAccessOfDocWithACP` should not have to do an ACP lookup it the access is done by the node itself

fredcarle:
## Relevant issue(s)

Resolves #3386 

## Description

This PR resolve the issue where re-adding docs to the p2p tropics on startup resulted in checking doc access. If we know the operation is node specific, meaning it is not to be returned to a client, then we don't need to check access through the ACP engine.

## Tasks

## How has this been tested?

Manual startup test.

- MacOS

### DIFF
--- a/internal/db/p2p_replicator.go
+++ b/internal/db/p2p_replicator.go
@@ -133,10 +133,12 @@ func (db *db) SetReplicator(ctx context.Context, rep client.ReplicatorParams) er
 	}
 
 	txn.OnSuccess(func() {
+		// This is a node specific action which means the actor is the node itself.
+		ctx := identity.WithContext(context.Background(), db.nodeIdentity)
 		db.events.Publish(event.NewMessage(event.ReplicatorName, event.Replicator{
 			Info:    rep.Info,
 			Schemas: storedSchemas,
-			Docs:    db.getDocsHeads(context.Background(), addedCols),
+			Docs:    db.getDocsHeads(ctx, addedCols),
 		}))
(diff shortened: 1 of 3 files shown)


internal/db/p2p_replicator.go
```diff
@@ -156,6 +157,8 @@ func (db *db) getDocsHeads(
 		}
 		defer txn.Discard(ctx)
 		ctx = SetContextTxn(ctx, txn)
+		// This is a node specific action which means the actor is the node itself.
+		ctx = identity.WithContext(ctx, db.nodeIdentity)
```
AndrewSisley, line 161: nitpick: I don't think this function can actually guarantee that this comment is actually true, it would be better to do this in `SetReplicator` immediately after creating the ctx.  Or remove teh ctx param and create the `context.Background()` here.

AndrewSisley (approved):
LGTM, good job tracking this down.

shahzadlone (approved):
LGTM

internal/db/p2p_replicator.go
fredcarle: It currently is guaranteed as only `SetReplicator` calls it and the action of getting the doc heads here is used only for the purpose of a node specific action.

fredcarle: moved it to `SetReplicator`.

Issue #3299, opened 2024-12-09
Sec. index: make compound filters utillize sec. indexes


islamaliev:
at the moment tests like this don't use indexes:
```go
func TestJSONIndex_WithCompoundFilterCondition_ShouldUseIndex(t *testing.T) {
	req := `query {
		User(filter: {_and: [
			{custom: {height: {_eq: 180}}},
			{custom: {weight: {_eq: 80}}}
		]}) {
			name
		}
	}`
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String 
						custom: JSON @index
					}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "John",
					"custom": {"height": 168, "weight": 70}
				}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Islam",
					"custom": {"height": 180, "weight": 80}
				}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Shahzad",
					"custom": {"height": 180, "weight": 75}
				}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Keenan",
					"custom": {"height": 190, "weight": 85}
				}`,
			},
			testUtils.Request{
				Request: req,
				Results: map[string]any{
					"User": []map[string]any{
						{"name": "Islam"},
					},
				},
			},
			testUtils.Request{
				Request:  makeExplainQuery(req),
				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(2),
			},
		},
	}

	testUtils.ExecuteTestCase(t, test)
}
```
The test uses json, but the behaviour should be the same for other field types.
Most likely majority of the work will be in function `findIndexByFilteringField` in `internal/planner/select.go`. But we should add a bunch of tests as well.

PR #3390, opened 2025-01-17
perf: Remove compound condition from 1-1 validation filter

### LINKED ISSUE #3190: Querying secondary relationships slow as db size increases
Responses to queries with secondary relationships become incrementally slower relative to the size of the db. (Including unrelated collections)

Example Schema
```gql
type Dog {
  name: String
} 

type Director {
  name: String 
  movies: [Movie] @relation(name: "directorMovies")
}

type Movie {
  title: String
  director: Director @primary @relation(name: "directorMovies")
}
```
*Note: Dogs is not related to Director or Movies + Not included in this query:

```gql
query {
  Director {
   movies {
     _docID
   } 
  }
}
```

Calling the above query gives the following response times (where number of documents in the db is):
```
Db seeded with:  0 Directors with 0 movies

0 Dogs 22ms
1000 Dogs 23ms
2000 Dogs 31ms
3000 Dogs 37ms
4000 Dogs 30ms
5000 Dogs 46ms
```

```
Db seeded with: 100 Directors with 1 movie each

0 Dogs 124ms
1000 Dogs 307ms
2000 Dogs 625ms
3000 Dogs 727ms
4000 Dogs 921ms
5000 Dogs 1177ms
```

```
Db seeded with: 200 Directors with 0 movies

0 Dogs 116ms
1000 Dogs 513ms
2000 Dogs 884ms
3000 Dogs 1322ms
4000 Dogs 1788ms
5000 Dogs 2199ms
```

islamaliev:
## Relevant issue(s)

Resolves #3190

## Description

Remove compound condition from 1-to-1 validation filter.

At the moment indexes don't work if compound conditions are present. For more details read https://github.com/sourcenetwork/defradb/issues/3299

### DIFF
--- a/internal/db/collection.go
+++ b/internal/db/collection.go
@@ -779,7 +779,7 @@ func (c *collection) validateOneToOneLinkDoesntAlreadyExist(
 	}
 
 	filter := fmt.Sprintf(
-		`{_and: [{%s: {_ne: "%s"}}, {%s: {_eq: "%s"}}]}`,
+		`%s: {_ne: "%s"}, %s: {_eq: "%s"}`,
 		request.DocIDFieldName,
 		docID,
 		fieldDescription.Name,


AndrewSisley (approved):
question: To clarify, this does not solve the performance issue, we still need to actually introduce an index?  This is just a prerequisite for the to-be-added index improving the performance?

islamaliev:
> question: To clarify, this does not solve the performance issue, we still need to actually introduce an index? This is just a prerequisite for the to-be-added index improving the performance?

yes, users still have to specify `@index` in order for this work fast.

But it's not a prerequisite. This would have worked automatically after [this](https://github.com/sourcenetwork/defradb/issues/3299) is merged, which I happened to work on for a few days already.

AndrewSisley:
> yes, users still have to specify `@index` in order for this work fast.
> 
> But it's not a prerequisite. This would have worked automatically after [this](https://github.com/sourcenetwork/defradb/issues/3299) is merged, which I happened to work on for a few days already.

Ah okay got it - thanks Islam, makes sense to just ask the partner to add the index themselves for now :)

jsimnz (commented):
> yes, users still have to specify @index in order for this work fast.

@islamaliev I thought the point was to have the DB build this index automatically for 1-1, or did I misunderstand the descision from yesterdays call?

islamaliev:
> @islamaliev I thought the point was to have the DB build this index automatically for 1-1, or did I misunderstand the descision from yesterdays call?

I also thought I would automatically create index on relations, but then bunch of questions popped up:
- what if the user already specified the index?
- what if the user doesn't want this because of some storage restrictions?
- so we expose this silently created index to the user and what if they decide to remote it?
- and so on.

I spoke today with @fredcarle regarding some other matter and it turned out he expected only validation check to utilize indexes (we didn't actually discuss the questions that I had).

Obviously silent index creation is more involved and this work won't be needed anyway after we get rig of `@primary` and make indexes dictate 1-to-1 vs. 1-to-many relations.

jsimnz:
> I also thought I would automatically create index on relations, but then bunch of questions popped up:

Those are all fair questions. I believe I have good answers to them, but as pointed out we can shelve the discussion in general 👍.

PR #3400, opened 2025-01-20
fix(i): Fetcher wrapper panic

nasdf:
## Relevant issue(s)

Resolves #3399

## Description

This PR fixes an issue where the `Start` function of `wrappingFetcher` would not return an error causing a seg fault when trying to access the fetcher in a subsequent `FetchNext` call.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

No added tests. I'm unsure if this is worth testing with mocks, but open to suggestions.

Specify the platform(s) on which this was tested:
- MacOS

PR #3403, opened 2025-01-21
test(i): Handle error in test setup, don't return nil

### LINKED ISSUE #3402: Non nill error should not return nil
In light of this PR:
https://github.com/sourcenetwork/defradb/pull/3400/files

I did a rip grep search:
```
 rg -U "if err != nil \{\n.*return nil$"
```

And found one more instance in the code base where not nill error was found and still returning `nil`.

shahzadlone:
## Relevant issue(s)
Resolves #3402 

## Description
In light of this PR:
https://github.com/sourcenetwork/defradb/pull/3400/files

I did a rip grep search:
```
 rg -U "if err != nil \{\n.*return nil$"
```

And found one more instance in the code base where non-nill error would still returning `nil`.

### DIFF
--- a/tests/integration/db.go
+++ b/tests/integration/db.go
@@ -111,7 +111,7 @@ func NewBadgerFileDB(ctx context.Context, t testing.TB) (client.DB, error) {
 	return node.DB, err
 }
 
-func getDefaultNodeOpts() []node.Option {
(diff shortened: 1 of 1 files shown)


AndrewSisley (approved):
LGTM, thanks for checking this Shahzad :)

PR #3404, opened 2025-01-23
feat: Add open telemetry metrics and traces

### LINKED ISSUE #74: Tracing system

### LINKED ISSUE #293: [Epic] Add Observability (metrics)
Using a combination of prometheus/statsd/etc

Add various metric tracking to the database node server

Initial metric ideas:
- CPU usage
- Memory usage
- Total storage (bytes)
- Total documents
- Total collections
- Size per collection (bytes)
- Per Collection
  - size (bytes)
  - total documents
  - total queries
  - queries per second
  - full scans
  - index scans
  - transactions
  - query response time (latency, histogram, 90th, 99th, etc)
  - size of query
- KV
  - kv ops
- ... more to come

Just an initial (non-exhaustive) list of possible stats

nasdf:
## Relevant issue(s)

Resolves #293
Resolves #74

## Description

This PR adds OpenTelemetry metrics and tracing. Telemetry is only enabled if the `telemetry` build flag is set.

Our default telemetry configuration uses the `http` exporter, but we can expand it to `grpc` and `console` if needed.

The exporters are configured through the standardized OpenTelemetry environment variables found here:

https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp
https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp

## Tasks

## How has this been tested?

Manually testing with Jaeger

```
docker run --rm --name jaeger \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  -p 5778:5778 \
  -p 9411:9411 \
  jaegertracing/jaeger:2.2.0
```

```
DEFRA_KEYRING_SECRET=secret \
 OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 \
 go run -tags telemetry ./cmd/defradb start
```

> Trace dashboard
Jaeger UI

> Detailed trace example
_ 3240f64_ GetCollections _DefraDB_ _ Jaeger UI

> Nested trace example
_ b146bd9_ ExecRequest _DefraDB_ _ Jaeger UI

- MacOS


internal/db/db.go
```diff
@@ -39,7 +39,8 @@ import (
 )
 
 var (
-	log = corelog.NewLogger("db")
+	log    = corelog.NewLogger("db")
+	tracer = otel.Tracer("github.com/sourcenetwork/defradb/internal/db")
```
AndrewSisley, line 43: thought: I think I agree with John's suggestion of extracting this out to a thin package so we can avoid scattering references throughout the code base to the same 3rd party package.  

If we ever swap it out we'll probably still need to change the func signatures (I wouldn't put much thought into designing them), but at least everything will be in the same place, and it makes adding/standardising any middleware much easier.

internal/db/db.go
```diff
@@ -212,6 +213,9 @@ func (db *DB) AddPolicy(
 	ctx context.Context,
 	policy string,
 ) (client.AddPolicyResult, error) {
+	ctx, span := tracer.Start(ctx, "AddPolicy")
```
AndrewSisley, line 216: suggestion: I do not know about performance/C-GO, but we can probably avoid having to manually specify `"AddPolicy"` by using the runtime package and stuff like [runtime.CallersFrames](https://golang.org/pkg/runtime/#CallersFrames) instead.

nasdf: I can look into that.

internal/db/db.go
nasdf: Update: the metrics were all moved to the internal `metric` package and put behind a build flag `telemetry`.

internal/db/db.go
nasdf: Update: I was able to get this working. Thanks for the suggestion!

AndrewSisley (changes requested):
praise: This looks really good considering my natural aversion to it, good job at minimising it's impact on readability.

I have not really reviewed *where* you have added the tracer (i.e. all the `tracer.Start` calls), as I would prefer to not have any by default and instead add them in when we really feel the need to (like debug logs) - I will let someone else review those.

todo: Please add a job to the CI test matrix so that we have one job that executes the tests with the tracer enabled.

internal/request/graphql/parser.go
```diff
@@ -46,7 +49,10 @@ func NewParser() (*parser, error) {
 	return p, nil
 }
 
-func (p *parser) BuildRequestAST(request string) (*ast.Document, error) {
+func (p *parser) BuildRequestAST(ctx context.Context, request string) (*ast.Document, error) {
+	_, span := tracer.Start(ctx)
```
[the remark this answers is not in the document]
nasdf: Yes, it was a linter warning because the context is only used for creating the span. The reason for adding the context to the function signature is that it allows the creation of nested spans by using context values.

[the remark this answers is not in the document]
AndrewSisley: Ah okay, thanks for the answer.

I *think* I have a slight preference to assigning it with a //nolint:foo comment as that would be less risky in case we later use `ctx` in this function, but it is a bit ugly so please don't feel any pressure to change it if you aren't certain in your agreement :)

cli/start.go
```diff
@@ -149,6 +150,13 @@ func MakeStartCommand() *cobra.Command {
 				}
 			}
 
+			if !cfg.GetBool("no-telemetry") {
+				err := metric.ConfigureTelemetry(cmd.Context())
+				if err != nil {
+					log.ErrorE("failed to configure telemetry", err)
```
shahzadlone, line 156: thought: Perhaps define in `cli/errors.go`

(thought because its a log).

internal/metric/metric_test.go
shahzadlone, line 1: question: We didn't need `count` and `hist` metric intrumentation?

internal/metric/noop.go
```diff
@@ -0,0 +1,40 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+//go:build !telemetry
```
shahzadlone, line 11: question: What does this do?

tests/integration/utils.go
```diff
@@ -2331,22 +2330,6 @@ func skipIfNetworkTest(t testing.TB, actions []any) {
...
-	if err != nil {
-		return nil, err
-	}
-	result := make(map[string]client.CollectionDefinition)
-	for _, col := range cols {
-		result[col.Description.Name.Value()] = col
-	}
-	return result, nil
-}
-
```
shahzadlone, line 2349: question: What happened here? was this dead code?

shahzadlone (approved):
Good stuff, just left some comments.


Please do add the CI test with the tracer enabled (doesn't have to be a matrix) just one workflow is good. You can do something like the `test-view` or `test-encryption` job in `.github/workflows/test-coverage.yml`.

internal/metric/noop.go
fredcarle: build will use this file if the `telemetry` build flag isn't used. So:
- noop.go will be used if no `telemetry` build flag.
- otel.go will be used if `telemetry` flag.

internal/metric/metric_test.go
nasdf: I'm planning on incrementally adding metrics as we need or want them. I'd like to keep the footprint as small as possible so that the telemetry doesn't get in the way of feature development.

tests/integration/utils.go
nasdf: I changed the `Parser` interface and when I went to fix the function signature there I noticed it was dead code.

docs/website/references/cli/defradb_start.md
```diff
@@ -21,6 +21,7 @@ defradb start [flags]
       --max-txn-retries int           Specify the maximum number of retries per transaction (default 5)
       --no-encryption                 Skip generating an encryption key. Encryption at rest will be disabled. WARNING: This cannot be undone.
       --no-p2p                        Disable the peer-to-peer network synchronization system
+      --no-telemetry telemetry        Disables telemetry reporting. Telemetry is only enabled in builds that use the telemetry flag.
```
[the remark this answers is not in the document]
nasdf: good catch! should be fixed now

.github/workflows/test-coverage.yml
```diff
@@ -236,6 +236,27 @@ jobs:
...
+      - name: Checkout code into the directory
+        uses: actions/checkout@v4
+
+      - name: Setup defradb
+        uses: ./.github/composites/setup-defradb
+
+      - name: Test coverage & save coverage report in an artifact
+        uses: ./.github/composites/test-coverage-with-artifact
+        with:
+          coverage-artifact-name: "coverage_encryption"
```
shahzadlone, line 258: issue: This typo is very problematic as it will fail all reports due to duplicate artifact name (or overwrite which was the old behavior). Please fix the name to something relevant.

.github/workflows/test-coverage.yml
```diff
@@ -236,6 +236,27 @@ jobs:
...
           coverage-path: coverage.txt
 
+  # This job runs the database with telemetry tests using default configuration, on linux.
+  test-telemetry:
+    name: Test telemetry job
+
+    runs-on: ubuntu-latest
+
+    env:
+      GO_FLAGS: -tags=telemetry
```
shahzadlone, line 246: thought: This is a neat way to pass go flags using env variables. Do you think this is more conventional that what we do with other testing types like `DEFRA_CLIENT_GO` for example. IMO if we choose this over `DEFRA_TELEMETRY` might be worth making other flags consistent 

question: What is the `tags` flag?

internal/metric/metric_test.go
shahzadlone: That's fine and I agree with. Just wondering why not make use of the the metric and histogram code that exists and has tests for it already? wouldn't you need to add it again soon anyway?

.github/workflows/test-coverage.yml
nasdf: nice catch! should be fixed now

.github/workflows/test-coverage.yml
shahzadlone: Ah! never mind thanks for clarifying

shahzadlone, line 246: question: is `GO_FLAGS=-tags=telemetry` and `GOFLAGS=-tags=telemetry` the same?

nasdf: Turns out the CI was just flaky. Coverage now shows that it is working.

shahzadlone: It wasn't a suggestion, but a question haha because I only saw that flag in documentation and not the one with underscore. Cool so it works eh?

shahzadlone: The configure telemetry function in otel.go file is untested? Is that expected?

nasdf: I've added a simple smoke test

internal/metric/metric_test.go
nasdf: I'm expecting metrics to be added as needed instead of pre-emptively. I can restore the code when we have a need for it, but I have a preference to not keep dead code.

I'm also curious why the previous code was dumping metrics locally. From all of the research I did, it seems like sending the metrics to a collector or exporter is the preferred setup.

shahzadlone: Cool then it's all good.

The reason it was done that way was because back then that's all that was implemented for GoLang support.

internal/metric/otel_test.go
```diff
@@ -0,0 +1,25 @@
...
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+//go:build telemetry
+
+package metric
```
shahzadlone, line 13: nitpick: Feel free to rename the package to telemetry at this point.

internal/request/graphql/parser.go
```diff
@@ -50,7 +50,7 @@ func NewParser() (*parser, error) {
 }
 
 func (p *parser) BuildRequestAST(ctx context.Context, request string) (*ast.Document, error) {
-	_, span := tracer.Start(ctx)
+	ctx, span := tracer.Start(ctx) //nolint:ineffassign,staticcheck
```
shahzadlone, line 53: todo: Why do you need to do this. 1) `ctx` is not used anywhere and 2) it is ineffectual assignment.

I like what you had before.

Same for all other instances

fredcarle: He change it because Andy asked him to do so I think. I'm also not a fan of this change.

internal/request/graphql/parser.go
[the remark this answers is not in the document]
shahzadlone: I do have a slightly stronger preference to leave as ot was and not do the linter suppression. Sorry Keenan

internal/request/graphql/parser.go
shahzadlone: Didn't seem like a strong preference by him, I commented on the original thread now.

internal/db/db.go
nasdf: The preview above is outdated. Should be fixed in the actual code.

internal/request/graphql/parser.go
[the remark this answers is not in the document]
nasdf: I ended up reverting the change since more of us preferred the original.

internal/metric/otel_test.go
nasdf: done

AndrewSisley (approved):
LGTM, thanks Keenan!

PR #3405, opened 2025-01-25
feat: Add contraints and embedding directives

### LINKED ISSUE #3350: Create embeddings based on document fields

### LINKED ISSUE #3351: Add size constraint to array types

fredcarle:
## Relevant issue(s)

Resolves #3350
Resolves #3351

## Description

Sorry for having 3 technically separate features as part of 1 PR. The reason for this is that I started working on embeddings and the size contraint was initially part of the embedding. Since we discussed that it could be applied to all array fields (and later even to String and Blob), I extracted it into a contraints directive that has a size parameter (more contraints can be added in the future). Furthermore, embeddings returned from ML models are arrays of float32. This caused some precision issues because we only supported float64. When saving the float32 array, querying it would return an float64 array with slight precision issues. I decided to add the float32 type.

You can review the first commit for contraint and embedding related code and the second commit for the float related changes. Some float stuff might have leaked in the first commit. Sorry for this. I tried hard to separate the float32 related changes.

Note that the `gql.Float` type is now `Float64` internally.
```graphql
type User {
  points: Float
}
```
is the same as
```graphql
type User {
  points: Float64
}
```

The embedding generation relies on a 3rd party package called `chromem-go` to call the model provider API. As long as one of the supported provider API is configured and accessible, the embeddings will be generated when adding new documents. I've added a step in the  test workflow that will run the embedding specific tests on linux only (this is because installation on windows and mac is less straight forward) using Ollama (because it runs locally).

The call to the API has to be done synchronously otherwise the docID/CID won't be representative of the contents. The only alternative would be for the system to automatically update the document when returning from the API call but I see that as a inferior option as it hides the update step from the user. It could also make doc anchoring more complicated as the user would have to remember to wait on the doc update before anchoring the doc at the right CID.

We could avoid having embedding generation support and let the users do that call themselves and store the embedding vectors directly. However, having it as a feature allows us to support RAG vector search which would let users get started with AI with very little work. This seems to be something our partners are looking forward to.

I don't see the 3rd party API call inline with a mutation as a problem since this is something that has to be configured by users and users will expect the mutation calls to take more time as a result.

If you're interested in running it locally, install Ollama and define a schema like so
```graphql
type User {
    name: String
    about: String
    name_v: [Float32!] @constraints(size: 768) @embedding(fields: ["name", "about"], provider: "ollama", model: "nomic-embed-text",  url: "http://localhost:11434/api") // contraint is optional and localhost:11434 is the default port for Ollama
}
```

Next steps: 
 - Support templates for the content sent to the model.
 - Add the `_similarity` operation to calculate the cosine similarity between two arrays.

## Tasks

## How has this been tested?

make test and manual testing

- MacOS


internal/db/collection.go
```diff
@@ -520,6 +525,11 @@ func (c *collection) update(
 	ctx context.Context,
 	doc *client.Document,
 ) error {
+	err := c.setEmbedding(ctx, doc, false)
```
AndrewSisley, line 528: question: This is making a remote call to a third party service no (ollama or openai)?

fredcarle: If embedding is defined yes.

AndrewSisley (changes requested):
I think the calls out to 3rd party APIs needs expanded documentation and discussion with the team.  I have reviewed very little so far but will continue whilst that issue is being resolved.

internal/db/collection.go
fredcarle: PR description has been updated

.github/workflows/test-coverage.yml
```diff
@@ -74,6 +74,12 @@ jobs:
       DEFRA_MUTATION_TYPE: ${{ matrix.mutation-type }}
 
     steps:
+      - name: Install ollama
+        run: curl -fsSL https://ollama.com/install.sh | sh
```
shahzadlone, line 78: question: This is only needed for basic test matrix?

client/normal_new.go
```diff
@@ -348,11 +351,52 @@ func convertAnyArrToNillableIntOrFloatArr(arr []any) (NormalValue, error) {
...
+		if arr[i] == nil {
+			return convertAnyArrToNillableFloat32Arr(arr)
+		}
+
+		var floatVal float32
+		switch v := arr[i].(type) {
+		case float32:
+			floatVal = v
+		case float64:
+			return convertAnyArrToFloat64Arr(arr)
```
shahzadlone, line 366: question: The whole array will be converted to Float64?

.github/workflows/test-coverage.yml
fredcarle: Yes. Why are you asking?

client/normal_new.go
fredcarle: Yes. There was a similar pattern before for the float64 function. It's the same thing here.

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+// setEmbedding sets the embedding fields on the document if the related fields are dirty.
+func (c *collection) setEmbedding(ctx context.Context, doc *client.Document, isCreate bool) error {
+	embeddingGenerated := false
+	for _, colField := range c.Definition().GetFields() {
+		if colField.Embedding != nil {
+			fieldsVal := make(map[string]client.NormalValue)
+			needsGeneration := false
```
AndrewSisley, line 52: question: It looks like if a user explicitly provides their own vector value it will be overwritten by the result of the remote call.  Would it not be better to allow users to provide their own values?

Besides being less surprising for them, it would give them greater control over the determinism of the resultant CID/DocID, which I believe can vary depending on the timing of the remote calls (under most AI models).

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+	default:
+		return nil, NewErrUnknownEmbeddingProvider(provider)
+	}
+}
+
+// setEmbedding sets the embedding fields on the document if the related fields are dirty.
+func (c *collection) setEmbedding(ctx context.Context, doc *client.Document, isCreate bool) error {
+	embeddingGenerated := false
+	for _, colField := range c.Definition().GetFields() {
+		if colField.Embedding != nil {
```
AndrewSisley, line 50: suggestion: Inverting this `if` and `continue`-ing would remove quite a long indentation, and IMO make the code a bit easier to read.

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+					colField.Embedding.URL,
+				)
+				if err != nil {
+					return err
+				}
+
+				text := ""
+				for _, fieldName := range colField.Embedding.Fields {
+					if val, ok := fieldsVal[fieldName]; ok {
+						text += fmt.Sprintf("%v\n", val.Unwrap())
```
AndrewSisley, line 97: suggestion: StringBuilder might be more appropriate here, it might be more readable, and should be more efficient, although the performance gains would probably be small compared to the cost of a remote call.

internal/request/graphql/schema/collection.go
```diff
@@ -596,6 +624,92 @@ func policyFromAST(directive *ast.Directive) (client.PolicyDescription, error) {
...
+	Size int
+}
+
+func contraintsFromAST(kind client.FieldKind, directive *ast.Directive) (constraintDescription, error) {
+	constraints := constraintDescription{}
+	for _, arg := range directive.Arguments {
+		switch arg.Name.Value {
+		case types.ConstraintsDirectivePropSize:
+			if !kind.IsArray() {
+				return constraintDescription{}, errors.New("size constraint can only be applied to array fields")
```
AndrewSisley, line 693: todo: This error should be declared in an errors.go file

internal/request/graphql/schema/collection.go
```diff
@@ -13,6 +13,7 @@ package schema
 import (
```
AndrewSisley, line 13: todo: There appears to be a lot of schema/collection description validation taking place in this file.  Please remove as much as you can and instead place it in `internal/db/definition_validation.go`.

This will allow all our schema validation to live in once place, and lets it operates on our lowest-level/core types, instead of within query-language dependent code-spaces.

internal/request/graphql/schema/types/base.go
```diff
@@ -216,53 +216,181 @@ func FloatListOperatorBlock(op *gql.InputObject) *gql.InputObject {
...
+			},
+		},
+	})
+}
+
+// Float32ListOperatorBlock filter block for [Float32] types.
+func Float32ListOperatorBlock(op *gql.InputObject) *gql.InputObject {
+	return gql.NewInputObject(gql.InputObjectConfig{
+		Name:        "Float32ListOperatorBlock",
+		Description: "These are the set of filter operators available for use when filtering on [Float32] values.",
```
AndrewSisley, line 329: todo: Please move these descriptions to the descriptions.go file

internal/request/graphql/schema/types/scalars.go
```diff
@@ -121,3 +122,281 @@ func JSONScalarType() *graphql.Scalar {
...
+	}
+
+	// If the value cannot be transformed into an float, return nil instead of '0.0'
+	// to denote 'no float found'
+	return nil
+}
+
+var Float32 = graphql.NewScalar(graphql.ScalarConfig{
+	Name: "Float32",
+	Description: "The `Float32` scalar type represents signed single-precision fractional " +
```
AndrewSisley, line 242: todo: Please move this description to descriptions.go

AndrewSisley (changes requested):
I've given the production code a first-pass review, and will shortly look through the tests.  The code looks good, most of the comments are about moving stuff around, or questions/thoughts RE the feature design.

thought: It looks very tedious to introduce new types atm, although numerics are probably most painful.  Perhaps we should review certain aspects in the future to improve this.

internal/db/embedding.go
fredcarle: It shouldn't overwrite. If it does it's a mistake on my part. Users should be free to use their own vectors.

AndrewSisley: Ah okay nice!  

todo: I've not read the tests yet, could you please make sure that there is at least one that covers that scenario?

suggestion: Please also note this behaviour in the func documentation

client/document.go
```diff
@@ -475,18 +522,22 @@ func getArray[T any](
...
-		return arr, nil
+		array = arr
 	case []T:
-		return val, nil
-	default:
-		return []T{}, nil
+		array = val
 	}
+	if size != 0 && len(array) != size {
+		return nil, NewErrArraySizeMismatch(array, size)
```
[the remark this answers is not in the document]
fredcarle: This function is called by `validateFieldSchema`. So it's done on write.

tests/integration/events.go
```diff
@@ -333,28 +333,6 @@ func getEventsForUpdateDoc(s *state, action UpdateDoc) map[string]struct{} {
 	}
 }
 
-// getEventsForCreateDoc returns a map of docIDs that should be
-// published to the local event bus after a CreateDoc action.
-func getEventsForCreateDoc(s *state, action CreateDoc) map[string]struct{} {
```
AndrewSisley, line 338: question: Why has this been removed?

tests/integration/mutation/create/constraints/size_constraint_test.go
```diff
@@ -0,0 +1,86 @@
...
+							_docID
+							name
+							numbers
+						}
+					}
+				`,
+				Results: map[string]any{
+					"Users": []map[string]any{
+						{
+							"_docID":  testUtils.NewDocIndex(0, 0),
```
AndrewSisley, line 50: suggestion: `_docID` is quite an expensive property maintenance-wise to include in a test, and it does not appear to have a use here, I suggest removing it from the query.

Same goes for the p[n]counter tests.

tests/integration/mutation/create/embeddings/embedding_test.go
```diff
@@ -0,0 +1,163 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+//go:build linux
```
AndrewSisley, line 11: todo: Please document in this file why these tests are Linux only.  I know you noted this in the PR description, but if this constraint persists it will take some figuring out as to why it is Linux only.

suggestion: Consider creating a ticket and linking it to and from here to remove this constraint.

tests/integration/mutation/create/embeddings/embedding_test.go
```diff
@@ -0,0 +1,163 @@
...
...94595, 0.02624271, 0.0066641932, 0.019060731, 0.05616408, 0.01946424, 0.050201155, 0.016927207, -0.038973548, 0.037738074, -0.056299437, 0.00967483, -0.018431908, 0.01668865, -0.047797382, 0.015456402, 0.018490652, -0.012237633, 0.012219139, 0.026376488, -0.023400653, -0.04556467, 0.064960316, 0.0019083915, 0.05282892, 0.043711416, 0.07134014, -0.029614225, -0.040969286, 0.045358557, -0.031595223, -0.01910029, -0.061545778, 0.024926651, 0.04099986, 0.01620945, 0.042243116, 0.04717351, -0.0048233513, -0.0019400177, -0.036986295, -0.039084256, 0.025065007, 0.03538771, -0.00042725189, 0.011488134, -0.018427324, -0.019081326, -0.03214962, 0.05602044, -0.07190716, 0.006701344, -0.027393665, 0.025218304, 0.021957463, -0.08623837, 0.026349807, -0.020874213, 0.016827395, -0.051633146, -0.016894199, 0.009786709, 0.026273636, 0.013940125, 0.011852815, 0.036727257, 0.055551354, 0.008069757, -0.015989762, 0.01829905, 0.060527466, 0.048552617, 0.01770399, -0.03225409, -0.042072583, -0.026071887, -0.036898546, -0.05571102, -0.0058531873, -0.047945924, -0.0003473616, 0.038228065, 0.018455006, 0.025350647, 0.00965173, 0.00064714643, 0.01915388, -0.015175882, -0.040475022, -0.0338664, 0.034016117},
```
AndrewSisley, line 150: thought: I'm not sure including the full array helps people read the test, and the only benefit to asserting anything beyond 'is it an array with some values' is perhaps the change detector. 

question: Do you have plans to introduce more tests like this in the near future?

suggestion: If you do want to add more, can we perhaps look at asserting on this value in a slightly different way, perhaps only checking that it is of an expected size, with non-zero values.

AndrewSisley (changes requested):
I've reviewed the tests, thanks for adding them, I think there are a couple of important test gaps though, please add tests for:

1. Patching a `size` constraint, including removing it 
2. Patching an `embedding` property, and it's children:
    2.1 Adding an embedding description to a field that didnt have one before
    2.2 Mutating each of the embedding properties on a field that declared one on create
    2.3 Removing an embedding property from a field that had one on create 
3. Including a `size` constraint in a View declaration, if you don't/haven't blocked it off, please include a test that shows what happens if that constraint is violated
4. Including an `embedding` property in a View declaration (I assume we want this to error for now).

client/document.go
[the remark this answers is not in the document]
AndrewSisley: Nice, thanks!

tests/integration/mutation/create/embeddings/embedding_test.go
AndrewSisley, line 11: todo: Please change the way you constrain these tests.  The current setup will prevent me, or any other linux users from running `go test ./...` or even `make test`, this could be a bit of a nuisance.  Perhaps introduce something similar to the `SupportedClients` `TestCase` property.

tests/integration/utils.go
```diff
@@ -1224,14 +1224,13 @@ func createDoc(
...
+
 	if action.ExpectedError == "" {
-		waitForUpdateEvents(
-			s,
-			action.NodeID,
-			action.CollectionID,
-			getEventsForCreateDoc(s, action),
-			action.Identity,
-		)
+		waitForUpdateEvents(s, action.NodeID, action.CollectionID, docIDMap, action.Identity)
```
shahzadlone, line 1233: question: curious why this needed to be done?

.github/workflows/test-coverage.yml
shahzadlone: because:
1) I don't understand why this is not needed for other test types (acp, view, encryption, etc.).
2) Is this a required dependency now? if so then:
	- document in `README`
	- add a make rule in `Makefile`
	- perhaps add this to the setup composite instead (`./.github/composites/setup-defradb`)

fredcarle: 1. When you see "Ollama", do you need more documentation to know what this is about? I might be wrong to assume that would be enough to understand that this is only for embedding related testing.
2. It's not a Defra dependency. Only a testing dependency for the embedding specific tests.

tests/integration/utils.go
fredcarle: Because the ids were already available on create and since the embedding vector is not know in advance, we cannot depend on the documents content in the test's definition to determine the docIDs.

.github/workflows/test-coverage.yml
shahzadlone: 1) Maybe I missed it but last I checked the embedding tests they didn't have `SupportedTypes` like config. So I assumed they will run with all configurations. For example when encryption or macos tests run, wouldn't they still run and hence be need this dep?

shahzadlone: 2. Yea so similar to how we have rust dep for lens test, this should have a make rule imo and will be part of test/dev deps

internal/db/collection.go
jsimnz: Follow up: Discussed during the standup. We'll leave the "in-request" embedding network call as is for now, and start expanding other options for async embedding generation soon. 

cc: @fredcarle can you create a follow up ticket for this?

client/collection_field_description.go
```diff
@@ -43,6 +43,22 @@ type CollectionFieldDescription struct {
...
+
+	// Embedding contains the configuration for generating embedding vectors.
+	//
+	// This is only usable with array fields.
+	//
+	// When configured, embeddings may call 3rd party APIs inline with document mutations.
+	// This may cause increase latency in the completion of the mutation requests.
+	// This is necessary to ensure that the generated docID is representative of the
+	// content of the document.
+	Embedding *EmbeddingDescription
```
jsimnz, line 61: suggestion: What if the structure/placement of the `EmbeddingDescription` was similar to how we have `IndexDescription`, where it actually lives on the `CollectionDescription` instead here on the `CollectionFieldDescription`.

This would make it much easier to iterate through the existing embedding descriptions when mutating documents to update/generate the embeddings. Whereas right now you have to iterate through all the fields on a doc. 

It also opens up the door to not having to actually have the embedding in the document graph itself, and instead is stored as an index outside the graph (future work).

internal/db/collection.go
jsimnz, line 528: question: Is there a particular reason why the first thing the `create/update` call does is generate the embeddings? in the `update` case, it gets the embeddings before the ACP check is even done to determine if a mutation is allowed? It could even be in the `save` func deeper in the call graph instead of duplicated in the `create` and `update` paths.

internal/db/embedding.go
jsimnz: related to my suggestion above regarding moving the `EmbeddingDescription` to the `CollectionDescription` instead of the `CollectionFieldDescription` would simplify this function, where it can go straight to the embeddings rather than scanning all the fields to see if there is a description on them.

internal/request/graphql/schema/collection.go
jsimnz: are you referring to the new code in this PR or in general for this file?

jsimnz (dismissed):
Marking this as "request for changes" (you can dismiss at any time), even tho there aren't any explcit `todos`. 

I have mainly focused on the definition/description code along with the collection/doc updating. Speed ran through the encoding section and the gql parsing.

Haven't reviewed the tests.

Main question/discussion points are "when" you are generating the embedding in the call path, and the location of the `EmbeddingDescription`.

internal/request/graphql/schema/collection.go
AndrewSisley: The new code in this PR.  We have existing validation in this file, but *only* for GQL/SDL specific validation.  The current PR contains a lot of general validation, such as validating that target is an array-float kind - this needs to move.

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+			missingFieldsForGeneration := []client.FieldDefinition{}
+			for _, embedField := range colField.Embedding.Fields {
+				if docField, ok := doc.Fields()[embedField]; ok {
+					if doc.Values()[docField].IsDirty() {
+						needsGeneration = true
+						fieldsVal[embedField] = doc.Values()[docField].NormalValue()
+					} else {
+						fieldDef, ok := c.def.GetFieldByName(embedField)
+						if !ok {
+							return errors.New("field not found", errors.NewKV("field", embedField))
```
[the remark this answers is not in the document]
AndrewSisley: Yes please :)

tests/integration/mutation/create/embeddings/embedding_test.go
islamaliev: as an idea, we can create a new `Validator` for asserting this. Similar to `NewUniqueCid` in `tests/integration/results.go`

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+						}
```
islamaliev, line 63: suggestion: I would move this part on the GraphQL side to catch error earlier and potentially make better user experience.

It should work pretty much the same way as we check fields for indexes. Find in the project `indexFieldFromAST`.

I would make it "todo" actually, but our experiment... :)

islamaliev: you can make `fields` in `@embedding` work as enum:
```graphql
@embedding(fields: ["name", "about"]
type User {
    name: String
    about: String
    name_v: [Float32!] @embedding(fields: ["name", "about"], ...
}
```
In this example GraqhQL would allow to choose only between "name" and "about" which (with schema inspection and suggestions) would give a better UX.

Note: I don't know if it would include also "name_v" itself. My guess is yes, but this can be filtered out as well.

question: what actually happens if an embedding refers to another embedding or even itself: `name_v: [Float32!] @embedding(fields: ["name", "name_v"], ...`
What about different embedding referring each other:
```
name_v: [Float32!] @embedding(fields: ["name", "about_v"], ...
about_v: [Float32!] @embedding(fields: ["about", "name_v"], ...
```
It seems to me that we need to forbid this. I'm sure this can be easily done by restricting number of available options on GraphQL side similar to what secondary indexes do: just take a list of available fields and filter out the ones with embeddings.

client/document.go
```diff
@@ -886,7 +939,11 @@ func (doc *Document) setDocID(docID DocID) {
 	doc.id = docID
 }
 
-// generateAndSetDocID generates the DocID and then (re)sets `doc.id`.
+// GenerateAndSetDocID generates the DocID and then (re)sets `doc.id`.
+func (doc *Document) GenerateAndSetDocID() error {
+	return doc.generateAndSetDocID()
```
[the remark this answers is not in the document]
fredcarle: The point is to be able to keep the internal (inside client package) usage using the private function and have the public function to use externally. This enables modification (removal or otherwise) of the public function without affecting the interal code. This is very common practice.

internal/db/collection.go
fredcarle: Issue created: https://github.com/sourcenetwork/defradb/issues/3427

internal/db/embedding.go
fredcarle: Done and done

internal/db/embedding.go
fredcarle: Inverted the `if` and moved the embedding description to the collection description.

internal/db/embedding.go
fredcarle: Changed to StringBuilder

internal/request/graphql/schema/collection.go
fredcarle: Moved to the collection/schema validation part of the repo.

tests/integration/mutation/create/constraints/size_constraint_test.go
fredcarle: Done

tests/integration/mutation/create/embeddings/embedding_test.go
fredcarle: Done

tests/integration/mutation/create/embeddings/embedding_test.go
fredcarle: Changed the way they are executed.

fredcarle: Done

client/collection_field_description.go
fredcarle: Done

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+						missingFieldsForGeneration = append(missingFieldsForGeneration, fieldDef)
+					}
+				}
+			}
+			if needsGeneration && len(missingFieldsForGeneration) > 0 {
```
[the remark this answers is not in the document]
fredcarle: Done

client/schema_field_description.go
```diff
@@ -322,28 +332,31 @@ func (k *NamedKind) IsArray() bool {
...
+	FieldKind_NILLABLE_JSON          ScalarKind      = 14
+	_                                ScalarKind      = 15 // safe to repurpose (was never used)
+	_                                ScalarKind      = 16 // Deprecated 2024-03-15, was FieldKind_FOREIGN_OBJECT
+	_                                ScalarKind      = 17 // Deprecated 2024-03-15, was FieldKind_FOREIGN_OBJECT_ARRAY
+	FieldKind_NILLABLE_BOOL_ARRAY    ScalarArrayKind = 18
+	FieldKind_NILLABLE_INT_ARRAY     ScalarArrayKind = 19
+	FieldKind_NILLABLE_FLOAT64_ARRAY ScalarArrayKind = 20
+	FieldKind_NILLABLE_STRING_ARRAY  ScalarArrayKind = 21
+	FieldKind_NILLABLE_FLOAT32       ScalarKind      = 22
+	FieldKind_FLOAT32_ARRAY          ScalarArrayKind = 23
```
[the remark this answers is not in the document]
fredcarle: Done

internal/db/embedding.go
fredcarle: Added validation that works for both GQL and patch. It takes care of validating types and validity of the fields.

internal/request/graphql/schema/collection.go
fredcarle: Done

internal/request/graphql/schema/types/base.go
fredcarle: I'm not a fan of not having this inline. I'd rather be able to read it with the rest of the object declaration. We currently have a mix of both so I'm not sure we've determined an official consensus to warrant a todo here.

internal/db/collection.go
fredcarle: It can't be in the `c.save` method because save is used with both create and update. It might be ok for update but on create, the `c.setEmbedding` bust be call before other actions that are create specific like ensuring the docID is unique and such.

So for create I'll leave it where it is and for update I'll move it bellow the ACP check. Good catch on that one.

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+				oldDoc, err := c.get(
+					ctx,
+					keys.DataStoreKeyFromDocID(doc.ID()).ToPrimaryDataStoreKey(),
+					missingFieldsForGeneration,
+					false,
+				)
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
fredcarle: Done

internal/db/embedding.go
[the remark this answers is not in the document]
fredcarle: Done

internal/db/embedding.go
```diff
@@ -0,0 +1,122 @@
...
+
+				text := ""
+				for _, fieldName := range colField.Embedding.Fields {
+					if val, ok := fieldsVal[fieldName]; ok {
+						text += fmt.Sprintf("%v\n", val.Unwrap())
+					}
+				}
+				embedding, err := embeddingFunc(ctx, text)
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
fredcarle: Done

client/normal_value_test.go
```diff
@@ -1658,11 +1992,11 @@ func TestArrayValue_IsNillable(t *testing.T) {
 	fieldKinds := []FieldKind{
 		FieldKind_BOOL_ARRAY,
 		FieldKind_INT_ARRAY,
-		FieldKind_FLOAT_ARRAY,
+		FieldKind_FLOAT64_ARRAY,
 		FieldKind_STRING_ARRAY,
 		FieldKind_NILLABLE_BOOL_ARRAY,
 		FieldKind_NILLABLE_INT_ARRAY,
-		FieldKind_NILLABLE_FLOAT_ARRAY,
+		FieldKind_NILLABLE_FLOAT64_ARRAY,
```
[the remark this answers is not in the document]
fredcarle: Nice catch. Done

client/schema_field_description_test.go
```diff
@@ -54,8 +54,8 @@ func TestField_ScalarArray_HasSubKind(t *testing.T) {
 		},
 		{
 			name:    "nillable float array",
-			arrKind: FieldKind_NILLABLE_FLOAT_ARRAY,
-			subKind: FieldKind_NILLABLE_FLOAT,
+			arrKind: FieldKind_NILLABLE_FLOAT64_ARRAY,
```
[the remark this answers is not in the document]
fredcarle: Nice catch. Done

tests/integration/collection_description/updates/add/collections_test.go
```diff
@@ -30,7 +30,7 @@ func TestColDescrUpdateAddCollections_WithUndefinedID_Errors(t *testing.T) {
 						{ "op": "add", "path": "/2", "value": {"Name": "Dogs"} }
 					]
 				`,
-				ExpectedError: "collection ID cannot be zero",
+				ExpectedError: "schema name can't be empty",
```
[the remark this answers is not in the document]
fredcarle: It's not really a bug. It's the order of validation. In this case the user would be trying to add a collection without a related schema which is not possible. 

I'm open to suggestions though if you have any regarding. How we should handle this.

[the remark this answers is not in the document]
AndrewSisley: Ignore me lol, I overestimated the usefulness of the original error :grin:  Problem is out of scope

tests/integration/collection_description/updates/add/collections_test.go
```diff
@@ -74,7 +74,7 @@ func TestColDescrUpdateAddCollections_Errors(t *testing.T) {
 						{ "op": "add", "path": "/2", "value": {"ID": 2, "Name": "Dogs"} }
 					]
 				`,
-				ExpectedError: "adding collections via patch is not supported. ID: 2",
+				ExpectedError: "schema name can't be empty",
```
AndrewSisley, line 77: todo: Sorry, whilst my last comment was misplaced, this error *was* useful to the user, please revert this test and change the production code to accomodate.

tests/integration/collection_description/updates/add/embeddings_test.go
```diff
@@ -0,0 +1,136 @@
...
+					{
+						Name:           immutable.Some("Users"),
+						IsMaterialized: true,
+						Embeddings: []client.EmbeddingDescription{
+							{
+								FieldName: "name_v",
+								Fields:    []string{"name", "about"},
+								Provider:  "ollama",
+								Model:     "nomic-embed-text",
+								URL:       "http://localhost:11434/api",
```
AndrewSisley, line 127: todo: Please add tests covering the omission of each property in a patch, and SDL ( e.g. what happens if the user does not provide a `url`?).

client/collection_description.go
```diff
@@ -103,6 +103,16 @@ type CollectionDescription struct {
...
+
+	// Embedding contains the configuration for generating embedding vectors.
+	//
+	// This is only usable with array fields.
+	//
+	// When configured, embeddings may call 3rd party APIs inline with document mutations.
+	// This may cause increase latency in the completion of the mutation requests.
+	// This is necessary to ensure that the generated docID is representative of the
+	// content of the document.
+	Embeddings []EmbeddingDescription
```
AndrewSisley, line 115: suggestion: Hosting this on the collection description makes it local.  Whilst I agree with hosting at least part of this here, because it allows local, independent changes, to stuff like `URL`, I am much less certain about including stuff like `fields` here.

It also allows different nodes to manage documents using different models, which can then be synced over P2P.  They will have the same type, yet may not be comparable.

This is a very large PR, and so whilst changing this later would an annoying and breaking change, I am uncertain as to whether the cost of delaying stuff like `Size`, and increasing author/reviewer-fatigue, is greater than the cost incurred by the potential misplacement of some of these properties, so, my suggestion is to document this potential clash to try and increase the likelihood of users being aware of this type-gap.  And perhaps create a ticket to properly review the placement of all of these properties.

client/document.go
```diff
@@ -886,7 +939,11 @@ func (doc *Document) setDocID(docID DocID) {
 	doc.id = docID
 }
 
-// generateAndSetDocID generates the DocID and then (re)sets `doc.id`.
+// GenerateAndSetDocID generates the DocID and then (re)sets `doc.id`.
+func (doc *Document) GenerateAndSetDocID() error {
```
AndrewSisley, line 943: todo: Code-cov now says this is untested, and I couldn't spot any references to it.  The documentation is probably not clear enough for users to know when they should be calling this.

Please either remove this function, or expand the documentation if you think users should be calling this.

internal/core/encoding.go
```diff
@@ -110,6 +128,19 @@ func NormalizeFieldValue(fieldDesc client.FieldDefinition, val any) (any, error)
 			case uint:
 				return float64(v), nil
 			}
+		case client.FieldKind_NILLABLE_FLOAT32:
+			switch v := val.(type) {
```
AndrewSisley, line 132: todo: Code-cov is flagging a lot of this as untested.  Is it dead code?  Will declaring an array `[1, 1.1]` not result in at least one of the `int` cases being hit?

Please include a test where some of the values lack decimal points if there are no such tests already.

If there are such tests, I *suggest* the deletion of the dead code branches.

Same todo/suggestion applies to `convertToFloat32` too.

internal/db/base/compare.go
```diff
@@ -39,6 +39,8 @@ func Compare(a, b any) int {
 		return compareInt(v, b.(int64))
 	case uint64:
 		return compareUint(v, b.(uint64))
+	case float32:
```
AndrewSisley, line 42: todo: Please either delete or cover this with an integration test (assuming code-cov isn't faulty, maybe add a panic and see if any tests fail).

internal/db/collection_define.go
```diff
@@ -120,15 +120,19 @@ func (db *DB) patchCollection(
 	if err != nil {
 		return err
 	}
-	txn := mustGetContextTxn(ctx)
-	cols, err := description.GetCollections(ctx, txn)
+	existingCols, err := db.getCollections(
+		ctx,
+		client.CollectionFetchOptions{IncludeInactive: immutable.Some(true)},
```
AndrewSisley, line 125: todo: Please document why you are fetching the entire collection version history for the entire database.  I cannot tell why we would want inactive versions here, and it could be a very large amount of data.

internal/request/graphql/schema/collection.go
```diff
@@ -596,6 +622,94 @@ func policyFromAST(directive *ast.Directive) (client.PolicyDescription, error) {
...
 
+func embeddingFromAST(directive *ast.Directive, fieldDef *ast.FieldDefinition) (client.EmbeddingDescription, error) {
+	embedding := client.EmbeddingDescription{
+		FieldName: fieldDef.Name.Value,
+	}
+	for _, arg := range directive.Arguments {
+		switch arg.Name.Value {
+		case types.EmbeddingDirectivePropFields:
+			val, ok := arg.Value.(*ast.ListValue)
+			if !ok {
```
AndrewSisley, line 633: todo: The cast checking (`!ok`) checks in `embeddingFromAST` and `contraintsFromAST` are untested, and it looks like they should be covered by the GQL lib when it parses the SDL.  Please either cover them with integration tests, or remove them.

internal/request/graphql/schema/collection.go
```diff
@@ -651,8 +767,10 @@ func astTypeToKind(
 				return client.FieldKind_NILLABLE_BOOL_ARRAY, nil
 			case typeInt:
 				return client.FieldKind_NILLABLE_INT_ARRAY, nil
-			case typeFloat:
-				return client.FieldKind_NILLABLE_FLOAT_ARRAY, nil
+			case typeFloat, typeFloat64:
+				return client.FieldKind_NILLABLE_FLOAT64_ARRAY, nil
+			case typeFloat32:
```
AndrewSisley, line 772: todo: I think I saw a comment from Shahzad asking you to test this, but just in case it was for something else, please make sure this is tested.

internal/request/graphql/schema/errors.go
```diff
@@ -182,3 +195,35 @@ func NewErrFieldTypeNotSpecified(objectName, fieldName string) error {
 		errors.NewKV("Field", fieldName),
 	)
 }
+
+func NewErrDirectiveWithUnknownArg(directive, arg string) error {
```
AndrewSisley, line 199: todo: After dealing with my comment regarding the GQL validation, please review these and remove any that are no longer used.

AndrewSisley (changes requested):
The features look good, but I think there are still a few more tests and documentation to write.

Depending on any new/pending feedback from the others, it is still not too late to break this up into smaller PRs, especially if we want to deal with the collection vs schema description problem here.

client/collection_description.go
fredcarle: I don't think it should be hosted anywhere else as this is a local configuration for ***convenience only***. Devs can manually set the vector to something incompatible with the configured model and there isn't much we can and should do about this. This extends to syncing over P2P. We can do all sorts of gymnastics to ensure that nodes are configured with the same model but a incompatible manually defined vector will propagate just the same. I'm currently very much against adding guardrails around this as it would restrict flexibility for devs.

client/document.go
fredcarle: That a bug with codecov because it is definitely used. It used anytime a new embedding is generated on document creation. This is what ensure the docID is representative of the contents.

internal/core/encoding.go
fredcarle: This function is used on unmarshalling from CBOR. As a result, I think a lot of this is defensive programming. I see this todo as out of scope of this PR. We can open an issue to review this at some point if you'd like though.

internal/db/collection_define.go
fredcarle: I'm fetching the entire collection version history because that's what it was previously doing. Only this time it includes the related schemas. It's needed because we are allowing patching of inactive collections as described in the tests you added for this.

client/collection_description.go
AndrewSisley: I understand what you mean, but caution against giving too much weight to that line of thinking.  Two nodes in a network may be owned by two different entities, they may have agreed upon a shared schema but may be managed by totally different teams.  Doing what we can to encode the model in the type could save a large amount of money for our users.

(just chatting here, not pushing for change)

internal/request/graphql/schema/collection.go
fredcarle: It was something else. I'll look into testing.

internal/core/encoding.go
AndrewSisley: I'm not sure I understand why adding an integration test with an array `[1, 1.1]` or similar would be out of scope as it would be covering behaviour introduced in this PR.

Or are you saying such a test would not cover that case?

internal/db/collection_define.go
AndrewSisley: Ah, so it does, thanks!  A comment would still be handy here, but fair enough if you want to dismiss that as out of scope.

internal/core/encoding.go
fredcarle: I think it's out of scope because it is not new behaviour. It's just a new case in the switch for a different type but the contents of the case are the same as the others which are also not covered. To cover all the possibilities, we would have to figure out what CBOR converts the values as when marshalling. So I'm not sur that the test you propose would cover much more.

client/collection_description.go
fredcarle: We can discuss at the next standup no problem.

internal/db/base/compare.go
fredcarle: Done

internal/request/graphql/schema/collection.go
fredcarle: Done

internal/request/graphql/schema/errors.go
fredcarle: Done

internal/request/graphql/schema/collection.go
fredcarle: I added a test that creates a schema with all supported types to ensure we hit a maximum of these.

client/document.go
fredcarle: The environment variable in the workflow had a typo. The embedding related test where not executed which affected code coverage.

Makefile
```diff
@@ -170,6 +170,19 @@ deps\:mocks:
 deps\:playground:
 	cd $(PLAYGROUND_DIRECTORY) && npm install --legacy-peer-deps && npm run build
 
+/PHONY: deps\:embedding
```
shahzadlone, line 173: todo:

```suggestion
.PHONY: deps\:embedding
```

Makefile
```diff
@@ -170,6 +170,19 @@ deps\:mocks:
...
+ifeq ($(OS_GENERAL),Linux)
+	curl -fsSL https://ollama.com/install.sh | sh
+else
+# run ollama in the background
+	nohup ollama serve > ollama.log 2>&1 &
+endif
+# make sure ollama is running before continuing
+	time curl --retry 5 --retry-connrefused --retry-delay 0 -sf http://localhost:11434
+# pull a model
+	ollama pull nomic-embed-text
```
shahzadlone, line 184: suggestion: I prefer the name to be `deps:ollama` instead of `deps:embedding`.

suggestion: Installing ollama, running ollama and pulling a model should be a separate command.
Perhaps:
- `make deps:ollama` to install ollama.
- `make ollama` to serve/start in background. 
- `make ollama:nomic` to pull the model

suggestion: If it is not Linux, can just print that the user must install it manually if have to use UI to download.

Note: the linux vs non-linux differences can be handled in the individual `ifeq/else` blocks in the separate make rules suggested above if you do want to provide them.

client/schema_field_description.go
```diff
@@ -32,6 +32,9 @@ type FieldKind interface {
 	// IsArray returns true if this FieldKind is an array type which includes inline arrays as well
 	// as relation arrays.
 	IsArray() bool
+
+	// IsNumericArray returns true if the FieldKind is an array that contains numerical scalars.
+	IsNumericArray() bool
```
islamaliev, line 37: suggestion: I don't feel good about adding this method to the interface. The interface provides all the necessary information and if you need some helper methods you can add them out of the interface.

Otherwise we will start polluting it all other types: `IsStringArray`, `IsBoolArray` and do on.

It's it's an array (i.e. `IsArray() == true`) you can probably cast it to `ScalarArrayKind` and read it's sub kind the same way it's done in `internal/keys/datastore_index.go` in function `DecodeIndexDataStoreKey`:

```go
if kind != nil && kind.IsArray() {
	if arrKind, ok := kind.(client.ScalarArrayKind); ok {
		kind = arrKind.SubKind()
	}
}
```

.github/workflows/test-coverage.yml
```diff
@@ -80,6 +83,9 @@ jobs:
       - name: Setup defradb
         uses: ./.github/composites/setup-defradb
 
+      - name: Setup vector embedding dependencies
+        run: make deps:embedding
+
```
shahzadlone, line 88: todo: I have a preference to not do fancy and extra stuff in the basic tests (unless you want to test both the new dimension with that flag on and off).

One of the reason for this is because then we loose the ability to actually have the `basic` tests run with all the fancy features turned off. For example in this case we will lose full suites running with embedding turned off completely (without relying on the other combination tests below that have embedding off but have other things enabled).

I understand that most tests shouldn't be effected by this and that this mostly applies to only the specific new embedding tests that are introduced, but imo we can still miss some unforeseen edge cases with other features that might fail when a feature (like embedding in this case) is off.

My suggestion is to leave the basic testing the way it was, in default state without any embedding tests (tests that might document embedding behavior usage when it is turned off are fine here of course), and introduce a new targeted test matrix that runs the full suite of tests with the embedding enabled.

All you would have to do is add this:

```yaml
jobs:
  # This job runs the vector embedding tests with different clients on linux.
  test-vector-embedding:
    name: Test job

    strategy:
      fail-fast: false
      matrix:
        client-type: [go, http, cli]

    runs-on: ubuntu-latest

    env:
      DEFRA_VECTOR_EMBEDDING: true
      DEFRA_CLIENT_GO: ${{ matrix.client-type == 'go' }}
      DEFRA_CLIENT_HTTP: ${{ matrix.client-type == 'http' }}
      DEFRA_CLIENT_CLI: ${{ matrix.client-type == 'cli' }}

    steps:
      - name: Checkout code into the directory
        uses: actions/checkout@v4

      - name: Setup defradb

      - name: Setup vector embedding dependencies
        run: make deps:embedding

      - name: Test coverage & save coverage report in an artifact
        uses: ./.github/composites/test-coverage-with-artifact
        with:
          coverage-artifact-name: "coverage_vector_embedding\
            _${{ matrix.client-type }}\
          "
          coverage-path: coverage.txt
```

This just need to add this new job under `upload-coverage` job as a dependency under `needs` attribute:

```yml
    needs:
      - test-basic            # 18 test(s)
      - test-os               #  1 test(s) [excluding windows]
      - test-acp              #  3 test(s)
      - test-lens             #  2 test(s)
      - test-view             #  1 test(s)
      - test-encryption       #  1 test(s)
      - test-vector-embedding #  3 test(s)
````


And would need to adjust this part in the new job if accepting my `Make` rule suggestions I made.

```yaml
      - name: Setup vector embedding dependencies
        run: make deps:embedding
```

tests/integration/mutation/create/embeddings/embedding_test.go
islamaliev, line 1: todo: please add a test with embedding referring to itself (i.e. it's own field) and a test with embedding referring to another field with embedding.

Whatever the behaviour should be, but I think these test cases are important to have.

tests/integration/utils.go
```diff
@@ -2331,6 +2338,21 @@ func skipIfNetworkTest(t testing.TB, actions []any) {
...
+	for _, act := range actions {
+		switch a := act.(type) {
+		case SchemaUpdate:
+			hasVectorEmbedding = strings.Contains(a.Schema, "@embedding")
+		}
+	}
+	if !runVectorEmbeddingTests && hasVectorEmbedding {
+		t.Skip("test involves vector embedding generation")
+	}
+}
```
shahzadlone, line 2354: todo: I dislike the automatic skipping based on detecting if it is a vector embedding test because this means we can not test the behavior that should happen when vector embedding is being used by the user but it was turned off (which as brought up by Kyle in the standup was an important behavior for us to ensure).

Instead I would like a usage like the `SupportedType` within the individual tests that specify that the tests are only testing the behavior for when "vector embedding is enabled" (Note: all your tests currently are only testing this behavior)

todo: Once above is done, I would add counter part tests to all (or most) embedding tests that were introduced in this PR and assert their behavior for when "vector embedding is turned off".

shahzadlone (commented):
Overall looks good, I have gone through most of the PR.

I left some comments, more notable is the testing request to be able to test and assert the behavior when vector embedding is turned off.

shahzadlone (commented):
Overall looks good, I have gone through most of the PR.

I left some comments, more notable is the testing request to be able to test and assert the behavior when vector embedding is turned off.

client/ctype.go
```diff
@@ -40,7 +40,7 @@ func (t CType) IsSupportedFieldCType() bool {
 func (t CType) IsCompatibleWith(kind FieldKind) bool {
 	switch t {
 	case PN_COUNTER, P_COUNTER:
-		if kind == FieldKind_NILLABLE_INT || kind == FieldKind_NILLABLE_FLOAT {
+		if kind == FieldKind_NILLABLE_INT || kind == FieldKind_NILLABLE_FLOAT64 || kind == FieldKind_NILLABLE_FLOAT32 {
```
shahzadlone, line 43: suggestion: An already untested behavior was changed, add a test if easy to do so.

client/document.go
```diff
@@ -274,14 +274,14 @@ func validateFieldSchema(val any, field FieldDefinition) (NormalValue, error) {
...
-		v, err := getArray(val, getString)
+		v, err := getArray(val, getString, field.Size)
 		if err != nil {
 			return nil, err
 		}
 		return NewNormalStringArray(v), nil
 
 	case FieldKind_NILLABLE_STRING_ARRAY:
-		v, err := getNillableArray(val, getString)
+		v, err := getNillableArray(val, getString, field.Size)
```
shahzadlone, line 284: thought: out of scope, can't believe these cases are untested, specially in a validation function, and all other cases in this function.

internal/db/collection_define.go
```diff
@@ -148,12 +152,27 @@ func (db *DB) patchCollection(
...
+		newDefinitions[i].Schema = col.Schema()
+		newDefinitions[i].Description = newColsByID[col.ID()]
+		updatedColsByID[col.ID()] = struct{}{}
+	}
+	// append new cols
+	for id, col := range newColsByID {
+		if _, ok := updatedColsByID[id]; ok {
+			continue
+		}
+		newDefinitions = append(newDefinitions, client.CollectionDefinition{Description: col})
```
shahzadlone, line 167: suggestion: Seems like there is no patch collection test that has the need to append new collection definitions, might be worth adding tests for if easy to do so.

internal/db/collection_define.go
```diff
@@ -148,12 +152,27 @@ func (db *DB) patchCollection(
...
+	}
 
-	err = db.validateCollectionChanges(ctx, cols, newColsByID)
+	err = db.validateCollectionChanges(ctx, existingDefinitions, newDefinitions)
 	if err != nil {
 		return err
 	}
 
+	txn := mustGetContextTxn(ctx)
```
shahzadlone, line 175: question: I would have expected this like to be hit by test, not sure why it is not. Maybe due to ci failures?

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+func (c *collection) setEmbedding(ctx context.Context, doc *client.Document, isCreate bool) error {
+	embeddingGenerated := false
+	for _, embedding := range c.Description().Embeddings {
+		vecValue, err := doc.GetValue(embedding.FieldName)
+		if err != nil && !errors.Is(err, client.ErrFieldNotExist) {
+			return NewErrGetEmbeddingField(err)
+		}
+		if vecValue != nil && vecValue.IsDirty() {
+			// If the vector has been explicitly set, no need to generate the embedding.
+			continue
```
shahzadlone, line 58: suggestion: This line shows untested, worth investigating why, if actually missing coverage, then add test please.

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+			if docField, ok := doc.Fields()[embedField]; ok {
+				if doc.Values()[docField].IsDirty() {
+					needsGeneration = true
+					fieldsVal[embedField] = doc.Values()[docField].NormalValue()
+				} else {
+					fieldDef, ok := c.def.GetFieldByName(embedField)
+					if !ok {
+						return NewErrEmbeddingFieldNotFound(embedField)
+					}
+					missingFieldsForGeneration = append(missingFieldsForGeneration, fieldDef)
```
shahzadlone, line 76: suggestion: This also seems like an important untested behavior to test if all the fields undefined are actually being tracked as expected.

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+				}
+			}
+		}
+
+		if !needsGeneration {
+			continue
```
shahzadlone, line 82: suggestion: Test if easy to test please.

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+				missingFieldsForGeneration,
+				false,
+			)
+			if err != nil {
+				return NewErrGetDocForEmbedding(err)
+			}
+			for _, embedField := range missingFieldsForGeneration {
+				if docField, ok := oldDoc.Fields()[embedField.Name]; ok {
+					fieldsVal[embedField.Name] = oldDoc.Values()[docField].NormalValue()
+				}
```
shahzadlone, line 100: suggestion: The whole block is showing as untested, please add tests if not there already (maybe due to ci failures, ignore if tested already).

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+			embedding.URL,
+		)
+		if err != nil {
+			return NewErrGetEmbeddingFunc(err)
+		}
+		var text strings.Builder
+		for _, fieldName := range embedding.Fields {
+			if val, ok := fieldsVal[fieldName]; ok {
+				text.WriteString(fmt.Sprintf("%v\n", val.Unwrap()))
+			}
```
shahzadlone, line 116: suggestion: The whole block is showing as untested, please add tests if not there already (maybe due to ci failures, ignore if tested already).

internal/db/embedding.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		embeddingVec, err := embeddingFunc(ctx, text.String())
+		if err != nil {
+			return err
+		}
+		err = doc.Set(embedding.FieldName, embeddingVec)
+		if err != nil {
+			return err
+		}
+		embeddingGenerated = true
```
shahzadlone, line 126: suggestion: The whole block is showing as untested, please add tests if not there already (maybe due to ci failures, ignore if tested already).

internal/db/fetcher/indexer_matchers.go
```diff
@@ -55,16 +55,34 @@ func (m *intMatcher) Match(value client.NormalValue) (bool, error) {
 	return false, NewErrUnexpectedTypeValue[int64](value)
 }
 
-type floatMatcher struct {
+type float32Matcher struct {
+	value    float32
+	evalFunc func(float32, float32) bool
+}
+
+func (m *float32Matcher) Match(value client.NormalValue) (bool, error) {
```
shahzadlone, line 63: suggestion: The whole function is showing as untested, please add tests if not there already (maybe due to ci failures, ignore if tested already).

client/schema_field_description.go
AndrewSisley: It is a shame Go does not allow the extension of interfaces in the same way a few other languages do such as C#. 

We would have to host `IsNumericArray` off of the object, but I agree with Islam in that this is still cleaner than expanding the interface here:
```
IsNumericArray(FieldKind) bool {
 ...
}
```

internal/db/collection_define.go
fredcarle: There should be as we have a test that tries to add a collection via patch. I'll investigate.

fredcarle: I can confirm that this is hit with intergration tests 👍

tests/integration/collection_description/updates/add/collections_test.go
fredcarle: Done

Makefile
fredcarle: Done

Makefile
fredcarle: Done

client/definitions.go
```diff
@@ -148,6 +148,12 @@ type FieldDefinition struct {
 
 	// DefaultValue contains the default value for this field.
 	DefaultValue any
+
+	// Size is a constraint that can be applied to fields that are arrays.
+	//
+	// Mutations on fields with a size constraint will fail if the size of the array
+	// does not match the constraint.
+	Size int
```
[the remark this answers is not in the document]
fredcarle: Interesting ideal :)

client/schema_field_description.go
fredcarle: I have an idea! Let's rewrite in `C#`! 😜

fredcarle: Done

.github/workflows/test-coverage.yml
```diff
@@ -42,6 +42,8 @@ env:
   DEFRA_ACP_TYPE: local
   DEFRA_VIEW_TYPE: cacheless
 
+  DEFRA_VECTOR_EMBEDDING: false
```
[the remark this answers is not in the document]
fredcarle: Done

.github/workflows/test-coverage.yml
fredcarle: As discussed in the standup, we will look ad changing the trigger for embedding tests in a subsequent PR.

tests/integration/utils.go
fredcarle: same as above, as discussed in the standup, we will look ad changing the trigger for embedding tests in a subsequent PR.

client/collection_description.go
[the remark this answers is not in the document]
fredcarle: Done

client/collection_description.go
```diff
@@ -268,3 +280,56 @@ func (c *CollectionDescription) UnmarshalJSON(bytes []byte) error {
 
 	return nil
 }
+
+// EmbeddingDescription hold the relevant information to generate embeddings.
+//
+// Embeddings are AI/ML specific vector representations of some content.
+// In the case of Defra, that content is one or multiple fields, optionaly added to a template.
```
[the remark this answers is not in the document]
fredcarle: Done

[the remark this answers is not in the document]
fredcarle: Done

internal/db/collection_define.go
fredcarle: Ci issue

internal/db/embedding.go
fredcarle: CI issue

internal/db/embedding.go
fredcarle: Done

tests/integration/mutation/create/embeddings/embedding_test.go
fredcarle: Done

client/ctype.go
fredcarle: This should be covered already. I'll mark it as out of scope for now. and will revisit coverage later

tests/integration/mutation/create/embeddings/embedding_test.go
islamaliev: I see you added them as schema patch tests. I think it makes sense to add them as part of initial schema as well.

Another point is that the patch actions are slightly harder to read.

AndrewSisley (approved):
LGTM, will be good to get this merged - thanks Fred, and thanks for all your explanations along the way :)

internal/db/embedding.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/keys"
+)
+
+const (
+	ollama string = "ollama"
+	openai string = "openai"
+)
+
+var supportedEmbeddingProviders = map[string]struct{}{
```
shahzadlone, line 31: nitpick: Using 'vector embedding' terminology over just 'embedding' here and other places in this file.

internal/db/definition_validation.go
```diff
@@ -763,6 +727,84 @@ func validateTypeAndKindCompatible(
 	return nil
 }
 
+func validateEmbeddingAndKindCompatible(
```
shahzadlone, line 730: nitpick: Using 'vector embedding' terminology over just 'embedding' here and other places in this file.

shahzadlone (approved):
Thanks for doing the requested suggestions :) great work overall!

LGTM

internal/db/embedding.go
fredcarle: Noted. I'll fix it in the _similarity PR coming soon.

islamaliev:
Bug bash result:
Launched ollama locally, ran different queries (create, update) embeddings work as expected

PR #3417, opened 2025-01-30
feat: Enable secondary index for compound filter conditions

### LINKED ISSUE #3299: Sec. index: make compound filters utillize sec. indexes

islamaliev:
## Relevant issue(s)

Resolves #3299

## Description

Utilize secondary indexes even when compound filter conditions are present.
For this to work new filter traversing utility function is introduced that can be configured to different needs.

And not that indexes are exposed to more complex conditions they started to produce more false positive docs that weren't checked by the filter because the index fetcher was not part of the new fetcher chain.

Make index fetcher implement new fetcher interface so that the documents it fetches can be checked against the scanner filter and permissions.

Change behavior of connor to recognize if a field exists. It's need to distinguish if _ne filter returns false because 2 values are different or becuase the document doesn't have the field.

Make fieldFetched explain metric count all fields fetched, not only fields that were requested.


shahzadlone:
question: Do you know the coverage takes a hit of -23% 😢? @islamaliev

internal/connor/key.go
```diff
@@ -1,13 +1,25 @@
...
 // in a filter.
 type FilterKey interface {
-	// PropertyAndOperator returns the data and operator that should be used
-	// to filter the value matching this key.
-	//
-	// If the key does not have an operator the given defaultOp will be returned.
-	PropertyAndOperator(data any, defaultOp string) (any, string, error)
+	// PropertyAndOperator returns [KeyResult] that contains data and operator that should be
+	// used to filter the value matching this key.
+	PropertyAndOperator(data any, defaultOp string) KeyResult
```
AndrewSisley, line 22: todo: Please return the `error` as a normal return parameter like it was before.  The gql-go lib holds `err` on a struct similar to what you are proposing here and it is really unpleasant to use,  readers do not always expect it, and the linter wont catch it when it is not handled.  We've had quite a few production bugs because of this pattern.

Please change this to:
```
PropertyAndOperator(data any, defaultOp string) (KeyResult, error)
```

internal/connor/eq.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // eq is an operator which performs object equality
 // tests.
-func eq(condition, data any) (bool, error) {
+func eq(condition, data any, propExists bool) (bool, error) {
```
AndrewSisley, line 16: todo: Please document what the `propExists` property is, and why anyone would want to call this function with a property that doesn't exist.

internal/connor/connor.go
```diff
@@ -49,13 +49,13 @@ func IsOpSimple(op string) bool {
...
 func Match(conditions map[FilterKey]any, data any) (bool, error) {
-	return eq(conditions, data)
+	return eq(conditions, data, true)
 }
 
 // matchWith can be used to specify the exact operator to use when performing
 // a match operation. This is primarily used when building custom operators or
 // if you wish to override the behavior of another operator.
-func matchWith(op string, conditions, data any) (bool, error) {
+func matchWith(op string, conditions, data any, propExists bool) (bool, error) {
```
AndrewSisley, line 58: todo: Please document what the `propExists` property is, and why anyone would want to call this function with a property that doesn't exist.

internal/db/fetcher/indexer.go
```diff
@@ -12,68 +12,68 @@ package fetcher
...
+	fieldsByID    map[uint32]client.FieldDefinition
 	indexDesc     client.IndexDescription
 	indexIter     indexIterator
-	execInfo      ExecInfo
+	currentDocID  immutable.Option[string]
+	execInfo      *ExecInfo
 }
 
-var _ Fetcher = (*IndexFetcher)(nil)
+// var _ Fetcher = (*IndexFetcher)(nil)
```
AndrewSisley, line 44: todo: Please remove the commented out line :)

internal/db/collection.go
```diff
@@ -71,7 +71,7 @@ func (c *collection) newFetcher() fetcher.Fetcher {
 	if c.fetcherFactory != nil {
 		innerFetcher = c.fetcherFactory()
 	} else {
-		innerFetcher = fetcher.NewDocumentFetcher()
+		innerFetcher = fetcher.NewDocumentFetcher(immutable.Option[client.IndexDescription]{})
```
AndrewSisley, line 74: suggestion: `immutable.None[client.IndexDescription]()` is more descriptive and expected than a default value, the current code looks like their is a value.

internal/db/fetcher/wrapper.go
```diff
@@ -45,8 +46,8 @@ type wrappingFetcher struct {
 
 var _ Fetcher = (*wrappingFetcher)(nil)
 
-func NewDocumentFetcher() Fetcher {
-	return &wrappingFetcher{}
+func NewDocumentFetcher(index immutable.Option[client.IndexDescription]) Fetcher {
```
AndrewSisley, line 49: todo: This is very much where I want all of the props currently passed into `Init` to be passed in, however the current code is not setup around this and `Init` is called multiple times per fetcher instance, potentially with properties that could in theory be different each time round.

If this is not currently a bug, it could easily become one if we do not fix how/where/when `Init` is called before someone changes something around there before the desired change (i.e. removal of) to `Fetcher.Init` is made.

I know it is miserable to do so, but please move this property into `Init` for the time being.  Or I guess, remove `Fetcher.Init` so we don't have this problem (please dont do this in this PR, please open and merge that change before merging this PR).

internal/db/fetcher/wrapper.go
```diff
@@ -148,31 +163,29 @@ func (f *wrappingFetcher) Start(ctx context.Context, prefixes ...keys.Walkable)
...
 func (f *wrappingFetcher) FetchNext(ctx context.Context) (EncodedDocument, ExecInfo, error) {
-	docID, err := f.fetcher.NextDoc()
-	if err != nil {
-		return nil, ExecInfo{}, err
-	}
-
-	if !docID.HasValue() {
-		execInfo := *f.execInfo
-		f.execInfo.Reset()
+	f.execInfo.Reset()
```
AndrewSisley, line 166: praise: This is a nice change, thanks :)

AndrewSisley (changes requested):
I've only reviewed the first few files and need to go and eat :)

Looks good so far,  just documentation requests.

internal/db/collection.go
islamaliev: removed

internal/db/fetcher/indexer.go
islamaliev: done

internal/connor/connor.go
islamaliev: done

internal/connor/eq.go
islamaliev: done

internal/connor/key.go
islamaliev: done

internal/db/fetcher/wrapper.go
islamaliev: moved to `Init()`

internal/connor/connor.go
```diff
@@ -49,13 +49,17 @@ func IsOpSimple(op string) bool {
...
 }
 
 // matchWith can be used to specify the exact operator to use when performing
 // a match operation. This is primarily used when building custom operators or
 // if you wish to override the behavior of another operator.
-func matchWith(op string, conditions, data any) (bool, error) {
+// It also takes a propExists boolean to indicate if the property exists in the data.
+// It's needed because the behavior of the operators can change if the property doesn't exist.
+// For example, _ne operator should return true if the property doesn't exist.
+// This can also be used in the future if we introduce operators line _has.
```
AndrewSisley, line 61: praise: This example is particularly useful - thanks Islam!

internal/db/fetcher/indexer.go
```diff
@@ -82,158 +81,64 @@ func (f *IndexFetcher) Init(
...
-				return nil, ExecInfo{}, err
-			}
-			f.execInfo.Add(execInfo)
-			if encDoc == nil {
-				continue
-			}
-			f.doc.MergeProperties(encDoc)
+	if f.indexDesc.Unique && !hasNilField {
+		f.currentDocID = immutable.Some(string(res.value))
+	} else {
```
AndrewSisley, line 108: question: This `else` block looks odd to me.  Why are you uncertain of the type?

todo: The bytes else-if block is untested. Please either remove it, or test it.

internal/db/fetcher/indexer_iterators.go
```diff
@@ -499,100 +485,109 @@ type fieldFilterCond struct {
...
-									// values picked by the index is random guessing. For example if we have doc1
-									// with array of [3, 5, 1] and doc2 with [7, 4, 8] the index first fetches
-									// value 1 of doc1, let it go through the filter and then fetches value 3 of doc1
-									// again, skips it (because it cached doc1 id) and fetches value 4 of doc2, and
-									// so on until it exhaust all prefixes in ascending order.
-									// It might be even less effective than just scanning all documents.
-									return nil, nil
+		filter.TraverseProperties(
+			f.indexFilter.Conditions,
+			func(prop *mapper.PropertyIndex, condMap map[connor.FilterKey]any) bool {
```
AndrewSisley, line 499: todo: IMO this is way too large a function to declare inline, I struggled to find where it ended.  Please make it a named function and add a line or so documenting what it is trying to do.

EDIT: I see that right at the very end of this very large inline function you mutate a variable (`found`) belonging to the encompassing scope.  I think this needs rework to allow you name this function, perhaps you will need to rename `TraverseProperties` and return a boolean from it, I am not sure, but please change this as it is a bit hard to follow IMO (especially due to the mutation of `found`).

internal/db/fetcher/indexer_iterators.go
```diff
@@ -499,100 +485,109 @@ type fieldFilterCond struct {
...
+								op, ok := key.(*mapper.Operator)
+								if ok && isArrayCondition(op.Operation) {
+									jsonPath = jsonPath.AppendIndex(0)
 								}
-								jsonPath = jsonPath.AppendIndex(0)
+								break jsonPathLoop
 							}
-							break jsonPathLoop
+							jsonPath = jsonPath.AppendProperty(prop.Name)
+							condMap = filterVal.(map[connor.FilterKey]any)
```
AndrewSisley, line 520: todo: Why is this cast safe (no `ok` check)?  Please add your answer as a code comment.

internal/db/fetcher/indexer_matchers.go
```diff
@@ -369,6 +353,78 @@ func createValueMatcher(condition *fieldFilterCond) (valueMatcher, error) {
...
 }
 
+func createComparingMatcher(condition *fieldFilterCond) valueMatcher {
+	// JSON type needs a special handling if the op is _ne, because _ne should check also
+	// difference of types
+	if v, ok := condition.val.JSON(); ok {
+		var matcher valueMatcher
+		// we have a separate branch for null matcher because default matching behavior
+		// is what we need: for filter `_ne: null` it will match all non-null values
+		if v.IsNull() {
```
AndrewSisley, line 363: suggestion: This would be a lot more readable if you returned early instead of the current if-else-if-else nesting.

for example:
```
if v.IsNull() {
    return &jsonNullMatcher{matchNull: condition.op == opEq}
}

if jsonVal, ok := v.Number(); ok {
    return &jsonComparingMatcher[float64]{
					value:        jsonVal,
					getValueFunc: func(j client.JSON) (float64, bool) { return j.Number() },
					evalFunc:     getCompareValsFunc[float64](condition.op),
				}
}

... etc
```

internal/db/fetcher/indexer_matchers.go
```diff
@@ -369,6 +353,78 @@ func createValueMatcher(condition *fieldFilterCond) (valueMatcher, error) {
...
+					}
+				}
+				matcher = newCompositeMatcher(typeMatcher, matcher)
+			}
+		}
+		return matcher
+	}
+
+	var matcher valueMatcher
+	if v, ok := condition.val.Int(); ok {
```
AndrewSisley, line 405: suggestion: Same as above, I think this would be easier to read if you returned early instead of the large if-else

AndrewSisley (changes requested):
Is looking good, I'm continuing my review but it is taking a while so I thought I'd submit the outstanding comments now.

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,139 @@
...
+//
+//	{
+//	    "author": {
+//	        "books": {
+//	            "title": {"_eq": "Sample"}
+//	        }
+//	    }
+//	}
+//
+// The callback would receive path=["author", "books", "title"] and value="Sample"
```
AndrewSisley, line 33: praise: This is excellent documentation and really helped me understand the functions here, thanks Islam :)

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,139 @@
...
+func TraverseFields(conditions map[string]any, f func([]string, any) bool) {
```
AndrewSisley, line 34: todo: The return value of this function is unused, and unexplained by the documentation.  Please remove it.

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,139 @@
...
+	traverseFields(nil, "", conditions, f)
+}
+
+func traverseFields(path []string, key string, value any, f func([]string, any) bool) bool {
+	isKeyOp := func(k string) bool { return len(k) > 0 && k[0] == '_' && k != request.DocIDFieldName }
+	canOpBeScalar := func(k string) bool {
+		switch k {
+		case request.AliasFieldName, request.FilterOpOr, request.FilterOpAnd, request.FilterOpNot:
```
AndrewSisley, line 42: suggestion: I am left wondering why alias fields cannot be scalar when I read this, I guess the alias targeting syntax is different to regular fields, but I don't remember it being so, and wound not want to rely on that guess when debugging.

I suggest adding a comment here.

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,139 @@
...
+			return false
+		}
+		return true
+	}
+	switch t := value.(type) {
+	case map[string]any:
+		for k, v := range t {
+			if !isKeyOp(k) {
```
AndrewSisley, line 50: suggestion:  There seems to be little benefit in inverting the if-else by negating `isKeyOp(k)`, if you remove the `!` it will reduce the cognitive load on the reader slightly.

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,139 @@
...
+				newPath := make([]string, len(path), len(path)+1)
+				copy(newPath, path)
+				newPath = append(newPath, k)
+				if !traverseFields(newPath, k, v, f) {
```
AndrewSisley, line 54: suggestion: `k` and `v` are short lived, but here, combined with the longer scope `f` the shortened variables read a little like algebra.  I suggest renaming them to `key` and `value`.

Same goes for the other similar areas in this function.

internal/planner/scan.go
```diff
@@ -152,46 +153,13 @@ func (n *scanNode) addField(field client.FieldDefinition) {
...
-	index immutable.Option[client.IndexDescription],
-) {
+func (scan *scanNode) initFetcher(cid immutable.Option[string]) {
 	var f fetcher.Fetcher
 	if cid.HasValue() {
 		f = new(fetcher.VersionedFetcher)
 	} else {
 		f = fetcher.NewDocumentFetcher()
 
-		if index.HasValue() {
```
AndrewSisley, line 165: praise: I was worried you'd have to put this back when you moved the index prop to init in the fetcher.  Thank you very much for cleaning this up.

internal/planner/select.go
```diff
@@ -308,17 +311,37 @@ func findIndexByFilteringField(scanNode *scanNode) immutable.Option[client.Index
...
+			}
 		}
+		return true
+	})
+	if len(indexCandidates) == 0 {
+		return immutable.Option[client.IndexDescription]{}
 	}
-	return immutable.None[client.IndexDescription]()
+	slices.SortFunc(indexCandidates, func(a, b client.IndexDescription) int {
+		switch {
```
AndrewSisley, line 333: todo: This looks like it is a re-implementation of `strings.Compare`, please use `strings.Compare` instead, or document why you are not using it, and what the differences between this and that function are.

internal/planner/select.go
```diff
@@ -308,17 +311,37 @@ func findIndexByFilteringField(scanNode *scanNode) immutable.Option[client.Index
...
+			indexes := colDesc.GetIndexesOnField(field.Name)
+			if len(indexes) > 0 {
+				indexCandidates = append(indexCandidates, indexes...)
+				return true
+			}
 		}
+		return true
+	})
+	if len(indexCandidates) == 0 {
+		return immutable.Option[client.IndexDescription]{}
```
AndrewSisley, line 330: suggestion: I think `None` is more descriptive than the default, which kind of looks like it has (or might have) a value.

AndrewSisley (changes requested):
Looks good Islam!  I'm nearly done reviewing but really need to eat :)

tests/integration/index/array_test.go
```diff
@@ -167,7 +167,7 @@ func TestArrayIndex_WithFilterOnIndexedArrayUsingNone_ShouldUseIndex(t *testing.
 			},
 			testUtils.Request{
 				Request:  makeExplainQuery(req),
-				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(9),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(0),
```
AndrewSisley, line 170: question: This is surprising, why has this changed? It looks like it is no longer using the index.

todo: If the change is correct, please update the name and/or document the test, as it doesnt make any sense to me atm.

tests/integration/index/json_test.go
```diff
@@ -1301,12 +1305,146 @@ func TestJSONIndex_WithCompoundFilterCondition_ShouldUseIndex(t *testing.T) {
...
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(2),
 			},
 		},
 	}
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestJSONIndex_WithNeFilterAgainstNonNullValue_ShouldFetchNullValues(t *testing.T) {
+	type testCase struct {
```
AndrewSisley, line 1318: suggestion: I am really not a fan of bundling multiple tests into the same test.  It makes debugging much harder and typically involves the repeated commenting and uncommenting of test cases in order to deal with one aspect at a time.

Please break this up.

tests/integration/index/query_with_index_combined_filter_test.go
```diff
@@ -48,7 +48,7 @@ func TestQueryWithIndex_IfIndexFilterWithRegular_ShouldFilter(t *testing.T) {
 			},
 			testUtils.Request{
 				Request:  makeExplainQuery(req),
-				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(3).WithIndexFetches(3),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(3),
```
AndrewSisley, line 51: question: Why have the field fetches here been removed?

tests/integration/index/query_with_index_only_filter_test.go
```diff
@@ -49,7 +49,7 @@ func TestQueryWithIndex_WithNonIndexedFields_ShouldFetchAllOfThem(t *testing.T)
 			},
 			testUtils.Request{
 				Request:  makeExplainQuery(req),
-				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(1).WithIndexFetches(1),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(1),
```
AndrewSisley, line 52: question: Regarding all the tests that had non-zero field fetches removed, why have the field fetches been removed?

tests/integration/index/query_with_index_only_filter_test.go
```diff
@@ -804,3 +804,213 @@ func TestQueryWithIndex_WithFilterOn2Relations_ShouldFilter(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQueryWithIndex_WithNeFilterAgainstNonNilValue_ShouldFetchNilValues(t *testing.T) {
+	type testCase struct {
```
AndrewSisley, line 809: suggestion: Same as other similar comment, I have a preference for this to be broken up.

tests/integration/query/json/with_ne_test.go
```diff
@@ -163,3 +163,129 @@ func TestQueryJSON_WithNotEqualFilterWithNullValue_ShouldFilter(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQueryJSON_WithNeFilterAgainstNonNullValue_ShouldFetchNullValues(t *testing.T) {
+	type testCase struct {
```
AndrewSisley, line 168: suggestion: Same suggestion as other comments, I have a preference for this to be broken up - it is multiple tests pretending to be a single test and this creates problems for anyone using it.

AndrewSisley (changes requested):
Reviewed! Overall it looks really good Islam, the provided documentation was very useful when reading the code.  I think all my requests are/were fairly localised, hopefully they all make sense to you :)

Thanks for resolving the issue.

tests/integration/query/json/with_ne_test.go
islamaliev: done

tests/integration/index/query_with_index_only_filter_test.go
islamaliev: done

tests/integration/index/query_with_index_only_filter_test.go
islamaliev: there is no point in having all index-related tests assert on fields fetched. The main indicator for indexes in index fetches.

The behaviour of field fetches did change few times in the past and might slightly change in the future again.
This usually had a huge impact on all index test. Which ideally it shouldn't.

I still left `WithFieldFetches` for some tests to make sure it works, but, as I said, having all index-related tests assert it doesn't make sense.

tests/integration/index/query_with_index_combined_filter_test.go
islamaliev: answered here https://github.com/sourcenetwork/defradb/pull/3417#discussion_r1942477505

tests/integration/index/json_test.go
islamaliev: done

tests/integration/index/array_test.go
islamaliev: how can this be a surprise to you? We had quite [a discussion](https://github.com/sourcenetwork/defradb/pull/3368#discussion_r1919323385) just recently :) 

But the name of the test was incorrect. Thanks.

internal/planner/type_join.go
```diff
@@ -721,28 +742,33 @@ func (join *invertibleTypeJoin) nextJoinedSecondaryDoc() (bool, error) {
 		join.encounteredDocIDs = append(join.encounteredDocIDs, secondaryDocID)
 	}
 
-	hasDoc, err := fetchDocWithID(secondSide.plan, secondaryDocID)
+	//secondaryDocOpt, err := fetchDocWithID(secondSide.plan, secondaryDocID)
```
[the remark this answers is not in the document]
islamaliev: done

internal/planner/type_join.go
```diff
@@ -372,12 +372,17 @@ func (p *Planner) newInvertableTypeJoin(
...
+	join := invertibleTypeJoin{
 		docMapper:  docMapper{parent.documentMapping},
 		parentSide: parentSide,
 		childSide:  childSide,
 		skipChild:  skipChild,
-	}, nil
+	}
+
+	// we store child's own filter in case an index kicks in and replaces it with it's own filter
+	join.subFilter = getScanNode(childSide.plan).filter
```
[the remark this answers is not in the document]
islamaliev: done

internal/planner/select.go
islamaliev: done

internal/planner/select.go
islamaliev: done

internal/planner/filter/traverse.go
islamaliev: done

internal/planner/filter/traverse.go
islamaliev: Renamed and added a comment

internal/db/fetcher/wrapper.go
```diff
@@ -117,17 +120,31 @@ func (f *wrappingFetcher) Start(ctx context.Context, prefixes ...keys.Walkable)
...
-	top, err = newPrefixFetcher(ctx, f.txn, dsPrefixes, f.col, fieldsByID, client.Active, &execInfo)
-	if err != nil {
-		return err
+	if f.index.HasValue() {
+		indexFetcher, err := newIndexFetcher(ctx, f.txn, fieldsByID, f.index.Value(), f.filter, f.col,
+			f.docMapper, &f.execInfo)
+		if err != nil {
+			return err
+		}
+		if indexFetcher != nil {
```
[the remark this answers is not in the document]
islamaliev: added more docs to the func

internal/db/fetcher/indexer_matchers.go
islamaliev: the reason why it can't be early-returned can be seen below: the matcher can be wrapped into another matcher.

The json null matcher can be early-returned, but it would be the only place the is returned this way in this function and will make it look inconsistent.

I don't see any problem here with debugging.

I could extract some of the blocks in this function into smaller functions which will make it more readable and this is what I usually prefer to do, but considering that majority of the team prefers to have larger functions I decided to write it like this.

Leaving it as is, unless you'd prefer to have 2 more smaller functions.

internal/db/fetcher/indexer_matchers.go
islamaliev: the same answer https://github.com/sourcenetwork/defradb/pull/3417#discussion_r1942643627

internal/db/fetcher/indexer_iterators.go
islamaliev: I simplified the closure by extracting some blocks into another functions.
Also added some comments to explain a few decisions.
Removed `found` as `result` is enough to carry this information.

Let me know if if looks better know.

I don't shouldn't need to change `TraverseProperties` as it does what it should do.

internal/db/fetcher/indexer.go
islamaliev: in theory docID can be represented as bytes as well, but we encode it as string. So, you are right, it can be removed.
I put it there just in case.

tests/integration/index/query_with_index_only_filter_test.go
AndrewSisley: Makes sense, thanks for the explanation Islam :)

internal/db/fetcher/indexer_matchers.go
```diff
@@ -369,6 +353,78 @@ func createValueMatcher(condition *fieldFilterCond) (valueMatcher, error) {
...
+	} else if v, ok := condition.val.Float(); ok {
+		matcher = &floatMatcher{value: v, evalFunc: getCompareValsFunc[float64](condition.op)}
+	} else if v, ok := condition.val.String(); ok {
+		matcher = &stringMatcher{value: v, evalFunc: getCompareValsFunc[string](condition.op)}
+	} else if v, ok := condition.val.Time(); ok {
+		matcher = &timeMatcher{value: v, op: condition.op}
+	} else if v, ok := condition.val.Bool(); ok {
+		matcher = &boolMatcher{value: v, isEq: condition.op == opEq}
+	} else {
+		return nil
```
AndrewSisley, line 416: question: This looks like it should be an error instead of nil, and code-cov says it is untested, is it dead code?  Can it be removed?

suggestion: If keeping it, I suggest making it an error, lest somewhere else nil-pointer-panics should it ever be reached.

internal/db/fetcher/indexer_matchers.go
AndrewSisley: > I don't see any problem here with debugging.

When I debug an issue (not with a literal code-executing-debugger-tool - I almost never use them), I read the code, if I come across a variable of interest and want to track it over it's lifetime my job is made much easier by minimising the scope of variables.  

Here, on any given execution, matcher is only ever interacted with twice I think - once on assignment, and once consumed by `condition.op == opNe`.

The number and complexity of the if-else blocks in here makes following this path fairly error prone (e.g. I made a mistake originally as you pointed out), and it takes a lot more effort than were it returned early.

Splitting into multiple functions is not free, and can result in the reader having to bounce around through the file, forgetting bits of context etc. however in this case, given the perceived possible scope reduction I would likely try and see what it looks like were I the author.

Is just a suggestion though, the code is readable in it's current state, I think it could be made even easier though and less error prone.

internal/db/fetcher/indexer_iterators.go
AndrewSisley: Looks much better, thanks Islam :)

AndrewSisley (approved):
LGTM, thanks Islam!  There are a couple of open questions, and a couple of suggestions that I would appreciate resolution before merge, but none are blocking.

Code looks really good, and I am glad we are able to close this issue :)

tests/integration/index/array_test.go
islamaliev: The previous PR ignored index when dealing with `_none` for json arrays. This time it's done for all arrays.

internal/db/fetcher/indexer_matchers.go
islamaliev: removed `else` as it can never get there

internal/db/fetcher/indexer_matchers.go
shahzadlone: suggestion: +1 for less nesting and returning early

internal/planner/type_join.go
```diff
@@ -769,36 +801,35 @@ func (join *invertibleTypeJoin) invertJoinDirectionWithIndex(
...
 ) error {
-	p := join.childSide.plan
-	s := getScanNode(p)
-	s.tryAddFieldWithName(join.childSide.relFieldDef.Value().Name + request.RelatedObjectID)
-	s.filter = fieldFilter
-	s.initFetcher(immutable.Option[string]{}, immutable.Some(index))
+	childScan := getScanNode(join.childSide.plan)
+	childScan.tryAddFieldWithName(join.childSide.relFieldDef.Value().Name + request.RelatedObjectID)
+	// replace child's filter with the filter that utilizes the index
+	// the original child's filter is stored in join.subFilter
```
shahzadlone, line 807: praise: Thanks for these comments

tests/integration/index/array_test.go
shahzadlone: Thanks for explaining again islam, to be fair I had the same confusion as Andy (and went through that previous conversation you guys had about `none` before as well).

nitpick: Perhaps add a comment here documenting again 1) that if it is `_none it doesn't make sense to use index`, 2) Why that is?

tests/integration/index/query_with_composite_index_only_filter_test.go
```diff
@@ -705,7 +705,7 @@ func TestQueryWithCompositeIndex_IfFirstFieldIsNotInFilter_ShouldNotUseIndex(t *
 							name
 						}
 					}`,
-				Asserter: testUtils.NewExplainAsserter().WithFieldFetches(20).WithIndexFetches(0),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(0),
```
shahzadlone, line 708: question: so before it was incorrectly using indexes?

shahzadlone (approved):
Great work, and appreciate the tests added. Also thanks for fixing the `fieldFetches` behavior in all the tests. Just left some suggestions and questions nothing blocking :)

tests/integration/index/array_test.go
islamaliev: ok

tests/integration/index/query_with_composite_index_only_filter_test.go
islamaliev: I would say expectedly, but inefficiently. It doesn't make sense to use index fetcher for composite index if the first field is not given ("age" is the second field)

I would add a comment, but the test name and description say it already.

internal/planner/type_join.go
```diff
@@ -470,6 +474,8 @@ type invertibleTypeJoin struct {
 	parentSide joinSide
 	childSide  joinSide
 
+	subFilter *mapper.Filter
```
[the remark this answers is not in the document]
islamaliev: done

internal/planner/filter/traverse.go
```diff
@@ -0,0 +1,140 @@
...
+					return false
+				}
+			} else {
+				newPath := make([]string, len(path), len(path)+1)
+				copy(newPath, path)
+				newPath = append(newPath, key)
+				if !traverseFields(newPath, key, value, f) {
+					return false
+				}
+			}
```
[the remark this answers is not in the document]
shahzadlone: > these are not equal. Ignoring

How so?

AndrewSisley:
Bug bash result: 
- Unrelated bug found during testing: https://github.com/sourcenetwork/defradb/issues/3469
- No related bugs found

AndrewSisley:
Bug bash result: No bugs found

PR #3446, opened 2025-02-13
chore(i): Bump golangci-lint to v1.64

### LINKED ISSUE #3445: Bump Linter to v1.64 and fix warnings
Bump Linter to v1.64 and fix warnings

shahzadlone:
## Relevant issue(s)
Resolves #3445

## Description
- Bump golangci-lint
- Fix deprecated warnings


tests/integration/results.go
```diff
@@ -123,8 +123,12 @@ func (ucid *UniqueCid) Validate(s *state, actualValue any, msgAndArgs ...any) {
 
 	if isNew {
 		require.IsType(s.t, "", actualValue)
+		value, ok := actualValue.(string)
+		if !ok {
+			require.Fail(s.t, "UniqueCid actualValue string cast failed")
```
AndrewSisley, line 128: suggestion: I think I prefer a `nolint` to this, this is dead code and makes reading the func more complicated than it needs to be

AndrewSisley: And if you really really don't want a no lint, please remove the `require.IsType` call on the line above

AndrewSisley (approved):
LGTM, and I have no problem with this sneaking in the release (no prod code changes).  Minor suggestion for you first if you dont mind looking before merge :)

tests/integration/results.go
shahzadlone: One can argue reading any go `if err != nil {` makes the code unreadable (which i agree it does, but we don't stop using it). I don't like the linter suppressing suggestion keeping as is.

- [x] Remove `require.IsType`

PR #3464, opened 2025-02-19
feat: Add cosine similarity query

### LINKED ISSUE #3349: Compute similarity between two vectors

fredcarle:
## Relevant issue(s)

Resolves #3349 

## Description

This PR adds the possibility to calculate the cosine similarity between a vector field and a given vector.

To achieve this we added the `_similarity` system field which take a target field (part of the parent object) and vector as parameter.
```gql
query {
  User{
    _similarity(pointsList: {vector: [1, 2, 0]})
  }
}
```

Note that the added code to mapper and planner is more of a "bolt on" addition given the current state of that part of the code base. A refactor is expected in the future.

Future work will allow giving a `content` parameter instead of the `vector` if the target field has embedding generation configured. This will enable out-of-the-box RAG queries.

## Tasks

## How has this been tested?

Added cosine similarity to integration testing

- MacOS

### DIFF
--- a/client/request/similarity.go
+++ b/client/request/similarity.go
@@ -1,0 +1,26 @@
+// Copyright 2025 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package request
+
+// Similarity is a functional field that defines the
+// parameters to calculate the cosine similarity between two vectors.
+type Similarity struct {
+	Field
+	// Vector contains the vector to compare the target field to.
+	//
+	// It will be of type Int, Float32 or Float64.
+	Vector any
+
+	// Target is the field in the host object that we will compare the the vector to.
+	//
+	// It must be a field of type Int, Float32 or Float64.
+	Target string
+}
--- a/client/request/similarity.go
+++ b/client/request/similarity.go
@@ -16,11 +16,11 @@ type Similarity struct {
 	Field
 	// Vector contains the vector to compare the target field to.
 	//
-	// It will be of type Int, Float32 or Float64.
+	// It will be of type Int, Float32 or Float64. It must be the same type and length as Target.
 	Vector any
 
 	// Target is the field in the host object that we will compare the the vector to.
 	//
-	// It must be a field of type Int, Float32 or Float64.
+	// It must be a field of type Int, Float32 or Float64. It must be the same type and length as Vector.
 	Target string
 }
--- a/internal/planner/mapper/similarity.go
+++ b/internal/planner/mapper/similarity.go
@@ -24,15 +24,3 @@ type Similarity struct {
 	// The vector to compare the target field to.
 	Vector any
 }
-
-func (a *Similarity) CloneTo(index int) Requestable {
-	return a.cloneTo(index)
-}
-
-func (a *Similarity) cloneTo(index int) *Similarity {
-	return &Similarity{
-		Field:            *a.Field.cloneTo(index),
-		DocumentMapping:  a.DocumentMapping,
-		SimilarityTarget: a.SimilarityTarget,
-	}
-}
--- a/internal/planner/mapper/similarity.go
+++ b/internal/planner/mapper/similarity.go
@@ -1,0 +1,38 @@
+// Copyright 2025 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package mapper
+
+import "github.com/sourcenetwork/defradb/internal/core"
+
+// Similarity represents an cosine similarity operation definition.
+type Similarity struct {
+	Field
+	// The mapping of this aggregate's parent/host.
+	*core.DocumentMapping
+
+	// The targetted field for the cosine similarity
+	SimilarityTarget Targetable
+
+	// The vector to compare the target field to.
+	Vector any
+}
+
+func (a *Similarity) CloneTo(index int) Requestable {
+	return a.cloneTo(index)
+}
+
+func (a *Similarity) cloneTo(index int) *Similarity {
+	return &Similarity{
+		Field:            *a.Field.cloneTo(index),
+		DocumentMapping:  a.DocumentMapping,
+		SimilarityTarget: a.SimilarityTarget,
+	}
+}
--- a/internal/planner/select.go
+++ b/internal/planner/select.go
@@ -61,6 +61,8 @@ type selectTopNode struct {
 	// selectNode is used pre-wiring of the plan (before expansion and all).
 	selectNode *selectNode
 
+	// This is added temporarity until Planner is refactored
+	// https://github.com/sourcenetwork/defradb/issues/3467
 	similarity []*similarityNode
 
 	// plan is the top of the plan graph (the wired and finalized plan graph).
--- a/internal/planner/select.go
+++ b/internal/planner/select.go
@@ -61,6 +61,8 @@ type selectTopNode struct {
 	// selectNode is used pre-wiring of the plan (before expansion and all).
 	selectNode *selectNode
 
+	similarity []*similarityNode
+
 	// plan is the top of the plan graph (the wired and finalized plan graph).
 	planNode planNode
 }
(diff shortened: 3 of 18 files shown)


client/request/similarity.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package request
+
+// Similarity is a functional field that defines the
+// parameters to calculate the cosine similarity between two vectors.
```
AndrewSisley, line 14: suggestion: It might be good to note that they must be the same length.

internal/planner/select.go
```diff
@@ -61,6 +61,8 @@ type selectTopNode struct {
 	// selectNode is used pre-wiring of the plan (before expansion and all).
 	selectNode *selectNode
 
+	similarity []*similarityNode
```
AndrewSisley, line 64: todo: Please add a comment (preferably linked to a follow-up ticket), noting what this is, and immediate future plans.

internal/request/graphql/parser/query.go
```diff
@@ -223,6 +223,31 @@ func parseAggregate(
...
+	field *ast.Field,
+) (*request.Similarity, error) {
+	fieldDef := gql.GetFieldDef(exe.Schema, parent, field.Name.Value)
+	arguments := gql.GetArgumentValues(fieldDef.Args, field.Arguments, exe.VariableValues)
+	var target string
+	var vector any
+	for _, argument := range field.Arguments {
+		target = argument.Name.Value
+		v := arguments[target].(map[string]any)
+		vector = v["vector"]
```
AndrewSisley, line 238: todo: Please create a const for `"vector"` in the correct file (IIRC there is one in the graphql package or one of it's child-packages)

internal/request/graphql/schema/generate.go
```diff
@@ -846,6 +856,40 @@ func (g *Generator) genAverageFieldConfig(obj *gql.Object) (gql.Field, error) {
 	return field, nil
 }
 
+func (g *Generator) genSimilarityFieldConfig(obj *gql.Object) (gql.Field, error) {
+	field := gql.Field{
+		Name:        request.SimilarityFieldName,
+		Description: "Returns the cosine similarity between the specified field and the provided vector.",
```
islamaliev, line 862: suggestion: I think the description should be defined in `schemaTypes` package

internal/request/graphql/schema/generate.go
```diff
@@ -846,6 +856,40 @@ func (g *Generator) genAverageFieldConfig(obj *gql.Object) (gql.Field, error) {
...
+			continue
+		}
+
+		inputObject := gql.NewInputObject(gql.InputObjectConfig{
+			Name:        genSimilaritySelectorName(obj.Name(), objectField.Name),
+			Description: objectField.Description,
+			Fields: gql.InputObjectConfigFieldMap{
+				"vector": &gql.InputObjectFieldConfig{
+					Type:        gql.NewNonNull(gql.NewList(listType.OfType)),
+					Description: "A vector of the same type as the field to compute the cosine similarity with.",
```
islamaliev, line 879: suggestion: I think this also should be extracted.

tests/integration/query/simple/with_similarity_test.go
```diff
@@ -0,0 +1,382 @@
...
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents a bug where having two aliases in a logical _or operator
+// return no results even though in the tests bellow 1 should be returned.
```
islamaliev, line 337: question: is there a ticket to tackle the bug? If yes, I would suggest to link it here.

tests/integration/query/simple/with_similarity_test.go
```diff
@@ -0,0 +1,382 @@
...
+					name: String
+					pointsList: [Int!]
+				}`,
+			},
+			testUtils.CreateDoc{
+				DocMap: map[string]any{
+					"name":       "John",
+					"pointsList": []int64{2, 4, 1},
+				},
+			},
```
islamaliev, line 131: todo: all your tests have well-defined types, but I think it's also important with `[]any`.

I'm not 100% sure, but `[]any` should be the type if instead of `DocMap` we use `Doc`:
```go
testUtils.CreateDoc{
	Doc: `{
		"name":       "John",
		"pointsList": [2, 4, 1],
	}`,
},
```

tests/integration/query/simple/with_similarity_test.go
islamaliev, line 1: todo: please add a test that tries to use similarity on non-numeric array, like `[String!]`

islamaliev (commented):
Overall looks good (ignoring the hacky part).

I have mostly doc suggestions and few test requests.

client/request/similarity.go
fredcarle: done

internal/request/graphql/parser/query.go
fredcarle: done

internal/planner/mapper/similarity.go
```diff
@@ -0,0 +1,38 @@
...
+	*core.DocumentMapping
+
+	// The targetted field for the cosine similarity
+	SimilarityTarget Targetable
+
+	// The vector to compare the target field to.
+	Vector any
+}
+
+func (a *Similarity) CloneTo(index int) Requestable {
```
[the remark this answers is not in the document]
fredcarle: These might not have been useful. They were copy pasted. I removed them and the tests all seem to pass without issue so it would have been dead code.

internal/request/graphql/schema/generate.go
fredcarle: We discussed not long ago that the consensus for description like these was to have them inline and that we would actually work towards returning the extracted ones back to inline.

internal/request/graphql/schema/generate.go
fredcarle: Same as above :)

islamaliev (approved):
Looks great. Thanks for quick adjustments

tests/integration/query/simple/with_similarity_test.go
fredcarle: Added an issue and linked it here :)

tests/integration/query/simple/with_similarity_test.go
fredcarle: done. Thanks for flagging the different length one. I had forgotten about ensuring they were the same length.

tests/integration/schema/default_fields.go
```diff
@@ -149,6 +150,14 @@ var aggregateFields = fields{
...
 }
 
+var similarityField = Field{
+	"name": "_similarity",
+	"type": map[string]any{
+		"kind": "SCALAR",
+		"name": "Float",
+	},
+}
+
```
fredcarle, line 160: > todo: Please make sure you add an introspection test(s) to make sure the new input type shows up in the correct shape in the correct place without breaking the GQL spec.

@AndrewSisley I added this here to ensure it's part of introspection test. Some test were actually failing because it was missing. But otherwise I don't know how I could meaningfully test it further. If you have any suggestions I'm happy to add them.

AndrewSisley: I actually forgot about the field, I am more concerned about it's params.

You need to test that the InputArgs are properly formed, and only present on fields of the relevant types.

The test directory is a bit of a mess, but `TestFilterForSimpleSchema` is an example of this for filters, `TestSchemaAggregateSimpleCreatesUsersAverage` does the same for `_avg`.

fredcarle: Let me know what you think of the added introspection tests.

tests/integration/schema/similarity_test.go
```diff
@@ -0,0 +1,322 @@
...
+										"type": map[string]any{
+											"inputFields": []any{
+												map[string]any{
+													"name": "vector",
+													"type": map[string]any{
+														"kind": "NON_NULL",
+														"name": any(nil),
+														"ofType": map[string]any{
+															"kind": "LIST",
+															"name": any(nil),
```
AndrewSisley, line 83: suggestion: `"name": nil,` should work fine

tests/integration/schema/default_fields.go
AndrewSisley: They look good, thanks Fred :)

tests/integration/schema/similarity_test.go
fredcarle: Indeed. Changed.

shahzadlone:
Bug bash + PR Review: I am leaving/ moving this back to the bug bash Todo because it can probably benefit from being tested by one more person, my testing might have missed some other edge cases:

However I noticed a minor bug that was introduced in the debug explain testing framework: https://github.com/sourcenetwork/defradb/issues/3471

Should be very easy to fix and add a debug explain test for it, explained it all in the issue.

islamaliev:
Bug bash result:
After manually updating the embedding field for one of the document I got this message:
"source and vector must be of the same length. Source: 3, Vector: 768"
This is expected, but if there are many document, it might be hard to find one, especially considering that we don't have a way of querying docs with array of certain length. So I would suggest to also return the doc's docID.

After deleting the document querying works as normal.

PR #3466, opened 2025-02-19
fix: Make --no-encryption flag work without keyring

### LINKED ISSUE #3465: Flag --no-encryption is ignored if keyring is enabled

islamaliev:
## Relevant issue(s)

Resolves #3465

## Description

Make --no-encryption flag work without keyring


islamaliev:
> LGTM - assuming you have manually tested this - if not, please do :)

yes, I tested it

Issue #3470, opened 2025-02-20
Updating a document via cli command doesn't work


islamaliev:
This simple command
```
$ defradb client collection update --name User --docID bae-07014aa9-7777-52a3-ba9e-505a8028a14e --updater
'{ "about": "Blah" }'
```
Always gives me 
```
Error: unexpected end of JSON input
```

shahzadlone:
This is odd, might be worth checking when this started happening, I am pretty sure I have been using cli to test commands like a couple months back which worked fine:

```go
defradb client collection update --name Users --docID "bae-a5830219-b8e7-5791-9836-2e494816fc0a" --updater '{ "name": "Update" }' --identity abcxyz
```

PR #3474, opened 2025-02-20
fix: Updating of document via cli

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/3470

## Description

Make `client collection update` always expect `--updater` parameter.
Change filter expectation from `{ "_gte": { "points": 100 } }` to `'{ "points": { "_gte": 100 } }`
Make db expect filter param as `map[string]any`.


tests/integration/collection/update/simple/with_filter_test.go
```diff
@@ -13,16 +13,41 @@ package update
...
 func TestUpdateWithInvalidFilterType_ReturnsError(t *testing.T) {
+	type invalidFilterType struct{ Number int }
 	test := testUtils.TestCase{
 		Description: "Test update users with invalid filter type",
+		// http and cli clients will pass the serialize filter into json which will result in
+		// the payload deserialized into map[string]any. With Go client the filter is passed as is.
+		SupportedClientTypes: immutable.Some(
+			[]testUtils.ClientType{testUtils.HTTPClientType, testUtils.CLIClientType}),
+		Actions: []any{
+			testUtils.UpdateWithFilter{
```
AndrewSisley, line 30: praise: I forgot we had this, thanks for using it and not doing whatever I suggested in the standup :grin:  Thanks for the comment too, it explains it nicely.

AndrewSisley (approved):
LGTM, thanks Islam!

PR #3508, opened 2025-03-03
refactor: Replace datastore package with corekv

### LINKED ISSUE #1813: Replace ds.Query interface with lower-level Iterator interface

AndrewSisley:
## Relevant issue(s)

Resolves #1813

## Description

Replaces the defra-repo datastore package with the corekv repo.

Switches some unit tests from the memory store to the badger-memory store, as the memory store is very deadlock prone and sorting that out is not worth blocking this PR over (high effort, current btree implementation would need replacing).

I think we can remove the Defra `datastore` package completely, there is not much left of it in this PR, although I want limit the scope of this PR to changing only what is necessary to bring in corekv.


tests/bench/storage/utils.go
```diff
@@ -200,89 +198,31 @@ func runStorageBenchPut(
 	return nil
 }
 
-func runStorageBenchPutMany(
```
fredcarle, line 203: question: What was wrong with this benchmark?

fredcarle (commented):
I'm happy to see this finally in the repo. Most of the changes are straight forward. What I'm wondering is the removal of some of the tests as you'll see from the questions bellow. Approving since they are not blockers but I just want to make sure they haven't been removed just to simplify the work for this PR.

tests/bench/storage/utils.go
AndrewSisley: corekv does not support Batching, it was benchmarking something that no longer exists (and wouldn't live in this repo anyway).

internal/db/index_test.go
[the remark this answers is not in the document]
AndrewSisley: I removed the unit tests that were deemed too painful to get compiling/running again. In my eyes the cost of maintaining them had exceeded the costs saved by having them.  Same with the other file with mock based tests removed.

internal/db/collection_index.go
```diff
@@ -422,34 +422,14 @@ func (c *collection) dropIndex(ctx context.Context, indexName string) error {
...
-	err = txn.Systemstore().Delete(ctx, key.ToDS())
+	err = txn.Systemstore().Delete(ctx, key.Bytes())
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (c *collection) dropAllIndexes(ctx context.Context) error {
```
shahzadlone, line 433: Question: This is no longer needed? What replaces this?

AndrewSisley: It is no longer needed (private function not called anywhere).

internal/db/index_test.go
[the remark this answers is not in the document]
fredcarle: I worry that moves like this could be detrimental in the long run. Have you discussed with the author of those tests to see if they took issue with removing them to make sure you're not missing anything?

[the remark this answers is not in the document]
AndrewSisley: >  Have you discussed with the author of those tests

No I have not.  The mocks removed were already quite painful to manage, but mocking an iterator is significantly more painful than a old-style query call, and whilst it would be possible, it would very annoying and time consuming (possibly doubling the effort of this PR) to do so for the sake of a couple of unit tests.

If you like I can create an issue to review the tests removed, and potentially find a testing solution that a little more cost effective?

jsimnz (approved):
LGTM. Awesome work!  Long chapter of work to make this PR happen.

AndrewSisley:
> LGTM. Awesome work! Long chapter of work to make this PR happen.

Thanks John!

PR #3545, opened 2025-03-17
feat: Make signing algorithm configurable

### LINKED ISSUE #3511: Config signing algorithms for the node
At the moment the running node will sign (if enabled) blocks with ECDSA, but it might be useful to make it configurable which algorithms to use.

Before proceeding with this task we should clarify if long-term nodes are supposed at all the sign any blocks after we implement [client signing](https://github.com/sourcenetwork/defradb/issues/3460).

islamaliev:
## Relevant issue(s)

Resolves #3511

## Description

Introduce a configurable signing algorithm for nodes that allows using different sig types (secp256k1 or ed25519 at the moment) for signing DAG blocks. 

Iintroduce `PublicKey` and `PrivateKey` interafaces for `crypto` package.

Extended identities tp allow the use of different key types (secp256k1 and ed25519) for JWT token signing.


acp/identity/identity.go
```diff
@@ -14,32 +14,30 @@ import (
 	"encoding/hex"
 	"time"
 
-	"github.com/cyware/ssi-sdk/crypto"
+	gocrypto "github.com/cyware/ssi-sdk/crypto"
```
AndrewSisley, line 17: suggestion: If you re-export `github.com/cyware/ssi-sdk/crypto.CreateDIDKey` from `github.com/sourcenetwork/defradb/crypto` we can avoid having to directly reference this here and anywhere outside of our crypto package

acp/identity/identity.go
```diff
@@ -112,7 +111,14 @@ func didFromPublicKey(publicKey *secp256k1.PublicKey, producer didProducer) (str
 
 // IntoRawIdentity converts an `Identity` into a `RawIdentity`.
 func (identity Identity) IntoRawIdentity() RawIdentity {
-	return newRawIdentity(identity.PrivateKey, identity.PublicKey, identity.DID)
+	privKeyBytes, _ := identity.PrivateKey.Raw()
+	pubKeyBytes, _ := identity.PublicKey.Raw()
```
AndrewSisley, line 115: todo: Please return the errors, or at least panic if they are not nil.  Please do not ignore them unless you have a very well documented reason to do so.

acp/identity/raw_identity.go
```diff
@@ -11,44 +11,38 @@
...
 	PublicKey string
-
-	// DID is `did:key` key generated from the public key address.
+	// DID is the actor's unique identifier.
+	//
+	// The address is derived from the actor's public key,
+	// using the did:key method
 	DID string
+	// KeyType is the type of the key (secp256k1 or ed25519)
+	KeyType string
```
AndrewSisley, line 34: suggestion: Can we type this a little more strongly than just a string?

If not, please word the documentation more strongly, as this is a user-public type, perhaps something like:
```
// KeyType is the type of the key
//
// Supported values are:
// - "secp256k1"
// - "ed25519"
```

crypto/keys.go
```diff
@@ -0,0 +1,280 @@
...
+func (k *secp256k1PrivateKey) Equals(other Key) bool {
+	if other.Type() != KeyTypeSecp256k1 {
+		return false
+	}
+	otherBytes, err := other.Raw()
+	if err != nil {
+		return false
+	}
+	myBytes := k.key.Serialize()
+	return string(myBytes) == string(otherBytes)
```
AndrewSisley, line 142: suggestion: I would expect `bytes.Equals` to be more performant than copying both plus equals, it also looks slightly better IMO.

crypto/keys.go
```diff
@@ -0,0 +1,280 @@
...
+	// KeyTypeSecp256k1 represents a secp256k1 key
+	KeyTypeSecp256k1 KeyType = "secp256k1"
+	// KeyTypeEd25519 represents an Ed25519 key
+	KeyTypeEd25519 KeyType = "ed25519"
+)
+
+// Key represents a cryptographic key
+type Key interface {
+	// Equals checks whether two keys are the same
+	Equals(Key) bool
```
AndrewSisley, line 37: suggestion: I've done this a few times... `Equal` is more common in Go I think, and Fred might have a word if he spots this :grin:

crypto/keys.go
```diff
@@ -0,0 +1,280 @@
...
+	// Raw returns the raw bytes of the key
+	Raw() ([]byte, error)
```
AndrewSisley, line 39: suggestion: I can't spot an implementation that returns an error, is this future-proofing?  Can we remove the `error` return value?

acp/identity/identity.go
```diff
@@ -193,13 +207,27 @@ func VerifyAuthToken(ident Identity, audience string) error {
...
-		[]byte(ident.BearerToken),
-		jws.WithKey(BearerTokenSignatureScheme, ident.PublicKey.ToECDSA()),
-	)
+	// For now we only support ECDSA with secp256k1 or Ed25519 for bearer tokens
+	if ident.PublicKey.Type() != crypto.KeyTypeSecp256k1 && ident.PublicKey.Type() != crypto.KeyTypeEd25519 {
+		return crypto.ErrUnsupportedSignatureType
+	}
+
+	pubKey := ident.PublicKey.Underlying()
+	if secpPubkey, ok := pubKey.(*secp256k1.PublicKey); ok {
```
AndrewSisley, line 216: suggestion: The `Underlying` func is a bit odd to me in general, and it is not easy to tell why this `if` is doing what it is doing here.  I think a comment explaining why could help a save bit of guess work.

cli/start.go
```diff
@@ -137,7 +138,14 @@ func MakeStartCommand() *cobra.Command {
 				}
 			}
 
-			opts = append(opts, db.WithBlockSigning(!cfg.GetBool("datastore.nosigning")))
+			if !cfg.GetBool("datastore.nosigning") {
+				signingAlg := cfg.GetString("datastore.defaultsigningalg")
```
AndrewSisley, line 142: todo: Please make sure you have tests covering two P2P nodes sharing signed docs with each other, with each node signing them with a different algo.

suggestion: Perhaps tests covering the changing the default signing algo on a single node and read/writing docs on ether side of the switch.   If you think they would be relevant (I'm not sure they would be, but you know this space way better than me :)).

cli/start.go
```diff
@@ -275,6 +283,11 @@ func MakeStartCommand() *cobra.Command {
 		"no-signing",
 		cfg.GetBool(configFlags["no-signing"]),
 		"Disable signing of commits.")
+	cmd.Flags().String(
+		"default-signing-alg",
+		cfg.GetString(configFlags["default-signing-alg"]),
+		"Default signature algorithm to use for signing commits. Valid values are 'secp256k1' and 'ed25519'."+
+			"If not specified, the default signing algorithm will be 'secp256k1'.")
```
AndrewSisley, line 290: praise: This is very clear, thanks Islam :)

AndrewSisley (changes requested):
Looks really good Islam, I think I just had a couple of hopefully small todos and minor suggestions for you before merge :)

acp/identity/identity.go
islamaliev: the code was dead, I removed it

acp/identity/identity.go
islamaliev: added more docs to `Underlying`

crypto/keys.go
islamaliev: used it initially, but not anymore. Thanks for spotting.

acp/identity/raw_identity.go
islamaliev: `RawIdentity` is supposed to be directly serializable.
Applied your suggested commend

crypto/keys.go
islamaliev: renamed

crypto/keys.go
islamaliev: right, totally forgot about this one. Thanks

acp/identity/identity.go
```diff
@@ -52,17 +45,17 @@ type Identity struct {
...
+	did, err := pubKey.DID()
 	if err != nil {
 		return Identity{}, err
 	}
 
 	return Identity{
 		DID:        did,
 		PrivateKey: privateKey,
-		PublicKey:  publicKey,
+		PublicKey:  pubKey,
```
shahzadlone, line 58: nitpick: lol why the rename?

shahzadlone (commented):
Thanks for this, looks nice.

Just a general question, what was the reason to not do the `Generate` step in the `crypto.Key` interface. This way the Generate switch would not be needed as we have it? And the new concrete `Key` type can be made first using `KeyType` and then can call Generate function on it.

Also would be nice to see that tests where two nodes share signed docs with each other using diff keys

internal/merkle/clock/clock.go
```diff
@@ -238,15 +240,28 @@ func (mc *MerkleClock) signBlock(
...
 
-	sigBytes, err := crypto.SignECDSA256K(ident.Value().PrivateKey, blockBytes)
+	var sigBytes []byte
+	var sigType string
+
+	switch signingAlg {
+	case crypto.KeyTypeSecp256k1:
+		sigType = coreblock.SignatureTypeECDSA256K
+	case crypto.KeyTypeEd25519:
+		sigType = coreblock.SignatureTypeEd25519
```
jsimnz, line 250: todo: I should've raised this also in the last PR, this signature type should be `EDDSA`

jsimnz (changes requested):
Leaving this broad `todo` before commenting on individual sections/lines.

I probably should've raised this on the first digital sig PR, but its made more significant in this PR.

You are currently confusing the usage of "key type" and "signature algorithm" in some places. There is a dependency between them, but they are not interchangeable.

The key type implies the underlying cryptographic primitive used to *generate* the private/public key, the signature is the algorithm that takes that keypair and a message to produce a signature. Not all key types work with all signature algorithms.

I know you know this, but some parts of the implementation use both interchangeably or incorrectly.

This is most notable in the `cli` and `db` packages. Eg `default-signing-alg` as a config but used with `secp256k1` and `ed25519` values, which are then passed to the DB via the `WithSigningAlg` option. 

Additionally, I think the flow for generating keys and choosing signature algorithms shouldn't be separate. We should choose the key type at the time of keyring generation, and the signature algorithm will be dependent on the key type.

internal/merkle/clock/clock.go
```diff
@@ -238,15 +240,28 @@ func (mc *MerkleClock) signBlock(
...
+	default:
+		return NewErrUnsupportedSignatureAlgorithm(signingAlg)
+	}
+
+	sigBytes, err = ident.Value().PrivateKey.Sign(blockBytes)
```
jsimnz, line 255: follow up from our discord convo:

This line is generating a signature based on the key, which at the moment is still hard coded in the CLI keyring generate to use secp256k1.

But the sigType variable in this scope is defined by the config/cli parameter, so it could in theory be ed25519.

So we would have a coreblock.Signature using a secp256k1 key, with a header that says ed25519.

internal/merkle/clock/clock.go
islamaliev: done

cli/start.go
islamaliev: done

acp/identity/identity.go
islamaliev: just did some refactoring. It was a byproduct.

internal/merkle/clock/clock.go
islamaliev: made users define key type instead of signature algorithm.

AndrewSisley (approved):
LGTM, thanks Islam, and sorry for forgetting to re-review this earlier!

jsimnz (approved):
lgtm, nice rework!

fredcarle:
Bug bash result:

No bug found. Selecting a different default key type results in the appropriate signature type.

PR #3546, opened 2025-03-17
fix: Change HTTP endpoints that are delete with body

### LINKED ISSUE #3158: OPEN Api incorrect request body for delete operations
OpenAPI should avoid requestBody for delete operations. 

https://github.com/sourcenetwork/defradb/blob/develop/docs/website/references/http/openapi.json

```
Semantic error at paths./acp/relationship.delete.requestBody
DELETE operations cannot have a requestBody.

Semantic error at paths./collections/{name}.delete.requestBody
DELETE operations cannot have a requestBody.

Semantic error at paths./p2p/collections.delete.requestBody
DELETE operations cannot have a requestBody.

Semantic error at paths./p2p/replicators.delete.requestBody
DELETE operations cannot have a requestBody.
```

ChrisBQu:
## Relevant issue(s)

Resolves #3158

## Description

There were certain HTTP endpoints that utilisted HTTP DELETE requests, but which took in a body. This goes against convention. The issue raised, lists four paths that should be altered:

/acp/relationship/ DELETE
/collections/{name} DELETE
/p2p/collections DELETE
/p2p/replicators DELETE

This PR alters how the requests are handled, so that instead of taking in a request body, any needed information is instead passed in as a query parameter.

## Tasks

## How has this been tested?

- Windows


docs/website/references/http/openapi.json
```diff
@@ -795,6 +795,17 @@
             "delete": {
                 "description": "Delete an actor relationship using acp system",
                 "operationId": "delete relationship",
+                "parameters": [
+                    {
+                        "description": "Parameters for the delete request",
```
AndrewSisley, line 800: suggestion: This is not very descriptive, I am left very unsure as to what this is expecting or used for.

Maybe it would be worth expanding the description and/or adding an example here.

playground/package-lock.json
```diff
@@ -2770,7 +2699,7 @@
       "version": "19.0.10",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.0.10.tgz",
       "integrity": "sha512-JuRQ9KXLEjaUNjTWpzuR231Z2WpIwczOkBEIvbHNCzQefFIT0L8IqE6NV6ULLyC1SI/i234JnDoMkfg+RjQj2g==",
-      "devOptional": true,
+      "dev": true,
```
AndrewSisley, line 2702: question: Why has this file changed in this PR?  And what is this significance of this `"dev"`/"`devOptional"` property in particular?

AndrewSisley (commented):
LGTM, I have a question RE the package-lock.json file changes though

playground/package-lock.json
ChrisBQu: This is a good question actually. This is not a file that I touched by hand, so let me look into what change specifically would have resulted in this. I will not merge or anything like that until I have an answer. Thank you for pointing this one out.

fredcarle: If you haven't made any changes to the package.json file I would just remove this from the PR.

ChrisBQu: Wanted to provide an update saying that I removed it from the PR as you suggested.

docs/website/references/http/openapi.json
fredcarle: This file is auto generated Andy 🙂

AndrewSisley: The text is not though? It is based off of something somewhere manually created.

AndrewSisley: The example looks good, thanks Chris! I assume the `\t`s etc will not be rendered like that in clients?

I didn't realise it was a complex object, there is no way to declare it's shape, and inner prop docs, like we do for POST body elements?

ChrisBQu: Well, I believe the two obvious approaches would be accepting a JSON string as a single query parameter (I did that one), or alternatively, taking in four separate query parameters: collectionName, DocID, Relation, and TargetActor.

I don't think either one is necessarily cleaner than the other, but I do think the latter might be a little more clearly descriptive in terms of what should be passed in. What do you think of that idea?

AndrewSisley: I don't know this function well enough to comment on whether it should be broken up or not.

I meant (if keeping the JSON), can we declare it's shape by adding to `openApiSchemas` in openapi.go and adding the accompanying descriptions etc in `bindRoutes` in handler_store.go?

fredcarle: Lets talk about it during the standup. There are some issues with this approach that I didn't think about before.

PR #3552, opened 2025-03-19
fix: Only give badger a path if not in-memory

### LINKED ISSUE #3551: Badger in-memory fails to start if provided a path
This is quite problematic as the CLI always provides a path.

AndrewSisley:
## Relevant issue(s)

Resolves #3551 

## Description

Only give badger a path if not in-memory.

I broke this bringing corekv in.  Keenan fixed this, I'm just doing the typing.

This is tested in https://github.com/sourcenetwork/defradb/pull/3550


shahzadlone (approved):
LGTM

PR #3574, opened 2025-03-26
feat: Make replicator retry intervals configurable

### LINKED ISSUE #3073: Replicator configurable retry

ChrisBQu:
## Relevant issue(s)

Resolves #3073

## Description

This adds some functionality for configuring the durations of time between the replicator retries. Originally, a hardcoded set of values was defined in `internal/db/config.go`. Now, this is a config option inside `cli/config.go` called `replicator.retryIntervals`. 

Additionally, there is a config flag, `replicator-retry-intervals` that can be passed in as a slice of integers via CLI:

`defradb start --replicator-retry-intervals=10,20,66`

Further, there are some new unit tests included to test for ways in which the flag might throw errors on bad input.

## Tasks

## How has this been tested?

- Windows


cli/root.go
```diff
@@ -105,6 +105,12 @@ Start a DefraDB node, interact with a local or remote node, and much more.
...
 		"secret-file",
 		cfg.GetString(configFlags["secret-file"]),
-		"Path to the file containing secrets")
+		"Path to the file containing secrets",
+	)
+	cmd.PersistentFlags().String(
+		"replicator-retry-intervals",
+		cfg.GetString(configFlags["replicator-retry-intervals"]),
+		"Retry intervals for the replicator. Format is a comma-separated list of durations. Example: \"10,20,40,80,160,320\"",
+	)
```
fredcarle, line 114: todo: This flag shouldn't be part of the root command persisted flags since it's only useful on start (it's only relevant to a running DefraDB instance and not to client commands). Please move it to the start command specifically.

cli/root.go
```diff
@@ -105,6 +105,12 @@ Start a DefraDB node, interact with a local or remote node, and much more.
 	cmd.PersistentFlags().String(
 		"secret-file",
 		cfg.GetString(configFlags["secret-file"]),
-		"Path to the file containing secrets")
+		"Path to the file containing secrets",
+	)
+	cmd.PersistentFlags().String(
```
fredcarle, line 110: todo: Please use something like the `Int64Slice` flag type. This will do the parsing work for us instead of having to implement hour own like you did in `start.go`. You'll still have to assign the duration though.

fredcarle (commented):
The outcome of the change is what we want but we can improve the implementation a little with a couple minor changes.

fredcarle (changes requested):
I forgot to select `Request changes` 🙃

cli/root.go
ChrisBQu: Done!

cli/root.go
ChrisBQu: Done!

fredcarle (approved):
LGTM. If you could add a test with valid arguments before merge that would be great.

cli/start_replicator_retry_config_test.go
```diff
@@ -0,0 +1,97 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
```
AndrewSisley, line 11: todo: The action/multiplier stuff has been merged.  I think any tests for the CLI package should be built using that.  It will make maintaining them a lot easier over the long term, and we will hopefully get a lot more coverage per test (as the multipliers will affect them).

It would also be a good review of the system to have someone who is not me, write a few tests now that it is in.

AndrewSisley (changes requested):
I don't think we should be adding any new tests directly into the CLI package, please feel super free to argue with me on this.

(note: I've not reviewed the production code, I'm happy with it as it has already been reviewed - I just saw you were adding more unit tests where I believe there should be none)

cli/start_replicator_retry_config_test.go
ChrisBQu: Let me see what I can do!

cli/test/action/start.go
```diff
@@ -18,12 +18,27 @@ import (
...
 	stateful
+	inlineArgs    []string
+	expectedError error
 }
 
 var _ Action = (*StartCli)(nil)
 
+// Start with additional CLI arguments, and/or expecting an error (for example, if you
+// are testing with flags that should cause the command to fail)
+func StartWithArgs(args []string, expectedErr error) *StartCli {
```
AndrewSisley, line 29: suggestion: Having two functions would remove the need to pass in `nil` in a few places, simplifying the code and making it look more deliberate.  I suggest:
```
func StartWithArgs(args []string) *StartCli {...}
func StartWithArgsE(args []string, expectedErr error) *StartCli {...}
```

AndrewSisley (approved):
LGTM, thanks Chris!  Sorry about the bother I caused with the test-bug.  One small suggestion for you to look at before merge if that's alright :)

AndrewSisley:
Bug bash result:

No bugs found, but there was a documentation issue we missed: https://github.com/sourcenetwork/defradb/issues/3623

PR #3589, opened 2025-04-01
chore: All comms with the net package through events

### LINKED ISSUE #3572: Replace db dependency in net package with events

fredcarle:
## Relevant issue(s)

Resolves #3572 

## Description

This PR removes the last direct calls to the net package other than at node setup and closing. For the `node.Node`, the network implementation only needs to implement `io.Closer`. Adding peers can now be communicated through the event bus.

`node.Node` now uses the concrete `db.DB` type instead of the `client.DB` interface. The reasoning behind this is that `db.DB` is the core implementation of DefraDB and is not something that should be interchangeable like `http` or `net`. It removes the unnecessary interface limitations.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -20,6 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/sourcenetwork/corekv"
 	"github.com/sourcenetwork/corelog"
 	"github.com/sourcenetwork/immutable"
@@ -233,6 +234,25 @@ func (db *DB) AddPolicy(
 	}
 
 	return client.AddPolicyResult{PolicyID: policyID}, nil
+}
+
+// Connect sends peer connection event on the intern event bus. It blocks until the connection is
+// successful or an error is returned.
+func (db *DB) Connect(ctx context.Context, addr peer.AddrInfo) error {
+	err := make(chan error)
+	db.events.Publish(event.NewMessage(
+		event.PeerConnectName,
+		event.PeerConnect{
+			Info: addr,
+			Err:  err,
+		},
+	))
+	select {
+	case e := <-err:
+		return e
+	case <-time.After(10 * time.Second):
+		return ErrTimoutOnPeerConnection
+	}
 }
 
 // publishDocUpdateEvent publishes an update event for a document.
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -20,7 +20,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/sourcenetwork/corekv"
 	"github.com/sourcenetwork/corelog"
 	"github.com/sourcenetwork/immutable"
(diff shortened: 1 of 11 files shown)


net/peer.go
```diff
@@ -191,7 +196,7 @@ func NewPeer(
 }
 
 // Close the peer node and all its internal workers/goroutines/loops.
-func (p *Peer) Close() {
+func (p *Peer) Close() error {
```
AndrewSisley, line 199: todo: Can we remove this return param? It appears to be unused.

If you added it because you want to implement `io.Closer`, can you please at least document this, and why it needs to implement that interface.

If there is no real technical reason for it to implement that interface, please remove it, returning errors where there are none adds a notable amount of unnecessary boilerplate/complexity and distracts readers from the important parts of the code.

AndrewSisley (approved):
LGTM, I agree with the core changes, but have a couple of comments that I would prefer being sorted before merge.

net/peer.go
fredcarle: I was thinking of using the stdlib interface to keep things simple but I do agree that if we can avoid the error it makes more sense.

node/node.go
```diff
@@ -12,34 +12,26 @@ package node
...
-	PeerID() peer.ID
-	PeerInfo() peer.AddrInfo
-	Connect(context.Context, peer.AddrInfo) error
-}
-
 // Node is a DefraDB instance with optional sub-systems.
 type Node struct {
 	// DB is the database instance
-	DB client.DB
+	DB *db.DB
```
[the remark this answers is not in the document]
fredcarle: The thing is its not just about the cast. It's about useful methods not being exposed by the client interface. The `Connect` method in this case to be more specific. I prefer making this change here and stop fighting the interface while we clean things up.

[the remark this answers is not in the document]
AndrewSisley: Ah fair enough, I'm happy keeping as-is then :)  Might be worth talking about this in the standup though.

net/peer.go
fredcarle: Check out the last commit and let me know what you think.

AndrewSisley: image

AndrewSisley: If we are only going to have two items in the array it is probably an over complication IMO and two simple props would be better.  If you think that list is likely to grow it is a nice solution :)

fredcarle: My thinking was that if we want to make things a bit more indepandant and give more control to users as to when things are closed, we could have ACP, Lens, Net and HTTP all be handled by the terminator instead of hidden in different places. It will be even more useful when users want to assemble their own nodes.

internal/db/db.go
```diff
@@ -235,6 +236,25 @@ func (db *DB) AddPolicy(
 	return client.AddPolicyResult{PolicyID: policyID}, nil
 }
 
+// Connect sends peer connection event on the intern event bus. It blocks until the connection is
+// successful or an error is returned.
+func (db *DB) Connect(ctx context.Context, addr peer.AddrInfo) error {
```
islamaliev, line 241: suggestion: maybe it makes more sense to call it `PublicsPeerConnect` or something like this since this is the only thing the function does.

net/peer.go
```diff
@@ -264,6 +269,24 @@ func (p *Peer) handleMessageLoop() {
 		case event.Replicator:
 			p.server.updateReplicators(evt)
 
+		case event.PeerConnect:
+			if err := p.Connect(p.ctx, evt.Info); err != nil {
+				if evt.Err != nil {
```
islamaliev, line 274: question: is there a reason why the channel might be nil? Maybe a comment?

islamaliev (approved):
Looks nice. I have a minor suggestion.

fredcarle:
Closing this PR as all I was doing was fighting the interfaces. Will work on refactoring that instead.

PR #3591, opened 2025-04-03
feat: Enable block verification by clients

### LINKED ISSUE #3526: Make block signatures verifiable by clients
At the moment all block signatures are verified only during DAG sync process.
It might be also handy for clients to be able to verify a block against a given signature.

islamaliev:
## Relevant issue(s)

Resolves #3526 

## Description

This PR exposes the block signature verification functionality through the client API, allowing external verification of block signatures independently of the DAG sync process. This enhancement makes the existing internal verification capability accessible to clients.


cli/block_verify.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func MakeBlockVerifyCommand() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "verify",
```
fredcarle, line 21: todo: `verify` is vague in the context of a block because one can verify the CID, the history of change, the contents, the signature, etc. What this command is really doing is validating that a given identity has been used to sign the block and that the signature is valid. I think the command should be named more precisely (this applies to the method names as well). 

My suggestion is to name it `validate-signature` of `verify-signature`. 

The todo here is for reaching consensus with the DX team before merging. @ChrisBQu and @nasdf.

cli/block_verify.go
```diff
@@ -0,0 +1,48 @@
...
+		
+Notes:
+  - The identity must be specified.
+
+Example to verify the signature of a block:
+  defradb client block verify -i <identity> <cid>
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return fmt.Errorf("cid is required")
```
fredcarle, line 33: todo: Errors should be defined in the `errors.go` file.

http/handler_block.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+const (
+	errMissingParameter = "missing required parameter"
+	blockCidParam       = "cid"
+)
+
+type blockHandler struct{}
+
+func (h *blockHandler) bindRoutes(router *Router) {
```
fredcarle, line 28: todo: Please keep the `bindRoutes` method at the bottom of the file.

http/handler_block.go
```diff
@@ -0,0 +1,71 @@
...
+	router.AddRoute("/block/verify", http.MethodGet, verifyBlock, h.verifyBlock)
+}
+
+// verifyBlock handles block signature verification requests
+func (h *blockHandler) verifyBlock(w http.ResponseWriter, r *http.Request) {
+	db := mustGetContextClientDB(r)
+	cid := r.URL.Query().Get(blockCidParam)
+	if cid == "" {
+		responseJSON(w, http.StatusBadRequest, errorResponse{
+			errors.New(errMissingParameter, errors.NewKV("Parameter", "cid")),
```
fredcarle, line 59: todo: Please move the error definition to `errors.go`

internal/db/verify.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return coreblock.NewErrCouldNotLoadSignatureBlock(err)
+	}
+
+	identityStr := string(sigBlock.Header.Identity)
+	if identityStr != ident.Value().PublicKey.String() {
+		return NewErrSignatureIdentityMismatch(ident.Value().PublicKey.String())
+	}
+
+	_, err = coreblock.VerifyBlockSignature(block, linkSys)
```
fredcarle, line 77: thought: This method will fetch the signature block that has already been fetched above. Shouldn't there be a version of it that can take the block and the signature instead of the block and the link system?

internal/db/verify_test.go
fredcarle, line 1: thought: These all seem like integration test that can be handled by the integration test framework. I don't think it's worth the maintenance overhead of having them here.

internal/db/verify.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/ipld/go-ipld-prime/storage/bsadapter"
+
+	"github.com/sourcenetwork/defradb/acp/identity"
+	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
+)
+
+// VerifyBlock verifies the signatures of a block.
+// The context should carry the identity which will be used to verify the signatures.
+// Returns an error if any signature verification fails.
+func (db *DB) VerifyBlock(ctx context.Context, blockCID string) error {
```
fredcarle, line 28: thought: Correct me if I'm wrong but the goal of this function is to validate the signature of a block given some identity and that identity may be different than the one of the user/client calling it. If this is true, then we shouldn't rely on the context identity as the identity used within. The function should then take an identity as parameter. The context identity being that of the user/client should be used to validate that the it has access to the block to do the signature validation.

This will impact both `cli` and `http`.

fredcarle (changes requested):
A few todos and thoughts to start :)

cli/block_verify.go
islamaliev: renamed to `verify-signature`

http/handler_block.go
islamaliev: didn't know it was a thing. Done

cli/block_verify.go
islamaliev: done

http/handler_block.go
islamaliev: done

internal/db/verify.go
islamaliev: I thought about it myself, but then I realized that this method is public and it makes no sense to provide your own identity to call it.

Then cli would be like `client block verify-signature <cid> <identity>`. Which you can also execute with `client block verify-signature --identity <your_identity> <cid> <identity>`

Is it a good UX/DX @nasdf @ChrisBQu?

internal/db/verify.go
islamaliev: had the same thoughts, but wasn't sure if that was necessary. Refactored a bit

internal/db/verify_test.go
islamaliev: yeah, there is a big overlap here, but there is also a couple test that cover some edge-cases that integration tests can't reach.
I can remove the overlapping ones, but then there will only those few left. Not sure if that makes sense

internal/db/verify.go
fredcarle: > it makes no sense to provide your own identity to call it.
Why not? If your identity is not that of the node, why should the node allow you to check a block if you don't have access to it?

client/db.go
```diff
@@ -144,6 +144,11 @@ type DB interface {
 
 	// GetNodeIdentity returns the identity of the node.
 	GetNodeIdentity(context.Context) (immutable.Option[identity.PublicRawIdentity], error)
+
+	// VerifyBlock verifies the signatures of a block.
+	// The context should carry the identity which will be used to verify the signatures.
+	// Returns an error if any signature verification fails.
+	VerifyBlock(ctx context.Context, blockCID string) error
```
AndrewSisley, line 151: suggestion: `VerifySignature` would be a less ambiguous name IMO here

internal/core/block/signature.go
```diff
@@ -97,57 +97,100 @@ func (sig *Signature) GenerateNode() ipld.Node {
...
+	if err != nil {
+		return false, err
+	}
+
+	return true, verifySignature(pubKey, signedBytes, sigBlock.Value)
+}
+
+// VerifyBlockSignatureWithKey verifies the signature of a block using a public key.
+// The first return value is true if it actually ran cryptographic verification.
+func VerifyBlockSignatureWithKey(block *Block, lsys *linking.LinkSystem, pubKey crypto.PublicKey) (bool, error) {
```
AndrewSisley, line 134: todo: The boolean return param here is unused, and always equals `err == nil`, please remove it.

internal/core/block/signature.go
```diff
@@ -97,57 +97,100 @@ func (sig *Signature) GenerateNode() ipld.Node {
 	return bindnode.Wrap(sig, SignatureSchema).Representation()
 }
 
-// VerifyBlockSignature verifies the signature of a block.
+// verifySignature performs the cryptographic verification and returns appropriate results
```
AndrewSisley, line 100: todo: `and returns appropriate results` is hopefully true of all our functions.  Please either remove this, or modify it so that it describes what those appropriate results might be :)

internal/core/block/signature.go
```diff
@@ -97,57 +97,100 @@ func (sig *Signature) GenerateNode() ipld.Node {
...
 	}
+	return signedBytes, sigBlock, nil
+}
 
+// getBlockBytesToSign returns the bytes to sign for a block
+func getBlockBytesToSign(block *Block) ([]byte, error) {
 	blockToVerify := *block
 	blockToVerify.Signature = nil
 
 	signedBytes, err := marshalNode(&blockToVerify, BlockSchema)
```
AndrewSisley, line 175: nitpick: `return marshalNode(&blockToVerify, BlockSchema)` would do.

internal/core/block/signature.go
```diff
@@ -97,57 +97,100 @@ func (sig *Signature) GenerateNode() ipld.Node {
...
+
+	// Verify that the identity matches the signature's identity
+	if string(sigBlock.Header.Identity) != pubKey.String() {
+		return false, ErrSignaturePubKeyMismatch
+	}
+
+	return true, verifySignature(pubKey, signedBytes, sigBlock.Value)
+}
+
+func getSignedDataAndSignature(block *Block, lsys *linking.LinkSystem) ([]byte, *Signature, error) {
```
AndrewSisley, line 152: thought: This is kind of an odd function, as it appears to return two unrelated properties - `signedBytes` is not used to get `sigBlock` and just gets returned untouched by the other lines in this function.  To me this suggests that this function should be broken up and should not deal with `signedBytes`.

internal/db/verify_test.go
AndrewSisley: Why are the remaining tests not integration tests? They look like they are testing public behaviour.

AndrewSisley (changes requested):
LGTM, just a couple of minor todos

client/db.go
fredcarle: Yes I had mentioned that in my comment :) Thanks for flagging it.

islamaliev: done

internal/core/block/signature.go
islamaliev: done

internal/core/block/signature.go
islamaliev: done

internal/core/block/signature.go
islamaliev: the bool param is independent of errors. It indicates if cryptographic verification was done.
It can return `(false, nil)` and `(true, err)` as well.

internal/db/verify.go
islamaliev: adjusted

internal/db/verify_test.go
islamaliev: added missing case to the integration tests. Remove these tests

internal/core/block/signature.go
AndrewSisley: > It can return...

Ah sorry, I missed that.

But why bother return a boolean from this internal function if it is never used?

AndrewSisley (approved):
LGTM :)

internal/core/block/signature.go
islamaliev: its sibling function `VerifyBlockSignature` return `bool` which is used. So one reason is the keep them consistent.
Another thing is that the caller know exactly the case where this function can return `(false, nil)`, it's when the block doesn't have the signature, and the caller already checks itself for this condition and returns an error.

http/handler_block.go
```diff
@@ -0,0 +1,92 @@
...
+const (
+	blockCidParam  string = "cid"
+	publicKeyParam string = "public-key"
+	typeParam      string = "type"
+)
+
+type blockHandler struct{}
+
+// verifyBlock handles block signature verification requests
+func (h *blockHandler) verifyBlock(w http.ResponseWriter, r *http.Request) {
```
fredcarle, line 30: todo: The name should be changed to match the clients. `verifySignature`

fredcarle (approved):
Nice work Islam. Just a couple minor todos before merge.

http/handler_block.go
```diff
@@ -0,0 +1,92 @@
...
+	cidQueryParam := openapi3.NewQueryParameter("cid").
+		WithDescription("Content ID of the block to verify").
+		WithRequired(true).
+		WithSchema(openapi3.NewStringSchema())
+
+	verifyBlock := openapi3.NewOperation()
+	verifyBlock.OperationID = "verify_block"
+	verifyBlock.Description = "Verify block signature"
+	verifyBlock.Tags = []string{"block"}
+	verifyBlock.AddParameter(cidQueryParam)
```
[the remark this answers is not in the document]
islamaliev: done

http/handler_block.go
islamaliev: done

tests/integration/signature/verify_test.go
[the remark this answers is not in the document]
fredcarle: The author maybe but access to the block is protected. Only authorized identities should have access.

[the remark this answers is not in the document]
islamaliev: good call! I will add some tests.
But isn't it the job of the server node to block to p2p sync for the protected block so that the client node doesn't even have it locally?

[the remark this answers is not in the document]
fredcarle: At the p2p level there is no verify signature call. It's just the exchange of blocks and yes that is protected.

Here we are exposing a new endpoint that can be called by any identity. We need to make sure that what we return is accessible by that identity.

fredcarle:
Bug bash result:

No bug found. Was able to verify signature as expected.

PR #3593, opened 2025-04-03
test(i): Wait for database to close before progressing

### LINKED ISSUE #3592: CLI test framework race problem with os.stderr

AndrewSisley:
## Relevant issue(s)

Resolves #3592

## Description

Wait for the database to close before progressing to the next test.

Previously the database would close in the background while the next test began, this created a race condition when setting `os.Stderr`.

It was not noticed earlier as I only added a single example test per package - the bug surfaces when there are at least 2.

Credit to Chris, as he suffered with this bug for a while and was a bug help in figuring out where it was. 

## How has this been tested?

I added temporary tests to the package and ran with the `-race` flag.

### DIFF
--- a/cli/test/action/action.go
+++ b/cli/test/action/action.go
@@ -28,6 +28,8 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 )
 
+var stdErr *os.File = os.Stderr
+
 type Action = action.Action
 type Actions = action.Actions
 type Stateful = action.Stateful[*state.State]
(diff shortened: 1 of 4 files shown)


cli/test/action/action.go
```diff
@@ -204,8 +201,21 @@ func executeUntil(ctx context.Context, args []string, until string) (string, err
...
+	originalWait := s.Wait
+	s.Wait = func() {
+		originalWait()
+		// Because we are overwriting a singleton (os.Stderr) we must wait for the full command
+		// to finish executing before we allow the next test to execute, else the next test may
+		// try and overwrite os.Stderr before this execution finishes closing down the database.
+		//
+		// That might not cause any real bother, but the go test `-race` flag will complain.
+		isDoneWg.Wait()
+		os.Stderr = stdErr
```
fredcarle, line 216: question: Why is this needed now but not before?

fredcarle (approved):
To be honest, I don't like this at all 😅. We can keep it in the interest of time for now but I think we should find a better solution. One thing I don't like for exemple is it will hinder future parallelization of tests.

ChrisBQu:
LGTM!

And by that, I mean, I had run into the race conditions while writing my own test code. I can confirm that the changes here to solve the issue. My tests behave as expected with the changes applied, and no race conditions are detected. This is not to say

cli/test/action/action.go
AndrewSisley: Because it can be, before we were essentially piping pipes - this way we actually reset the state (this line is not required for the fix)

PR #3604, opened 2025-04-09
fix: Prevent infinite float values in documents

### LINKED ISSUE #2569: http.responseJSON fails to encode inf float and drops error
`json.NewEncoder(rw).Encode(out) //nolint:errcheck`

I'll add more info/tests tomorrow, found in https://github.com/sourcenetwork/defradb/pull/2567

ChrisBQu:
## Relevant issue(s)

Resolves #2569

## Description
It was possible to store an integer so large, or so largely negative, that it overflowed and was stored as `Inf` or `-Inf`. This resulted in future Updates to silently fail, and future Get requests to fail with an error. Both behaviors are undesirable. The proposed solution in this fix is to explicitly check that floats are not `NaN`, `Inf`, or `-Inf` when Creating or Updating a document. The proposed solution also adds logic to check for an overflow when incrementing a PNCounter.

**This is still in progress, do not merge.**

## Tasks

## How has this been tested?

I have written a set of tests inside `2569_test.go`. These tests do the following:

- Prevent creation of a document with a positively infinite float value
- Prevent creation of a document with a negatively infinite float value
- Prevent the updating of a document that would set a value to `Inf`
- Prevent the updating of a document that would set a value to `-Inf`
- Prevent the updating of a document that would increment a PNCounter to `Inf`
- Prevent the updating of a document that would decrement a PNCounter to `-Inf`

- Windows

### DIFF
--- a/internal/core/crdt/counter.go
+++ b/internal/core/crdt/counter.go
@@ -149,6 +149,7 @@ func (c Counter) Increment(ctx context.Context, value []byte) (*CounterDelta, er
 // Merge implements ReplicatedData interface.
 // It merges two CounterRegisty by adding the values together.
 func (c Counter) Merge(ctx context.Context, delta core.Delta) error {
+
 	d, ok := delta.(*CounterDelta)
 	if !ok {
 		return ErrMismatchedMergeType
@@ -230,6 +231,20 @@ func validateAndIncrement[T Incrementable](
 	}
 
 	newValue := curValue + value
+
+	// Handle float-specific overflow/NaN for float32 and float64
+	switch any(newValue).(type) {
+	case float32:
+		f := float64(newValue)
+		if math.IsNaN(f) || math.IsInf(f, 0) {
+			return nil, errCounterInfiniteOverflowOperation
+		}
+	case float64:
+		f := float64(newValue)
+		if math.IsNaN(f) || math.IsInf(f, 0) {
+			return nil, errCounterInfiniteOverflowOperation
+		}
+	}
 	return cbor.Marshal(newValue)
 }
 
--- a/client/document.go
+++ b/client/document.go
@@ -12,6 +12,7 @@ package client
 
 import (
 	"encoding/json"
+	"math"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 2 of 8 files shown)


tests/integration/issues/2566_test.go
```diff
@@ -1,213 +0,0 @@
...
-					Users {
-						Age
-					}
-				}`,
-				Results: map[string]any{
-					"Users": []map[string]any{
-						{
-							// Node 0 overflows before subtraction, and because subtracting from infinity
-							// results in infinity the value remains infinate
-							"Age": math.Inf(1),
```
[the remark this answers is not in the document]
ChrisBQu: This is definitely true. I've added two, one for positive, one for negative.

tests/integration/mutation/update/crdt/pncounter_test.go
```diff
@@ -315,22 +315,7 @@ func TestPNCounterUpdate_FloatKindWithPositiveIncrementOverflow_PositiveInf(t *t
...
-				}`,
-				Results: map[string]any{
-					"Users": []map[string]any{
-						{
-							"name":   "John",
-							"points": math.Inf(1),
-						},
-					},
-				},
+				ExpectedError: "error merging delta: operation results in a NaN, Inf, or -Inf value.",
```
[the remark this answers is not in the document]
ChrisBQu: This would be a major change of direction, I think. Let's discuss it in standup, I think.

internal/core/crdt/counter.go
```diff
@@ -230,6 +230,20 @@ func validateAndIncrement[T Incrementable](
 	}
 
 	newValue := curValue + value
+
+	// Disallow increments that cause overflow, and throw error
```
AndrewSisley, line 234: question: I thought we agreed to store infinitely precise floats, merge them, and then return NaN when queried?

ChrisBQu: I misunderstood. I thought the plan was to go ahead and merge this, and that this change would take place on a separate PR in the future. Clearly, I am mistaken. :)

ChrisBQu: Also, I see Fred reacted to the message, but I hope the notifications are not interrupting his vacation.

Issue #3607, opened 2025-04-10
Configure fallback signer


islamaliev:
Currently there is no way to create a signed document by a client over HTTP because in order to sign, the client identity's private key is required which we do not send over HTTP.

This makes it problematic for cases when ACP is activate, because users won't be able to create signed document at all when passing an identity (the node will receive only identity's public key).

We need to add a configuration to the db that allows specifying an fallback signer identity that should be used if the server node receives an identity (in `identity.FromContext`) that has no private.

PR #3617, opened 2025-04-14
feat: Configure fallback signer

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/3607

## Description

Allow configuring fallback identity that is used when signing is activated and the identity provided for a request is missing the private key. This happens when requests are sent over HTTP where we do not pass private keys.

Sign deletion blocks (forgot to include it in the previous commits).

### DIFF
--- a/internal/merkle/clock/clock.go
+++ b/internal/merkle/clock/clock.go
@@ -112,8 +112,9 @@ func (mc *MerkleClock) AddDelta(
 		dagBlock.Encryption = &encLink
 	}
 
-	if EnabledSigningFromContext(ctx) {
-		err = mc.signBlock(ctx, dagBlock)
+	signingEnabled, fallbackSigner := SigningFromContext(ctx)
+	if signingEnabled {
+		err = mc.signBlock(ctx, dagBlock, fallbackSigner)
 		if err != nil {
 			return cidlink.Link{}, nil, err
 		}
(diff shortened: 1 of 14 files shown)


internal/merkle/clock/clock.go
```diff
@@ -221,6 +222,7 @@ func encryptBlock(
 func (mc *MerkleClock) signBlock(
 	ctx context.Context,
 	block *coreblock.Block,
+	fallbackSigner immutable.Option[identity.Identity],
```
fredcarle, line 225: thought: I'm struggling a little bit with the fact that we have the main identity for signing taken from `ctx` inside the function but that the fallback one is taken from a function parameter. I makes it seem like the fallback one should be more important. Why not take them both from `ctx`?

islamaliev: no big reason. It's just that I used to store the flat that signing is enabled in the context, now instead of the bool flag that was always `true`, [I store in the context the `Option[Identity]`](https://github.com/sourcenetwork/defradb/pull/3617/files#diff-35ea5f48e29c2231c4f5331c5366332692b6c3f9c3b10523066d4eab230e51feR378).
I can make it use 2 different values with `EnabledSigningFromContext` and `FallbackSignerFromContext` it's better for understanding.

fredcarle: I think it would keep things more consistent and easy to understand if you did this :)

fredcarle:
Bug bash result:

The flag is not working as expected since the node was already falling back to using it's own identity if the request did not provide a private key. This means that setting the flag or not results in the same outcome.

PR #3629, opened 2025-04-18
fix(i): Node identity as default instead of fallback signer

### LINKED ISSUE #3628: Fallback signer flag has no impact on block signature.
Found during bug bash of #3617.

The flag is not working as expected since the node was already falling back to using it's own identity if the request did not provide a private key. This means that setting the flag or not results in the same outcome.

We should probably return an error instead if the request has no private key and the fallback identity hasn't been set.

fredcarle:
## Relevant issue(s)

Resolves #3628 

## Description

The PR revert the inclusion of the fallback signer configuration. There was a bug where if the request did not provide an identity, then the node identity would be used instead of the fallback signer. This was an opaque bug because in its current state, the fallback signer is always the node identity.

At standup we concluded that it would be best to just keep the node identity as the default at all times for now.

## Tasks

## How has this been tested?

make test and manual testing.

- MacOS

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -55,7 +55,6 @@ var configFlags = map[string]string{
 	"store":                      "datastore.store",
 	"no-encryption":              "datastore.noencryption",
 	"no-signing":                 "datastore.nosigning",
-	"use-fallback-signer":        "datastore.usefallbacksigner",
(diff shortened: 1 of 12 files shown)


nasdf (approved):
LGTM

jsimnz (approved):
Lgtm

Issue #3556, opened 2025-03-20
Unify the presentation of Schema/Collection


AndrewSisley:
Internally they can, and probably should, remain distinct.

Do not worry about multiple collections from a single schema.

PR #3644, opened 2025-04-28
refactor: Replace CollectionDescription.Root with ColID

### LINKED ISSUE #3641: Replace col.Root with CollectionID (SchemaRoot)

AndrewSisley:
## Relevant issue(s)

Resolves #3641

## Description

Removes CollectionDescription.Root from user-space, replacing it in most cases with the schema root.  Internally, in a few performance sensitive areas (like datastore keys), a short-id is used (the original id, only stored elsewhere.

Please make sure you have read https://github.com/sourcenetwork/defradb/issues/3556 before reviewing this.

I recommend reviewing commit by commit, and reading the extra info contained within the commit body.


client/collection_description.go
```diff
@@ -36,17 +27,12 @@ type CollectionDescription struct {
...
-	// Collections sharing the same RootID will be compatable with each other, with the documents
-	// within them shared and yielded as if they were in the same set, using Lens transforms to
-	// migrate between schema versions when provided.
-	RootID uint32
-
 	// The immutable ID of this collection version.
 	ID string
 
+	// The immutable ID of this collection, consistent across all versions.
+	CollectionID string
```
islamaliev, line 34: suggestion: after reading these 2 comments it doesn't become clear how they are different. Maybe it has something to do with naming. `ID` and `CollectionID` within `CollectionDescription` sound the same. Maybe it makes sense to rename `ID` to `VersionID`?

client/errors.go
```diff
@@ -158,10 +158,10 @@ func NewErrMaxTxnRetries(inner error) error {
 	return errors.Wrap(errMaxTxnRetries, inner)
 }
 
-func NewErrCollectionNotFoundForSchemaVersion(schemaVersionID string) error {
+func NewErrCollectionNotFoundForCollectionVersion(collectionVersionID string) error {
```
islamaliev, line 161: question: why did you rename it? The value that is passed here is `block.Delta.GetSchemaVersionID()`

internal/db/collection.go
```diff
@@ -483,7 +475,18 @@ func (c *collection) create(
...
 		txn := mustGetContextTxn(ctx)
-		valueKey := c.getDataStoreKeyFromDocID(docID)
+
+		shortID, err := id.ShortCollectionID(ctx, txn, c.Description().CollectionID)
+		if err != nil {
+			return err
+		}
+
+		valueKey := keys.DataStoreKey{
+			CollectionShortID: shortID,
```
islamaliev, line 485: question: is it just to save storage space?

internal/db/id/collection.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/db/sequence"
+	"github.com/sourcenetwork/defradb/internal/keys"
+)
+
+// ShortCollectionID returns the local, shortened, internal, collection id, which is used
+// only in locations where using the full CID would be a waste of storage space.
+//
+// If there is no short id found for the given full id, a new one will be generated and saved.
+func ShortCollectionID(
```
islamaliev, line 29: note: I was confused a bit by a few places where it's used until I read the documentation. Didn't know it was mutating the db state.

I don't suggest to rename it to `GetOrGenerateCollectionID` as it seems too long, but if you have a better idea, would be nice to have more clarity.

question: do you really need to have get-or-generate functionality in 1 function? I'm haft through the code and so far it looked like there is clear distinction "get" and "generate" parts. For example, in `SaveCollection` we want to generate (it even ignores the returned short id).

suggestion: Also before I read the implementation I assumed that it just a trimmed version of the CID. So far we mostly used "id" for hash-like identifiers and not serial numbers. Maybe it make sense to reconsider how we name it.

internal/db/schema.go
```diff
@@ -473,6 +472,7 @@ func (db *DB) updateSchema(
 			// they will be activated later along with any existing collection versions.
 			col.Name = immutable.None[string]()
 			col.ID = schema.VersionID
+			col.CollectionID = schema.Root
```
islamaliev, line 475: question: if it's effectively a schema root why not to call it `SchemaRoot` instead of `CollectionID`?

internal/keys/systemstore_collection_id.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"strings"
+
+	ds "github.com/ipfs/go-datastore"
+)
+
+// CollectionID indexes collection short ids by the full id.
+type CollectionID struct {
+	CollectionID string
+}
```
islamaliev, line 22: suggestion: why not 
```go
type CollectionID string
```
?

islamaliev (approved):
Good job, Andy. I like it overall. I just have some concerns regarding some names. I hope we clarify them before the merge.

client/collection_description.go
AndrewSisley: Long term plan is to rename `ID` to `VersionID`, I was going to do it later (this sprint) - do you feel that it is confusing enough need doing now?

client/errors.go
AndrewSisley: `Schema` is disappearing as a concept (minus GQL-SDL schema, which is not a defra specific thing), so I've been converting stuff away from 'Schema' names when I can.  It will remain a little confused until SchemaDescription (and fields) are removed (converted into Collection blocks).

internal/db/collection.go
AndrewSisley: Space, memory, and read/write time.

internal/db/id/collection.go
AndrewSisley: > do you really need to have get-or-generate functionality in 1 function

I do not think this is required, I think I wrote it this way to avoid having to think about that (and handle errors when key is not found).

I'll have a look at this though, as I think I agree with you in that the illusion/possibility of mutation on most gets is an unwelcome complication.  This would also alleviate your naming concern.

- [ ] Consider splitting func

> So far we mostly used "id" for hash-like identifiers and not serial numbers

I'm not sure I consider that to be true, Collection ID, and CollectionField ID have been called 'ID' and have been serial numbers since before I joined Source and are only now becoming CIDs.

Do let me know if you have any alternative names that you like though.

internal/db/schema.go
AndrewSisley: Schema is to be removed. `CollectionID` will be the CID of the collection block.

internal/db/id/collection.go
AndrewSisley: Thanks Islam!  Exploring this surfaced a bug (introduced in this PR) RE the versioned fetcher that would have been a real pain to track down later once this change was out of mind.  `ShortCollectionID` has also been split into two.

PR #3647, opened 2025-04-29
docs: Update outdated disclaimer about lack of encryption

getify:
per @jsimnz

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ Read the documentation on [docs.source.network](https://docs.source.network/).
    * [Contributors](#contributors)
 <!--te-->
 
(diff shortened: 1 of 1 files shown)


jsimnz (changes requested):
todo: I think for the disclaimer, we can leave it the rest about production ready. For the "do not use it for production purposes", we can lighten up this language a bit since we do currently have some partners using in prod. 

Instead we can say something like: "we encourage teams looking to deploy to production to reach out to us". Or something to that affect.

jsimnz (approved):
LGTM

shahzadlone (approved):
Perhaps `docs(i): ..` in title.

LGTM

jsimnz:
> Perhaps docs(i): .. in title.
I don't think the `i` is necessary. Its fine to have this in the changelog

Issue #3650, opened 2025-04-30
Sec. index: utilize sec. indexes for sorting


islamaliev:
Currently when a query contains `order` field, it would be handled by `planner.orderNode` which sits on top of the planner tree hierarchy and collects all the documents in the collection in order to sort them. This is very resource-consuming for large collections and it also makes the `limit` query field useless in terms of node's resources.

We should utilize secondary indexes whenever possible for sorting as well.

PR #3652, opened 2025-04-30
feat: Utilize secondary indexes for ordering

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/3650

## Description

This change utilizes secondary indexes for order queries, so that not only filtering but, also sorting is optimized.
Whenever a secondary index takes care of ordering on the documents, the planner optimization pass will detect it and ignore `planner.orderNode`.

## Tasks

## How has this been tested?

Tests

- MacOS

### DIFF
--- a/internal/db/fetcher/indexer_iterators.go
+++ b/internal/db/fetcher/indexer_iterators.go
@@ -158,6 +158,25 @@ func (iter *indexPrefixIterator) Close() error {
 	return iter.resultIter.Close()
 }
 
+func (f *indexFetcher) newPrefixIterator(
+	indexKey keys.IndexDataStoreKey,
+	matchers []valueMatcher,
+	execInfo *ExecInfo,
+) *indexPrefixIterator {
+	return &indexPrefixIterator{
+		indexDesc:     f.indexDesc,
+		indexedFields: f.indexedFields,
+		indexKey:      indexKey,
+		matchers:      matchers,
+		execInfo:      execInfo,
+	}
+}
+
+func (f *indexPrefixIterator) Reverse(reverse bool) *indexPrefixIterator {
+	f.reverse = reverse
+	return f
+}
+
 type eqSingleIndexIterator struct {
 	indexKey keys.IndexDataStoreKey
 	execInfo *ExecInfo
@@ -313,7 +332,7 @@ func (iter *memorizingIndexIterator) Close() error {
 	return iter.inner.Close()
 }
 
-// newPrefixIteratorFromConditions creates a new eqPrefixIndexIterator for fetching indexed data.
+// newPrefixIteratorFromConditions creates a new indexPrefixIterator for fetching indexed data.
 // It can modify the input matchers slice.
 func (f *indexFetcher) newPrefixIteratorFromConditions(
 	fieldConditions []fieldFilterCond,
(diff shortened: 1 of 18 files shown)


internal/db/fetcher/indexer_iterators.go
```diff
@@ -428,6 +440,32 @@ func (f *indexFetcher) newIndexDataStoreKeyWithValues(values []client.NormalValu
...
+		if err != nil {
+			return nil, err
+		}
+		iter := f.newPrefixIterator(key, nil, f.execInfo).Reverse(reverse)
+		return iter, nil
+	}
+	return nil, nil
+}
+
+func (f *indexFetcher) isIndexOrdered() (bool, bool) {
```
AndrewSisley, line 456: suggestion: A comment describing what each returned boolean represents might speed up future reading

internal/planner/planner.go
```diff
@@ -301,33 +301,76 @@ func findFilteredByRelationFields(
...
+				// At the moment we just take the first index, but later we want to run some kind of analysis to
+				// determine which index is best to use. https://github.com/sourcenetwork/defradb/issues/2680
+				err := node.invertJoinDirectionWithIndex(indexes[0], fieldFilter, nil)
+				if err != nil {
+					return err
+				}
+				break
+			}
+		}
+	} else if parentPlan.order != nil && len(parentPlan.order.ordering) > 0 {
```
AndrewSisley, line 359: todo: Decent documentation explaining why each block is thought to be optimal would be really handy here, as both blocks are located quite some distance from their main code, and it is very reasonable to expect this function to receive significant work in the future as we get a little smarter about query optimization.

internal/planner/planner.go
```diff
@@ -301,33 +301,76 @@ func findFilteredByRelationFields(
 	return filteredSubFields
 }
 
+func findOrderedByRelationFields(
+	ordering mapper.OrderCondition,
+	mapping *core.DocumentMapping,
+) (string, error) {
+	fieldIndex := ordering.FieldIndexes[0]
+	_, found := mapping.TryToFindNameFromIndex(fieldIndex)
```
AndrewSisley, line 309: suggestion: A quick comment here would be handy in order to tell why we are bothering with this check.

internal/planner/type_join.go
```diff
@@ -825,6 +827,28 @@ func (join *invertibleTypeJoin) invertJoinDirectionWithIndex(
 	return nil
 }
 
+// isOrderedByIndex checks if the plan is ordered by an index.
+func isOrderedByIndex(plan planNode) bool {
```
AndrewSisley, line 831: suggestion: This file seems an odd location for this function, I suggest moving it somewhere else, probably to `planner.go` - the only place it is currently consumed.

internal/planner/type_join.go
```diff
@@ -842,12 +866,11 @@ func addFilterOnIDField(f *mapper.Filter, propIndex int, docID string) *mapper.F
 	return f
 }
 
-func getScanNode(plan planNode) *scanNode {
+func getNode[T planNode](plan planNode) T {
```
AndrewSisley, line 869: praise: The generic is a really nice way of handling this, thanks :)

AndrewSisley (approved):
LGTM, only long-term documentation requests from me I think.  Might be work checking the Petastic-Slack to see if it is worth mentioning to them on merge.

internal/planner/type_join.go
```diff
@@ -825,6 +827,28 @@ func (join *invertibleTypeJoin) invertJoinDirectionWithIndex(
...
+			scan = getNode[*scanNode](j.getFirstSide().plan)
+		}
+	} else {
+		scan = getNode[*scanNode](plan)
+	}
+	if scan == nil || !scan.index.HasValue() {
+		return false
+	}
+	fieldIndexes := scan.documentMapping.IndexesByName[scan.index.Value().Fields[0].Name]
+	return len(fieldIndexes) > 0 && len(scan.ordering) > 0 && scan.ordering[0].FieldIndexes[0] == fieldIndexes[0]
```
shahzadlone, line 849: nitpick: `return` statement a bit hard to read, split the `len > 0` checks perhaps.

shahzadlone (approved):
LGTM

internal/planner/type_join.go
islamaliev: done

internal/planner/planner.go
islamaliev: that's a good point. I don't need this check.

internal/planner/planner.go
islamaliev: done

AndrewSisley:
Bug bash result, https://github.com/sourcenetwork/defradb/issues/3774 found

PR #3653, opened 2025-04-30
refactor: Replace schema-only collections with boolean

### LINKED ISSUE #3646: Remove schema-only definitions

AndrewSisley:
## Relevant issue(s)

Resolves #3646

## Description

Replaces schema-only collections with a boolean `IsEmbeddedOnly` boolean.

I've kept the existing `interface` GQL syntax as-is, we can change that later if we wish, but I'd rather not get into that right now.

### DIFF
--- a/client/collection_description.go
+++ b/client/collection_description.go
@@ -84,6 +84,12 @@ type CollectionDescription struct {
 	// that will change in the future.
 	IsBranchable bool
 
+	// IsBranchable defines whether this collection exists only as a child object embedded within
+	// another collection or not.
+	//
+	// If true, it will not be directly queriable.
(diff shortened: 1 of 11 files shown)


client/collection_description.go
```diff
@@ -84,6 +84,12 @@ type CollectionDescription struct {
 	// that will change in the future.
 	IsBranchable bool
 
+	// IsBranchable defines whether this collection exists only as a child object embedded within
```
islamaliev, line 87: typo: "IsBranchable" -> "IsEmbeddedOnly"

islamaliev (approved):
LGTM

PR #3657, opened 2025-05-02
refactor(i): Reorganize existing document acp

### LINKED ISSUE #3656: Clean up and re-organize document access control
Clean up and re-organize document access control to build aac on top of in a cleaner way.

shahzadlone:
## Relevant issue(s)
Resolves #3656 

## Description
As discussed in the standup last week breaking this PR out of my admin acp work. The admin acp work builds on top of this. 

What was done:
- a small new interface was used to make document specific permissions more general (to accommodate admin perms)
- renamed all generic existing `acp` terminology to `documentACP` to avoid confusion with `adminACP`
- refactored sourcehub client to be a more general acp system client (which already had 2 clients, admin will be another client soon).
- moved the document acp logic to the `acp/document` directory
- keep the common acp logic in the main `acp/` folder like the resource validation stuff.
- moved the acp types to the `acp/types` directory to avoid circular dependency
- made some errors and acp types public to be used by document and admin acp systems
- rename permission names to be namespaced by `Document` in the acp/types` to avoid clashing with admin permissions that will be introduced.
- change the dpi terminology to document resource interface  terminology (to avoid confusion with admin resource interface and also because dpi was wrong it should have been dri to begin with).
- rename some document specific things in sourcehub client, like docID to objectID, as it should be acp client specfic and not document specfic
- adjust test file names, package names and function names
- change ci workflow name to reflect document acp clearly
- change env var from `DEFRA_ACP_TYPE` to `DEFRA_DOCUMENT_ACP_TYPE`

What was not done:
- no implementation or behavior changes were done
- document identity stuff was not renamed
- no refactoring done through introducing new abstractions (just moved existing ones)
- the acp cli stuff was not moved under `defradb client acp dac ...` yet, will be done when aac is introduced
- did not include new `acp/README.md` changes because I don't want to confuse @ashutosh-src already in transition to updating to current develop documentation state + will be nicer to have all those changes included with aac implementation IMO


acp/acp.go
```diff
@@ -13,23 +13,18 @@ package acp
...
-	log = corelog.NewLogger("acp")
-
-	// NoACP is an empty ACP, this is used to disable access control.
-	NoACP = immutable.None[ACP]()
-)
-
-// ACP is the interface to all types of access control that might exist.
-type ACP interface {
+// ACPSystemClient is an abstraction to allow multiple types of ACP systems to share DefraDB specific logic.
+type ACPSystemClient interface {
```
shahzadlone, line 25: Note: diff looks odd but it's just copy pasted of what sourcehub client was before (+ some renames) and this `ACP` interface is now `DocumentACP` under `acp/dac`. Both being used just like they were before.

internal/db/permission/check.go
```diff
@@ -115,10 +116,15 @@ func CheckDocAccessWithIdentityFunc(
 		identityValue = identity.Value().DID
 	}
 
+	documentResourcePerm, ok := permission.(acpTypes.DocumentResourcePermission)
+	if !ok {
+		return false, ErrInvalidResourcePermissionType
```
islamaliev, line 121: thought: I wonder if there is a way to test it

node/acp_dac.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/sourcenetwork/defradb/acp/dac"
+)
+
+type DocumentACPType string
+
+const (
+	// NoDocumentACPType disables the document ACP subsystem.
+	NoDocumentACPType DocumentACPType = "none"
+	// DefaultDocumentACPType uses the default ACP implementation for this build.
+	DefaultDocumentACPType DocumentACPType = ""
```
islamaliev, line 28: question: is it supposed to be empty?

acp/dac/bridge.go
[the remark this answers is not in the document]
shahzadlone: I understand where you are coming from but it's copy pasted from before nothing changed.

internal/db/permission/check.go
shahzadlone: Can possibly pass an invalid document resource permission in a unit test. However this is not going to be something we have to worry about much as it's fairly internal.

node/acp_dac.go
shahzadlone: It is the current develop behavior is copy-pasted/moved, nothing changed

acp/dac/bridge.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+func (a *bridgeDocumentACP) ValidateResourceInterface(
+	ctx context.Context,
+	policyID string,
+	resourceName string,
+) error {
+	var err error
+	switch a.clientACP.(type) {
+	case *LocalDocumentACP:
```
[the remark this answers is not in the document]
shahzadlone: Yes, only place for now, but with admin stuff it will have at least 1 more location I believe. Nonetheless will double check to see if I can avoid it.

- [ ] Check if can avoid type assertion.

acp/acp.go
```diff
@@ -40,102 +35,98 @@ type ACP interface {
...
-	// CheckDocAccess returns true if the check was successfull and the request has access to the document. If
-	// the check was successful but the request does not have access to the document, then returns false.
+	// VerifyAccessRequest returns true if the check was successfull and the request has access to the object. If
+	// the check was successful but the request does not have access to the object, then returns false.
 	// Otherwise if check failed then an error is returned (and the boolean result should not be used).
-	//
-	// Note(s):
-	// - permission here is a valid DPI permission we are checking for ("read" or "update" or "delete").
-	CheckDocAccess(
+	VerifyAccessRequest(
```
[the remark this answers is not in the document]
shahzadlone: It's all copy pasted from before haha, most of these terms were adopted from acp/sourcehub lang in first implementation. There were sourcehub/acp-core related reasons for having the `Request` part existing in `VerifyAccessRequest` (I am also not a huge fan haha).

While I wanted this PR to focus mainly on fixing document access control specific stuff, If you have concrete suggestion improvements for what to change/improve documentation with, you can use the `suggestions` code block here and I will just apply them.

acp/dac/bridge.go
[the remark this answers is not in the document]
shahzadlone: Revisited, New interface method would be required on the general acp interface that shouldn't depend on defradb logic so I am hesitant to do that approach to avoid this assertion. I might revisit this in the admin acp work if a more cleaner solution can be had. Leaving as it until then.

Issue #3699, opened 2025-05-12
Remove CollectionFieldDescription.ID


AndrewSisley:
And replace it with a short id.  Index can be keyed by name in the short term before being replaced with cid.

AndrewSisley:
Work has begun on this, branch `3699-rm-col-field-id`

PR #3701, opened 2025-05-13
feat: Replace commit query fieldID with fieldName

### LINKED ISSUE #1453: Add `fieldName` param to commits queries
e.g.:
```
commits (fieldName: "age") {
	cid
}
```

AndrewSisley:
## Relevant issue(s)

Resolves #1453

## Description

Replaces the commit query fieldID parameter with a new fieldName param.  Also removes the commit query fieldID field for the same reason (a fieldName field already exists).

Whilst fieldName is less efficient (for now) when providing anything but the composite identifier (the most common/important case and includes the `_version` object in doc queries), it is much simpler for users.  FieldID is soon to become an internal-only concept, and will make even less sense to users than it currently does.

FieldID may return at some point, using the field CID that will be introduced in later PR (hopefully this sprint).

Using `_C` to identify composites is questionable, IMO, it is good enough for now, and lets me crack on with my primary goal.  Long term we might want to split this out to another param, e.g. '_compositesOnly'.

This PR was broken out of https://github.com/sourcenetwork/defradb/issues/3699 and so is part of the epic https://github.com/sourcenetwork/defradb/issues/3556

### DIFF
--- a/client/request/commit.go
+++ b/client/request/commit.go
@@ -32,11 +32,12 @@ type CommitSelect struct {
 	// belonging to the given document.
 	DocID immutable.Option[string]
 
-	// FieldID is an optional filter which when provided will limit commits to those
+	// FieldName is an optional filter which when provided will limit commits to those
 	// belonging to the given field.
 	//
-	// `C` may be provided for document-level (composite) commits.
-	FieldID immutable.Option[string]
+	// `_C` may be provided for document-level (composite) commits. An explicit empty
+	// value will return branchable collection commits only.
+	FieldName immutable.Option[string]
 
 	// Depth limits the returned commits to being X places in the history away from the
 	// most current.
--- a/client/request/consts.go
+++ b/client/request/consts.go
@@ -19,13 +19,15 @@ const (
 	// that corresponds to the related object's join relation id, i.e. `Author_id`.
 	RelatedObjectID = "_id"
 
-	Cid         = "cid"
-	Input       = "input"
-	CreateInput = "create"
-	UpdateInput = "update"
-	FieldName   = "field"
-	FieldIDName = "fieldId"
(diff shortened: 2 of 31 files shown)


client/request/commit.go
```diff
@@ -32,11 +32,12 @@ type CommitSelect struct {
...
 
-	// FieldID is an optional filter which when provided will limit commits to those
+	// FieldName is an optional filter which when provided will limit commits to those
 	// belonging to the given field.
 	//
-	// `C` may be provided for document-level (composite) commits.
-	FieldID immutable.Option[string]
+	// `_C` may be provided for document-level (composite) commits. An explicit empty
+	// value will return branchable collection commits only.
+	FieldName immutable.Option[string]
```
fredcarle, line 40: suggestion: Since we are introducing a name for a system level value (the composite), I think it would be more intuitive for users to also have a name for the branchable collection as well. It would be more consistant. `_B` would be fine I think. Otherwise why would ove have a name and the other be `nil`.

fredcarle (commented):
LGTM. Just one suggestion more related to DX than implementation.

client/request/commit.go
AndrewSisley: Oh that's a good spot.  The composite name is not new though, it existed as `C` on the old fieldID, so I'm going to leave this as-is for now, but we should figure out what we want long term here at some point (the PR description suggests a `compositeOnly` input param, we could also have a `collectionOnly`, or perhaps an enum `type` with values `All` (default), `Composite`, `Collection`, `Field`, etc).

Thanks for the idea :)

fredcarle (approved):

fredcarle:
Bug bash result: `FieldName` is available as a requested field and filter. Working as expected

PR #3703, opened 2025-05-13
perf: Cache collection short ids within txn context

### LINKED ISSUE #3702: Add context cache for collection short ids
Working on https://github.com/sourcenetwork/defradb/issues/3699 I feel like the field ids are used in too hot  paths to tolerate fetching each usage (like collections do) - we need to sort out caching first.

Current (wip) thought is a `descriptions` cache on `ctx`, it can be a strongly typed complex object with separate maps for collection versions and short ids.  It would (optionally) be possible to use this cache/ctx across operations too, but we'd then need to ensure that the transaction remains constant.

AndrewSisley:
## Relevant issue(s)

Resolves #3702

## Description

Caches collection short ids within txn context, introducing a framework for all future txn-scoped caches to use (such as collection versions, and field short ids).

Performance impact of this PR will be quite small, this will have a much bigger impact when applied to collection versions, and field short-ids.

### DIFF
--- a/internal/db/id/collection.go
+++ b/internal/db/id/collection.go
@@ -26,6 +26,12 @@ func GetShortCollectionID(
 	txn datastore.Txn,
 	collectionID string,
 ) (uint32, error) {
+	cache := getCollectionShortIDCache(ctx)
+	shortID, ok := cache[collectionID]
+	if ok {
+		return shortID, nil
+	}
+
 	key := keys.NewCollectionID(collectionID)
 
 	valueBytes, err := txn.Systemstore().Get(ctx, key.Bytes())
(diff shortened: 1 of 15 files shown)


internal/db/id/collection.go
```diff
@@ -61,10 +76,38 @@ func SetShortCollectionID(
...
 
-	return txn.Systemstore().Set(ctx, key.Bytes(), []byte(strconv.Itoa(int(shortID))))
+	cache[collectionID] = shortID
+
+	return nil
+}
+
+type collectionShortIDCacheKey struct{}
+
+var collectionShortIDCacheKeyValue = collectionShortIDCacheKey{}
```
fredcarle, line 97: nitpick: I'm not a fan of this. It's really normal to use the type directly and having the extra variable is a bit confusing and unnecessary.

fredcarle (approved):
LGTM with one nitpick and assuming you'll fix the js failure.

internal/db/id/collection.go
AndrewSisley: Fair enough, I'll remove it.

- [ ] Remove unnecessary var

Thanks for spotting the JS failure, the job has been very flaky lately and I didn't bother checking this one

- [ ] Fix JS build

PR #3705, opened 2025-05-14
refactor: Allow retrieval of txn from ctx everywhere

### LINKED ISSUE #3704: Allow fetching of txn from ctx anywhere
Is a bit annoying and pointless to have to pass both ctx and txn everywhere

AndrewSisley:
## Relevant issue(s)

Resolves #3704

## Description

Allows the retrieval of txn from ctx everywhere.  Does not remove the many function params that could/should be removed because of this change (we can do that gradually over time).

Is annoying and messy to have to pass both independently into most functions in our codebase, especially when one lives within the other.

Broken out of #3699


fredcarle (approved):
LGTM. The only thing that I don't like about this is the package name. I'm wondering if we should have a `ctx` package and have most related things in there. Like the id caches and the context init function.

AndrewSisley:
> The only thing that I don't like about this is the package name

Yeah, it's not ideal - I called it `txn` at first but it of course was shadowed by everything. `ctx` would have the same problem.

>  I'm wondering if we should have a ctx package and have most related things in there. Like the id caches and the context init function.

I'm currently against that as it forces unrelated stuff into the same space and IMO goes against the spirit of `ctx`.  It would however allow us to create a strongly typed `ctx`, but I'm still very weary of the coupling it would introduce.

PR #3707, opened 2025-05-14
chore: Improve node startup sync and feedback

### LINKED ISSUE #3706: Improve node start testing
Remove reliance on logs for node status check when testing.

fredcarle:
## Relevant issue(s)

Resolves #3706

## Description

This PR introduces a message channels struct to use the cli command context to relay node information outside of the `cli.Start` command. This removes the need to use logs to figure out node state. Some improvement to the synchronousness of the startup process was introduced.

## Tasks

## How has this been tested?

make start

- MacOS

### DIFF
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -51,6 +51,8 @@ var (
 	// If a transaction exists, all operations will be executed
 	// in the current transaction context.
 	colContextKey = contextKey("col")
+	// messageChansContextKey is the context key for the messageChans
+	messageChansContextKey = contextKey("messageChans")
 )
 
 const (
@@ -93,6 +95,24 @@ func mustGetContextConfig(cmd *cobra.Command) *viper.Viper {
 	return cmd.Context().Value(cfgContextKey).(*viper.Viper) //nolint:forcetypeassert
 }
 
+// MessageChans contains message channels that can be used to relay important information
+// after calling `cli.Start`.
+type MessageChans struct {
+	APIURL chan string
+}
+
+// tryGetContextMessageChans returns the message channels for the current command context
+// and a boolean indicating if the message channels struct was set.
+func tryGetContextMessageChans(ctx context.Context) (*MessageChans, bool) {
+	node, ok := ctx.Value(messageChansContextKey).(*MessageChans)
+	return node, ok
+}
+
+// SetContextMessageChans sets the message channels for the current command context.
+func SetContextMessageChans(ctx context.Context, w *MessageChans) context.Context {
+	return context.WithValue(ctx, messageChansContextKey, w)
+}
+
 // mustGetContextRootDir returns the rootdir for the current command context.
 //
 // If a rootdir is not set in the current context this function panics.
@@ -119,7 +139,7 @@ func setContextDB(cmd *cobra.Command) error {
 	return nil
 }
 
-// setContextConfig sets teh config for the current command context.
+// setContextConfig sets the config for the current command context.
 func setContextConfig(cmd *cobra.Command) error {
 	rootdir := mustGetContextRootDir(cmd)
 	cfg, err := loadConfig(rootdir, cmd.Flags())
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -51,8 +51,6 @@ var (
 	// If a transaction exists, all operations will be executed
 	// in the current transaction context.
 	colContextKey = contextKey("col")
-	// messageChansContextKey is the context key for the messageChans
-	messageChansContextKey = contextKey("messageChans")
 )
 
 const (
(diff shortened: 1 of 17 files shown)


cli/utils.go
```diff
@@ -93,6 +95,24 @@ func mustGetContextConfig(cmd *cobra.Command) *viper.Viper {
 	return cmd.Context().Value(cfgContextKey).(*viper.Viper) //nolint:forcetypeassert
 }
 
+// MessageChans contains message channels that can be used to relay important information
+// after calling `cli.Start`.
+type MessageChans struct {
```
AndrewSisley, line 100: suggestion: I don't think I like defining this in the utils.go file (as that really doesn't scale), and I don't think I like the idea of lots of non-command files in the `cli` package, leaving me with the suggestion of moving this into a `message.go` file in a new `utils` sub-package of `cli`.

http/client.go
```diff
@@ -479,6 +479,16 @@ func (c *Client) Purge(ctx context.Context) error {
 	return err
 }
 
+func (c *Client) HealthCheck(ctx context.Context) error {
```
AndrewSisley, line 482: question: This appears to be unrelated, is it?

AndrewSisley, line 482: suggestion: This looks like the CLI would benefit from having this command too.

AndrewSisley, line 482: thought: This looks very similar to the previously removed `ping` command that we used to have.

node/node_api.go
```diff
@@ -43,5 +43,15 @@ func (n *Node) startAPI(ctx context.Context) error {
 			log.ErrorContextE(ctx, "HTTP server stopped", err)
 		}
 	}()
+	n.APIURL = n.server.Address()
+	// Check that the server is ready before returning.
+	c, err := http.NewClient(n.APIURL)
+	if err != nil {
+		return err
+	}
+	err = c.HealthCheck(ctx)
```
AndrewSisley, line 52: question: Is this what allowed the removal of the sleep in `updateNetworkState`?

todo: Please document why we are making this call here.

cli/utils.go
AndrewSisley, line 100: question: Do you imagine production code use cases for this chan?

AndrewSisley (approved):
LGTM, the code is quite a bit simpler than the horrible test hack that existed before (thanks for cleaning that up), although this was done by adding (some) complexity to the production code.

question: What motivated you to deal with this now?

fredcarle:
> question: What motivated you to deal with this now?

I was playing around with this while on vacation so I decided to do it first thing as a way to get back into things.

cli/utils.go
fredcarle: If devs decide to use `node` in a goroutine in their project and they want to get back information from the `node`.

node/node_api.go
fredcarle: No. I think with the changes we've made over the last few months, it's no longer necessary.

http/client.go
fredcarle: Yes but I didn't want to introduce it in this PR. Planning to add it later. #3708

cli/utils.go
AndrewSisley: This is defined in the CLI package though, which they would probably not be using if they were using `node`?

http/client.go
fredcarle: It's not. In order to ensure that the `node.Start` method returns only when the server is fully started, we need to make a call to the server. Basically what happens when starting the server is that we instantiate the listener which secures the port number and allow clients to connect. The `server.Serve()` method is the one that enables handling the clients' requests and that can take a bit of time to setup and is done in a goroutine because it blocks until the server is closed. So to ensure that `server.Serve()` is all good to go, we send a `HealthCheck` request that returns only when the server is truly operational.

Doing this ensure that subsequent operations will behave as expected.

node/node_api.go
fredcarle: > todo: Please document why we are making this call here.

There is already a comment for that but I can expand it a little.

AndrewSisley: > There is already a comment for that but I can expand it a little.

Ah, my brain didn't register the comment on the constructor call as relevant for this line for some reason :)  I'm happy as-is.

cli/utils.go
AndrewSisley: >  Let me know if you're ok with this.

LGTM, thanks :)

> ...

Is a nice thought, although I have mixed feelings - mostly I think because I see `node` as a simplified wrapper around the stuff that *does* allow the more complex customisation - if they want to play around with different event busses they should probably not be using the `node` package.

PR #3710, opened 2025-05-15
refactor: Remove collection field id

### LINKED ISSUE #3699: Remove CollectionFieldDescription.ID

AndrewSisley:
## Relevant issue(s)

Resolves #3699

## Description

Removes the public `CollectionVersion` field id property, replacing it with an internal short-id.

The internal code has gotten slightly more complex as a result.  Some of this complexity might be removed soon, once we gain (public) collection field CIDs as the collection short id will the no longer be required in order to fetch the field short id.  This is caveated by the fact that I'm still unsure if I want to host the field CIDs on the collection version fields (would be consistent with the collection version), or keep them as internal-only (the links between field stuff get more complex if we do this as we have to rely on field Name+CollectionVersionID instead of just field CID).


internal/keys/systemstore_field_id.go
```diff
@@ -0,0 +1,79 @@
...
+var _ Key = (*SchemaRootKey)(nil)
+
+func NewFieldID(collectionShortID uint32, fieldID string) FieldID {
+	return FieldID{
+		CollectionShortID: collectionShortID,
+		FieldID:           fieldID,
+	}
+}
+
+func NewFieldIDP(collectionShortID uint32) FieldID {
```
fredcarle, line 35: suggestion: This function is confusing. Probably mostly because of the name. It makes you wonder what `P` stands for. Either `NewFieldIDPrefix` or just using `NewFieldID(1, "")` would be better.

internal/db/id/field.go
```diff
@@ -12,69 +12,160 @@ package id
...
+
+	cache := getFieldShortIDCache(ctx)
+	shortID, ok := cache[uniqueKey]
+	if ok {
+		return shortID, nil
+	}
+
+	// If we miss the cache, load the entire collection's worth - it is almost always
+	// going to be more efficient than loading the field short-ids one by one, and we'll
+	// usually want most of them.
```
fredcarle, line 44: question: Have you thought about caching these globally instead of per call? I'm thinking even caching the collection descriptions would improve performance.

internal/db/id/field.go
```diff
@@ -12,69 +12,160 @@ package id
...
+		if key.FieldID == fieldID {
+			shortID = sID
+		}
+	}
+
+	return shortID, iter.Close()
+}
+
+// SetShortFieldID sets and stores the short field id, if it does not already exist.
+func SetShortFieldID(
```
fredcarle, line 90: suggestion: Unless you know that this will be used outside this package eventually, I suggest making it private.

internal/db/fetcher/versioned.go
```diff
@@ -381,7 +386,7 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 			keys.DataStoreKey{
 				CollectionShortID: shortID,
 				DocID:             string(block.Delta.GetDocID()),
-				FieldID:           fmt.Sprint(field.ID),
+				FieldID:           fmt.Sprint(fieldShortID),
```
fredcarle, line 389: todo: Please use `strconv.Itoa` instead of `fmt.Sprint`. It has 0 allocations instead of 2 and is about an order of magnitude faster.

internal/db/merge.go
```diff
@@ -489,7 +495,7 @@ func (mp *mergeProcessor) initCRDTForType(ctx context.Context, crdtUnion crdt.CR
 			keys.DataStoreKey{
 				CollectionShortID: shortID,
 				DocID:             docID,
-			}.WithFieldID(fd.ID.String()),
+			}.WithFieldID(fmt.Sprint(fieldShortID)),
```
fredcarle, line 498: todo: Same as above.

internal/lens/fetcher.go
```diff
@@ -316,7 +317,13 @@ func (f *lensedFetcher) updateDataStore(ctx context.Context, original map[string
...
 			continue
 		}
-		fieldKey := datastoreKeyBase.WithFieldID(fieldDesc.ID.String())
+
+		fieldShortID, err := id.GetShortFieldID(ctx, shortID, fieldDesc.Name)
+		if err != nil {
+			return err
+		}
+
+		fieldKey := datastoreKeyBase.WithFieldID(fmt.Sprint(fieldShortID))
```
fredcarle, line 326: todo: Same as above.

internal/planner/mapper/mapper.go
```diff
@@ -365,6 +366,15 @@ func resolveAggregates(
 	def client.CollectionDefinition,
 	store client.Store,
 ) ([]Requestable, error) {
+	var collectionShortID uint32
+	if def.Version.CollectionID != "" {
```
fredcarle, line 370: question: Could it be `""` and what happens if it is?

internal/keys/systemstore_field_id.go
AndrewSisley: I meant to change this and then forgot about this, and missed it scanning through :) Will change.

- [ ] `NewFieldIDP`

internal/db/id/field.go
AndrewSisley: That would get fairly complicated RE transactions.  Long term we'll probably want to do that, but right now I do not think it is worth the dev-effort.  Doing it this way in the medium term should not waste much time (actually implementing the cache was a handful of minutes here).

internal/db/fetcher/versioned.go
AndrewSisley: This is not new, but I might have a look in this PR.

internal/planner/mapper/mapper.go
AndrewSisley: Yes, this surprised me too - I think it is `""` for commit queries and some other edge cases.

AndrewSisley: The `collectionShortID` block is only hit when this function is called with a collection.  It would panic before the `collectionShortID` fetching when retrieving a field description from a nil/empty def

internal/db/id/field.go
AndrewSisley: I think it will be, and the func below that consumes it will either be removed or (less likely) moved.

internal/db/fetcher/versioned.go
AndrewSisley: My attention needs to be on Lens in order to unblock Shinzo so I'm going to leave this as-is.

fredcarle (approved):
LGTM

PR #3716, opened 2025-05-16
refactor: Store index description in collection version

### LINKED ISSUE #3715: Sec. index: Store sec. index description in CollectionVersion
At the moment secondary indexes are serialized/deserialized separately from collection version. We should change it.

islamaliev:
## Relevant issue(s)

Resolves #3715

## Description

Store secondary index description in `CollectionVersion` and rely on it's state instead of storing them separately.


internal/db/collection.go
```diff
@@ -221,13 +225,10 @@ func (db *DB) getAllActiveDefinitions(ctx context.Context) ([]client.CollectionD
 			return nil, err
 		}
 
-		collection := db.newCollection(col, schema)
-
-		err = collection.loadIndexes(ctx)
```
AndrewSisley, line 226: question: Can we now remove `loadIndexes` and `fetchCollectionIndexDescriptions`?  If not, why not?

islamaliev: are you just half way thorough the PR? They are removed :)

internal/db/collection.go
```diff
@@ -57,11 +57,19 @@ type collection struct {
...
 // newCollection returns a pointer to a newly instantiated DB Collection
-func (db *DB) newCollection(desc client.CollectionVersion, schema client.SchemaDescription) *collection {
-	return &collection{
+func (db *DB) newCollection(desc client.CollectionVersion, schema client.SchemaDescription) (*collection, error) {
+	col := &collection{
 		db:  db,
 		def: client.CollectionDefinition{Version: desc, Schema: schema},
 	}
+	for _, index := range desc.Indexes {
+		colIndex, err := NewCollectionIndex(col, index)
```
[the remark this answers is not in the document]
islamaliev: no. What's a benefit of doing this?

internal/core/parser.go
```diff
@@ -32,6 +32,11 @@ type SchemaDefinition struct {
 	Body []byte
 }
 
+type ParsedCollection struct {
```
AndrewSisley, line 35: thought: I do not think this should exist, firstly, the `parsed` prefix is misleading as it implies that everything else is not parsed (they are).  

Secondly, it is preserving a lot of the oddities around secondary index definitions where creating/editing them is very special compared to everything else on collection - this is very visible in `db.createCollection`, a function which will hopefully become public at some point (similar thoughts RE the GQL parsing stuff too).

suggestion: If keeping this in the short term, please consider renaming it.

suggestion: Please try and remove this.  The PR overall makes the code more robust, however it is retaining some of the underlying secondary-index exceptionalism that caused the original 'loadIndexes` issue, and making it visible in the signatures of some fairly important, semi-public, global-context, functions.

AndrewSisley (approved):
I'm approving as whilst I think some of the changes will need to be reworked in the future, it does solve the primary issue and the codebase is much better off for this change as-is.

Thank you very much for sorting this out Islam!

internal/core/parser.go
islamaliev: the problem is that so far the schema that is being parsed (with `client add schema`) happens to align with what we store. But for example with with searchable encryption the syntax is similar to secondary index, e.g. you just specify `@encryptedIndex` on a field. The parser will spit out a description for SE that is attached to a field with a specific name. But we don't want to store is like this. What we store at the moment is the field id (and later will be CID-based id).

So that parsed object and stored one are different things. There is some processing that needs to be done before it's stored.

PR #3736, opened 2025-05-26
docs: Add data flow documentation

islamaliev:
Add new documentation that explains how data flows through DefraDB, from client requests to distributed synchronization across peers.

This documentation serves as a technical reference for developers working with DefraDB, providing clear insights into the system's distributed architecture.

### DIFF
--- a/internal/db/data_flow.md
+++ b/internal/db/data_flow.md
@@ -1,6 +1,6 @@
 # DefraDB Data Flow: Create Document Request Lifecycle
 
-This document provides a comprehensive overview of how data flows through DefraDB, from initial client request to distributed synchronization across peers. It serves as a technical reference for developers at all levels working with the codebase.
+This document provides a comprehensive overview of how data flows through DefraDB for creating and updating a document,from initial client request to distributed synchronization across peers. It serves as a technical reference for developers at all levels working with the codebase.
 
 ## Table of Contents
 
--- a/internal/db/data_flow.md
+++ b/internal/db/data_flow.md
@@ -1,4 +1,4 @@
-# DefraDB Data Flow: Request Lifecycle
+# DefraDB Data Flow: Create Document Request Lifecycle
 
 This document provides a comprehensive overview of how data flows through DefraDB, from initial client request to distributed synchronization across peers. It serves as a technical reference for developers at all levels working with the codebase.
 
@@ -245,18 +245,13 @@ DefraDB uses a multi-store architecture to separate different types of data for 
 - `internal/keys/` - Key formatting for each store type
 
 The `MultiStore` interface defines all available stores:
-
-```go
-type MultiStore interface {
-    Rootstore()   DSReaderWriter  // Base transactional store
-    Datastore()   DSReaderWriter  // Document data (/data namespace)
-    Encstore()    Blockstore      // Encryption keys (/enc namespace)
-    Headstore()   DSReaderWriter  // Document heads (/head namespace)
-    Peerstore()   DSReaderWriter  // Peer information (/peers namespace)
-    Blockstore()  Blockstore      // IPLD blocks (/blocks namespace)
-    Systemstore() DSReaderWriter  // System metadata (/system namespace)
-}
-```
+- Rootstore - base transactional store
+- Datastore - document data (/data namespace)
+- Encstore - encryption keys (/enc namespace)
+- Headstore - document heads (/head namespace)
+- Peerstore - peer information (/peers namespace)
+- Blockstore - iPLD blocks (/blocks namespace)
+- Systemstore - system metadata (/system namespace)
 
 ### AddDelta Function Flow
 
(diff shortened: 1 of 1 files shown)


AndrewSisley:
> You could make the same argument about any docs - they all need maintenance, but we still write them because they're valuable.

Other docs are more narrow in scope and are located alongside the code they describe - this doc exists away from what it is describing and yet is very tightly coupled to the internals.  It is significantly more fragile than, for example, the ACP doc.

> This doc describes the core architecture of DefraDB, which honestly isn't going to change that much.

It changes pretty regularly IMO - the CRDT flow your opening diagram was changed two weeks ago, and the P2P is probably less than a year old.  We should be encouraging the ease of change RE the core internal architecture to that it can adapt itself as pain free as possible to new ideas and requirements - this documents adds overhead to changing it.

> The real value is for onboarding

I do not think this has been a significant drain on resources the last 4 years, very few people have been onboarded.  I would fully expect the maintenance cost of always keeping this document up to date to significantly outweigh the time saved by it's always-updated existence within the repo.

>  I'm fine spending 15 minutes doing it.

Part of what I'm driving at is that it doesn't really matter how enthusiastic any single team member is in maintaining this, given its very broad scope, and locating it within the repo, everyone needs to be on board with maintaining it.  And even if everyone is on board with that, the lack of build checks will mean it will regularly be broken accidentally.

internal/db/data_flow.md
```diff
@@ -0,0 +1,544 @@
...
+type MultiStore interface {
+    Rootstore()   DSReaderWriter  // Base transactional store
+    Datastore()   DSReaderWriter  // Document data (/data namespace)
+    Encstore()    Blockstore      // Encryption keys (/enc namespace)
+    Headstore()   DSReaderWriter  // Document heads (/head namespace)
+    Peerstore()   DSReaderWriter  // Peer information (/peers namespace)
+    Blockstore()  Blockstore      // IPLD blocks (/blocks namespace)
+    Systemstore() DSReaderWriter  // System metadata (/system namespace)
+}
+```
```
AndrewSisley, line 259: thought: I disagree with the inclusion of all the code blocks present in this document, the read should just go and read the actual functions instead of us maintaining them twice.

IMO, this `MultiStore` one is particularly bad, as including it here appears to have very little benefit, whilst at the same time, discourages the urge for the user to actually go and look at the actual code and read the actual documentation on it, as well as all the other duplication/maintenance issues talked about earlier.

The goal of this document is largely to make readers comfy exploring the codebase - IMO it should very much be encouraging them to visit the actual code, otherwise they will exist in their own little fake safe-space until the end of onboarding/this-doc until being suddenly and violently cast into the real world all in one go.

internal/db/data_flow.md
```diff
@@ -0,0 +1,544 @@
...
+   - Atomic updates ensure consistency
+   - Clear error propagation to clients
+
+## Summary
+
+The DefraDB data flow represents a sophisticated distributed system that orchestrates multiple components to provide a decentralized database experience.
+
+### Design Benefits
+
+This architecture enables DefraDB to:
```
AndrewSisley, line 530: thought: The architecture does not really enable these things, the code does.  This section feels like marketing fluff/self-congratulations and encourages readers to ignore them and devalues the rest of the document.

You could perhaps keep these points but present them as a feature/requirement set of Defra.

internal/db/data_flow.md
```diff
@@ -0,0 +1,544 @@
...
+   - Encrypts block content (`internal/encryption/encryptor.go`)
+   - Replaces plaintext with ciphertext
+3. **Signing**: If enabled, signs the block for integrity verification
+4. **Storage**:
+   - **Encrypted/signed block → Blockstore**: The transformed block (with encryption and/or signature) is stored in the blockstore. This is the version that gets distributed to other peers during synchronization.
+   - **Original block → Local processing**: The unencrypted block data is used for local operations like updating indexes and merging CRDT states. This ensures the local node can work with the data without decryption overhead.
+5. **Local Processing**:
+   - Merges CRDT changes into the current document state
+   - Updates document heads in headstore to track the latest version
+   - Triggers index updates for query optimization
```
AndrewSisley, line 295: suggestion: It would be worth adding a point here (under `5.`) to note that this is where stuff is written to the datastore too, it feels like an important omission.

AndrewSisley (approved):
Discussed in stand up.

We agreed to merge it and see what happens.  The explicit func/file names are to prompt AI and Islam wants to play around with that.

Of the document itself, I think it is very well structured and presented.

I've tried not to comment too much on the body, as I have a tendency to want considerably briefer documents than perhaps the majority of the team.  I have instead highlighted my two largest concerns outside of the maintenance-space.

internal/db/data_flow.md
```diff
@@ -0,0 +1,544 @@
+# DefraDB Data Flow: Request Lifecycle
```
AndrewSisley, line 1: todo: Mentioned by Fred in the standup, and I agree - this does need to be renamed.  I think (for an overview doc) the current scope is good, but the name does not reflect this.  Neither does the first paragraph.

internal/db/data_flow.md
islamaliev: done

Issue #3754, opened 2025-06-03
Doc Encryption: enable doc encryption for embedded go


islamaliev:
At the moment there is no way for embedded go client to enable encryption when creating a document.

PR #3755, opened 2025-06-03
fix: Enable doc encryption for embedded client

islamaliev:
## Relevant issue(s)

Resolves [#3754](https://github.com/sourcenetwork/defradb/issues/3754)

## Description

Add create doc options to collection create-doc-related methods. 
The options allow specifying parameters for doc encryption.

As opposed to the previous approach with passing the context, this moves parameters handing into the interface and unifies it for all clients.

### DIFF
--- a/http/client_collection.go
+++ b/http/client_collection.go
@@ -23,7 +23,6 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 
 	"github.com/sourcenetwork/defradb/client"
-	"github.com/sourcenetwork/defradb/internal/encryption"
 )
 
 var _ client.Collection = (*Collection)(nil)
@@ -61,6 +60,7 @@ func (c *Collection) Definition() client.CollectionDefinition {
 func (c *Collection) Create(
 	ctx context.Context,
 	doc *client.Document,
+	opts ...client.DocCreateOption,
 ) error {
 	methodURL := c.http.apiURL.JoinPath("collections", c.Version().Name)
 
@@ -74,7 +74,7 @@ func (c *Collection) Create(
 		return err
 	}
 
-	setDocEncryptionFlagIfNeeded(ctx, req)
+	setDocEncryptionFlagIfNeeded(req, opts...)
 
 	_, err = c.http.request(req)
 	if err != nil {
@@ -87,6 +87,7 @@ func (c *Collection) Create(
 func (c *Collection) CreateMany(
 	ctx context.Context,
 	docs []*client.Document,
+	opts ...client.DocCreateOption,
 ) error {
 	methodURL := c.http.apiURL.JoinPath("collections", c.Version().Name)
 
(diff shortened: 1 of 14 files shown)


http/client_collection.go
```diff
@@ -122,16 +123,18 @@ func (c *Collection) CreateMany(
...
-	encConf := encryption.GetContextConfig(ctx)
-	if encConf.HasValue() {
-		q := req.URL.Query()
-		if encConf.Value().IsDocEncrypted {
-			q.Set(docEncryptParam, "true")
-		}
-		if len(encConf.Value().EncryptedFields) > 0 {
-			q.Set(docEncryptFieldsParam, strings.Join(encConf.Value().EncryptedFields, ","))
-		}
+func setDocEncryptionFlagIfNeeded(req *http.Request, opts ...client.DocCreateOption) {
```
jsimnz, line 126: suggestion: The `opts` param for the internal util function doesn't need to use a variadic type here, can just be a regular slice. 

The public APIs use it so that it becomes optional from the external POV, but once the `...client.DocCreateOption` type is being passed around internally, it can be treated as a regular `[]client.DocCreateOption`

The same applies to the `setContextDocEncryption` function

jsimnz (approved):
LGTM, minor suggestion raised on some of the param types.

One thing I wanted to highlight as a general convo (non blocker to this PR) - Long term, how much do we want to expose the *manual* API for (field-level) doc encryption vs soley relying on the policy/schema to determine and inform the rest of the system when docs and fields should be encrypted. (We can have this discussion in discord if prefered)

AndrewSisley:
> Are you suggesting abandoning the idea of having field-level encryption individually configurable?

This does sound like an easily avoidable user-complication, but I've not thought a whole load about it and I cant remember a previous discussion around it.

internal/db/collection.go
```diff
@@ -508,6 +513,16 @@ func (c *collection) create(
 	return c.registerDocWithACP(ctx, doc.ID().String())
 }
 
+func setContextDocEncryption(ctx context.Context, opts ...client.DocCreateOption) context.Context {
+	createOptions := client.DocCreateOptions{}
+	createOptions.Apply(opts...)
+	if !createOptions.EncryptDoc && len(createOptions.EncryptedFields) == 0 {
+		return ctx
+	}
+	ctx = encryption.SetContextConfigFromParams(ctx, createOptions.EncryptDoc, createOptions.EncryptedFields)
```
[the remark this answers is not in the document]
islamaliev: the way it was didn't allow users of embedded client to created encrypted docs. You had to add the encryption configuration to the context before calling collection.Create with some internal values. Now all clients are forced to pass the encryption config in a unified way.

http/client_collection.go
islamaliev: done

PR #3757, opened 2025-06-04
test: Add Datastore test action for direct value assertions

### LINKED ISSUE #3756: Allow asserting datastore state in integration tests
For some integration tests it would be very beneficial to assert that some key-value records are stored in the datastore.

islamaliev:
## Relevant issue(s)

Resolves #3756

## Description

This PR introduces a new Datastore test action to the DefraDB integration test framework that enables direct fetching and assertion of values from the datastore.
The action provides a flexible builder pattern for constructing keys.

The implementation adds a Datastore struct that accepts a KeyBuilder interface for key construction and uses gomega.OmegaMatcher for value assertions.
This allows tests to verify the presence or absence of specific keys in the datastore and assert their values with flexible matchers.

A builder pattern API is provided through the NewKey() factory function, which supports building different key types:

- DatastoreDoc() builder creates keys for document field values
- DatastoreIndex() builds keys for index entries

Both builders use collection and document indices that are resolved to actual IDs at runtime, maintaining consistency with the existing test framework patterns.


tests/integration/store_key.go
```diff
@@ -0,0 +1,208 @@
...
+package tests
+
+import (
+	"strconv"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/db"
+	"github.com/sourcenetwork/defradb/internal/db/id"
+	"github.com/sourcenetwork/defradb/internal/keys"
```
AndrewSisley, line 20: thought: I understand what and why you are doing, but this makes me quite uncomfortable.

Our tests should not be dependent on internal code (yes bits already are, but lets not make it worse), and relying on internals creates invasive and fragile tests, as well as unit tests posing as integration tests.

We could argue that `keys` is a bit of a grey area as it is kind of in the file system, which kind of makes it public, which kind of means `keys` should be moved out of internal.

If we end up keeping things roughly as-is, I think this new stuff should come with a big warning in its documentation discouraging its use, so that it is only used when absolutely unavoidable.

islamaliev: hey Andy, I'm sure we are both on the page in seeing accessing the storage as observable behaviour, because in theory one can just connect to the database (like badger) and directly access the data circumventing defradb.

Your concern is about using internal keys. I totally agree with you in that tests should have as little knowledge about internal state is possible.

In this case all the db internal code is internal to the testing framework and tests themselves don't have access to them, as they operate via KeyBuilder's. In case we decide to change the way we form the keys, we will have to change it in just one place.

In theory we could build keys from scratch in the testing framework, so we don't depend on the internal code, but it seems to be less optimal.

This test action is not supposed be use extensively. They just might be handy for few scenarios when you want to make sure some data is stored or not stored in when for example you can't assert it with existing capabilities like fetching docs with normal query. This might be particularly useful for my searchable encryption tests where you can't run normal queries on the untrusted replicator node, but want to make sure it stored (or didn't) data correctly.

tests/clients/cli/wrapper.go
```diff
@@ -525,6 +525,10 @@ func (w *Wrapper) Encstore() datastore.Blockstore {
 	return w.node.DB.Encstore()
 }
 
+func (w *Wrapper) Datastore() datastore.DSReaderWriter {
+	return w.node.DB.Datastore()
```
AndrewSisley, line 529: todo: This, and the tests passing for CLI/Http feels wrong and misleading - the test code is passing, but the production code will panic.

IMO the tests need to be limited by `SupportedClientTypes` to embedded only, and this function should panic.

islamaliev: I think this is something you are proposing for another ticket. Here I've just done the same as we do for other methods. If it's the correct way or not shouldn't be part of this task.

tests/integration/test_case.go
```diff
@@ -883,3 +886,33 @@ type VerifyBlockSignature struct {
...
+
+// Datastore is an action that fetches a value from the datastore for a given key.
+type Datastore struct {
+	// NodeID may hold the ID (index) of a node to fetch from.
+	// If not provided, the fetch will be done on all nodes.
+	NodeID immutable.Option[int]
+
+	// Key is built using the KeyBuilder interface.
+	// Use KeyFactory with NewKey() to create a new key builder.
+	Key KeyBuilder
```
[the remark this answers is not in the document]
AndrewSisley: It should be possible to describe and test real behaviour.  Utilities can be added to aid that (like `KeyBuilder`), but it should always be possible to describe and test the actual production interface.

`[]byte("/foo/bar/dnsagag/njsgsgag")` is pretty descriptive, test don't have to type out the raw machine bytes.

AndrewSisley (changes requested):
There are a couple of todos that I think need resolving if we as a team agree to the idea of testing via store in this (or similar) form, so I'm going to Request Changes.

tests/clients/cli/wrapper.go
islamaliev: I don't understand what you mean by "bad tests". Which tests are bad?
In the DB interface we exposed all types of storages, but Datastore, so I just added the missing one and implemented it the same way all others are implemented. 

Is the fact that this one is used by the testing framework bothering you? I can find another way of retrieving datastore. But we already use production-panicking stuff in tests like `MaxTxnRetries` which makes all integration tests "bad".

This action is not client-specific, it should work for all clients, because they all have datastore.

This is not the problem of testing framework, but of our interface that is hopefully going to be improved with the refactoring that @fredcarle is working on.

AndrewSisley: > Which tests are bad?

If you go by test name - all of them.  They say the test passes with the CLI and http clients, which is wrong.  The tests are testing test cod, not production.

AndrewSisley:
Hey Islam, I'm closing this PR as I think we agreed to a few weeks ago in a standup.  You should still be able to reference or re-open it in the the 'closed' tab if you need to in the future.

PR #3776, opened 2025-06-11
fix: Correct index fetching order

### LINKED ISSUE #3774: Ordering ASC on a DESC secondary index with a limit returns the wrong results
It appears to apply the limit, but the order matches the index, not the query.
```
func TestFoo(t *testing.T) {
	req := `query {
		User(order: {age: ASC}, limit: 3) {
			name
			age
		}
	}`
	test := testUtils.TestCase{
		Actions: []any{
			testUtils.SchemaUpdate{
				Schema: `
					type User {
						name: String 
						age: Int @index(direction: DESC)
					}`,
			},
			testUtils.CreatePredefinedDocs{
				Docs: getUserDocs(),
			},
			testUtils.Request{
				Request: req,
				Results: map[string]any{
					"User": []map[string]any{
						{
							"name": "Chris",
							"age":  int64(55),
						},
						{
							"name": "Keenan",
							"age":  int64(48),
						},
						{
							"name": "Roy",
							"age":  int64(44),
						},
					},
				},
			},
			testUtils.Request{
				Request:  makeExplainQuery(req),
				Asserter: testUtils.NewExplainAsserter().WithLimit().WithIndexFetches(3),
			},
		},
	}

}
```

islamaliev:
## Relevant issue(s)

Resolves #3774

## Description

Correctly determine if based on query order and ordering of the field index the fetching should be in reverse order.

### DIFF
--- a/internal/db/fetcher/indexer_iterators.go
+++ b/internal/db/fetcher/indexer_iterators.go
@@ -454,7 +454,7 @@ func (f *indexFetcher) tryCreateOrderedIndexIterator() (indexIterator, error) {
 }
 
 // isIndexOrdered checks if the index is ordered by the first field in the ordering array.
-// The first return value specifies ordered-by field is indexed
+// The first return value specifies if ordered-by field is indexed
 // The second one specifies if the index is ordered in the opposite direction to the given ordering, i.e. reversed.
 func (f *indexFetcher) isIndexOrdered() (bool, bool) {
 	if len(f.ordering) > 0 {
@@ -462,7 +462,8 @@ func (f *indexFetcher) isIndexOrdered() (bool, bool) {
 		if found {
 			indexField := f.indexDesc.Fields[0]
 			if indexField.Name == orderField {
-				return true, f.ordering[0].Direction == mapper.DESC && !indexField.Descending
+				isOrderDesc := f.ordering[0].Direction == mapper.DESC
+				return true, isOrderDesc != indexField.Descending
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/db/fetcher/indexer_iterators.go
```diff
@@ -454,15 +454,16 @@ func (f *indexFetcher) tryCreateOrderedIndexIterator() (indexIterator, error) {
 }
 
 // isIndexOrdered checks if the index is ordered by the first field in the ordering array.
-// The first return value specifies ordered-by field is indexed
+// The first return value specifies if ordered-by field is indexed
 // The second one specifies if the index is ordered in the opposite direction to the given ordering, i.e. reversed.
 func (f *indexFetcher) isIndexOrdered() (bool, bool) {
```
AndrewSisley, line 459: question: What happens in a composite if the index order is reversed compared to the query? What happens if the first field's index order matches the query, but subsequent fields do not?

shahzadlone (approved):
LGTM

internal/db/fetcher/indexer_iterators.go
islamaliev: the subsequent fields will be just ignored

AndrewSisley: Confirmed on latest develop (after merging this). It is still broken: https://github.com/sourcenetwork/defradb/issues/3779

islamaliev: oh, I see what you mean now. Thanks, I will take a look

PR #3787, opened 2025-06-16
feat: Provide C Bindings

### LINKED ISSUE #3362: C bindings

ChrisBQu:
## Relevant issue(s)

Resolves #3362

## Description

There are several things occurring in this PR. 

**- C Bindings:**

There exists a directory, cbindings, which contains various .go files, as well as a `defra_structs.h` file. These files implement the functionality that will be exported to the shared object. In other words, these are the functions that will be made available to the users in C. Currently, they are grouped in a similar to way as the subcommands in the Defra CLI, although there has already been some feedback regarding whether or not that should be the case. These functions will likely be re-grouped, re-named, or both.

To allow for the building, a new section of the makefile has been added, `build-c-shared-linux` which is where current development is focused. Building for Windows and/or Android are not currently included in this PR.

**- Integration Test Support:**

A new directory exists called `tests/integration/clients/c`. This contains code that wraps a large portion of the C binding code for use in our integration tests. There also exists a number of files which contain functions duplicating the behavior of those found in the `cbindings` directory. The reason for this, is that we have a layered problem:

1) The integration tests require that we be able to pass complex objects (such as a Node), which is not possible through calling into a .so
2a) We can, as a workaround, call the functions directly using C-Types, to test the logic...
2b) ...but to be exportable, the functions must exist in a package named "main", which we cannot import and use...
2c) ...and additionally, due to the way Go handles C Types, we cannot pass a C structure from one package to another. 

Therefore, it's a little bit messy, but we have the exported functions, and the appropriate .h file, all in the `tests/integration/clients/c` folder. This allows the integration tests to run. Hopefully, we can find a better approach, for the long-term, moving forward. 

**- Key Takeaway / Where we are at:**

Testing indicates that the C functionality is working well. However, properly incorporating a C client into the integration test suite that we have is a bit hacky. This PR attempts to create a base to work from, where the core logic works correctly, and is testable.

**- Important Note on P2P Functionality:**

Currently, there exist wrapper functions for P2P related functionality (i.e.  P2PsetReplicator). However, network tests are skipped. The architected design of the C bindings relies on a singular globalNode which is accessed and used by all of the exported functions. Even if two client wrappers were instantiated, they would be utilizing the same node which is problematic. 

*This is the biggest open issue to be addressed. But as per the team discussion, this will be addressed in a separate PR*

## Tasks

## How has this been tested?

- WSL


cbindings/node.go
```diff
@@ -0,0 +1,166 @@
...
+	"strconv"
+	"time"
+
+	"github.com/sourcenetwork/defradb/internal/db"
+	"github.com/sourcenetwork/defradb/node"
+
+	netConfig "github.com/sourcenetwork/defradb/net/config"
+)
+
+var globalNode *node.Node
```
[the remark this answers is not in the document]
ChrisBQu: I like this approach. Originally I didn't do it because I am relying on a struct defined in C (C.Result) as the return type. Even if two packages use identical header files, the type will not be treated as the same. So, for example, I could not return a function from the cbindings function, to the stub function, and return it from there to the user. But with a little bit of bridge code, it will be doable. And I think it would clean up the work a great deal.

Good call.

fredcarle (commented):
Thanks for putting this together Chris. Here are a few comments to start. There are a few things I want to investigate before I finalize my review.

cbindings/collection.go
```diff
@@ -0,0 +1,456 @@
...
+	identityStr := C.GoString(cOptions.identity)
+	if identityStr != "" {
+		newctx, err := contextWithIdentity(ctx, identityStr)
+		if err != nil {
+			return ctx, err
+		}
+		return newctx, nil
+	}
+	return ctx, nil
+}
```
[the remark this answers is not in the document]
ChrisBQu: This was done purposefully to factor out some code that has to be repeated in a lot of the function calls. I do think it results in fewer lines of code being repeated, but... not convinced it's necessarily worth having the functions? If you dislike it, I have no push back, and will remove them.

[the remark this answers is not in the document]
fredcarle: I think it's exactly the same number of lines 🙂
```go
// Attach the identity to the context
ctx, err := setIdentityOfCollectionCommand(ctx, cOptions)
if err != nil {
	return returnC(1, err.Error(), "")
}
// vs
ctx, err = contextWithIdentity(ctx, C.GoString(cOptions.identity))
if err != nil {
	return returnC(1, err.Error(), "")
}

// Set the transaction
ctx, err = setTransactionOfCollectionCommand(ctx, cOptions)
if err != nil {
	return returnC(1, err.Error(), "")
}
// vs
ctx, err = contextWithTransaction(ctx, cOptions.tx)
if err != nil {
	return returnC(1, err.Error(), "")
}
```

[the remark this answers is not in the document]
ChrisBQu: Oh I see what you meant, now. I misunderstood before; I thought it was unsafe to call contextWithIdentity with a blank string. This will clean up very nicely then.

cbindings/acp.go
```diff
@@ -0,0 +1,129 @@
...
+	identityStr := C.GoString(cIdentity)
+
+	// Attach the identity to the context
+	newctx, err := contextWithIdentity(ctx, identityStr)
+	if err != nil {
+		return returnC(1, err.Error(), "")
+	}
+	ctx = newctx
+
+	// Set the transaction
```
islamaliev, line 76: suggestion: all these comments don't add any value, they state the obvious thing. We can remove them

cbindings/crypto.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/sourcenetwork/defradb/crypto"
+)
+
+const (
+	KeyTypeEd25519   = "ed25519"
+	KeyTypeSecp256k1 = "secp256k1"
+)
+
+//export cryptoGenerateKey
+func cryptoGenerateKey(cKeyType *C.char) *C.Result {
```
islamaliev, line 33: question: why is this function exposed? We don't have it on the interface. Where is it used?

cbindings/errors.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Schema
+	cerrAddingSchema    string = "error adding schema: %v"
+	cerrGettingSchema   string = "error getting schema: %v"
+	cerrPatchingSchema  string = "error patching schema: %v"
+	cerrSetActiveSchema string = "error setting active version of schema: %v"
+	cerrEmptyPatch      string = "patch cannot be empty"
+
+	// Collection
+	cerrGettingCollection    string = "error getting collection: %v"
```
islamaliev, line 35: suggestion: I'm not sure prefixing these with "c" adding more clarity. It's already in `cbindings` package

cbindings/query.go
```diff
@@ -0,0 +1,137 @@
...
+func getSubscription(id string) (*client.RequestResult, bool) {
+	val, ok := subscriptionStore.Load(id)
+	if !ok {
+		return nil, false
+	}
+	//nolint:forcetypeassert
+	return val.(*client.RequestResult), true
+}
+
+// Helper function
```
islamaliev, line 48: todo: please remove such comments or give a proper comments to functions

tests/clients/c/acp.go
```diff
@@ -0,0 +1,124 @@
...
+	ctx = newctx
+
+	// Set the transaction
+	newctx, err = contextWithTransaction(ctx, cTxnID)
+	if err != nil {
+		return returnC(1, err.Error(), "")
+	}
+	ctx = newctx
+
+	// Try to add the policy
```
islamaliev, line 41: suggestion: reminder again to remove unnecessary comments

tests/clients/c/wrapper_collection.go
```diff
@@ -0,0 +1,592 @@
...
+
+	result := CollectionCreate(cJSON, cIsEncrypted, cEncryptedFields, copts)
+
+	defer C.free(unsafe.Pointer(cIdentity))
+	defer C.free(unsafe.Pointer(cName))
+	defer C.free(unsafe.Pointer(cVersion))
+	defer C.free(unsafe.Pointer(cCollectionID))
+	defer C.free(unsafe.Pointer(cJSON))
+	defer C.free(unsafe.Pointer(cEncryptedFields))
+	defer freeCResult(result)
```
islamaliev, line 98: todo: schedule destructors right after resource is allocated to avoid memory leaks. Which in this case can happen if `err != nil` few lines above.

Also applies to `CreateMany`, `UpdateWithFilter` and `DeleteWithFilter`

cbindings/crypto.go
ChrisBQu: I saw that the CLI had analogous functionality, by calling `defradb keyring generate` so I also exposed the functionality to the C bindings. Do we not want to do this?

cbindings/errors.go
ChrisBQu: Changed!

cbindings/crypto.go
ChrisBQu: I brought this up in today's sync meeting. The decision was remade to remove it.

cbindings/collection.go
```diff
@@ -0,0 +1,456 @@
...
+	"github.com/sourcenetwork/immutable"
+)
+
+// Helper struct
+type docIDResult struct {
+	DocID string `json:"docID"`
+	Error string `json:"error"`
+}
+
+// Helper function
```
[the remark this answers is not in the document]
ChrisBQu: I have made the changes across all files, but I'm interested in any potential feedback regarding where things are at with them now.

cbindings/collection.go
```diff
@@ -0,0 +1,456 @@
...
+		if err != nil {
+			return returnC(1, err.Error(), "")
+		}
+	}
+
+	return returnC(0, "", "")
+}
+
+//export collectionDelete
+func collectionDelete(cDocID *C.char, cFilter *C.char, cOptions C.CollectionOptions) *C.Result {
```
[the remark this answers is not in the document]
nasdf: It's not ideal but I can't think of a better alternative for namespacing functions in C. 

Since our target use case is for Android and iOS, we shouldn't worry too much about the structure of the C bridge layer. When we wrap the C bindings in a higher level language these problems won't exist.

cbindings/query.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Try to marshall the JSON and return it
+	dataMap, ok := res.GQL.Data.(map[string]any)
+	if !ok || dataMap == nil {
+		return returnC(1, "GraphQL response data is nil or invalid.", "")
+	}
+	wrapped := map[string]any{
+		"data": dataMap,
+	}
+	return marshalJSONToCResult(wrapped)
```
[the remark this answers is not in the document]
ChrisBQu: Do you feel strongly about this one? If so, I am willing to change it. I think I have a minor preference towards leaving it as is. For all other function calls you can check result.status to see if the function returned an error, and having this be the only bound function that does not use the same result struct seems a little bit odd to me? But this is not a strong push back.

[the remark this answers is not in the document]
ChrisBQu: Following our discussion today, I have made the suggested change.

cbindings/collection.go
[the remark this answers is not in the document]
ChrisBQu: As discussed in the sync meeting, we will leave these for the time being.

cbindings/logic/query.go
```diff
@@ -0,0 +1,105 @@
...
+
+	res := globalNode.DB.ExecRequest(ctx, query, opts...)
+
+	// The return is either a subscription ID, or a GQL result. The status indicates
+	// which: 0 for GQL, 2 for subscription. 1 is not used because this cannot error; the
+	// error is part of the GQL result, to be GQL-compliant.
+	if res.Subscription != nil {
+		id := storeSubscription(res)
+		return returnGoC(2, "", id)
+	}
```
fredcarle, line 103: todo: This will leak go routines that are generated from a subscriptions because it provides no way to tell the go routine that it's no longer necessary. The only way to do this is to close the provided context. 

The way I suggest to do this is to create a new type that looks like this:
```go
type subscription struct {
	ctxCancel context.CancelFunc
	resultChan <-chan client.GQLResult
}
```

Instead of creating the context above with `ctx := context.Background()`, use `ctx, cancel := context.WithCancel(context.Background())`. Now if you use the `subscription` type with `storeSubscription`, you will then be able to call  `ctxCancel()`  within `removeSubscription` before removing it from the `sync.Map`.

cbindings/collection.go
```diff
@@ -0,0 +1,456 @@
...
+
+	// Range over the chan, filling out an array of docIDResults (ID/error pairs)
+	var results []docIDResult
+	for doc := range docCh {
+		result := docIDResult{
+			DocID: doc.ID.String(),
+		}
+		if doc.Err != nil {
+			result.Error = doc.Err.Error()
+		}
```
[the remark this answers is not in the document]
fredcarle: The thing is the two are different in that the CLI is streaming the results to the terminal and the C binding is returning an array of results. That being said, if you think its ok to return an array of results with the last item containing the error, if there is one, then I'm fine with leaving it as is (same as it is currently in the CLI).

fredcarle (approved):
Thanks for all the exploratory work involved for getting this done. It's a nice initial introduction of the C binding. Before merging, rhere is one todo that you'll see bellow and I would like to resolve the set-on-context function calls too that was mentionned in a previous comment.

islamaliev (changes requested):
Looks great so far. I have some minor request and some questions that I'd like to be clarified before the merge.

cbindings/bridge/main.go
```diff
@@ -0,0 +1,442 @@
...
+	return returnC(gcr)
+}
+
+//export P2PdocumentSync
+func P2PdocumentSync(cCollection *C.char, cDocIDs *C.char, cTxnID C.ulonglong, cTimeout *C.char) *C.Result {
+	gcr := cbindings.P2PdocumentSync(C.GoString(cCollection), C.GoString(cDocIDs), uint64(cTxnID), C.GoString(cTimeout))
+	return returnC(gcr)
+}
+
+func main() {}
```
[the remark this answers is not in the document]
ChrisBQu: Done.

cbindings/logic/node.go
```diff
@@ -23,15 +23,9 @@ import (
...
+var GlobalNodes = make(map[int]*node.Node)
 
-// GetGlobalNode is a function to accommodate integration test support
-// It is not made available as an exported C function
-func GetGlobalNode() *node.Node {
-	return globalNode
-}
-
-func NodeInit(cOptions GoNodeInitOptions) GoCResult {
+func NodeInit(n int, cOptions GoNodeInitOptions) GoCResult {
```
islamaliev, line 28: suggestion: why not to have something like `NewNode(cOptions GoNodeInitOptions) (int, GoCResult) that automatically generates the next index?

tests/clients/c/wrapper.go
```diff
@@ -30,18 +30,23 @@ import (
 	"github.com/sourcenetwork/immutable"
 )
 
+var wrapperCount int = 0
```
islamaliev, line 33: todo: please make it thread-safe or at least leave a comment why it's not done

cbindings/logic/node.go
ChrisBQu: I felt that the caller of the function should have control over which index is assigned to which node. There is definitely room for discussion here, but its also likely to be a detail that is inconsequential as we wrap in a higher level language. Let's keep this in mind for the time being, but I'm going to first focus on addressing the other comments before this one.

cbindings/logic/node.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	ctx := context.Background()
+
+	if globalNode != nil {
+		err := globalNode.Close(ctx)
+		if err != nil {
+			return returnGoC(1, fmt.Sprintf(errClosingNode, err), "")
+		}
+		globalNode = nil
```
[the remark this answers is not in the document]
ChrisBQu: I have, today, made changes so that instead of a single globalNode we use a map of nodes. Access to it uses a mutex. I'm interested in knowing if you are happy with where it is at now.

tests/clients/c/utils.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	opName := ""
+	if config.OperationName != "" {
+		opName = config.OperationName
+	}
+
+	varsJSON := ""
+	if config.Variables != nil {
+		data, _ := json.Marshal(config.Variables)
```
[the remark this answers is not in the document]
ChrisBQu: This seems like the way. Done.

tests/clients/c/wrapper.go
ChrisBQu: Good call. Done.

cbindings/logic/query.go
ChrisBQu: I have incorporated this approach, and made the changes.

cbindings/logic/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/crypto"
+	"github.com/sourcenetwork/defradb/internal/datastore"
+)
+
+const (
+	KeyTypeEd25519   = "ed25519"
+	KeyTypeSecp256k1 = "secp256k1"
+)
```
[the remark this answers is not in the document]
ChrisBQu: Done.

cbindings/logic/query.go
```diff
@@ -0,0 +1,122 @@
...
+	return id
+}
+
+func getSubscription(id string) (*Subscription, bool) {
+	val, ok := subscriptionStore.Load(id)
+	if !ok {
+		return nil, false
+	}
+	//nolint:forcetypeassert
+	return val.(*Subscription), true
```
fredcarle, line 48: thought: I'm pretty certain this will panic as the type will be a `Subscription` and not `*Subscription`.

todo: You can either store a pointer or the struct but make sure the asserted type is correct. Note that it doesn't need to be a pointer since both `cancelFunc` and `resultChan` are references.

cbindings/logic/query.go
```diff
@@ -0,0 +1,122 @@
...
+	res := GetNode(n).DB.ExecRequest(ctx, query, opts...)
+	sub := &Subscription{
+		ctxCancel:  cancelFunc,
+		resultChan: res.Subscription,
+	}
+	// The return is either a subscription ID, or a GQL result. The status indicates
+	// which: 0 for GQL, 2 for subscription. 1 is not used because this cannot error; the
+	// error is part of the GQL result, to be GQL-compliant.
+	if res.Subscription != nil {
+		id := storeSubscription(*sub)
```
fredcarle, line 118: thought: It doesn't make much sense to create a pointer to a struct only to dereference it in the only location it's used. Choose one or the other.

cbindings/logic/query.go
```diff
@@ -27,39 +27,50 @@ import (
 
 var subscriptionStore sync.Map // map[string]*client.RequestResult
 
+type Subscription struct {
```
islamaliev, line 30: todo: please document

cbindings/logic/query.go
```diff
@@ -27,39 +27,50 @@ import (
...
+	return val.(*Subscription), true
 }
 
 func removeSubscription(id string) {
-	subscriptionStore.Delete(id)
+	val, ok := subscriptionStore.LoadAndDelete(id)
+	if ok {
+		//nolint:forcetypeassert
+		sub := val.(*Subscription)
+		sub.ctxCancel()
```
islamaliev, line 56: suggestion: all these subtleties are hard to reach with integration tests I believe. Would be nice to have unit tests for some edge cases. Or at least a github issue to cover it later

tests/clients/c/wrapper.go
```diff
@@ -39,10 +40,10 @@ type CWrapper struct {
 }
 
 func NewCWrapper() *CWrapper {
-	nodeNum := wrapperCount
-	setupTests(nodeNum)
+	nodeNum := atomic.AddInt32(&wrapperCount, 1) - 1
+	setupTests(int(nodeNum))
 	wrapperCount++
```
islamaliev, line 45: todo: I believe this increment is not needed any more

cbindings/logic/node.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	"github.com/sourcenetwork/defradb/internal/db"
+	"github.com/sourcenetwork/defradb/node"
+
+	netConfig "github.com/sourcenetwork/defradb/net/config"
+)
+
+var (
+	GlobalNodes   = make(map[int]*node.Node)
```
islamaliev, line 28: todo: unless you are using some kind of thread-safe map, make it private so that the package is in control of access to it.

ChrisBQu: Done.

tests/clients/c/wrapper.go
ChrisBQu: Good catch. Removed.

cbindings/logic/query.go
ChrisBQu: I've added some verbose comments. Let me know what you think about them when you get a chance.

cbindings/logic/query.go
ChrisBQu: I had this short discussion with Fred on Discord, but I'll put it here for anyone to see.

That cast was a mistake, so that was a good catch, and I've changed it. I also agree that consistency is important. Given the two choices, I decided to store pointers in the store so that getSubscription can return nil instead of a zero'd-out object. I think this makes it a little more clear what the intentions are, but I am open to changing it if anyone feels strongly.

cbindings/logic/query.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	select {
+	case msg, ok := <-res.Subscription:
+		if !ok {
+			removeSubscription(id)
+			return returnGoC(0, "", "") // closed
+		}
+		return marshalJSONToGoCResult(msg)
+	case <-time.After(time.Second):
```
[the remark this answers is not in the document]
ChrisBQu: (Which should not be documented in the comments.)

cbindings/logic/query.go
islamaliev: the comment gives some information, but mostly internal about it's fields and stuff.
Normally documentation should start with some introduction explaining what is the object/struct/interface and what it's for. For individual fields (especially the ones that are private) you can use a comment right above them (not as part of the struct comment).

cbindings/bridge/utils.go
```diff
@@ -0,0 +1,62 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+//go:build cgo
+// +build cgo
+
+package main
+
+/*
+#include "defra_structs.h"
```
shahzadlone, line 17: Todo: remove?

cbindings/bridge/main.go
```diff
@@ -72,6 +72,60 @@ func ACPDeleteRelationship(
...
+}
+
+//export ACPNodeReEnable
+func ACPNodeReEnable(n int, cIdentity *C.char, cTxnID C.ulonglong) *C.Result {
+	gcr := cbindings.ACPNodeReEnable(n, C.GoString(cIdentity), uint64(cTxnID))
+	return returnC(gcr)
+}
+
+//export ACPNodeReRelationshipAdd
+func ACPNodeReRelationshipAdd(
```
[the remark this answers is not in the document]
ChrisBQu: Just a typo; fixed.

cbindings/logic/acp.go
```diff
@@ -0,0 +1,209 @@
...
+
+	result, err := GetNode(n).DB.AddDACActorRelationship(ctx, collectionArg, docIDArg, relationArg, targetActorArg)
+	if err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+
+	return marshalJSONToGoCResult(result)
+}
+
+func ACPDeleteRelationship(
```
shahzadlone, line 67: Todo: these should be renamed likely to indicate they are DAC ACP (to separate them from NAC ACP.

islamaliev (approved):
Approving. There are still some points that I would like you to address, but it would be fine if you created a GitHub issues and linked them in the code so that the PR lands faster.

cbindings/logic/acp.go
```diff
@@ -0,0 +1,209 @@
...
+
+	result, err := GetNode(n).DB.DeleteDACActorRelationship(ctx, collectionArg, docIDArg, relationArg, targetActorArg)
+	if err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+
+	return marshalJSONToGoCResult(result)
+}
+
+func ACPNodeDisable(n int, identityPrivateKey string, TxnID uint64) GoCResult {
```
shahzadlone, line 96: Nitpick: why not stick closer to the original  names like `ACPDisableNAC`

cbindings/logic/acp.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	if err := GetNode(n).DB.ReEnableNAC(ctx); err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+
+	return marshalJSONToGoCResult(client.SuccessResponse{Success: true})
+}
+
+func ACPNodeRelationshipAdd(
```
shahzadlone, line 136: Suggestion: now that you have fixed the name of `ACPAddDACActorRelationship` this should match similar style i.e. `ACPAddNACActorRelationship`

cbindings/logic/acp.go
```diff
@@ -0,0 +1,209 @@
...
+
+	addNACActorRelationshipResult, err := GetNode(n).DB.AddNACActorRelationship(ctx, relationArg, targetActorArg)
+	if err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+
+	return marshalJSONToGoCResult(addNACActorRelationshipResult)
+}
+
+func ACPNodeRelationshipDelete(
```
shahzadlone, line 163: Suggestion: now that you have fixed the name of `ACPDeleteDACActorRelationship` this should match similar style i.e. `ACPDeleteNACActorRelationship`

cbindings/logic/acp.go
```diff
@@ -0,0 +1,209 @@
...
+
+	deleteNACActorRelationshipResult, err := GetNode(n).DB.DeleteNACActorRelationship(ctx, relationArg, targetActorArg)
+	if err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+
+	return marshalJSONToGoCResult(deleteNACActorRelationshipResult)
+}
+
+func ACPNodeStatus(n int, identityPrivateKey string, TxnID uint64) GoCResult {
```
shahzadlone, line 190: Nitpick: similar comment why not `ACPGetNACStatus`? And for all other `AcpNode` functions

tests/integration/signature/commit_test.go
```diff
@@ -49,6 +50,14 @@ func TestSignature_WithCommitQuery_ShouldIncludeSignatureData(t *testing.T) {
 
 	test := testUtils.TestCase{
 		EnableSigning: true,
+		SupportedClientTypes: immutable.Some([]state.ClientType{
+			// C bindings do not support calling functions with non-Secp256k key yet
+			testUtils.CClientType,
```
shahzadlone, line 55: Question: the comment says c bindings doesn't support but then c binding client is listed in the supported types?

tests/integration/db_setup.go
```diff
@@ -52,6 +52,7 @@ func setupNode(
 	s *state.State,
 	identity immutable.Option[acpIdentity.Identity],
 	testCase TestCase,
+	enableNAC bool,
```
shahzadlone, line 55: Question: Why is the needed? Could you not use the opts option?

tests/integration/db_setup.go
```diff
@@ -140,13 +141,13 @@ func setupNode(
 
 	s.Ctx = acpIdentity.WithContext(s.Ctx, identity)
 	err = node.Start(s.Ctx)
-	resetStateContext(s)
+	defer resetStateContext(s)
```
shahzadlone, line 144: Suggestion: I would just move this until after the last call you need the identity at when enable flag is true for nac, rather than defer, because identity shouldn't leak into other operations that are done after start.

tests/integration/signature/commit_test.go
ChrisBQu: Nice catch. I had added it back to make sure the test was being skipped without it; and I guess I forgot to hit save after putting it back! Fixing this one immediately.

shahzadlone:
@ChrisBQu Do you know why the patch coverage is 0 (is that expected?)

image

tests/integration/db_setup_js.go
```diff
@@ -57,17 +57,27 @@ func setupNode(
...
-	c, err := setupClient(s, node)
+
+	// Get enableNAC boolean from the opts
+	tmpOptions := node.DefaultNodeACPOptions()
+	for _, opt := range opts {
+		if opt, ok := opt.(node.NodeACPOpt); ok {
+			opt(tmpOptions)
+		}
+	}
+	enableNAC := tmpOptions.IsEnabled()
```
shahzadlone, line 78: question: Why do you need to do this here?

In `tests/integration/client_setup_js.go`
```go
func setupClient(_ *state.State, node *node.Node, _ bool) (impl clients.Client, err error) {
	return js.NewWrapper(node)
}
```

The value is ignored anyway?

tests/integration/db_setup.go
```diff
@@ -133,20 +133,29 @@ func setupNode(
...
+	// Get enableNAC boolean from the opts
+	tmpOptions := node.DefaultNodeACPOptions()
+	for _, opt := range opts {
+		if opt, ok := opt.(node.NodeACPOpt); ok {
+			opt(tmpOptions)
+		}
+	}
+	enableNAC := tmpOptions.IsEnabled()
+
+	c, err := setupClient(s, nodeObj, enableNAC)
```
shahzadlone, line 157: question: do you think it makes sense to maybe move this further into the `setupClient` function (i.e. pass `opts` as arg and remove the `enableNAC` arg). Specially in light of my other comment, if the `enableNAC` arg is ignored anyway.

ChrisBQu: You are certainly right about the enableNAC arg being ignored, so I can/should/will remove the logic for actually getting it. But I don't think we can move it further into the setupClient function. We could pass in all of the opts as an arg instead, but that's still going to mean passing in one argument that gets ignored for the JS client. If I understand what you mean correctly.

cbindings/logic/acp.go
ChrisBQu: Done

tests/integration/utils.go
```diff
@@ -786,6 +789,7 @@ func setStartingNodes(
 			s,
 			acpIdentity.None,
 			testCase,
+			false,
```
shahzadlone, line 792: todo: Don't forget to remove this, please

tests/integration/db_setup.go
ChrisBQu: Done

tests/integration/db_setup.go
ChrisBQu: This was a good idea, as we discussed. Thank you. Done

tests/integration/utils.go
ChrisBQu: Done

shahzadlone (changes requested):
todo: Add running of tests (I would prefer with coverage, but at least have it run even if without).

tests/integration/signature/commit_test.go
```diff
@@ -49,6 +50,14 @@ func TestSignature_WithCommitQuery_ShouldIncludeSignatureData(t *testing.T) {
 
 	test := testUtils.TestCase{
 		EnableSigning: true,
+		SupportedClientTypes: immutable.Some([]state.ClientType{
```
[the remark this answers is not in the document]
shahzadlone: Made an issue https://github.com/sourcenetwork/defradb/issues/3915

.github/workflows/test-c-bindings.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+    branches:
+      - master
+      - develop
+
+jobs:
+  check-data-format-changes:
```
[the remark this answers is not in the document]
ChrisBQu: You caught this one faster than I did.

tests/integration/signature/branchable_test.go
```diff
@@ -14,14 +14,24 @@ import (
...
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
 )
 
 func TestSignature_WithBranchableCollection_ShouldSignCollectionBlocks(t *testing.T) {
 	test := testUtils.TestCase{
 		EnableSigning: true,
+		SupportedClientTypes: immutable.Some([]state.ClientType{
+			// C bindings do not support calling functions with non-Secp256k key yet
+			testUtils.CClientType,
```
shahzadlone, line 29: Todo: Incase missed, it says not supported hut still has the client type under supported

tests/integration/client_setup.go
```diff
@@ -17,31 +17,40 @@ import (
...
 )
 
 func init() {
-	if !goClient && !httpClient && !cliClient {
+	if !goClient && !httpClient && !cliClient && !cClient {
 		// Default is to test go client type.
 		goClient = true
 	}
+	if cClient {
+		// TODO: Network test support for C client
```
shahzadlone, line 32: Todo: all todos must be linked to an issue. Dont forget to resolve this

tests/integration/signature/branchable_test.go
ChrisBQu: Taken care of.

cbindings/bridge/utils.go
shahzadlone: @ChrisBQu will this be done before merge?

.github/workflows/test-c-bindings.yml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+name: Check C Bindings Workflow
```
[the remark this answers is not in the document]
shahzadlone: @ChrisBQu doing this?

Makefile
```diff
@@ -432,3 +436,11 @@ fix:
 	@$(MAKE) tidy
 	@$(MAKE) mocks
 	@$(MAKE) docs
+	
+build-c-shared-linux:
```
shahzadlone, line 440: question: Where in the pipe line does this run? do we need to add this to `make build`? or run it in the CI?

ChrisBQu: This doesn't actually run in the pipeline. It's included as the way to build the shared library, but this library is not directly tested (at least, not yet -- we will hopefully eventually do something with it.) But it's included because the C bindings are incomplete without... well, a way to bind to C.

cbindings/logic/identity.go
```diff
@@ -0,0 +1,44 @@
...
+func NodeIdentity(n int) GoCResult {
+	ctx := context.Background()
+	identity, err := GetNode(n).DB.GetNodeIdentity(ctx)
+	if err != nil {
+		return returnGoC(1, err.Error(), "")
+	}
+	if identity.HasValue() {
+		return marshalJSONToGoCResult(identity.Value())
+	}
+	return returnGoC(0, "", "Node has no identity assigned to it.")
```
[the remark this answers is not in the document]
ChrisBQu: This is not an error. Granted, it is a constant, and could be defined somewhere -- but it is only used one time, in this one place. Take a look at cli/node_identity.go, on line 45, and you'll see the same thing there. This is actually what I modelled this function after. I think if we wish to change one, we should change both. But I also think its a very minor issue.

Makefile
shahzadlone: Perhaps make an issue for this to best testing or be part of pipeline somehow in the future.

fredcarle:
Bug bash: Created a C program with basic operations and it seems to work nicely

PR #3794, opened 2025-06-23
feat: JavaScript event system

### LINKED ISSUE #3793: JavaScript event implementation
Implement a JavaScript version of the event bus using EventTarget and EventListener.

nasdf:
## Relevant issue(s)

Resolves #3793

## Description

This PR adds a JavaScript based event system. This is a pre-requisite for the JavaScript p2p system.

Note: Some of the events contain channels and will not be compatible. This will be fixed in a future PR.

## Tasks

## How has this been tested?

Added unit tests.

- MacOS

### DIFF
--- a/event/bus_test.go
+++ b/event/bus_test.go
@@ -11,6 +11,7 @@
 package event
 
 import (
+	"errors"
 	"sync"
 	"time"
 
@@ -59,7 +60,7 @@ func (b *BusTestSuite) TestBus_IfSubscriptionIsUnsubscribedTwice_ItShouldNotPani
 	bus := b.setup(0)
 	defer bus.Close()
 
-	sub, err := bus.Subscribe(WildCardName)
+	sub, err := bus.Subscribe("test")
 	b.Assert().NoError(err)
 
 	bus.Unsubscribe(sub)
@@ -71,6 +72,9 @@ func (b *BusTestSuite) TestBus_IfSubscribedToWildCard_ItShouldNotReceiveMessageT
 	defer bus.Close()
 
 	sub, err := bus.Subscribe("test", WildCardName)
+	if errors.Is(err, ErrWildcardNotSupported) {
+		b.T().Skipf("wildcard not supported")
+	}
 	b.Assert().NoError(err)
 
 	msg := NewMessage("test", 1)
@@ -91,8 +95,8 @@ func (b *BusTestSuite) TestBus_IfMultipleSubscriptionsToTheSameEvent_EachSubscri
 	bus := b.setup(0)
 	defer bus.Close()
 
-	msg1 := NewMessage("test", 1)
-	msg2 := NewMessage("test", 2)
+	msg1 := NewMessage("test", false)
+	msg2 := NewMessage("test", true)
 
 	sub1, err := bus.Subscribe("test")
 	b.Assert().NoError(err)
@@ -145,8 +149,8 @@ func (b *BusTestSuite) TestBus_IfMultipleBufferedSubscribersWithMultipleEvents_E
 	bus := b.setup(2)
 	defer bus.Close()
 
-	msg1 := NewMessage("test", 1)
-	msg2 := NewMessage("test", 2)
+	msg1 := NewMessage("test", false)
+	msg2 := NewMessage("test", true)
 
 	sub1, err := bus.Subscribe("test")
 	b.Assert().NoError(err)
--- a/event/bus_test.go
+++ b/event/bus_test.go
@@ -12,82 +12,93 @@ package event
 
 import (
 	"sync"
-	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
 )
 
-func TestBus_IfPublishingWithoutSubscribers_ItShouldNotBlock(t *testing.T) {
-	bus := NewBus(0, 0)
+type BusTestSuite struct {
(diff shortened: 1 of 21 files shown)


event/bus_test.go
```diff
@@ -12,82 +12,93 @@ package event
...
-	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
 )
 
-func TestBus_IfPublishingWithoutSubscribers_ItShouldNotBlock(t *testing.T) {
-	bus := NewBus(0, 0)
+type BusTestSuite struct {
```
AndrewSisley, line 21: praise: Whilst if we get serious about testing this package long term I might not want us to test it like this, this is a very nice and low risk way of testing multiple implementations without the cost of setting up an action based int. test system for the package, and it keeps any would-be-effort largely out of the rod-code PR.

Thanks Keenan :)

event/channel_bus.go
```diff
@@ -94,18 +106,22 @@ func (b *Bus) Subscribe(events ...Name) (*Subscription, error) {
...
 	b.closeMutex.RLock()
 	defer b.closeMutex.RUnlock()
 
 	if b.isClosed {
 		return
 	}
-	b.commandChannel <- unsubscribeCommand(sub)
+	s, ok := sub.(*channelSub)
+	if !ok {
+		return
```
AndrewSisley, line 118: thought: Did you you consider panicking here instead? I worry that this might cause hidden bugs long term, with a silent return.

suggestion: I think a comment here either warning that this is not something we want long term (maybe linked to a new issue), or, if you do want this long term, a description of why you think it is a good solution as otherwise it looks a bit broken.

event/event.go
```diff
@@ -15,6 +15,25 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+type Bus interface {
```
AndrewSisley, line 18: todo: This is a user-public type, and should really have type level documentation, as well as the stuff on the functions.

Same for `Subscription`

event/event_target_bus.go
```diff
@@ -0,0 +1,162 @@
...
+// manage subscribers and publish messages.
+type eventTargetBus struct {
+	target       goji.EventTargetValue
+	bufferSize   uint
+	subscriberID atomic.Uint64
+	subscribers  map[uint64]*eventTargetSub
+	closeMutex   sync.Mutex
+}
+
+// NewEventTargetBus creates a new bus from the given event target value.
```
AndrewSisley, line 45: suggestion: Took me a few seconds to realise what `event target` meant here (I checked this before the documentation on the private type, which is also not going to be readily available to users).  I suggest tweaking this slightly to `given JavaScript EventTarget value.`

suggestion: The `js.Value` type does cover this slightly, but it might be worth expanding the documentation to describe when this bus should probably be used over the other bus (and I guess future buses).  It might be worth adding a similar line to `NewChannelBus` too.

nitpick: Whilst it is JS, it might be worth explicit in noting that it communicates across the runtime boundary using JSON.

event/event_target_bus.go
```diff
@@ -0,0 +1,162 @@
...
+// unmarshalMessage unmarshals a message from a JS EventValue.
+func unmarshalMessage(event goji.EventValue) Message {
+	message := Message{
+		Name: Name(event.Type()),
+	}
+	detail := js.Value(event).Get("detail")
+	if detail.IsUndefined() {
+		return message
+	}
+	switch message.Name {
```
AndrewSisley, line 127: suggestion: Might be worth adding a `//no-op` `default` case here with a comment detailing why that is okay/the-lesser-evil.  And perhaps a similar comment for the `detail.IsUndefined()` early return.  Or perhaps note it in the function signature.

Without this, the reader is left guessing, and in the case of the `switch` it wouldn't be too hard to guess that it was accidental.

event/event_target_bus.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+//go:build js
```
AndrewSisley, line 11: question: Why this restriction?  And do we need an inverse restriction on `channelBus`?

event/event_target_bus.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Message returns the message channel for the subscription.
+func (s *eventTargetSub) Message() <-chan Message {
+	return s.value
+}
+
+// eventTargetBus uses a JavaScript EventTarget to
+// manage subscribers and publish messages.
+type eventTargetBus struct {
```
AndrewSisley, line 37: question: Did you consider decoupling this from JavaScript?  Could it perhaps be useful for non-browser hosted WASM builds, and other use-cases?

AndrewSisley (approved):
LGTM, all comments from me are fairly localised - would appreciate responses before merge though :)

event/event_target_bus.go
nasdf: It's required to use the `syscall/js` package. The `channelBus` works fine in WASM so I don't see a need to restrict it.

shahzadlone:
Bug bashed: No bugs

PR #3822, opened 2025-07-07
feat: Add explicit p2p doc subscription

### LINKED ISSUE #3816: Remove automatic pubsub subscription for all created docs
The current system adds a p2p topic (a pubsub subscription) for every new document created. Since topic subscription creates a (multiple?) new goroutine for every topic, this can create a lot of overhead. We should remove that automatic subscription and instead have an opt-in system.

### LINKED ISSUE #3823: Support P2P document subscription
Add support for p2p subscription to document via docIDs. The API should be the same as for the p2p collections.

fredcarle:
## Relevant issue(s)

Resolves #3816 
Resolves #3823

## Description

This PR introduced a new feature that allows users to explicitly subscribe to DocIDs on the P2P pubsub network. With this we removed the automatic subscription to every docID which created a lot of potentially unnecessary overhead as every topic subscription creates at least one new go routine.

## Tasks

## How has this been tested?

Added new document subscription integration tests.

- MacOS

### DIFF
--- a/cli/p2p_document.go
+++ b/cli/p2p_document.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Democratized Data Foundation
+// Copyright 2025 Democratized Data Foundation
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt.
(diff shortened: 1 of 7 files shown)


nasdf (approved):
No blockers from me. I think this is a much more intuitive way to handle document subscriptions.

PR #3868, opened 2025-07-24
test(i): Move int. test state to state package

### LINKED ISSUE #3867: Move int. test state to own directory

AndrewSisley:
## Relevant issue(s)

Resolves #3867

## Description

Moves the integration test `state` type and dependencies to a new state package.

The only behaviour change is in the first commit, where `Description` was removed from a lot of failure messages.

The new `state` package lies outside of the `integration` directory - the state, actions and multipliers should all be available for use for other stuff like benchmarks.  

This begins the process of converting our main test framework to `testo`.  Later PRs will convert the actions and multipliers.

I do not agree with the location of all of the properties that are currently on `State` - many of them will need to be removed as the framework is converted.  This includes the `DatabaseType`, `KMSType` and `ClientType` types that were moved to the `state` package out of short-term necessity.  

If tolerable, I would very much like to avoid any changes requiring serious thought in this PR - I am looking to start the process by a fairly simple/dumb move of the `state` type, nothing more.

### DIFF
--- a/tests/integration/identity.go
+++ b/tests/integration/identity.go
@@ -24,21 +24,21 @@ import (
 	"github.com/sourcenetwork/defradb/crypto"
 )
 
-type identityType int
+type IdentityType int
 
 const (
-	clientIdentityType identityType = iota
-	nodeIdentityType
+	ClientIdentityType IdentityType = iota
+	NodeIdentityType
 )
 
 // Identity helps specify Identity type info and selector/index of Identity to use in a test case.
 type Identity struct {
 	// type of identity
-	kind identityType
+	Kind IdentityType
 
-	// selector can be a valid identity index or a selecting pattern like "*".
+	// Selector can be a valid identity index or a selecting pattern like "*".
 	// Note: "*" means to select all identities of the specified [kind] type.
-	selector string
+	Selector string
 }
 
 // NoIdentity returns an reference to an identity that represents no identity.
@@ -50,8 +50,8 @@ func NoIdentity() immutable.Option[Identity] {
 func AllClientIdentities() immutable.Option[Identity] {
 	return immutable.Some(
 		Identity{
-			kind:     clientIdentityType,
-			selector: "*",
+			Kind:     ClientIdentityType,
+			Selector: "*",
 		},
 	)
 }
@@ -60,8 +60,8 @@ func AllClientIdentities() immutable.Option[Identity] {
 func ClientIdentity(indexSelector int) immutable.Option[Identity] {
 	return immutable.Some(
 		Identity{
-			kind:     clientIdentityType,
-			selector: strconv.Itoa(indexSelector),
+			Kind:     ClientIdentityType,
+			Selector: strconv.Itoa(indexSelector),
 		},
 	)
 }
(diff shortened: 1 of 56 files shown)


tests/integration/identity.go
```diff
@@ -12,124 +12,55 @@ package tests
...
 // ClientIdentity returns a node identity at the given index.
-func NodeIdentity(indexSelector int) immutable.Option[Identity] {
+func NodeIdentity(indexSelector int) immutable.Option[state.Identity] {
 	return immutable.Some(
-		Identity{
-			kind:     nodeIdentityType,
-			selector: strconv.Itoa(indexSelector),
+		state.Identity{
+			Kind:     state.NodeIdentityType,
+			Selector: strconv.Itoa(indexSelector),
```
fredcarle, line 54: question: If moving `Identity` to `state`, why not these functions too?

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+	Col int
+	Doc int
+}
+
+func NewColDocIndex(col, doc int) ColDocIndex {
+	return ColDocIndex{col, doc}
+}
+
+// P2pState contains all p2p related testing state.
+type P2pState struct {
```
fredcarle, line 56: suggestion: `P2PState`.

I'll add it here only but the reference for this suggestion and the others to follow: https://go.dev/wiki/CodeReviewComments#initialisms

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+	// This tracks composite commits for documents, and collection commits for
+	// branchable collections
+	ExpectedDAGHeads map[string]cid.Cid
+}
+
+// DocHeadState contains the state of a document head.
+// It is used to track if a document at a certain head has been decrypted.
+type DocHeadState struct {
+	// The actual document head.
+	Cid cid.Cid
```
fredcarle, line 98: suggestion: `CID`

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// NodeState contains all testing state for a node.
+type NodeState struct {
+	// The node's client active in this test.
+	clients.Client
+	// Event contains all Event node subscriptions.
+	Event *EventState
+	// P2p contains P2p states for the node.
+	P2p *P2pState
```
fredcarle, line 155: suggestion: `P2P`

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+	// The network configurations for the nodes
+	NetOpts []netConfig.NodeOpt
+	// The path to any file-based databases active in this test.
+	DbPath string
```
fredcarle, line 159: suggestion: `DBPath`

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+// State contains all testing State.
+type State struct {
+	// The test context.
+	Ctx context.Context
+
+	// The Go Test test state
+	T testing.TB
+
+	// The type of KMS currently being tested.
+	Kms KMSType
```
fredcarle, line 178: suggestion: `KMS`

tests/state/state.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// The type of database currently being tested.
+	Dbt DatabaseType
```
fredcarle, line 181: suggestion: This one is a bit cryptic in this form. Maybe `DBType` would clearer for readers.

fredcarle (approved):
Approving with comments to resolve before merge.

tests/state/state.go
AndrewSisley: Leaving this one as is, as it is not an acronym, just a shortened name. Database is one word.

tests/state/state.go
AndrewSisley: I've renamed this as suggested

PR #3873, opened 2025-07-29
feat: Searchable encryption

### LINKED ISSUE #3681: Implement basic searchable encryption
Implement the foundational Searchable Encryption capability for basic equality (_eq) queries on designated fields. This involves client-side tag generation using a deterministic, keyed approach (similar to HMAC beacons), storage of the index in the remote peer's KV store via replication updates, and a dedicated query mechanism.

Deliverables:
- Key Management: Introduce a new persistent, symmetric searchable-encryption-key per user identity within the defradb keyring system.
- Schema Definition: Implement the @encryptedIndex(type: "equality") directive (or simply @encryptedIndex defaulting to equality) for marking fields in the SDL.
- Client-Side Index Metadata: Implement local storage and management of metadata for defined encrypted indexes (tracking IndexID, target field(s), type, config).
- Search Tag Computation: Implement client-side tag generation using HMAC-SHA256 (truncated) with the SE key and appropriate domain separation (using CollectionID + IndexID).
- Client Indexing Logic: Integrate logic into commit processing to identify indexed fields (initially focusing on LWW CRDTs), compute old/new tags, and queue index update operations.
- Replication Channel: Define and implement IndexAddOperation and IndexDeleteOperation P2P messages; modify the active replicator to send these after block commits.
- Remote Node KV Handling: Implement logic on the remote node's P2P service to receive index update messages and apply corresponding Put/Delete operations to its local KV store using the defined key structure (/se/<UserID>/<CollectionID>/<IndexID>/<SearchTag>/<DocID>).
- Query API: Auto-generate dedicated GraphQL mutations per collection (e.g., encrypted_User(filter: {...})) for SE queries. The filter input type should only allow _eq on indexed fields. Configure the target remote peer implicitly (e.g., via startup config).
- Query Execution: Implement the client-side logic to compute the query tag/prefix and make a P2P RPC call (e.g., requestKVPrefixScan) to the remote node. Implement the remote node RPC handler to scan the KV index and return matching DocIDs.
- Documentation: Basic documentation on how to enable and use equality-based SE.

islamaliev:
## Relevant issue(s)

Resolves #3681


internal/se/replication_coordinator.go
```diff
@@ -0,0 +1,671 @@
...
+// DB interface required by ReplicationCoordinator
+type DB interface {
+	Rootstore() corekv.TxnStore
+	Events() event.Bus
+	MaxTxnRetries() int
+	GetCollections(context.Context, client.CollectionFetchOptions) ([]client.Collection, error)
+}
+
+// SERetryInfo stores retry information for failed SE replications
+type SERetryInfo struct {
```
islamaliev: note: retry logic is almost identical to the regular one, but there is [a ticket](https://github.com/sourcenetwork/defradb/issues/4016) to tackle this.

client/encrypted_index.go
```diff
@@ -0,0 +1,36 @@
...
+	// FieldName is the name of the field that is being indexed.
+	FieldName string
+
+	// Type is the type of searchable encryption.
+	// Currently only "equality" is supported.
+	Type EncryptedIndexType
+}
+
+// EncryptedIndexCreateRequest describes an encrypted index creation request.
+type EncryptedIndexCreateRequest struct {
```
AndrewSisley: question: Why does both `EncryptedIndexCreateRequest` and `EncryptedIndexDescription` exist? They appear to be identical.

internal/db/collection_define.go
```diff
@@ -102,7 +110,27 @@ func (db *DB) createCollections(
...
 }
 
-// PatchCollection takes the given JSON patch string and applies it to the set of CollectionVersions
+func processCreateEncryptedIndexRequest(
+	definition client.CollectionVersion,
+	createEncryptedIndex client.EncryptedIndexCreateRequest,
+) (client.EncryptedIndexDescription, error) {
+	indexType := createEncryptedIndex.Type
+	if indexType == "" {
+		indexType = client.EncryptedIndexTypeEquality
```
AndrewSisley: todo: This should not take place here.  This is an artefact of GQL schema - everything similar to this is done in the GQL package when mapping to a `client.CollectionVersion`.

The `client.CollectionVersion` received by `db.createCollections` should be valid - any inpput language, user-layer, omissions should be handled before reaching here.

This ensures that the core logic around managing collections is decoupled from any user interface, and makes that core logic much simpler than it would be it it also had to take care of all the possible user-layer quirks.

Please move this logic to somewhere within the `fromAst` func in `internal/request/graphql/schema/collection.go`. 

The `processCreateIndexRequest` block above where this is called is legacy, it will be removed in time too, please do not propagate that pattern.

client/collection_description.go
```diff
@@ -69,6 +69,9 @@ type CollectionVersion struct {
 	// Indexes contains the secondary indexes that this Collection has.
 	Indexes []IndexDescription
 
+	// EncryptedIndexes contains the encrypted indexes that this Collection has.
+	EncryptedIndexes []EncryptedIndexDescription
```
AndrewSisley, line 77: todo: This, as it currently is in this PR, is exposed through `PatchCollection`.

Please add tests to cover this, and ensure that it is either blocked off using a validation rule in the short term, or handled (long term).

internal/planner/mapper/mapper.go
```diff
@@ -930,6 +938,12 @@ func getTopLevelInfo(
 		return mapping, client.CollectionVersion{}, nil
 	}
 
+	if rootSelectType == EncryptedSearchSelection {
+		mapping.Add(0, request.DocIDsFieldName)
```
AndrewSisley: todo: Please you the ``DocIDFieldIndex` const instead of a magic `0` :)

internal/planner/se_scan.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+		}
+
+		if encIdx == nil {
+			continue
+		}
+
+		condMap, ok := condition.(map[string]any)
+		if !ok {
+			return NewErrInvalidEncryptedFieldCondition(fieldName)
```
AndrewSisley: question: How is this possible?

todo: If it is not, please remove it.  If it is, please add an integration test that covers it.

internal/planner/se_scan.go
```diff
@@ -0,0 +1,159 @@
...
+
+		n.fieldSearchTags[fieldName] = artifact.SearchTag
+	}
+
+	return nil
+}
+
+func (n *seScanNode) queryRemoteNodes() ([]string, error) {
+	if len(n.fieldSearchTags) == 0 {
+		return nil, nil
```
AndrewSisley: question: How is this possible?

todo: If it is not, please remove it.  If it is, please add an integration test that covers it. And please document why returning `nil`, `nil` is desirable in such a case.

internal/planner/se_scan.go
```diff
@@ -0,0 +1,159 @@
...
+	queries := make([]se.FieldQuery, 0, len(n.fieldSearchTags))
+	for fieldName, searchTag := range n.fieldSearchTags {
+		queries = append(queries, se.FieldQuery{
+			FieldName: fieldName,
+			IndexID:   fieldName,
+			SearchTag: searchTag,
+		})
+	}
+
+	msg, responseChan := se.NewQuerySEArtifactsMessage(n.collectionID, queries)
```
AndrewSisley: todo: Please document how this works - by submitting each search tag like this it looks like it only ever works as `or` - there *appears* to be no `and` support and future readers of this function may be concerned/confused by this.

I assume it is handled elsewhere (if not that looks like a bug? I didnt spot any user facing documentation on this either, but it is a big PR and I may have missed it).

AndrewSisley (commented):
Gave a few key bits a quick look over, it looks good, but I have some small todos.  Not marking as `Request changes` as I'm going away soon and don't want you to have to dismiss my review before merging :)

cli/encrypted_index_create.go
```diff
@@ -0,0 +1,58 @@
...
+func MakeEncryptedIndexCreateCommand() *cobra.Command {
+	var collectionArg string
+	var fieldArg string
+	var typeArg string
+	var cmd = &cobra.Command{
+		Use:   "create -c --collection <collection> --field <field> [--type <type>]",
+		Short: "Creates an encrypted index on a collection's field",
+		Long: `Creates an encrypted index on a collection's field.
+		
+The --type flag is optional. If not provided, the default value will be "equality".
```
fredcarle, line 28: todo: Please document what are the supported types. It might make sense to have a list of xor type flags that cover the supported types. That would be one way of documenting this.

client/encrypted_index.go
islamaliev: because they are different concepts: `EncryptedIndexDescription` is the stored data about the index, while `EncryptedIndexCreateRequest` is the creation instructions which might be different. For example with secondary indexes creation instruction didn't have `ID` field because it automatically generated.
Current these 2 happen to have the same structure, but with the followup change they will most likely diverge.

internal/db/collection_define.go
islamaliev: honestly, I don't see a problem with this code being here. I don't see how it's related specifically to GQL schema. `EncryptedIndexType` is part of the `EncryptedIndexDescription`.
If you mean specifically default value assignment, I wouldn't put it to the schema parsing part, as parser should be just do sheer parsing, i.e. conversion from one representation into another.
If we move this part of code to client-related input (gql, cli, http...) all of them will have to know about default value, which breaks single-source-of-truth principle.

internal/planner/mapper/mapper.go
islamaliev: done

internal/planner/se_scan.go
islamaliev: I can't think of any scenario. Removed

internal/planner/se_scan.go
islamaliev: just additional precaution. Removed

internal/db/collection_define.go
AndrewSisley: It is going to be wildly inefficient if we keep pulling in opposite directions here.

The index stuff is coded in a very different way to everything else on `collection_version`, and this complicates the code for everyone, including users, as this has public elements.  It has had a very tangible cost already in terms of my time when trying to make changes to the schema/collection system, index-exceptionalism prevents a number planned features in its current state, and if you keep adding stuff that does not match the system design it is only going to get worse and more frustrating for everyone.

Please lets have one way of doing things here.

I do not have the time to have a proper conversation around this right now, maybe we can have one when I get back, or you and the rest of the team can have one without me.

cli/encrypted_index_list.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func MakeEncryptedIndexListCommand() *cobra.Command {
+	var collectionArg string
+	var cmd = &cobra.Command{
+		Use:   "list [-c --collection <collection>]",
+		Short: "Shows the list encrypted indexes in the database or for a specific collection",
```
shahzadlone: suggestion:

```suggestion
		Short: "Lists the encrypted indexes in the database or for a specific collection",
```

internal/se/README.md
shahzadlone, line 1: nitpick: Not sure if I like the folder name to be `se` haha

tests/integration/utils.go
```diff
@@ -445,6 +446,15 @@ func performAction(
 	case GetIndexes:
 		getIndexes(s, action)
 
+	case CreateEncryptedIndex:
+		createEncryptedIndex(s, action)
+
+	case GetEncryptedIndexes:
+		getEncryptedIndexes(s, action)
```
shahzadlone: suggestion: Similar to another suggestion, stick with `list` instead of `get` here?

internal/db/collection_index.go
```diff
@@ -395,6 +395,114 @@ func (c *collection) GetIndexes(context.Context) ([]client.IndexDescription, err
...
+		return client.EncryptedIndexDescription{}, err
+	}
+
+	return c.def.EncryptedIndexes[len(c.def.EncryptedIndexes)-1], nil
+}
+
+// GetEncryptedIndexes returns all the encrypted indexes that exist on the collection.
+func (c *collection) GetEncryptedIndexes(ctx context.Context) ([]client.EncryptedIndexDescription, error) {
+	return c.Version().EncryptedIndexes, nil
+}
```
shahzadlone: suggestion: Probably want to stick with the `list` naming like you did in client cli commands for example, due to : https://github.com/sourcenetwork/defradb/issues/3881

```suggestion
// ListEncryptedIndexes returns all the encrypted indexes that exist on the collection.
func (c *collection) ListEncryptedIndexes(ctx context.Context) ([]client.EncryptedIndexDescription, error) {
	return c.Version().EncryptedIndexes, nil
}
```

internal/planner/se_scan.go
islamaliev: it is documented here internal/se/README.md. I will also add a few notes here

islamaliev: I added more documentation to gql query request/graphql/schema/generate.go. Saying that results are intersection (aka AND).

cli/encrypted_index_list.go
islamaliev: done

internal/db/collection_index.go
islamaliev: done

tests/integration/utils.go
islamaliev: done

internal/se/README.md
islamaliev: We can discuss it on the standup if you want. For now I will keep it as is

client/collection_description.go
islamaliev: blocked during patch validation and added tests

internal/se/README.md
shahzadlone: Is a non-blocker nitpick nothing important if you have a hard preference for `se`

cbindings/encrypted_index.go
```diff
@@ -0,0 +1,102 @@
...
+			return returnC(returnGoC(1, err.Error(), ""))
+		}
+		indices, err := col.ListEncryptedIndexes(ctx)
+		if err != nil {
+			return returnC(returnGoC(1, err.Error(), ""))
+		}
+		return returnC(marshalJSONToGoCResult(indices))
+	// Get all of the encrypted indices, because no collection was specified
+	default:
+		indices, err := store.GetAllEncryptedIndexes(ctx)
```
shahzadlone: suggestion: thanks for doing the previous suggestion, since you did that likely should update all other instances to list aswell?

islamaliev: done

internal/db/p2p/protocol/comm_channel.go
```diff
@@ -0,0 +1,144 @@
...
+	protocolBase           = "/defradb/"
+	protocolRequestSuffix  = "_req/" + protocolVersion
+	protocolResponseSuffix = "_resp/" + protocolVersion
+)
+
+// CommProcessor defines the interface for processing requests and replies.
+// Uses 4 type parameters to solve the embedded interface pointer receiver problem:
+// - Req/Reply: Value types for stack allocation and clean processor signatures
+// - ReqP/ReplyP: Pointer types that implement message.Message interface
+type CommProcessor[Req any, Reply any, ReqP interface {
```
fredcarle: todo: Please create an interface that you can reuse where needed so that it makes it less difficult to read and understand what's going on.
```go
type messagePointer[T any] interface {
	*T
	message.Message
}

...
type CommProcessor[Req any, Reply any ReqP messagePointer[Req], ReplyP messagePointer[Reply]](...
```

internal/db/p2p/p2p.go
```diff
@@ -90,6 +94,22 @@ type P2P struct {
...
+	p2p *P2P
+}
+
+func (proc *pushLogCommProcessor) ProcessRequest(
+	ctx context.Context,
+	req protocol.PushLogRequest,
+) (protocol.PushLogReply, error) {
+	f := func(ctx context.Context, req *protocol.PushLogRequest) error {
+		return proc.p2p.processPushlogRequest(ctx, req, true)
+	}
```
fredcarle: question: Why the closure instead of just
```go
return protocol.PushLogReply{}, proc.p2p.processPushlogRequest(ctx, req, true)
```

internal/db/p2p/replicator.go
```diff
@@ -302,13 +306,23 @@ func (p *P2P) pushLogToReplicators(lg event.Update) {
...
+				if _, err := p.replicatorProtocol.SendRequest(ctx, pushLogReq, peerID); err != nil {
 					log.ErrorE(
 						"Failed pushing log",
 						err,
 						corelog.String("DocID", lg.DocID),
 						corelog.Any("CID", lg.Cid),
 						corelog.Any("PeerID", peerID))
+					if !lg.IsRetry {
+						_ = p.handleReplicatorFailure(ctx, peerID, lg.DocID)
+					}
```
fredcarle, line 336: praise: Much better place to handle this.

internal/db/collection_define.go
fredcarle: I see what you mean Andy but I think this argument only holds water if this function was only called from `createCollection` when in fact it is also called from the encrypted index API (CreateEncryptedIndex)

internal/db/db.go
```diff
@@ -92,6 +93,12 @@ type DB struct {
 	// If true, block signing is disabled. By default, block signing is enabled.
 	signingDisabled bool
 
+	// The key used for searchable encryption.
```
fredcarle: todo: Please expand this comment so that someone who doesn't know what a key for searchable encryption is can still understand what this byte array contains. Like is it a KV store key, a cryptographic key or something else.

internal/db/errors.go
```diff
@@ -20,173 +20,178 @@ import (
...
+	errFieldNotFound                             string = "field not found"
+	errGetDocForEmbedding                        string = "failed to get previous document for embedding generation"
+	errMissingSignature                          string = "block is missing required signature"
+	errNoIdentityInContext                       string = "no identity found in context"
+	errMissingPermission                         string = "missing permission"
+	errCollectionNameMutated                     string = "collection name cannot be mutated"
+	errUnsupportedTxnType                        string = "unsupported transaction type"
+	errEncryptedIndexUnknownField                string = "encryptedIndex on non-existent field"
+	errEncryptedIndexAlreadyExists               string = "encryptedIndex already exists on this field"
+	errEncryptedIndexDoesNotExist                string = "encryptedIndex does not exist on this field"
```
fredcarle: suggestion: In the error message it would be better to user `encrypted index` instead of `encryptedIndex`.

internal/db/txn.go
```diff
@@ -187,6 +187,13 @@ func (txn *Txn) VerifySignature(ctx context.Context, blockCid string, pubKey cry
 	return txn.db.VerifySignature(ctx, blockCid, pubKey)
 }
 
+func (txn *Txn) GetSearchableEncryptionKey() []byte {
```
fredcarle: question: Why is this defined on the transaction?

internal/request/graphql/parser/query.go
```diff
@@ -90,11 +92,15 @@ func parseSelect(
 	parent *gql.Object,
 	field *ast.Field,
 ) (*request.Select, error) {
+	isEncrypted := strings.HasPrefix(field.Name.Value, request.EncryptedCollectionPrefix)
+	fieldName := field.Name.Value
```
fredcarle: suggestion: There is no reason to extract the value only to use it once bellow. I think you wanted to define it above the `HasPrefix` call to use in there as well?

internal/request/graphql/schema/errors.go
```diff
@@ -16,29 +16,32 @@ import (
...
+	errPolicyInvalidResourceProp      string = "policy directive with invalid resource property"
+	errDefaultValueType               string = "default value type must match field type"
+	errDefaultValueNotAllowed         string = "default value is not allowed for this field type"
+	errDefaultValueInvalid            string = "default value is invalid"
+	errDefaultValueOneArg             string = "default value must specify one argument"
+	errFieldTypeNotSpecified          string = "field type not specified"
+	errInvalidTypeForConstraint       string = "size constraint can only be applied to array fields"
+	errEncryptedIndexUnknownArgument  string = "encryptedIndex with unknown argument"
+	errEncryptedIndexInvalidArgument  string = "encryptedIndex with invalid argument"
+	errEncryptedIndexTypeNotSupported string = "encryptedIndex type not supported"
```
fredcarle: suggestion: Same as the other errors. `encrypted index`

internal/se/core/tag.go
```diff
@@ -0,0 +1,40 @@
...
+	// - fieldName ensures tags are unique per field
+	// This prevents cross-field and cross-collection tag collisions
+	domainSeparator := fmt.Sprintf("eq:%s:%s", collectionID, fieldName)
+
+	h := hmac.New(sha256.New, key)
+	h.Write([]byte(domainSeparator))
+	h.Write(value)
+	tag := h.Sum(nil)
+
+	// Truncate to 16 bytes for efficiency (128-bit security)
```
fredcarle: question: Is there some documentation on the safety/correctness of truncating the tag's hash?

internal/se/se.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/encoding"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	secore "github.com/sourcenetwork/defradb/internal/se/core"
+)
+
+// StoreArtifacts stores SE artifacts directly in the datastore.
+// This is called by the server when receiving artifacts from peers.
+func StoreArtifacts(ctx context.Context, ds corekv.ReaderWriter, artifacts []secore.Artifact) error {
```
fredcarle: question: Most of these functions except for the last one (`CreateEncryptedIndex`) are only ever used within this package. Any reasons for making them public?

internal/se/dto.go
fredcarle, line 1: question: What does `dto` stand for?

internal/se/coordinator.go
```diff
@@ -0,0 +1,369 @@
...
+
+var log = corelog.NewLogger("defra.se.replication")
+
+type P2P interface {
+	GetReplicatorsIDs(collectionID string) []string
+	Host() client.Host
+}
+
+// Coordinator manages SE artifact replication and storage
+type Coordinator struct {
```
fredcarle, line 48: thougth: I think this is in the wrong location. Having it here makes it so it has to duplicate some of the work the `db/p2p` is doing. I think the whole package should be split into 2 where one part lives within `db` and the other within `db/p2p`. I also don't agree with the use of the event bus the handle  `Coordinator` actions.

fredcarle (changes requested):
Here in an initial set of comments. More to come depending on the outcome of some of these.

cli/encrypted_index_create.go
islamaliev: Mentioned that only "equality" is supported.

internal/db/db.go
islamaliev: done

internal/db/errors.go
islamaliev: done

internal/db/p2p/p2p.go
islamaliev: good catch. Looks like some refactoring leftovers

internal/db/p2p/protocol/comm_channel.go
islamaliev: good idea. Done

internal/db/txn.go
islamaliev: forgot to delete.

internal/request/graphql/parser/query.go
islamaliev: done

internal/request/graphql/schema/errors.go
islamaliev: done

internal/se/dto.go
islamaliev: Data Transfer Object. I thought it was well-known abbreviation. Shall rename it something else?

internal/se/se.go
islamaliev: it used to be called from another package. I will make it private

internal/se/core/tag.go
islamaliev: added more documentation

internal/db/collection_define.go
AndrewSisley: ~~@fredcarle This function is called a few lines up from `createCollections`.~~

I misread :)

-------

Why not have both work like the existing `createCollections` stuff?  And then the code will also work by default for creating indexes via `PatchCollection` too.

The code that works on the standard, non request, `collection_version` stuff needs to be written either way.  Why not write it the first time instead of introducing view-model style request objects into `collection_version`.

`collection_version` very deliberately avoids using request stuff (it is not a request object).  I feel very strongly that some sub-systems should not leak request objects into it (for whatever reason, here the reasons feel particularly weak).

Request objects are a syntax layer, analogous to JSON patch, and GQL schema. Request objects should not be leaked into `collection_version` and its children.

If we want to change this, then we should be very very deliberate about this and design it properly for all collection sub-systems, instead of mangling the existing system with a mismatch of competing systems - each of which significantly increases complexity and hinders development.

I have no problem with request objects being used in `CreateEncryptedIndex` or similar functions, that is how those APIs chose to express themselves to users. `PatchCollection` does not chose this though, and it makes no sense to do so. `collection_version` and `db.createCollections` also very deliberately does not chose to do this, and, exists at a different layer of sorts.  Both are designed to work with reality (`collection_version`), instead of with API/User level request objects.

Then there is also the issue that introducing request objects into the compile-time reality used across Defra introduces ambiguity into every other system, where the compile-time type says encrpyted-indexes *might* actually be a request object, instead of the 'real' data object.

cli/encrypted_index_create.go
fredcarle: Where? Did you forget to push that change? I only see that the default value is "equality".

event/event.go
```diff
@@ -14,6 +14,9 @@ import (
 	"github.com/ipfs/go-cid"
 	"github.com/libp2p/go-libp2p/core/peer"
 
+	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/immutable"
```
AndrewSisley: nitpick: These imports arent correctly formatted, the linter is just confused.

`immutable` is external to this repo and belongs in the same block as `github.com/libp2p/go-libp2p/core/peer`.  `acpIdentity` should be in the same block as `client`

internal/db/collection_index.go
```diff
@@ -414,6 +524,24 @@ func checkExistingFieldsAndAdjustRelFieldNames(
 	return nil
 }
 
+// validateNewEncryptedIndex validates, if encrypted index can be created on the field.
+// It checks if the field exists in the collection schema and if an encrypted index already exists on the field.
+func validateNewEncryptedIndex(
```
AndrewSisley, line 529: suggestion: If this is tweaked and added to the main collection validation rules, it will automatically validate anything added by collection patch - could be handy for when that support is enabled.

It is also a more expected location for it IMO, but I'm probably quite biased there.

internal/request/graphql/schema/generate.go
```diff
@@ -1118,6 +1143,46 @@ func (g *Generator) GenerateQueryInputForGQLType(
...
+) (*gql.Field, error) {
+	var collection *client.CollectionVersion
+	for _, col := range collections {
+		if col.Name == obj.Name() {
+			collection = &col
+			break
+		}
+	}
+
+	if collection == nil || len(collection.EncryptedIndexes) == 0 {
```
AndrewSisley: question: When could `collection` ever be nil?

todo: If it can be nil, please document how.

todo: If it cant be nil, please remove this check as it confuses readers into thinking that something very odd is happening :)

internal/request/graphql/schema/generate.go
```diff
@@ -1327,6 +1392,104 @@ func (g *Generator) genLeafFilterArgInput(obj gql.Type) *gql.InputObject {
...
+			for _, encIdx := range encryptedIndexes {
+				var objField *gql.FieldDefinition
+				for f, field := range obj.Fields() {
+					if f == encIdx.FieldName {
+						objField = field
+						break
+					}
+				}
+				if objField == nil {
+					// This should not happen, as we should have already validated the schema
```
AndrewSisley: todo: I would much prefer a panic to pretending everything is fine here, it should make fixing whatever bug led to this being nil slightly easier.  As well reducing the risk to the users system where they may be saving a corrupt state.

Please remove the if-continue

I have similar thoughts on the handling of the `exists` returned by `encryptedFilterType, exists := g.manager.schema.TypeMap()`

internal/se/coordinator.go
```diff
@@ -0,0 +1,377 @@
...
+				err,
+				corelog.String("CollectionID", collectionID),
+				corelog.Any("PeerID", pid))
+			continue
+		}
+
+		break
+	}
+
+	if err != nil {
```
AndrewSisley, line 207: todo: This looks incorrect for two reasons:

1. You are handling the error by logging and so shouldn't be returning it
2. You are overwriting this `err` var on the next iteration of the for loop, meaning `err` will only ever be returned if the error is raised by the last `pid` in `peerIDs`.

EDIT: I am mistaken, I misread.

todo: Please document why you are doing this, it is a little unexpected, and an explanation as to why it is okay/desired would be handy

internal/se/coordinator.go
```diff
@@ -0,0 +1,377 @@
...
+		return fmt.Errorf("failed to deserialize block: %w", err)
+	}
+
+	if !block.Delta.IsComposite() {
+		return nil
+	}
+
+	updatedFields := []string{}
+	for _, link := range block.Links {
+		if link.Name != "" && link.Name != "_head" {
```
AndrewSisley: todo: This looks like the copy-paste of legacy code, there should be no head links in `block.Links`.   Please either remove the check, or document why it is needed.

internal/se/coordinator.go
```diff
@@ -0,0 +1,377 @@
...
+func (rc *Coordinator) generateSEArtifacts(
+	ctx context.Context,
+	docID, collectionID string,
+	fieldNames []string,
+) ([]secore.Artifact, error) {
+	cols, err := rc.p2p.DB().GetCollections(ctx, client.CollectionFetchOptions{
+		CollectionID: immutable.Some(collectionID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get collection: %w", err)
```
AndrewSisley: todo: Please use the errors package instead of `fmt`, there are a few in this file, including further up.

internal/se/se.go
```diff
@@ -0,0 +1,212 @@
...
+// FieldQuery represents a query for a specific encrypted field
+type FieldQuery struct {
+	FieldName string
+	IndexID   string
+	SearchTag []byte
+}
+
+// generateDocArtifacts generates SE artifacts for specified fields of a document.
+// If fieldNames is empty or nil, artifacts are generated for all encrypted fields.
+func generateDocArtifacts(
```
AndrewSisley, line 131: suggestion: This function is called in one place, quite a few files up - it would have been much easier to find if it was declared near where it is called, or if it did not exist at all.

I have a preference for it's removal.

internal/se/se.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, nil
+	}
+
+	collectionID := col.VersionID()
+	docID := doc.ID().String()
+
+	var artifacts []secore.Artifact
+	for _, encIdx := range encryptedIndexes {
+		// Skip if fieldNames is specified and this field is not in the list
+		if len(fieldNames) > 0 && !slices.Contains(fieldNames, encIdx.FieldName) {
```
AndrewSisley: todo: `len(fieldNames) > 0` can be performed outside of the for loop and doing so will make reading this easier.

suggestion: I'm fairly confident that `fieldNames` can *never* have a len of 0, as that would require a document composite block without any field blocks.  It takes a long walk back up through the call stack to see this, and that to me is a code smell.  I suggest removing the check completely, as it is dead code, and I suggest shrinking the call stack so that stuff like this is easier to see.

internal/se/se.go
```diff
@@ -0,0 +1,212 @@
...
+			continue
+		}
+
+		fieldValue, err := doc.GetValue(encIdx.FieldName)
+		if err != nil {
+			return nil, NewErrFailedToGetFieldValue(encIdx.FieldName, err)
```
AndrewSisley: suggestion: It looks like the `err` returned from `doc.GetValue` should probably be impossible, and even then the errors look like they will be pretty for the user.  I suggest removing the new error type and just return `err`.

internal/se/core/artifact.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Artifact represents a searchable encryption operation to be replicated
+type Artifact struct {
+	Type         ArtifactType
+	CollectionID string
+	FieldName    string
+	DocID        string
+	Operation    OperationType
+	IndexID      string
+	SearchTag    []byte
```
AndrewSisley: suggestion:  This feels like a good location to document what the SearchTag is.

internal/se/core/tag.go
```diff
@@ -0,0 +1,50 @@
...
+// For fields with low cardinality (e.g., boolean, status codes), consider the
+// privacy implications.
+//
+// The tag is computed as: HMAC-SHA256(key, "eq:collectionID:fieldName" || value)[:16]
+func GenerateEqualityTag(
+	key []byte,
+	collectionID string,
+	fieldName string,
+	value []byte,
+) ([]byte, error) {
```
AndrewSisley: todo: There is never an error returned from this function, please remove the `error` return param - you are adding complexity to everywhere that calls this function.

internal/se/dto.go
AndrewSisley: I think it is a well known abbreviation, but I also consider the name totally useless, as every state bearing thing is transferring data from one part of the system to another.

If we are happy with a useless name I would prefer one that we already use, such as `utils` or `core` :grin:

internal/se/se.go
```diff
@@ -0,0 +1,212 @@
...
+			return nil, err
+		}
+		artifacts = append(artifacts, artifact)
+	}
+
+	return artifacts, nil
+}
+
+// GenerateFieldArtifact generates a single SE artifact for a specific field value.
+func GenerateFieldArtifact(
```
AndrewSisley: question: This looks like it is only called from within this package, but I might still have some collapsed files - why is it public?

tests/bench/query/planner/utils.go
```diff
@@ -22,10 +22,23 @@ import (
 	"github.com/sourcenetwork/defradb/internal/core"
 	"github.com/sourcenetwork/defradb/internal/planner"
 	"github.com/sourcenetwork/defradb/internal/request/graphql"
+	"github.com/sourcenetwork/defradb/node"
 	benchutils "github.com/sourcenetwork/defradb/tests/bench"
 	"github.com/sourcenetwork/defradb/tests/bench/fixtures"
 )
 
+type dbWrapper struct {
```
AndrewSisley: question: This looks concerning, why are you feeling the need to do this?

internal/db/p2p/p2p.go
```diff
@@ -126,9 +159,35 @@ func New(ctx context.Context, db DB, host client.Host) (*P2P, error) {
...
+
 	return &p, nil
 }
 
+func (p *P2P) Host() client.Host {
+	return p.host
+}
+
+func (p *P2P) DB() se.DB {
```
AndrewSisley: todo: This feels very wrong.  I don't think `P2P` should be returning a `DB` type defined by the `se` package.

internal/db/db.go
```diff
@@ -533,3 +548,30 @@ func printStore(ctx context.Context, store corekv.ReaderWriter) error {
 
 	return iter.Close()
 }
+
+func (db *DB) GetSearchableEncryptionKey() []byte {
+	return db.searchableEncryptionKey
+}
+
+// QueryDocIDsByValues queries SE artifacts from replicators based on field values.
+func (db *DB) QueryDocIDsByValues(
```
AndrewSisley: todo: I think this function name needs to reflect that it is use SE.  At the moment the name is very generic.

internal/se/coordinator.go
```diff
@@ -0,0 +1,377 @@
...
+		PublicKey:    publicKey,
+		KeyType:      keyType,
+	}
+
+	b, err := cbor.Marshal(retryInfo)
+	if err != nil {
+		return err
+	}
+
+	ps := datastore.PeerstoreFrom(rc.p2p.DB().Rootstore())
```
AndrewSisley: todo: This feels wrong, `se` should not be having to redefine the peerstore.

internal/se/coordinator_protocol.go
```diff
@@ -0,0 +1,135 @@
...
+	for i, netArtifact := range req.Artifacts {
+		artifacts[i] = secore.Artifact{
+			DocID:        netArtifact.DocID,
+			IndexID:      netArtifact.IndexID,
+			SearchTag:    netArtifact.SearchTag,
+			CollectionID: req.CollectionID,
+		}
+	}
+
+	if err := storeArtifacts(ctx, datastore.DatastoreFrom(rc.p2p.DB().Rootstore()), artifacts); err != nil {
```
AndrewSisley: todo: `se` should not be redefining the datastore.  Please rework this, and other similar stuff to make use of the controlled types that expose the sub-stores - such as multi store and txn.

event/event.go
islamaliev: done

internal/se/coordinator.go
islamaliev: done

internal/db/collection_index.go
islamaliev: done

internal/db/db.go
islamaliev: done

internal/db/p2p/p2p.go
islamaliev: done

internal/request/graphql/schema/generate.go
islamaliev: removed

internal/request/graphql/schema/generate.go
islamaliev: done

internal/se/coordinator.go
islamaliev: done

internal/se/coordinator.go
islamaliev: added comments

internal/se/coordinator.go
islamaliev: removed the check

internal/se/coordinator.go
islamaliev: done

internal/se/coordinator_protocol.go
islamaliev: done

internal/se/core/artifact.go
islamaliev: done

internal/se/core/tag.go
islamaliev: done

internal/se/se.go
islamaliev: this function is called by coordinator, but it's not a part of it as it's more generic. That's why it's here. I'd prefer to keep it here

internal/se/se.go
islamaliev: done

internal/se/se.go
islamaliev: It used to be called from planner, but not anymore. Made it private

internal/se/se.go
islamaliev: it used to be called by a public method. That's why there this precaution. But I made the method private and removed this check.
Good catch!

tests/bench/query/planner/utils.go
islamaliev: it's because `planner.DB` expects `QueryDocIDsWithSETags` method, but it's not defined in `node.DB` the instance of which bench uses. So I didn't feel like adding the method to the `node.DB` just for benches

internal/se/dto.go
fredcarle: > I think it is a well known abbreviation

My education must be lacking then lol

internal/db/collection_index.go
fredcarle: You marked it as done but I only see it called above. Did you forget to push a commit?

internal/db/errors.go
```diff
@@ -20,175 +20,183 @@ import (
...
+	ErrNACIsEnabledButIsMissingPolicyInfo               = errors.New(errNACIsEnabledButIsMissingPolicyInfo)
+	ErrNACNodeObjectToGateIsNotRegistered               = errors.New(errNACNodeObjectToGateIsNotRegistered)
+	ErrNACIsEnabledButInstanceIsNotAvailable            = errors.New(errNACIsEnabledButInstanceIsNotAvailable)
+	ErrNACRelationshipOperationRequiresIdentity         = errors.New("node acp relationship operation requires identity")
+	ErrRelationNameEmpty                                = errors.New(errRelationNameEmpty)
+	ErrInvalidCID                                       = errors.New(errInvalidCID)
+	ErrUnknownCID                                       = errors.New(errUnknownCID)
+	ErrNoP2P                                            = errors.New("no p2p system configured")
+	ErrBadDocsResultType                                = errors.New("bad docs result type")
+	errMigrationBetweenNonAdjacentVersions       string = "cannot migrate between non-adjacent collection versions"
```
fredcarle: todo: Please move this with the other string definitions above.

internal/planner/planner.go
```diff
@@ -84,21 +85,34 @@ type PlanContext struct {
...
 
+type DB interface {
+	client.TxnStore
+	Events() event.Bus
+	// QueryDocIDsWithSETags queries SE artifacts from replicators based on field values.
+	QueryDocIDsWithSETags(ctx context.Context, collectionID string, fieldValues []SEFieldValueQuery) ([]string, error)
+}
+
+// SEFieldValueQuery represents a field value to query for SE artifacts.
+type SEFieldValueQuery struct {
```
fredcarle: thought: This really seems like it should live in the client package.

internal/planner/planner.go
```diff
@@ -84,21 +85,34 @@ type PlanContext struct {
 	context.Context
 }
 
+type DB interface {
+	client.TxnStore
+	Events() event.Bus
```
fredcarle: todo: `Events()` isn't used so please remove it from the interface.

internal/request/graphql/schema/collection.go
```diff
@@ -447,6 +460,66 @@ func defaultFromAST(
...
+			return client.EncryptedIndexDescription{}, NewErrEncryptedIndexWithUnknownArg(arg.Name.Value)
+		}
+	}
+
+	return encryptedIndex, nil
+}
+
+// validateEncryptedIndexes validates that encrypted indexes can be created on the given fields.
+// It checks that all fields exist in the collection schema and that there is at most one index per field.
+func validateEncryptedIndexes(
```
fredcarle: thought: I'm not sure why it should be the responsibility of the graphql package. It seems like it should live in the collection validation part of the code base.

internal/se/coordinator.go
```diff
@@ -0,0 +1,389 @@
...
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/event"
+	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
+	"github.com/sourcenetwork/defradb/internal/datastore"
+	"github.com/sourcenetwork/defradb/internal/db/p2p/protocol"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	secore "github.com/sourcenetwork/defradb/internal/se/core"
+)
+
+var log = corelog.NewLogger("defra.se.replication")
```
fredcarle: suggestion: just `se` or something like that should be enough. That pattern of logger naming isn't used anywhere else in the codebase.

internal/se/coordinator.go
```diff
@@ -0,0 +1,389 @@
...
+type P2P interface {
+	GetReplicatorsIDs(collectionID string) []string
+	Host() client.Host
+}
+
+// Coordinator manages SE artifact replication and storage
+type Coordinator struct {
+	retryIntervals []time.Duration
+	encKey         []byte // Encryption key for SE artifacts
+	p2p            P2P
```
fredcarle, line 51: suggestion: Seems like you don't need the `P2P` interface. you could have the `client.Host` directly here and a field `replicatorIDsGetter func(collectionID string) []string`. Would make this even more independant.

internal/se/coordinator.go
```diff
@@ -0,0 +1,389 @@
...
+
+	return rc, nil
+}
+
+func (rc *Coordinator) Close() {
+	rc.cancel()
+}
+
+// reconstructIdentity reconstructs an Identity from stored public key information
+func (rc *Coordinator) reconstructIdentity(
```
fredcarle: thougth: The fact that you named the receiver `rc` get me confused quite a bit because I always forget that it represents a `Coordinator`.

internal/se/se.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		artifacts = append(artifacts, artifact)
+	}
+
+	return artifacts, nil
+}
+
+// generateFieldArtifact generates a single SE artifact for a specific field value.
+func generateFieldArtifact(
+	ctx context.Context,
```
fredcarle: suggestion: `ctx` is never used. Might as well remove it.

internal/db/collection.go
```diff
@@ -782,6 +783,7 @@ func (c *collection) save(
 			Cid:          link.Cid,
 			CollectionID: c.Version().CollectionID,
 			Block:        headNode,
+			Identity:     ident,
```
fredcarle: question: Why are you adding the identity to the update? Or more importantly, why is this needded for SE?

internal/se/coordinator.go
```diff
@@ -0,0 +1,389 @@
...
+func (rc *Coordinator) generateSEArtifacts(
+	ctx context.Context,
+	docID, collectionID string,
+	fieldNames []string,
+) ([]secore.Artifact, error) {
+	cols, err := rc.db.GetCollections(ctx, client.CollectionFetchOptions{
+		CollectionID: immutable.Some(collectionID),
+	})
+	if err != nil {
+		return nil, NewErrFailedToGetCollection(err)
```
fredcarle: thought: Wrapping errors doesn't add much value if the error comes from a Defra function that "should" already provide that information.

internal/se/coordinator_query_test.go
```diff
@@ -0,0 +1,172 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package se_test
```
fredcarle: question: Why is this a different package?

internal/se/se.go
```diff
@@ -0,0 +1,201 @@
...
+	docIDs := make([]string, 0, len(docIDSet))
+	for docID := range docIDSet {
+		docIDs = append(docIDs, docID)
+	}
+
+	return docIDs, nil
+}
+
+// FieldQuery represents a query for a specific encrypted field
+type FieldQuery struct {
```
fredcarle: suggestion: This doesn't seem like it needs to be public.

internal/se/coordinator_retry.go
```diff
@@ -0,0 +1,227 @@
...
+		log.ErrorContextE(ctx, "Failed to create transaction on retry", err, corelog.String("PeerID", peerID))
+		return
+	}
+	defer clientTxn.Discard()
+	txn := datastore.MustGetFromClientTxn(clientTxn)
+	ctx = datastore.CtxSetTxn(ctx, txn)
+
+	log.InfoContext(ctx, "Retrying SE replicator", corelog.String("PeerID", peerID))
+
+	identity, err := rc.reconstructIdentity(retryInfo.PublicKey, retryInfo.KeyType)
```
fredcarle: question: The identity never seems to actually be used. Why do we need to reconstruct it?

internal/db/config.go
```diff
@@ -25,11 +25,12 @@ const (
...
 type dbOptions struct {
-	maxTxnRetries  immutable.Option[int]
-	identity       immutable.Option[identity.Identity]
-	disableSigning bool
-	p2p            immutable.Option[client.Host]
-	retryIntervals []time.Duration
+	maxTxnRetries           immutable.Option[int]
+	identity                immutable.Option[identity.Identity]
+	disableSigning          bool
+	searchableEncryptionKey []byte
```
fredcarle, line 31: suggestion: It seems like there is at least one case where this would be set to `nil`. Might be worth making it it and optional type so that by default it would be `None`.

tests/integration/encryption/peer_acp_test.go
```diff
@@ -220,6 +220,23 @@ func TestDocEncryptionACP_IfUserHasAccessButNotNode_ShouldNotFetch(t *testing.T)
 					"Users": []map[string]any{},
 				},
 			},
+			// If the node doesn't have rights to the doc, it can't do merge
```
fredcarle: question: Is this a new behaviour? Why do you think this extra request adds any value to the test?

tests/integration/searchable_encryption/peer_create_test.go
```diff
@@ -0,0 +1,90 @@
...
+			testUtils.Request{
+				NodeID: immutable.Some(0),
+				Request: `
+					query {
+						encrypted_User(filter: {age: {_eq: 33}}) {
+							docIDs
+						}
+					}`,
+				Results: map[string]any{
+					"encrypted_User": gomega.Not(gomega.BeEmpty()),
```
fredcarle, line 48: todo: I'm not sure I understand what you're testing here. Can you please document it. Some for the one bellow.

tests/integration/searchable_encryption/peer_create_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/sourcenetwork/immutable"
+)
+
+func TestEncryptedIndexCreatePeer_SchemaWithEncryptedIndex_ShouldGenerateGQL(t *testing.T) {
+	test := testUtils.TestCase{
+		KMS:                        testUtils.KMS{Activated: true},
+		EnableSearchableEncryption: true,
+		Actions: []any{
+			testUtils.RandomNetworkingConfig(),
+			testUtils.RandomNetworkingConfig(),
```
fredcarle, line 30: question: Why create two instances if the test only uses one?

tests/integration/searchable_encryption/peer_delete_test.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+			testUtils.Request{
+				NodeID: immutable.Some(0),
+				Request: `
+					query {
+						encrypted_User(filter: {age: {_eq: 21}}) {
+							docIDs
+						}
+					}`,
+				ExpectedError: "Cannot query field \"encrypted_User\" on type \"Query\".",
```
fredcarle, line 66: thought: I'm not sure that's a great DX. We can bring it up for discussion during the standup.

tests/integration/searchable_encryption/peer_query_test.go
```diff
@@ -0,0 +1,426 @@
...
+			},
+			testUtils.CreateDoc{
+				NodeID: immutable.Some(0),
+				Doc: `{
+					"name":	"John",
+					"age":	21
+				}`,
+				IsDocEncrypted: true,
+			},
+			testUtils.Wait{
```
fredcarle: question: Do you intend to use the event bus to wait more efficiently?

tests/integration/searchable_encryption/replicator_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/onsi/gomega"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+func TestSEReplicator_WithEncryptedIndexAndCreate_ShouldReplicateArtifacts(t *testing.T) {
```
fredcarle: suggestion: This test name is misleading. It not testing replication anymore so than the tests in `peer_query_test.go`. It is however testing the retry mechanics. Please rename to reflect this.

tests/integration/searchable_encryption/replicator_test.go
```diff
@@ -0,0 +1,92 @@
...
+	test := testUtils.TestCase{
+		KMS:                        testUtils.KMS{Activated: true},
```
fredcarle: question: Why is the KMS necessary in this case?

internal/db/collection.go
islamaliev: in order to retry pushing SE tags to replicators we need identity as it's used for generation of tags. It's being stored with retry info

internal/db/config.go
islamaliev: I don't understand which case you mean. And I'm not sure it's makes much sense to make optional something that intrinsically optional.

internal/db/errors.go
islamaliev: oh, I think it was some kind of git merge leftover

internal/planner/planner.go
islamaliev: done

internal/planner/planner.go
islamaliev: SE stuff is exposed to users only through gql. I don't think this struct should be there.

internal/request/graphql/schema/collection.go
islamaliev: you are right, I think I just forgot to move it there.

internal/se/coordinator.go
islamaliev: done

internal/planner/planner.go
AndrewSisley: I agree with Islam, if it don't need to be user-public, it shouldn't be.

internal/se/coordinator.go
islamaliev: I changed the constructor to take client.Host explicitly, but passing a closure doesn't see to be very elegant.

internal/se/coordinator.go
islamaliev: it used to be called `ReplicatorCoordinator` :\
That's why I personally prefer to call receivers `this` or `self` (don't care about the idiomatic Go), because it makes things clearer and immune to struct renames.

internal/se/coordinator_query_test.go
islamaliev: I had some issues with mocks that are used in unit tests that caused circular dependencies. I changed stuff since then, but forgot to rename the package

internal/se/coordinator.go
islamaliev: true, remove unnecessary ones

internal/se/se.go
islamaliev: done

tests/integration/searchable_encryption/peer_create_test.go
islamaliev: added a comment:
`// Add peers to enable p2p so that SE gql queries are generated`

tests/integration/searchable_encryption/replicator_test.go
islamaliev: it's not, removed

tests/integration/searchable_encryption/replicator_test.go
islamaliev: renamed

tests/integration/encryption/peer_acp_test.go
islamaliev: actually this test should ideally be testing "replication" permission that we don't have at the moment. I just decided to keep it here for now.
Are you concerned about the test in general or specifically "commits" query?

tests/integration/searchable_encryption/peer_create_test.go
islamaliev: well, the test says "ShouldGenerateGQL". So we just test that "encrypted_User" can be queried without errors.
The difference between this test and the one below is that this specifies the encrypted index in the schema while the other one with the dedicated command to make sure both ways trigger gql generation.

tests/integration/searchable_encryption/peer_delete_test.go
islamaliev: Why dot you think it's not great and what exactly do you propose?
Sure we can discuss it.

tests/integration/searchable_encryption/peer_query_test.go
islamaliev: I did already. Just forgot to replace them here

internal/se/se.go
islamaliev: you are right. Changed

internal/se/coordinator_retry.go
islamaliev: holy crap, you are right. I was sure I added it. Added now. Thanks for the catch!

internal/db/collection_index.go
islamaliev: I usually mark as done when it's committed locally.

internal/planner/planner.go
fredcarle: It's used on DB and thus would become public. I think either it has to be moved or the pattern needs to change.

internal/db/collection.go
fredcarle: But this is the node identity. Why do you need to store that same info multiple times. And why would it need to be published on the bus?

internal/db/config.go
fredcarle: Discussed in person. Not using Optional here is fine.

tests/integration/encryption/peer_acp_test.go
fredcarle: Discussed in person. Only the comment should be updated.

internal/db/collection.go
islamaliev: it is externally provided identity, but thinking about it I realized you are right, using node's identity makes more sense. I switched.
Thanks for bringing this up

internal/db/collection_index.go
AndrewSisley: @islamaliev This still appear to not be done - have you forgotten to push?

AndrewSisley: :) They appear in the validation rules, but are still defined here - did you forget to delete these?

internal/db/p2p/protocol/comm_channel.go
```diff
@@ -0,0 +1,131 @@
...
+
+	replyPtr := ReplyP(&reply)
+	replyPtr.SetMessageID(reqPtr.GetMessageID())
+	err = message.SendAndForget(ctx, c, replyPtr, peerID, c.responseEndpoint)
+}
+
+func (c *commChannel[Req, Reply, ReqP, ReplyP]) onResponse(stream io.Reader, peerID string) {
+	var reply Reply
+	replyPtr := ReplyP(&reply)
+	_ = message.Receive(stream, peerID, c, replyPtr)
```
AndrewSisley: todo: Please either handle this error (logging counts, if you feel like that is more appropriate), or document why it is desirable/okay to discard this.

internal/db/p2p/replicator.go
```diff
@@ -297,18 +301,36 @@ func (p *P2P) pushLogToReplicators(lg event.Update) {
...
+				}
+				if _, err := p.replicatorProtocol.SendRequest(ctx, pushLogReq, peerID); err != nil {
 					log.ErrorE(
 						"Failed pushing log",
 						err,
 						corelog.String("DocID", lg.DocID),
 						corelog.Any("CID", lg.Cid),
 						corelog.Any("PeerID", peerID))
+					if !lg.IsRetry {
+						_ = p.handleReplicatorFailure(ctx, peerID, lg.DocID)
```
AndrewSisley: todo: Please document why discarding this error is desirable

internal/se/coordinator.go
```diff
@@ -0,0 +1,361 @@
...
+	ctx          context.Context
+	cancel       context.CancelFunc
+	nodeIdentity immutable.Option[acpIdentity.Identity]
+}
+
+// NewCoordinator creates a new coordinator
+func NewCoordinator(
+	p2p P2P,
+	host client.Host,
+	db DB,
```
AndrewSisley, line 65: praise: This looks a much better dependency tree, thank you very much Islam, and thanks for removing the rootstore access!

internal/se/coordinator_retry.go
```diff
@@ -0,0 +1,217 @@
...
+			if err := txn.Peerstore().Set(ctx, iter.Key(), b); err != nil {
+				log.ErrorContextE(ctx, "Failed to update SE retry info", err)
+				continue
+			}
+
+			if err = txn.Commit(); err != nil {
+				log.ErrorContextE(ctx, "Failed to commit transaction on retry", err)
+			}
+
+			go coordinator.retrySEArtifacts(ctx, key.PeerID, retryInfo)
```
AndrewSisley: question: Why is this spawning a new Go routine?  It *looks* unwanted to me, as the spawning func is already async, and this will just complicate the data flow as well as permit the SE stuff to consume more resources than it should be permitted to (these new routines will compete with core database operations, many of which are in user-time).

todo: Please document why this absolutely must be done in a new routine.

suggestion: Remove the routine.

internal/se/se.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/encoding"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	secore "github.com/sourcenetwork/defradb/internal/se/core"
+)
+
+// storeArtifacts stores SE artifacts directly in the datastore.
+// This is called by the server when receiving artifacts from peers.
+func storeArtifacts(ctx context.Context, ds corekv.ReaderWriter, artifacts []secore.Artifact) error {
```
AndrewSisley, line 32: todo: I think this comment, and maybe the function name is misleading, and they should be adjusted.

There are 7 properties on `secore.Artifact` but only 4 of them are stored here.

internal/db/collection_index.go
islamaliev: This function is used by `CreateEncryptedIndex` method that is part of public API and tests if a new encrypted index can be added.

The other function `validateEncryptedIndexesOnCollection` is testing if the collection with defined on it encrypted indexes is in a valid state.

They are similar, but different.

internal/db/p2p/protocol/comm_channel.go
islamaliev: done

internal/db/p2p/replicator.go
islamaliev: I just moved this stuff from other place. Added log

internal/se/se.go
islamaliev: This used to be a public method. Remove unnecessary line of documentation.

Great spot of 7 vs 4. These fields are actually not needed. I removed them.

internal/se/coordinator_retry.go
islamaliev: it was just more or less copy&pasted from regular replication code. Btw we plan to merge them in follow up work.

I removed goroutine for now.

internal/db/db.go
```diff
@@ -538,3 +551,5 @@ func printStore(ctx context.Context, store corekv.ReaderWriter) error {
 
 	return iter.Close()
 }
+
+// QueryDocIDsWithSETags queries SE artifacts from replicators based on field values.
```
fredcarle: todo: You seem to have forgotten to remove this function doc. Please remove it.

internal/planner/planner.go
```diff
@@ -84,26 +85,38 @@ type PlanContext struct {
 	context.Context
 }
 
+type DB interface {
```
fredcarle: question: Why do you want to wrap the `client.TxnStore` interface instead of using it directly?

internal/request/graphql/schema/errors.go
```diff
@@ -66,7 +71,7 @@ var (
 	ErrPolicyInvalidIDProp       = errors.New(errPolicyInvalidIDProp)
 	ErrPolicyInvalidResourceProp = errors.New(errPolicyInvalidResourceProp)
 	ErrFieldTypeNotSpecified     = errors.New(errFieldTypeNotSpecified)
-	ErrInvalidTypeForContraint   = errors.New(errInvalidTypeForContraint)
+	ErrInvalidTypeForContraint   = errors.New(errInvalidTypeForConstraint)
```
fredcarle: todo: Could you also fix the variable name? It also has the typo.

internal/se/coordinator_retry.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+const (
+	// retryLoopInterval is the interval at which the retry handler checks for
+	// SE artifacts that are due for a retry. Same as document replicator.
+	retryLoopInterval = 2 * time.Second
+)
+
+// SERetryInfo stores retry information for failed SE replications
+type SERetryInfo struct {
```
fredcarle: suggestion: This doesn't need to be public.

tests/integration/encryption/peer_acp_test.go
```diff
@@ -220,6 +220,23 @@ func TestDocEncryptionACP_IfUserHasAccessButNotNode_ShouldNotFetch(t *testing.T)
 					"Users": []map[string]any{},
 				},
 			},
+			// If the instance doesn't have rights to the doc, it can't do block sync
+			// and therefore has no heads. So commits should be empty.
```
fredcarle: suggestion: `... and therefore doesn't have the related commit blocks`. I think we had discussed this previously but saying it `has no heads` isn't the reason why `commits` returns empty.

tests/integration/searchable_encryption/peer_acp_test.go
```diff
@@ -0,0 +1,171 @@
...
+				Identity: testUtils.NodeIdentity(0),
+				Request: `
+					query {
+						commits {
+							delta
+						}
+					}
+				`,
+				// this replicator doesn't have access to the document, so it can't even
+				// store heads. Once we introduce a dedicated permission for replication,
```
fredcarle: suggestion: Similar to above, please rephrase to remove the mention of heads.

fredcarle (approved):
Thanks for your patience through this long review process. I know we've discussed this at length in a standup but I'll reiterate that we should try to keep the PRs smalled in the future and that should help a lot to make the process quicker and less painful for everyone.

Just a few comments to resolve before merge.

internal/db/db.go
islamaliev: done

internal/planner/planner.go
islamaliev: removed

internal/request/graphql/schema/errors.go
islamaliev: done

internal/se/coordinator_retry.go
islamaliev: made private

tests/integration/encryption/peer_acp_test.go
islamaliev: done

tests/integration/searchable_encryption/peer_acp_test.go
islamaliev: done

PR #3905, opened 2025-08-02
feat: Gate schema add op with NAC

### LINKED ISSUE #3904: Gate NAC Access to AddSchema Op
Gate NAC Access to AddSchema Op

shahzadlone:
## Relevant issue(s)
Resolves #3904

## Description
- Follow-up to #3758 
- Gate Schema Add Node Operation
- Ensure `admin` behaves properly with new operation.

## Additional Context
- Original PR when opened only had this commit: https://github.com/sourcenetwork/defradb/pull/3905/commits/b307e6a1c22c429972176a24883f4ff698078774
- This PR (https://github.com/sourcenetwork/defradb/pull/3918) moved the test action under `tests/action` folder
- So, the first commit handles the refactoring of testing framework to add the missing identity util functions in the `action` package.

## How has this been tested?
Integration tests.

- WSL 2 (Manjaro)

### DIFF
--- a/acp/types/types.go
+++ b/acp/types/types.go
@@ -88,6 +88,7 @@ const (
 	NodeNACStatusPerm
 	NodeNACRelationAddPerm
 	NodeNACRelationDeletePerm
+	NodeSchemaAddPerm
 )
 
 // List of all valid resource interface permissions for node access control, the order of
@@ -108,6 +109,7 @@ var RequiredResourcePermissionsForNode = []string{
 	"nac-status",
 	"nac-relation-add",
 	"nac-relation-delete",
+	"schema-add",
 }
 
 const NodeACPObject = "NodeObject"
@@ -151,6 +153,8 @@ resources:
       nac-relation-add:
         expr: owner + admin
       nac-relation-delete:
+        expr: owner + admin
+      schema-add:
         expr: owner + admin
 
     relations:
--- a/tests/action/identity.go
+++ b/tests/action/identity.go
@@ -8,66 +8,23 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-package tests
+package action
 
 import (
 	"context"
-	"strconv"
+	"time"
 
 	"github.com/sourcenetwork/immutable"
 	"github.com/stretchr/testify/require"
 
 	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
-	"github.com/sourcenetwork/defradb/tests/action"
 	"github.com/sourcenetwork/defradb/tests/state"
 )
 
-// NoIdentity returns an reference to an identity that represents no identity.
-func NoIdentity() immutable.Option[state.Identity] {
-	return immutable.None[state.Identity]()
-}
-
-// AllClientIdentities returns user identity selector specified with the "*".
-func AllClientIdentities() immutable.Option[state.Identity] {
-	return immutable.Some(
-		state.Identity{
-			Kind:     state.ClientIdentityType,
-			Selector: "*",
-		},
-	)
-}
-
-// ClientIdentity returns a user identity at the given index.
-func ClientIdentity(indexSelector int) immutable.Option[state.Identity] {
-	return immutable.Some(
-		state.Identity{
-			Kind:     state.ClientIdentityType,
-			Selector: strconv.Itoa(indexSelector),
-		},
-	)
-}
-
-// ClientIdentity returns a node identity at the given index.
-func NodeIdentity(indexSelector int) immutable.Option[state.Identity] {
-	return immutable.Some(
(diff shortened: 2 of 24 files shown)


tests/action/identity.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+
+	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+const (
+	// AuthTokenExpiration is the expiration time for auth tokens.
+	AuthTokenExpiration = time.Minute * 1
+)
```
shahzadlone, line 27: question: Are people happy with this being here or should this be in state?

AndrewSisley: I have a preference to not place constants in state - any reason why you might want it there?

tests/state/state.go
```diff
@@ -183,6 +191,12 @@ type State struct {
 	// The type of client currently being tested.
 	ClientType ClientType
 
+	// The type of Document ACP
+	DocumentACPType DocumentACPType
```
AndrewSisley, line 195: thought: I was actually wanting to use this today and was wondering what the status was - I didnt realize it had been sat waiting for review :P

Thanks for moving it!

tests/action/identity.go
shahzadlone: I have no preference. So you are happy in it's current spot?

acp/types/types.go
```diff
@@ -152,6 +154,8 @@ resources:
         expr: owner + admin
       nac-relation-delete:
         expr: owner + admin
+      schema-add:
```
AndrewSisley, line 157: suggestion: I think `PatchCollection` should also use this permission, as long term users will be able to add collections via patches (artificially disabled atm).

There is perhaps a case for splitting up the permissions within `Patchollection`, dpending on what exactly the user is patching, but that complication can probably be introduced later if we want.

This (these?) permission(s) will probably also need to be checked during the P2P sync of collection versions (when we add the feature later this sprint).

tests/action/identity.go
AndrewSisley: yes :) Thanks

acp/types/types.go
shahzadlone: Here is what the patch collection gate commit will look like roughly (has been changed a little bit according to some PR feedback from the NAC PR):
https://github.com/sourcenetwork/defradb/pull/3811/commits/72866d83197bbce78d7820b6dc918622e0133737

What is the suggestion in this PR though?

> P2P sync of collection versions (when we add the feature later this sprint).

I am not a 100% sure of the location and feature you are referring to, can you expand more?

tests/state/state.go
shahzadlone: 😄 Cheers

acp/types/types.go
AndrewSisley: > the patch collection gate commit

I don't think that makes sense. If users lack schema-add permission but have collection-patch they will be able to bypass schema-add and just add collection versions via patches.

> I am not a 100% sure of the location and feature you are referring to, can you expand more?

By the end of this sprint it is planned that CollectionVersions will be syncable across nodes using the blockstore/p2p system. This needs to respect NAC permissionspermissions (CRUD collection version permissions).

shahzadlone: Sorry, accidentally edited your previous comment rather than making a comment of my own, I am leaving this as a comment here even though we discussed on call, so can ignore parts that might not make sense:

> I don't think that makes sense.  If users lack `schema-add` permission but have `collection-patch` they will be able to bypass `schema-add` and just add collection versions via patches.

Things like this is what we want to really optimize once all gates are in, but since you have found this already we can use the same permission for schema add and patch. Will update the patch gate implementation accordingly. Also do you have a preference for the name of the permission?

> By the end of this sprint it is planned that `CollectionVersions` will be syncable across nodes using the blockstore/p2p system.  This needs to respect NAC permissions (CRUD collection version permissions).

Gotcha! Yes we will then likely need to gate somewhat similarly how we do for maybe DAC documents that need to sync?

AndrewSisley: > Will update the patch gate implementation accordingly

Thanks Shahzad :)

> Also do you have a preference for the name of the permission?

Nothing strong, other than that it should be `collection` or `collection-version` instead of `schema`, `schema` is kind of correct in the short term for `AddSchema`, but incorrect for patches (and probable long-term).

> Yes we will then likely need to gate somewhat similarly how we do for maybe DAC documents that need to sync

Yes, I'll need to be careful with this when implementing (it seems very likely that you'll have merged the NAC stuff before I implement the collection-p2p stuff)

PR #3918, opened 2025-08-06
test(i): Convert schema update action to testo

### LINKED ISSUE #3916: Migrate SchemaUpdate test action to testo

AndrewSisley:
## Relevant issue(s)

Resolves #3916

## Description

Converts the schema update action to testo.

Also renames it to `AddSchema`, finally matching the production function's name.  Also reworks the `Replace` stuff so it now works with everything and without tests having to declare the previous boilerplate around it.

Everything in the new `utils.go` and `results.go` is copy-pasted, the originals will be deleted when the last old-type actions that reference them get converted to testo. 

I suggest reviewing commit by commit, the second commit is just a find and replace and isnt worth reading IMO.


tests/action/replace.go
```diff
@@ -0,0 +1,66 @@
...
+//
+// Supporting action properties will replace any templated elements with data drawn from these
+// sets.
+var templateDataGenerators = map[string]func(*state.State, int) map[string]string{
+	"Policy": func(s *state.State, nodeID int) map[string]string {
+		nodesPolicyIDs := s.PolicyIDs[nodeID]
+
+		res := map[string]string{}
+		for i, policyID := range nodesPolicyIDs {
+			res["Policy"+strconv.Itoa(i)] = policyID
```
shahzadlone, line 36: Praise: Thanks, removing the need to specify the replaces manually per test.

tests/action/add_schema.go
```diff
@@ -0,0 +1,69 @@
...
+package action
+
+import (
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+// AddSchema is an action that will add the given GQL schema to the Defra nodes.
+type AddSchema struct {
```
shahzadlone, line 20: Praise: thanks renaming to this!

shahzadlone (approved):
This is a very nice PR! Thanks for the extra goodies + the refactor to test repo.

shahzadlone:
question: What are the remaining tests that still use the old type? And is there a blocker on why they can't transition?

AndrewSisley:
> question: What are the remaining tests that still use the old type? And is there a blocker on why they can't transition?

I think you were reviewing an old commit?  The old type has been deleted.

tests/action/add_schema.go
AndrewSisley: Only took me ~4 years :rofl:

shahzadlone:
> > question: What are the remaining tests that still use the old type? And is there a blocker on why they can't transition?
> 
> I think you were reviewing an old commit?  The old type has been deleted.

I was reading the PR description:

> the originals will be deleted when the last old-type actions that reference them get converted to testo.

PR #3926, opened 2025-08-07
test(i): Remove all test descriptions

### LINKED ISSUE #3925: Remove the test Description field from all the tests
Remove the test Description field from all the tests

shahzadlone:
Resolves #3925 

## Description

To reproduce the 1st commit:

1) Put his bash function in your bashrc:
```
DeleteLineRg() {
    \rg "${1}" --files-with-matches | \xargs \sed -i "/${1}/d";
}
```

2) cd into `tests` folder and do:
```
DeleteLineRg "\tDescription: \".*\"\,"
```

3) To remove extra lines, go back to root directory and do:
`make lint:fix`## Relevant issue(s)

The rest was done manually (2nd commit).


tests/integration/utils.go
```diff
@@ -263,7 +263,7 @@ func executeTestCase(
 		logAttrs = append(logAttrs, corelog.Any("KMS", kms))
 	}
 
-	log.InfoContext(ctx, testCase.Description, logAttrs...)
+	log.InfoContext(ctx, "", logAttrs...)
```
shahzadlone, line 266: question: Are people happy with this? 

I can even remove this completely or log something else.

tests/integration/query/simple/with_order_test.go
```diff
@@ -827,7 +814,6 @@ func TestQuerySimple_WithInvalidOrderEnum_ReturnsError(t *testing.T) {
 func TestQuerySimple_WithMultipleOrderFields_ReturnsError(t *testing.T) {
 	tests := []testUtils.TestCase{
 		{
-			Description: "Simple query with multiple order fields and a single entry",
```
AndrewSisley, line 830: todo: I do think this test, and all other tests that are actually multiple tests squashed into one  test function, need to be split up before we can remove `Description`.  Without the description it becomes even harder to tell which test failed, and it is already hard enough.

Happy enough to argue on this though, as often when one of them fails I find it least difficult to delete/comment out tests until I figure out which one is failing - IIRC some of the have the same description anyway, and sometimes the logs are noisy/unclear enough to make test-deletion easier than trying to read the output.

shahzadlone: I like this thought, but isn't it a different issue to address?

AndrewSisley: It is a different issue that IMO needs to be addressed before we can remove description, or during this PR, as there arent too many tests left like this now I think.

shahzadlone: They are spanned over 10+ files, I have fixed like 3. Will continue to push through, but can be merged right after this PR I don't see why this needs to block it as it is in progress.

AndrewSisley: > I don't see why this needs to block it as it is in progress.

If you merge this PR before breaking them up, it becomes practically impossible to tell which sub test is failing (by reading the logs, you can still delete sub tests until they pass...).  It is not great atm, but you would be significantly worsening the situation.

I see that as a pretty good reason to not merge this PR until the sub tests are broken up.

shahzadlone: They wouldn't be failing though as I am just splitting them, I am not going to rebase this PR with that branch until all the tests are broken apart (the files that need to be changed are all opened in my vim buffer haha) I am converting the tests as we speak.

AndrewSisley: > They wouldn't be failing though as I am just splitting them

That's not what I'm concerned about.  I am worried about this:

- You merge this PR
- Someone is doing dev and causes a sub test to fail
- There is no reasonable way to tell which sub test has failed reading the test output, they must delete sub tests in order to figure out which one is failing

At the moment, whilst it is progressing towards the end goal, I see this PR as having a detrimental effect on the codebase.  This can be avoided by splitting the sub tests first.  Neither this PR, or the sub-test PR is a significant amount of work and IMO we lose very little by doing things in the order avoids causing short term damage to the codebase.

shahzadlone: While the probability of that is super low IMO. I understand your concern. I will just do the splitting in this PR then.

shahzadlone: - [x] Done in the commit: ` PR(FIX): Split all multiple tests`

PR #3937, opened 2025-08-11
test: Add TestSignature support to C client

### LINKED ISSUE #3917: Create Coverage Test for C Bindings
The C binding will need test coverage.

Context: #3787

### LINKED ISSUE #3919: TestSignature Support for C Client
The initial PR for the C bindings lacks support for several tests:

_TestSignatureVerify_WithValidData_ShouldVerify
TestSignature_WithBranchableCollection_ShouldSignCollectionBlocks
TestSignatureVerify_WithDifferentKeyType_ShouldVerify
TestSignatureVerify_WithWrongIdentity_ShouldError_

The reason for this is, I believe, that the integration test harness handles the creation of the nodes that are used by the other client test wrappers, but the C wrapper uses its own node on the back end. As a result, there is an identity mismatch. One proposed solution is to create a way to inject node into the C binding logic for the purpose of testing.

ChrisBQu:
## Relevant issue(s)

Resolves #3919
Resolves #3917

## Description

The initial PR of the C Bindings merged in with the following tests disabled, because they did not work correctly:

_TestSignatureVerify_WithValidData_ShouldVerify
TestSignatureVerify_WithDifferentKeyType_ShouldVerify
TestSignatureVerify_WithWrongIdentity_ShouldError_

They didn't work correctly because of the way our test harness managed nodes, and the identities of those nodes. The C Wrapper would initialize its own node upon creation, which led to an incompatibility with some of the tests. This has been addressed by allowing for the injection of the node object that is created by the test utility into the -- bringing the C Wrapper more in line with how the other wrappers behave.

Incidentally, making this change also allows us to run unit coverage tests, which run to completion as expected (no SIGTERM.) So this PR does include the test coverage. However, this has revealed that the `test-document-acp` job, specifically, throws errors. I believe addressing that is beyond the scope of this PR, so I will be creating a new issue for it.

## Tasks

## How has this been tested?

- WSL


shahzadlone:
Thanks for getting the coverage working :)

However, the coverage report shows that the bump of `+1.05` brings coverage to `74.25` :

image

But the initial c-bindings PR dropped the coverage by `-2.16` to `73.2` from `75.36`:
image

If the 200-300 lines that aren't covered are all error paths, then we can be a little less worried. I will review this in a bit more depth to see in case accidentally some coverage is being missed (I also am guessing some of that might be because we are not running the other configuration combinations with c-bindings and only the default case).

For example right now it runs only `collection-save` mutation type on only `linux` with document acp type local only. Do you thing coverage can benefit by running this with other configurations?

shahzadlone:
suggestion: You want to do a test by adding this run to run in combination with other configurations by adding it to `client-type`s in both the `test-basic` job and the `test-document-acp` job of the workflow.

headsup: Don't forget to change other relevant things in the config otherwise coverage won't be accurate and will be overwritten, for example update the command that it can run c-binding tests like previous other clients, can take a look at the composite action here: `.github/composites/test-coverage-with-artifact/action.yml` (in other words don't call `make test:c` but instead just make use of the `ENV` var like other clients ). Also the `coverage-artifact-name` needs to be updated for both the jobs.

.github/workflows/test-coverage.yml
shahzadlone, line 1: todo: Don't forget to remove the extra workflow we introduced in your previous PR

PR #3950, opened 2025-08-15
refactor: Utilize CGO handles for C bindings

### LINKED ISSUE #3917: Create Coverage Test for C Bindings

### LINKED ISSUE #3919: TestSignature Support for C Client

### LINKED ISSUE #3920: Network Test Support for C Client
Currently, network tests are disabled and skipped for the C client. This must be investigated, and necessary features to allow for the testing must be implemented.

### LINKED ISSUE #3961: Refactor C Bindings
The C bindings code was previously separated into bridge and logic packages. The bridge code exposed C functions, while the logic code was done using Go function signatures. The integration tests only used the Go function signatures, so that was one area that was lacking in the approach.

It should be refactored such that the integration tests can use the actual function signatures, using C types.

ChrisBQu:
## Relevant issue(s)

Resolves #3961
Resolves #3919
Resolves #3917
Resolves #3920

## Description

This refactor changes how the C binding package works. Before, it was split into two packages, bridge, and logic. The bridge package is the one that exported functions to C, while the logic package contained all of the logic, and utilized Go primitives and structs in all of the function signatures. This had several downsides, not least of which was that the integration testing only ever used the Go functions. This addresses that.

The new version uses CGO handles to be able to return `uintptr_t` that refer to the transactions and node objects. This allows us to consolidate everything into a single package, which now also contains the wrapper used for integration tests. There is one trick employed to allow this to work: the makefile build command will, at build time, rename the package from "cbindings" to "main", do the build, and then rename the package from "main" to "cbinding." This is to meet a requirement of CGO: that the package being built have the name "main."

The major revision to the c bindings is also applied to identities. Previously we passed in identity by private key, as a string. We now utilize CGO handles to be able to pass in identity as an object. See, also, that the return type of `IdentityNew` has been adjusted to fit this design.

**Other fixes**
This new approach to the c binding package allows us to pass in a node for use during integration tests in a simple, and direct way. This results in issue 3919 (https://github.com/sourcenetwork/defradb/issues/3919#issue-3298445597) , in which certain integration tests that were related to node identity did not work, being solved. Those tests now behave appropriately.

Additionally, coverage test has been included, addressing issue 3917 (https://github.com/sourcenetwork/defradb/issues/3917).

Further, this PR includes an implementation for the `P2PConnect` function, and a small adjustment to the `withRetryOnNode` function inside `tests/integration/utils.go`. The former was the missing piece that allows networking tests to be enabled for the C client. The ladder addresses an oddity of the C client: the C client does not return error objects, it returns stringified errors. So a small adjustment to some error catching code had to be made. The outcome of these changes is that networking tests are no longer skipped. As such, this PR addresses issue 3929 (https://github.com/sourcenetwork/defradb/issues/3920)

## Tasks

## How has this been tested?

- WSL


Makefile
```diff
@@ -436,10 +443,29 @@ fix:
 	@$(MAKE) mocks
 	@$(MAKE) docs
 	
+build-c-shared-linux: SHELL := /bin/bash
```
nasdf: thought: I think it'd be a bit cleaner to move this to`tools/scripts/build-c-shared-linux.sh` and simply call that script from the make command.

cbindings/build.sh
nasdf, line 1: nitpick: should be moved to `tools/scripts` if possible

cbindings/wrapper.go
nasdf, line 1: todo: If possible I would add `go:build test` to all of the test wrapper and test util files so they are not accidentally included in production builds.

nasdf (approved):
Great work! This is a nice refactor. 

I left a few minor things to consider, but no major blockers from me.

Makefile
```diff
@@ -194,8 +195,7 @@ deps:
 
 .PHONY: mocks
 mocks:
-	@$(MAKE) deps:mocks && \
-	find . -type d -name "mocks" -exec rm -r {} + && \
+	@$(MAKE) deps:mocks
```
shahzadlone: todo: Why was this removed? I just added this fix while generating mocks. Please add it back

ChrisBQu: I must have accidentally done that while resolving merge conflicts. My mistake. Will fix.

.github/workflows/test-coverage.yml
```diff
@@ -300,6 +300,30 @@ jobs:
...
+
+      - name: Run C bindings test coverage
+        run: make test:coverage-c
+
+      - name: Save coverage report in an artifact
+        uses: ./.github/composites/upload-coverage-artifact
+        with:
+          coverage-artifact-name: "coverage_c"
+          coverage-path: coverage.txt
+
```
shahzadlone: question/todo: I don't understand why we need to call this (c client) in a special way using `make test:coverage-c`.

More details, in the review comment: https://github.com/sourcenetwork/defradb/pull/3950#pullrequestreview-3133713749

shahzadlone (changes requested):
question: I don't understand why we need to call this (c client) in a special way using `make test:coverage-c` in the CI workflow? I am fine with the make rule like the other client make rules, so you can leave the make rule in.

thought: Isn't that the point of adding `c` client to the `client-type: [go, http, cli, c]` in the `test-basic` job. I don't see the purpose for the `test-c` job.

todo: Please remove the extra `test-c` job that is just repeating the same job for no reason.

todo: Also add this c client test to run in combination with document acp running with sourcehub. You should just need to add the `c` to the `client-type: [go, http, cli]` config value under the `test-document-acp` job.

shahzadlone:
issue: I also just noticed none of the `c` actions in the `test-basic` job are running properly, this is because you only added `c` to the `client-type` but didn't set the `DEFRA_CLIENT_C: ${{ matrix.client-type == 'c' }}` below. You can check that before there was 3 X 2 X 3 (18) jobs being fired up by the `test-basic` job, now there should be `4 X 2 X 3` (24).

Also, incase there are some matrix variances that c bindings is not supported with (then you can make it's custom matrix). Otherwise it looks like you are wanting to run it with combos of [(file, memory) X (gql, collection-named, collection-save)] i.e. `6`cmore jobs.

todo: Don't forget to also do the same with `test-document-acp` job when you do the previous `todo` I made.

You can check if this is fixed or not, once you do the above changes by looking under the running c job (make sure it is `true` not `false`):
image

Makefile
```diff
@@ -74,6 +74,7 @@ COVERAGE_DIRECTORY=$(PWD)/coverage
 COVERAGE_FILE=coverage.txt
 COVERAGE_FLAGS=-covermode=atomic -coverpkg=./... -args -test.gocoverdir=$(COVERAGE_DIRECTORY)
 
+PLAYGROUND_DIRECTORY=playground
```
fredcarle: todo: Accidentally re-added. Please remove.

Makefile
```diff
@@ -170,7 +171,7 @@ deps\:mocks:
 
 .PHONY: deps\:playground
 deps\:playground:
-	go generate -tags playground ./playground/...
+	cd $(PLAYGROUND_DIRECTORY) && npm install --legacy-peer-deps && npm run build
```
fredcarle: todo: Accidentally changed. Please revert.

Makefile
```diff
@@ -427,6 +435,7 @@ docs\:godoc:
 .PHONY: toc
 toc:
 	bash tools/scripts/md-toc/gh-md-toc --insert --no-backup --hide-footer --skip-header README.md
+	bash tools/scripts/md-toc/gh-md-toc --insert --no-backup --hide-footer --skip-header playground/README.md
```
fredcarle: todo: Accidentally re-added. Please remove.

cbindings/testutils.go
```diff
@@ -28,86 +34,49 @@ import (
 	"github.com/sourcenetwork/defradb/internal/datastore"
 )
 
-// setupTests is a function that initializes and starts the globalNode (in memory), for the tests
-func setupTests(n int, identityPrivateKey string, enableNAC bool) {
-	var nodeOpts cbindings.GoNodeInitOptions
```
shahzadlone: question: where did this stuff move to btw?

cbindings/testutils.go
shahzadlone, line 1: suggestion: Rename to `test_utils.go`?

cbindings/testutils.go
ChrisBQu: This was actually removed, not relocated. Let me try to explain what is happening now.

If you look at line 53 of tests/integration/client_setup..go, you see that a node is passed in as a parameter to the NewCWrapper function. The CWrapper function, on line 97 of cbindings/wrapper.go, calls cgo.NewHandle on this node. This is the key. Because if you then look at any of the functions throughout the wrapper, you'll see that when they call one of the C functions, they pass in this handle (cast as a uintptr_t). For example, line 124:

res := ConvertAndFreeCResult(C.P2PsetReplicator(C.uintptr_t(w.handle), colStr, peerStr))

What happens, is that the C functions no longer utilize a map containing global nodes. The C functions actually take a pointer to the node, and then work on the given node. When you are using the C bindings, you can create one of these nodes and get the handle by calling the NewNode function (in cbindings/node.go). But here, we already have a node. The test harness has created the node. So instead of a function that sets up the tests... we just pass it in.

I hope this explanation was clear.

shahzadlone: I gotcha! Thanks for the explaination!

tests/integration/acp/nac/schema_add_test.go
```diff
@@ -13,26 +13,12 @@ package test_acp_nac
...
 	test := testUtils.TestCase{
-		SupportedClientTypes: immutable.Some(
-			[]state.ClientType{
-				// TODO: C binding test harness must be reworked to support this test
-				// See: https://github.com/sourcenetwork/defradb/issues/3919
-				state.GoClientType,
-				state.CLIClientType,
-				state.HTTPClientType,
-				state.JSClientType,
-			},
```
shahzadlone, line 33: praise: Thanks for sorting this out

tests/integration/acp/nac/relation_admin/schema_add_test.go
```diff
@@ -13,26 +13,12 @@ package test_acp_nac_relation_admin
...
-		SupportedClientTypes: immutable.Some(
-			[]state.ClientType{
-				// TODO: C binding test harness must be reworked to support this test
-				// See: https://github.com/sourcenetwork/defradb/issues/3919
-				state.GoClientType,
-				state.CLIClientType,
-				state.HTTPClientType,
-				state.JSClientType,
-			},
-		),
```
shahzadlone, line 34: praise: Thanks for sorting this out

Makefile
shahzadlone, line 1: nitpick: I see you removed the makefile rule you made, you could have left it in haha.

shahzadlone (approved):
LGTM

Thanks for sorting out running with sourcehub, and doing all the other requested changes. Looks really good now.

Let me know when ready to merge and I will remove the required job that is blocking this, and after merge I will make all the 7 new job actions (6+1) as required.

PR #3954, opened 2025-08-15
refactor: Cache collections in context

### LINKED ISSUE #3952: Cache CollectionVersions on ctx
Like short ids.

AndrewSisley:
## Relevant issue(s)

Resolves #3952

## Description

Cache collection versions in context.  Just like short IDs are handled.

## How has this been tested?

I ran the tests a bunch of times, averaging out the execution times using `time` (IM store and badger file).  PR branch was very slightly faster for both stores, but not by much, although I think badger-file essentially runs in memory for most of tests due to the badger cache size.


internal/db/description/collection.go
```diff
@@ -136,6 +184,15 @@ func GetCollectionsByCollectionID(
 	ctx context.Context,
 	collectionID string,
 ) ([]client.CollectionVersion, error) {
+	cache := getCollectionCache(ctx)
+	if cache.IsFullyPopulated {
+		return cache.CollectionsByID[collectionID], nil // todo - handle panic?
```
nasdf: question: if the panic mentioned here is from a nil map, should the map be pre-initialized?

nasdf (approved):

internal/db/description/collection.go
AndrewSisley: sorry - this todo should not have made it in to the PR.  Is a WIP comment.

The map should never be nil, however it will panic if an unknown collection id is provided (via user call, not possible internally).

I'll sort it out shortly, thanks for spotting.

- [x] Sort out todo (might be a second one somewhere too)

PR #3966, opened 2025-08-25
feat: Handle invalid CGO handles in C bindings

### LINKED ISSUE #3965: Create safety checks in C bindings
Currently, there are portions of the C binding codebase that would benefit from better error-catching. For example, most of the functions assume that the node or transaction handle is valid, and will then proceed to use the handle. This will crash in a way that does not return an error message to the user. Cases like this should be caught and returned gracefully.

ChrisBQu:
## Relevant issue(s)

Resolves #3965 

## Description

There were many portions of the c binding code base where a pointer (CGO handle) to a node, transaction, or identity CGO handle was passed in, and assumed valid without any error checking. This would result in a panic and crash if the node were actually invalid. While there is always the chance of the user passing in an invalid handle, it needs to be handled gracefully, which is what now happens. The functions will now return the error as part of the `Result` object, and return the error message to the user to help indicate what happened.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/cutils.go
+++ b/cbindings/cutils.go
@@ -166,8 +166,6 @@ func getIdentityFromPointer(identityPtr C.uintptr_t) (ident identity.Identity, e
 	switch v := v.(type) {
 	case identity.Identity:
 		return v, nil
-	case *identity.Identity:
-		return *v, nil
 	default:
 		return nil, fmt.Errorf(errInvalidTxnPointer, uintptr(identityPtr))
 	}
--- a/cbindings/cutils.go
+++ b/cbindings/cutils.go
@@ -138,7 +138,7 @@ func getStoreFromPointer(nodePtr C.uintptr_t) (store client.Store, err error) {
 	case client.Txn:
 		return v, nil
 	default:
-		return nil, fmt.Errorf(errInvalidStorePointer, uintptr(nodePtr))
+		return nil, fmt.Errorf(errInvalidCGOHandle, uintptr(nodePtr))
 	}
 }
 
@@ -150,7 +150,7 @@ func getNodeFromPointer(nodePtr C.uintptr_t) (n *node.Node, err error) {
 	}
 	n, ok := v.(*node.Node)
 	if !ok || n == nil {
-		return nil, fmt.Errorf(errInvalidStorePointer, uintptr(nodePtr))
+		return nil, fmt.Errorf(errInvalidCGOHandle, uintptr(nodePtr))
 	}
 	return n, nil
 }
@@ -167,7 +167,7 @@ func getIdentityFromPointer(identityPtr C.uintptr_t) (ident identity.Identity, e
 	case identity.Identity:
 		return v, nil
 	default:
-		return nil, fmt.Errorf(errInvalidTxnPointer, uintptr(identityPtr))
+		return nil, fmt.Errorf(errInvalidCGOHandle, uintptr(identityPtr))
 	}
 }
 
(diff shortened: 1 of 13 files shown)


cbindings/cutils.go
```diff
@@ -93,53 +94,95 @@ func returnNewIdentityResultC(status int, error string, n identity.Identity) C.N
...
 		EnableNodeACP:            int(cOptions.enableNodeACP),
-	}
+	}, nil
 }
 
-func getStoreFromPointer(nodePtr C.uintptr_t) client.Store {
+// getStoreFromPointer should be used by functions that can work on a node pointer or
+// on a transaction pointer.
+func getStoreFromPointer(nodePtr C.uintptr_t) (store client.Store, err error) {
+	// Protect against invalid handles
```
AndrewSisley: question: Which line(s?) are going to panic if the handle is invalid?

suggestion: Might be worth pulling the line(s) and defer out into a `tryFoo() error` func to limit the `recover` scope and make it easier to tell which call(s) may fail in a way that we cannot otherwise protect against.

EDIT: I see the same couple of lines are duplicated in a few funcs in this file - the suggestion should also help with the removal of said duplication (only as a happy bonus, I don't think it is worth making that a focus of any refactor).

cbindings/errors.go
```diff
@@ -45,7 +45,9 @@ const (
 	errGEttingSubscription   string = "error: could not retrieve subscription"
 
 	// Generic
-	errInvalidLensConfig string = "invalid lens configuration: %v"
-	errMarshallingJSON   string = "error marshalling JSON: %v"
-	errInvalidKeyType    string = "invalid key type: %v"
+	errInvalidLensConfig   string = "invalid lens configuration: %v"
+	errMarshallingJSON     string = "error marshalling JSON: %v"
+	errInvalidKeyType      string = "invalid key type: %v"
+	errInvalidStorePointer string = "invalid store pointer: %v"
```
AndrewSisley: todo: These should really be following the same error pattern that we use everywhere else in our code base.

AndrewSisley: question: Are you sure it is worth having separate errors for `errInvalidStorePointer` and `errInvalidTxnPointer`?

AndrewSisley (approved):
LGTM, just a couple of small comments.

question: Was this broken out of a main C binding refactor, or did you discover that this was entirely responsible for the problems found and you chose to go with this instead of a more comprehensive refactor?

cbindings/cutils.go
ChrisBQu: Good catch. I think I will factor that out. To specifically answer your question though:

EDIT: `h.Value()` will panic if called after `h = cgo.Handle(x)` where `x` is an invalid handle..

Also, because I can't respond to it directly on this thread chain, I'll answer your first question: this was not specifically broken out of the main refactor, per se. This is not an issue that affects the integration testing, for example, and there was no existing issue referencing it. So it was never *removed* from that PR. However, while writing some test code, I triggered the panic in question, and I made note of it as an area we could make a clear improvement in.

AndrewSisley: Thanks for the explanation (and for not bundling it into the branch you found it in) :)

cbindings/errors.go
ChrisBQu: It might not be. I don't have strong feelings either way on this one, but for now I've taken your suggestion and combined them into one.

ChrisBQu: For everyone's context: Andy and I discussed this issue a bit through direct message. This change will happen, but it will happen through a separate PR/issue.

PR #3970, opened 2025-08-26
refactor: Standardize errors in C bindings

### LINKED ISSUE #3969: Standardize error usage in cbindings package
The way error constants are utilized in the c bindings is a bit different than how they are utilized in other parts of the code base. Currently, the constants are only string constants (some of which contain format verbs.) 

The `errors.go` file should contain vars defined by `errors.New` and functions that build and return more complex errors, using `errors.Wrap`, `errors.New` and `errors.NewKV`

ChrisBQu:
## Relevant issue(s)

Resolves #3969 

## Description

The way that errors were utilized in the C bindings was somewhat different than in most other parts of our codebase. It largely just utilized strings with format verbs; it also used them in a lot of locations that perhaps it should not have used them. This PR seeks to do a few things. **One)** C binding functions should now prefer returning the error received from the Defra function, wrapped as a `Result` structure, where possible, instead of wrapping it as a new error type first. **Two)** When using a new error type unique to the C binding package, the code now avoids using strings with format verbs, instead opting to utilize functions which return new errors. When applicable, these functions will use `errors.NewKV` and `errors.Wrap` to attach additional information to the error. **Three**) Errors that had been left over, from previous commits, but which are no longer directly used anywhere, have simply been removed.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/collection.go
+++ b/cbindings/collection.go
@@ -19,8 +19,6 @@ import "C"
 import (
 	"context"
 	"encoding/json"
-	"errors"
-	"fmt"
 	"strings"
 
 	"github.com/sourcenetwork/immutable"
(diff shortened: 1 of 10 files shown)


shahzadlone (approved):
Nice cleanup!

LGTM

PR #3977, opened 2025-08-29
fix: Prevent blocking on server close

### LINKED ISSUE #3976: HTTP server close blocks when graphql subscription is active.

fredcarle:
## Relevant issue(s)

Resolves #3976 

## Description

This PR fixes an issue where the Defra would block on close when a graphql subscription was active. This was only problematic for graphql subscriptions over HTTP.

## Tasks

## How has this been tested?

Changed the test util and clients using HTTP to not automatically exit when all actions are done.

- MacOS

### DIFF
--- a/http/handler_store.go
+++ b/http/handler_store.go
@@ -294,9 +294,28 @@ func (s *storeHandler) ExecRequest(rw http.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(http.StatusOK)
 	flusher.Flush()
 
+	serverCtx, hasServerCtx := tryGetContexCtx(req)
+	var serverDone <-chan struct{}
+	if hasServerCtx {
+		serverDone = serverCtx.Done()
+	}
 	for {
 		select {
 		case <-req.Context().Done():
+			return
+		case <-serverDone:
+			// We need to check for closure of the server context
+			// otherwise the server won't gracefully shutdown until all
+			// connections are closed.
+			_, err := fmt.Fprintf(rw, "event: complete\n")
+			if err != nil {
+				return
+			}
+			_, err = fmt.Fprintf(rw, "data: {}\n\n")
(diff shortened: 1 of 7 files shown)


http/handler_store.go
```diff
@@ -294,10 +294,29 @@ func (s *storeHandler) ExecRequest(rw http.ResponseWriter, req *http.Request) {
...
 			return
+		case <-serverDone:
+			// We need to check for closure of the server context
+			// otherwise the server won't gracefully shutdown until all
+			// connections are closed.
+			_, err := fmt.Fprintf(rw, "event: complete\n")
+			if err != nil {
+				return
+			}
+			_, err = fmt.Fprintf(rw, "data: {}\n\n")
```
shahzadlone, line 314: question: why are there two newlines (`\n\n`)?

tests/integration/subscription/subscription_test.go
```diff
@@ -344,3 +344,22 @@ func TestSubscriptionWithUpdateAllMutations(t *testing.T) {
 
 	execute(t, test)
 }
+
+func TestSubscription_WithClose_WontBlock(t *testing.T) {
```
shahzadlone, line 348: Question: This test would fail before this PR, correct?

shahzadlone (approved):
Just some non-blocking questions.

LGTM

http/handler_store.go
fredcarle: Thats how SSE determines that the message is complete.

tests/integration/subscription/subscription_test.go
fredcarle: With the changes to the test clients and util but without the fix, yes. The test clients were forcing client connections to close which hid this issue.

getify:
awesome work!

PR #3992, opened 2025-09-08
fix: Allow JSON variables in GQL

### LINKED ISSUE #3989: JSON value not via external variable, must be inlined in graphql
schema setup:

```graphql
type KV {
  key: String @index(unique: true)
  value: JSON
  updatedAt: DateTime @index
}
```

that works. then I do this:

```graphql
mutation setupFoo($key:String!) {
  create_KV(input:[{ key:$key, value:{ bar:1 } }]) { key }
}
```

with variables:

```json
{ "key": "foo" }
```

and that works fine...

```graphql
query { KV(filter: { key: {_eq:"foo"} }) { key value } }
```

returns:

```json
{
  "data": {
    "KV": [
      {
        "key": "foo",
        "value": {
          "bar": 1
        }
      }
    ]
  }
}
```

Great. But if I try to externalize that JSON value via variables...

```graphql
mutation setupWoz($key:String!,$value:JSON) {
  create_KV(input:[{ key:$key, value:$value }]) { key }
}
```

with variables:

```json
{ "key": "woz", "value": { "bar": 1 } }
```

I get this error back:

```
[
  {
    "message": "Variable \"$value\" of type \"JSON\" used in position expecting type \"JSON\".",
    "locations": [
      {
        "line": 1,
        "column": 34
      },
      {
        "line": 2,
        "column": 41
      }
    ]
  }
]
```

I verified the field definition like this:

```graphql
{
  __type(name:"KVMutationInputArg"){
    inputFields{
      name
      type{ kind name ofType{ kind name ofType{kind name} } }
    }
  }
}
```

and that returns:

```json
{
  "data": {
    "__type": {
      "inputFields": [
        {
          "name": "key",
          "type": {
            "kind": "SCALAR",
            "name": "String",
            "ofType": null
          }
        },
        {
          "name": "updatedAt",
          "type": {
            "kind": "SCALAR",
            "name": "DateTime",
            "ofType": null
          }
        },
        {
          "name": "value",
          "type": {
            "kind": "SCALAR",
            "name": "JSON",
            "ofType": null
          }
        }
      ]
    }
  }
}
```

so the `value` field is a SCALAR JSON.

when the `{ bar: 1 }` comes in inline, that's compatible, but when it comes in as a variable, it doesn't pass the type matching.

fredcarle:
## Relevant issue(s)

Resolves #3989 

## Description

This PR fixes an issue where a JSON variable in GQL would cause a parsing error. This was due to the JSON scalar type being defined twice in the process and the parser interpreting it as two different types. The Blob scalar type usage was also preemptively changed.

## Tasks

## How has this been tested?

Added a test with a query that would fail before applying the fix.

- MacOS

### DIFF
--- a/internal/request/graphql/schema/collection.go
+++ b/internal/request/graphql/schema/collection.go
@@ -434,9 +434,9 @@ func defaultFromAST(
 	case types.DefaultDirectivePropDateTime:
 		value = gql.DateTime.ParseLiteral(arg.Value, nil)
 	case types.DefaultDirectivePropJSON:
-		value = types.JSONScalarType().ParseLiteral(arg.Value, nil)
+		value = types.JSON.ParseLiteral(arg.Value, nil)
 	case types.DefaultDirectivePropBlob:
-		value = types.BlobScalarType().ParseLiteral(arg.Value, nil)
+		value = types.Blob.ParseLiteral(arg.Value, nil)
 	}
 	// If the value is nil, then parsing has failed, or a nil value was provided.
 	// Since setting a default value to nil is the same as not providing one,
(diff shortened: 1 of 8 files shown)


AndrewSisley (approved):
LGTM, thanks for fixing this so quickly :)

shahzadlone (approved):
Cheers on the speedy fix. Other than some actions failing, LGTM

Issue #3993, opened 2025-09-08
Rename document specific client collection commands to be more consistent


shahzadlone:
Rename document specific client collection commands to be more consistent

Context: Discussed in standup and reached consensus on renaming the following `client collection` commands:
 
- Rename `collection update` -> `document update` 
- Rename `collection delete` -> `document delete` 
- Rename `collection get` -> `document read`
- Rename `collection docIDs` -> `document list`
- Rename `collection create` -> `document create`
 
- Keep `collection describe` as is, because it operates on collection.
- Keep `collection patch` as is because it operates on collection, and had pushback regarding not changing to `collection update` as `patch != update`.

Note: these weren't explicitly discussed: 
- ~~get -> read~~ (Discussed on discord, consensus is to stick with `get`: https://discord.com/channels/1374835078781468722/1466897550371061791/1470446696231211283)
- `docIDs` -> `list`

But naming here is according to the other related issue we want to do before v1: https://github.com/sourcenetwork/defradb/issues/3881

Once above is done, then do:
- Rename `schema add` -> `collection create` (#3994)

Discord: https://discord.com/channels/1374835078781468722/1390650826317627502/1413272524778770642

Here are some other related ones:
- https://github.com/sourcenetwork/defradb/issues/4107

fredcarle:
> get -> read

`get` will be more familiar to most people. It matches the HTTP verb.

shahzadlone:
> `get` will be more familiar to most people. It matches the HTTP verb.

I do personally like `get` more. But the reason I put `read` was because:
1) CRUD = create, read, update, and delete
2) We have `delete`, `update`, `read` ACP permissions

However, these are not strong reasons. Happy to follow consensus.

ChrisBQu:
> > `get` will be more familiar to most people. It matches the HTTP verb.
> 
> I do personally like `get` more. But the reason I put `read` was because:
> 
> 1. CRUD = create, read, update, and delete
> 2. We have `delete`, `update`, `read` ACP permissions
> 
> However, these are not strong reasons. Happy to follow consensus.

I would like to offer that my preference is also for `get`. I agree with Fred that it is more familiar.

[...]

PR #3996, opened 2025-09-09
feat: Gate collection patch op with NAC

### LINKED ISSUE #3995: Gate NAC Access to Collection Patch Op
Gate NAC Access to Collection Patch Op

shahzadlone:
Resolves #3995

## Description
- Gates collection patch command with `Collection Patch` perm
- Uses the same `Collection Patch` perm for `Schema Add` operation as well (soon to be `Collection Create`).
- Sharing the same perm between them because, patch command can delete/update/create collections.

## How has this been tested?
- Wrote tests for:
  - Admin relation
  - Restart persistent state maintained
  - Normal use

- Manjaro WSL2


acp/types/types.go
```diff
@@ -109,7 +109,7 @@ var RequiredResourcePermissionsForNode = []string{
 	"nac-status",
 	"nac-relation-add",
 	"nac-relation-delete",
-	"schema-add",
+	"collection-update",
```
AndrewSisley: todo: I thought we agreed on `collection-patch`?  `collection-update` is misleading unless you plan on adding `collection-create` and `collection-delete` and have us write and maintain a bunch of logic within `CollectionPatch` to make sure the net-result of the patch operation(s) is asserted against the correct NAC rules.

tests/integration/acp/nac/collection_patch_test.go
```diff
@@ -0,0 +1,73 @@
...
+package test_acp_nac
+
+import (
+	"testing"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestNAC_GatesCollectionPatch_AllowIfAuthorizedElseError(t *testing.T) {
```
AndrewSisley: thought: I think I have raised this before, but I do disagree with this testing strategy.  It requires too much conscious (and fallible) effort from maintainers, and adds a non-trivial amount of tests to read/maintain, and execute against unwanted multipliers.

IMO we can do much better than this and automate the NAC testing so that existing tests combined with a multiplier (will have to be coded in the old, non-testo framework for now), automatically cover all endpoints, including future ones - eliminating room for human error, and saving us from having to deal with a bunch of extra tests.

AndrewSisley: suggestion: This test appears to be several tests posing as a single one.   This makes debugging more complicated, and IMO causes the name to be incorrect as it says it errors, yet it also includes the test for a valid permissioned call.

tests/integration/utils.go
```diff
@@ -1116,9 +1116,12 @@ func patchCollection(
 	s *state.State,
 	action PatchCollection,
 ) {
-	_, nodes := getNodesWithIDs(action.NodeID, s.Nodes)
-	for _, node := range nodes {
+	nodeIDs, nodes := getNodesWithIDs(action.NodeID, s.Nodes)
+	for index, node := range nodes {
+		nodeID := nodeIDs[index]
+		s.Ctx = getContextWithIdentity(s.Ctx, s, action.Identity, nodeID)
```
AndrewSisley: todo: I can't spot why you are overwriting `s.Ctx` here, and arent just doing the below instead of this set/rest pattern:
```
ctx := getContextWithIdentity(s.Ctx, s, action.Identity, nodeID)
err := node.PatchCollection(ctx, action.Patch, action.Lens)
```
I assume there are a bunch of places that could be simplified like this.

AndrewSisley (dismissed):
I think this needs a couple of tweaks before merge.

acp/types/types.go
shahzadlone: You and I had a dedicated zoom call about changing to `collection update` for permission name to be CRUD oriented where we both reached consensus, however then when we discussed in standup regarding the user facing client commands where we said we want to stick to `collection patch`. Due to that confusion I think I forget to transition the permission terminology.

Thinking about this I agree with you, as it stands if we are merging the permissions, then this should reflect the same naming (we can make it more granular later if we want once/if we have `collection delete` and want to spend the time to maintain the logic within the patch implementation to detect the different delete/update/create).

On a similar note, would you like the `read` permission terminology for DAC as we have it now to also be renamed to `get`?  As it seems like people including me prefer the client command to be `get` not `read`. Context: https://github.com/sourcenetwork/defradb/issues/3993

- [x] Rename terminology to `collection-patch`

tests/integration/acp/nac/collection_patch_test.go
shahzadlone: I am curious as to how you want to automate the NAC testing, with which multiplier? Because testing NAC requires specific type of calls, and can not simply just be enabled and run with the same tests.

shahzadlone: To be fair it says: `AllowIfAuthorized` `ElseError` haha, I am doing both error and valid tests though so I understand what you mean. I did this to avoid having to duplicate the test setup. However I don't mind splitting the test if there is preference, so to understand your point, you are suggesting to break apart the error asserts into a separate test, and valid ones in another?

- [x] Split error and valid tests into separate ones.

tests/integration/utils.go
shahzadlone: This was likely after math of a bad conflict resolution, I had that way initially, will look into and change all occurrences.

- [x] Simplify set/rest pattern

tests/integration/acp/nac/collection_patch_test.go
AndrewSisley: The same N tests need to be written for every endpoint.  They all need to behave the same way, at least by default (we could do fancier stuff long term and this can be handle with an exclusion list).

We already have tests that call every endpoint.

Instead of writing the same tests for every endpoint, including new ones that get added in the future, we can write a rule (a multiplier(s)) that asserts this behaviour for everything. Mutate the existing test actions so that they execute with/without NAC permissions and assert that the expected errors (and lack of).

This way nothing gets forgotten (especially when adding new endpoints).  The cost of NAC to stuff is minimised (by default, for a new endpoint, the tests will fail until the 1-2 lines of prod code are added).  And we dont have a whole load of very boring and repetitive new tests to maintain.

It feels exactly the kind of problem that the matrix/multiplier system is there to solve.

AndrewSisley: > AllowIfAuthorized

Does anyone read the main body of test names :laughing: ?

> you are suggesting to break apart the error asserts into a separate test, and valid ones in another?

I am suggesting to test one thing in each test.  Test one identity, and one error (or lack of) in each test.  This applies to the multiplier thread too - one multiplier for each behaviour.

shahzadlone: > We already have tests that call every endpoint.

Here did you mean have some dedicated tests that call every endpoint within them (as in a centralized test that calls every known end point or that in general we have test coverage / tests testing every endpoint overall)? 

> Rest

If I understand you correctly you want to do a similar matrix approach that we use to run the CI test coverage matrix? i.e. using env vars like `DEFRA_DOCUMENT_ACP_TYPE` to configure a combination of behaviors that will then execute every existing test with that.

In that case the node acp is a little different, while it will be nice to do something that tests without us having to write more tests, I am not sure if it's that straight forward for the nac case, here are some questions/concerns:

1) You can't just flip on NAC by an ENV flag (nor should you be able to). The current system requires that in order to set an identity you need to configure it with a dedicated FBO("flesh-and-blood owner identity") that will become the owner, at the `defradb start` step.

2) Only authorized identities can then flip the state on and off.

3) Every time the database is started (it recovers the previous state, which is not controllable/configurable just by the start config flags once nac has been configured once by FBO, unless purged).

4) Even if we have one of the matrix that configures NAC using a default hardcoded FBO identity on start, in that case every existing previous test would now fail because they are (and should continue to be) running with no identity and therefore will have no access as once NAC is started, everything is banned without identity.

5) We still would need to have dedicated and situation specific tests that would test NAC state flipping tests (which we already have). Like `status`, `re-enable`, `disable`, `purge`, `start`, it would be so awkward to block some state tests to run when NAC state is automatically configured and on within the matrix, and some state tests to run when NAC state is off in the matrix.

Moreover:

- We could solve (4) by having every test magically use a hardcoded identity when NAC state in matrix is on, but I would be super against that as it will now be adding some magic config options in our testing that would not be super obvious to catch if a test is automatically using an identity when it should not be, and will ruin one of the most important parts of our current testing framework according to me, which is that you can see what the test action is being configured with right then and there without magic abstractions in the back.

Alternate solution(s):
I do have some cleaner solutions planned to improve the NAC testing without causing the matrix mess once most of the gates are in and we have a somewhat exhaustive list of all endpoints. Using a dedicated test action/utility to assert all existing endpoint behaviors are consistent, for example:
1) Test that every endpoint is blocked once NAC is configured, and using endpoint with no identity.
2) Test that every endpoint is blocked once NAC is configured, and using endpoint with unauthorized identity.
3) Test that every endpoint allows access when NAC is on and identity is authorized.
4) Test that every endpoint allows access when NAC is off identity has access. (note: this is what every existing test currently does so doesn't need to be added)
5) Test that input relationship with a relation (`admin` for example) if added, now gives access to all target endpoints that are in that test pattern.
6) Test that input relationship with a relation (`admin` for example) if removed, now revokes access to all target endpoints that are in that test pattern.


Note(s):
- This way we can do many other types of endpoint specific behavior testing very nicely and quickly. 
- (1) (2) (3) will all replace the pattern of testing that is in this file (i.e. the test this pr comment is on).
- (4) (5) This will make adding new relations and testing them be super clean.

shahzadlone: TBH I preferred the previous approach as I could see the behaviors of all three commands with different Identities in the same test, and required the setup once:
```go
			// We haven't authorized non-identities. So, this should error.
			testUtils.PatchCollection{
				Identity: testUtils.NoIdentity(),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} }
					]
				`,
				ExpectedError: "not authorized to perform operation",
			},

			// Wrong user/identity will also not be authorized.
			testUtils.PatchCollection{
				Identity: testUtils.ClientIdentity(2),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} }
					]
				`,
				ExpectedError: "not authorized to perform operation",
			},

			// This should work as the identity is authorized.
			testUtils.PatchCollection{
				Identity: testUtils.ClientIdentity(1),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} }
					]
				`,
			},
``` 

Nonetheless, since I didn't have a hard preference I split them up according to the suggestion.

Note: There are few other gates that will follow after this PR and will follow the same "blueprint" for their testing.

acp/types/types.go
```diff
@@ -88,7 +88,7 @@ const (
 	NodeNACStatusPerm
 	NodeNACRelationAddPerm
 	NodeNACRelationDeletePerm
-	NodeSchemaAddPerm
+	NodeCollectionPatchPerm
```
AndrewSisley, line 91: question: Why does permission lack the `NAC` prefix? And why do all the permissions have a `Node` prefix?

tests/integration/acp/nac/setup_then_start/collection_patch_test.go
```diff
@@ -0,0 +1,82 @@
...
+	test := testUtils.TestCase{
+		SupportedDatabaseTypes: immutable.Some(
+			[]state.DatabaseType{
+				// This test only supports file type databases since the setup steps will be done before
+				// the node is re-started with nac enabled (if it's in-memory it will loose setup state).
+				testUtils.BadgerFileType,
+			},
+		),
+		Actions: []any{
+			// Note: Since this is not an in-memory test, we can do the setup steps before nac is enabled.
```
AndrewSisley: question: Why complicate things though?

The existence of a comment, and unusual action location makes it look significant, yet the comment suggests it is not.

tests/integration/acp/nac/setup_then_start/collection_patch_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+func TestNAC_GatesCollectionPatchPreSetup_AllowIfAuthorizedElseError(t *testing.T) {
```
AndrewSisley: todo: This test does not make a lot of sense to me.  I think it either needs more/better documentation, or it needs to disappear.

Why would adding a GQL schema before restart, affect a NAC permission on an unrelated function after restart?

It feels like this test does not justify it's maintenance cost, and is just going to add to everyone's dev cycles (is another test to wait for, locally and in the CI).

It also tests multiple things (same conversation as an early comment on another test).

tests/integration/acp/nac/collection_patch_test.go
AndrewSisley: I think you have misunderstood how multipliers/test-environment-variables work.  They are not passed to the production code.

Mentioned briefly over discord, writing here for tracking - I can have a call with you at somepoint.

tests/integration/acp/nac/setup_then_start/collection_patch_test.go
fredcarle: question: Why is the behaviour different (is it anything other than the store being wiped clean on restart)? Why introduce a different setup pattern that risks confusing readers?

acp/types/types.go
shahzadlone: Depends how you look at it. From left to right it reads "For Node ACP, Collection Patch permission". `NAC` is in the names of enums is representing the NAC operations being gated, my the NAC system it self (for example: turning on nac / or off, is a nac op, that is also gated by nac system).

The current naming format is `<Client Facing ACP Type Name><Required Permission Name><The String "Perm">`

Here are DAC permission names:
```go
var RequiredResourcePermissionsForDocument = []string{
	"read",
	"update",
	"delete",
}
```

So the permission enum names:
```go
const (
	DocumentReadPerm DocumentResourcePermission = iota
	DocumentUpdatePerm
	DocumentDeletePerm
)
```

Similarly, NAC permission string names:

```go
var RequiredResourcePermissionsForNode = []string{
	"dac-bypass",
	"dac-enable",
	"dac-disable",
	"dac-purge",
	"dac-status",
	"dac-relation-add",
	"dac-relation-delete",
	"dac-policy-add",
	"nac-re-enable",
	"nac-disable",
	"nac-purge",
	"nac-status",
	"nac-relation-add",
	"nac-relation-delete",
	"schema-add",
    ...
}
```

Hence node acp permission enum names:
```go
const (
	NodeDACBypassPerm NodeResourcePermission = iota
	NodeDACEnablePerm
	NodeDACDisablePerm
	NodeDACPurgePerm
	NodeDACStatusPerm
	NodeDACRelationAddPerm
	NodeDACRelationDeletePerm
	NodeDACPolicyAddPerm
	NodeNACReEnablePerm
	NodeNACDisablePerm
	NodeNACPurgePerm
	NodeNACStatusPerm
	NodeNACRelationAddPerm
	NodeNACRelationDeletePerm
	NodeSchemaAddPerm
	...
)
```

tests/integration/acp/nac/setup_then_start/collection_patch_test.go
shahzadlone: The comment exists in case people are confused why the setup is being done before here.

I was testing this because wanting to make sure that a feature like NAC on defradb restart, still works as expected and no bugs are slipped through that may mess with the `NAC` system to fail on restart.

IMO this is important to have as most tests just test single running db instance and not nac working properly even after restart, when setup was done before the restart.

Nonetheless, this seems to be too controversial, and in the hoping of not delaying in merging this PR furthur, I will accept the suggestion asking to remove this (for this and all other gates).

shahzadlone: - [x] Remove this test

tests/integration/acp/nac/setup_then_start/collection_patch_test.go
shahzadlone: Explained in the other comment why I added this test, tldr: even if chance of this nac op failing on restart is low I still wanted a restart test, but I will remove it to adhere to this todo.

shahzadlone: - [x]  Removed

fredcarle (approved):
LGTM

PR #4004, opened 2025-09-11
fix: Signature selection panic

### LINKED ISSUE #4003: No signature selection causes panic
**Describe the problem**
When signatures are enabled and you query the `Commit` type either through a `_version` sub selection of a user type or through a direct `commits` query, but you omit the `signature` field on the selection, it panics.

**To Reproduce**
Ensure block signatures are enabled
```
query {
  User {
    _version {
      cid
    }
  }
}
```

Specifically do ***not*** include the `signature` field on the selection.

**Expected behavior**
It should return the query results as expected. But it panics instead.

**Platform**
`defradb dev-develop (b9c1fb48 2025-09-01) built with Go 1.24.6 linux/amd64`

jsimnz:
## Relevant issue(s)

Resolves #4003 

## Description

When document/block signature are enabled, and you query for a `Commit` type without selecting the `signature` type, it will panic.

This is caused by the `addSignatureFieldToDoc` code *always* attempting to add the signature whenever the `Commit` type is requested, but if you don't include the `signature` type, the mapper indexer lookup panics as it doesn't exist.

Note: The fix is applied in two spots. Proactively to prevent adding the signature entirely, and again reactively when trying to add the signature.

The surprising thing is that this wasn't caught by the test suite. Turns out, the only tests that use the `EnableSigning` either A) Never query the `Commit` type without also including the `signature` field B) Use the builtin `testUtils.VerifyBlockSignature` and therefore skip this bug altogether.

## Tasks

## How has this been tested?

Updated tests to cover this panic

- Ubuntu (WSL)

### DIFF
--- a/internal/planner/commit.go
+++ b/internal/planner/commit.go
@@ -380,7 +380,8 @@ func (n *dagScanNode) dagBlockToNodeDoc(block *coreblock.Block) (core.Doc, error
 		n.commitSelect.DocumentMapping.SetFirstOfName(&commit, request.DeltaFieldName, nil)
 	}
 
-	if block.Signature != nil {
+	if block.Signature != nil &&
+		n.commitSelect.DocumentMapping.IndexesByName[request.SignatureFieldName] != nil {
 		err := n.addSignatureFieldToDoc(*block.Signature, &commit)
 		if err != nil {
 			return core.Doc{}, err
@@ -446,8 +447,11 @@ func (n *dagScanNode) addSignatureFieldToDoc(link cidlink.Link, commit *core.Doc
 	if err != nil {
 		return err
 	}
-	sigFieldIndex := n.commitSelect.DocumentMapping.IndexesByName[request.SignatureFieldName][0]
-	sigMapping := n.commitSelect.DocumentMapping.ChildMappings[sigFieldIndex]
+	sigFieldIndexes, exists := n.commitSelect.DocumentMapping.IndexesByName[request.SignatureFieldName]
+	if !exists {
+		return NewErrMissingFieldSelection(request.SignatureFieldName)
+	}
+	sigMapping := n.commitSelect.DocumentMapping.ChildMappings[sigFieldIndexes[0]]
 
 	sigDoc := sigMapping.NewDoc()
 	sigMapping.SetFirstOfName(&sigDoc, request.SignatureTypeFieldName, sigBlock.Header.Type)
--- a/internal/planner/errors.go
+++ b/internal/planner/errors.go
@@ -17,6 +17,7 @@ const (
 	errFailedToClosePlan              string = "failed to close the plan"
 	errFailedToCollectExecExplainInfo string = "failed to collect execution explain information"
 	errSubTypeInit                    string = "sub-type initialization error at scan node reset"
+	errMissingFieldSelection          string = "missing field selection"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


shahzadlone (approved):
Thanks for squashing this bug!

LGTM

internal/planner/commit.go
```diff
@@ -446,8 +447,11 @@ func (n *dagScanNode) addSignatureFieldToDoc(link cidlink.Link, commit *core.Doc
 	if err != nil {
 		return err
 	}
-	sigFieldIndex := n.commitSelect.DocumentMapping.IndexesByName[request.SignatureFieldName][0]
-	sigMapping := n.commitSelect.DocumentMapping.ChildMappings[sigFieldIndex]
+	sigFieldIndexes, exists := n.commitSelect.DocumentMapping.IndexesByName[request.SignatureFieldName]
```
AndrewSisley, line 450: todo: This is dead code, and very clearly so.  If you don't remove it now, someone will likely do so the next time they have to touch this function. I have a preference for it to be removed now.

internal/planner/commit.go
```diff
@@ -380,7 +380,8 @@ func (n *dagScanNode) dagBlockToNodeDoc(block *coreblock.Block) (core.Doc, error
 		n.commitSelect.DocumentMapping.SetFirstOfName(&commit, request.DeltaFieldName, nil)
 	}
 
-	if block.Signature != nil {
+	if block.Signature != nil &&
```
AndrewSisley, line 383: todo: It looks like there might be another bug here, if the user requests the signature field and the block is not signed, instead of returning `"SignatureFieldName": nil` it will omit the field.

I think this `block.Signature != nil` should be moved to within the `IndexesByName[request.SignatureFieldName] != nil` check and `SignatureFieldName` with an explicit nil if `block.Signature == nil`.  After adding a test to cover this, assuming there isn't one already.

This is really a different bug, so if you don't want to deal with this now, can you please open a ticket and link to it from here?

AndrewSisley (approved):
Thanks for the quick fix! I do think the dead code should be removed before merge though.

internal/planner/commit.go
AndrewSisley: I created the issue, feel free to pick it up here if you want, but it is really a different bug https://github.com/sourcenetwork/defradb/issues/4005

internal/planner/commit.go
AndrewSisley: Fred pointed out that the `sigFieldIndexes` is used and my comment location is confusing - this todo is regarding the `exists` var, the `if` below, and the new error `NewErrMissingFieldSelection`.

jsimnz: had some convos on this point, and thought through it more. Im going to leave the code as is. Happy to bring it up and talk through on the DB sync. Theres another PR I want to open, and it depends on this, it is also semi-blocking Kyle, so prefer to merge it soon it rather later.

For posterity, my reasoning for keeping the code as is (beyond the logistics reasons mentioned above) boils down to 2 reasons. 1) The fix at the call site referenced here on this code comment is unsafe, due to the direct index lookup, so that needs to be fixed, all the other callsites that use `ChildMappings` handle this bug slightly differently but they do directly handle the bug so there isn't an option to panic on invalid indexes 2) The higher up code also just prevents this function from running entirely if there is no need to extract the signature values... this has the affect that the first point is moot and is "dead code" but I think they both add their own value to the fix/change.

getify:
💯💙

internal/planner/commit.go
jsimnz: From DB Sync call. Decided that its better to remove the dead code, and add a pre-condition note to the function documentation about needing to make sure the signature is actually requested.

PR #4009, opened 2025-09-12
refactor: Make net.Host interface easier to consume

### LINKED ISSUE #4007: Tweak net.Host interface to make it easier to declare/consume elsewhere
A trimmed version will be used by Lens, and they need to match before Defra integrates the latest Lens as the Lens feature branch wants to make Defra.net a test dependency.

AndrewSisley:
## Relevant issue(s)

Resolves #4007

## Description

Tweaks the `net.Host` interface easier to consume by removing elements that require direct reference to the exact concrete type.

This is required by Lens, so that a trimmed down `net.Host` interface can be consumed without making Lens dependant on Defra.  The alternative is to extract `net` to a new repository - this may be desirable long term, so Lens can spin up a `net.Host` instance, but even if we go that route I'd be in favour of this change as it allows external apps to consume the interface without a direct reference.

`net.Host.PeerInfo()` should now be the only interface member containing a a concrete custom type, I have left it, as this function is syntax sugar around `ID()` and `Addrs()` and is never going to be required by anything consuming the type.

The Lens code may be viewed in this PR: https://github.com/sourcenetwork/lens/pull/122 - see `host-go/p2p/host.go` for the re declaration of a subset of the `net.Host` interface.

### DIFF
--- a/client/p2p.go
+++ b/client/p2p.go
@@ -106,8 +106,8 @@ type Host interface {
 	PeerInfo() PeerInfo
 	// Pubkey return the byte slice representation of the host's public key.
 	Pubkey() ([]byte, error)
-	// Connect tries to connect to the peer with the given [PeerInfo].
-	Connect(ctx context.Context, info PeerInfo) error
+	// Connect tries to connect to the peer with the given peer info.
+	Connect(ctx context.Context, id string, addresses []string) error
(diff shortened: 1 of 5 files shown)


islamaliev (approved):
LGTM

PR #4012, opened 2025-09-12
fix: Passive sync with node chain

### LINKED ISSUE #2744: Network document updates do not send subscription events
Document updates from the p2p network do not send subscription events. Should be easy to fix by changing merge event types.

### LINKED ISSUE #4010: Document not syncing with chain of nodes on pubsub
This fails as the document never reaches node 2
```golang
func TestP2PCreateWithP2PCollection2(t *testing.T) {
    test := testUtils.TestCase{
        Actions: []any{
            &action.AddSchema{
                Schema: `
                    type Users {
                        Name: String
                        Age: Int
                    }
                `,
            },
            testUtils.ConnectPeers{
                SourceNodeID: 1,
                TargetNodeID: 0,
            },
            testUtils.ConnectPeers{
                SourceNodeID: 2,
                TargetNodeID: 1,
            },
            testUtils.SubscribeToCollection{
                NodeID:        1,
                CollectionIDs: []int{0},
            },
            testUtils.SubscribeToCollection{
                NodeID:        2,
                CollectionIDs: []int{0},
            },
            testUtils.CreateDoc{
                NodeID: immutable.Some(0),
                Doc: `{
                    "Name": "John",
                    "Age": 21
                }`,
            },
            testUtils.WaitForSync{},
            testUtils.Request{
                Request: `query {
                    Users {
                        Age
                    }
                }`,
                Results: map[string]any{
                    "Users": []map[string]any{
                        {
                            "Age": int64(21),
                        },
                    },
                },
            },
        },
    }

}
```

fredcarle:
## Relevant issue(s)

Resolves #4010 
Resolves #2744  

## Description

This PR ensures that every node in a chain of connected nodes receives a document update if they subscribe to the appropriate topic. 

The issue was caused by the nodes receiving the update message trying to fetch the new blocks, but the nodes that they were connected to didn't yet have it themselves. We solve this by sending the update message again once a node completes a sync.

By fixing this, we also fixed the gql subscription missing new doc updates when those updates came from the p2p system.

Additionally, the blockstore now as the capability to track if a block has been merged or not. This helps to prevent unnecessarily walking the dag upon receiving an update and trying to sync.

## Tasks

## How has this been tested?

New integration tests

- MacOS

### DIFF
--- a/internal/core/block/store.go
+++ b/internal/core/block/store.go
@@ -226,11 +226,21 @@ func updateHeads(
 		}
 	}
 
+	err := txn.Blockstore().MarkAsMerged(ctx, blockLink.Cid)
+	if err != nil {
+		return NewErrMarkingAsMerged(blockLink.Cid, err)
+	}
+
 	for _, l := range block.AllLinks() {
 		linkCid := l.Cid
 		isHead, err := headset.IsHead(ctx, linkCid)
 		if err != nil {
 			return NewErrCheckingHead(linkCid, err)
+		}
+
+		err = txn.Blockstore().MarkAsMerged(ctx, blockLink.Cid)
+		if err != nil {
+			return NewErrMarkingAsMerged(blockLink.Cid, err)
 		}
 
 		if isHead {
--- a/internal/core/block/store.go
+++ b/internal/core/block/store.go
@@ -238,7 +238,7 @@ func updateHeads(
 			return NewErrCheckingHead(linkCid, err)
 		}
 
-		err = txn.Blockstore().MarkAsMerged(ctx, blockLink.Cid)
+		err = txn.Blockstore().MarkAsMerged(ctx, linkCid)
 		if err != nil {
 			return NewErrMarkingAsMerged(blockLink.Cid, err)
 		}
--- a/internal/core/block/store.go
+++ b/internal/core/block/store.go
@@ -226,6 +226,8 @@ func updateHeads(
 		}
 	}
 
+	// Marking the block as merged removes the to-merge index. It signals that nothing
+	// else needs to be done for that block.
 	err := txn.Blockstore().MarkAsMerged(ctx, blockLink.Cid)
 	if err != nil {
 		return NewErrMarkingAsMerged(blockLink.Cid, err)
@@ -238,6 +240,8 @@ func updateHeads(
 			return NewErrCheckingHead(linkCid, err)
 		}
 
+		// Marking the block as merged removes the to-merge index. It signals that nothing
+		// else needs to be done for that block.
 		err = txn.Blockstore().MarkAsMerged(ctx, linkCid)
 		if err != nil {
 			return NewErrMarkingAsMerged(blockLink.Cid, err)
--- a/internal/db/config.go
+++ b/internal/db/config.go
@@ -30,6 +30,8 @@ type dbOptions struct {
 	disableSigning bool
 	p2p            immutable.Option[client.Host]
 	retryIntervals []time.Duration
+	// timeout duration for syncing block links.
+	p2pBlockSyncTimeout time.Duration
 }
 
 func defaultDBOptions() *dbOptions {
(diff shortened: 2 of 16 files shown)


internal/db/p2p/p2p.go
```diff
@@ -368,12 +378,33 @@ func (p *P2P) processPushlogRequest(
...
+	}
+	if merged {
+		return nil
+	}
+
+	// Calls to syncDAG should not overlap for a given CID. If they do, they will use the same
+	// underlying pubsub topic and this brings along potential pitfalls. One of them being that
+	// if this initial sync call had a negative response for a given link, the subsequent calls will
+	// assume a negative response for that same link without retrying.
+	p.processQueue.add(headCID.String())
```
AndrewSisley: suggestion: Does this need to call `String()`?  Can we not use the raw cid (either bytes or `KeyString()`)?

internal/db/p2p/p2p.go
```diff
@@ -431,3 +476,44 @@ func (p *P2P) SendUpdate(evt event.Update) error {
 
 	return nil
 }
+
+// processQueue is synchronization source to ensure that concurrent
+// document merges do not cause transaction conflicts.
+type processQueue struct {
```
AndrewSisley, line 487: praise: This is quite (b)locky, but a neat and code-efficient way of handling the problem - thanks :)

tests/integration/net/simple/peer/with_create_test.go
```diff
@@ -195,3 +195,131 @@ func TestP2PCreateWithP2PCollection(t *testing.T) {
...
+			},
+			testUtils.ConnectPeers{
+				SourceNodeID: 2,
+				TargetNodeID: 1,
+			},
+			testUtils.ConnectPeers{
+				SourceNodeID: 3,
+				TargetNodeID: 2,
+			},
+			testUtils.ConnectPeers{
```
AndrewSisley, line 230: suggestion: It might be worth noting the importance of having 4 somewhere in this test. I almost always assume that 3 and 4 would have the same problems, and we might lose a node from this test (and any future derivatives) over time after forgetting the bug you found at 4.

AndrewSisley (approved):
LGTM, approving now but might want to review any blockstore/index rework that may or may not happen before merge.  Thanks for this Fred :)

internal/core/block/store.go
```diff
@@ -226,13 +226,23 @@ func updateHeads(
 		}
 	}
 
+	err := txn.Blockstore().MarkAsMerged(ctx, blockLink.Cid)
+	if err != nil {
```
shahzadlone, line 232: Nitpick: one liner comment here saying why its important to mark this as merged right here is important would be nice.

internal/db/config.go
```diff
@@ -45,6 +47,7 @@ func defaultDBOptions() *dbOptions {
 			time.Minute * 16,
 			time.Minute * 32,
 		},
+		p2pBlockSyncTimeout: time.Second * 5,
```
shahzadlone, line 50: Nitpick: would be nice to see all these time defaults defined at the top perhaps

shahzadlone (approved):
LGTM, just some nitpicks

nasdf:
@fredcarle does this fix #2744 as well?

fredcarle:
> @fredcarle does this fix #2744 as well?

It does :)

internal/db/config.go
fredcarle: I prefer not separating the default declaration from the defined values. Specially if they are not used anywhere else.

internal/db/p2p/p2p.go
fredcarle: Done

tests/integration/net/simple/peer/with_create_test.go
fredcarle: Done

internal/core/block/store.go
fredcarle: Done

AndrewSisley (approved):
LGTM, thanks Fred!

PR #4019, opened 2025-09-16
feat: Support Commit API on GQL Subscriptions

### LINKED ISSUE #3986: Improve error handling for invalid subscriptions received over SSE
Currently, if you make an SSE subscription attempt to a DefraDB node, but use a subscription that's malformed (or just not supported, such as `subscription { Commit { .. } }` (`Commit` subscriptions not currently supported!), this causes DefraDB to skip the subscription and return a `200` with a GraphQL-style errors payload like:

```json
{
  "errors": [
    {
      "message": "Cannot query field \"commits\" on type \"Subscription\"."
    }
  ],
  "data": null
}
```

This is mostly because the SSE and GraphQL endpoints are overloaded on the same URL path.

However, this type of error reporting is [not according to the SSE standard](https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md). Currently, the GraphQL-style errors body is likely just ignored/swallowed by SSE clients (like [Node.js GraphQL-SSE](https://the-guild.dev/graphql/sse)).

Per @jsimnz, instead, the DefraDB node should establish/upgrade the SSE stream, and report the error(s) via a "next" event (and then kill the stream, since it's fatal). That's how SSE client libraries will expect to be notified, and thus allow developers to detect an error.

Per @fredcarle, DefraDB can detect that the request is for SSE by inspecting the `Accept` header for `text/event-stream`.

@jsimnz suggested this simplified pseudo-code for the fix:

```
if request.headers.accept == "text/event-stream" {
  upgradeToSSE()
  result := db.ExecRequest(request.body)
  if result.errors != nil {
    respondSSE(result)
  }

  for data := range result.subscription {
    respondSSE(data)
  }
}

result := db.ExecRequest(request.body)
respond(result)
```

### LINKED ISSUE #4017: Infinite loop on bad commit `fieldName` filter
**Describe the problem**
If you create a `commits` query with a `fieldName` argument that either
A) Doesn't exist
B) If used in conjunction with a `cid` argument that doesn't match the `fieldName`
C) Other?

Then it creates an infinite loop of `Next()` calls on the `dagScanNode`

**To Reproduce**
```graphql
query {
  commits(fieldName: "BAD_NAME") {
    cid
  }
}
```

### LINKED ISSUE #4018: Support `commits` API on subscription
Add the ability to subscribe to `commits` stream.

```graphql
subscription {
  commits {
    cid
    docID
  }
}
```

jsimnz:
## Relevant issue(s)

Resolves #4018 
Resolves #3986
Resolves #4017

## Description

Primarily adds support for the commit API on GQL subscriptions. Additionally it fixes a bug in how we were reporting certain kind of GQL validation errors to SSE clients. Lastly, it fixes a bug in the `dagScanNode` planner that resulted in an infinite loop if you requested a bad field on a `commits` query. 

Additionally, not marked as a "resolved" issue as I felt it wasn't necessary, fixed the "Dead Code" issue referenced [here](https://github.com/sourcenetwork/defradb/pull/4004#discussion_r2351488606).

For the actual implementation, I made the conscious choice to only emit a ***single*** event for a mutation operation (create/update) instead of one for each individual commit block that is generated (which is N+1, where N is the number of fields mutated). This simplifies and streamlines the commit subscription, and you can use the `links` in the `_version` field to access the corresponding commit blocks anyway. It does however limit the ability to subscribe to an individual fields updates, and instead can only subscribe to the document.

Another important note that I would like feedback on... For the Subscription schema, I am reusing the existing `commits` type from the Query schema (this is also how the user-defined types work as well, single type on both the Query and Subscription). This leads to an odd behavior in both the commits API and the collection subscription... There are some arguments that don't do anything in *some* contexts. Most notable, as referenced above we can't subscribe on individual fields, but the `fieldName` argument is still there, its just silently ignored. I would like to deal with this out of scope, as some solution might also affect how the collection subscription stuff works.

Lastly, in the `handler_store.go:execSSESubscription()` function theres some errors that are silently ignored, should we add some error logging here? There wasn't any before, and my refactor there was a "plain text refactor".

PS. Open to more suggestions for tests that can be added to the commit subscription suite.

## Tasks

## How has this been tested?
Manually, updated integration suite.

- Ubuntu (WSL2)


internal/request/graphql/schema/schema.go
```diff
@@ -77,15 +66,23 @@ func defaultMutationType() *gql.Object {
...
+func defaultSubscriptionType(fields ...*gql.Field) *gql.Object {
+	return defaultOperationType("Subscription", fields...)
+}
+
+func defaultOperationType(name string, fields ...*gql.Field) *gql.Object {
+	fieldsCfg := make(gql.Fields, len(fields)+1)
+	for _, field := range fields {
+		fieldsCfg[field.Name] = field
+	}
+	fieldsCfg["_"] = &gql.Field{
```
jsimnz: question(all): We don't actually need these `_` query placeholders anymore, they are a very old artifact of how the GQL library works, but its no longer necessary, for either Subscription or Query.

tests/integration/query/commits/branchables/cid_doc_id_test.go
```diff
@@ -49,7 +49,7 @@ func TestQueryCommitsBranchables_WithCidAndDocIDParam(t *testing.T) {
 				Results: map[string]any{
 					"commits": []map[string]any{},
 				},
-				ExpectedError: "cid does not belong to document",
+				// ExpectedError: "cid does not belong to document",
```
jsimnz: question(all): To make the commit subscription work properly with `docID` filters, this behavior needed to change. Happy to discuss, but I don't think theres much we can do here. It also feels appropriate to simply not return any results if the `cid` filter doesn't match the `docID` filter, instead of an error.

http/handler_store.go
```diff
@@ -240,61 +242,80 @@ type GraphQLRequest struct {
...
 
-	case req.Body != nil:
-		if err := requestJSON(req, &request); err != nil {
-			responseJSON(rw, http.StatusBadRequest, errorResponse{err})
-			return
-		}
-	default:
-		responseJSON(rw, http.StatusBadRequest, errorResponse{ErrMissingRequest})
+	// if at this point the we get a subscription query, it isn't using
+	// the correct accpet headers, and we error
```
islamaliev: typo: "accpet"

http/handler_store.go
```diff
@@ -326,19 +339,70 @@ func (h *storeHandler) ExecRequest(rw http.ResponseWriter, req *http.Request) {
...
+
+		request.Query = req.URL.Query().Get("query")
+
+		request.OperationName = req.URL.Query().Get("operationName")
+
+		variablesFromQuery := req.URL.Query().Get("variables")
+		if variablesFromQuery != "" {
+			var variables map[string]any
+			if err := json.Unmarshal([]byte(variablesFromQuery), &variables); err != nil {
+				// responseJSON(rw, http.StatusBadRequest, errorResponse{err})
```
islamaliev: todo: remove commented out code

internal/planner/commit.go
```diff
@@ -267,7 +275,7 @@ func (n *dagScanNode) Next() (bool, error) {
 		len(n.visitedNodes) == 0 &&
 		n.commitSelect.DocID.HasValue() &&
 		currentDocID != n.commitSelect.DocID.Value() {
-		return false, ErrIncorrectCIDForDocId
+		return false, nil // ErrIncorrectCIDForDocId
```
islamaliev: todo: remove commented out code

internal/planner/commit.go
```diff
@@ -236,6 +237,13 @@ func (n *dagScanNode) Next() (bool, error) {
 	}
 
 	if n.commitSelect.FieldName.HasValue() {
+		// early catch for CID based filtering
+		// since we are only concerned about this one CID lookup
+		if n.commitSelect.Cid.HasValue() &&
+			n.commitSelect.FieldName.Value() != dagBlock.Delta.GetFieldName() {
+			return false, nil
+		}
```
islamaliev, line 245: question: the same check `n.commitSelect.FieldName.Value() != dagBlock.Delta.GetFieldName()` is also done 10 lines below, but it will call `return n.Next()`. Is it intentional? Why different handling in presence of CID?

internal/request/graphql/parser/subscription.go
```diff
@@ -42,7 +42,14 @@ func parseSubscriptionOperationDefinition(
 // parseSubscription parses a typed subscription field
 // which includes sub fields, and may include
 // filters, IDs, etc.
-func parseSubscription(exe *gql.ExecutionContext, field *ast.Field) (*request.ObjectSubscription, error) {
+func parseSubscription(exe *gql.ExecutionContext, field *ast.Field) (request.Selection, error) {
+	if field.Name.Value == "commits" {
```
islamaliev, line 46: nitpick: I think we should have somewhere a const for this value. Would be better to use it

internal/request/graphql/schema/schema.go
```diff
@@ -38,31 +40,18 @@ func defaultSchema() (gql.Schema, error) {
...
 		Mutation:     defaultMutationType(),
 		Directives:   defaultDirectivesType(crdtEnum, explainEnum, orderEnum, indexFieldInput),
-		Subscription: defaultSubscriptionType(),
+		Subscription: defaultSubscriptionType(queryCommits),
 	})
+
+	return sch, err
 }
 
 // @todo: Use a better default Query type
```
islamaliev: question: is this todo still up-to-date?

tests/integration/query/commits/with_field_test.go
```diff
@@ -139,3 +139,88 @@ func TestQueryCommitsWithCompositeFieldIdWithReturnedSchemaVersionID(t *testing.
...
+					},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestQueryCommitsWithWrongFieldAndCID(t *testing.T) {
```
islamaliev: suggestion: the tests' names don't follow our agreed-upon style `<prefix>_<condition>_<expectation>`
This one for example can be named `TestQueryCommits_WithWrongFieldAndCID_ReturnEmptyList`

tests/integration/query/commits/with_field_test.go
```diff
@@ -139,3 +139,88 @@ func TestQueryCommitsWithCompositeFieldIdWithReturnedSchemaVersionID(t *testing.
...
+			},
+			testUtils.Request{
+				Request: `query {
+						commits (fieldName: "NOT_A_FIELD", cid: "bafyreiae763hq5srsefplqrehpsuyieuwmbvblgzdma7srss522yciumhu") {
+							cid
+						}
+					}`,
+				Results: map[string]any{
+					"commits": []map[string]any{},
+				},
```
islamaliev, line 220: suggestion: I squint a bit when I see this. It's might be okay to show empty result if there is no document that matches given field and cid, but I would say for explicitly invalid input we should show an error.

tests/integration/subscription/with_commit_test.go
```diff
@@ -0,0 +1,244 @@
...
+							{
+								"cid": "bafyreihsducixg7n6wdbqoyjao4pecsalt4zjx2ybyncizqhq2ci46gyoa",
+							},
+						},
+					},
+				},
+			},
+			testUtils.Request{
+				Request: `mutation {
+					create_User(input: {name: "John", age: 27, points: 42.1, verified: true}) {
```
islamaliev: question: why didn't you use `testUtils.CreateDoc`?

tests/integration/subscription/with_commit_test.go
```diff
@@ -0,0 +1,244 @@
...
+				},
+			},
+		},
+	}
+
+	execute(t, test)
+}
+
+func TestCommitSubscriptionWithDocFilterAndMultipleMutations(t *testing.T) {
+	// create1Links := testUtils.NewSameValue()
```
islamaliev: todo: remove commented out code

tests/integration/subscription/with_commit_test.go
```diff
@@ -0,0 +1,244 @@
...
+	execute(t, test)
+}
+
+func TestCommitSubscriptionLinksWithCreateMutations(t *testing.T) {
+	create1Links := testUtils.NewSameValue()
+	create2Links := testUtils.NewSameValue()
+
+	test := testUtils.TestCase{
+		Actions: []any{
+			testUtils.SubscriptionRequest{
```
islamaliev, line 75: thought: I wonder if we want and/or makes sense to have a test action with long-running subscription, so that instead of testing immediate result we can test updates on subscription.

The test framework can close subscription after all test actions are executed and assert that all expected docs are received.

fredcarle: Isn't that exactly what this current Subscription test action is doing?

AndrewSisley: yes

http/handler_store.go
```diff
@@ -326,19 +339,70 @@ func (h *storeHandler) ExecRequest(rw http.ResponseWriter, req *http.Request) {
...
+	}
+	_, err = fmt.Fprintf(rw, "data: %s\n\n", data)
+	if err != nil {
+		return err
+	}
+	flusher.Flush()
+	return nil
+}
+
+func extractGraphQLRequest(rw http.ResponseWriter, req *http.Request) (GraphQLRequest, []client.RequestOption, error) {
```
fredcarle: todo: We should remove the response writer param as it's not needed in the function.

http/handler_store.go
```diff
@@ -240,61 +242,80 @@ type GraphQLRequest struct {
...
+			return
+		}
+
+		err = emitSSENextEvent(rw, flusher, string(data))
+		if err != nil {
+			return
+		}
+
+		_ = emitSSECompleteEvent(rw, flusher)
+		return
```
fredcarle, line 319: suggestion: The marshalling error shouldn't be silently ignored. I think it should be returned to the caller. It should never happen because the gql result is controlled by us but if for some odd reason it did, the caller should know about it. 

As for the other two ignored errors in this block, I think it's correct to ignore them since an error there would have to be related to the connection/responseWriter which we can't really do anything about. Same for the ones in the select block.

http/handler_store.go
fredcarle, line 1: todo: The CI is failing for the HTTP and CLI clients because the Accept header is not set. This means the subscription code in the graphql handler is never tiggered. Please add the header in the `client.ExecRequest` method after `c.http.setDefaultHeaders(req)`.
```go
// This works but you might find a better way to do this.
if strings.HasPrefix(strings.ToLower(strings.TrimSpace(query)), "subscription") {
	req.Header.Set("Accept", sseAcceptHeader)
}
```

tests/integration/query/commits/branchables/cid_doc_id_test.go
fredcarle: This is a degradation of DX in my opinion. We should probably explore a better solution.

internal/request/graphql/schema/schema.go
fredcarle: Might as well remove it then.

tests/integration/subscription/with_commit_test.go
islamaliev: I didn't check the implementation. Just assumed it didn't. Good to know.

fredcarle (dismissed):
I've changed my mind on limiting the comits subscription to only the composite. The reason is that we have a fieldName filter for this and if users want all commits to be returned then we shouldn't artificially prevent it.

http/handler_store.go
jsimnz: Without fully parsing the entire GraphQL query (or at least part of it if that is an option), there isn't a reliable way to determine if the operation type is a subscription.

Examples:
```graphql
# Hello World
subscription
{
    User...
}
```

This is a valid potential query, the comment at the beginning and the brackets on a new line make this very hard to determine operation type with basic string operations.

But fully parsing the entire string feels needlessly expensive :/. Ive added a naive solution but it will fail in the above example.

http/handler_store.go
jsimnz: > The marshalling error shouldn't be silently ignored. I think it should be returned to the caller.

This is a cyclic problem though, no? The marshal code is for the errors, if there is error when marshaling the error, we need to create another error, and marshal it to a string to send it to the user, which might have an error...

Unless I've misunderstood what your suggesting. Alternatively we could craft a JSON value by hand to return to the user without marshaling.

internal/planner/commit.go
jsimnz: If there is a user defined CID, then the `dagScanNode` is only ever concerned with the one block, which we then check if there is a user defined field name and early exit if it doesn't match.

The check 10 lines below is checking if *this particular loop of the planNode* and its CID have a matching field, if it doesn't it continues on trying to find the next potential CID with the matching field.

internal/request/graphql/parser/subscription.go
jsimnz: I think theres a bunch of places the parser code can use consts (take a look at `graphql/schema/types/commits.go`). I'd rather do that all together seperately.

internal/request/graphql/schema/schema.go
jsimnz: Nope :)

tests/integration/query/commits/branchables/cid_doc_id_test.go
jsimnz: I don't believe this is a degredation of DX, but actually adhering to what the query is trying todo. But, this entirely depends on how we (and our users) treat the non `filter` args that act as another filter. 

Eg should `User(docID: "...")` be semantically (external not internal) different from `User(filter: {docID: { _eq: "..."}})` other than being more convenient/syntactic surgar?

If the answer is Yes, they are different, then you are right and this change should be addressed.
If the answer is No, they are the same, then this new behavior is more accurate.

tests/integration/query/commits/with_field_test.go
jsimnz: Ahh, great catch! This was originally copied from a relatively old test that uses this older format. Theres an uncountable number of those, but new test should certainly use the agreed upon format.

tests/integration/query/commits/with_field_test.go
jsimnz: I agree, it should be part of the input validation. There is a *lot* of places we need semantic validation on top of the GQL syntactic validation. 

This particular PR doesn't change this behavior, just tests/documents it for the first time. Therefore I'd prefer to fix this in a followup PR.

tests/integration/subscription/with_commit_test.go
jsimnz: No particular reason, was writting a bunch of these all over, and frequently copying/pasting from the playground as I manually tested. Is this a nitpick or a suggestion to change it?

http/client.go
```diff
@@ -337,19 +337,23 @@ func (c *Client) ExecRequest(
...
+	err = c.http.setDefaultHeaders(req)
 	if err != nil {
 		result.GQL.Errors = append(result.GQL.Errors, err)
 		return result
 	}
 
+	if strings.Contains(query, "subscription {") {
+		req.Header.Set("Accept", sseAcceptHeader)
+	}
+
```
fredcarle: suggestion: if you want it to be more robust without reading the whole thing.
```go
iter := strings.Lines(query)
for line := range iter {
	trimmedLine := strings.TrimSpace(line)
	if strings.HasPrefix(trimmedLine, "#") {
		// we skip the comment lines
		continue
	}
	if strings.HasPrefix(trimmedLine, "subscription") {
		req.Header.Set("Accept", sseAcceptHeader)
		break
	}
}
```

jsimnz: The annoying part is that GQL supports "Description" strings that precede operations. So need to apply a quick scan to handle a single line and multi line block string description 🥲

getify: could this provide a false match if the "subscription {" shows up in a comment or description string?

AndrewSisley: We got rid of this hack from planner, and I don't think performance is too great a concern here.  We can remove this re-introduction of the hack by using the gql ast lib (will probably be more expensive computationally).

It will be slightly more dev-effort to do so, so if John is really pushed for time and if this PR is urgent, we can probably clean it up later, before v1.

getify: it seems exactly the sort of corner-cutting that we could do now and very likely get bitten by later. it might be worth that to move forward, but... I suspect that we'll be back here at some point.

I know I *often* comment things out -- though maybe not inside of graphql snippets, but... possibly!? -- and expect they get ignored while I'm testing other things, so I can certainly see myself having accidentally commented out a `subscribe{ .. }` that wasn't working and trying some other `query { .. }` temporarily. ;-)

jsimnz: Ive opted to write a very quick linear scan 0 allocation GQL parser (it's a stretch to call it a parser tbh) that adheres to the full gql spec, and pulls out the operation name, accounting for edge cases.

I'll push this up momentarily. The original solution was always temp to this PR, I just didn't know how deep I wanted/needed to go for this.

jsimnz: > Never spend millions of dollars and years of development making an advanced pen that can write in space, when you could've used a pencil.

:sweat_smile:
After attempting to get my "parser" to accept all my unit tests AND not be a mess of hard to read spaghetti... I opted for a simple solution. User provided context.

Happy to discuss, but it works, its easy, and its replaceable in the future.

tests/integration/subscription/with_commit_test.go
islamaliev: it's a question :) but now that you asked I would be leaning toward suggestion. The reason is that we usually use `CreateDoc` and using mutation request explicitly makes the reader feel that there is something special about GQL requests which is incorrect.

http/client.go
getify: for my learning processes, can you elaborate a bit as to what "user provided context" means here?

tests/integration/query/commits/branchables/cid_doc_id_test.go
jsimnz: Follow up from team discussion: I will keep the error as is, and on the subscription handler, we'll do a error introspection and filter out this kind of error so it doesn't bleed into the user subscription unintentionally.

http/client.go
jsimnz: > for my learning processes, can you elaborate a bit as to what "user provided context" means here?

The user provides a bool that indicates if its a subscription or not, using the `GQLOptions` parameter.

But, this is moot. We discussed on the DB call, and decided that we can just send both accept headers `json` and `event-stream`, then leave it up to the server to upgrade the connection or not. 

In a follow up PR, we will add a dedicated `ExecSubscription` method to the client API to properly segment/inform the semantic differences between a subscription vs non-subscription request.

jsimnz:
> I've changed my mind on limiting the comits subscription to only the composite. The reason is that we have a fieldName filter for this and if users want all commits to be returned then we shouldn't artificially prevent it.

Follow up from DB call. We'll leave it as is in this PR (only one composite event per doc mutation), and discuss with the team on ways to rework the `commits` API to accept more nuanced combination of docIDs, CIDs, and fieldNames.

http/client.go
jsimnz: Changed my mind *again*. To gurantee correctness and "simplicty" the client is going to parse the query and determine the operation type.

I added some benchmarks to determine the perf implications, and its negligable.

http/handler_store.go
fredcarle: Fair point. I wonder if there is some merit to returning a 500 status in that case. Maybe a conversation for the DX team.

tests/integration/subscription/subscription_test.go
```diff
@@ -15,10 +15,14 @@ import (
 
 	"github.com/sourcenetwork/defradb/tests/action"
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+	"github.com/sourcenetwork/immutable"
 )
 
 func TestSubscriptionWithCreateMutations(t *testing.T) {
 	test := testUtils.TestCase{
+		SupportedClientTypes: immutable.Some(
```
fredcarle: question: Any reason why you're limiting the supported client types? It seems to works fine locally with the go client.

fredcarle (commented):
LGTM. Just one question that might require a change.

tests/integration/subscription/subscription_test.go
jsimnz: Great catch, this is a debug artifact from early in the PR. Forgot about it, thanks

client/request/commit.go
```diff
@@ -60,3 +62,20 @@ func (c CommitSelect) ToSelect() *Select {
 		ChildSelect: c.ChildSelect,
 	}
 }
+
+// Implements the subscriptionSelector interface in internal/db/subscriptions.go
+// We can safely ignore the first parameter (docID) for now
+// since its always copied from the original subscription request
+func (c CommitSelect) ToSubscriptionSelect(_, cid string) Selection {
```
shahzadlone: nitpick: I want to enable a linter that catches typos and wrong (non-matching) function names as first word somewhere in the nesr future, so in that spirit leaving this nitpick

```suggestion
// ToSubscriptionSelect implements the subscriptionSelector interface in internal/db/subscriptions.go
// We can safely ignore the first parameter (docID) for now
// since its always copied from the original subscription request
func (c CommitSelect) ToSubscriptionSelect(_, cid string) Selection {
```

internal/planner/commit.go
```diff
@@ -435,6 +443,8 @@ func (n *dagScanNode) dagBlockToNodeDoc(block *coreblock.Block) (core.Doc, error
 	return commit, nil
 }
 
+// pre-condition: the signature needs to have been requested in the query selection
```
shahzadlone, line 448: Nitpick: similar to other comment

tests/integration/mutation/delete/simple_test.go
```diff
@@ -55,7 +55,7 @@ func TestMutationDeletion_WithoutSubSelectionFields(t *testing.T) {
 
 					}
 				}`,
-				ExpectedError: "Syntax Error GraphQL request (2:17) Unexpected empty IN {}",
```
shahzadlone, line 58: Question: What happened here?

tests/integration/net/simple/peer/with_create_test.go
```diff
@@ -314,7 +314,7 @@ func TestP2PCreate_WithP2PCollectionAndSubscription_ShouldSucceed(t *testing.T)
 				},
 			},
 			testUtils.CreateDoc{
-				NodeID: immutable.Some(0),
+				NodeID: immutable.Some(1),
```
shahzadlone: Question: what was the reason for this change?

shahzadlone (approved):
Just some non-blocking nitpicks and questions.

LGTM

tests/integration/mutation/delete/simple_test.go
jsimnz: It was behaving weird, if I left the `request (2:17)` in it would fail on some tests matrix but pass others. Basically, some of the tests were including the `request` substring, and others weren't. I dont know why its acting up, but it doesn't semantically change anything. The `(2:17) Unexpected empty IN {}` is still there though.

tests/integration/net/simple/peer/with_create_test.go
jsimnz: Debug artifact, reverted

PR #4023, opened 2025-09-17
fix: Allow configuring node identity with embedded Defra

### LINKED ISSUE #4022: Can't set node identity when using embedded DefraDB
This came up when a team member tried to use embedded Defra and they noticed that the document block didn't contain any signature. The reason for this is that, even though signing is set to `true` by default, the node identity has to be configured to be able to produce the signature. 

When looking into configuring the node identity, they noticed that the only way to do so programatically is to call `db.WithNodeIdentity` which is an internal function that can't be used by `defradb` consumers.

fredcarle:
## Relevant issue(s)

Resolves #4022

## Description

This PR simply exposes a node config to set the internal db node identity config.

Our test framework couldn't catch this since it has access to the internal package.

We had discussed making DB public again. I'll be looking into doing this in order to prevent these kinds of problems in the future.

## Tasks

## How has this been tested?

manual embedded defra test

- MacOS

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -11,6 +11,8 @@
 package node
 
 import (
+	"github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/internal/db"
 	"github.com/sourcenetwork/defradb/internal/kms"
 
 	"github.com/sourcenetwork/immutable"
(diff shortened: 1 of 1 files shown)


node/config.go
```diff
@@ -70,6 +72,12 @@ func WithEnableDevelopment(enable bool) NodeOpt {
 	}
 }
 
+// WithNodeIdentity sets the identity for the node. This is the identity that
+// will be used for things like block signatures and P2P sync operations.
+func WithNodeIdentity(ident identity.Identity) db.Option {
```
AndrewSisley, line 77: suggestion: If it is simple to do, can you please use this in the test framework instead of the internal call you mentioned in the PR description?

AndrewSisley (approved):
LGTM

node/config.go
fredcarle: Since we plan on making `db` public, I'd rather not change it now only to change it again later.

PR #4025, opened 2025-09-17
chore: Reduce size of store prefixes

### LINKED ISSUE #4024: Use full cid in blockstore keys
The blockstore keys currently use only the hash of the cid. This means that future versions of defra would have no information allowing them to recompute the cid of a given block for cid verification.

We should instead store the full cid (version+codec+hash).

### LINKED ISSUE #4026: Reduce size of store prefixes
We currently have human readable store prefixes and that is not needed internally. We can add a function to make them human readable when needed.

fredcarle:
## Relevant issue(s)

Resolves #4026 

## Description

~~This PR changes the blockstore keys to use the full cid information (version+codec+hash instead of only hash). It also uses raw bytes instead of a base32 encoded string which reduces storage overhead.~~

This PR no only focuses on reducing the size of the store prefixes. The blockstore portion won't be needed shortly as Andy is moving it outside of Defra. This PR will wait for that change to happen before getting merged.

Using raw bytes has the drawback of being difficult to use for debugging. A `HumanReadableKey` function was added and used with `printStore`.

## Tasks

## How has this been tested?

make test

- MacOS

### DIFF
--- a/internal/datastore/blockstore.go
+++ b/internal/datastore/blockstore.go
@@ -46,11 +46,15 @@ func (bs *bstore) AsIPLDStorage() IPLDStorage {
 
 const (
 	objectMarker       = byte(0xff)
-	toMergeIndexPrefix = "m"
+	toMergeIndexPrefix = byte('m')
 )
 
-func newToMergeKey(cid string) []byte {
-	return []byte(toMergeIndexPrefix + cid)
+func newToMergeKey(cid []byte) []byte {
+	l := len(cid)
+	key := make([]byte, l+1)
+	copy(key[1:], cid)
+	key[0] = toMergeIndexPrefix
+	return key
 }
 
 func (bs *bstore) IsMerged(ctx context.Context, cid cid.Cid) (bool, error) {
@@ -61,7 +65,7 @@ func (bs *bstore) IsMerged(ctx context.Context, cid cid.Cid) (bool, error) {
 	if !hasBlock {
 		return false, nil
 	}
-	notMerged, err := bs.store.Has(ctx, newToMergeKey(cid.String()))
+	notMerged, err := bs.store.Has(ctx, newToMergeKey(cid.Bytes()))
 	if err != nil {
 		return false, err
 	}
@@ -69,7 +73,7 @@ func (bs *bstore) IsMerged(ctx context.Context, cid cid.Cid) (bool, error) {
 }
 
 func (bs *bstore) MarkAsMerged(ctx context.Context, cid cid.Cid) error {
-	return bs.store.Delete(ctx, newToMergeKey(cid.String()))
+	return bs.store.Delete(ctx, newToMergeKey(cid.Bytes()))
 }
 
 type p2pBlockStore struct {
@@ -85,7 +89,7 @@ func (bs *p2pBlockStore) Put(ctx context.Context, block blocks.Block) error {
 	if err == nil && exists {
 		return nil // already stored.
 	}
-	err = bs.store.Set(ctx, newToMergeKey(string(key)), []byte{objectMarker})
+	err = bs.store.Set(ctx, newToMergeKey(key), []byte{objectMarker})
 	if err != nil {
 		return err
 	}
@@ -99,7 +103,7 @@ func (bs *p2pBlockStore) PutMany(ctx context.Context, blocks []blocks.Block) err
 		if err == nil && exists {
 			continue
 		}
-		err = bs.store.Set(ctx, newToMergeKey(string(key)), []byte{objectMarker})
+		err = bs.store.Set(ctx, newToMergeKey(key), []byte{objectMarker})
 		if err != nil {
 			return err
 		}
--- a/internal/datastore/blockstore.go
+++ b/internal/datastore/blockstore.go
@@ -46,11 +46,11 @@ func (bs *bstore) AsIPLDStorage() IPLDStorage {
 
 const (
 	objectMarker       = byte(0xff)
-	toMergeIndexPrefix = "/tm"
+	toMergeIndexPrefix = "m"
 )
 
 func newToMergeKey(cid string) []byte {
-	return []byte(toMergeIndexPrefix + "/" + cid)
+	return []byte(toMergeIndexPrefix + cid)
 }
 
 func (bs *bstore) IsMerged(ctx context.Context, cid cid.Cid) (bool, error) {
@@ -85,7 +85,7 @@ func (bs *p2pBlockStore) Put(ctx context.Context, block blocks.Block) error {
 	if err == nil && exists {
 		return nil // already stored.
 	}
-	err = bs.store.Set(ctx, newToMergeKey(block.Cid().String()), []byte{objectMarker})
+	err = bs.store.Set(ctx, newToMergeKey(string(key)), []byte{objectMarker})
 	if err != nil {
 		return err
 	}
@@ -99,7 +99,7 @@ func (bs *p2pBlockStore) PutMany(ctx context.Context, blocks []blocks.Block) err
 		if err == nil && exists {
 			continue
 		}
-		err = bs.store.Set(ctx, newToMergeKey(b.Cid().String()), []byte{objectMarker})
+		err = bs.store.Set(ctx, newToMergeKey(string(key)), []byte{objectMarker})
 		if err != nil {
 			return err
 		}
--- a/internal/datastore/blockstore.go
+++ b/internal/datastore/blockstore.go
@@ -89,7 +89,7 @@ func (bs *p2pBlockStore) Put(ctx context.Context, block blocks.Block) error {
 	if err == nil && exists {
 		return nil // already stored.
 	}
-	err = bs.store.Set(ctx, newToMergeKey(key), []byte{objectMarker})
+	err = bs.store.Set(ctx, newToMergeKey(block.Cid().Bytes()), []byte{objectMarker})
 	if err != nil {
 		return err
 	}
@@ -103,7 +103,7 @@ func (bs *p2pBlockStore) PutMany(ctx context.Context, blocks []blocks.Block) err
 		if err == nil && exists {
 			continue
 		}
-		err = bs.store.Set(ctx, newToMergeKey(key), []byte{objectMarker})
+		err = bs.store.Set(ctx, newToMergeKey(b.Cid().Bytes()), []byte{objectMarker})
 		if err != nil {
 			return err
 		}
--- a/internal/datastore/multi_test.go
+++ b/internal/datastore/multi_test.go
@@ -1,0 +1,96 @@
+// Copyright 20225Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package datastore
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corekv/memory"
+)
+
+func TestMultistore_HumanReadableKeys_ShouldSucceed(t *testing.T) {
+	ctx := context.Background()
+	rootstore := memory.NewDatastore(ctx)
+
+	ms := NewMultistore(rootstore)
+
+	err := ms.Blockstore().Put(ctx, blocks.NewBlock([]byte("123")))
+	require.NoError(t, err)
+	err = P2PBlockstoreFrom(rootstore).Put(ctx, blocks.NewBlock([]byte("1234")))
+	require.NoError(t, err)
+	err = ms.Datastore().Set(ctx, []byte("/123"), []byte("123"))
+	require.NoError(t, err)
+	err = ms.Encstore().Put(ctx, blocks.NewBlock([]byte("123")))
+	require.NoError(t, err)
+	err = ms.Headstore().Set(ctx, []byte("/123"), []byte("123"))
+	require.NoError(t, err)
+	err = ms.Peerstore().Set(ctx, []byte("/123"), []byte("123"))
+	require.NoError(t, err)
+	err = ms.Systemstore().Set(ctx, []byte("/123"), []byte("123"))
+	require.NoError(t, err)
+
+	iter, err := rootstore.Iterator(ctx, corekv.IterOptions{KeysOnly: true})
+	require.NoError(t, err)
+
+	expectedKVs := []struct {
+		key string
+	}{
+		{
+			key: "blocks/",
+		},
+		{
+			key: "blocks/",
+		},
+		{
+			key: "blocks/to_merge/",
+		},
+		{
+			key: "data/",
+		},
+		{
+			key: "encryption",
+		},
+		{
+			key: "heads/",
+		},
+		{
+			key: "peers/",
+		},
+		{
+			key: "system/",
+		},
+	}
+
+	expectedIndex := 0
+	for {
+		hasNext, err := iter.Next()
+		require.NoError(t, err)
+
+		if !hasNext {
+			break
+		}
+
+		key, err := HumanReadableKey(iter.Key())
+		require.NoError(t, err)
+
+		require.True(t, strings.HasPrefix(key, expectedKVs[expectedIndex].key), key, expectedKVs[expectedIndex].key)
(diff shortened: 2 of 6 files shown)


internal/datastore/blockstore.go
```diff
@@ -45,6 +44,22 @@ import (
 // respective substores don't need to optimize or worry about Batching/Txn.
 // Hence the simplified ReaderWriter.
 
+func newCIDKeyBytes(c cid.Cid) ([]byte, error) {
+	sep := []byte("/")
```
AndrewSisley: question: Why bother?  Why not drop the `"/"` completely?

If you are worried about human readability you can just add the `"/"` when dumping, no need to waste bytes and code storing it.

internal/datastore/blockstore.go
```diff
@@ -45,6 +44,22 @@ import (
...
+	b := make([]byte, 0, c.ByteLen()+len(sep))
+	buf := bytes.NewBuffer(b)
+	_, err := buf.Write(sep)
```
getify: in what circumstance would this fail? and how would that be different from the `WriteBytes()` call a few lines later?

internal/datastore/blockstore.go
```diff
@@ -107,26 +126,32 @@ func (bs *bstore) Get(ctx context.Context, k cid.Cid) (blocks.Block, error) {
...
 func (bs *bstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
 	for _, b := range blocks {
-		k := dshelp.MultihashToDsKey(b.Cid().Hash())
-		exists, err := bs.store.Has(ctx, k.Bytes())
+		key, err := newCIDKeyBytes(b.Cid())
+		if err != nil {
+			return err
+		}
+
+		exists, err := bs.store.Has(ctx, key)
```
getify: this code seems the same as in `Put()`... should it be in a separate internal helper method?

moreover, it's duplicative of the internals of the public `Has()` so is it possible that the public method could be reused by `Put()` and `PutMany()`?

internal/datastore/blockstore.go
AndrewSisley: It can't fail.  The `err` return param is a compatibility thing so that `buffer.Buffer` satisfies an interface (maybe in the `io` std lib package, I cant remember).

Generally in defra we just handle the impossible error as it looks nicer than a `//nolint` comment that would be needed to avoid our linter from freaking out due to the unhandled error.  Handling it also requires less thought, both from reader and writer, as long as it isnt propagated and the calling func returns non-nil errors anyway.

Although, this particular function can never return an error, and so we are propogating this smell to a fair few more places - if we keep this function, which I don't want to (depending on Fred's justification for it), then a couple of `//nolint`s might be better than having `newCIDKeyBytes` return an always nil error param.

internal/datastore/blockstore.go
fredcarle: It's our prefix separator. We use that with every key.

internal/datastore/blockstore.go
fredcarle: It's not the same. Note what happens when `exists` is true.

internal/datastore/blockstore.go
AndrewSisley: We dont need to, especially not for the blockstore, where every key is a single value

AndrewSisley: Upgrading this discussion point to a todo, as I don't think the answer is a good enough justification for this.  As well as wasting resources and dev time, this also impacts anything else using the blockstore (such as lens), forcing the smell downstream.

AndrewSisley (dismissed):
I don't think keeping the `"/"` prefix is justified based on the answer received

internal/datastore/blockstore.go
getify: if the code doesn't have a way to semantically describe "an error can't actually happen here", and the linter would complain without the subsequent dummy `if` statement (or muting the linter), would there be a code readability benefit to some other approach that makes it more clear (e.g., a code comment explaining the same)?

----

I'm nitpicking here, for sure, as I'm not at all familiar enough with our codebase -OR- with the go language, to understand the various pros/cons of available idioms, but... as an exercise in my learning (since I was asked to code review) I am just "thinking outloud" if there's some approach that better balances code readability (to those who aren't in possession of the tribal knowledge *you all* have!).

For example....

```go
_, _ := SomeOperationWithImpossibleErrors()
```

Does that work? Is that possible to tell go, "hey, run this side effecting procedure, which always works without failure, and ignore the type-implied error that could (but won't) come back"?

If that's NOT possible, can you instead do something like:

```go
_, err := SomeOperationWithImpossibleErrors()
AssertNoError(err)
```

That `AssertNoError()` util would basically be a no-op, where it has an `if` statement (to satisfy the linter), but everybody knows it'll never run, by virtue of how the utility is named, and it would consistently be used in all places where this "error is impossible, so we're just no-op squashing it" happens, as a better more descriptive idiom.

And if neither of those could work, would it be too burdensome to do something like:

```go
_, err := SomeOperationWithImpossibleErrors()
if err != nil {
   panic("We should never get here", err)
}
```

IOW, I would prefer we weren't doing things that were known dead-code idioms and just relying on the fact that everyone just knows that intrinsically without the code giving any hints to the same. And to your point, by putting it into the return value, it's propagating that confusion outward, which makes it worse.

internal/datastore/blockstore.go
getify: hmmm... I am missing it I guess.

I see this code:

```go
key, err := newCIDKeyBytes(b.Cid())
if err != nil {
	return err
}

exists, err := bs.store.Has(ctx, key)
```

And I think I see that same code in both functions. Do I not?

getify: Moreover, the `Has()` function has:

```go
func (bs *bstore) Has(ctx context.Context, k cid.Cid) (bool, error) {
	key, err := newCIDKeyBytes(k)
	if err != nil {
		return false, err
	}
	return bs.store.Has(ctx, key)
}
```

That looks the same to me as the same snippet in both `Put()` and `PutMany()`?

fredcarle: `newCIDKeyBytes` will be removed so for context I'm answering this as if if was still going to exist.

We need `key, err := newCIDKeyBytes(k)` for the `Set` operation bellow regardless. So whether we use the `bs.store.Has` or `bstore.Has` method doesn't change much but if we used the `bstore.Has` version it would call `newCIDKeyBytes` twice for no added benefit.

Also, we can't just use the internals of `Put` in `PutMany` as what happens on `if exists` is different. One `continues` and the other `returns`.

internal/datastore/blockstore.go
fredcarle:
```go
_, _ := SomeOperationWithImpossibleErrors()
```
This works. In general it's better to handle possible errors in case the assumptions we've held about the inputs not causing errors become no longer valid. If we are certain that this will never be the case, there is nothing stopping us from using the above.

getify: does Go (and the linter) treat these two the same?

```go
_, _ := SomeOperationWithImpossibleErrors()
```

```go
SomeOperationWithImpossibleErrors()
```

If so, I'm not sure which of the two of those better communicates that you're intentionally ignoring errors (ostensibly because they cannot happen), but my *instinct* is that the former is a little better (even if more verbose) than the latter.

internal/datastore/blockstore.go
getify: but I was wondering -- sketching here, please bear with my rudimentary attempts at writing go code -- if a helper like this could be useful:

```go
func bikeshedMyName(ctx context.Context, k cid.Cid) (bool, []byte, error) {
   key, err := newCIDKeyBytes(k)
   if err != nil {
      return false, nil, err
   }
   exists, err := bs.store.Has(ctx, key)
   if err != null {
      return false, key, err
   }
   return exists, key, nil
}
```

And then it would be used in `Put()`, `PutMany()`, and `Has()`, like this:

```go
func (bs *bstore) Put(ctx context.Context, block blocks.Block) error {
   exists, key, err := bikeshedMyName(ctx, block.Cid())
   if err != nil {
      return err
   }
   if exists {
      return nil
   }
   return bs.store.Set(ctx, key, block.RawData())
}
```

```go
func (bs *bstore) PutMany(ctx context.Context, blocks []blocks.Block) error {
   for _, b := range blocks {
      exists, key, err := bikeshedMyName(ctx, b.Cid())
      if err != nil {
         return err
      }
      if exists {
         continue
      }
      err = bs.store.Set(ctx, key, b.RawData())
      if err != nil {
         return err
      }
   }
   return nil
}
```

```go
func (bs *bstore) Has(ctx context.Context, k cid.Cid) (bool, error) {
   exists, key, err := bikeshedMyName(ctx, k)
   if err != nil {
      return false, err
   }
   return exists, nil
}
```

---

Now, this helper abstraction may very well *not be worth it*, I'm not experienced enough here to judge. If so, happy to move on.

But I was mostly just asking to exercise my understanding of go code and of our code base. I *am* curious if the above sketch would/could work, separately from whether you think it *should* be done?

AndrewSisley: It would work, and I would be fairly against doing it :grin: 

The `bs.store.Has` call is an optional optimisation in `Put` and `PutMany` - they shouldn't be forced to use it, and it shouldnt be hidden from anyone reading those functions.  I'd see `bikeshedMyName` as unhelpful coupling and abstraction.

internal/datastore/blockstore.go
fredcarle: They are treated differently. With `_, _` we've technically handled the error. We made a deliberate choice to discard it. With the other one its not clear if it was forgotten or done on purpose.

internal/datastore/blockstore.go
fredcarle: > I'd see bikeshedMyName as unhelpful coupling and abstraction.

Same :)

getify: My apologies for prolonging a point that it seems you all are ready to move on from.

My question was triggered by the fact that this snippet of code (using `newCIDKeyBytes()` and `bs.store.Has()`, and the associated `if` conditionals) is essentially repeated 3 times (in `Has()`, `Put()`, and `PutMany()`) all in the same file, and that seems to match the typical best-practice guideline coined by [Martin Fowler (et al), as "Rule of Three"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

Again, I'm not raising a blockage to insist on us making this change, but trying to use this as a learning opportunity for myself. To that point, I don't understand these bits of your reply:

> The bs.store.Has call is an optional optimisation in `Put` and `PutMany`

"Optional" meaning what? If the optimization is helpful and important (as the code comment suggests it is), and it deserves to be in the code, then it's good to be there, and not really "optional" in the most strict sense. In general/basic terms, all optimizations are "optional", but... if they're important, then they *should* be there.

The inverse also holds... if these optimizations aren't that important (and are thus truly "optional") should they be removed? If they're not helping much, they might be worse being left there (larger code size, more code to learn by new team members, more code to maintain, etc).

> they shouldn't be forced to use it

I don't know what "forced" means in this case? They're "forced" to use it if the code is there, and they're not "forced" to do it if the code is removed or commented out. At present, they're "forced" to use it since the code is there.

> it shouldnt be hidden from anyone reading those functions

It's not "hidden" by any reasonable definition of that word that I can muster. The helper is right there next to the functions being used, AND I'd advocate that the name of the helper (to be bikeshedded) should in fact describe what it's doing, which in this case means it should indicate in some way that it's doing this cheaper "has" lookup.

Abstractions (like functions) which are named well, don't "hide" things, they separate concerns so they can be reasoned about in smaller chunks.

> coupling and abstraction.

If you don't mind, would you elaborate on what "coupling" you're concerned with? In all three functions `Put()`, `PutMany()`, and `Has()`, they all do the `newCIDKeyBytes()` and `bs.store.Has()` calls together, along with the surrounding `if` conditionals. In that sense, those two operations are already "coupled", so I don't think factoring out to a private helper changes that.

> unhelpful

I certainly trust and defer to your subjective judgement call here. I'm just trying to learn what design perspectives here inform that, so that I am better prepared to make similar subjective judgements going forward. :)

AndrewSisley: > "Optional" meaning what? 

I think of calling `Has` as optional for two main reasons:

1. It is an un-benchmarked guess made by the original author that this *might* improve performance.  It probably doesn't.  If it would have improved performance, it doesn't belong here, and would have or should have been added to the underlying store `Set` functions.  

It is for this reason you will not find this `Has` call in the `corekv` stores.  Even within that library, I see this as something that belongs a layer lower (unless we have to work around the 3rd party libs for practical reasons) - if it made sense for `badger` to perform a `Has` call before inserting data, they probably would have, if not, there are probably reasons we haven't thought of yet that could even necessitate the removal of the `Has` calls here, regardless of performance.

TLDR: I dislike it regardless, but removing it is probably out of scope of this PR, and I doubt Fred wanted to think about it right now :)

2. I see no reason for it to be the same across all three functions.  Even if right now, the `Has` is beneficial for all three, maybe in the future we find a way of optimising one of the three that does not require `Has`.

e.g. (and this is really just an example) if `corekv` were to in the future expose `HasAll` and `SetAll` functions, we may find that it is much more performant for `PutAll` to use those instead of looping through and calling `Has` and `Set`.

Obscuring the `Has` call in `bikeshedMyName`, and binding the three functions together with it makes this harder to spot, and harder to implement than if the `Has` call was inline as it is now.  

This is in contrast to `newCIDKeyBytes` which I see as mandatory - the behaviour *must* be the same across all three functions, otherwise they will not work.

My view is somewhat coloured by my past experiences, where I have managed to do a lot of very useful simplification and performance work by *unrefactoring* code that really should not have been tangled up in well meaning but unhelpful abstractions - to the extent that unrefactoring basically became my default approach when dealing with spagettified or unperformant code.  

In my experience, there are far worse code crimes than code duplication, and overly aggressive abstraction is one of the most common causes of it.

Other developers may have had other experiences that push them in different directions, and they probably have valid reasons for that.

getify: thank you @AndrewSisley!

AndrewSisley: Thank you @getify  - I like talking about this kind of thing :grin:  Hopefully Fred doesnt mind that we are doing it in his PR - if the conversation continues we should really pull it out into a discord thread (I should really have done that instead of replying here...).

fredcarle: > If it would have improved performance, it doesn't belong here, and would have or should have been added to the underlying store Set functions.

I don't fully agree with this. The underlying `Set` function should just worry about "setting" the value for the provided key. It shouldn't behave any differently if the key already exists. For the blockstore however, if a key already exists, it doesn't make sense to waste CPU cycles on writing the bytes to storage once more as they would be the exact same bytes. And if the block is large, it would make a sizeable difference on performance. No need to benchmark this to know that there would be an impact proportional to the size of the data written. This is why this optimization has been added to `Put` and `PutMany` blockstore level functions and not to the lower level corekv.

AndrewSisley: > corekv Set with Has

Ah true, corekv cant assume a key can only ever have one valid value in the same way the blockstore can :)

> No need to benchmark 

I still disagree with you :)  We are slowing down the normal/happy path with an extra io operation.  It probably speeds up an unhappy path if we try and write the same value multiple times.  We have not measured how frequently either of those paths are taken (if the unhappy path is never hit, `Has` is detrimental to performance), and we have not measured the gain/loss per path with the `Has` in/out.  We also have not checked when those paths are taken, and decided how much we care about them (e.g. if it is detrimental to user-time, we may wish to take the throughput hit and remove `Has` even if overall it reduces CPU-time/system-throughput).

shahzadlone (approved):
I am happy with the changes although I know there are some on going discussions for further size reduction.

Should this really be `chore` instead of `refac`?

fredcarle:
> I am happy with the changes although I know there are some on going discussions for further size reduction.

The discussion has been resolved :) It's fully reduced now.

> Should this really be `chore` instead of `refac`?

Probably. The initial intent of this PR was a chore. It evolved into a refactor.

internal/datastore/blockstore.go
fredcarle: Done

internal/datastore/multi_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"testing"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corekv/memory"
+)
+
+func TestMultistore_HumanReadableKeys_ShouldSucceed(t *testing.T) {
```
islamaliev, line 25: praise: thanks for the tests

islamaliev (approved):
LGMT

PR #4029, opened 2025-09-20
fix: Remove recursive issue in versioned fetcher

### LINKED ISSUE #3988: SSE next event payload incorrectly reports value of CRDT `pcounter` field
SSE next events after a `mutation` affecting a `@crdt(type: pcounter)` field indicate a different (wrong) value for the field compared to the query response, and the actual state of the field in the database.

I issued this:

```graphql
type User {
  name: String
  email: String @index(unique: true)
  counter: Int @crdt(type: pcounter)
  createdAt: String
}
```

Then this:

```graphql
mutation CreateUser {
  create_User(input:{ name:"Ada", email:"ada@example.com", counter:1, createdAt:"2025-09-02T18:10:27.601Z" }) {
    _docID name email counter createdAt
  }
}
```

The client returns back:

```json
{
  "data": {
    "create_User": [
      {
        "_docID": "bae-91f25db6-8f0d-5eb3-b0a2-efe7820aa3bc",
        "counter": 1,
        "createdAt": "2025-09-02T18:07:45.162Z",
        "email": "ada@example.com",
        "name": "Ada"
      }
    ]
  }
}
```

And the SSE event that comes through has:

```js
{
  User: [
    {
      _docID: 'bae-91f25db6-8f0d-5eb3-b0a2-efe7820aa3bc',
      counter: 1,
      email: 'ada@example.com',
      name: 'Ada'
    }
  ]
}
```

That's just as expected. But then if I issue this:

```graphql
mutation UpdateUser {
  update_User(
    filter:{ email:{ _eq:"ada@example.com" } },
    input:{ name:"Ada Lovelace", counter:1 }
  ){ _docID name email counter }
}
```

The client returns:

```json
{
  "data": {
    "update_User": [
      {
        "_docID": "bae-91f25db6-8f0d-5eb3-b0a2-efe7820aa3bc",
        "counter": 2,
        "email": "ada@example.com",
        "name": "Ada Lovelace"
      }
    ]
  }
}
```

Again, as expected, however the SSE event says:

```js
{
  User: [
    {
      _docID: 'bae-91f25db6-8f0d-5eb3-b0a2-efe7820aa3bc',
      counter: 4,
      email: 'ada@example.com',
      name: 'Ada Lovelace'
    }
  ]
}
```

Hmmm... why does the client see `counter: 2` (as expected), but the SSE event shows `counter: 4`?

And moreover, if I then issue:

```graphql
query FindUser {
  User(filter:{ email:{ _eq:"ada@example.com" } }) { _docID name email counter }
}
```

Indeed, I see in the client, as expected:

```json
{
  "data": {
    "User": [
      {
        "_docID": "bae-91f25db6-8f0d-5eb3-b0a2-efe7820aa3bc",
        "counter": 2,
        "email": "ada@example.com",
        "name": "Ada Lovelace"
      }
    ]
  }
}
```

So it's only that the SSE event had the surprising/wrong value for that `counter` value, but what's actually in the DB (and coming back from the client) seems to be correct.

Further, if I issue several additional mutations (incrementing the counter by 1 each time), the values of that `counter` field as they come through the SSE next events progress from `4` (should have been `2`) to `10` (should have been `3`), to `20` (should have been `4`), to `35` (should have been `5`). Interesting pattern of value sequence there, not sure what is going on.

fredcarle:
## Relevant issue(s)

Resolves #3988 

## Description

This PR resolves an issue where a version query (`query { User (cid: "someCID") {...}}`) would merge each commit recursively all the way down to the root. 

Assume updates 1, 2 and 3 on field A. When specifying version 3 in the query, for example,  the versioned fetcher loads Composites `C1`, `C2` and `C3` into memory and merges to the datastore waling from `C1` to `C3`. 

What happened was this:
Merge `C1`-`A1` -> Merge `C2`- `C1` - `A1` - `A2` - `A1` -> Merge `C3` - `C2` - `C1` - `A1` -  `A2` - `A1` - `A3` - `A2` - `A1`. 

Basically the merge function was walking all links (heads included) instead of only the non-head links.

We never detected this because all our tests that do a version query (including subscriptions) only use LWW CRDTs and this recursion yields the same result for the type.

With Counter CRDTs, imagine 3 increments of 1. The create would yield 1, the first update would yield 4 and the second update would yield 10.

## Tasks

## How has this been tested?

Added 2 new integrations tests

- MacOS

### DIFF
--- a/internal/db/fetcher/versioned.go
+++ b/internal/db/fetcher/versioned.go
@@ -407,7 +407,9 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 		return err
 	}
 
-	// handle subgraphs
+	// Handle subgraphs. We range over `Links` only (not `Heads``) because the trunk is already accounted for
+	// by the initial caller or `merge`. Including `Heads` would result in unnecessary recursion and possible
+	// wrong final value for the fields.
 	for _, l := range block.Links {
 		err = vf.merge(l.Cid)
 		if err != nil {
--- a/internal/db/fetcher/versioned.go
+++ b/internal/db/fetcher/versioned.go
@@ -408,7 +408,7 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 	}
 
 	// handle subgraphs
-	for _, l := range block.AllLinks() {
+	for _, l := range block.Links {
 		err = vf.merge(l.Cid)
 		if err != nil {
 			return err
--- a/tests/integration/query/simple/with_cid_test.go
+++ b/tests/integration/query/simple/with_cid_test.go
@@ -118,3 +118,44 @@ func TestQuerySimpleWithCid_MultipleDocs(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQuerySimpple_WithCIDAndCounterAfterUpdate_ShouldSucceed(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			action.AddSchema{
+				Schema: `
+					type User {
+						counter: Int @crdt(type: pcounter)
+					}
+				`,
+			},
+			testUtils.CreateDoc{
+				CollectionID: 0,
+				DocMap: map[string]any{
+					"counter": int64(1),
+				},
+			},
+			testUtils.UpdateDoc{
+				CollectionID: 0,
+				DocID:        0,
+				Doc:          `{"counter": 1}`,
+			},
+			testUtils.Request{
+				Request: `query {
+					User(cid: "{{.CID0_0_1}}") {
+						counter
+					}
+				}`,
+				Results: map[string]any{
+					"User": []map[string]any{
+						{
+							"counter": int64(2),
+						},
+					},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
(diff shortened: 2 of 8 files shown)


tests/integration/query/simple/with_cid_test.go
```diff
@@ -118,3 +118,44 @@ func TestQuerySimpleWithCid_MultipleDocs(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestQuerySimpple_WithCIDAndCounterAfterUpdate_ShouldSucceed(t *testing.T) {
```
getify: TestQuerySimpple -> TestQuerySimple

internal/db/fetcher/versioned.go
```diff
@@ -408,7 +408,7 @@ func (vf *VersionedFetcher) merge(c cid.Cid) error {
 	}
 
 	// handle subgraphs
-	for _, l := range block.AllLinks() {
+	for _, l := range block.Links {
```
getify, line 413: nitpick: should we update the code comment above this line to explain why `Links` is used?

tests/integration/replace.go
```diff
@@ -0,0 +1,60 @@
...
+	"CID": func(s *state.State, nodeID int) map[string]string {
+
+		docIDsToCIDs := s.Nodes[nodeID].Composites
+
+		res := map[string]string{}
+		for colIndex, docIndexes := range s.DocIDs {
+			for dIndex, docID := range docIndexes {
+				cids := docIDsToCIDs[docID.String()]
+				for cidIndex, cid := range cids {
+					res["CID"+strconv.Itoa(colIndex)+"_"+strconv.Itoa(dIndex)+"_"+strconv.Itoa(cidIndex)] = cid.String()
```
getify: nit:

would the code be clearer here if this long string concatenation were assigned to a local variable, named for example `templateCIDRef` or something, and then used as `res[templateCIDRef] = cid.String()`?

getify (changes requested):

getify (approved):
LGTM :)

shahzadlone (approved):
Nice! This was a good find.

LGTM

PR #4034, opened 2025-09-24
feat: Add Android build support

### LINKED ISSUE #4033: Allow builds for Android
Currently, the C bindings can be built for Linux and the amd64 architecture. We must also support builds for Android. This would mean supporting the arm64-v8a for modern devices, but ideally it should also support x86_64 architecture, to aid developers (it would allow many of the emulators inside Android Studio to run our code.)

ChrisBQu:
## Relevant issue(s)

Resolves #4033

## Description

This adds support for generating shared object files for Android, using both arm64-v8a architecture (which is used on modern devices), and x86_64 architecture (which is used in Android Studio emulators on Windows.) To do this, a build script was added at `tools/scripts/build-c-shared-android.sh` and a build command was added inside the makefile which invokes it, `build-c-shared-android`.

To actually allow the build to work, and then run correctly, some changes have been made to the code base. See that `lens_wasmer.go` , `lens_wasmtime.go` and `lens_wazero.go` now use build directives to not build when we are targeting Android. Additionally, there is a new file, `lens_wazero_android.go` which is similar to `lens_wazero.go` except that it sets the DefaultLens.

Lastly, `go.mod` now has one replace directive:

```
```

## Tasks

## How has this been tested?

- WSL
- Windows


node/lens_wazero_android.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/sourcenetwork/lens/host-go/engine/module"
+	"github.com/sourcenetwork/lens/host-go/runtimes/wazero"
+)
+
+const Wazero LensRuntimeType = "wazero"
+
+func init() {
+	runtimeConstructors[DefaultLens] = func() module.Runtime { return wazero.New() }
```
fredcarle, line 23: thought: I think you could add this line to `lens_wazero.go` instead of creating a new file where that is the only difference.

Makefile
```diff
@@ -436,6 +436,12 @@ fix:
...
-	
+
+.PHONY build-c-shared-linux:	
 build-c-shared-linux:
 	@tools/scripts/build-c-shared-linux.sh $(BUILD_FLAGS)
+
+API_LEVEL ?= 21
+.PHONY: build-c-shared-android
+build-c-shared-android:
+	@tools/scripts/build-c-shared-android.sh $(ANDROID_NDK) $(API_LEVEL) "$(BUILD_FLAGS)"
```
fredcarle, line 452: suggestion: It would be nice to document what `ADROID_NDK` and `API_LEVEL` refer to.

node/lens_wazero.go
```diff
@@ -20,5 +20,6 @@ import (
 const Wazero LensRuntimeType = "wazero"
 
 func init() {
+	runtimeConstructors[DefaultLens] = func() module.Runtime { return wazero.New() }
```
AndrewSisley: todo: This makes the default lens unpredictable for non-js builds, as the default will now depend on which init function executes first (see `lens_wasmtime.go`).  Please dont do this, and instead set the default for android only (like it was originally in the PR I think)

AndrewSisley (approved):
LGTM, assuming the Lens-default conversation is resolved before merge please.

node/lens_wazero.go
ChrisBQu: I understand the concern now. Let's raise it as a discussion point in the stand up, so that everyone can get on the same page. I will not merge for now.

fredcarle: My bad. I forgot that the build tags were only negatively filtering. You can revert back to what you had.

PR #4035, opened 2025-09-24
chore: Allow test framework to test message relaying

### LINKED ISSUE #4030: Allow testing framework to test message relaying
The DHT currently doesn't relay messages for loopback addresses (`127.0.0.1`). We should try to get P2P enabled nodes to use their local address instead.

fredcarle:
## Relevant issue(s)

Resolves #4030 

## Description

~~This PR fixes a devex issue where bootstrap instance would not be connected to until a message was sent over the pubsub network. This resulted it the bootstrap instance not knowing about the new instance until later than expected.~~

~~The workaround for this was to call `Connect` instead of (or in addition to) defining a bootstrap instance.~~

~~With this fix, the bootstrapped instance will know about the new instance as soon as it goes online.~~

Edit: The above was taken care of in `go-p2p` which will be brought in from another PR. Removing the fix from this branch.

In addition to this, this PR makes it so the test tries to use a local address instead of the loopback. This is to make sure that the DHT can relay messages across the network. Loopback addresses are ignored by the DHT. My assumption is that the creator of the package assumed that peers using the loopback address would be using mDNS instead when trying to discover other local peers.

## Tasks

## How has this been tested?

New multi-instance chain test

- MacOS

### DIFF
--- a/internal/db/p2p/p2p.go
+++ b/internal/db/p2p/p2p.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/fxamacker/cbor/v2"
 	"github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
 
 	"github.com/sourcenetwork/corelog"
 	"github.com/sourcenetwork/immutable"
@@ -189,7 +190,9 @@ func (p *P2P) hasAccess(ctx context.Context, pid string, c cid.Cid) bool {
 
 	rawblock, err := txn.Blockstore().Get(ctx, c)
 	if err != nil {
-		log.ErrorE("Failed to get block", err)
+		if !ipld.IsNotFound(err) {
+			log.ErrorE("Failed to get block", err)
+		}
 		return false
 	}
 
--- a/internal/db/p2p/p2p.go
+++ b/internal/db/p2p/p2p.go
@@ -13,6 +13,7 @@ package p2p
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


internal/db/p2p/p2p.go
```diff
@@ -410,40 +414,37 @@ func (p *P2P) processPushlogRequest(
...
+		Cid:          headCID,
+		CollectionID: req.CollectionID,
+	}
+	err = p.db.Merge(ctx, mergeEvt)
+	if err != nil {
+		// log.ErrorContextE(
+		// 	ctx,
+		// 	"Failed to execute merge",
+		// 	err,
+		// 	corelog.Any("Event", mergeEvt))
```
islamaliev: todo: please remove commented out code

net/peer.go
```diff
@@ -129,6 +129,16 @@ func NewPeer(
 		return nil, err
 	}
 
+	for _, peer := range peers {
```
islamaliev: nitpick: could be in a separate function

tests/integration/p2p.go
```diff
@@ -464,12 +465,33 @@ func reconnectPeers(s *state.State) {
...
 		}
 	}
 }
 
+func getIPString() string {
+	loopbackIP := "127.0.0.1"
+
+	conn, err := net.Dial("udp", "8.8.8.8:80")
+	if err != nil {
+		// If getting the local address fails, we simply return the loopback address.
```
islamaliev, line 479: nitpick: would be also nice to know in which case it could fail and what it would mean for a test.

tests/integration/net/simple/peer/with_create_test.go
```diff
@@ -275,6 +275,71 @@ func TestP2PCreate_WithP2PCollectionWithNodeChain_ShouldSucceed(t *testing.T) {
 	testUtils.ExecuteTestCase(t, test)
 }
 
+func TestP2PCreate_WithP2PCollectionOnLastNodeInNodeChain_ShouldSucceed(t *testing.T) {
```
islamaliev: nitpick: the test's name doesn't really tell what is expected. 
"ShouldSucceed" should be used when it's clear from the context what is expected. Here it's different.
Something like this I think would be better `TestP2PCreate_WithP2PCollectionOnLastNodeInNodeChain_ShouldPropagateUpdate` or 
`TestP2PCreate_WithP2PNodeChainAndUpdateOnOneEnd_ShouldPropagateToAnotherEnd`

islamaliev (approved):
Looks good. I have some minor suggestions.

internal/db/p2p/p2p.go
fredcarle: done

net/peer.go
fredcarle: reverted

tests/integration/p2p.go
fredcarle: Added documentation

tests/integration/net/simple/peer/with_create_test.go
fredcarle: Done

shahzadlone (approved):
LGTM

PR #4044, opened 2025-09-26
test: Test blocking off of patching secondary indexes

### LINKED ISSUE #3890: Check for test coverage of sec. indexes after collection patch
And make sure it works.  I think it might have been explicitly blocked off, but I'm not sure.

islamaliev:
## Relevant issue(s)

Resolves #3890

## Description

Add tests for patching collection's `Indexes` property.


tests/integration/index/patch_test.go
```diff
@@ -0,0 +1,124 @@
...
+                                        "Name": "email",
+                                        "Descending": false
+                                    }
+                                ],
+                                "Unique": false
+                            }
+						}
+					]
+				`,
+				ExpectedError: "collection indexes cannot be mutated",
```
fredcarle, line 49: question: Why are we not allowing adding secondary indexes from the patch operation?

tests/integration/index/patch_test.go
```diff
@@ -0,0 +1,124 @@
+// Copyright 2025 Democratized Data Foundation
```
AndrewSisley, line 1: thought: This file does feel out of place - all the existing patch collection tests are currently in `tests/integration/collection_version/updates`, including a single one testing secondary indexes (`TestColVersionUpdateReplaceIndexes_Errors`).

I do have a preference to keep them all in one place, but that is likely biased by me largely working these tests from the perspective of the patch system, instead of the various feature/components.  Do people feel like many of them would be better placed in their own feature directories, such as here?

tests/integration/index/patch_test.go
```diff
@@ -0,0 +1,124 @@
...
+				`,
+				ExpectedError: "collection indexes cannot be mutated",
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestPatchCollection_ModifySecondaryIndex_ShouldError(t *testing.T) {
```
AndrewSisley, line 85: thought: I think this test is the only new test that increases code-line coverage, the other two cover lines already covered by `TestColVersionUpdateReplaceIndexes_Errors`.

How many tests we need for this is kind of thing is vague to me, I'm not sure the code complexity warrants covering every patch operation for secondary index items at the moment.

I am unsure of my preferences regarding this at the moment, but I thought I'd share the thought.

AndrewSisley (approved):
LGTM, but it is probably worth resolving the discussion regarding the location of the tests before merge.

tests/integration/index/patch_test.go
AndrewSisley: To save time in the short term - the priority of allowing patch operations on secondary index definitions was/is different to the priority of introducing the secondary index feature.  Blocking it off in the short term allowed the more important feature to be merged earlier, without worrying too much about patch.

Encrypted indexes are doing the same I think, and I support this approach.

tests/integration/index/patch_test.go
islamaliev: as discussed in the standup I'm leaving it as is, because the majority of team members are also in favour of this approach.

PR #4047, opened 2025-10-02
refactor: Use 64bit cid encoding in blockstore

### LINKED ISSUE #4046: Replace Defra blockstore with corekv

AndrewSisley:
## Relevant issue(s)

Resolves #4046

## Description

Use 64bit cid encoding in the blockstore instead of the old `dshelp` 32bit encoding, and preserve the version/codec.

Also replaces the Defra blockstore with one from corekv (the corekv blockstore uses the raw 64bit encoding).


fredcarle (approved):
LGTM

Issue #4021, opened 2025-09-16
Server 500 from cid+_version query


getify:
When I issue this query:

```graphql
query {
  User(cid:"-whatever-") {
    _version {
      cid height signature { __typename }
    }
  }
}
```

I got back a server 500 error (with no error text).

But when I removed the `_version` stuff (and selected other fields), it worked fine.

AndrewSisley:
Possibly related to https://github.com/sourcenetwork/defradb/issues/1709 and #4144 - consider testing both when picking up this ticket

PR #4051, opened 2025-10-04
chore: Change from net to go-p2p

### LINKED ISSUE #4031: Drefra should automatically try to connect to bootstrap nodes on startup
The configured bootstrap nodes get added to the DHT but are only connected to when the node emits a message. This means that the node that we want to bootstrap with will not know of our existence until we decide to connect to them. 

We should instead ensure that we try to connect to bootstrap nodes on startup to let them be aware of our existence.

### LINKED ISSUE #4048: Replace Defra `net` package with `go-p2p` repo

fredcarle:
## Relevant issue(s)

Resolves #4048
Resolves #4031

## Description

This PR removes the `defradb/net` package in favour of `sourcenetwork/go-p2p`. It brings in the latest changes from `go-p2p` which implies making changes to the `client.Host` interface.

## Tasks

## How has this been tested?

make test

- MacOS


cli/p2p_connect.go
```diff
@@ -11,30 +11,36 @@
...
-  defradb client p2p connect '{"ID": "12D3", "Addrs": ["/ip4/0.0.0.0/tcp/9171"]}'
-  		`,
+		Use:   "connect <addresses>",
+		Short: "Connect to one or more peers",
+		Long: `Connect to one or more peers with the given addresses
+
+Example: Connect to a peer
+  defradb client p2p connect /ip4/0.0.0.0/tcp/9171/p2p/12D3KooW...
+  
+Example: Connect to multiple peers
```
AndrewSisley, line 26: praise: Thanks for adding this

cli/p2p_connect.go
```diff
@@ -11,30 +11,36 @@
...
+		`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliClient := mustGetContextCLIClient(cmd)
 
-			var info client.PeerInfo
-			if err := json.Unmarshal([]byte(args[0]), &info); err != nil {
-				return err
+			var addresses []string
+			for _, id := range strings.Split(args[0], ",") {
```
AndrewSisley: thought: This split feels odd to find in a CLI command, and seems undocumented. It is also only performed on the first arg - yet the command documentation suggests a space (` `) delimitation, which is more common in CLIs.

question: What is the motivation for comma delimitation here?

todo: If both comma and space delimination is kept, please ensure that it is applied to all args.

todo: If space delimination is kept (with or without comma), as the documentation suggests, please make sure `args[N>1]` are added to `addresses` - it looks like they are ignored.

suggestion: It looks like the delimination (spaces and commas) are untested, please add CLI integration tests for this (in `cli/test`), unless unreasonably difficult to do so (I think you'll need to define actions, as only a handful have been added so far).

cli/p2p_replicator_set.go
```diff
@@ -27,17 +25,21 @@ func MakeP2PReplicatorSetCommand() *cobra.Command {
...
 `,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliClient := mustGetContextCLIClient(cmd)
 
-			var info client.PeerInfo
-			if err := json.Unmarshal([]byte(args[0]), &info); err != nil {
-				return err
+			var addresses []string
+			for _, id := range strings.Split(args[0], ",") {
```
AndrewSisley: todo: The requests in `p2p_connect` all look like they also apply here.

go.mod
```diff
@@ -42,10 +38,13 @@ require (
 	github.com/philippgille/chromem-go v0.7.0
 	github.com/pkg/errors v0.9.1
 	github.com/sourcenetwork/acp_core v0.4.1
-	github.com/sourcenetwork/corekv v0.1.2
-	github.com/sourcenetwork/corekv/blockstore v0.0.0-20250917232726-9384d90f13b5
+	github.com/sourcenetwork/corekv v0.2.3
+	github.com/sourcenetwork/corekv/badger v0.0.0-20251002183059-6d8366118c02
```
AndrewSisley: todo: I added the tag for badger, please update this dependency to `v0.2.3` - it should work now

internal/db/p2p/p2p.go
```diff
@@ -71,8 +71,11 @@ type P2P struct {
 	db   DB
 	host client.Host
 
-	// replicators is a map from collection CollectionID => peerId
-	replicators map[string]map[string]client.PeerInfo
+	// replicators is a map from collection CollectionID => peerId => list of addresses.
+	// This is a cached in-memory copy of the persisted replicators in the database.
+	// It is used to quickly find the replicators for a given collection when sending updates.
+	// The map is protected by repMu.
```
AndrewSisley, line 101: praise: Thanks for the documentation here, the type alone is not ideal.

thought: Perhaps aliasing (not typing) `[]string` and and `map[string][]string` is worth it here.  Something like the below, but with proper names and some of the documentation from your original PR-code comment.
```
// note: I think `=` is important - no reason to lock down the types, they are descriptive only
type addresses = []string  
type peerAddresses = map[string]addresses
...
P2P {
replicatorsByCollectionID map[string]peerAddresses 
}
```

internal/db/p2p/replicator.go
```diff
@@ -99,36 +94,71 @@ func (p *P2P) SetReplicator(ctx context.Context, repInfo client.PeerInfo, collec
...
+			return err
 		}
-	}
 
-	// persist replicator to the datastore
-	newRepBytes, err := json.Marshal(storedRep)
-	if err != nil {
-		return err
-	}
+		storedRep := client.Replicator{}
```
AndrewSisley, line 112: suggestion: This change does not look like it is noted in the PR description - please add it for future reference.

todo: It looks like an optimisation, can you please document in this block why we are doing it.

AndrewSisley (approved):
LGTM, thanks for this Fred :)

I have a couple of requests that I would like resolving before merge, but they are all pretty localised.

cli/p2p_connect.go
fredcarle: That is a mistake of mine. It should be spaces and that would be different arguments. The comma part was a copy pasting mistake. I'll fix that

go.mod
fredcarle: will do.

internal/db/p2p/p2p.go
fredcarle: I'll look into it. Good suggestion!

internal/db/p2p/replicator.go
fredcarle: Not sure what you are referring to. The line that has the comment hasn't changed. It's not an optimization. It's simply checking if a replicator with that ID exists and if so, it will update the list of collections instead of overwriting it. If that is not clear from the code I can document it.

cli/p2p_connect.go
fredcarle: Done

cli/p2p_replicator_set.go
fredcarle: Done

go.mod
fredcarle: done

internal/db/p2p/p2p.go
fredcarle: done

fredcarle: Let me know what you think.

cli/p2p_connect.go
fredcarle: Let me know if you like the added tests

cli/test/integration/p2p/connect_test.go
```diff
@@ -0,0 +1,95 @@
...
+		"12D3KooWGj2wWqxSKB2Lqg6K6ye7N7W4YcpoeNpdXakHkgGjUqHC",
+		"12D3KooWBVR4AGs7pTwdjUH1Pd9bgh85hyf1xtV3rg7bj3VfZGoT",
+	}
+	invalidPeerID = "invalid-peer-id"
+	addresses     = []string{
+		fmt.Sprintf("/ip4/127.0.0.1/tcp/9000/p2p/%s", peerIDs[0]),
+		fmt.Sprintf("/ip4/127.0.0.1/tcp/9001/p2p/%s", peerIDs[1]),
+	}
+	addressWithInvalidPeerID = fmt.Sprintf("/ip4/127.0.0.1/tcp/9000/p2p/%s", invalidPeerID)
+	addressWithInvalidIP     = "/ip4/999.999.999.999/tcp/9000/p2p/12D3KooWGj2wWqxSKB2Lqg6K6ye7N7W4YcpoeNpdXakHkgGjUqHC"
```
AndrewSisley, line 32: thought: I dislike having these here, away from the test declarations, but I assume most will move once support for multiple nodes in the cli tests is added.

question: What actually broke when you added multiple?  I thought it would work and was already used (it is not, I must have mixed memories with Lens).

cli/p2p_connect.go
AndrewSisley: The tests look really good, and thanks for documenting the multi-node problem in the relevant tests :)

internal/db/p2p/p2p.go
```diff
@@ -40,6 +40,13 @@ var (
 	tracer = telemetry.NewTracer()
 )
 
+type (
```
AndrewSisley, line 48: thought: I had no idea we could do this, although it makes sense from a syntax perspective :) Thanks :)

internal/db/p2p/p2p.go
AndrewSisley: Looks good, and thanks for aliasing the collectionID, it keeps the name short and is more descriptive than my suggestion.

internal/db/p2p/replicator.go
AndrewSisley: > checking if a replicator with that ID exists and if so, it will update the list of collections instead of overwriting it

question: This is a behaviour change?  Previously is `SetReplicators` was called multiple times for a given peer, previous collections would be wiped, but now they will be kept?

suggestion: If this is a behaviour change, as per my question, please consider changing the PR title - noting the change in the release-change-log/git-history is more important than the chore IMO.

suggestion: If `SetReplicators` is now only ever going to append replicators, perhaps a name change is needed, as `Set` does not really reflect this.  Please make sure any high level public layers on top of this func are changed too if need be, including CLI/http.

fredcarle: > question: This is a behaviour change? Previously is SetReplicators was called multiple times for a given peer, previous collections would be wiped, but now they will be kept?

This was always the behavior. Nothing changed for a given peer. The only change is that since we receive a list of addresses only, we have to extract the peer IDs from the addresses and it is possible to get more than one peer. So we do the same process as before except that we do it for all the peers that we found.  The order of operations in that function changed a little to accommodate the multiple possible peers but the code itself remains pretty much the same.

> suggestion: If this is a behaviour change, as per my question, please consider changing the PR title - noting the change in the release-change-log/git-history is more important than the chore IMO.

No behavior change :)

> suggestion: If SetReplicators is now only ever going to append replicators, perhaps a name change is needed, as Set does not really reflect this. Please make sure any high level public layers on top of this func are changed too if need be, including CLI/http.

Its always just appended. Happy to change the name if I get a good suggestion.

cli/test/integration/p2p/connect_test.go
fredcarle: > thought: I dislike having these here, away from the test declarations, but I assume most will move once support for multiple nodes in the cli tests is added.

The exact address is not important to the tests other than the fact that its valid or not and they are reused across all the p2p tests. So it made sense to me to have them defined this way.

> question: What actually broke when you added multiple? I thought it would work and was already used (it is not, I must have mixed memories with Lens).

Can you clarify this question please. Not sure what you are referring to.

internal/db/p2p/replicator.go
AndrewSisley: Ah okay - my mistake, thanks for the explanation Fred!

cli/test/integration/p2p/connect_test.go
AndrewSisley: > Can you clarify this question please. Not sure what you are referring to.

There are a couple of tests with the below comment, I read it as you tried to setup multiple nodes and the test action couldn't handle it somehow.

> // TODO: Add capability to have multiple defradb instances in tests, so we can
// actually test successful connections. #4021

cli/test/integration/p2p/connect_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	test.Execute(t)
+}
+
+// NOTE: This test currently fails because there is no peer listening at the given address.
+// However, it does at least verify that a single address can be passed in.
+//
+// TODO: Add capability to have multiple defradb instances in tests, so we can
+// actually test successful connections. #4021
```
shahzadlone: suggestion: Mark the issue number #4021 in a similar format like our other todos (url)? Here and other places

cli/test/action/p2p_replicator_delete.go
```diff
@@ -0,0 +1,59 @@
...
+
+package action
+
+import (
+	"strings"
+
+	"github.com/stretchr/testify/require"
+)
+
+// P2PReplicatorDelete executes the `client p2p replicator set` command.
```
shahzadlone: todo: incorrect documentation, please fix.

```suggestion
// P2PReplicatorDelete executes the `client p2p replicator delete` command.
```

cli/test/action/p2p_replicator_getall.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+)
+
+// P2PReplicatorGetAll executes the `client p2p replicator set` command.
```
shahzadlone: todo: same as above

cli/test/integration/p2p/connect_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	test.Execute(t)
+}
+
+// NOTE: This test currently fails because there is no peer listening at the given address.
+// However, it does at least verify that multiple addresses can be passed in.
+//
+// TODO: Add capability to have multiple defradb instances in tests, so we can
+// actually test successful connections. #4021
```
shahzadlone: suggestion: Fix todo issue number like the other comment

internal/db/p2p/replicator.go
```diff
@@ -99,36 +94,71 @@ func (p *P2P) SetReplicator(ctx context.Context, repInfo client.PeerInfo, collec
...
 
-	addedCols := []client.Collection{}
-	for _, col := range fetchedCollections {
-		if _, ok := storedCollectionIDs[col.CollectionID()]; !ok {
-			storedCollectionIDs[col.CollectionID()] = struct{}{}
-			addedCols = append(addedCols, col)
-			storedRep.CollectionIDs = append(storedRep.CollectionIDs, col.CollectionID())
+	// Update the list of collections for each replicator prior to persisting.
+	storedRepCollectionIDs := make(map[string]map[string]struct{}) // replicatorID => collectionID
+	addedCols := make(map[string][]client.Collection)              // peerID => list of collections added
```
shahzadlone, line 100: question: What does this comment mean?  `peerID => list of collections added`

cli/test/integration/p2p/connect_test.go
fredcarle: > question: What actually broke when you added multiple? I thought it would work and was already used (it is not, I must have mixed memories with Lens).

ah ok. So the problem is that technically we could call start multiple times and it would create instances on different ports but the test state is made in a way that only one instance is accounted for at a time. So the test framework needs to be updated to handle multiple instances at once if we want to test p2p related things.

internal/db/p2p/replicator.go
fredcarle: it means its a map of peerIDs to the list of collections added.

cli/test/integration/p2p/connect_test.go
fredcarle: done

cli/test/integration/p2p/connect_test.go
fredcarle: done

cli/test/action/p2p_replicator_delete.go
fredcarle: done

cli/test/action/p2p_replicator_getall.go
fredcarle: done

cli/test/integration/p2p/connect_test.go
AndrewSisley: Nice :) That is an easier problem to sort out when we get the time, thanks Fred

Issue #4054, opened 2025-10-06
Update Defra to latest Lens version


AndrewSisley:
Defra needs Lens store and p2p.

AndrewSisley:
wip branch 4054-lens-09

AndrewSisley:
todo: If the LensRegistry survives - https://github.com/sourcenetwork/defradb/pull/4062#discussion_r2411170973

PR #4056, opened 2025-10-06
feat: Remove unused context from Txn funcs

### LINKED ISSUE #4055: Remove unused context from Txn funcs
It is never used, and it complicates a large number of calls, including in downstream projects.

AndrewSisley:
## Relevant issue(s)

Resolves #4055

## Description

Removes the unused context parameter from `NewTxn`, `NewConcurrentTxn`, `Commit` and `Discard`.

It has already been removed from Lens, as it was quite an annoying and unwanted complication there.  The `New` funcs also made it look like the txn was scoped to the context.

Broken out of https://github.com/sourcenetwork/defradb/issues/4054 as this is a breaking API change and deserves its own commit.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -49,7 +49,7 @@ func NewClient(rawURL string) (*Client, error) {
 	return &Client{httpClient}, nil
 }
 
-func (c *Client) NewTxn(ctx context.Context, readOnly bool) (client.Txn, error) {
+func (c *Client) NewTxn(readOnly bool) (client.Txn, error) {
 	query := url.Values{}
 	if readOnly {
 		query.Add("read_only", "true")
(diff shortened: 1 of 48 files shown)


shahzadlone (approved):
Easy review LGTM

http/client.go
```diff
@@ -58,7 +58,7 @@ func (c *Client) NewTxn(ctx context.Context, readOnly bool) (client.Txn, error)
 	methodURL := c.http.apiURL.JoinPath("tx")
 	methodURL.RawQuery = query.Encode()
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, methodURL.String(), nil)
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodPost, methodURL.String(), nil)
```
fredcarle, line 61: thought: You merged this already but this is a bad idea as it no longer provided a means for the caller to set a timeout out to otherwise cancel the request that could hang indefinitely.

AndrewSisley: Maybe we can add a http-only overload for this, I really don't think the internal and embedded code, and anything built on it should have to worry about context here.

fredcarle: > I really don't think the internal and embedded code, and anything built on it should have to worry about context here.

I agree. But not having a better solution for it yet was the reason I didn't previously remove the context from those methods. Would love to find a better alternative.

PR #4061, opened 2025-10-06
fix: Allow docs to contain nillable arrays of nillable types

### LINKED ISSUE #2950: Order by aggregate values
```gql
type User {
  name: String
  books: Int
  dvds: Int
  posters: Int
}
```
It would be useful for example to aggregate the stats and order by that aggregated value.
```gql
query {
   User(order: {total: DESC}) {
    total: _sum(books, dvds, posters)
  }
}
```

Given GQL specs, we may need to do something like this for GQL client to support it:
```gql
query {
   User(order: {_alias("total"): DESC}) {
    total: _sum(books, dvds, posters)
  }
}
```

### LINKED ISSUE #4059: Fix support for docs containing arrays of nillable types
The existing version of the `toMap` function on line 870 of `client/document.go` handles sub-documents being present recursively, but for all other cases, it sets the doc map to `value.Value()`. This breaks down in the case of a nillable array being present, and it is possible for two different documents hashing to the same doc ID, and the creation of the second document not being possible. This must be fixed.

To illustrate this bug, suppose we create a schema:

```
defradb client schema add '
  type User {
    books: [Int]
    movies: [Int]
    games: [Int]
  }
'
```

If we try to add the following two documents, only the first will be added:

```
defradb client query '
  mutation {
      create_User(input: {books:[1,2,3], movies:[4,5,6], games:[7,8,9]}) {
          _docID
      }
  }
'

defradb client query '
  mutation {
      create_User(input: {books:[10,11,12], movies:[13,14,15], games:[16,17,18]}) {
          _docID
      }
  }
'
```

An error will be returned saying that a document with the given docID already exists.

ChrisBQu:
## Relevant issue(s)

Resolves #4059
Resolves #2950

## Description

For a detailed explanation and example of the problem at hand, see: https://github.com/sourcenetwork/defradb/issues/4059

To address the issue:

In `client/value.go` there is a function called `Bytes()` which contains pre-existing logic for properly handling arrrays of nillable types. This logic has been duplicated into the `toMap` function in `client/document.go`, cleanly handling the issue.

In addressing this issue, we actually find that another outstanding issue is also addressed. We wanted to be able to execute queries that order by aggregate fields. With the previously mentioned fix, this will just work *on arrays*. To illustrate this, some integration tests have been written and included with this PR. 

**This is a breaking change. I document this in `i4061-nillable-aray-docs.md`**

Because the logic of the `toMap` function has changed, existing integration tests that explicitly utilized doc IDs, and which had schemas containing arrays of nillable types, needed to be adjusted to use new doc IDs. There were also tests which expected documents in a certain default order (that is, it did not explicitly order the results), which needed to be adjusted because of a side-effect of the fix described above.

The following tests had to change because of hardcoded doc IDs:

`TestQueryInlineNillableFloatArray_WithDocIDMaxAndPopulatedArray_Succeeds`
`TestQueryInlineNillableFloatArray_WithDocIDAndMin_Succeeds`
`TestQueryInlineNillableFloatArray_WithDocIDAndMax_Succeeds`
`TestQueryInlineNillableFloatArray_WithDocIDMaxAndPopulatedArray_Succeeds`

The following tests had to change the expected order of its result:

`TestQueryInlineArrayWithGroupByString`
`TestQueryInlineArrayWithGroupByArray`

And the structure of the following tests' `ContainsData` block was also reworked:

`TestSchemaAggregateSimpleCreatesUsersAverage`
`TestSchemaAggregateSimpleCreatesUsersSum`
`TestSchemaAggregateSimpleCreatesUsersCount`
`TestSchemaAggregateInlineArrayCreatesUsersSum`
`TestSchemaAggregateInlineArrayCreatesUsersAverage`

**Ordering aggregate fields of scalar values**

As mentioned, the previous work allows the aggregate functions (_sum, _count, _min, _max, _avg) to be performed, and for the query to be ordered in ascending or descending value. However, leaving the code like this would result in an issue that was only half-solved. We also want these aggregate functions, and ordering, to work on scalar values. So additional work was done to make this work. Specifically, the `getNumericFields` function of `internal/request/graphql/schema/generate.go` file was modified. Then, the `Next` function of the following functions were modified as well: `internal/planner/average.go`, `internal/planner/max.go`, `internal/planner/min.go`, `internal/planner/sum.go`. Each of these were changed so as to catch the cases of scalar values, and use them. This PR contains tests to illustrate this functionality.

**DO NOT MERGE. TESTS ARE STILL BEING WORKED ON**

## Tasks

## How has this been tested?

- WSL


client/document.go
```diff
@@ -879,16 +881,39 @@ func (doc *Document) toMap(excludeEmpty bool) (map[string]any, error) {
...
+		if v, ok := normValue.NillableStringArray(); ok {
+			innerValue = convertImmutable(v)
+		} else if v, ok := normValue.NillableIntArray(); ok {
+			innerValue = convertImmutable(v)
+		} else if v, ok := normValue.NillableFloat64Array(); ok {
+			innerValue = convertImmutable(v)
+		} else if v, ok := normValue.NillableFloat32Array(); ok {
+			innerValue = convertImmutable(v)
+		} else if v, ok := normValue.NillableBoolArray(); ok {
+			innerValue = convertImmutable(v)
```
islamaliev, line 910: suggestion: why not to use 
```go
if normValue.IsArray() && normValue.IsNillable() {
```

EDIT: If you really meant array of nillable types, ignore this suggestion. 
Well, in theory we could add another method to `NormalValue` interface to facilitate that.
I think I still have a git branch or stash with work that I did to add something like `SubType` that would return array's element type. It's out of scope anyway.

client/field_kind.go
```diff
@@ -304,6 +304,7 @@ const (
 	FieldKind_NILLABLE_FLOAT64_ARRAY ScalarArrayKind = 20
 	FieldKind_NILLABLE_STRING_ARRAY  ScalarArrayKind = 21
 	FieldKind_NILLABLE_FLOAT32_ARRAY ScalarArrayKind = 22
+	// TO DO: Add nillable array types. See: https://github.com/sourcenetwork/defradb/issues/4060
```
islamaliev: nitpick: please join "TO" and "DO" because this is what devs search for.

docs/data_format_changes/i4061-nillable-aray-docs.md
islamaliev, line 1: todo: this file looks empty to me. Please add some text

tests/integration/query/inline_array/utils.go
```diff
@@ -28,6 +28,9 @@ var userCollectionGQLSchema = (`
 		pageRatings: [Float]
 		preferredStrings: [String!]
 		pageHeaders: [String]
+		books: [Int]
+		movies: [Int]
+		games: [Int]
```
islamaliev: todo: why do you need this ones here? You already have `testScores` of type `[Int]`.

And the tests you added do not use the new set of fields with the old ones. It makes sense to create another collection with the new fields if you really need them.

client/document.go
```diff
@@ -879,16 +881,39 @@ func (doc *Document) toMap(excludeEmpty bool) (map[string]any, error) {
...
 		}
 
-		docMap[k] = value.Value()
+		normValue := value.NormalValue()
+		if normValue.IsNil() {
+			docMap[k] = nil
+			continue
+		}
+
+		// In the case of nillable arrays, we need to convert to the underlying value.
```
islamaliev: todo: we have 3 types of nillables when it comes to arrays:
- nillable array
- array of nillables
- nillable array of nillables

Method you use below like `normValue.NillableStringArray` is array of nillable bools, not nillable array, but the comment states it is a nillable array.

In the PR description you say "arrrays of nillable types". So please make it clear. And please check other place that might be confusing.

client/document.go
```diff
@@ -865,6 +865,8 @@ func (doc *Document) Clean() {
 //
 // If `true` is provided, properties with nil values will be ommited from
 // the result.
+// Assumes in same package, so we can access field.Value() or the unexported field.
+// If value is exposed via method, use that method instead.
```
islamaliev: todo: sorry, but I don't understand this comment. Please rephrase

islamaliev (changes requested):
Look good overall. I left some minor requests.

client/document.go
fredcarle: We currently use, for example,  `FieldKind_NILLABLE_BOOL_ARRAY` for what is actually a nillable bool nillable array (`[Bool]`). We currently don't support nillable bool arrays (`[BOOL]!`) so the mistake in field kind naming was not a high priority to change if you remember.

fredcarle: Just to clarify from your comment, we do not currently support `array of nillables`.

client/document.go
fredcarle: We can't just use `if normValue.IsArray() && normValue.IsNillable() {` because the type needs to be known.

fredcarle: We do the same with `FieldValue.Bytes()`

client/document.go
fredcarle: To be clearer the PR title could be `Allow docs to contain nillable arrays of nillable types`

client/document.go
ChrisBQu: Good catch. This looks like a left over portion of a comment from a helper function that I was using in an earlier version of the code. I think no need to rewrite it, I'll just remove it.

client/document.go
ChrisBQu: It is as Fred says. I don't want to mark this as resolved if you still have concerns though. Please let me know if you are comfortable with the given approach.

islamaliev: I said "If you really meant array of nillable types, ignore this suggestion" :) 
So just ignore it

islamaliev (approved):

shahzadlone (approved):
LGTM

PR #4062, opened 2025-10-07
refactor: Limit previous collection versions to one or zero

### LINKED ISSUE #4057: Compile-time limit collection VersionSources to zero or one source
It is currently an array, with a runtime limit of a single or zero and some documentation.  This makes it confusing, and quite a lot harder to build off of.

Collection versions now form a DAG - use it.  Currently unsure if the prior version id needs to be on collection version, or we can get away with keeping it internal (in the actual DAG).  It is probably nicer to have the previous version on CollectionVersion, even if it duplicates data a little.

Broken out of https://github.com/sourcenetwork/defradb/issues/4054

AndrewSisley:
## Relevant issue(s)

Resolves #4057

## Description

Limit previous collection versions to one or zero at compile time.

Previously the model supported multiple sources, but was temporarily restricted at runtime for short-term simplification.  Particularly now that the collection is represented by a DAG in the blockstore, this is an unwanted complication that makes a lot of things harder.


cli/test/action/collection_describe.go
```diff
@@ -64,10 +64,22 @@ func (a *CollectionDescribe) Execute() {
...
-			require.Equal(a.s.T, expected.VersionSources, actual.VersionSources)
+		require.Equal(a.s.T, expected.PreviousVersion.HasValue(), actual.PreviousVersion.HasValue())
+		if expected.PreviousVersion.HasValue() {
+			require.Equal(
+				a.s.T,
+				expected.PreviousVersion.Value().SourceCollectionID,
+				actual.PreviousVersion.Value().SourceCollectionID,
+			)
+
+			if expected.PreviousVersion.Value().Transform.HasValue() {
```
shahzadlone, line 80: suggestion: Shouldn't you also check if `actual.PreviousVersion.Value().Transform.HasValue()`? before doing the `require.Equal`?`

internal/db/collection_id.go
```diff
@@ -483,29 +483,13 @@ func saveBlocks(
...
-					break
-				}
-			}
+		if oldCol.VersionID != "" {
+			collection.PreviousVersion = immutable.Some(
+				client.CollectionSource{
+					SourceCollectionID: oldCol.VersionID,
+					Transform:          migration,
+				},
+			)
```
shahzadlone, line 492: praise: Much simpler logic now

cli/test/action/collection_describe.go
AndrewSisley: Nice spot! That's never actually asserted... I'll fix this, and the 2 other places that have the same problem.

- [x] Assert HasValue

internal/db/definition_validation.go
```diff
@@ -430,21 +430,15 @@ func validateSourcesNotRedefined(
 			continue
 		}
 
-		if len(newCol.VersionSources) != len(oldCol.VersionSources) {
+		if newCol.PreviousVersion.HasValue() != oldCol.PreviousVersion.HasValue() {
 			errs = append(errs, NewErrCollectionSourcesCannotBeAddedRemoved(newCol.VersionID))
```
shahzadlone, line 434: question: If there is an error why don't we return rather than keep appending errors and continuing?

internal/db/definition_validation.go
```diff
@@ -430,21 +430,15 @@ func validateSourcesNotRedefined(
...
-					oldCol.VersionSources[i].SourceCollectionID,
-				))
-			}
+		} else if newCol.PreviousVersion.HasValue() &&
+			newCol.PreviousVersion.Value().SourceCollectionID != oldCol.PreviousVersion.Value().SourceCollectionID {
+			errs = append(errs, NewErrCollectionSourceIDMutated(
+				newCol.VersionID,
+				newCol.PreviousVersion.Value().SourceCollectionID,
+				oldCol.PreviousVersion.Value().SourceCollectionID,
+			))
```
shahzadlone, line 441: question: Same question related to error appending here.

internal/lens/fetcher.go
```diff
@@ -96,14 +96,11 @@ func (f *lensedFetcher) Init(
 	f.lens = new(ctx, f.registry, f.col.Version().VersionID, history)
 	f.txn = txn
 
-historyLoop:
```
shahzadlone, line 99: praise: Nice to see this gone haha

internal/lens/registry.go
```diff
@@ -56,19 +56,19 @@ func (r *LensRegistry) ReloadLenses(ctx context.Context) error {
...
 
 	for _, col := range cols {
-		if len(col.VersionSources) == 0 {
+		if !col.PreviousVersion.HasValue() {
 			continue
 		}
 
 		// WARNING: Here we are only dealing with the first source in the set, this is fine for now as
 		// currently collections can only have one source, however this code will need to change if/when
 		// collections support multiple sources.
```
shahzadlone, line 65: suggestion: Update/remove comment, as it suggests that we can have more than the "first source in the set"?

tests/integration/collection_version/get_schema_test.go
```diff
@@ -107,6 +107,9 @@ func TestGetSchema_ReturnsAllSchema(t *testing.T) {
 						Name:           "Users",
 						IsActive:       false,
 						IsMaterialized: true,
+						PreviousVersion: immutable.Some(client.CollectionSource{
+							SourceCollectionID: "bafyreihdbjfazsx5vq2tpzedqdktrjyn6lq22qle7el2s42b3q4zpxmwqq",
```
shahzadlone, line 111: question/nitpick: Is there a way to use indexing like we do for other actions to avoid having to maintain the actual cids in the tests?

internal/db/definition_validation.go
AndrewSisley: The validation rules were changed to do this a little while ago, it lets users receive all the errors at once, instead of  drip feeding them one at a time as they fix them

internal/lens/registry.go
AndrewSisley: This entire file is deleted in my wip branch, so I'm going to leave this as-is.  I'll add a link to this discussion to that issue, to try and stop me from forgetting in case I end up reverting that deletion.

AndrewSisley: https://github.com/sourcenetwork/defradb/issues/4054#issuecomment-3377610383

tests/integration/collection_version/get_schema_test.go
AndrewSisley: We can add it, I just don't want to spend time on that right now

internal/lens/registry.go
shahzadlone: Thanks for the clarification! Happy with that

tests/integration/collection_version/get_schema_test.go
shahzadlone: Cool perhaps make an issue for a follow-up later.

shahzadlone (approved):
Thanks for answering all the questions and adding the missing asserts :)

LGTM

tests/integration/collection_version/get_schema_test.go
AndrewSisley: I'd rather not, that feels like a low priority use of the backlog - it'll get done when it is annoying enough :grin:

PR #4064, opened 2025-10-09
fix: Default json value

### LINKED ISSUE #3146: JSON object default value causes panic

islamaliev:
## Relevant issue(s)

Resolves #3146

## Description

Fix default json value by properly converting it's value into a corresponding type.

### DIFF
--- a/internal/request/graphql/schema/collection.go
+++ b/internal/request/graphql/schema/collection.go
@@ -11,6 +11,7 @@
 package schema
 
 import (
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


ChrisBQu (approved):
LGTM.

internal/request/graphql/schema/collection.go
```diff
@@ -434,14 +435,28 @@ func defaultFromAST(
...
+			value = string(jsonBytes)
+		}
 	case types.DefaultDirectivePropBlob:
 		value = types.Blob.ParseLiteral(arg.Value, nil)
 	}
 	// If the value is nil, then parsing has failed, or a nil value was provided.
 	// Since setting a default value to nil is the same as not providing one,
 	// it is safer to return an error to let the user know something is wrong.
-	if value == nil {
+	// Exception: JSON fields can have nil as a valid default value
```
fredcarle, line 458: question: Why do we treat the JSON type differently?

PR #4069, opened 2025-10-13
fix: Forbid default null value for json field type

### LINKED ISSUE #4068: Disable null default json value
Only for top level json.

islamaliev:
## Relevant issue(s)

Resolves #4068

## Description

This change forbids setting `null` as default value for json field type (only top level) so that it behaves as all other field types.


AndrewSisley (approved):
LGTM, thanks Islam :)

fredcarle (commented):
Thanks for fixing this.

fredcarle (approved):

PR #4070, opened 2025-10-14
feat: Remove LensRegistry

### LINKED ISSUE #3889: Lens repository/registry looks like it respects transactions but does not
Either document it heavily (short-term), or handle transactions.  Most/all(?) of the txn code already exists in the Lens repo, hopefully just needs to be hooked up and tested.

### LINKED ISSUE #4054: Update Defra to latest Lens version

AndrewSisley:
## Relevant issue(s)

Resolves #4054
Resolves #3889

## Description

Removes the LensRegistry, and updates to Lens v0.9.  This also reworks the Lens source properties on `client.CollectionVersion` so they now hold the Lens ID instead of the full configuration.

Long term we'll almost certainly want to expose the lens store `Add` and `List` functions currently on `db.DB`, but that can be done later IMO (argue back if need you want :)).


internal/db/lens.go
shahzadlone, line 1: Suggestion: Perhaps consider all the public functions with warning comments to first begin starting with the go style functions name with a quick one liner description of the function, followed by the warning comment, in the next line.

Moreover, there is a linter I want to enable in the future to detect typos/mismatched function name vs function documentation leading word (which will complain)

internal/db/config.go
```diff
@@ -32,6 +34,8 @@ type dbOptions struct {
 	retryIntervals []time.Duration
 	// timeout duration for syncing block links.
 	p2pBlockSyncTimeout time.Duration
+	LensRuntimeType     LensRuntimeType
```
shahzadlone, line 38: Question: these two need to be public?

AndrewSisley: I think all options should be public, they *are* set by the user, and it is good to let them check them.  All of Lens' are public.

They can be made private if there is a strong demand for consistency here though.

shahzadlone: Was just a bit odd seeing some private and some public.

todo: Just want consistency either way you like.

internal/db/lens.go
AndrewSisley: Adding a linter will likely strengthen my dislike for that Go convention...

This might be a good example as to why it would be a bad linter to add - it is going to cost dev time to places that dont benefit from it, similar to the unchecked casting one (which I curse on a semi-regular basis).

AndrewSisley: Now I want to rant about that :laughing: 

Casts are unsafe by default.  There is no linter required to flag that.  Sometimes, in their given context, they are safe, and documentation should be provided.  The linter does not care about the documentation, and forces either code-complexity, or a waste of eye-space `//nolint` comment (and the time it takes to find the exact syntax and rule name to disable).

I hate it and want it removed.

shahzadlone: Haha, I think I agree to most of what you said.

One of the reasons for having that linter is that we have a lot of typos currently, and there have been 3-4 external contributions IIRC by people just openning PRs fixing these.

http/handler_store.go
```diff
@@ -141,12 +141,13 @@ func (h *storeHandler) SetMigration(rw http.ResponseWriter, req *http.Request) {
...
 
-	err := db.SetMigration(req.Context(), cfg)
+	lensID, err := db.SetMigration(req.Context(), cfg)
 	if err != nil {
 		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
 		return
 	}
-	rw.WriteHeader(http.StatusOK)
+
+	responseJSON(rw, http.StatusOK, lensID)
```
fredcarle: thought: Just the `lensID` in the payload is a little weird when returning an `applicaton/json` content type. Might be worth returning an object that is more descriptive: `{"lensID": "1234"}`.

internal/db/collection_define.go
```diff
@@ -286,33 +286,14 @@ func (db *DB) patchCollection(
...
 		}
 
-		if col.Query.HasValue() && col.Query.Value().Transform.HasValue() {
-			err = db.LensRegistry().SetMigration(ctx, col.VersionID, col.Query.Value().Transform.Value())
+		if col.PreviousVersion.HasValue() && migration.HasValue() {
+			_, err = db.setMigration(ctx, client.LensConfig{
+				SourceSchemaVersionID:      col.PreviousVersion.Value().SourceCollectionID,
+				DestinationSchemaVersionID: col.VersionID,
+				Lens:                       migration.Value(),
+			})
```
fredcarle, line 296: praise: That is much nicer that what we had previously.

internal/db/db.go
```diff
@@ -162,9 +164,23 @@ func newDB(
...
-	if lens != nil {
-		lens.Init(db)
+	lensRuntime, err := newLensRuntime(opts.LensRuntimeType)
+	if err != nil {
+		return nil, err
+	}
+
+	// Overwrite a few key Lens options for now, by appending them to the end of the option
+	// slice.
+	opts.LensOptions = append(opts.LensOptions, lensNode.WithP2PDisabled(true))
```
fredcarle, line 174: thought: I'm not a fan of extra functionality having to be disabled because they are on by default. I don't think Lens should have P2P on by default. Users using Lens directly shouldn't have the side effect of P2P without explicitly opting in.

internal/db/db.go
```diff
@@ -162,9 +164,23 @@ func newDB(
...
+	opts.LensOptions = append(opts.LensOptions, lensNode.WithRootstore(rootstore))
```
fredcarle, line 175: todo: I don't think its safe to pass the rootstore to an external package (even if it's our own). There are too many risks for mistakes that could wipe the whole rootstore. A namespace should be applied here before passing it to Lens.

internal/db/lens.go
```diff
@@ -13,22 +13,68 @@ package db
...
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/description"
 )
 
-func (db *DB) setMigration(ctx context.Context, cfg client.LensConfig) error {
+// WARNING - This function does not create a txn ctx if one does not exist
+// and so should not be exposed to users.
+func (db *DB) AddLens(ctx context.Context, cfg model.Lens) (cid.Cid, error) {
```
fredcarle: todo: This function is never used outside of this package. Instead of a warning not to expose it to users, you should simply make it private.

internal/db/lens.go
```diff
@@ -13,22 +13,68 @@ package db
...
+	ctx context.Context,
+	source enumerable.Enumerable[map[string]any],
+	id string,
+) (enumerable.Enumerable[map[string]any], error) {
+	return db.GetLensStore(ctx).Inverse(ctx, source, id)
+}
+
+// WARNING - This function does not create a txn ctx if one does not exist
+// and so should not be exposed to users.
+func (db *DB) GetLensStore(ctx context.Context) store.Store {
```
fredcarle: todo: Please make this private as it's only used in this package.

internal/db/lens.go
```diff
@@ -13,22 +13,68 @@ package db
...
+	ctx context.Context,
+	source enumerable.Enumerable[map[string]any],
+	id string,
+) (enumerable.Enumerable[map[string]any], error) {
+	return db.GetLensStore(ctx).Transform(ctx, source, id)
+}
+
+// WARNING - This function does not create a txn ctx if one does not exist
+// and so should not be exposed to users.
+func (db *DB) Inverse(
```
fredcarle: question: These two seem to never be used. Why do we even need them?

tests/integration/db.go
```diff
@@ -69,14 +71,16 @@ func init() {
 
 func defaultNodeOpts() []node.Option {
 	return []node.Option{
-		node.WithLensPoolSize(lensPoolSize),
+		db.WithLensOpts(
+			lensNode.WithPoolSize(lensPoolSize),
+		),
+		db.WithLensRuntime(lensType),
```
fredcarle, line 77: question: Why is this option not part of the `WithLensOpts`?

tests/integration/replace.go
```diff
@@ -78,3 +86,30 @@ func replace(s *state.State, nodeId int, input string) string {
 
 	return buf.String()
 }
+
+func replaceMap(s *state.State, nodeId int, inputSet []string) map[string]string {
+	templateData := map[string]string{}
```
fredcarle, line 91: question: Why not return early like `replace` if there is no `{{`?

fredcarle (changes requested):
I really like this change overall. Just a few things I'd like resolve before approval.

http/handler_store.go
AndrewSisley: I understand what you mean, it is still valid json though, and atm we never seem to return `text/foo` from any http client methods (making this an outlier if we change the Mime type).

Returning an object here feels a bit odd too, as that is inconsistent with the embedded api.

Going to leave as-is for now as I don't feel enthusiastic about any of the alternatives right now.  It might also be that we have a lot of hidden inconsistencies in the http API similar to this and we need to adjust a few of them (v1?).

internal/db/db.go
AndrewSisley: This is what `lensNode` is for.  P2P is the only reason you'd use it over `store`.  We use it because we are going to use lens P2P in the next few days.

IIRC Defra follows this pattern too, and the feeling I got from a recent meeting was that the dev-ex guys like this approach (node serving the everything on/ready by default, lower level libs largely off by default).

internal/db/db.go
AndrewSisley: This was driven by us wanting to use the same blockstore for Lens.  I have been very much starting to lean away from that recently, but I thought you were in favour of that (you spoke about handling the Lens block sync in the Defra code using links long term, instead of letting Lens sync those blocks).  Given that the blockstore is an immediate child of root, and lens spins up another prefix, it must be given root.

internal/db/lens.go
AndrewSisley: Ah sweet, I accidentally removed all external references to these!  Thanks for flagging, a few of them can even bee deleted.

- [x] Make private/delete funcs

internal/db/lens.go
AndrewSisley: Handling against comment on `AddLens`

internal/db/lens.go
AndrewSisley: Handling against comment on `AddLens`

tests/integration/db.go
AndrewSisley: The lens repo packages take an actual Lens runtime instance, the db option takes an aliased string.

tests/integration/replace.go
AndrewSisley: Because I didn't think to do so at the time :)  I'll add it.

- [x] Add early return for `replaceMap`

AndrewSisley: Actually it doesn't really make sense here, as this function takes a set, and looping through each one and adding an extra check may not be more performant overall than just trying to do it like it currently does.

http/handler_store.go
AndrewSisley: Discussed on a call, I will have a look at NewTxn/ID() result.

- [x] Revisit

internal/db/db.go
AndrewSisley: Discussed on a call, code will remain as is, and is acceptable for v1, but https://github.com/sourcenetwork/lens/issues/139 has been opened and should be sorted out once the collection-p2p feature is done.

fredcarle (approved):
Thanks for the changes and the conversation earlier.

PR #4071, opened 2025-10-14
fix: Version query with delete

### LINKED ISSUE #4008: subscription for deletes doesn't work
Example schema:

```graphql
type User {
  name: String
  email: String @index(unique: true)
  counter: Int @crdt(type: pcounter)
  createdAt: String
}
```

Then this subscription (via SSE)...

```graphql
subscription {
   User (showDeleted: true) { .. }
}
```

...then this mutation:

```graphql
mutation {
  delete_User(filter:{ email:{ _eq:"whatever@whatever.whatever" } }){ _docID }
}
```

successfully deletes:

```json
{
  "data": {
    "delete_User": [
      {
        "_docID": "bae-9e371ee2-d366-5ea1-b69d-4f6232ad0a43"
      }
    ]
  }
}
```

...but doesn't actually fire on the SSE subscription.

fredcarle:
## Relevant issue(s)

Resolves #4008

## Description

This PR fixed an issue where doing a version query on a document that had been deleted cause the query to hang. This also appeared in subscriptions since subscriptions use a version query to return the appropriate document state.

The problem stems from the version fetcher using the in-memory store and having a btree implementation that doesn't allow modifying the btree at the same time as iterating over it. There is a way that this could be done by modifying the in-memory store implementation but this PR provides a good enough fix for the short term.

## Tasks

## How has this been tested?

Add two new integration tests.

- MacOS


AndrewSisley (approved):
LGTM, hope that wasn't too painful to track down...

internal/core/crdt/composite.go
```diff
@@ -184,23 +191,36 @@ func (m DocComposite) deleteWithPrefix(ctx context.Context, key keys.DataStoreKe
 			return errors.Join(err, iter.Close())
 		}
 
+		var value []byte
```
getify: question: in Go, does this allocate an array (that is GC'd) if the subsequent `if err != nil` block early returns?

internal/core/crdt/composite.go
```diff
@@ -184,23 +191,36 @@ func (m DocComposite) deleteWithPrefix(ctx context.Context, key keys.DataStoreKe
...
 		if dsKey.InstanceType == keys.ValueKey {
-			value, err := iter.Value()
+			value, err = iter.Value()
 			if err != nil {
 				return errors.Join(err, iter.Close())
 			}
+		}
+		kvArray = append(kvArray, kv{
```
getify, line 198: what is the intended significance of adding an entry to `kvArray` here if `value` is the empty `[]byte` (or `nil`?)? does it represent an entry that should be *deleted* instead of renamed?

like, in what cases will `if dsKey.InstanceType == keys.ValueKey` not be true?

if so, a code comment here would make that more obvious.

internal/core/crdt/composite.go
```diff
@@ -184,23 +191,36 @@ func (m DocComposite) deleteWithPrefix(ctx context.Context, key keys.DataStoreKe
 			return errors.Join(err, iter.Close())
 		}
 
+		var value []byte
 		if dsKey.InstanceType == keys.ValueKey {
-			value, err := iter.Value()
```
getify, line 187: what is the significance of removing `:` here?

I assume it's because `value` is already declared... but `err` is not yet?

internal/core/crdt/composite.go
```diff
@@ -184,23 +191,36 @@ func (m DocComposite) deleteWithPrefix(ctx context.Context, key keys.DataStoreKe
...
+	err = iter.Close()
+	if err != nil {
+		return err
+	}
+
+	for _, item := range kvArray {
+		if item.value != nil {
+			err = m.store.Set(ctx, item.key.WithDeletedFlag().Bytes(), item.value)
 			if err != nil {
 				return errors.Join(err, iter.Close())
```
getify: `iter.Close()` has already been called (line 207) before this loop... we don't need to close it again do we?

internal/core/crdt/composite.go
```diff
@@ -184,23 +191,36 @@ func (m DocComposite) deleteWithPrefix(ctx context.Context, key keys.DataStoreKe
...
 			}
 		}
-
-		err = m.store.Delete(ctx, dsKey.Bytes())
+		err = m.store.Delete(ctx, item.key.Bytes())
 		if err != nil {
 			return errors.Join(err, iter.Close())
```
getify: same question about `iter.Close()` as above.

tests/integration/query/simple/with_cid_test.go
```diff
@@ -159,3 +159,46 @@ func TestQuerySimple_WithCIDAndCounterAfterUpdate_ShouldSucceed(t *testing.T) {
...
+					}
+				`,
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"name": "John"
+				}`,
+			},
+			testUtils.DeleteDoc{
+				DocID: 0,
```
getify, line 179: nit: I know we talked about this, but it's still confusing to me that these are *indexes* not actual "doc ID" values.

getify (commented):
seems like a reasonable solution to the issue.

have we considered "in the real world" how big this `kvArray` could build up to? could there be a case where there's a significant performance hit from running through the iterator fully once, collecting all the pending deletes, and then running through that list of pending deletes a second time?

would/could it reasonably be hundreds, thousands, or even millions? since it's for every field of every document, just wondering if it might explode (memory or cpu) as compared to being "lazy" interleaved with the iterator.

fredcarle:
> have we considered "in the real world" how big this `kvArray` could build up to? could there be a case where there's a significant performance hit from running through the iterator fully once, collecting all the pending deletes, and then running through that list of pending deletes a second time?
> 
> would/could it reasonably be hundreds, thousands, or even millions? since it's for every field of every document, just wondering if it might explode (memory or cpu) as compared to being "lazy" interleaved with the iterator.

The impact would we proportional strictly to the number of fields on a doc. If one has hundreds, thousands or millions of fields on a doc, we have other problems :)

internal/core/crdt/composite.go
fredcarle: Yes

internal/core/crdt/composite.go
fredcarle: I wanted to assigned to `value` that was created on like 194 so removing `:` prevents from allocating to a new variable. `err` was created above as well so I didn't need to recreated with `var err error` before the `iter.Value` call.

internal/core/crdt/composite.go
fredcarle: Good catch. I'll remove.

internal/core/crdt/composite.go
fredcarle: I'll remove.

tests/integration/query/simple/with_cid_test.go
fredcarle: I agree. We should probably change that at some point.

internal/core/crdt/composite.go
getify: I know it's only a nit, but... I prefer not to allocate things that I am not sure I'm going to use. I wonder if there are idioms in go here that can defer the allocation until we're assigning to it? in JS we might use a ternary in such a scenario.

internal/core/crdt/composite.go
fredcarle: > what is the intended significance of adding an entry to kvArray here if value is the empty []byte (or nil?)? does it represent an entry that should be deleted instead of renamed?

Everything is deleted, if there is a value, a new kv is created to represent the deleted field.

> like, in what cases will if dsKey.InstanceType == keys.ValueKey not be true?

To be honest, I can't remember why we have to check that. I did a few test and it seems to always be a value key. I removed the check.

getify (approved):
LGTM! thanks :)

internal/core/crdt/composite.go
fredcarle: Sorry spoke too fast here. I didn't think this through properly.
```go
var b []byte
```
creates a `nil` slice which require no allocation. Allocation only happens when subsequently assigning to it.
```go
b := []byte{}
```
On the other hand would allocate a zero-length backing array. 

So no that line (no longer present in latest version) did not allocate memory.

client/request/select.go
```diff
@@ -107,6 +109,16 @@ func (s *Select) validateGroupBy() []error {
 	return result
 }
 
+func (s *Select) ToSubscriptionSelect(docID, cid string) Selection {
```
AndrewSisley, line 116: todo: All the other `ToFoo` funcs clone the source object - this is mutating it - it could be quite an unpleasent side-effect to debug.  Please return a new, clean object instead of the old one.

AndrewSisley (approved):
LGTM - thanks for simplifying the subscription stuff!

Please sort out the one todo before merge though.

client/request/select.go
fredcarle: I think you missed the last commit :)

client/request/subscription.go
```diff
@@ -1,47 +0,0 @@
...
-		DocIDsFilter: DocIDsFilter{
-			DocIDs: immutable.Some([]string{docID}),
-		},
-		CIDFilter: CIDFilter{
-			immutable.Some(cid),
-		},
-		ChildSelect: m.ChildSelect,
-		Filterable:  m.Filterable,
-	}
-}
```
jsimnz, line 47: question: one of the reasons we had a separate dedicated type is because there are some args the don't make sense in the context of the subscription (the way theyre currently implemented).

Eg: Limiting and offset

Since the subscriptions do a point lookup on both DocID and CID (eg: `User(docID: X, cid: Y) { ... }`). Offset and Limit have no effect.

We don't currently, but we should prob update the `Subscription` type for the collection queries to not expose these as well.

(This decision also affects the change further down replacing `parseSubscriptionOperationDefinition` with  `parseQueryOperationDefinition`

fredcarle: > We don't currently, but we should prob update the Subscription type for the collection queries to not expose these as well.

I think that's what matters most in this conversation. I agree that breath of the input parameters needs to be limited to what is supported/makes sense for subscriptions. The underlying logic to execute the query however doesn't need to be duplicated. I think it's much cleaner to use the same underlying logic. It has the benefit of automatically making available to subscriptions new features we add to queries if it makes sense to do so.

getify:
🙏🎉💙

PR #4075, opened 2025-10-16
feat: Gate collection get op with NAC

### LINKED ISSUE #4074: Gate NAC Access to Collection List/Get
Gate NAC Access to Collection List/Get

shahzadlone:
## Relevant issue(s)
Resolves #4074 (Part of #2640)

## Description
- Cherry pick the previous schema get and version work and fix it to work for collection get/list gates.
- Share collection-get permission between the fetch option commands that are separate right now (like the name command) but in future will be consolidated.

Note:

- The command that currently returns all the collections will really become the `Collection List` command. But was discussed in standup and we are currently happy to tie them to one permission for now.
- Some more context: https://discord.com/channels/1374835078781468722/1422306930348523560/1423423824019325118


AndrewSisley (approved):
LGTM, with the same thought as the last PR, there are a lot of tests here that shouldn't need to be written, and they will impact the run time of `make test` and every multiplier (including CI).

shahzadlone:
> LGTM, with the same thought as the last PR, there are a lot of tests here that shouldn't need to be written, and they will impact the run time of `make test` and every multiplier (including CI).

Happy to discuss further, we can do a call sometime?

EDIT: We had a nice call, discussion summarized:

We can have potentially the 3-4 repetitive test cases as their own matrix of sorts with the new testing matrix stuff which we plan to migrate to soon. We discussed that this is not a priority for v1, even though it might help writing less tests, there is some thought required. 
When we do tackle it, we want to for sure run/multiply it with c-bindings and can decide what other actions to multiply it with.

Tracking issue: https://github.com/sourcenetwork/defradb/issues/4084

PR #4077, opened 2025-10-16
feat: Gate collection set active version op with NAC

### LINKED ISSUE #4076: Gate NAC Access to Collection Set Active Version
Gate NAC Access to Collection Set Active Version

shahzadlone:
## Relevant issue(s)
- Resolves #4076 (Part of #2640)
- Built on top of #4074 (so that should merge first)

## Description
- Cherry pick the previous schema set active version and update to current collection set active version
- Ignore all the commits before this commit: `PR(CHERRY-PICK): Gate - Set Active Collection Version`


acp/types/types.go
```diff
@@ -110,6 +112,8 @@ var RequiredResourcePermissionsForNode = []string{
 	"nac-relation-add",
 	"nac-relation-delete",
 	"collection-patch",
+	"collection-get",
+	"collection-set-active-version",
```
AndrewSisley: todo: I don't think this permission should exist, `collection-patch` allows the same operation, and they should be covered by the same permission.

You could perhaps instead require `collection-set-active-version` when changing which version is active when patching, but that feels like a maintenance cost that we shouldn't be taking on right now.

AndrewSisley (changes requested):
I don't think `collection-set-active-version` should exist.

acp/types/types.go
shahzadlone: I am happy to tie it to the same permission (collection patch).

shahzadlone:
> I don't think `collection-set-active-version` should exist.

Will remove it and tie to `collection-patch` thanks.

acp/types/types.go
shahzadlone: - [x] done

AndrewSisley (approved):
LGTM.

Note, it is much easier for reviewers if you squash the base branch's commits into a single `SQUASH - #9999` commit.

shahzadlone:
> LGTM.
> 
> Note, it is much easier for reviewers if you squash the base branch's commits into a single `SQUASH - #9999` commit.

Sadly, if I squashed it here my [Rebasing stacked branches with --update-refs](https://andrewlock.net/working-with-stacked-branches-in-git-is-easier-with-update-refs/#rebasing-stacked-branches-with-update-refs) wouldn't work properly :(. Instead maybe if I had one commit in the base PR that would be ideal.

Thanks for the reminder / sharing preference though, will try to keep in mind for future :)

cbindings/collection.go
ChrisBQu, line 1: Praise: Thank you for changing this. This looks correct to me.

ChrisBQu (approved):
LGTM

PR #4079, opened 2025-10-16
feat: Include query source in collection block

### LINKED ISSUE #4053: Include query source in collection block

AndrewSisley:
## Relevant issue(s)

Resolves #4053

## Description

Includes the query source delta in the collection blocks.

Required in order to send Views over P2P.

Based off of #4070 - please review that first.

### DIFF
--- a/internal/core/crdt/collection_definition.go
+++ b/internal/core/crdt/collection_definition.go
@@ -95,11 +95,7 @@ func (c *CollectionDefinition) Delta(
 			if !bytes.Equal(newQuery, oldQuery) {
 				queryDelta = newQuery
 			}
-		} else {
-			queryDelta = newQuery
 		}
-	} else if old.Query.HasValue() {
-		queryDelta = []byte{}
 	}
 
 	var transformDelta *string
--- a/internal/core/crdt/collection_definition.go
+++ b/internal/core/crdt/collection_definition.go
@@ -24,7 +24,7 @@ type CollectionDefinitionDelta struct {
 	Priority uint64
 
 	Name           string
-	QuerySelect    *[]byte
+	QuerySelect    []byte
 	QueryTransform *string
 }
 
@@ -78,7 +78,7 @@ func (c *CollectionDefinition) Delta(
 		return &CollectionDefinitionDelta{}, false, nil
 	}
 
-	var queryDelta *[]byte
+	var queryDelta []byte
 	if new.Query.HasValue() {
 		newQuery, err := cbor.Marshal(new.Query.Value().Query)
 		if err != nil {
@@ -92,13 +92,13 @@ func (c *CollectionDefinition) Delta(
 			}
 
 			if !bytes.Equal(newQuery, oldQuery) {
-				queryDelta = &newQuery
+				queryDelta = newQuery
 			}
 		} else {
-			queryDelta = &newQuery
+			queryDelta = newQuery
 		}
 	} else if old.Query.HasValue() {
-		queryDelta = &[]byte{}
+		queryDelta = []byte{}
 	}
 
 	var transformDelta *string
--- a/internal/core/crdt/collection_definition.go
+++ b/internal/core/crdt/collection_definition.go
@@ -11,7 +11,10 @@
 package crdt
 
 import (
+	"bytes"
 	"context"
+
+	"github.com/fxamacker/cbor/v2"
 
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/internal/keys"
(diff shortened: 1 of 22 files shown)


internal/core/crdt/collection_definition.go
```diff
@@ -66,14 +73,55 @@ func (c *CollectionDefinition) HeadstorePrefix() keys.HeadstoreKey {
...
+	}
+
+	var queryDelta *[]byte
+	if new.Query.HasValue() {
+		newQuery, err := cbor.Marshal(new.Query.Value().Query)
+		if err != nil {
+			return &CollectionDefinitionDelta{}, false, err
+		}
+
+		if old.Query.HasValue() {
```
fredcarle, line 89: question: Codecov tells me that this is never true in out tests. Would it be easy enough to add one that covers the `true` condition?

internal/core/crdt/collection_definition.go
```diff
@@ -66,14 +73,55 @@ func (c *CollectionDefinition) HeadstorePrefix() keys.HeadstoreKey {
...
+			oldQuery, err := cbor.Marshal(old.Query.Value().Query)
+			if err != nil {
+				return &CollectionDefinitionDelta{}, false, err
+			}
+
+			if !bytes.Equal(newQuery, oldQuery) {
+				queryDelta = &newQuery
+			}
```
fredcarle, line 97: question: Same question here.

internal/core/crdt/collection_definition.go
```diff
@@ -66,14 +73,55 @@ func (c *CollectionDefinition) HeadstorePrefix() keys.HeadstoreKey {
...
+		} else {
+			queryDelta = &newQuery
+		}
+	} else if old.Query.HasValue() {
+		queryDelta = &[]byte{}
+	}
```
fredcarle: question: Same question here.

internal/core/crdt/collection_definition.go
```diff
@@ -66,14 +73,55 @@ func (c *CollectionDefinition) HeadstorePrefix() keys.HeadstoreKey {
 func (c *CollectionDefinition) Delta(
 	new client.CollectionVersion,
 	old client.CollectionVersion,
-) (*CollectionDefinitionDelta, bool) {
+) (*CollectionDefinitionDelta, bool, error) {
 	if new.Name == old.Name {
-		return &CollectionDefinitionDelta{}, false
+		return &CollectionDefinitionDelta{}, false, nil
```
fredcarle: thought: This seems wrong. It made sense when only the `Name` was affecting the `Detla` but what if the query source changes but not the `Name`?

internal/core/crdt/collection_definition.go
```diff
@@ -66,14 +73,55 @@ func (c *CollectionDefinition) HeadstorePrefix() keys.HeadstoreKey {
...
+	}
+
+	var queryDelta *[]byte
```
fredcarle: question: Why do you define a pointer to the byte slice? If it's because you've set the field in the ipld schema to `optional`, it still doesn't need to be defined as a pointer. With `var queryDelta []byte`, `queryDelta` is `nil` and will be ignored when marshalling.

internal/core/crdt/collection_definition.go
AndrewSisley: I've would not be very easy, as we have a validation rule that prevents this from being reached.  I think that is unlikely to change, perhaps ever actually, so I'm going to remove this deadcode - thanks :)

- [x] Remove deadcode

internal/core/crdt/collection_definition.go
AndrewSisley: This is an old test gap, and IMO there is no reason to close it in this PR.  I've added a ticket, thanks for spotting: https://github.com/sourcenetwork/defradb/issues/4086

internal/core/crdt/collection_definition.go
AndrewSisley: I commented on the wrong code block.  Basically, adding/removing queries is blocked and I will remove those dead code paths.  Mutating an existing query is permitted, but it looks like there is a test gap there - a ticket has been opened and it can be solved outside of this PR.

internal/core/crdt/collection_definition.go
AndrewSisley: I missed that... Thanks Fred - I was surprised everything seemed to work first attempt and threw a panic in to test the other key blocks were being hit, they must have only been hit on view create.

This bug will have turned up handling P2P, but I'm glad it has been spotted earlier.

- [x] Actually handle updates

internal/core/crdt/collection_definition.go
AndrewSisley: Nice, thanks - I'll change that.

- [x] rm ptr

internal/core/crdt/collection_definition.go
AndrewSisley: Required test changes make a lot more sense now :)

fredcarle (approved):
LGTM. Thanks for fixing the bug :)

PR #4088, opened 2025-10-21
feat: Allow syncing of collection versions over P2P

### LINKED ISSUE #3892: Allow P2P syncing of Collections
Including view queries and Lenses.  Required for Shinzo.

AndrewSisley:
## Relevant issue(s)

Resolves #3892

## Description

Allows the syncing of collection versions over P2P, by version ID.


client/p2p.go
```diff
@@ -71,6 +71,12 @@ type P2P interface {
 	// to the documents or their collection for future updates.
 	// context.WithTimeout can be used to set a timeout for the operation.
 	SyncDocuments(ctx context.Context, collectionName string, docIDs []string) error
+
+	// SyncCollections syncs the given collection versions to the local node.
+	//
+	// It will not complete until a version is not found, so it is strongly recommended
+	// to set a timeout using `context.WithTimeout`.
+	SyncCollections(ctx context.Context, versionIDs ...string) error
```
shahzadlone, line 79: question: Guessing this needs to be gated with NAC too? (out of scope of this PR ofcourse)

cbindings/p2p.go
```diff
@@ -207,6 +207,38 @@ func P2PdocumentSync(nodePtr C.uintptr_t, collection *C.char, docIDs *C.char, ti
 	return returnC(returnGoC(0, "", ""))
 }
 
+//export P2PcollectionSync
+func P2PcollectionSync(nodePtr C.uintptr_t, versionIDs *C.char, timeoutStr *C.char) C.Result {
+	ctx := context.Background()
+	versionArgs := splitCommaSeparatedString(C.GoString(versionIDs))
+	timeoutDuration := time.Duration(0)
```
shahzadlone, line 214: question: @ChrisBQu do we need to apply the similar identity fix we did for collection set active version? (https://www.loom.com/share/25979d65369a41d4b85f3a0f50e11b33?sid=8b566e4a-e944-4e98-b0c1-1a2f73987c36) 


Probably for these all the p2p functions we want to gate?

shahzadlone (approved):
LGTM (ofc with exception of ci failures)
 
Some questions regarding NAC follow-up work (out of scope).

internal/datastore/txn.go
```diff
@@ -91,9 +92,9 @@ type BasicTxn struct {
 var _ Txn = (*BasicTxn)(nil)
 
 // newTxnFrom returns a new Txn from the rootstore.
-func NewTxnFrom(rootstore corekv.TxnStore, id uint64, readonly bool) *BasicTxn {
+func NewTxnFrom(rootstore corekv.TxnStore, id uint64, readonly bool, chunkSize immutable.Option[int]) *BasicTxn {
 	rootTxn := rootstore.NewTxn(readonly)
-	multistore := NewMultistore(rootTxn)
+	multistore := NewMultistore(rootTxn, chunkSize)
```
shahzadlone, line 97: nitpick: rename to `multiStore`
```suggestion
	multiStore := NewMultistore(rootTxn, chunkSize)
```

internal/db/lens_p2p_config.go
shahzadlone, line 1: nitpick: Not sure if filename is a typo `lens_not_jscript.go`. Why not `lens_not_js.go` or `lens_not_javascript.go`.

internal/db/p2p/p2p.go
```diff
@@ -266,10 +276,22 @@ func (p *P2P) hasAccess(ctx context.Context, pid string, c cid.Cid) bool {
 
 	block, err := coreblock.GetFromBytes(rawblock.RawData())
 	if err != nil {
+		if strings.Contains(err.Error(), "invalid key: \"modules\" is not a field in type Block") ||
+			strings.Contains(err.Error(), "invalid key: \"lens\" is not a field in type Block") ||
+			strings.Contains(err.Error(), "invalid key: \"wasmBytes\" is not a field in type Block") {
```
shahzadlone: question: Is there a better way to do these checks, perhaps through error codes?

internal/db/p2p/sync_col.go
```diff
@@ -0,0 +1,229 @@
...
+				kind = &client.CollectionKind{
+					CollectionID: *fieldDelta.CollectionID,
+					// The secondary side of CollectionKind relationships are never represented by
+					// blocks, so we can safely hardcode this to `false`
+					Array: false,
+				}
+			} else {
+				kind = client.IntToFieldKind(uint8(*fieldDelta.ScalarKind))
+			}
+		*/
```
shahzadlone, line 147: question: You want to leave this code comment in?

http/handler_p2p.go
```diff
@@ -402,6 +459,7 @@ func (h *p2pHandler) bindRoutes(router *Router) {
 	router.AddRoute("/p2p/collections", http.MethodGet, getPeerCollections, h.GetAllP2PCollections)
 	router.AddRoute("/p2p/collections", http.MethodPost, addPeerCollections, h.AddP2PCollections)
 	router.AddRoute("/p2p/collections", http.MethodDelete, removePeerCollections, h.RemoveP2PCollections)
+	router.AddRoute("/p2p/collections/sync", http.MethodPost, syncCollections, h.SyncCollections)
```
fredcarle, line 462: suggestion: `/p2p/collection-versions/sync` feels a lot more appropriate here. `/p2p/collections` represents a different thing so it doesn't make much sense to use it for this use case.

cli/cli.go
```diff
@@ -38,6 +38,7 @@ func NewDefraCommand() *cobra.Command {
 		MakeP2PCollectionAddCommand(),
 		MakeP2PCollectionRemoveCommand(),
 		MakeP2PCollectionGetAllCommand(),
+		MakeP2PCollectionSyncCommand(),
```
fredcarle, line 41: suggestion: Similar to my comment on the http handler url, this seems to be on the wrong parent command. This does not sync a p2p collection. It syncs a collection version so it would make sense for it to have it's own p2p sub-command.

internal/datastore/txn.go
fredcarle: None of the stores use a camel case format. This was changed a while ago and I don't think we should reintroduce this convention here.

internal/datastore/concurrent_txn.go
```diff
@@ -29,10 +30,15 @@ type concurrentTxn struct {
 }
 
 // newConcurrentTxnFrom creates a new Txn from rootstore that supports concurrent API calls
-func NewConcurrentTxnFrom(rootstore corekv.TxnStore, id uint64, readonly bool) *BasicTxn {
+func NewConcurrentTxnFrom(
+	rootstore corekv.TxnStore,
+	id uint64,
+	readonly bool,
+	chunkSize immutable.Option[int],
```
fredcarle, line 37: thought: Seeing `chunkSize` added to this and other functions feels intrusive. It forces the caller to include an extra param that is rarely ever used. Currently only used by Lens. It seems like a variadic func param or a config struct would be much nicer and also help prevent some breaking changes if we add/change options.

client/p2p.go
AndrewSisley: Yes, assuming `SyncDocs` is.

cbindings/p2p.go
AndrewSisley: I copy-pasted this from `SyncDocs`, if a fix is needed here, it is probably needed there too.

internal/db/lens_p2p_config.go
AndrewSisley: Annoyingly `lens_not_js.go` did not compile, as the compiler got confused and tried to build it for the js build target.  I didn't try `lens_not_javascript.go`, if you have a notable preference I'm happy to change it assuming I have other changes to push before merge :)

internal/db/p2p/p2p.go
AndrewSisley: Not through error codes, but at some point, we could perhaps do what signatures do, and try and fully parse them to their proper types (I do not want to do this now, nor perhaps ever, it is coupling to Lens internals somewhat), and then return true if successful.

I do see both methods as non-ideal and error prone, and I think my current preference is to remove both checks, and simply return true if we fail to parse a block to `core.Block`.  I didn't feel like thinking too hard on that right now though, but happy to change if people feel like it is important enough to do now (these solutions are all non-breaking on disk, other potential solutions might be though - e.g. add an enum to *all* blocks, and parse that first in order to determine type).

internal/db/p2p/sync_col.go
AndrewSisley: Yes, it will be back very soon.  Please see the code-comment immediately below for some extra info, otherwise we can have a quick call if you want.

http/handler_p2p.go
AndrewSisley: This was asked in discord, and the dev-ex guys were happy with the current path.  I'll @ you there and we can continue this discussion after you read it if need be :)

cli/cli.go
AndrewSisley: Resolving this comment, handling along side the other open one.

internal/datastore/concurrent_txn.go
AndrewSisley: > variadic func param or a config struct

I dislike both of these options for roughly the same reason - they will only have one item at the moment, and this feels like premature over-engineering that will end remaining forever a single item for its lifetime.

I think the time for adding either system is when adding a second option, assuming we don't know that there will be a second option right now.

This concern of mine is compounded by the fairly rushed throwing together of the chunk store, and it's controversy - it might be that we remove it for v2, increasing the likelihood of the variadic or config being removed before it is ever added to.

I agree that it is ugly and intrusive, but IMO it is still better than those alternatives.

http/handler_p2p.go
AndrewSisley: Discord conversation opened

- [x] Wait for discussion

internal/db/lens_p2p_config.go
shahzadlone: Just a nitpick. I do prefer slightly the other alternative then.

fredcarle: The compiler detects the suffix similarly to using `//go:build ...`:
```
lens.go // the default
lens_js.go // only compiles when `GOOS=js`
lens_linux.go // only compiles when `GOOS=linux`
lens_windows.go // only compiles when `GOOS=windows`
lens_darwin.go // only compiles when `GOOS=darwin`
```
So in my opinion it's preferable to keep the default `lens.go` and not `lens_not_jscript.go` and have it include `//go:build !js`

AndrewSisley: lens.go includes a bunch of stuff that the js build needs, and I'd rather not duplicate the whole file (103 lines)

http/handler_p2p.go
AndrewSisley: Discussed over discord, `/p2p/collections/sync` will remain as-is, the other paths will probably be changed at somepoint though.

internal/db/lens_p2p_config.go
fredcarle: sorry I was looking at the changed files and thought that that was replacing `lens.go`.

Perhaps `lens_p2p_config.go` would be more appropriate?

AndrewSisley: I like that, I'll rename the js one to match.

- [x] Rename files

client/p2p.go
```diff
@@ -71,6 +71,12 @@ type P2P interface {
 	// to the documents or their collection for future updates.
 	// context.WithTimeout can be used to set a timeout for the operation.
 	SyncDocuments(ctx context.Context, collectionName string, docIDs []string) error
+
+	// SyncCollections syncs the given collection versions to the local node.
+	//
+	// It will not complete until a version is not found, so it is strongly recommended
```
fredcarle: question: Should it be `until a version is found`?

http/client_p2p.go
```diff
@@ -234,3 +234,28 @@ func (c *Client) SyncDocuments(
...
 	return err
 }
+
+func (c *Client) SyncCollections(ctx context.Context, versionIDs ...string) error {
+	methodURL := c.http.apiURL.JoinPath("p2p", "collections", "sync")
+
+	req := map[string]any{
+		"versionIDs": versionIDs,
+	}
+
```
fredcarle, line 244: question: I'm curious to know why you prefer a map to a struct here?

http/client_p2p.go
```diff
@@ -234,3 +234,28 @@ func (c *Client) SyncDocuments(
...
+	deadline, hasDeadline := ctx.Deadline()
+	if hasDeadline {
+		req["timeout"] = time.Until(deadline).String()
```
fredcarle, line 247: question: Why are you converting this to a string instead of using the duration directly? This would prevent having to convert from string to duration in the handler.

internal/db/p2p/sync_col.go
```diff
@@ -0,0 +1,229 @@
...
+	if linkBlock.Delta.CollectionDefinitionDelta.QuerySelect != nil {
+		var q request.Select
+		err = json.Unmarshal(linkBlock.Delta.CollectionDefinitionDelta.QuerySelect, &q)
+		if err != nil {
+			return client.CollectionVersion{}, err
+		}
+
+		var transform immutable.Option[string]
+		if linkBlock.Delta.CollectionDefinitionDelta.QueryTransform != nil {
+			err = p.lens.P2P.Value().SyncLens(ctx, *linkBlock.Delta.CollectionDefinitionDelta.QueryTransform)
```
fredcarle: thought: Just realized that the `QueryTransfrom` should be defined as a `Link`/`CID` in the struct. Would you agree?

internal/db/p2p.go
```diff
@@ -220,3 +220,26 @@ func (db *DB) SyncDocuments(ctx context.Context, collectionName string, docIDs [
 	}
 	return db.p2p.SyncDocuments(ctx, collectionName, docIDs)
 }
+
+// SyncCollections syncs the given collection versions to the local node.
+//
+// It will not complete until a version is not found, so it is strongly recommended
```
fredcarle: question: Same thing here. Shouldn't it say `until a version is found`?

fredcarle (approved):
LGTM. A few comments to resolve before merge.

client/p2p.go
AndrewSisley: Yes, sorted

http/client_p2p.go
AndrewSisley: Because I copy-pasted from sync doc, and that used a map :)

http/client_p2p.go
AndrewSisley: This is copy-paste, I didn't think very hard about it, and would prefer not to :)

internal/db/p2p/sync_col.go
AndrewSisley: I have mixed feelings here.  This would make calling `SyncLens` harder unless that took a cid too.  At the moment, with `SyncLens` taking a string, `QueryTransfrom` should really be of type `string` too, even though internally to lens it is a link/cid.

I think I prefer Lens to take a string than a CID, similar to how Defra does.

internal/db/p2p.go
AndrewSisley: done

http/client_p2p.go
fredcarle: This should be fixed before v1 though, both of them. String is the wrong type here and changing post v1 would be an API breaking change.

fredcarle: That last comment was in reference to the timeout.

internal/db/p2p/sync_col.go
fredcarle: If it's a link, it should probably be a Link type within the Delta. This allows IPLD compatible block services to sync blocks and their links without needing to know about the schema of the block. SyncDoc can still use a string.

AndrewSisley: I disagree with the change, but I have made it.

PR #4093, opened 2025-10-23
feat: Gate verify signature op with NAC

### LINKED ISSUE #4092: Gate Verify Signature op with NAC
Gate Verify Signature op with NAC

shahzadlone:
## Relevant issue(s)
Resolves #4092

## Description
Gates Verify Signature operation with NAC


tests/integration/acp/nac/signature_verify_test.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func TestNAC_GatesVerifySignature_GoClient_AllowAccess(t *testing.T) {
+	test := testUtils.TestCase{
+		EnableSigning: true,
+		SupportedClientTypes: immutable.Some(
+			[]state.ClientType{
+				// Creating of signed documents over HTTP is not supported yet, because signing
+				// requires a private key which we do not pass over HTTP.
+				state.GoClientType,
```
ChrisBQu, line 81: question: Should the c client also support this? If so, we should list it here either as a supported client, or a to-do comment linking the issue you created. If not, how come?

ChrisBQu (commented):
The PR mostly looks good to me, but I have one question I would like to know the answer to before I feel good to approve it.

tests/integration/acp/nac/signature_verify_test.go
shahzadlone: I first did put a todo, but wasn't sure as it is a go specific test, and C client would I think work similar to cli and http flow than go in this case (hence the todo on the above test).

ChrisBQu (approved):

Issue #4091, opened 2025-10-23
Fix all c-binding identity bugs.


shahzadlone:
There are 2 or 3 types of c-binding bugs I found related to access control and identity passing between the c wrapper step.

Bugs:

1) Memory leaks: There are a lot of instances where the c type pointers are not freeing the identity pointer properly like it is for other properties.
For example in `cbindings/wrapper.go` consider:
```go
func (w *CWrapper) AddDACPolicy(
	ctx context.Context,
	policy string,
) (client.AddPolicyResult, error) {
	cIdentity := identityFromContext(ctx)
	cPolicy := C.CString(policy)
	defer C.free(unsafe.Pointer(cPolicy))
...
```
Note: the `cIdentity` does not have a defer `C.free` call (I am assuming that is a memory bug, and not purposeful @ChrisBQu correct me if I am wrong please). There are many other instances like this.

2) No identity being passed: The second bug is that the identity is not being passed at all. This means for all NAC instances where this happens, we would see `"not authorized to perform operation"` error even when access is given. Kind of the same bug as the one highlighted in this PR comment: https://github.com/sourcenetwork/defradb/pull/4088#discussion_r2449972401

Solution should be similar to this: https://www.loom.com/share/25979d65369a41d4b85f3a0f50e11b33?sid=8b566e4a-e944-4e98-b0c1-1a2f73987c36
Some NAC tests would have a `todo` marked this ticket where we need to add `state.CClientType` in the `SupportedClientTypes` list(`[]state.ClientType`) 

As of writing, this bug is at least in 3 places (likely way more places). `VerifySignature` and `GetIdentityNode` commands + the newly added commands all have this issue.

- Also fix for: https://github.com/sourcenetwork/defradb/pull/4088#discussion_r2451838155

- Also fix for p2p nac cbinding tests introduced in: https://github.com/sourcenetwork/defradb/pull/4111

ChrisBQu:
You are right about point one. It is a memory leak bug. And while it shouldn't ever cause tests to fail, its best to address and correct it.

ChrisBQu:
Question, after looking into it some. It seems this is a little different than the issue described in the loom video. It's not that identity is being passed in and disregarded; identity is just not being passed in. So this is not so much a logical error as it was a design mistake of the bindings(?) Does that sound right?

I'm happy to tackle this and make the changes/fixes. I will reach out to you about getting a bit more information on this issue.

PR #4097, opened 2025-10-23
feat: Gate sec. index ops with NAC

### LINKED ISSUE #4096: Gate Index ops with NAC
Gate Index ops with NAC

shahzadlone:
## Relevant issue(s)
Resolves #4096 

## Description
- Gate Index ops with NAC

~~## TODO~~
~~- [x] After rebase some http tests give 404 error [fix it].~~
~~- [x] Mark todo to fix the C-binding issue in future #4091~~


internal/db/store.go
```diff
@@ -98,6 +98,10 @@ func (db *DB) GetAllIndexes(
 	ctx, span := tracer.Start(ctx)
 	defer span.End()
 
+	if err := db.checkNodeAccess(ctx, acpTypes.NodeIndexListPerm); err != nil {
+		return nil, err
```
ChrisBQu, line 102: thought: Codecov picks up that this line is not covered by tests. It seems specific enough to me that maybe it should be specifically covered by a test. I don't think this is a blocker to me, but I'm interested in what you think.

ChrisBQu (approved):

internal/db/store.go
shahzadlone: This is because the current testing framework doesn't seem to make use of it. It will be one of the future todo tasks, I have make note of a few of these in issues. Thanks for reminder :)

shahzadlone: Tracking: https://github.com/sourcenetwork/defradb/issues/4102

PR #4104, opened 2025-10-24
fix: Encode identity as hex in commit query

### LINKED ISSUE #3999: Hex encode identity bytes in commit query
The identity value in the block signature returned on commit queries is a blob value that isn't compatible with the verify block command. This makes it awkward for devs to use. We should ensure it's returned hex encoded so that it's directly usable in the block verification query.

ChrisBQu:
## Relevant issue(s)

Resolves #3999 

## Description

Doing a commit query that gets the `Identity` field previously returned a string of text that was base64-encoded. This PR adjusts it so that it returns a hex-encoded string instead. This is so it is in a format that is more directly-usable by developers. Note that the VerifyBlockSignature function expects a hex-encoded string.

Changing this was a one-liner, but did result in some tests breaking, because they expected the old form. To fix them, adjustments to the `IdentityMatcher` have been made.

## Tasks

## How has this been tested?

- WSL


shahzadlone (approved):
question: I see this is still a draft and there are no tests for the previous and the new changes, is there a reason for that?

thought: Should we have a flag for returning identity in the base64-encoded blob value like we have before as well. I do think the default should be in hex

Giving LGTM as all the above comments are non-blocking.

fredcarle:
> thought: Should we have a flag for returning identity in the base64-encoded blob value like we have before as well. I do think the default should be in hex

I don't think we should add that functionality unless we have a use case for it. If partners want it in the future we can easily make the addition.

ChrisBQu:
> question: I see this is still a draft and there are no tests for the previous and the new changes, is there a reason for that?

Sorry it took so long to get back to this because of build week stuff, and then the priority placed on resolving bugs with C binding identity stuff. I have added a new integration test. There's not a lot going on with this PR, I feel, to test. But it does explicitly show and test the the identity should come back as a hex-encoded string. Let me know how you feel about it? I'll hold off on merging until I hear back.

tests/integration/signature/commit_test.go
```diff
@@ -455,3 +455,78 @@ func TestSignature_WithClientIdentity_ShouldUseItForSigning(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+func TestSignature_WithCommitQuery_ShouldBeHexEncoded(t *testing.T) {
+
+	test := testUtils.TestCase{
+		EnableSigning: true,
+		SupportedClientTypes: immutable.Some([]state.ClientType{
+			state.GoClientType,
```
shahzadlone: Suggestion: doesn't this mean all clients are supported? In that case why specify them all?

shahzadlone:
> > question: I see this is still a draft and there are no tests for the previous and the new changes, is there a reason for that?
> 
> Sorry it took so long to get back to this because of build week stuff, and then the priority placed on resolving bugs with C binding identity stuff. I have added a new integration test. There's not a lot going on with this PR, I feel, to test. But it does explicitly show and test the the identity should come back as a hex-encoded string. Let me know how you feel about it? I'll hold off on merging until I hear back. 
> 
> 

Was just a question, nothing blocking. Thanks for the reply.

tests/integration/signature/commit_test.go
ChrisBQu: ...I suppose it does mean that. Good catch. Let me fix that up.

PR #4106, opened 2025-10-24
ci: Enable Runs-On action runners

### LINKED ISSUE #4105: Enable Runs-On runner orchestration

fredcarle:
## Relevant issue(s)

Resolves #4105 

## Description

This PR makes our linux based test coverage workflows use AWS EC2 instead of Github Runners which will allow us to run faster machines at much lower costs.

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -65,7 +65,7 @@ jobs:
         database-type: [file, memory]
         mutation-type: [gql, collection-named, collection-save]
 
-    runs-on: runs-on=${{ github.run_id }}/runner=2cpu-linux-x64/spot=lowest-price/cpu=8+16/family=m7*+c7*/extras=s3-cache
+    runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/spot=pco/cpu=16+32/family=m7*+c7*/extras=s3-cache
 
     # Overwrite the defaults based on the basic matrix
     env:
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -65,7 +65,8 @@ jobs:
         database-type: [file, memory]
         mutation-type: [gql, collection-named, collection-save]
 
-    runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/spot=pco/cpu=16+32/family=m7*+c7*/extras=s3-cache
+    runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/\
+      spot=pco/cpu=16+32/family=m7*+c7*/extras=s3-cache
 
     # Overwrite the defaults based on the basic matrix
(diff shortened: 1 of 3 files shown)


.github/workflows/test-coverage.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         database-type: [file, memory]
         mutation-type: [gql, collection-named, collection-save]
 
-    runs-on: ubuntu-latest
+    runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/spot=pco/cpu=16+32/family=m7*+c7*/extras=s3-cache
```
shahzadlone: question: why not also other `runs-on` as well?

AndrewSisley (approved):
LGTM, thanks Fred!

.github/workflows/test-coverage.yml
AndrewSisley: Looks like it would be sensible to explore that in another chunk of work if it is not straight forward to do so right now.  This is a very good improvement to get in in itself, and I think Fred mentioned a problem/hassle with the mac runner.

shahzadlone: I am not concerned about the `mac-runner`(`test-os` job)  as that is not a required action anyway. But for example all other linux jobs like the `test-document-acp`, the `test-lens`, the `test-view`, the `test-encryption` the `test-telemetry` and the `test-js` all use linux jobs.

Without changing all of them I don't see how this will help us, because even if one required job is taking 17 minutes we have to wait for it anyway?

shahzadlone: Similar note, also should do this for change detector, it took 17+ mins

fredcarle: I'll add it to the other linux based jobs :)

shahzadlone (changes requested):
Thanks a bunch for this awesome improvement, I love the changes but we should do this for change detector and the other linux jobs in test coverage that are required (all of them are I think, can leave macos one as is)

image

Made a more descriptive comment here: https://github.com/sourcenetwork/defradb/pull/4106#discussion_r2462246617

.github/workflows/test-coverage.yml
shahzadlone: Thanks Fred!

fredcarle: @shahzadlone, Change has been made.

.github/workflows/test-coverage.yml
```diff
@@ -101,7 +107,7 @@ jobs:
             _${{ matrix.client-type }}\
             _${{ matrix.database-type }}\
             _${{ matrix.mutation-type }}\
-          "
+            "
```
shahzadlone, line 110: question: Why lol?

shahzadlone (approved):
PERFECTO!!! THANKSSSS LGTM

.github/workflows/test-coverage.yml
fredcarle: Cuz my IDE was showing red and it annoyed me.

PR #4113, opened 2025-10-25
ci: Fine tune self-hosted workflows

### LINKED ISSUE #4112: Cancel workflow if subsequen push before completion

fredcarle:
## Relevant issue(s)

Resolves #4112

## Description

This PR changes the instance type to use from memory optimized to cpu optimized to reduce workflow costs. This shouldn't affect performance since the memory optimized instances only used at most 25% of their memory.

We also make it so pushing a new commit on a PR will cancel the previously running workflows that were using self-hosted runners. This should avoid unnecessary costs.

### DIFF
--- a/.github/workflows/check-data-format-changes.yml
+++ b/.github/workflows/check-data-format-changes.yml
@@ -1,4 +1,4 @@
-# Copyright 2024 Democratized Data Foundation
+# Copyright 2025 Democratized Data Foundation
 #
 # Use of this software is governed by the Business Source License
 # included in the file licenses/BSL.txt.
@@ -25,7 +25,8 @@ on:
 
 # Cancel the workflow if its in progress when there is a new commit. This will reduce unecessary resource usage.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref == github.ref_protected && github.run_id || github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-\
+    ${{ github.ref == github.ref_protected && github.run_id || github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/check-data-format-changes.yml
```diff
@@ -23,12 +23,17 @@ on:
...
+  group: ${{ github.workflow }}-${{ github.ref == github.ref_protected && github.run_id || github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check-data-format-changes:
     name: Check data format changes job
 
     runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/\
-      spot=pco/cpu=16+32/family=m7*+c7*/extras=s3-cache
+      spot=pco/cpu=16+32/family=c7*+c6*/extras=s3-cache
```
shahzadlone: question: what is this?

shahzadlone (approved):
I understand the issue, but I don't follow how this fixes it :P I believe you will test that it actually works, so giving LGTM

.github/workflows/check-data-format-changes.yml
fredcarle: Removing the memory optimized family (they are more expensive and we don't benefit from the extra memory).

fredcarle:
> I understand the issue, but I don't follow how this fixes it :P I believe you will test that it actually works, so giving LGTM

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref == github.ref_protected && github.run_id || github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

This config makes it so you can't have the workflow running twice for the same PR. So when you push a new commit on the PR branch, it will start the workflow once again as it did previously but it will cancel the previous run if it hasn't completed yet.

Issue #296, opened 2022-03-15
Sec. Indexes: Allow backfill 


AndrewSisley:
Part of https://github.com/sourcenetwork/defradb/issues/201

PR #4114, opened 2025-10-27
fix: Lens-migrations-aware secondary indexes

### LINKED ISSUE #1597: Handle Lens migration of indexed collections
Currently migrations do not affect the indexes, this is undesirable. Resolving this is non-trival and requires thought, we should probably use the same mechanic for modifying/creating an index on an existing collection.

islamaliev:
## Relevant issue(s)

Resolves #1597

## Description

This PR make sure the secondary indexes are up-to-date for different collection versions and migrations.
Current implementation is not ideal as it reindexes the entire collection on every migration or version change, but this can be optimized only after [Backfill feature](https://github.com/sourcenetwork/defradb/issues/296) is implemented.

This PR also adds a new Increment lens module that can be used to increment/decrement doc field values.


internal/db/collection_define.go
```diff
@@ -504,6 +507,13 @@ func (db *DB) setActiveCollectionVersion(
 		}
 	}
 
+	if newActiveCol.HasValue() {
+		err = db.reindexNewActiveVersion(ctx, newActiveCol.Value())
```
AndrewSisley: todo: Please document why we may need to reindex here.

tests/integration/collection_version/migrations/query/simple_increment_test.go
```diff
@@ -0,0 +1,366 @@
...
+				}`,
+			},
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "level", "Kind": "Int"} }
+					]
+				`,
+			},
+			testUtils.ConfigureMigration{
```
AndrewSisley: suggestion: If you include the migration in the `PatchCollection` call you wont need to specify the version ids.

As well as being a little simpler, it also means the test wont break if the version ids change.

tests/integration/collection_version/migrations/query/simple_increment_test.go
```diff
@@ -0,0 +1,366 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package query
```
AndrewSisley: todo: The tests in this file do not appear to be testing any new Defra functionality not already covered by existing tests, and instead appear to be testing the new test Lens.

I strongly suggest removing them.  If you want to test the test lens, you can instead add unit tests for it - at least one of the test lenses (maybe in the lens repo) contains an example of how to do this.

tests/integration/lens.go
```diff
@@ -71,4 +71,9 @@ func configureMigration(
 	}
 
 	s.LensIDs = append(s.LensIDs, lensID)
+
+	// After setting migration the collection's Version.Previous.Value().Transform should be set.
+	// that's why we need to refresh collections, so that the in-memory collection versions are updated.
+	// Originally was added for [CreateIndex] to be able to index docs with migrated values.
+	refreshCollections(s)
```
AndrewSisley, line 78: praise: Nice catch, and thanks for the documentation!

tests/integration/results.go
```diff
@@ -426,45 +426,50 @@ func assertCollectionVersions(
 	expected []client.CollectionVersion,
 	actual []client.CollectionVersion,
 ) {
```
AndrewSisley, line 443: praise: Thanks for expanding the assert info - I had been forever procrastinating on that :grin:

tests/integration/view/simple/with_transform_test.go
```diff
@@ -326,3 +326,169 @@ func TestView_SimpleWithTransformReturningFewerDocsThanInput(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestView_SimpleWithIncrementTransform(t *testing.T) {
```
AndrewSisley: todo: Similar to other comment, these tests don't seem cover anything not already covered - please remove them.

internal/db/lens.go
```diff
@@ -99,5 +99,10 @@ func (db *DB) setMigration(ctx context.Context, cfg client.LensConfig) (string,
 		return "", err
 	}
 
+	err = db.reindexNewActiveVersion(ctx, dstCol)
```
AndrewSisley: suggestion: It might be okay for now, but this is not a great place to do this IMO.

Re-indexing is very slow, and unlikely to fail.  Here it is being done before other, quick, and error-prone operations have been performed.

This means that a user may wait an hour for the re-indexing for the first collection to complete, only to receive an error when processing the patch of the second collection.

This will likely become more important in the future, if/when the re-indexing needs to be done on a different transaction and/or async - here, the re-indexing is submitted before the original transaction processing the collection update may be committed.

internal/db/collection_index.go
```diff
@@ -692,3 +692,27 @@ func (db *DB) listAllEncryptedIndexDescriptions(
 
 	return indexes, nil
 }
+
+// reindexNewActiveVersion reindexes all documents in the collection for the new active version.
+func (db *DB) reindexNewActiveVersion(ctx context.Context, col client.CollectionVersion) error {
```
AndrewSisley, line 697: todo: This is only needed if there is a migration, which, as far as I know, is really quite rare, or if the two versions (newly activated, and newly deactivated) have different indexes (and then you only need to fix the differences, but maybe that logic can be written later).

Avoiding unnecessary re-indexing could save a lot of user/cpu time, and should be quite easy to check for.

Consider documenting that this check will need to be expanded if/when we allow fields to be mutated.

AndrewSisley (changes requested):
Looks good - is very easy to follow :)  I do think a couple of tweaks should be made before merge though, please argue back on any you disagree with :)

internal/db/collection_define.go
islamaliev: done

internal/db/collection_index.go
islamaliev: I added better checks if reindexing should happen

internal/db/lens.go
islamaliev: yeah, it's far from ideal and there a ticket to track this, but for now it's fine.

tests/integration/collection_version/migrations/query/simple_increment_test.go
islamaliev: we have tests for all other lenses. This is one of them. I just added it to a separate file because I covered different cases.
Maybe we don't need this many cases. I remove most of them and move the 1 or 2 remaining to `simple_test.go` where all other similar tests live.
What do you think?

tests/integration/view/simple/with_transform_test.go
islamaliev: I removed them

tests/integration/collection_version/migrations/query/simple_increment_test.go
islamaliev: done. Thank to this suggestion I found a bug in collections caching

tests/integration/collection_version/migrations/query/simple_increment_test.go
AndrewSisley: > we have tests for all other lenses.

The other lens are used to test defra, that is why they exist - the tests they are used in are not there to test them, they are test code, those tests use those Lenses to test Defra.  The Lenses are test code, they are not for users to use.

There are a few of them so that we can test different aspects of Defra code - e.g. normal iteration, more results than inputs being returned, fewer results than inputs being returned, etc.

> I [can?] remove most of them and move the 1 or 2 remaining to simple_test.go where all other similar tests live.

My preference remains the removal of most/all of them - one of your other comments suggested that it caught a bug, so it sounds like we must have had a test gap before, so that test should be kept, as should any others that test Defra (not Lens code) behaviour that isn't already tested.

internal/db/collection_define.go
```diff
@@ -502,8 +525,70 @@ func (db *DB) setActiveCollectionVersion(
...
+		if err != nil {
+			return err
+		}
 	}
 
 	// Load the schema into the clients (e.g. GQL)
 	return db.loadSchema(ctx)
 }
+
+func hasMigrationBetweenVersions(
```
AndrewSisley: todo: I think this function is incorrect - it only checks in a single straight line (in both directions).

The collection version history may branch, and this function needs to handle this.
```
       => C
A => B => D
```
The newly active version may be `D`, and the only other migration might be from `B` to/from `C` - this migration will still be run if a document of version `C` is received over P2P - the inverse will be applied from `C` down to `B`, then the new migration from `B` to `D` will run.

Please add a test for this.

internal/db/lens.go
```diff
@@ -99,5 +99,82 @@ func (db *DB) setMigration(ctx context.Context, cfg client.LensConfig) (string,
...
+		return false, client.CollectionVersion{}, err
+	}
+
+	isInChain := isMigrationInActiveChain(dstCol, activeCol, colsWithRoot)
+	return isInChain, activeCol, nil
+}
+
+// isMigrationInActiveChain checks if dstCol (where migration was just added) is in the
+// history chain of the activeCol. This means the migration affects the active version.
+func isMigrationInActiveChain(
```
AndrewSisley: todo: Similar to the other comment, this function looks incorrect and does not take into consideration branching versions.  It also doesn't account for documents that may have been migrated down from higher versions.

If you are not wanting to check for the existence of documents of various versions, I think the only correct way to check for migrations is to get all versions of a collection, and if any of them, regardless of history-location, have a migration, return true.

The same applies to the other comment.

AndrewSisley (changes requested):
I still think there are a couple of bugs here, code is looking good though :)

tests/integration/collection_version/migrations/query/simple_increment_test.go
islamaliev: > The other lens are used to test defra

well, I didn't create that lens just for fun. It's purely there for testing sec. indexes. So I don't see how this argument applies.

AndrewSisley: > well, I didn't create that lens just for fun

I didn't say you did.  What I'm saying is that a lot of the new tests added are essentially duplicates of existing tests whos primary purpose is to test the test Lens, not Defra.

For example, `TestSchemaMigrationIncrement_WithPositiveValueMigration_ShouldIncrement` just applies a very simple transform to a document - it will catch nothing that other tests would not.

The test name also makes it sound like the `increment` is important.  It absolutely is not in terms of Defra.  What is important is that it requires reading an existing value, and writing an output based on that value - everything else involved is internal to the test Lens.  

I think there is a very strong argument to be had that the new Lens does not allow the testing of anything new that the `CopyModule` lens does not, and that the new lens could be removed completely, but I'm not going to push you on that.

internal/db/collection_define.go
islamaliev: that's a nice catch. Andy. I added tests and adjusted implementation. Please take a look

internal/db/lens.go
islamaliev: done

tests/integration/collection_version/migrations/query/simple_increment_test.go
islamaliev: ok, I will remove it

tests/lenses/rust_wasm32_increment/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+pub struct Parameters {
+    pub field: String,
+    pub value: serde_json::Value,
+}
+
+static PARAMETERS: RwLock<Option<Parameters>> = RwLock::new(None);
+
+fn try_transform(
+    iter: &mut dyn Iterator<Item = lens_sdk::Result<Option<HashMap<String, serde_json::Value>>>>,
+) -> Result<StreamOption<HashMap<String, serde_json::Value>>, Box<dyn Error>> {
```
AndrewSisley: todo: This type looks like it was copy-pasted from a bad lens - there is no reason to use `HashMap` in either the input or return types.  Please use `Value` directly, it is accessible just like a map.

internal/db/collection_define.go
```diff
@@ -504,6 +527,55 @@ func (db *DB) setActiveCollectionVersion(
...
+// to a new active version by examining the full version history DAG using the lens
+// package's GetTargetedCollectionHistory function.
+//
+// This properly handles branching version histories by checking if any version
+// reachable from the new active version has a migration.
+func (db *DB) shouldReindexForVersionSwitch(
+	ctx context.Context,
+	newActiveCol client.CollectionVersion,
+) (bool, error) {
+	history, err := lens.GetTargetedCollectionHistory(
```
AndrewSisley: thought: I don't think this is a good enough reason to call into the lens package. `GetTargetedCollectionHistory` is not a lens specific function, and lens is not a utility package.

I think I prefer duplicating any code that needs to be used by both packages, but you could also extract this function to `core`.

tests/integration/collection_version/migrations/query/with_index_test.go
```diff
@@ -0,0 +1,1316 @@
...
+				`,
+			},
+			testUtils.ConfigureMigration{
+				LensConfig: client.LensConfig{
+					SourceSchemaVersionID:      schemaV1,
+					DestinationSchemaVersionID: schemaV2,
+					Lens: model.Lens{
+						Lenses: []model.LensModule{
+							{
+								Path: lenses.IncrementModulePath,
```
AndrewSisley, line 70: thought: I remain unconvinced that these new tests could not have been written using one of the existing lenses, instead of introducing a new one that will need to be understood when reading, and maintained.

AndrewSisley (approved):
LGTM :)

tests/lenses/rust_wasm32_increment/src/lib.rs
islamaliev: done

tests/integration/collection_version/migrations/query/with_index_test.go
islamaliev: This lens is mentally easier to wrap ones head around. At least for me. Especially in complex scenarios with migrations and indexes dancing together having this part straightforward help a lot.

internal/db/collection_define.go
islamaliev: move to description package

Issue #3524, opened 2025-03-10
Refactor integration tests to use `Action` interface


AndrewSisley:
I have a preference for the `Action` interface to be defined in a new repo, maybe `testing/action` so we can keep the multiplier and matcher stuff in the same place - discuss with team.

Islam's `SetState(s *state)` func might be superiour in a shared library, as it avoids the need for a generic in the `Execute` function.

For example:
```
// testing/state

type Stateful[TState any] interface {
   SetState(s *TState)
}
```
```
// testing/action

type Action interface {
	Execute()
}
```
```
// defra/test

type CreateDoc struct { ... }

var _  Stateful[defra.state] = (*CreateDoc)(nil)  // `defra.state` illustration only, implementation of `Stateful` is optional
var _  Action = (*CreateDoc)(nil) 
```

PR #4115, opened 2025-10-28
feat: Add ability to bypass DAC for performance

### LINKED ISSUE #3886: Ability to by-pass DAC for admin-identity
A request for documents coming from an admin identity should not trigger a DAC check.

shahzadlone:
## Relevant issue(s)
Resolves #3886

## Description
- Implementation of DAC bypass ~~using a cache to store~~ if an identity has the ability to dac bypass ~~or not~~. 
- There are a lot of tests because in previous work I skipped testing DAC access in combination with NAC. I have tested most if not all edge case concerns for DAC (public-doc, private-owned doc, private-not-owned own) <> NAC (on and temporarily turned off) <> Relationship (granting, revoking/not having)
- It is a bit odd to 100% assert DAC-bypass happens in the tests, as we are seeing "behavior" of dac-bypass working, not asserting it actually occurred (perhaps can find a better way to assert that in future). 

### Concerns / Discussion
I do have some concerns with the caching approach:
1) local acp calls are fast anyways so can get rid of caching and just make per request call to acp.
2) the caching approach will need to be scrapped if/when we implement sourcehub/global nac.
3) maintaining the cache state can be messy (atm we are clearing it through relationship add/delete calls so it's not too bad).
4) seems like we are holding/replicating the acp state that is part of the acp systems job perhaps not the application's.

However despite above, I am happy to merge this as is, because it works for what we have now (local NAC) and the testing covers the concerns I value the most.

## Todo
- [x] Removed cache completely according the standup discussion
- [x] Rename `db/permission` package to `db/acp`


internal/db/permission/dac_bypass.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"context"
+	"sync"
+
+	acpTypes "github.com/sourcenetwork/defradb/acp/types"
+	"github.com/sourcenetwork/defradb/client"
+)
+
+var identityBypassState = struct {
```
AndrewSisley: suggestion: I think it is unlikely that we want to cache this globally, for all contexts, ignoring transactions.

It would really surprise me if we didn't ever want this to respect transactions, especially local-ACP implementations or NAC (Source-hub being a little harder to sort out).

I strongly suggest scoping this cache to the current context in the same way we handle collections and their short ids.  It means the cache will be missed once per transaction, weakening it slightly, but it should be much less likely to cause problems now or in the future.

shahzadlone: - [x] Removed cache completely according the standup discussion

tests/integration/acp/nac/dac_access_by_empty_user_nac_off_test.go
```diff
@@ -0,0 +1,175 @@
+// Copyright 2025 Democratized Data Foundation
```
AndrewSisley, line 1: thought: This, and the others, have very long file names, and this reduces readability.   A lot of the time all people are going to see in their IDEs is `dac_access_by_empty...[hidden]`.  There are also a lot of files in this directory, and I think we would benefit from a few more sub directories, with short names in the future.

tests/integration/acp/nac/dac_access_by_empty_user_nac_off_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestNAC_Disabled_WithDACEnabled_AccessByEmptyUser_PrivateDocumentOwnedByNodeOwner_CanNotAccess(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			// Starting with NAC, so only authorized user(s) can perform operations from here on out.
+			testUtils.Close{},
```
AndrewSisley, line 24: thought: Starting the tests with a `Close` action makes no sense to me.  I think you tried to explain it once, but I can't remember.  Maybe the next issue you pick up could be to remove these, and change the test framework so that these no longer need to start, then close, then start an empty database.

AndrewSisley (approved):
LGTM, thanks Shahzad :)

fredcarle (approved):
LGTM

tests/integration/acp/nac/dac_access_by_empty_user_nac_off_test.go
shahzadlone: - [x] Made them a little shorter

tests/integration/acp/nac/dac_access_by_empty_user_nac_off_test.go
shahzadlone: We discussed this previously and you asked me to add a sub-task under this issue after I suggested that we should likely improve the `Start` testing framework action by consolidating the `configureNode`, you mentioned this should be sorted out when we move to testo:
https://github.com/sourcenetwork/defradb/issues/3524

Because in default case start is set through setStartingNodes(s) function that calls the setupNode if not a network test

I still see that sub-task as unresolved.

PR #4124, opened 2025-11-01
fix: Windows build

### LINKED ISSUE #4123: Windows release is broken
- The playground fails to download
- Build fails because windows tries to use wasmtime

fredcarle:
## Relevant issue(s)

Resolves #4123

## Description

This PR fixes the windows build by making the playground download compatible with windows and by ensuring windows doesn't try to use wasmtime.

### DIFF
--- a/internal/db/lens_wasmtime.go
+++ b/internal/db/lens_wasmtime.go
@@ -8,7 +8,7 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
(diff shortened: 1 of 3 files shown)


jsimnz (approved):

jsimnz:
~~Discussed with Fred, technically this puts a requirement to build with powershell only on Windows, but it's OK, since it's just the playground and can be fixed later~~ Nevermind, works in both `cmd` and `powershell`

PR #2040, opened 2023-11-11
ci: Add goreleaser workflow

fredcarle:
## Relevant issue(s)

Resolves #2039 

## Description

This PR adds a Goreleaser workflow to automate the publication of releases on Github and DockerHub.

Note that the release header and the milestone name should be updated manually for every release.

This is what the release would look like on Github:
Screenshot 2023-11-10 at 8 55 10 PM
Screenshot 2023-11-10 at 8 55 57 PM

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

With Goreleaser locally and through the CI on my defradb fork.

Specify the platform(s) on which this was tested:
- Debian Linux
- MacOS

PR #4129, opened 2025-11-04
fix: Glibc error with Docker container

### LINKED ISSUE #3953: Validate containerfile job fails due to glibc
Validate containerfile job fails due to glibc

Error: version `GLIBC_2.39' not found

Run that failed: https://github.com/sourcenetwork/defradb/actions/runs/16948611287/job/48035809394

fredcarle:
## Relevant issue(s)

Resolves #3953

## Description

This PR fixes the `GLIBC` error seen in with the docker image. This was caused buy the build image using Debian 13 and the deployed image using Debian 12.

The PR also adds a workflow so that we can release Docker images outside of the normal release workflow. This will be useful in situations like this where something needs to be fixed for the docker image but the rest of the release is fine.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -1,0 +1,101 @@
+# Copyright 2025 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+name: Docker release workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'New tag name'
+        required: true
+
+# Cancel the workflow if its in progress when there is a new commit. This will reduce unecessary resource usage.
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+  packages: write
+  issues: write
+
+jobs:
+  docker-build-push:
+    name: Build and push Docker image
+    runs-on: runs-on=${{ github.run_id }}-${{ github.run_attempt }}-${{ strategy.job-index }}/\
+      spot=pco/cpu=16+32/family=c6*+c7*/disk=large/extras=s3-cache
+    steps:
+      - name: Enable RunsOn action
+        uses: runs-on/action@v2
+        with:
+          metrics: cpu,network,memory,disk,io
+
+      - name: Checkout code into the directory
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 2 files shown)


tools/defradb.containerfile
```diff
@@ -4,7 +4,7 @@
 
 # Stage: build
 # Several steps are involved to enable caching and because of the behavior of COPY regarding directories.
-FROM docker.io/golang:1.24 AS build
```
AndrewSisley, line 7: question: Why?

AndrewSisley (approved):
LGTM, but I feel unqualified to review this - either way, my approval is there if you need it :)

tools/defradb.containerfile
AndrewSisley: Sorry, this unhelpful on this side - moving the comment.

tools/defradb.containerfile
```diff
@@ -4,7 +4,7 @@
...
+FROM docker.io/golang:1.24-bookworm AS build
```
AndrewSisley, line 10: question: Why?

fredcarle: Because otherwise it uses `1.24-trixie` which is Debian 13.

fredcarle: And as you can see, the run image using `bookworm`.

AndrewSisley: suggestion: Might be worth noting this in a comment here - `bookworm` is not explicitly referenced in this file as far as I can see, and it is important to get right.

fredcarle: Good point.

fredcarle: Done. Let me know if you think it should say something else.

AndrewSisley: Looks good, thanks Fred :)

.github/workflows/docker-release.yml
shahzadlone, line 1: question: Why not have a release test like action run with every PR (doesn't have to be required) to notify us whenever that breaks.

question: Did you test this action manually?

fredcarle: The container validation was giving this exact warning. We just ignored it until it became a problem after release. It used to be a required job.

It's the same workflow as in the release. Just without all the other release stuff.

shahzadlone, line 1: nitpick: All our actions use an action work as the first word, would like to see it perhaps be consistent

shahzadlone: > It used to be a required job.

When we had the 2 separate jobs before this PR?: https://github.com/sourcenetwork/defradb/pull/2040

> It's the same workflow as in the release. Just without all the other release stuff.

The reason I ask is because the input to that workflow was asking for a description tag. I am not sure where that is used in the script, is that relevant ?

fredcarle: > When we had the 2 separate jobs before this PR?: https://github.com/sourcenetwork/defradb/pull/2040

No. Before #3953 was opened.

> The reason I ask is because the input to that workflow was asking for a description tag. I am not sure where that is used in the script, is that relevant ?

It's used on lines 74 and 76.

fredcarle: I'll fix that in a future PR.

shahzadlone: > No. Before #3953 was opened.

Oh that's what you meant, gotcha.

shahzadlone: > I'll fix that in a future PR.

For what it is worth, I think I did like the way we had separate two jobs that we removed here: https://github.com/sourcenetwork/defradb/pull/2040/files

Of the top of my head I don't remember what pros and cons there are, so happy with your judgement on what you want to settle on :)

Thanks for resolving this

PR #4131, opened 2025-11-05
chore: Expand git version support for Makefile build version

### LINKED ISSUE #3997: Quirk in make tagging the github version + date
One one of my vm's, when I run `make && make install` and then `defradb version`, it reports:

```
defradb dev-develop (f1d79744 %cs) built with Go 1.24.6 linux/amd64
```

Notice the "%cs" there instead of the date?

But on another of my vm's, `defradb version` reports:

```
defradb dev-develop (f1d79744 2025-09-08) built with Go 1.24.6 linux/amd64
```

Both these vm's are debian, though the one with the quirk is old (debian 9.x) and the other is newer (debian 12.x).

Not a big deal really, just a quirk I noticed and wanted to point out so someone could figure out why this might be failing.

getify:
`%cs` was a more "recent" git formatting feature, but the `%cd` formatting (along with `--date=short`) does the same thing and works even on older git versions (such as 2.11).

## Relevant issue(s)

Resolves #3997

## Description

For makefile build version tagging, switching from `%cs` formatting string to `%cd` and `--date=short`, which extends support to older versions of git, and does the same thing as our current `%cs` usage.

## Tasks

- [X] I made sure the code is well commented, particularly hard-to-understand areas.
- [X] I made sure the repository-held documentation is changed accordingly.
- [X] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [X] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

Made this change locally and rebuilt, and it worked fine.

- Debian Linux 9.13
- Debian Linux 12

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ $(error "No git in $(PATH), version information won't be included")
 else
 VERSION_GOINFO=$(shell go version)
 VERSION_GITCOMMIT=$(shell git rev-parse HEAD)
-VERSION_GITCOMMITDATE=$(shell git show -s --format=%cs HEAD)
+VERSION_GITCOMMITDATE=$(shell git show -s --date=short --format=%cd HEAD)
 ifneq ($(shell git symbolic-ref -q --short HEAD),master)
 VERSION_GITRELEASE=dev-$(shell git symbolic-ref -q --short HEAD)
 else


AndrewSisley (approved):
LGTM, but I'm assuming there is not really much more to know besides what's in the description :)

You'd need prefix the PR title with `fix: ` or maybe `tools: ` (available options visible ln 16 of `tools/scripts/validate-conventional-style.sh` and shorten it so that it is 60 chars or less before you can merge.

getify:
61 characters was the shortest I could make it and still be accurate/descriptive. ¯\\_(ツ)_/¯

I mean, yeah, I could use "fix" and the problem goes away, but... now I'm just stubbornly resisting this arbitrariness on principle. chore is the more semantic type label, and without oddly abbreviating or omitting words, the description is what it should be IMO.

PR #4143, opened 2025-11-10
fix: Allow Lenses to pull docs from source

### LINKED ISSUE #4133: Query transforms appear to be running to completion, per document
They should not - they should allow the transforms to pull in as many documents as they like, before yielding items.

We have a test gap here, as none of our test modules aggregate doc state, some do filter and append docs, but they will not catch this problem.

For example, see the module https://codeberg.org/AndrewSisley/Gulistan/src/branch/init/src-tauri/modules/std_dev - it calculates the standard deviation, but Defra is messing things up when using it in a View so that the calc is run per document queried, instead of for all documents.

AndrewSisley:
## Relevant issue(s)

Resolves #4133

## Description

Allow Lenses to pull docs from source, instead of pushing them in.  Allowing Lenses to correctly aggregate data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ tests/lenses/rust_wasm32_prepend/pkg
 tests/lenses/rust_wasm32_filter/Cargo.lock
 tests/lenses/rust_wasm32_filter/target
 tests/lenses/rust_wasm32_filter/pkg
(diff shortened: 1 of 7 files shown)


islamaliev (approved):
LGTM!

fredcarle (approved):
LGTM

PR #4111, opened 2025-10-25
feat: Gate p2p ops with NAC

shahzadlone:
## Relevant issue(s)
Resolves #4110

## Description
Gate the following P2P ops:
- Connect
- SetReplicator
- DeleteReplicator
- GetAllReplicators
- AddP2PCollections
- RemoveP2PCollections
- GetAllP2PCollections
- AddP2PDocuments
- RemoveP2PDocuments
- GetAllP2PDocuments

## For Reviewers:
- Commit by commit should be clean (the first commit is just refactor and splitting of a large testing framework file)

## Future Todo(s) (ones with tick boxes will likely be done in this PR):
- [x] Finish testing p2p peer connect op
- [x] Finish testing p2p replicator ops
- [x] Finish testing p2p documents subscription ops
- [x] Finish testing p2p collection subscription ops
- [x] Finish testing p2p peer connect op with `admin` relation
- [x] Finish testing p2p replicator ops with `admin` relation
- [x] Finish testing p2p documents subscription ops with `admin` relation
- [x] Finish testing p2p collection subscription ops with `admin` relation
- Gate PeerInfo
- Gate SyncDocuments
- Test gap for `GetAllReplicators`: https://github.com/sourcenetwork/defradb/issues/4109

PR #4145, opened 2025-11-11
fix: Fix identity-related bugs in c bindings

### LINKED ISSUE #4091: Fix all c-binding identity bugs.

ChrisBQu:
## Relevant issue(s)

Resolves #4091 

## Description

This pull request does a few different things, which are all related to bugs in the way the c bindings, and the c binding integration test wrapper, used and passed identity.

Firstly:

A new function is available, `IdentityFree`, which takes an identity pointer, checks whether or not it is valid, and if it is valid, calls `Delete` on the associate CGO handle. This function is exported and made available to users, however we also now use it throughout the C client wrapper. Previously, we were creating a CGO handle every time we extracted an identity from a context, and made a pointer out of it, but we were never actually deleting the handle. This is probably harmless on the scale our tests exist, but it is a memory leak, and it is wrong. This has been fixed.

Secondly:

Several functions never took in identity, when in reality, they were supposed to. These are `BlockVerifySignature` ,`IndexList` , and `IndexDrop` , and all of the P2P-related functions. Because identity was never being passed into these functions, errors creeped in with regards to permissions of actions. To address this, these functions now take either a `CollectionOptions` object, or an identity pointer. 

*There has been a bit of questioning around whether or not the use of CollectionOptions is ideal in all of these locations. I think we should have a team-wide discussion on this topic, however, I think it should be a separate PR.*

Thirdly:

These changes have resulted in integration tests that were not behaving correctly before, to work.

 The C client type has been enabled for the following tests, and they will pass:

`TestNAC_AdminRelation_CanIndexCreate`
`TestNAC_AdminRelation_CanIndexDrop`
`TestNAC_AdminRelation_CanIndexList`

Tests prefixed with `TestNAC_GatesP2P*`
Tests prefixed with `TestNAC_AdminRelation*`

The to-do has been removed from the following tests, as it will not (and should not) ever pass:

`TestNAC_AdminRelation_CanVerifySignature`
`TestNAC_GatesVerifySignature_AuthorizedIdentity_AllowAccess`

And instead, the C client has been added as a supported client for the following tests, where it will pass:

`TestNAC_AdminRelation_GoClient_CanVerifySignature`
`TestNAC_GatesVerifySignature_GoClient_AllowAccess` (renamed: `TestNAC_GatesVerifySignature_GoClient_AuthorizedIdentity_AllowAccess`)

## Tasks

## How has this been tested?

- WSL


cbindings/wrapper.go
```diff
@@ -706,11 +720,25 @@ func (w *CWrapper) GetCollections(
...
-	defer C.free(unsafe.Pointer(colName))
+	cVersion := C.CString("")
+	cCollectionID := C.CString("")
+	cName := C.CString("")
+	cIdentity := identityFromContext(ctx)
+	defer C.free(unsafe.Pointer(cVersion))
+	defer C.free(unsafe.Pointer(cCollectionID))
+	defer C.free(unsafe.Pointer(cName))
+	defer C.IdentityFree(cIdentity)
+	defer C.IdentityFree(cIdentity)
```
shahzadlone: todo: There is a duplicate free.

```suggestion
	defer C.IdentityFree(cIdentity)
	defer C.IdentityFree(cIdentity)
```

tests/integration/acp/nac/signature_verify_test.go
shahzadlone, line 1: praise: Cheers for sorting this out

shahzadlone (approved):
Thanks for sorting this out bud, we discussed most of this in the call today so giving an LGTM and summarizing the todo(s) that need to be resolved pre-merge:

1) Rebase with develop, then resolve the missing todo(s) introduced in: https://github.com/sourcenetwork/defradb/pull/4111

2) Rename `TestNAC_GatesVerifySignature_GoClient_AllowAccess` to `TestNAC_GatesVerifySignature_GoClient_AuthorizedIdentity_AllowAccess`

3) Fix the double/extra free-ing

cbindings/index.go
```diff
@@ -112,15 +124,21 @@ func IndexList(nodePtr C.uintptr_t, collectionName *C.char) C.Result {
...
+
+	collectionName := C.GoString(options.name)
 
 	store, err := getStoreFromPointer(nodePtr)
 	if err != nil {
 		return returnC(returnGoC(1, err.Error(), ""))
 	}
 
-	col, err := store.GetCollectionByName(ctx, C.GoString(collectionName))
+	col, err := store.GetCollectionByName(ctx, collectionName)
```
AndrewSisley, line 141: question: Why are you doing this manually, instead of the `parseCollectionOptions` => `getCollection` flow I've seen elsewhere? e.g. in `CollectionUpdate`.

cbindings/p2p.go
```diff
@@ -176,8 +211,16 @@ func P2PdocumentGetAll(nodePtr C.uintptr_t) C.Result {
...
-func P2PdocumentSync(nodePtr C.uintptr_t, collection *C.char, docIDs *C.char, timeoutStr *C.char) C.Result {
+func P2PdocumentSync(nodePtr C.uintptr_t,
+	collection *C.char,
+	docIDs *C.char,
+	timeoutStr *C.char,
+	identityPtr C.uintptr_t) C.Result {
 	ctx := context.Background()
+	ctx, err := contextWithIdentity(ctx, identityPtr)
+	if err != nil {
+		return returnC(returnGoC(1, err.Error(), ""))
```
AndrewSisley, line 222: suggestion: The code seems to do this a lot - perhaps it would be better to write a function for this - this will reduce the noise from the usually unimportant error returns.

e.g.
```
func returnCErr(error) C.Result

func P2PdocumentSync() {
  ...
  if err != nil {
    return returnCErr(err)
  }
}
```

AndrewSisley (approved):
LGTM, but IMO this really should have been 2 different PRs, and two different items in the change log.

cbindings/index.go
ChrisBQu: That's a good catch. The honest answer is the same reasoning behind the order of parameters in the function name: because this was the first one I fixed, while working through it in real time, on a call with Shahzad. :) Can definitely do better. Let me just re-tackle this function and make a fresh commit.

ChrisBQu: Actually taking a closer look at it, I have some thoughts. It was a quick-change from taking in a collectionName to taking in a CollectionOptions, and this line is a result of that change. But I think the reason it called GetCollectionByName, taking in that parameter, is because that is how the CLI does it (see: `cli/index_list.go` line 34.)

I'm not against refactoring this for consistency if you feel strongly that the other approach is cleaner. However, your other comment already suggests maybe this PR should have been broken into two. I think that would further dilute it. Perhaps if its deemed important, we make a separate issue?

AndrewSisley: > because that is how the CLI does it

I think a lot of the c-bindings functions have this inconsistency - here we are trading inconsistencies against each other :)  I also think the c-bindings way is superiour and should probably be ported to our other clients.

If you feel like this needs more thought, you are probably right to not do it in this PR though.  Happy either way as this is just a couple of lines here and can be merged immediately after.

PR #4147, opened 2025-11-12
fix: Handle _version property in time travel queries

### LINKED ISSUE #4021: Server 500 from cid+_version query

AndrewSisley:
## Relevant issue(s)

Resolves #4021

## Description

Handle the `_version` property in time travel queries.

### DIFF
--- a/internal/planner/select.go
+++ b/internal/planner/select.go
@@ -446,18 +446,11 @@ func (n *selectNode) initFields(selectReq *mapper.Select) ([]aggregateNode, []*s
 				commitSlct := &mapper.CommitSelect{
 					Select: *f,
 				}
-				// handle _version sub selection query differently
-				// if we are executing a regular Scan query
-				// or a TimeTravel query.
+
 				if selectReq.Cid.HasValue() {
-					// for a TimeTravel query, we don't need the Latest
-					// commit. Instead, _version references the CID
-					// of that Target version we are querying.
-					// So instead of a LatestCommit subquery, we need
-					// a commits query with max depth starting from the
-					// target CID version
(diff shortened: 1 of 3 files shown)


internal/planner/select.go
```diff
@@ -446,18 +446,11 @@ func (n *selectNode) initFields(selectReq *mapper.Select) ([]aggregateNode, []*s
...
-				// or a TimeTravel query.
+
 				if selectReq.Cid.HasValue() {
-					// for a TimeTravel query, we don't need the Latest
-					// commit. Instead, _version references the CID
-					// of that Target version we are querying.
-					// So instead of a LatestCommit subquery, we need
-					// a commits query with max depth starting from the
-					// target CID version
-					commitSlct.DocID = immutable.Some(selectReq.DocIDs.Value()[0]) // @todo check length
```
AndrewSisley, line 459: note: The docID filtering of the parent is handled by the parent select/filter node(s).

getify:
I'm confused as to how this bug fix only was removal of code comments?

shahzadlone:
> I'm confused as to how this bug fix only was removal of code comments?

Line 459 is not a comment haha

shahzadlone (approved):
LGTM

getify:
ahhh I did miss that. sorry. 

might be helpful to document here why that removal of code fixes the bug, as that's not as common as needing to add code to fix bugs.

if I missed it, perhaps others might also miss it without more clear roadsigns.

PR #4167, opened 2025-11-13
fix: Bump sonic to version 14.2 with Go 1.24 support

### LINKED ISSUE #3982: Sonic dependency has a version mismatch with declared go module version
When compiling latest (as of 9/2/2025, git commit b9c1fb48f5d6f53ae9fd3193e1254cf540185eae), the compilation succeeds but every time you run the defradb binary, this warning is printed: 

> WARNING:(ast) sonic only supports go1.17~1.23, but your environment is not suitable

The module declaration for defra requires go 1.24.6, but the Sonic version we're pinned to is 1.12.3, and that version advertises (as the warning says) that it only supports go versions 1.17 - 1.23.

islamaliev:
## Relevant issue(s)

Resolves #3982

## Description

Bump sonic to version 14.2 with Go 1.24 support


shahzadlone (approved):
LGTM

question: But was just wondering, that since this is an indirect dependency, is the direct dependency that pulls this indirect dependency in compatible with this bump?

islamaliev:
> question: But was just wondering, that since this is an indirect dependency, is the direct dependency that pulls this indirect dependency in compatible with this bump?

yes, it is compatible

PR #4172, opened 2025-11-14
ci(i): Add a limited resource job

### LINKED ISSUE #4168: Introduce the slow CI actions running on github in a separate workflow but make them not required
Introduce the slow CI actions running on github in a separate workflow but make them not required

This was discussed in today's discussion to find flakeyness and bugs specific to way Edge devices.

shahzadlone:
## Relevant issue(s)
Resolves #4168 

## Description
~~- Runs tests with faster runner without coverage, should make them even faster (these will be required on PRs to merge).~~
~~- Runs tests with coverage using slower github runner (these will not be required on PRs to merge).~~
~~- Runs macos test with coverage only, as there is no benefit to it running with faster runs (and is not required anyway)~~
Edit:
- Introduce a job that runs on default (slower) github runner with http client, on badger store, with embedding, encryption, and collection-named mutation.

## Some findings when I introduced coverage and normal runs together of all the matrix.
I am not sure why the generating with coverage report and uploading artifact for some jobs seems to be faster than just running the tests without coverage (perhaps because of the number of actions running github might throttle resources?)

1.5 minute difference in the example below:

image

## Todo(s) before merge:
- Change branch rules required test names according to new changes (I don't think this is needed because the names of required jobs remain same).


.github/workflows/test-coverage.yml
```diff
@@ -8,6 +8,10 @@
 # by the Apache License, Version 2.0, included in the file
 # licenses/APL.txt.
 
+# Note: Unlike the [test.yml] workflow, the coverage jobs in this workflow are not required
```
AndrewSisley: suggestion: This is quite a lot of compute for something that is optional, and probably isn't going to get looked at a whole lot.

Perhaps it is time to ditch code-coverage from the the automatically executed PR jobs.  We can perhaps add it to the nightly job at some point.  And allow it's manual, on-demand, execution in PRs if/when people really want it.

fredcarle: It might make sense to reduce the items in the matrix too. We want the slower running tests to catch flakyness but it might not be necessary to cover all options to achieve this.

shahzadlone: @AndrewSisley 

> suggestion: This is quite a lot of compute for something that is optional, and probably isn't going to get looked at a whole lot. Perhaps it is time to ditch code-coverage from the the automatically executed PR jobs. We can perhaps add it to the nightly job at some point. And allow it's manual, on-demand, execution in PRs if/when people really want it.

I don't like the suggestion to remove code coverage completely, as mentioned in the PR description, our current tests (that run with coverage) are surprisingly running faster than the non-coverage tests in this PR (possibly due to running too many jobs + github throttling + github queueing).

I would rather keep what we currently have as is rather than remove code coverage and introduce the slower actions.

Code coverage in PRs has caught some quite gnarly bugs and very miss able human errors that would have otherwise not have been easy to catch.

Instead if we really want to go towards that route perhaps we should have some scheduled nightly slow runs to catch failures then instead of code coverage.

shahzadlone: @fredcarle 

> It might make sense to reduce the items in the matrix too. We want the slower running tests to catch flakyness but it might not be necessary to cover all options to achieve this.

You mean only for the slower running tests or even other tests? If only for the slower running ones I agree (but I wonder if that effects coverage paths?). We could then just run coverage tests with the custom runner like we do currently and keep the current matrix, and instead make the slower tests just be the ones that run with coverage + shrink the matrix jobs (so code coverage is as accurate as possible)

fredcarle: What I mean is that slower running tests hit flakiness in the internals of defra. For example, it's not related to the client used. So we could limit to http client with badger datastore and that would cover what we want..

shahzadlone: > What I mean is that slower running tests hit flakiness in the internals of defra. For example, it's not related to the client used. So we could limit to http client with badger datastore and that would cover what we want..

Yes I agree with that, but right now the slower tests are doing 2 things (As this PR currently stands):
- Coverage Generation
- Test flakeyness with slow runner

I was saying that we can keep the coverage to run like we have before this PR (with faster runners). And just do what you suggested (limit to http client with badger datastore) with the normal non-coverage tests on the slower runners.

shahzadlone:
Did the agreed upon changes, LMK if any more suggestions @AndrewSisley @fredcarle

fredcarle (approved):
LGTM

PR #4173, opened 2025-11-14
chore: Utilize default db path if blank string passed in

### LINKED ISSUE #4160: C-Bindings CreateNode path error
From build week feedback: https://discord.com/channels/1374835078781468722/1432471927553003690

> There seemed to be no way to use the default (`~/.defra`) database path when using the C-bindings using Rust `bindgen` to call them.  
> 
> Rust requires assignment to every property, and passing in a null pointer (`std::ptr::null()`) to `NodeInitOptions.dbPath` resulted in the following error:
> ```
> Creating Dir: "" err: mkdir : no such file or directory
> ```

ChrisBQu:
## Relevant issue(s)

Resolves #4160 

## Description

Previously, the `CreateNode` function of the c bindings would return an error if the `inMemory` flag was set to 0, but the `dbPath` field was a blank string. This PR changes the behavior to rely on Go's `os.UserHomeDir()` function to create a `.defradb` directory in the home directory, instead of erroring out.

## Tasks

## How has this been tested?

- WSL
- Android

### DIFF
--- a/cbindings/node.go
+++ b/cbindings/node.go
@@ -18,8 +18,6 @@ import "C"
 
 import (
 	"context"
-	"os"
-	"path/filepath"
 	"runtime/cgo"
 	"strconv"
 	"time"
@@ -42,23 +40,11 @@ func NewNode(cOptions C.NodeInitOptions) C.NewNodeResult {
 
 	ctx := context.Background()
 
-	// Use a database path if one is provided, otherwise try to determine the
-	// home directory and use that.
-	var defraPath string
-	if gocOptions.DbPath == "" {
-		home, err := os.UserHomeDir()
-		// This error should not happen on any major platform.
-		if err != nil {
-			return returnNewNodeResultC(1, errDatabasePathNotSet, nil)
-		}
-		defraPath = filepath.Join(home, ".defradb")
-	} else {
-		defraPath = gocOptions.DbPath
-	}
-
 	opts := []node.Option{
-		node.WithStorePath(defraPath),
 		db.WithLensRuntime(db.Wazero),
+	}
+	if gocOptions.DbPath != "" {
+		opts = append(opts, node.WithStorePath(gocOptions.DbPath))
 	}
 	if len(listeningAddresses) > 0 {
 		opts = append(opts, p2p.WithListenAddresses(listeningAddresses...))
--- a/cbindings/node.go
+++ b/cbindings/node.go
@@ -18,6 +18,8 @@ import "C"
 
 import (
 	"context"
+	"os"
+	"path/filepath"
 	"runtime/cgo"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


cbindings/node.go
```diff
@@ -40,20 +42,22 @@ func NewNode(cOptions C.NodeInitOptions) C.NewNodeResult {
...
+			return returnNewNodeResultC(1, errDatabasePathNotSet, nil)
+		}
+		defraPath = filepath.Join(home, ".defradb")
+	} else {
+		defraPath = gocOptions.DbPath
+	}
 
 	opts := []node.Option{
-		node.WithStorePath(gocOptions.DbPath),
+		node.WithStorePath(defraPath),
```
AndrewSisley: todo: This feels like the wrong way of doing this - you are having to duplicate a bunch of Defra business-logic in the c-bindings.

Instead, if `gocOptions.DbPath` is empty, IMO you should not be calling `node.WithStorePath` at all.

AndrewSisley (changes requested):
I feel like this is the wrong way of doing this - you shouldn't be providing the Go option if the user hasn't given it.

cbindings/node.go
ChrisBQu: Hmm. Something has to be going on under the hood that I don't completely understand, yet. If we try as you say, for example, something like this:

```
	opts := []node.Option{
		db.WithLensRuntime(db.Wazero),
	}
	if gocOptions.DbPath != "" {
		opts = append(opts, node.WithStorePath(gocOptions.DbPath))
	}
```

it will work as expected... on Linux. But it will fail on Android. That is, you can pass in a blank path on the Linux build, and call `NewNode`, and the default path will work. But Android will still give the error: `Error creating node: Error Creating Dir: "" err: mkdir : no such file or directory`. So I suspect that even though the idea of determining a default path must appear in two locations, it is not being done the same way.

If I search the codebase for `os.UserHomeDir()`, the only other spot I see it in, is the `setContextRootDir` function in `cli/utils.go`. So I don't think this is, necessarily, how the default path is being determined for other clients -- but it is the (at least, *a*) safe way to do so on Android, that is compatible with other platforms.

It could be argued that it makes sense to adjust how the default path is determined if a blank string is put in to begin with. I'd like to have a look at that, maybe. Do you know where in the codebase it happens? I'm digging, and not quite seeing it.

AndrewSisley: Oh that's not so good - it looks like `node` is not handling this at all, and the default is probably being set by the CLI.

I have a notable preference for this to be handled by `node`, and not duplicated (with potential variations) by all our clients.  I assume the embedded Go client just errors if not given a path/!in-memory.

ChrisBQu: If that's the case, it sounds like we either need to expand this PR, or use the second approach, from my above comment, in this PR and then make a new PR(?) Do you have a preference on that?

AndrewSisley: I have a preference for this PR to not be merged without my preferred changes - as at the moment it is actually introducing an inconsistency between the embedded C client, and the embedded Go client (which will error as the C client in develop does).

I think the team should be consulted over discord/standup first though, others might have thoughts on this - there is an argument to be had that defaults are client specific.

AndrewSisley (approved):
LGTM, thanks Chris - sorry about the run around it took to get this in :)

PR #4174, opened 2025-11-14
fix: Handle _version alongside type joins

### LINKED ISSUE #1709: queries on type with requested relational object and _version don't work

AndrewSisley:
## Relevant issue(s)

Resolves #1709

## Description

Handles `_version` properties that exist alongside type joins.

### DIFF
--- a/internal/planner/multi.go
+++ b/internal/planner/multi.go
@@ -81,6 +81,29 @@ func (p *parallelNode) Kind() string {
 }
 
 func (p *parallelNode) Init() error {
+	newChildren := make([]planNode, len(p.children))
+	newChildIndexes := make([]int, len(p.childIndexes))
+
+	endIndex := len(p.children) - 1
+	startIndex := 0
+	for i, child := range p.children {
+		switch child.(type) {
+		case *dagScanNode:
+			// Any node types that result in `nextAppend` calls must be executed last, as they are
+			// dependent on the docID set by the `nextMerge` calls.
+			newChildren[endIndex] = child
+			newChildIndexes[endIndex] = p.childIndexes[i]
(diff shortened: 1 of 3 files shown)


internal/planner/multi.go
```diff
@@ -81,6 +81,29 @@ func (p *parallelNode) Kind() string {
...
+			endIndex--
+		default:
+			newChildren[startIndex] = child
+			newChildIndexes[startIndex] = p.childIndexes[i]
+			startIndex++
+		}
+	}
+
+	p.children = newChildren
+	p.childIndexes = newChildIndexes
```
islamaliev, line 109: suggestion: the significance of this change is not clear here. Maybe a comment explaining an examples case it solves. Even better would to extract this block into a function with a descriptive name like `orderChildrenExecution`

islamaliev (approved):
LGTM!

internal/planner/multi.go
AndrewSisley: You don't think the existing comment covers it?  Do you have a rough example as to what you are looking for?
> // Any node types that result in `nextAppend` calls must be executed last, as they are
  // dependent on the docID set by the `nextMerge` calls.

fredcarle (approved):
Thank for fixing this. It would be nice to document this a little more before merge. The comment in the switch case is not giving much information unless the reader is very familiar with that part of the code.

internal/planner/multi.go
AndrewSisley: Fred's comment gave me an idea, as this code is in a different block to the one with the `nextAppend`/`nextMerge` calls - I've linked to it more explicitly.

PR #4175, opened 2025-11-15
fix: Introspection error and badly formatted GQL types

### LINKED ISSUE #4132: GraphQL introspection doesn't work since v0.20
In Altair I get 
_Looks like the GraphQL schema is invalid. Please check that your schema in your GraphQL server conforms to the latest GraphQL spec._

In Postman I get
_Could not load GraphQL schema.
Unknown type reference: { kind: "INPUT_OBJECT", name: "", ofType: null }._

jsimnz:
## Relevant issue(s)

Resolves #4132 

## Description

Fixes the introspection issue. Theres room for discussion with the solution. Specifically the new `ScalarAggregateNumericBlock` type. Which includes a "noop" field `_`, since the `gql.InputObject` type requires at least one field to be defined.

This wasn't needed before, but the method to get around this, also causes the introspection error in the first place. There might be away around it in the depths of the gql library but its not self evident how.

In reality, this is a symptom of *how* we target/select fields for aggregates, and not a issue with the gql lib itself.

This PR also introduces a new CI test workflow to handle using nodejs in a specific unit test. We need to test the actual GraphQL client that most JS based client use to verify introspection results, which means we need to use the actual library via node. There is a new test in the `internal/request/graphql/schema` package specifically verifying the introspection output. Note this is a unit test, and therefore obviously doesn't cover some of the code paths between an external introspection request finding its way all the way to the gql lib, just the important 90% of it. I'm open if people have strong feelings to include a full integration test for this.

## Tasks

- [ x I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).

## How has this been tested?

Updated existing integration tests. Added a gql introspection conformance test.

- Ubuntu (WSL2)


internal/request/graphql/schema/generate.go
```diff
@@ -914,8 +918,15 @@ func (g *Generator) getNumericFields(obj *gql.Object) map[string]gql.Type {
 	for _, field := range obj.Fields() {
 		listType, isList := field.Type.(*gql.List)
 		var inputObjectName string
-		if !isList {
-			fieldTypes[field.Name] = gql.NewInputObject(gql.InputObjectConfig{})
+
+		col := g.typDefCollectionMap[obj.Name()]
+
+		// fmt.Println("Collection:", col.Name)
```
ChrisBQu: Nitpick: I think we should remove these debug print comments before merging.

tests/integration/collection_version/aggregates/inline_array_test.go
ChrisBQu, line 1: Praise: Nice to see these tests get cleaned up and simplified so much. Maybe the size of them before should have been a flag to us (me) that the approach to get them to work was fishy. Sorry you had to deal with them how they were.

ChrisBQu (approved):
LGTM. I left just one nitpick. It is not a blocker at all.

internal/request/graphql/schema/generate.go
jsimnz: This isn't a nitpick. Definitely a `todo` thanks for flagging. I'm notorious for accidentally leaving them in

internal/request/graphql/schema/introspection_nodejs_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	tempDir := t.TempDir()
+	resultFileName := filepath.Join(tempDir, "introspection_data2.json")
+	filebuf, err := json.Marshal(r.Data)
+	require.NoError(t, err)
+	err = os.WriteFile(resultFileName, filebuf, 0644)
+	require.NoError(t, err)
+
+	// this requires running `make deps:test:js` before running this test
+	cmd := exec.Command("../../../../tests/node_modules/.bin/graphql-introspection-json-to-sdl", resultFileName)
```
nasdf: thought: you could replace this with `npx -y graphql-introspection-json-to-sdl`

fredcarle (approved):
Approving with the expectation that Keenan's comment will be resolved :)

jsimnz:
@fredcarle Keenan's comment was marked as a "thought" do you feel stronger for it to be a suggestion/todo?

fredcarle:
> @fredcarle Keenan's comment was marked as a "thought" do you feel stronger for it to be a suggestion/todo?

I forgot to point that out in my comment. For me it's at the suggestion level. I think it will result in less potential mistakes/failures in the future, specifically related to the path that is hard coded.

internal/request/graphql/schema/introspection_nodejs_test.go
jsimnz: Great suggestion, certainly made things simpler

PR #4186, opened 2025-11-19
fix: Make node apply a default store path

### LINKED ISSUE #4184: Node doesn't set default store path
It was discovered that creating a defra node without the `inMemory` option, and then also *not* passing in the `node.WithStorePath` option results in no default path ever being set. Currently, the CLI does set its own default path. But this is behavior that should happen inside the node package, so that it's consistent across the clients.

ChrisBQu:
## Relevant issue(s)

Resolves #4184

## Description

If a Defra node was created without the `node.WithStorePath` option, the default store path, `$HOME/.defradb`, would never actually get set. The `cli` package did contain a function, `setContextRootDir` which set such a path in its context, but this logic was unique to that package. This PR changes this by creating a new function in the `node` package, `GetDefaultStorePath`. This function is used by the `DefaultStoreOptions` function, to apply the default value to the `path` field. Additionally, this function is called by the CLI inside the `setContextRootDir` function so as not to break any of the ways context is passed around or used, while also making sure the logic remains consistent across clients, without ever being repeated in different locations.

One note, is that this PR does not yet adjust how the way in which the C bindings package pass in the `node.WithStorePath` option to the new node. The reason for this, is that a PR already exists for exactly that issue. Presumably, it can be closed and merged after this one gets merged and closed.

## Tasks

## How has this been tested?

I used the C bindings on WSL and on Android, and the CLI with WSL.

- WSL
- Android

### DIFF
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -16,7 +16,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -32,6 +31,7 @@ import (
 	"github.com/sourcenetwork/defradb/http"
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/keyring"
+	"github.com/sourcenetwork/defradb/node"
 )
 
 const (
@@ -172,11 +172,7 @@ func setContextRootDir(cmd *cobra.Command) error {
 		return err
 	}
 	if rootdir == "" {
-		home, err := os.UserHomeDir()
-		if err != nil {
-			return err
-		}
-		rootdir = filepath.Join(home, ".defradb")
+		rootdir = node.GetDefaultStorePath()
 	}
 	ctx := context.WithValue(cmd.Context(), rootDirContextKey, rootdir)
 	cmd.SetContext(ctx)
--- a/cbindings/node.go
+++ b/cbindings/node.go
@@ -40,21 +40,11 @@ func NewNode(cOptions C.NodeInitOptions) C.NewNodeResult {
 
 	ctx := context.Background()
 
-	// // Create the directory if it doesn't exist, and inMemory flag is not set
-	// For now this is not done, but we leave it here because we might need it in
-	// the future, when running on mobile platforms.
-	// if !inMemoryFlag {
-	// 	if _, err = os.Stat(gocOptions.DbPath); os.IsNotExist(err) {
-	// 		err := os.MkdirAll(gocOptions.DbPath, 0755)
-	// 		if err != nil {
-	// 			return returnGoC(1, err.Error(), "")
-	// 		}
-	// 	}
-	// }
-
 	opts := []node.Option{
-		node.WithStorePath(gocOptions.DbPath),
 		db.WithLensRuntime(db.Wazero),
+	}
+	if gocOptions.DbPath != "" {
+		opts = append(opts, node.WithStorePath(gocOptions.DbPath))
 	}
 	if len(listeningAddresses) > 0 {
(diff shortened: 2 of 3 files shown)


cli/utils.go
```diff
@@ -172,11 +172,7 @@ func setContextRootDir(cmd *cobra.Command) error {
 		return err
 	}
 	if rootdir == "" {
-		home, err := os.UserHomeDir()
-		if err != nil {
-			return err
-		}
-		rootdir = filepath.Join(home, ".defradb")
+		rootdir = node.GetDefaultStorePath()
```
AndrewSisley, line 175: > this function is called by the CLI inside the setContextRootDir function so as not to break any of the ways context is passed around or used

question: What broke and how?

ChrisBQu: Sorry, I think my wording was poor. Let me try again.

The `setContextRootDir` function, inside `cli/utils.go`, sets the context of the command such that the `rootDirContextKey` key is set to the root directory. The CLI will later rely on this key being set inside the `MakeStartCommand` function. For example, it uses it to know how to set `node.WithDocumentACPPath` or `node.WithNodeACPPath`. What I meant to imply was, this PR does not adjust those behaviors of the CLI. Instead, it just factors out the way that the path was being determined. That logic no longer belongs to the CLI's utility function.

Does this make sense?

AndrewSisley: Ah okay, so this is still only used within the Start command in order to set other options?

If so, I have a preference (sorry, because it is bothersome) for those other options to be adjusted so that their defaults too are handled inside the `node` package (execution-wise, not just code-location).  It sounds like they have the same problem that `WithStorePath` has in develop.

ChrisBQu: I might not completely understand what you are asking for. It is true that this context key is only used by the Start command. However, it is the mechanism by which non-default values would be assigned to the options.

It is true, now that you mention it, that function `DefaultNodeACPOptions()` also does not have a default value. That is something that we could potentially change. However, I think it is a separate issue from this one (although, admittedly, they share similarities.)

Of further note, the C bindings actually do not have any flag options to adjust the NodeACPPath *or* the DocumentACPPath. They are both hardcoded as blank.

I am not convinced that this issue should change in scope -- though I am open to it. However, I am becoming increasingly convinced that there is a follow-up issue relating to ACP and Document paths. What do you think?

AndrewSisley: My opinion is that the logic for all options should be handled by the `node` package or lower, except in rare cases where we really want a client specific default or something.

This means the CLI should not contain, nor execute, logic for setting `WithDocumentACPPath`, `WithNodeACPPath` or `WithStorePath`.

If the ACP options are dependent on the store path, then that logic should be handled from within `node` or deeper (e.g. `acp`).

IMO this means that this PR does not resolve the issue - the default path logic is still being executed/managed by the CLI, and fed into ACP and other options.  It does make it slightly better by moving the code, but it has not resolved the core underlying issue.

It is unfortunate to have hit some old tech debt that we missed from the ACP options, but that is an inherent risk in any issue we pick up.  To me fixing this properly means removing the tech debt, and then resolving the issue - possibly in two PRs, with the tech debt removal being merged first.  However, time is finite, and now we know about the debt, we can either deliberately chose to stack more on top of it, or remove it - might be worth raising with the team.  

The workaround does solve the CLI dev-ex issue for `WithStorePath`, but the solution could be cleaner if the tech debt is removed first.

AndrewSisley: Discussed in the standup.  The team is happy enough in the short term with the current change in the PR, the long term solution is more effort than we are willing to commit right now.

AndrewSisley (approved):
LGTM, sorry I forgot to approve the after we talked last week :)

PR #4187, opened 2025-11-19
fix: Mutation of branchable collection

### LINKED ISSUE #4185: Not possible to mutate branchable collection
```go
func TestBranchableCollection_AddingFieldViaPatching_ShouldSucceed(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type User @branchable {
						name: String
					}
				`,
			},
			testUtils.CreateDoc{
				NodeID: immutable.Some(0),
				DocMap: map[string]any{
					"name": "John",
				},
			},
			testUtils.CreateDoc{
				NodeID: immutable.Some(0),
				DocMap: map[string]any{
					"name": "Islam",
				},
			},
			testUtils.PatchCollection{
				NodeID: immutable.Some(0),
				Patch: `
					[
						{ "op": "add", "path": "/User/Fields/-", "value": {"Name": "email", "Kind": 11} }
					]
				`,
			},
		},
	}

}
```
Get this result:
```
Error:     Received unexpected error:
           mutating IsBranchable is not supported. Collection: User
```

islamaliev:
## Relevant issue(s)

Resolves #4185 

## Description

This fix makes it possible to mutate branchable collections

### DIFF
--- a/tests/integration/collection_version/updates/add/branchable_test.go
+++ b/tests/integration/collection_version/updates/add/branchable_test.go
@@ -15,11 +15,69 @@ import (
 
 	"github.com/sourcenetwork/immutable"
 
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/client/request"
 	"github.com/sourcenetwork/defradb/tests/action"
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
 )
 
-func TestBranchableCollection_AddNewField_ShouldAddField(t *testing.T) {
+func TestBranchableCollection_AddNewField_ShouldUpdateCollectionDefinition(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddSchema{
+				Schema: `
+					type User @branchable {
+						name: String
+					}
+				`,
+			},
+			testUtils.PatchCollection{
+				NodeID: immutable.Some(0),
+				Patch: `
+					[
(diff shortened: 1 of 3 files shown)


tests/integration/collection_version/updates/add/branchable_test.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"testing"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestBranchableCollection_AddNewField_ShouldAddField(t *testing.T) {
```
AndrewSisley: suggestion: You could simplify this test by removing the create document actions and query and instead call `GetCollections`.  At the moment there is quite a lot going on and I would argue that a lot of it is irrelevant (multiple docs, one on the far side of the patch) - even if we felt like we want to test that situation, my preference would be to have the simple test followed by a more complicated document test with documentation detailing exactly why we felt the need to create documents in that order.

By creating docs in the current manner the code is suggesting that it *is* important, without explaining why.

AndrewSisley (approved):
LGTM, but I suggest simplifying the test before merge

tests/integration/collection_version/updates/add/branchable_test.go
islamaliev: added simpler test

PR #4188, opened 2025-11-19
feat: Sync branchable collection

### LINKED ISSUE #4155: Sync branchable collection documents
From build week feedback: 
- https://discord.com/channels/1374835078781468722/1435369228289966111
- https://discord.com/channels/1374835078781468722/1433484447872712936
- https://discord.com/channels/1374835078781468722/1435530561207799831

> We have /p2p/documents/sync http endpoint that given docID will update the document's DAG.
> Now what seemed to analogous function for branchable collection /p2p/collections/sync is not. Instead of syncing the collection's DAG, it just fetches the collection's definition from peers.

> I do not want a replicator, I want users to be able to on demand sync all, or a range, of docs in a collection that they have potentially subscribed to.  For example when they start defra after being offline for a while and wish to catch up.  Accuracy/completeness of sync is not important.
> 
> Something using branchable collections would work, but I would prefer that it was not limited to them.
> 
> Perhaps supplying no docIDs to `SyncDocs` could trigger this, or it could eventually take an object similar to `CollectionFilterOptions`.

islamaliev:
## Relevant issue(s)

Resolves #4155

## Description

This PR adds `SyncBranchableCollection` method to client interface.


Makefile
```diff
@@ -437,10 +437,14 @@ fix:
...
 	@$(MAKE) docs
 
-.PHONY build-c-shared-linux:	
+.PHONY build-c-shared-linux:
 build-c-shared-linux:
 	@tools/scripts/build-c-shared-linux.sh $(BUILD_FLAGS)
 
+.PHONY build-c-shared-macos:
+build-c-shared-macos:
+	@tools/scripts/build-c-shared-macos.sh $(BUILD_FLAGS)
```
AndrewSisley: todo: There appears to be no reason for this to be in this PR, and it really does deserve it's own entry in the change-log IMO, it is something that people might go looking for.

Please remove this and the script from this PR.

internal/db/p2p/sync_branchable_col.go
```diff
@@ -0,0 +1,335 @@
...
+// SyncBranchableCollection initiates a request for the latest version of the branchable
+// collection's DAG from the network.
+//
+// This function call will block until there is a response for the collection.
+// It is the responsibility of the caller to set an appropriate timeout on the context.
+func (p *P2P) SyncBranchableCollection(ctx context.Context, collectionID string) error {
+	log.InfoContext(ctx, "Starting branchable collection sync",
+		corelog.String("PeerID", p.host.ID()))
+
+	cols, err := p.db.GetCollections(
```
AndrewSisley, line 55: suggestion: If you are fetching the local collections anyway, we can simplify the internal code, and provide more functionality by allowing users to provide a `client.CollectionFetchOptions` object themselves, and iterate through the returned `cols` (removing the `"collection is not branchable"` error below).

internal/db/p2p/sync_branchable_col.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		return err
+	}
+	if len(cols) == 0 {
+		return client.NewErrCollectionNotFoundForCollectionVersion(collectionID)
+	}
+
+	col := cols[0].Version()
+	if !col.IsBranchable {
+		return errors.New("collection is not branchable", errors.NewKV("CollectionID", collectionID))
```
AndrewSisley: todo: Please return an error from `errors.go` instead of declaring inline like this.

internal/db/p2p/sync_branchable_col.go
```diff
@@ -0,0 +1,335 @@
...
+			corelog.String("CID", cidStr))
+
+		syncedHeads[cidStr] = colCid
+	}
+
+	return nil
+}
+
+// syncCollectionAndMerge synchronizes a branchable collection from a remote peer and publishes a merge event.
+func (p *P2P) syncCollectionAndMerge(
```
AndrewSisley, line 207: todo: This does not appear to be doing any merging, please rename this, and tweak the `"Successfully synced and merged head"` log above as it is misleading.

tests/integration/net/sync/branchable_collection/complex_network_test.go
```diff
@@ -0,0 +1,396 @@
...
+						},
+					),
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+func TestBranchableCollectionSync_WithDocumentsFromPeers_ShouldHaveIdenticalDAG(t *testing.T) {
```
AndrewSisley, line 229: todo: There is something wrong with this test, or lack of companions and I am struggling to describe it to myself.

It passes because the `SyncBranchableCollection` actions run in serial.

Node 0 syncs all the available documents and collection heads from the other nodes and adds them on top of it's branchable.  This works, and is shown to work by other tests.  However, none of the tests show that the resultant DAG is deterministic - the order in which the documents are added to the collection might essentially be random, based on the order in which they are received over the network.

The 2nd, 3rd, and 4th `action.SyncBranchableCollection` actions for nodes 1, 2, and 3, will then (serially) fetch the latest collection commits from the other nodes.  Because node 0 already has the highest height, and contains the other document commits, they will take the node 0 head to be the source of truth, and *that* is why the test passes, and why the DAG appears consistent.  It does not show strong eventual consistency.

At least that is what it looks like is happening, I could be wrong - a few more tests should be able to prove/disprove this.  It could be as easy as hardcoding the expected cids and then see them change depending on the order in which blocks are received over the network.

tests/action/sync_branchable_collection.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+// SyncBranchableCollection is an action that syncs a branchable collection's DAG
+// from another node.
+type SyncBranchableCollection struct {
```
AndrewSisley, line 21: praise: Thank you for adding this over here, instead of a legacy action!

And thank you for documenting it well :)

AndrewSisley (changes requested):
The code was very readable, but I'm not sure it completely works - there might be a discussion to be had with the team RE whether or not strong eventual consistency is a requirement or not.

Makefile
islamaliev: removed

internal/db/p2p/sync_branchable_col.go
islamaliev: I'm not sure what you mean. As I understood you propose to expose `client.CollectionFetchOptions` in this method. This will give callers more flexibility in what collection they want to sync. But how will it lead to removing the "collection is not branchable" error? We still need to check it. The method is called `SyncBranchableCollection` after all

internal/db/p2p/sync_branchable_col.go
islamaliev: done

internal/db/p2p/sync_branchable_col.go
islamaliev: > does not appear to be doing any merging

what do you mean? It calls `p.db.Merge`

internal/db/p2p/sync_branchable_col.go
AndrewSisley: Ah I see what you mean - I made an assumption in my hastily scribbled design :)  I figured it would be fine to just `continue` on collections that are not branchable, as they are in a sense already synced.

This is even more questionable if in the future we add some kind of syncing mechanic for non-branchable collections though.

internal/db/p2p/sync_branchable_col.go
AndrewSisley: `p.db.Merge` doesn't really merge the collections.  It can I think merge documents in non-branchable collections, but for branchables I can't picture any merging being done ever by that function.

tests/integration/net/sync/branchable_collection/complex_network_test.go
islamaliev: Added nice tests

internal/db/p2p/p2p.go
```diff
@@ -167,7 +167,19 @@ func New(
 
 	host.SetBlockAccessFunc(p.hasAccess)
 
-	err := p.host.AddPubSubTopic(docSyncTopic, true, p.docSyncMessageHandler)
+	err := p.host.AddPubSubTopic(docSyncTopic, true, p.docSyncMessageHandler, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	err = p.host.AddPubSubTopic(syncBranchableCollectionTopic, true, p.syncBranchableCollectionMessageHandler,
```
AndrewSisley, line 191: todo: Please document why we are publishing events for this pub-sub topic, and not all of the other ones that we are subscribing to.

suggestion: I suggest removing the cast to `client.PubsubPeerEventHandler(`, that type is an alias and doesn't require a cast, and including the cast here makes an already very busy line harder to read.

islamaliev: done

internal/db/p2p/p2p.go
```diff
@@ -167,7 +167,21 @@ func New(
...
+		// for sync-branchable-collection topic peer events are tracked primarily for testing purposes
+		// because tests need a reliable way to know when peers joined the topic in order to proceed with the sync
```
AndrewSisley: todo: This IMO is not a good reason to change the production code.  You should be able to solve the test issue fairly easily by directly connecting to the relevant nodes in the network instead of relying on libp2p propagation - that propagation is not something that is really guaranteed anyway, at least in a large network, and shouldn't be relied upon for our tests.

We can (and do) test the propagation of connections in dedicated tests.  It should not be the responsibility of the SyncBranchableCollection tests to test this.

It is also very weird from a user perspective to only get these public events for the branchable collection topic and not all the other topics.

AndrewSisley (changes requested):
I had a look at new prop to `host.AddPubSubTopic` trying to review the go-p2p PR and think things need changing.

Happy to talk about it, maybe in the standup tomorrow if you prefer to wait.

internal/db/p2p/p2p.go
islamaliev: > directly connecting to the relevant nodes

what do you mean? Elaborate please.

The comment explain the original need for the event to exist, but it can be used by clients to do whatever they want to do.

> It should not be the responsibility of the SyncBranchableCollection tests to test this.

`SyncBranchableCollection` doesn't test it. It uses it pretty much the same other other actions use each other.

> only get these public events for the branchable collection topic and not all the other topics.

I'm sure it's going to be useful also for `SyncDocuments` and `SyncCollectionVersions` if we had right tests. But we can add them later.

AndrewSisley: > what do you mean? Elaborate please.

`TestBranchableCollectionSync_WithMultipleDocsInComplexLinkedNetwork_ShouldSyncAll` is flaky at the moment without this test-supporting production change because it is relying on libp2p to asynchronously propagate all the connections to all nodes from all the nodes that are not directly connected to in the tests.

If instead of relying on this 3rd party async functionality, you directly connect the nodes in the tests, the flakiness caused by the async connection propagation goes away as the tests are no longer reliant on it.

E.g. the `ConnectPeers` actions in `TestBranchableCollectionSync_WithMultipleDocsInComplexLinkedNetwork_ShouldSyncAll` would look like:
```
testUtils.ConnectPeers{
	SourceNodeID: 0,
	TargetNodeIDs: int[]{1, 2, 3},
},
testUtils.ConnectPeers{
	SourceNodeID: 1,
	TargetNodeIDs: int[]{0, 2, 3},
},
testUtils.ConnectPeers{
	SourceNodeID: 3,
	TargetNodeIDs: int[]{0, 1, 2},
},
```

This will solve the test flakiness without introducing test-only code to production.  The event stuff is probably useful to users if exposed in a sensible, consistent way for users, *and* tested, but that is not done well enough in this PR, and should not be, as with the proposed change above, it becomes totally irrelevant to the current issue/feature and can be implemented and tested independently.

islamaliev: direct peer connection doesn't solve this problem, the tests are still flaky when multiple peers are involved. For example the test `TestBranchableCollectionSync_WithDocumentsFromPeers_ShouldHaveIdenticalDAG` has 4 nodes directly connected to one another but it fails sometimes https://github.com/sourcenetwork/defradb/actions/runs/19677926282/job/56366060203

Peer connection establishment and information exchange on listened-to topics are different things.

AndrewSisley: Do you know why `TestBranchableCollectionSync_WithDocumentsFromPeers_ShouldHaveIdenticalDAG` fails more regularly than all our other existing P2P tests?

And roughly how regularly does it fail?

islamaliev: I'm not 100% sure, but I noticed that when we use more than 3 nodes in tests the flakiness increases drastically. As I understood after the node 0 establishes connection with all other nodes and immediately calls `SyncBranchableCollection` it's DHT doesn't always manage to get all the information about which peers are listening to the topic and therefore send it only to those that it knows.

AndrewSisley: Ah that's a shame, thanks for the explanation Islam.

In that case I think we should flesh out and test the feature properly - at the moment the public event is only fired off for one of many topics, and there are no dedicated tests for it (documenting the feature, as well as ensuring correctness).

My preference is to do that work in another PR and base this PR off of that - they (sync-cols, and the new events) are two different features, should have two different entries in the change log, and the smaller PRs will more focused instead of introducing the event-feature as a side effect of sync-col.

internal/db/p2p.go
```diff
@@ -292,3 +292,17 @@ func (db *DB) SyncCollectionVersions(ctx context.Context, versionIDs ...string)
...
 	return txn.Commit()
 }
+
+// SyncBranchableCollection requests the latest version of the branchable collection's DAG
+// from the network and synchronizes it locally. This syncs the collection-level history
+// for branchable collections (collections marked with @branchable directive).
+// It doesn't automatically subscribe to the collection for future updates.
+// context.WithTimeout can be used to set a timeout for the operation.
+//
+// WARNING: This function does not respect transactions.
```
AndrewSisley, line 302: thought: Don't worry about this now, idea just popped into my head - for this and the similar funcs, I think we can and should respect transactions (implicit and explicit) for the `datastore` related operations.  The syncing of blocks should be the only bit to ignore a transaction.

internal/db/p2p/sync_branchable_col.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	waitCtx := ctx
+	if _, hasDeadline := ctx.Deadline(); !hasDeadline {
+		var cancel context.CancelFunc
+		waitCtx, cancel = context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+	}
+
+	return p.waitAndHandleSyncBranchableCollectionResponse(waitCtx, collectionID, pubSubRespChan)
```
AndrewSisley: question: This function looks like it will always block for 5 seconds, unless the user has already provided their own timeout, or an error is returned.  Once the context timeout has been breached, syncing will cease, even if more responses have been received locally and are queued for processing (the processing appears limited to a single routine, so they will be handled one at a time).

This feels inconsistent with the user-visible documentation, which simply notes that a timeout may be provided.  This is not how the collection version sync works.  How does doc sync work?

Have I misread the code?  Is this the behaviour we want? 

todo: If it does work like this, and we want this, I think the user-visible documentation needs to be clearer.

internal/db/p2p/sync_col_versions_js.go
```diff
@@ -8,6 +8,8 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+//go:build js
```
AndrewSisley: todo: This line is not required.  The build tag is handled by the `_js` in the file name.  I think it is horrible that Go does this, but it is standard, and this new line is essentially dead code.

AndrewSisley (changes requested):
Code looks great, I just have one fairly large question that I'd like answered before approving :)

internal/db/p2p/sync_branchable_col.go
islamaliev: collection version sync works entirely different then the other 2: there you provide the CID that you want to fetch.
With docs and branch-col sync we ask the network for all known blocks.
The docs sync "worked" because we don't have complex tests for it. We requesting sync for doc1, after the first response from any peer we stop. But what if right after it another response comes from another peer with with a more up-to-date doc version, it will be ignored. I even created an issue for it https://github.com/sourcenetwork/defradb/issues/4265

We probably should discuss it in the team.

AndrewSisley (approved):
LGTM, approving now assuming that the timeout is discussed in the the standup or discord before merge - thanks Islam!

PR #4200, opened 2025-11-19
feat: Allow deletion of collections via PatchCollection

### LINKED ISSUE #4181: Allow deletion of empty collections via `PatchCollection`

AndrewSisley:
## Relevant issue(s)

Resolves #4181 

## Description

Allows the deletion of collections via `PatchCollection`.

### DIFF
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -735,16 +735,7 @@ func collectionHasDocuments(
 		return false, errors.Join(err, iter.Close())
 	}
 
-	if hasValue {
-		return true, iter.Close()
-	}
-
-	err = iter.Close()
-	if err != nil {
-		return false, err
-	}
-
-	return false, nil
+	return hasValue, iter.Close()
 }
 
 func validateCollectionDoesNotHaveHigherVersion(
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -700,6 +700,12 @@ func collectionHasDocuments(
 	ctx context.Context,
 	version client.CollectionVersion,
 ) (bool, error) {
+	if !version.IsMaterialized {
+		// Assume that if the collection *was* materialized, and is no longer materialized, that the cached
+		// state was properly disposed of (it should have been).
+		return false, nil
+	}
+
 	txn := datastore.CtxMustGetTxn(ctx)
 
 	shortID, err := id.GetShortCollectionID(ctx, version.CollectionID)
@@ -716,30 +722,26 @@ func collectionHasDocuments(
 		}
 	}
 
-	// Assume that if the collection *was* materialized, and is no longer materialized, that the cached
-	// state was properly disposed of (it should have been).
-	if version.IsMaterialized {
-		iter, err := txn.Datastore().Iterator(ctx, corekv.IterOptions{
-			Prefix:   prefixKey.ToDS().Bytes(),
-			KeysOnly: true,
-		})
-		if err != nil {
-			return false, errors.Join(err, iter.Close())
-		}
+	iter, err := txn.Datastore().Iterator(ctx, corekv.IterOptions{
+		Prefix:   prefixKey.ToDS().Bytes(),
+		KeysOnly: true,
+	})
+	if err != nil {
+		return false, errors.Join(err, iter.Close())
+	}
 
-		hasValue, err := iter.Next()
-		if err != nil {
-			return false, errors.Join(err, iter.Close())
-		}
+	hasValue, err := iter.Next()
+	if err != nil {
+		return false, errors.Join(err, iter.Close())
+	}
 
-		if hasValue {
-			return true, iter.Close()
-		}
+	if hasValue {
+		return true, iter.Close()
+	}
 
-		err = iter.Close()
-		if err != nil {
-			return false, err
-		}
+	err = iter.Close()
+	if err != nil {
+		return false, err
 	}
 
 	return false, nil
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -18,15 +18,21 @@ import (
 	"unicode"
 
 	jsonpatch "github.com/evanphx/json-patch/v5"
+	"github.com/ipfs/go-cid"
 
+	"github.com/sourcenetwork/corekv"
 	"github.com/sourcenetwork/immutable"
 	"github.com/sourcenetwork/lens/host-go/config/model"
 
 	"slices"
 
 	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/description"
+	"github.com/sourcenetwork/defradb/internal/db/id"
+	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
 func (db *DB) createCollections(
@@ -169,6 +175,21 @@ func (db *DB) patchCollection(
 		return err
 	}
 
+	removedCollectionVersions := []client.CollectionVersion{}
+existingVersionLoop:
+	for versionID, version := range existingColsByID {
+		if _, ok := newColsByID[versionID]; !ok {
+			for _, newCol := range newColsByID {
+				if newCol.VersionID == versionID {
+					// If the missing version id is found in another location, we do not wish to delete the collection,
+					// it has essentially been moved by the JSON patch for reasons known only to the user.
+					continue existingVersionLoop
+				}
+			}
+			removedCollectionVersions = append(removedCollectionVersions, version)
+		}
+	}
+
 	for _, col := range newColsByID {
 		// Automatically add any id fields for object fields added by the patch, if the patch did not explicitly
 		// add one.
@@ -272,7 +293,29 @@ func (db *DB) patchCollection(
 		return err
 	}
 
+	err = db.deleteCollectionVersions(ctx, removedCollectionVersions)
+	if err != nil {
+		return err
+	}
+
 	for _, col := range newCollections {
+		isDeleted := false
+		for _, removedCol := range removedCollectionVersions {
+			if col.VersionID == removedCol.VersionID {
+				isDeleted = true
+				break
+			}
+		}
+		if isDeleted {
+			// We need to make sure we dont save any collections that we have just deleted.
+			// This check is needed due to the unfortunate way mutated collections have their
+			// originals re-added to `newCollections` on line 260.
+			//
+			// This re-adding, and this check, are planned to be removed post v1 in issue:
+			// https://github.com/sourcenetwork/defradb/issues/4197
+			continue
+		}
+
 		existingCol, ok := existingColsByID[col.VersionID]
 		if ok && col.Equal(existingCol) {
 			continue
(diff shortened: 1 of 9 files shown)


internal/db/description/collection.go
```diff
@@ -459,3 +459,58 @@ func GetRelatedCollection(
...
+
+	txn := datastore.CtxMustGetTxn(ctx)
+
+	key := keys.NewCollectionKey(version.VersionID)
+	err = txn.Systemstore().Delete(ctx, key.Bytes())
+	if err != nil {
+		return err
+	}
+
+	if version.IsActive {
```
islamaliev, line 483: question: if you delete the active collection shouldn't we active the previous one? I thought we should always have an active version. Explain please how it works

EDIT: by further reading the PR I noticed that it will just be left as un-queriable, but lack of tests leaves me wondering what this state really means

internal/db/description/collection_cache.go
```diff
@@ -148,6 +148,47 @@ func (cache *collectionCache) Add(col client.CollectionVersion) {
...
+
+	if cols, ok := cache.CollectionsByID[version.CollectionID]; ok {
+		var indexToRemove int
+		for i, col := range cols {
+			if col.VersionID == version.VersionID {
+				indexToRemove = i
+				break
+			}
+		}
+		cache.CollectionsByID[version.CollectionID] = append(cols[:indexToRemove], cols[indexToRemove+1:]...)
```
islamaliev, line 162: suggestion: 
```go
slices.DeleteFunc(cols, func(c client.CollectionVersion) bool {
    c.VersionID == version.VersionID
}
```
And similar for deletion below

internal/db/id/field.go
```diff
@@ -153,6 +168,49 @@ func SetShortFieldIDs(ctx context.Context, collection client.CollectionVersion)
...
+		if field.FieldID == "" {
+			// Short field IDs only exist for fields with full ids
+			continue
+		}
+
+		if _, ok := previouslyExistingFieldIDs[field.FieldID]; ok {
+			// We should only delete short IDs for fields that are unique to this version,
+			// because short IDs remain consistent across all collection versions.
+			continue
+		}
```
islamaliev, line 203: question: what if the active version in the middle of history and the current field was introduced in the version that we are deleting? Then deletion of the field should be legit, no?

EDIT: ignore this comment. I saw that it's not allowed from tests

internal/db/collection_define.go
```diff
@@ -272,7 +293,29 @@ func (db *DB) patchCollection(
...
+		for _, removedCol := range removedCollectionVersions {
+			if col.VersionID == removedCol.VersionID {
+				isDeleted = true
+				break
+			}
+		}
+		if isDeleted {
+			// We need to make sure we dont save any collections that we have just deleted.
+			// This check is needed due to the unfortunate way mutated collections have their
+			// originals re-added to `newCollections` on line 260.
```
islamaliev, line 312: todo: please do not reference any line numbers. This code changes quite ofter and even now I think you meant line 259. If we had the logical blocks of this huge function extracted into smaller function, it would make sense to reference a function.

internal/db/collection_define.go
```diff
@@ -272,7 +293,29 @@ func (db *DB) patchCollection(
...
+		isDeleted := false
+		for _, removedCol := range removedCollectionVersions {
+			if col.VersionID == removedCol.VersionID {
+				isDeleted = true
+				break
+			}
+		}
+		if isDeleted {
+			// We need to make sure we dont save any collections that we have just deleted.
+			// This check is needed due to the unfortunate way mutated collections have their
```
islamaliev, line 311: suggestion: I didn't quite get what you mean here. Would you rephrase please?

internal/db/collection_define.go
```diff
@@ -578,3 +621,184 @@ func (db *DB) shouldReindexForVersionSwitch(
...
+		prefixKey = keys.NewViewCacheColPrefix(shortID)
+	} else {
+		prefixKey = keys.PrimaryDataStoreKey{
+			CollectionShortID: shortID,
+		}
+	}
+
+	// Assume that if the collection *was* materialized, and is no longer materialized, that the cached
+	// state was properly disposed of (it should have been).
+	if version.IsMaterialized {
```
islamaliev: suggestion: change into early return and move to the top of the funcion

internal/db/collection_define.go
```diff
@@ -578,3 +621,184 @@ func (db *DB) shouldReindexForVersionSwitch(
...
+		}
+
+		if hasValue {
+			return true, iter.Close()
+		}
+
+		err = iter.Close()
+		if err != nil {
+			return false, err
+		}
```
islamaliev: suggestion:
```go
return hasValue, iter.Close()
```

internal/db/collection_define.go
```diff
@@ -578,3 +621,184 @@ func (db *DB) shouldReindexForVersionSwitch(
...
+func validateCollectionDoesNotHaveHigherVersion(
+	ctx context.Context,
+	version client.CollectionVersion,
+) error {
+	allVersions, err := description.GetCollectionsByCollectionID(ctx, version.CollectionID)
+	if err != nil {
+		return err
+	}
+
+	for _, newVersion := range allVersions {
```
islamaliev, line 750: nitpick: `newVersion` is probably not the right name here

internal/db/collection_define.go
```diff
@@ -578,3 +621,184 @@ func (db *DB) shouldReindexForVersionSwitch(
...
+		return false, err
+	}
+
+	var prefixKey keys.Key
+	if version.Query.HasValue() {
+		prefixKey = keys.NewViewCacheColPrefix(shortID)
+	} else {
+		prefixKey = keys.PrimaryDataStoreKey{
+			CollectionShortID: shortID,
+		}
```
islamaliev, line 722: question: I might have missed some recent changed but do we allow now documents hard deletion? As I recall we did only soft deletes, so this would always return docs even if they are "deleted"

tests/integration/collection_version/updates/remove/collections_test.go
```diff
@@ -39,12 +40,583 @@ func TestColVersionUpdateRemoveCollections(t *testing.T) {
...
+				}`,
+				ExpectedError: `Cannot query field "Users" on type "Subscription".`,
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestColVersionUpdateRemoveCollectionWithData(t *testing.T) {
```
islamaliev, line 144: suggestion: change test name to follow our convention.

Like this `TestRemoveCollection_IfHasDocs_ReturnError`

tests/integration/collection_version/updates/remove/collections_test.go
```diff
@@ -39,12 +40,583 @@ func TestColVersionUpdateRemoveCollections(t *testing.T) {
...
+							"op": "copy",
+							"from": "/Users",
+							"path": "/UsersV2"
+						},
+						{
+							"op": "add", "path": "/UsersV2/Fields/-", "value": {"Name": "email", "Kind": 11}
+						},
+						{
+							"op": "remove",
+							"path": "/Users"
```
islamaliev, line 238: question: weren't you supposed to remove `/UsersV2`?

tests/integration/collection_version/updates/remove/collections_test.go
```diff
@@ -39,12 +40,583 @@ func TestColVersionUpdateRemoveCollections(t *testing.T) {
...
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{
+							"op": "remove",
+							"path": "/bafyreiciz2hrrmt7ritk5gf5fyruw46v2tfhq5dc7qto4wgpzluben2smu"
+						}
+					]
+				`,
+				ExpectedError: "cannot delete a version that is used by a newer version, first delete the new version",
```
islamaliev, line 333: suggestion: would be nice to use text from the error itself so that if we change the error text the test doesn't break

tests/integration/collection_version/updates/remove/collections_test.go
```diff
@@ -39,12 +40,583 @@ func TestColVersionUpdateRemoveCollections(t *testing.T) {
...
+				Patch: `
+					[
+						{
+							"op": "add", "path": "/Users/Fields/-", "value": {"Name": "email", "Kind": 11}
+						}
+					]
+				`,
+			},
+			testUtils.PatchCollection{
+				// Remove the active version, leaving the collection un-queriable
```
islamaliev, line 361: todo: add a test that show what happens when you patch such an un-queriable collection

tests/integration/collection_version/updates/remove/collections_test.go
islamaliev, line 1: todo: there should be tests with deleting a collection that has deleted docs

tests/integration/collection_version/updates/remove/view_test.go
```diff
@@ -0,0 +1,323 @@
...
+				},
+				ExpectedResults: []client.CollectionVersion{},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestColVersionUpdateRemoveNonMaterializedViewWithData(t *testing.T) {
```
islamaliev, line 67: suggestion: would be nice to have tests in our agreed-upon naming convention

`TestRemoveView_IfNotMaterializedWithData_ShouldRemove`

tests/integration/collection_version/updates/remove/view_test.go
```diff
@@ -0,0 +1,323 @@
...
+				},
+				ExpectedResults: []client.CollectionVersion{},
+			},
+			testUtils.Request{
+				Request: `query {
+					UserView {
+						name
+					}
+				}`,
+				ExpectedError: `key not found`,
```
islamaliev, line 257: note: this is the worst ever, it gives 0 value. I'm not sure if it's out of scope for this task.

islamaliev, line 257: question: should this return something like 	`Cannot query field "UserView" on type "Query".` like with normal collection that don't exist?

tests/integration/collection_version/updates/remove/view_test.go
```diff
@@ -0,0 +1,323 @@
...
+			testUtils.PatchCollection{
+				Patch: `
+					[
+						{
+							"op": "remove",
+							"path": "/Users"
+						}
+					]
+				`,
+			},
```
islamaliev, line 299: question: is this the desired end behaviour? Do we want to allow view's source collection to be removed? Or is it just temporary solution? 
What are the implications? Will the materialized view just linger in storage?

islamaliev (changes requested):
Looks good overall. I'd like some questions answered before the merge and also requested couple change (mostly tests).

internal/db/description/collection_cache.go
AndrewSisley: I'm not sure I like this yet :)

It is more expensive (`slices.DeleteFunc` does not `break` on match), and I like the explicitness of the raw way (readers do not need to learn a new function).

Thanks for the suggestion though, I didn't know this existed.

internal/db/collection_define.go
AndrewSisley: I'm going to leave this as-is for now.

I hate the line number reference too, for the same reasons.  It is however also linked by variable name, and git-commit, and I couldn't think of a better way.

It is this or no comment IMO.

internal/db/description/collection_cache.go
islamaliev: there is nothing to learn there. It's very obvious. And now that it's in standard library, people should know it.
About early break you are right, but are there going to be that many collections to go though?
Anyway, no need to discuss it further. It's a minor suggestion. Use your best judgement.

internal/db/collection_define.go
AndrewSisley: Good shout, is weird I ended up coding it this way, thanks Islam :)

- [x] Move ` version.IsMaterialized`

internal/db/collection_define.go
AndrewSisley: This will make sense when I move the ` version.IsMaterialized` check  - thanks for flagging :)

- [x] simplify return

internal/db/collection_define.go
AndrewSisley: It kind of is :) We are looking for version higher up in the dag, it can only ever be newer than previous version.

I say kind of, because we havent found that it is the newer version until the `if`.

Leaving as is as it is not worth spending time on IMO.

internal/db/collection_define.go
islamaliev: you can at least say "few lines above". Nobody is going to maintain this reference up-to-date.
Or we can create a marker-comment like `// reAddingLabel` which you can use in this comment, which in my opinion is stupid and I never saw it in practice, but it solves the problem of maintaining references up-to-date.
This is a general problem of having large functions. It's well-known code smell. And this function is going to grow even larger with time.

internal/db/collection_define.go
AndrewSisley: > do we allow now documents hard deletion?

We do not, yet.

>  this would always return docs even if they are "deleted"

If documents have been added.  If the collection is empty it will return no docs and can be deleted.

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: Leaving as-is, the only conversation I recall around the `_ReturnError` convention was that it was entirely optional.

internal/db/collection_define.go
islamaliev: > If documents have been added. If the collection is empty it will return no docs and can be deleted

Are you sure this is what you wanted to say? I think there is a mistake here, because I didn't get your point

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: Nope, The original is being removed.

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: I would like the test to break in that case, and I like to see the error text when reading the test.

tests/integration/collection_version/updates/remove/collections_test.go
islamaliev: FYI: for me personally this style of naming is very hard to hear and it's easier for me to just ready the test body which I actually did. 80% of the name is just a common prefix, it contains almost no useful information

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: I'm not sure I understand your comment.

There are existing tests for collections with only inactive versions.

 > what happens when you patch such an un-queriable collection
 
Patching an unqueriable collection works exactly like patching a queriable collection, there is no reason to duplicate tests for this.  There are tests that test the patching of specific properties on such collections though, such as toggling `IsActive`.

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: Not a problem, I'll add one.

- [x] Add a test with a deleted doc

tests/integration/collection_version/updates/remove/view_test.go
AndrewSisley: Covered by earlier comment, resolving.

tests/integration/collection_version/updates/remove/view_test.go
AndrewSisley: Covered by previous comment, this is out of scope, and is of course not ideal :)

tests/integration/collection_version/updates/remove/collections_test.go
islamaliev: it sounds like your personal preference which is strange to me. Changing error text should not be the reason for this test to break. It should be the behaviour of the system. 

What if you had 1000 of such tests? That would be fun. It creates somewhat similar issue to changing CIDs for tests. I don't think you like to adjust those.

tests/integration/collection_version/updates/remove/view_test.go
AndrewSisley: There is no direct link between a view and the collection(s) that it queries.  The view queries by query string, and which collection that query string fetches documents is free to change over time at the moment.

This is also not a new problem - you can create a view that queries a non-existent collection.  And such a view can be received over P2P before its target is received/created.

> Will the materialized view just linger in storage?

Yes, and this is probably desirable in a few use cases.

internal/db/collection_define.go
AndrewSisley: Yes :)

If you create a collection, until documents are added to it, it will have no documents.  During this time period it can be deleted.

tests/integration/collection_version/updates/remove/collections_test.go
AndrewSisley: > Changing error text should not be the reason for this test to break. It should be the behaviour of the system.

Changing error text is a semver breaking change and should cause the tests to fail.  The exact text returned *is* the behaviour of the system.  It is also what gets acted upon when calling `errors.Is`.

> What if you had 1000 of such tests?

If we had a million tests repeating the same error text it would still be a single find and replace. It is not the same as the CID changes, and I'm not sure you understand the problems we've been having with those.

internal/db/collection_define.go
AndrewSisley: I'm going to resolve this discussion thread as it is regarding a relatively unimportant part of the code, and at the moment the conversation is not headed in a constructive direction.

internal/db/description/collection.go
AndrewSisley: > if you delete the active collection shouldn't we active the previous one?

No.  A collection does not need to have an active version.  This is not new to this PR.

> Explain please how it works

If a collection has no active versions it will not be queriable.  It may still have data.  Longer term it should be accessible by explicitly targeting the collection version. 

> but lack of tests leaves me wondering what this state really means

There are not many tests for this in this PR because it is not a new feature.  There are tests for this if you go looking in develop, although given the v0.20 pressure-experiment there may only be a handful of them.

fredcarle: > It may still have data

Didn't we say that this shouldn't be allowed?

AndrewSisley: > Didn't we say that this shouldn't be allowed?

I do not recall such a conversation.  And I'm not sure why we would want that.  Either way, IMO that is out of scope of this PR by quite a way - can open a discord thread if you like?

internal/db/collection_define.go
AndrewSisley: This is a tricky concept perhaps.  It is referring to some tech debt, this tech debt has started to cause problems and should be removed (ticket linked).

The linked issue doesn't directly correspond to this line, instead resolving the issue will naturally cause this line to disappear - the (corrected) internal model will not permit this line to have any meaning, as well as solving a couple of other pain points.

The code is perhaps the best description of the problem, for this comment to properly describe it would take up a large amount of space and would not aid 90% of readers of this function.

I can remove the comment, or shrink it if you like?  I do understand that it may cause more alarm to readers than is perhaps warranted.

internal/db/description/collection.go
fredcarle: We definitely said that we shouldn't delete a collection that has documents. I'm trying to find the thread but it might have been in a standup.

AndrewSisley: Ah okay, I think you may have quoted the wrong block of text then, or there were multiple and I found the wrong one :laughing: 

We do not permit the deletion of collections with documents.

internal/db/collection_define.go
islamaliev: use your best judgement

islamaliev (approved):

PR #4206, opened 2025-11-21
feat: Add ActivePeers function

### LINKED ISSUE #4159: List connected peers
From build week feedback: https://discord.com/channels/1374835078781468722/1433455098591187138

> At first I thought this was a documentation issue, but looking through the defra code there is no public way of doing this.  It would be very useful.
> 
> It is possible to fetch the configured replicators, and but not peers.
> 
> Without this, users are left guessing as to whether they are connected to anything and may end up aggressively spamming `Connect` to try and make sure.

AndrewSisley:
## Relevant issue(s)

Resolves #4159

## Description

Adds the ActivePeers function, returning the set of current connections.

## Tasks

- [x] Update dependency once https://github.com/sourcenetwork/go-p2p/pull/13 is merged

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/sourcenetwork/corekv/memory v0.2.4
 	github.com/sourcenetwork/corekv/namespace v0.2.4
 	github.com/sourcenetwork/corelog v0.0.8
-	github.com/sourcenetwork/go-p2p v0.1.4
+	github.com/sourcenetwork/go-p2p v0.1.5
 	github.com/sourcenetwork/goji v0.0.8
 	github.com/sourcenetwork/graphql-go v0.7.10-0.20241003221550-224346887b4a
 	github.com/sourcenetwork/immutable v0.3.0
--- a/internal/db/p2p/p2p.go
+++ b/internal/db/p2p/p2p.go
@@ -206,6 +206,10 @@ func (p *P2P) PeerInfo() ([]string, error) {
 	return p.host.Addresses()
 }
 
+func (p *P2P) ActivePeers(ctx context.Context) ([]string, error) {
+	return p.host.ActivePeers()
+}
+
 // Connect initiates a connection to the peer with the given addresses.
 func (p *P2P) Connect(ctx context.Context, addresses []string) error {
 	return p.host.Connect(ctx, addresses)
--- a/tests/action/active_peers.go
+++ b/tests/action/active_peers.go
@@ -29,7 +29,7 @@ type ActivePeers struct {
 	// The expected set of results.
 	//
 	// Respects `replace`, and should typically be provided a string similar to
-	// `{{.Peer1Address0}}`.
+	// `{{.Peer1_Address0}}`.
 	//
 	// The order of elements in the given slice is not asserted.
 	Expected []string
--- a/tests/action/active_peers.go
+++ b/tests/action/active_peers.go
@@ -23,10 +23,8 @@ import (
 type ActivePeers struct {
 	stateful
 
-	// NodeID may hold the ID (index) of a node to get active peers for.
-	//
-	// If a value is not provided the update will be applied to all nodes.
-	NodeID immutable.Option[int]
+	// NodeID holds the ID (index) of a node to get active peers for.
+	NodeID int
 
 	// The expected set of results.
 	//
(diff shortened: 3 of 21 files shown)


tests/action/active_peers.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// NodeID may hold the ID (index) of a node to get active peers for.
+	//
+	// If a value is not provided the update will be applied to all nodes.
+	NodeID immutable.Option[int]
+
+	// The expected set of results.
+	//
+	// Respects `replace`, and should typically be provided a string similar to
+	// `{{.Peer1Address0}}`.
```
islamaliev: suggestion: I think `Peer1_Address0` would be visually more distinguashable.

tests/action/active_peers.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ActivePeers is an action that will get the active peers from the given node(s)
+// and assert that the given expect result matches the actual.
+type ActivePeers struct {
+	stateful
+
+	// NodeID may hold the ID (index) of a node to get active peers for.
+	//
+	// If a value is not provided the update will be applied to all nodes.
+	NodeID immutable.Option[int]
```
islamaliev: todo: "update will be applied to all nodes" seems to be copy&pasted.

Also I don't think `NodeID` should be optional. I don't see a point of requesting actives peers by all nodes.

internal/db/p2p/p2p.go
```diff
@@ -206,6 +206,10 @@ func (p *P2P) PeerInfo() ([]string, error) {
 	return p.host.Addresses()
 }
 
+func (p *P2P) ActivePeers(ctx context.Context) ([]string, error) {
+	return p.host.ActivePeers()
```
islamaliev, line 210: question: didn't you say libp2p returns a list of peers including the current peers?

I though we had a consensus that we don't want it included.

go.mod
```diff
@@ -412,3 +412,4 @@ require (
 )
 
 replace github.com/wlynxg/anet => github.com/sourcenetwork/anet v0.0.0-20250417190629-7c87cba7799e
+replace github.com/sourcenetwork/go-p2p => ../p2p
```
islamaliev: todo: this is not supposed to be committed.

islamaliev (changes requested):
Looks good, Andy! I disapprove for now to make sure few minor things are addressed.

tests/action/active_peers.go
AndrewSisley: I think you are right, it didn't like a `.`, but hopefully an `_` will be fine :)

- [x] Tweak replace format

tests/action/active_peers.go
AndrewSisley: > todo: "update will be applied to all nodes" seems to be copy&pasted.

Thanks... And sorry for the bother

- [x] Fix copy-pasta

> I don't see a point of requesting actives peers by all nodes.

Now that the host is excluded I think you are right, as they should all be unique anyway.  I think at some point they will have all been consistent, unless I was just on auto-pilot adding this :)

- [x] Remove optional node id

internal/db/p2p/p2p.go
AndrewSisley: This filtering is done in the `go-p2p` package - the link to the PR (with filtering) is in this PR's description

go.mod
AndrewSisley: Correct, this is noted in the PR description and will be removed before merge.  The CI will not pass until it is removed.

tests/integration/net/info/connect_peers_test.go
```diff
@@ -0,0 +1,125 @@
...
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+func TestNetInfoPeers_NoP2PConfigured(t *testing.T) {
+	test := testUtils.TestCase{
+		SupportedClientTypes: immutable.Some(
+			// Everything besides the JS client is supported, as the JS client does not have
+			// an `ActivePeers` function to call.
+			[]state.ClientType{
```
islamaliev, line 29: note: would be nice to have something like `NotSupportedClientTypes` where we explicitly list them. This way when a new client gets added it won't be automatically skipped.

islamaliev (approved):

tests/integration/net/info/connect_peers_test.go
AndrewSisley: Absolutely, and I miss that a lot - the new framework has it :)

PR #4210, opened 2025-11-25
feat: Support UTC_NOW as a DateTime value

### LINKED ISSUE #4152: Support current time for DateTime default value
From build week feedback: https://discord.com/channels/1374835078781468722/1433527809254359192

> Most DBs have a feature to set a date time field to the current time automatically on creation and update. We could most likely support this via a special default field value such as `now` or `current_time`.

ChrisBQu:
## Relevant issue(s)

Resolves #4152 

## Description

This allows the use of a literal value, `UTC_NOW` to be used as a value for `DateTime` objects. For example, users will be able to do the following:

```
defradb client schema add '
 type ObjectWithTimestamp {
   timestamp: DateTime
 }
'

defradb client query '
 mutation {
     create_ObjectWithTimestamp(input: {timestamp: UTC_NOW}) {
         _docID
     }
 }
 `
 ```
 
 The logic accommodating this actually contained within the following two PRs from our `graphql-go` fork.
 
 https://github.com/sourcenetwork/graphql-go/pull/15
 https://github.com/sourcenetwork/graphql-go/pull/16
 
 This PR changes are `go.mod` file to point at the correct commit. It also changes three integration tests that utilize the new functionality.

As discussed in standup, this PR leaves a following issue to be addressed: ensuring that all instances of `UTC_NOW` inside of a single transaction resolve to the same timestamp.

## Tasks

## How has this been tested?

- WSL

### CODE REVIEW

**Review by jsimnz — CHANGES_REQUESTED** [review]
As we discussed on our call, everything looks great so far. 

However I have since realized that this wont work for `@default` directives AFAICT. The default values are saved to the collection, but are saved based on the output of `scalar.ParseLiteral` which in the case of `UTC_NOW` for `DateTime` scalar will produce the *current* time, as in the time the collection is created and the default values cached, not the time of the mutation.

I don't know what the most obvious answer to this is, short of adding some annoying/specific code to the default system to handle `UTC_NOW` different from other default values.

I would consider this a blocker on this PR. If we cant find an easy solution, then I would specifically prohibit `UTC_NOW` as a valid default value for the directive, and we document that

**Comment by getify** [comment]
on the topic of `UTC_NOW` being resolved at the server instead of client...

if I fire off two create mutations, in order, from a client... and there's a `createdAt` field in the document that I set to `UTC_NOW`... if those arrive at the server in a different order (and possibly delay between them), the timestamps will be filled in by the server, countering the expectation of what the client asked for.

is that accurate?

in this way, it's sort of a "CRDT" function rather than a scalar value, no?

**Comment by jsimnz** [comment]
@getify this seems to be the desirable behavior and expected from any "now" like semantic provided by a server. If the client wants things from it's perspective it's free to do so. 

This seems how any other SQL and NonSQL DB handles their "NOW" semantics.

Is there something Im missing/not understand the concern?

**Comment by getify** [comment]
`UTC_NOW` syntactically presents as a variable that is filled in, and I think that's part of the problem. It leads to the presumption that it's filled by the client, as opposed to being filled in by the server.

Contrast that with `UTC_NOW()` that looks like a function that's being called. Indeed, the other databases we compared to, such as SQLite and MySQL, present their timestamping as a function: `NOW()`, `CURRENT_TIMESTAMP()`, etc.

Since it's behaving like a function (deferred until it reaches the server), I was thinking it would be better for it to syntactically look like that function call. And I also thought that maybe this might offer an alternative path for how the @default handling you were discussing would be done.

And while I'm at making suggestions here, `NOW(UTC)` (where `NOW()` is the function and `UTC` is like an enum that is activating what kind of timestamp to obtain) reads better to me than `UTC_NOW` / `UTC_NOW()`.

**Comment by jsimnz** [comment]
I understand the original point now, thanks for the clarification. The downside we're facing is that we are painfully limited by GraphQL syntax, which unfortunately doesn't offer us any sort of function like declaration/invocation in the naive and straightforward sense. 

Going down this line of reasoning for completeness sake, we could just use a regular string value and present that as the "function" call. So it would be `mutation(input: {createdAt: "utc_now()"})` (or whatever we would want the function/args to be), in comparison to `mutation(input: {createdAt: UTC_NOW})`.

We opted for the enum value `UTC_NOW` without additional quotes since it felt more *intentional* and *special* relative to a regular string.

I'm not a huge fan of the string approach (function invocation syntax or not), and I feel like this can be clearly established in the docs.

Its open for debate how people might interpret the semantics without additional docs/context, but my gut says it'll skew to people interpreting this from the servers POV, not the clients.

File: tests/integration/mutation/create/field_kinds/date_time_test.go [hunk 1]
```diff
@@ -104,3 +104,32 @@ func TestMutationCreateFieldKinds_WithDateTimesNanoSecondsAppart(t *testing.T) {
...
+			testUtils.Request{
+				Request: `mutation {
+                    create_User(input: {time: UTC_NOW}) {
+						time
+                    }
+                }`,
+				Results: map[string]any{
+					"create_User": []map[string]any{
+						{
+							"time": testUtils.CurrentTimestamp(),
```
islamaliev, line 127: suggestion: you could you `gomega` matcher instead to be more explicit:
```go
"time": BeTemporally("~", time.Now().UTC(), 5*time.Second)
```

https://pkg.go.dev/github.com/onsi/gomega#BeTemporally

ChrisBQu: I wasn't aware of this, but it's exactly what I want. I tried it locally, and it works. Thank you for this.

Edit: Actually, I spoke a little too soon. This won't work in all places. The `gomega` matcher wants to compare a `time.Time` but sometimes our result is a string. I think even if we used `gomega`'s logic, we would have to wrap it in a lightweight custom matcher(?) In which case, I'm not sure we gain much, but maybe you disagree?

internal/request/graphql/schema/collection.go
```diff
@@ -38,6 +38,9 @@ const (
 	typeString   string = "String"
 	typeBlob     string = "Blob"
 	typeJSON     string = "JSON"
+
+	// Special case directives
+	directive_UTC_NOW string = "UTC_NOW"
```
jsimnz: suggestion: Careful using the word `directive` here, given its location, people might assume this will be used to define a new directive (such as the `@explain` directive). In actuality this is a special value used in GQL inputs. Dont have an exact suggestion on an alternative name

jsimnz (approved):
LGTM. Like it!

I wish Islams suggested worked out in all cases, since its so simple, but the string type limitation is a blocker.

Good test coverage

internal/request/graphql/schema/collection.go
getify: is "enum_" appropriate? or "keyword_"?

ChrisBQu: I can see how it would potentially confuse someone. It's kind of awkward that it has to exist at all -- it's just there to appease the linter, by not using the same literal twice. I don't think its name will have far-reaching implications. I am happy to make whatever change to its name people feel is appropriate.

I am okay with "enum_" if you guys are.

jsimnz: that works 👍

PR #4212, opened 2025-11-26
refactor: Remove p2p transactions

### LINKED ISSUE #4211: P2P system uses a lot of transactions
Discovered this when trying to implement the LevelDB backed key-store. Since only 1 transaction can be active at a time it causes all p2p tests to fail. I believe most of these transactions can be removed without any down-side.

nasdf:
## Relevant issue(s)

Resolves #4211

## Description

This PR removes as many transactions from the p2p system as possible to allow for the LevelDB store to pass the integration tests.

This is part of a stacked PR and will be undrafted once I open the other PRs.

## Tasks

## How has this been tested?

Integration tests

- MacOS


fredcarle (approved):
Hopefully this doesn't create problems down the road but in any case I want to improve this code after v1. And the only place I can see this causing issues is on retries which I'm not too concerned with at the moment.

islamaliev (approved):

shahzadlone (approved):
LGTM

internal/db/p2p/collection.go
```diff
@@ -122,7 +122,10 @@ func (p *P2P) GetAllP2PCollections(ctx context.Context) ([]string, error) {
 	ctx, span := tracer.Start(ctx)
 	defer span.End()
 
-	iter, err := p.db.Multistore().Systemstore().Iterator(ctx, corekv.IterOptions{
+	clientTxn := datastore.CtxMustGetClientTxn(ctx)
```
fredcarle, line 125: question: What caused the walk back on this one?

nasdf: My goal was to remove only the background process transactions. This one is user invoked.

fredcarle: Ah yes that makes sense. Thanks.

Issue #2968, opened 2024-08-28
Creating document from primary side of one-one very very slow


AndrewSisley:
Creating a document from the primary side of a one-one is very very slow, even on a near empty database.

Over the http client the below takes about `8ms`:
```gql
mutation {
  create_Player(
    input: { email: "test2" }
  ) {
    _docID
  }
}
```
However, the below takes about `500ms`, and gets slower the more Players are created:
```gql
mutation {
  create_Player(
    input: { email: "test", contact_id: "bae-000163f5-a445-5909-ae67-2958d02a74dc" }
  ) {
    _docID
  }
}
```
Schema (trimmed):
```gql
type Player {
  email: String @index(unique: true)
  contact: Contact @primary @relation(name: "playerContact")
}
```
Adding an index to `contact_id` does not appear to improve things.

Flagged as a bug, as this is far slower than is reasonable, suggesting there is a bug in the code somewhere.

fredcarle:
This was initially raised by @islamaliev in June https://github.com/sourcenetwork/defradb/issues/2690

Issue #2689, opened 2024-06-06
Include metrics for 1-to-1 links validation scan


islamaliev:
upon saving a document to a collection for every relation 1-to-1 field we run a full scan to find out if there is any other document that refers to the same object. This is done to make sure 1-to-1 invariant is kept valid.
But it uses a full scan (which probably needs another ticket to optimize) and it is not being tracked by our explain.

Check `validateOneToOneLinkDoesntAlreadyExist` in `internal/db/collection.go`

PR #4218, opened 2025-11-28
feat: Create unique index for 1-to-1 relations

### LINKED ISSUE #2690: Optimize 1-to-1 links validation upon saving a document
upon saving a document to a collection for every relation 1-to-1 field we run a full scan to find out if there is any other document that refers to the same object. This is done to make sure 1-to-1 invariant is kept valid.

It uses the full scan. I didn't run any benchmarks yet, but this is potentially a huge performance hit and one of the reasons creation of multiple documents with related objects is so slow.

islamaliev:
## Relevant issue(s)

Resolves #2690 #2968 https://github.com/sourcenetwork/defradb/issues/2689

## Description

Remove full scan validation of 1-to-1 relation and replace it with auto-created unique secondary indexes.

This has now impact on concurrent transactions. If one transaction queries a secondary doc that doesn't have related primary the sec. index will try to fetch it anyway, and then if the second transaction that adds this related primary document commits, then the first one will be rejected. A test for it is `TestTransactionalCreationAndLinkingOfRelationalDocumentsForward` in `tests/integration/mutation/create/field_kinds/one_to_one_to_one/with_txn_test.go`


internal/request/graphql/schema/collection.go
```diff
@@ -891,9 +891,67 @@ func finalizeRelations(
 				}
 			}
 
+			// For one-to-one relations, ensure a unique index exists on the primary side's _id field.
+			// This is determined by: (1) it's a 1-to-1 (other field exists and is not an array), and
+			// (2) this field is marked as primary (either via @primary directive or implicit).
+			isOneToOne := hasOtherColFieldDescription && !otherColFieldDescription.Kind.IsArray()
```
AndrewSisley: todo: I think this is the wrong place to do this, or at least, you will need to duplicate this to other places, which I also think I will not like.

This location will only ever affect collections created using GQL schemas, it will not affect collections/relations created via patches, or if we expose the `db.createCollection` function.

I'm guessing based on your discord question this morning you are changing this - please request changes once those changes are pushed :)

AndrewSisley (changes requested):

internal/request/graphql/schema/collection.go
islamaliev: yes, I though so too when I realized the the test for adding relation via patch is missing. I added test, but first we need to fix this https://github.com/sourcenetwork/defradb/issues/4219

AndrewSisley: Just in case this was thought - IMO being unable to add a patch test should not block the production code in this PR being moved, tested, and merged into develop before adding tests (and possible code) for patch.

tests/integration/mutation/create/field_kinds/one_to_one_to_one/with_txn_test.go
```diff
@@ -130,15 +130,20 @@ func TestTransactionalCreationAndLinkingOfRelationalDocumentsForward(t *testing.
...
 				},
 			},
-			// Commit the transactions before querying the end result
 			testUtils.TransactionCommit{
 				TransactionID: 0,
 			},
+			// The second commit fails with a transaction conflict due to SSI semantics:
+			// - Txn0 writes index key for Website publisher, reads index key for Online publisher (via query)
+			// - Txn1 writes index key for Online publisher, reads index key for Website publisher (via query)
+			// - This creates an anti-dependency cycle that SSI detects as a conflict
```
shahzadlone, line 139: praise: Thanks for these comments

shahzadlone (approved):
LGTM I did have a similar thought/concern as Andy about having to duplicate that logic in other place, but it's a non-blocker for me. Thanks for the test changes and comments, they make it nice to understand the changes :)

docs/data_format_changes/i4218-unique-index-for-1to1.md
```diff
@@ -0,0 +1,3 @@
+# Enforce unique index for 1:1 relationships
+
+Now all 1:1 relationships will have a unique index created on the foreign key.
```
AndrewSisley: todo: Adding indexes for new relations shouldn't be a breaking change - if the change detector failed, it was probably for a different reason (potentially you modified an existing test, changing the definition between the branches under test).

Can you please find out why this broke, and then adjust this file to reflect that?  If it was just test changes or the change detector false-failing, please make sure that you note that this is not a breaking change.

internal/db/collection_define.go
```diff
@@ -327,6 +336,25 @@ existingVersionLoop:
 			return err
 		}
 
+		if ok {
```
AndrewSisley: question: Why are you only adding indexes if the collection already existed?  This code will break as soon as we unlock the ability to create collections via patch (it might already be unlocked, I lose track).

internal/db/collection_define.go
```diff
@@ -875,6 +898,155 @@ func finalizeRelations(
...
+					newCollections[i].CreateIndexes,
+					nil,
+					newCol.Definition.Name,
+					field.Name,
+				)
+				if err != nil {
+					return err
+				}
+				if newIndex != nil {
+					newCollections[i].CreateIndexes = append(newCollections[i].CreateIndexes, *newIndex)
```
AndrewSisley, line 909: question: Why are indexes added in two different ways, one via patches and one via AddSchema?

internal/db/collection_define.go
```diff
@@ -875,6 +898,155 @@ func finalizeRelations(
...
+	// No user-defined index exists, create one automatically
+	return &client.IndexCreateRequest{
+		Fields: []client.IndexedFieldDescription{{Name: idFieldName}},
+		Unique: true,
+	}, nil
+}
+
+// ensureOneToOneIndexesForPatch ensures unique indexes exist for one-to-one relations
+// added via collection patch. This is needed because patches don't go through the
+// standard schema creation flow that calls finalizeRelations.
```
AndrewSisley, line 988: todo: This feels like tech debt added on top of tech debt - instead I really think we should fix the underlying problem.  `AddSchema` and `PatchCollection` work (at the db level, after parsing), operate on exactly the same data model, and both allow the adding of all stuff (ignoring temporary restrictions to patch for testing reasons).

Please change this so that they both work the same way.

AndrewSisley (changes requested):
Its looking good, but I think we are adding a notable amount of future work for ourselves by coding addSchema and patchCollection in two different ways.

docs/data_format_changes/i4218-unique-index-for-1to1.md
islamaliev: It's not breaking change. I thought for breaking change one needs to put "BREAKING CHANGE" or something like this to make it explicit.
I'm not exactly sure why it breaks. The test that complained was changed. I though for changes tests it would ignore checks. Here is the example https://github.com/sourcenetwork/defradb/actions/runs/20098802697/job/57663933882 . The test is `TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelation_ShouldFilter`

AndrewSisley: > I thought for breaking change one needs to put "BREAKING CHANGE" or something like this to make it explicit.

Ah no, everything in the `data_format_changes` directory is by default a breaking change - a handful of the files explicitly state that they are test artefacts and not a breaking change.

I had a look at `TestQueryWithIndexOnOneToOnePrimaryRelation_IfFilterOnIndexedFieldOfRelation_ShouldFilter` - you have changed the test definition in this PR, that is why it failed - the setup stage is done by `develop` branch, without an index, then queried and asserted expecting an index - it is a test artefact.  IMO there is no need for details in the file, just note that it is not a breaking change.

But please do verify that all the other change detector tests are passing, as this file will suppress all test failures, potentially hiding an actual breaking change.

internal/db/collection_define.go
AndrewSisley: Discussed over discord - I was tripped up by the `addNewIndex` call later in `PatchCollection` - Islam and I are both unsure as to why that is there (my review has been slow, so it was added a while ago) and he is going to look into it.

The tech-debt issue I was worried about is not an issue, but the new `addNewIndex` call might be.

internal/db/collection_define.go
```diff
@@ -327,7 +336,22 @@ existingVersionLoop:
...
 
-		if ok {
+		if col.IsActive {
+			if indexReqs, hasReqs := oneToOneIndexRequests[col.Name]; hasReqs {
+				colObj, err := db.newCollection(col)
+				if err != nil {
+					return err
+				}
+				for _, indexReq := range indexReqs {
+					if _, err := colObj.createIndex(ctx, indexReq); err != nil {
```
AndrewSisley, line 346: praise: This all makes much more sense to me - thanks for this change Islam!

AndrewSisley (approved):
LGTM, thanks Islam!

Issue #4219, opened 2025-11-28
Creating secondary relation fields via patch doesn't work


islamaliev:
This test should pass, however it errors with `relation missing field. Object: Author, RelationName: author_book`
```go
func TestSchemaUpdatesAddFieldKindForeignObject_RelationBetweenTwoExistingCollections(t *testing.T) {
    test := testUtils.TestCase{
        Actions: []any{
            &action.AddSchema{
                Schema: `
                    type Author {
                        name: String
                    }
                    type Book {
                        title: String
                    }
                `,
            },
            testUtils.PatchCollection{
                Patch: `
                    [
                        { "op": "add", "path": "/Author/Fields/-", "value": {
                            "Name": "published", "Kind": "Book", "RelationName": "author_book", "IsPrimary": true
                        }},
                        { "op": "add", "path": "/Book/Fields/-", "value": {
                            "Name": "author", "Kind": "Author", "RelationName": "author_book"
                        }}
                    ]
                `,
            },
        },
    }
    testUtils.ExecuteTestCase(t, test)
}
```
Splitting the patch operations into two calls results in the same error (on the second patch action), so there is probably no workaround if you want to add a secondary relation field:
```
func TestSchemaUpdatesAddFieldKindForeignObject_RelationBetweenTwoExistingCollections(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
                    type Author {
                        name: String
                    }
                    type Book {
                        title: String
                    }
                `,
			},
			testUtils.PatchCollection{
				Patch: `
                    [
                        { "op": "add", "path": "/Author/Fields/-", "value": {
                            "Name": "published", "Kind": "Book", "RelationName": "author_book", "IsPrimary": true
                        }}
                    ]
                `,
			},
			testUtils.PatchCollection{
				Patch: `
                    [
                        { "op": "add", "path": "/Book/Fields/-", "value": {
                            "Name": "author", "Kind": "Author", "RelationName": "author_book"
						}}
                    ]
                `,
			},
		},
	}
	testUtils.ExecuteTestCase(t, test)
}
```

This is probably the same bug as https://github.com/sourcenetwork/defradb/issues/4118 - please test both when picking this up.

PR #4221, opened 2025-11-28
fix: Sec. side relation via patch and AddSchema

### LINKED ISSUE #4118: One sided relations can only be defined in same AddSchema call as the object they reference
E.g. Existing test `TestQueryOneToMany_OneSided` works, however, the following errors with `relation missing field. Object: Author, RelationName: author_book`:
```
func TestQueryOneToMany_OneSided2(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type Author {
						name: String
					}
				`,
			},
			&action.AddSchema{
				Schema: `
					type Book {
						name: String
						author: Author
					}
				`,
			},
		},
	}

}
```

islamaliev:
## Relevant issue(s)

Resolves #4118 #4219

## Description

Fix creating secondary side relation between collection via patch and AddSchema.


internal/db/definition_validation.go
```diff
@@ -30,8 +30,9 @@ import (
 type definitionState struct {
 	collections []client.CollectionVersion
 
-	collectionsByID         map[string]client.CollectionVersion
-	activeCollectionsByName map[string]client.CollectionVersion
+	collectionsByID          map[string]client.CollectionVersion
+	activeCollectionsByName  map[string]client.CollectionVersion
+	activeCollectionsByColID map[string]client.CollectionVersion
```
fredcarle, line 35: praise: That was a much simpler solution than I was expecting. Good job!

tests/integration/collection_version/updates/add/field/kind/foreign_object_test.go
fredcarle, line 1: todo: If this PR resolves #4118, please add a test that adds two types in different `AddSchema` calls with the second one creating a relationship.

```go
func TestSchemaUpdatesAddFieldKindForeignObject...(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type Author {
						name: String
					}
				`,
			},
			&action.AddSchema{
				Schema: `
					type Book {
						title: String
						author: Author
					}
				`,
			},
			...
```

fredcarle (approved):
Thanks for fixing this. Just one todo before merge.

ChrisBQu:
thought: The test you wrote shows that issue 4219 is resolved. This is good. However, I think we should also have a test like the one Andy wrote inside issue 4118, to show that that issue is also resolved by this.

Edit: I had the tab open, and it in the time it took me to write this, Fred said the same thing.

islamaliev:
@ChrisBQu @fredcarle turns out I didn't cover entirely https://github.com/sourcenetwork/defradb/issues/4118. I did now though and covered with tests. You can look again.

tests/integration/collection_version/updates/add/field/kind/foreign_object_test.go
```diff
@@ -178,3 +178,575 @@ func TestSchemaUpdatesAddFieldKindForeignObject_Succeeds(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestSchemaUpdatesAddFieldKindForeignObject_WithAddSchemaCreatingOneToManyRelationToExistingCollection_ShouldSucceed(t *testing.T) {
```
AndrewSisley: todo: This test, and a couple more below, do not belong here (no updates to collection versions), please move it somewhere else.

internal/db/collection_id.go
```diff
@@ -47,12 +48,21 @@ func setCollectionIDs(
 	collectionSets := getCollectionSets(newCollectionPtrs)
 	collectionSets = sortCollectionSets(collectionSets)
 
+	// Build a list of all collection sets including existing collections for relationship resolution.
+	// This is needed to convert NamedKind to CollectionKind for relations pointing to existing collections.
+	existingCollectionPtrs := make([]*client.CollectionVersion, len(existingCollections))
```
AndrewSisley: todo: This looks like it should not be an array of pointers, but just a regular array - having it be an array of pointers in this context suggests that the items within it will be mutated, which they really should not be here (they are existing collections, not new ones).

internal/db/collection_define.go
```diff
@@ -795,3 +797,75 @@ func deleteCollectionBlocks(
...
+				continue
+			}
+
+			if field.IsPrimary {
+				// Already marked as primary, skip
+				continue
+			}
+
+			if _, inBatch := newByName[namedKind.Name]; inBatch {
+				// The relation is within the batch, finalizeRelations in the parser already handled this
```
AndrewSisley: todo: I do not think any code within `db` should reference the parser, code-comment or otherwise - the parser (or any other pathway into it) should be providing stuff in the format expected by the `db` methods, and the `db` code should not be guessing where it came from.

internal/db/collection_define.go
```diff
@@ -795,3 +797,75 @@ func deleteCollectionBlocks(
...
+			)
+
+			if !hasCorrespondingField {
+				newCollections[i].Definition.Fields[fieldIndex].IsPrimary = true
+
+				// Also set IsPrimary=true on the corresponding _id field.
+				// The parser creates _id fields for all non-array relations, but doesn't set
+				// IsPrimary=true until finalizeRelations runs. Since finalizeRelations only
+				// sees collections in the current batch, we need to set IsPrimary on the _id
+				// field here for cross-batch relations.
```
AndrewSisley: todo: It is fine for this logic to exist in `db` if we want it to, but it should only live here because we *chose* for it to live here, because we want this package to do this.  It should not live here because of anything to do with a package that sits on top of this layer.

So similar to my other comment, please remove or reword this comment - the `db` package should not reference any of the packages layered on top of it.

internal/db/collection_define.go
```diff
@@ -795,3 +797,75 @@ func deleteCollectionBlocks(
 
 	return nil
 }
+
+// finalizeRelationsWithExistingCollections handles relation fields that point to existing collections.
+// When a new collection is added with a relation field pointing to an existing collection,
+// this function sets IsPrimary=true on the relation field and on the corresponding _id field.
+// This enables one-sided relations to be defined in separate AddSchema calls.
+func finalizeRelationsWithExistingCollections(
```
AndrewSisley: todo: I'm pretty sure there was a discussion around this behaviour a few years ago - it is automatically declaring which side of the relation is primary, without being visible to the user.

Which side is primary is pretty important, and is something that I think should be an explicit user choice - it should not be something we set automatically.

This is also only done for `AddSchema`, this is inconsistent with `PatchCollection`.

I also think there is an argument to be had that this should really be in the GQL layer, as it feels like it is a language layer thing (a default application of the `isPrimary` directive), and not a core-model default in the same way that field IDs are.

I suggest creating a thread on discord about this if you wish to keep the automatic selection of `IsPrimary`, as that is an important user-visible feature.

AndrewSisley (changes requested):
There is an important user change here that needs to be discussed before I can approve.

internal/db/collection_define.go
```diff
@@ -795,3 +797,75 @@ func deleteCollectionBlocks(
...
+
+		for fieldIndex, field := range newCol.Definition.Fields {
+			namedKind, ok := field.Kind.(*client.NamedKind)
+			if !ok || namedKind.IsArray() {
+				// Only process non-array object relations
+				continue
+			}
+
+			if field.IsPrimary {
+				// Already marked as primary, skip
```
AndrewSisley: todo: This comment adds nothing but eye space to the reader, so does the one above (`// Only process non-array object relations`).  The code already says this in very clear terms, if the comments wants to justify their existence in these places they need to detail something that the code does not already do - they should describe *why* we want to `continue` here.

internal/db/collection_define.go
fredcarle: If adding a new collection from `AddSchema` with a relationship to a collection that already exists, then it has no other choice but to be the primary. The will also be the behavior when we move to the new syntax for defining relationship (only the primary side will ever be defined).

AndrewSisley: Ah true!  Thanks Fred.

suggestion: Please consider noting this in the function-level documentation :)

internal/db/collection_define.go
islamaliev: removed

internal/db/collection_define.go
islamaliev: removed reference

internal/db/collection_define.go
islamaliev: hey guys, after looking into these methods (`finalizeRelations` from `internal/request/graphql/schema/collection.go`) more carefully I decided to merge them because they do more or less the thing - finalize or resolve relations. It's better to have everything in one place and the logic doesn't really belong to parser side.

internal/db/collection_define.go
islamaliev: updated

internal/db/collection_id.go
islamaliev: you are right. Adjusted

tests/integration/collection_version/updates/add/field/kind/foreign_object_test.go
islamaliev: done

tests/integration/collection_version/updates/add/field/kind/foreign_object_test.go
islamaliev: done

internal/request/graphql/schema/collection.go
```diff
@@ -815,85 +807,3 @@ func genRelationName(t1, t2 string) (string, error) {
...
-						len(result.Definition.Fields)+1,
-					)
-					copy(result.Definition.Fields, existingFields[:fieldIndex+1])
-					copy(result.Definition.Fields[fieldIndex+2:], existingFields[fieldIndex+1:])
-
-					// An _id field is added for every 1-1 or 1-N relationship from this object if the relation
-					// does not point to an embedded object.
-					//
-					// It is inserted immediately after the object field to make things nicer for the user.
-					result.Definition.Fields[fieldIndex+1] = client.CollectionFieldDescription{
```
AndrewSisley, line 886: question: This logic appears to have disappeared, and I cannot find it's duplication anywhere where on collection create.  Where is this now happening for `AddSchema`?

islamaliev: I looked up the call stack and realized that we always add _id field for `client.NamedKind` fields in `fieldsFromAST` in `internal/request/graphql/schema/collection.go` so this code was dead code. And I also checked in codecov https://app.codecov.io/gh/sourcenetwork/defradb/blob/develop/internal%2Frequest%2Fgraphql%2Fschema%2Fcollection.go#L874
So I just removed it.

AndrewSisley: Ah sweet! I wonder why my Ctrl+f didn't find that.  Thanks for removing this then :)

internal/request/graphql/schema/collection.go
```diff
@@ -815,85 +807,3 @@ func genRelationName(t1, t2 string) (string, error) {
 	}
 	return fmt.Sprintf("%s_%s", t2, t1), nil
 }
-
-func finalizeRelations(
```
AndrewSisley, line 828: praise: The removal of this is a great change, thanks Islam :)

AndrewSisley (approved):
LGTM, thanks for fixing these Islam :)

suggestion: I suggest rewording the title a little bit before merge - adding a relation via patch has always worked, this fixes adding the secondary side.  And more importantly, it fixes #4118 (unrelated to patch) which is something people are more likely to go looking for in the change log IMO.

PR #4225, opened 2025-12-01
fix: Remove enabling NAC option from config

### LINKED ISSUE #4127: Remove the node acp enable from config file options at start
Remove the node acp enable from config file options at start (improve dev-ex)

The `acp.node.enable` config file option should likely be removed from the configuration file because it can confuse people into thinking it's value is being used even after the first start of defradb node that enables node acp.

Note: Once the node acp is enabled on the defradb node start, all other restarts (closing of a node and restarting) will not really care about the value of that config parameter (it's only relevant for the first start). So, we should have it only as a cli flag, to avoid confusing logs and also to avoid confusing users because they might assume by setting it to false will disable NAC (which it would not).

shahzadlone:
## Relevant issue(s)
Resolves #4127 

## Description
-  Remove enabling NAC option from config

## How has this been tested?
- Tests should still pass

- WSL2 (Manjaro)

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -69,7 +69,6 @@ var configFlags = map[string]string{
 	"keyring-backend":            "keyring.backend",
 	"keyring-path":               "keyring.path",
 	"no-keyring":                 "keyring.disabled",
-	"node-acp-enable":            "acp.node.enable",
 	"document-acp-type":          "acp.document.type",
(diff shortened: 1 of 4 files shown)


ChrisBQu (approved):
LGTM. Seems simple and clean.

PR #4226, opened 2025-12-01
refactor: Replace SE key collection root cid with short id

### LINKED ISSUE #4223: Use col short id for DatastoreSE instead of cid
`DatastoreSE` uses the full collection root cid instead of the short id, this takes up more space than it needs to, and differs from the other datastore keys complicating related code.  Make it use the short id instead.

AndrewSisley:
## Relevant issue(s)

Resolves #4223

## Description

Replaces the collection root cid in the SE artifact key with the collection short id.

This saves storage space and makes the key consistent with the rest of the datastore keys.


shahzadlone (approved):
LGTM

Issue #4207, opened 2025-11-24
Allow truncation of collection


AndrewSisley:
Hard deleting all local document stuff (blocks, datastore, indexes)

AndrewSisley:
WIP branch `4207-truncate`

PR #4242, opened 2025-12-04
refactor: Wrap Datastore in key-typed wrapper

### LINKED ISSUE #4233: Wrap datastore in key-typed Defra owned implementation
Defra needs to manage locks for various aspects of the store(s).  The immediate need is a collection-wide lock for the datastore as part of #4207.  Wrapping the corekv store is the safest way to enforce the locks.

This could probably be done in corekv, but that would complicate the lock management, and this is perhaps something that we do not want shared across projects., so I'll be implementing it in Defra for now (easy enough to move later should we wish).

This issue will replicate the corekv interface but swap out the `key []byte` params for the `internal/keys.Key` interface (or probably just part of it), as the lock logic that will later be introduced needs to access parts of the strongly typed keys (probably via an interface).

AndrewSisley:
## Relevant issue(s)

Resolves #4233

## Description

Wraps the Datastore in a new `Keyedstore` wrapper around `corekv.ReaderWriter`.

This gives us a single location from which we can manage Defra-specific access to the datastore using they typed keys, which for #4207, will mean enforcing read-locks for collection-specific keys whilst a truncate is being performed.

The new `Keyedstore` interface is store agnostic and can be used for our other stores, and may need to be at some point.  I have only used it for the Datastore in this PR, as I think that is the only store that I will need/want to lock in such a way for `truncate`.  

The only implementation of `Keyedstore` would work for for all stores at the moment, however that will change soon, within the `truncate` PR, as the locking will be very specific to the datastore.

This PR is a pure refactor, no user-behaviour has changed.  The first two commits are semi-unrelated dead-code cleans that I spotted as they got in the way.  Similarly, a couple of unit tests were removed.

### DIFF
--- a/internal/datastore/datastore.go
+++ b/internal/datastore/datastore.go
@@ -1,0 +1,75 @@
+// Copyright 2025 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package datastore
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corekv/namespace"
+)
+
+type datastore struct {
+	underlying corekv.ReaderWriter
+}
+
+var _ Keyedstore = (*datastore)(nil)
+
+func newDatastore(rootstore corekv.ReaderWriter) *datastore {
+	return &datastore{
+		underlying: namespace.Wrap(rootstore, []byte{dataStoreKey}),
+	}
+}
+
+func (s *datastore) Get(ctx context.Context, key Key) ([]byte, error) {
+	keyBytes := key.Bytes()
+	return s.underlying.Get(ctx, keyBytes)
+}
+
+func (s *datastore) Has(ctx context.Context, key Key) (bool, error) {
+	keyBytes := key.Bytes()
+	return s.underlying.Has(ctx, keyBytes)
+}
+
+func (s *datastore) Iterator(ctx context.Context, opts IterOptions) (corekv.Iterator, error) {
+	var prefix []byte
+	var start []byte
+	var end []byte
+
+	if opts.Prefix != nil {
+		prefix = opts.Prefix.Bytes()
+	}
+	if opts.Start != nil {
+		start = opts.Start.Bytes()
+	}
+	if opts.End != nil {
+		end = opts.End.Bytes()
+	}
+
+	ckvOpts := corekv.IterOptions{
+		Prefix:   prefix,
+		Start:    start,
+		End:      end,
+		KeysOnly: opts.KeysOnly,
+		Reverse:  opts.Reverse,
+	}
+	return s.underlying.Iterator(ctx, ckvOpts)
+}
+
+func (s *datastore) Set(ctx context.Context, key Key, value []byte) error {
+	keyBytes := key.Bytes()
+	return s.underlying.Set(ctx, keyBytes, value)
+}
+
+func (s *datastore) Delete(ctx context.Context, key Key) error {
+	keyBytes := key.Bytes()
+	return s.underlying.Delete(ctx, keyBytes)
+}
--- a/internal/datastore/keyedstore.go
+++ b/internal/datastore/keyedstore.go
@@ -1,0 +1,98 @@
+// Copyright 2025 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package datastore
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/corekv"
+)
+
+// Key represents a typed key for a key-value pair or a prefix within a `Keyedstore`.
+type Key interface {
+	// Bytes returns the serialized for of the key, as it is stored within a `Keyedstore`.
+	Bytes() []byte
+}
+
+// Keyedstore is a corekv.ReaderWriter that takes typed `Key`s instead of `[]byte`
+// for it's key function parameters.
+type Keyedstore interface {
+	// Get returns the value at the given key.
+	//
+	// If no item with the given key is found, nil, and an [ErrNotFound]
+	// error will be returned.
+	Get(ctx context.Context, key Key) ([]byte, error)
+
+	// Has returns true if an item at the given key is found, otherwise
+	// will return false.
+	Has(ctx context.Context, key Key) (bool, error)
+
+	// Iterator returns a read-only iterator using the given options.
+	Iterator(ctx context.Context, opts IterOptions) (corekv.Iterator, error)
+
+	// Set sets the value stored against the given key.
+	//
+	// If an item already exists at the given key it will be overwritten.
+	Set(ctx context.Context, key Key, value []byte) error
+
+	// Delete removes the value at the given key.
+	//
+	// If no matching key is found the behaviour is undefined:
+	// https://github.com/sourcenetwork/corekv/issues/36
+	Delete(ctx context.Context, key Key) error
+}
+
+// IterOptions contains the full set of available iterator options,
+// it can be provided when creating an [Iterator] from a [Store].
+type IterOptions struct {
+	// Prefix iteration, only keys beginning with the designated prefix
+	// with the given prefix will be yielded.
+	//
+	// Keys exactly matching the provided `Prefix` value will not be
+	// yielded.
+	//
+	// Providing a Prefix value should cause the Start and End options
+	// to be ignored, although this is currently untested:
+	// https://github.com/sourcenetwork/corekv/issues/35
+	Prefix Key
+
+	// If Prefix is nil, and Start is provided, the iterator will
+	// only yield items with a key lexographically greater than or
+	// equal to this value.
+	//
+	// Providing an `End` value equal to or smaller than this value
+	// will result in undefined behaviour:
+	// https://github.com/sourcenetwork/corekv/issues/32
+	Start Key
+
+	// If Prefix is nil, and End is provided, the iterator will
+	// only yield items with a key lexographically smaller than this
+	// value.
+	//
+	// Providing an End value equal to or smaller than Start
+	// will result in undefined behaviour:
+	// https://github.com/sourcenetwork/corekv/issues/32
+	End Key
+
+	// Reverse the direction of the iteration, returning items in
+	// lexographically descending order of their keys.
+	Reverse bool
+
+	// Only iterate through keys. Calling Value on the
+	// iterator will return nil and no error.
+	//
+	// This option is currently untested:
+	// https://github.com/sourcenetwork/corekv/issues/34
+	//
+	// It is very likely ignored for the memory store iteration:
+	// https://github.com/sourcenetwork/corekv/issues/33
+	KeysOnly bool
+}
--- a/internal/db/index.go
+++ b/internal/db/index.go
@@ -221,14 +221,14 @@ func (index *collectionBaseIndex) deleteIndexKey(
 ) error {
 	txn := datastore.CtxMustGetTxn(ctx)
 	ds := txn.Datastore()
-	exists, err := ds.Has(ctx, key.Bytes())
+	exists, err := ds.Has(ctx, &key)
 	if err != nil {
 		return err
 	}
 	if !exists {
 		return NewErrCorruptedIndex(index.desc.Name)
 	}
-	return ds.Delete(ctx, key.Bytes())
+	return ds.Delete(ctx, &key)
 }
 
 // RemoveAll remove all artifacts of the index from the storage, i.e. all index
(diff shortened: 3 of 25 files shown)


internal/datastore/keyedstore.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Key represents a typed key for a key-value pair or a prefix within a `Keyedstore`.
+type Key interface {
+	// Bytes returns the serialized for of the key, as it is stored within a `Keyedstore`.
+	Bytes() []byte
+}
+
+// Keyedstore is a corekv.ReaderWriter that takes typed `Key`s instead of `[]byte`
+// for it's key function parameters.
+type Keyedstore interface {
```
islamaliev, line 27: suggestion: the name consist of two words. Why not to make it `KeyedStore`. We won't need to teach our IDE a new work

internal/datastore/datastore.go
```diff
@@ -0,0 +1,75 @@
...
+package datastore
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corekv/namespace"
+)
+
+type datastore struct {
```
islamaliev, line 20: note: it's a bit confusing to have a datastore as one of the namespaces inside rootstore and another datastore as overarching storage wrapper.

I think we should discuss how we want to go about it long term.

internal/datastore/datastore.go
```diff
@@ -0,0 +1,75 @@
...
+	underlying corekv.ReaderWriter
+}
+
+var _ Keyedstore = (*datastore)(nil)
+
+func newDatastore(rootstore corekv.ReaderWriter) *datastore {
```
islamaliev, line 26: suggestion: there is nothing in this `datastore` that makes it specific for datastore except for `dataStoreKey` that can be easily passed as an argument. Then this `datastore` (that can be renamed to something like `defaultKeyedstore`) can be reused by other stores.

internal/datastore/datastore.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+
+	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corekv/namespace"
+)
+
+type datastore struct {
+	underlying corekv.ReaderWriter
+}
```
islamaliev, line 22: question: I presume in follow up PRs you plan to add here some locking mechanism. Am I right? If not, then I misunderstood your plan and the purpose of this struct.

internal/db/index.go
```diff
@@ -244,14 +244,38 @@ func (index *collectionBaseIndex) RemoveAll(ctx context.Context) error {
 	prefixKey.IndexID = index.desc.ID
 
 	txn := datastore.CtxMustGetTxn(ctx)
-	ds := txn.Datastore()
-	keys, err := datastore.FetchKeysForPrefix(ctx, prefixKey.Bytes(), ds)
```
islamaliev, line 248: suggestion: the only other place where `datastore.FetchKeysForPrefix` is used is `deleteReplicatorRetryIfNoMoreDocs` in `internal/db/p2p/replicator.go`. There it's just used to check if there are any keys so it's lighter usage.

I would say you change `datastore.FetchKeysForPrefix` to take `Keyedstore` and adjust the usage in `internal/db/p2p/replicator.go`

internal/keys/datastore_index.go
```diff
@@ -34,9 +34,12 @@ type IndexDataStoreKey struct {
 	IndexID uint32
 	// Fields is the values of the fields in the index
 	Fields []IndexedField
+	// Offset can be set in order to add the given value to the end of the resultant bytes,
+	// allowing fine-grained control over key selection.
+	Offset uint64
```
islamaliev: todo: I don't understand how this is related to the PR.

todo: the documentation is not quite clear. Which "given value" to which "end"? It doesn't explain why it's needed.

internal/keys/datastore_index_test.go
```diff
@@ -140,33 +140,12 @@ func TestIndexDataStoreKey_PrefixEnd_Ordering(t *testing.T) {
...
-					CollectionShortID: 10,
-					IndexID:           2,
-				},
-				{
-					CollectionShortID: 11,
-					IndexID:           1,
-				},
-			},
-			shouldMatch: []bool{true, true, false},
-		},
```
islamaliev, line 163: question: why did you delete this test? It's not possible anymore to iterate over collection indexes?

internal/keys/datastore_index.go
```diff
@@ -180,23 +183,36 @@ func EncodeIndexDataStoreKey(key *IndexDataStoreKey) []byte {
 
 	b := encoding.EncodeUvarintAscending([]byte{'/'}, uint64(key.CollectionShortID))
 
-	if key.IndexID == 0 {
-		return b
+	if key.IndexID != 0 {
```
islamaliev, line 189: todo: why would you allow index key without index id?

This should never happen. Even if it was possible how would you decode the same values for 2 different indexed fields?
```json
{
  "num1": 1, // indexed
  "num2": 1 // indexed
}
```

islamaliev (changes requested):
I don't quite understand the necessity of changing index keys and the way it was done. We need to clarify it first.

internal/datastore/keyedstore.go
AndrewSisley: All our stores following this casing convention.  For what it's worth, I'm pretty sure I was in favour of capitalization too, BlockStore, DataStore etc.  I don't want to revisit that conversation here though, it is low value and out of scope.

internal/datastore/datastore.go
AndrewSisley: I agree, but I'd rather not rename the package right now.  The name has always been a bit odd, this is not a new problem introduced by this PR.

internal/datastore/datastore.go
AndrewSisley: This is noted in the PR description.  Leaving as is.

internal/datastore/datastore.go
AndrewSisley: Yes

internal/db/index.go
AndrewSisley: `deleteReplicatorRetryIfNoMoreDocs` operates on the `Peerstore`, and wrapping and converting the `Peerstore` to  a `Keyedstore` is unnecessary and well out of scope from this PR.

internal/keys/datastore_index.go
AndrewSisley: > todo: I don't understand how this is related to the PR.

Sorry, I had a commit specifically for this change, accidentally squashed part of it, and decided against bothering to unsquash it.

Many of our keys implement the `Walkable` interface, it essentially `bytesPrefixEnd` but before serialization.  This allows the store to be walked whilst preserving the typed key instead of operating on a `[]byte`.

Operating on the typed key, as noted in the description, is required in order to reliably block access to the datastore based on key components (e.g. the collection short id).

`Offset` was required by this key type (unlike the others implementing `Walkable`), as the last serialized value is a `bool` and so the `Walkable` cannot be implemented by incrementing it in the same way that an `int` can be.

> Which "given value"

`given value` refers to the value given to the property, this is a fairly common way of wording documentation IMO. 

> to which "end"

When refering to a slice or an array, `end` in programming refers to the last element(s) - I'm not sure people are going to be confused here and think that it refers to the first element(s).

It's also the naming convention of the function that this property controls - `bytesPrefixEnd`.

> It doesn't explain why it's needed.

`allowing fine-grained control over key selection` attempts to cover this without being overly specific - there are a few references to this (via `PrefixEnd()`, as well as a few other potential use-cases that the other `Walkable` keys have.  I'd really prefer to avoid being too concrete here example wise, as this is a fairly generalised mechanic shared with a decent number of other keys and is used for various different reasons.

internal/keys/datastore_index_test.go
AndrewSisley: I think this one got deleted because it was failing whilst the integration tests were passing, and it was taking more effort to tell what it was testing and why than I think is justified for a test of internal code behaviour.  I don't know if the failure was caused by the test being overly specific, a test bug, or whatever, and it seemed very unlikely to be testing any production-useful behaviour that wouldn't have been caught by our integration tests.

internal/keys/datastore_index.go
AndrewSisley: This is how the original code worked.  You can see this on the removed line 184 where it returns early.  The change on this line is to allow the `Offset` to be applied even in this case.

I'm pretty sure there were unit tests that complained before the prod code was changed, and it is very possible I made this change under the assumption that I would receive less pushback accommodating the unit tests, than I would get if I deleted them.

internal/keys/datastore_index_test.go
islamaliev: just because we have integration tests, doesn't mean we should delete unit tests. The test is quite simple and doesn't test something extraordinary: simple bytes range check.

todo: please figure out why it fails and fix it

internal/keys/datastore_index.go
islamaliev: I think I mostly concerned here with fields being added to the the bytes if no `IndexID` is provided. This should not happen. If there is not `IndexID` fields also should be skipped.

internal/keys/datastore_index.go
islamaliev: > Offset was required by this key type (unlike the others implementing Walkable), as the last serialized value is a bool and so the Walkable cannot be implemented by incrementing it in the same way that an int can be.

By "value" do you mean the value of a bool field on a document that is indexed? Buy why does it matter if it's bool or not? You operate with bytes. Can you give an example where the problem shows?

> I'm not sure people are going to be confused here and think that it refers to the first element(s).

Being sarcastic is not helpful here and counterproductive. I wouldn't recommend to go this route.

Please do not assume that if something is obvious to you, it's obvious to everybody else. Even if one person has a question (let alone the original implementor of this code), it should be a good enough sign that there is not enough clarity.

> `allowing fine-grained control over key selection` attempts to cover this without being overly specific

But it doesn't say why you need this control. It worked fine before. If it has something to do with bool fields, that's a valuable info to add.

> Offset can be set

I can't see why anyone would set it explicitly. It's a side effect of calling `PrefixEnd()` and I believe it can be event made private.

internal/keys/datastore_index.go
```diff
@@ -180,23 +183,36 @@ func EncodeIndexDataStoreKey(key *IndexDataStoreKey) []byte {
...
 	for _, field := range key.Fields {
 		b = append(b, '/')
 		b = encoding.EncodeFieldValue(b, field.Value, field.Descending)
 	}
 
+	if key.Offset != 0 {
+		for i := 0; i < int(key.Offset); i++ {
+			b = bytesPrefixEnd(b)
+		}
+	}
```
islamaliev: suggestion: 
```go
for range int(key.Offset) {
	b = bytesPrefixEnd(b)
}
```
no need for `if`

internal/keys/datastore_index.go
AndrewSisley: > You operate with bytes

This PR changes that, this key, in keeping with our other keys, and the new `Keyedstore` interface no longer prefix-ends using bytes.

> Being sarcastic is not helpful here and counterproductive.

I wasn't being sarcastic.  I don't think people will think `end` will refer to the start of a slice, which is how I read your comment.  Did you mean something else?

> But it doesn't say why you need this control. It worked fine before. If it has something to do with bool fields, that's a valuable info to add.

It worked fine before because the store interface it was being passed into took an array of bytes.  This PR changes that, the PrefixEnd state-change needs to occur before converting the typed representation into a `[]byte`.

> I can't see why anyone would set it explicitly. It's a side effect of calling PrefixEnd() and I believe it can be event made private.

It is not a side effect, it is the primary/only effect.  I don't expect people to set it explicitly either.  I think it is important that people/code can read the typed representation of the key though, which is why it, and all the other properties on all of our keys, are public.

Would you prefer that it was private?

internal/keys/datastore_index.go
AndrewSisley: Nice spot - thanks :)  I think this was left in after I moved some stuff about.

- [x] Remove pointless if

internal/keys/datastore_index.go
AndrewSisley: I see what you mean.  But perhaps that validation would be better placed in the constructor? IMO calling `Bytes()` should not be validating the key, and should serialise it as best as it can - this means including fields if they were specified without an index id.

I think we are also discussing dead code here, as this doesn't happen in production.

internal/keys/datastore_index_test.go
AndrewSisley: The unit test is asserting internal behaviour that no longer exists.  The behaviour that it asserts does not need to exist, and so neither does the unit test.

Just because a unit test currently describes dead behaviour does not mean that the behaviour or the test needs to be preserved.

internal/keys/datastore_index.go
AndrewSisley: Discussed on a call.

`bytesPrefixEnd` is a bit confusing already, and this adds a new layer of complexity by essentially delaying it's execution.  We are both a bit unsure as how to better word this - we talked about perhaps making it even less specific, but noting it's relation to `bytesPrefixEnd`, and trying to expand on *why* the end is ever prefixed.

Islam does not use `given value` for members, only function params - I'll try and avoid this wording for members in the future and will change it here.

We noted briefly about how this could be extracted from this particular key type and embedded, in this and others, however we both agreed that this was seen as an unnecessary complication right now.

- [x] Reword

internal/keys/datastore_index.go
```diff
@@ -180,23 +186,34 @@ func EncodeIndexDataStoreKey(key *IndexDataStoreKey) []byte {
...
-	if key.IndexID == 0 {
-		return b
+	if key.IndexID != 0 {
+		b = append(b, '/')
+		b = encoding.EncodeUvarintAscending(b, uint64(key.IndexID))
 	}
-	b = append(b, '/')
-	b = encoding.EncodeUvarintAscending(b, uint64(key.IndexID))
 
 	for _, field := range key.Fields {
```
islamaliev: todo: please process fields only if `key.IndexID != 0`

AndrewSisley: I do disagree (see here https://github.com/sourcenetwork/defradb/pull/4242/files#r2593681025 for why), but it is dead-code and not worth either of us spending time arguing the pros/cons on IMO :)

- [x] Only serialize fields if IndexID != 0

islamaliev (approved):
LGTM! Thanks Andy for adjustments.

Issue #4208, opened 2025-11-25
Establish peers connection acknowledgement mechanism in tests


islamaliev:
Currently when settings up a test with complex network topology like this
```
// Node 0 ──── Node 1 ──── Node 2
//    │
//    └─────── Node 3 ──── Node 4
```
with sequential `testUtils.ConnectPeers` actions they trigger connections too fast which doesn't allow libp2p to fully exchange information between the peers which lead to test flakiness when test assume the network configuration is complete.

A workaround for this at the moment is either to put a long way `testUtils.Wait` after all the connections or to have a short wait after every connection.

Would be great to have some kind of handshake between peers when connection is established.

fredcarle:
> Would be great to have some kind of handshake between peers when connection is established.

The `Connect` function blocks until connection is fully established. The issue lies with the asynchronous propagation of pubsub topic subscriptions.

PR #4247, opened 2025-12-05
feat: Push topic peer events to event bus

### LINKED ISSUE #4246: Push topic peer events to event bus
originated from https://github.com/sourcenetwork/defradb/pull/4188.

We want to listen to peer event on all topic and push them to event bus so that it can be used for fine-tuning network behaviour

islamaliev:
## Relevant issue(s)

Resolves #4246 #4208

## Description

Make all topic that the node subscribed to handle peer event and push them to event bus.
Add integration tests for TopicPeerEvent published to the event bus when peers join/leave pubsub topics.


internal/db/p2p/p2p.go
```diff
@@ -142,6 +142,22 @@ func (proc *pushLogCommProcessor) ProcessRequest(
 	return protocol.PushLogReply{}, proc.p2p.processPushlogRequest(ctx, &req, true)
 }
 
+// peerEventHandlingHost wraps a Host to add a PeerEventHandler to pubsub topics.
+// It's added so that KMS doesn't need to bother with event handling and keeps it independent
+// from e
```
fredcarle: todo: This sentence seems to have been cut short.

islamaliev: done

tests/action/wait_for_peer_events.go
```diff
@@ -0,0 +1,146 @@
...
+			if peerEvent.EventType != eventType {
+				continue
+			}
+			if topicPeers, topicExists := expectedPeers[peerEvent.Topic]; topicExists {
+				if topicPeers[peerEvent.PeerID] {
+					delete(topicPeers, peerEvent.PeerID)
+					totalExpected--
+				}
+			}
+		case <-timer.C:
```
fredcarle, line 123: question: Why did you not use 
```go
case <-time.After(timeout):
```

fredcarle: I just realized it's because you want to apply the timeout to the whole loop and not per message. Never mind the question :)

tests/action/wait_for_peer_events.go
```diff
@@ -0,0 +1,146 @@
...
+				"source node %d did not receive %s events for: %v",
+				a.NodeID, eventType, remaining)
+			return
+		}
+	}
+}
+
+// extractPeerID extracts the peer ID from a multiaddr string.
+// Example: /ip4/127.0.0.1/tcp/4001/p2p/12D3KooWExample -> 12D3KooWExample
+func extractPeerID(addr string) string {
```
fredcarle: thought: There are some utilities that help with extracting the peerID that could provide more context if the address is wrongly formatted.
```go
maddr, err := multiaddr.NewMultiaddr(addr)
if err != nil {
  return err
}
id, err := peer.IDFromP2PAddr(maddr)
if err != nil {
  return err
}
```

fredcarle (approved):
Thanks for adding this. It should give use the ability to substantially reduce test flakiness in the future.

tests/action/wait_for_peer_events.go
islamaliev: done

PR #4248, opened 2025-12-05
fix: Make resolution of UTC_NOW the same across transaction

### LINKED ISSUE #4231: Multiple instances of UTC_NOW in a transaction should resolve to the same value
Consider running the following:

```
defradb client schema add '
  type User {
    name: String 
    created: DateTime @default(dateTime: UTC_NOW) 
  }
'

defradb client query '
mutation {
    bob: create_User(input: { name: "Bob" }) {
        _docID
        name
        created
    }

    alice: create_User(input: { name: "Alice" }) {
        _docID
        name
        created
    }
}
'
```

Currently, we expect the `created` timestamp on both documents to be slightly different from one another, as instances of `UTC_NOW` are resolved separately for each mutation. This is not the desired behavior. All instances of it should resolved to exactly the same value.

ChrisBQu:
## Relevant issue(s)

Resolves #4231 

## Description

This PR makes it so that across a transaction, all instances of `UTC_NOW` will resolve to the same value. 

The core work is done by adding a new field to `datastore.BasicTxn`, called `ts` for timestamp. We also add function to the `datastore.Txn` interface, `StartTS()` which returns the timestamp. This value is then propogated down inside a context. The new `clock` package defines the context key that is used for this purpose, and contains the functions that add a timestamp to, and get a timestamp from, a given context.

You can see that many function signatures have been touched to facilitate this passing down-and-through of the context: `client.NewDocFromMap`, `client.NewDocWithID`, `doc.SetWithJSON`, etc. In most of these places, the change will be straightforward, and the reason for the change will hopefully be self-evident.

Additionally, this PR includes several integration tests which illustrate the new functionality. Further, the client wrappers have all been adjusted accordingly, such that the Txn wrappers fit the new interface.

## Tasks

## How has this been tested?

- WSL


internal/datastore/txn.go
```diff
@@ -45,6 +46,9 @@ type Txn interface {
 	// ID returns the unique immutable identifier for this transaction.
 	ID() uint64
 
+	// StartTS returns the start timestamp of the transaction
+	StartTS() time.Time
```
AndrewSisley, line 50: question: This very much looks like a different timestamp to the GQL `UTC_NOW`, and suggests that other future timestamps are possible, such as `CommitTS`.

Do you have plans for other timestamps? If not, are you sure you want this held on the transaction?

ChrisBQu: That code was actually written by John. He took a swing at the problem that was very different than my original approach. But let's discuss it in the standup.

fredcarle (approved):
LGTM

PR #4254, opened 2025-12-08
fix: Migrations with filtering

### LINKED ISSUE #4119: Setting migration between old versions doesn't work
When setting migration between previous versions it is not applied.
```go
func TestSchemaMigrationQuery_SetMigrationBetweenOldVersions_ShouldApplyMigration(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
				type Users {
					name: String
					age: Int
				}
			`,
			},
			testUtils.CreateDoc{
				DocMap: map[string]any{
					"name": "Andy",
					"age":  20,
				},
			},
			testUtils.CreateDoc{
				DocMap: map[string]any{
					"name": "John",
					"age":  30,
				},
			},
			testUtils.CreateDoc{
				DocMap: map[string]any{
					"name": "Fred",
					"age":  25,
				},
			},
			testUtils.CreateDoc{
				DocMap: map[string]any{
					"name": "Islam",
					"age":  32,
				},
			},
			testUtils.PatchCollection{
				Patch: `
				[
					{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "level", "Kind": "Int"} }
				]
			`,
			},
			testUtils.PatchCollection{
				Patch: `
				[
					{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "points", "Kind": "Int"} }
				]
			`,
			},
			testUtils.PatchCollection{
				Patch: `
				[
					{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "rank", "Kind": "Int"} }
				]
			`,
			},
			testUtils.PatchCollection{
				Patch: `
				[
					{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "score", "Kind": "Int"} }
				]
			`,
			},
			testUtils.ConfigureMigration{
				LensConfig: client.LensConfig{
					SourceSchemaVersionID:      "bafyreibz2wolrhx2vnrvyh7vg5rlehyekhnivthqom5zft5ku4mhpzi2fa", // v3
					DestinationSchemaVersionID: "bafyreigpp4pjidheelridixgeppcki44t33dugru5b4hqnicyue5jqtudq", // v4
					Lens: model.Lens{
						Lenses: []model.LensModule{
							{
								Path: lenses.IncrementModulePath,
								Arguments: map[string]any{
									"field": "age",
									"value": 5,
								},
							},
						},
					},
				},
			},
			testUtils.Request{
				Request: `query {
					Users(filter: {age: {_eq: 30}}) {
						name
					}
				}`,
				Results: map[string]any{
					"Users": []map[string]any{
						{
							"name": "Fred",
						},
					},
				},
			},
		},
	}

}
```
In this example "Fred" should be returned (it's original value 25 + 5 from migration), but currently it returns "John".

islamaliev:
## Relevant issue(s)

Resolves #4119

## Description

Originally the work was supposed to be done migration between old collection version, but during the development few other bigger problems were spotted.

First, the migrations would work incorrectly if filter is present. The problem was that filter is applied in the fetcher that is triggered before migration which is wrong and produces unexpected results. So now we check if there is a migration we skip filtering and just rely on filtering on select node which happen latter in the call chain.

Second, applying patch with inline migration and a follow up patch without migration would cause the first migration to be executed twice because the first one would be re-added to the next one for bookkeeping. I assume it didn't show before because most (or all) of the lens modules are idempotent and could be applied any number or times. With `increment` module some other issues come to surface.

Third, we tried to apply migration to every version without checking if there is a migration between versions. Added proper `Transform.HasValue()` checks before attempting to apply migrations. When no transform exists at a version step, documents pass through unchanged but the traversal continues to the next version.


tests/integration/collection_version/migrations/query/simple_test.go
```diff
@@ -1102,3 +1102,106 @@ func TestSchemaMigrationQueryMigrationCopiesExistingFieldWhenSrcAndDstFieldNotRe
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestSchemaMigrationQuery_MigrationBetweenOldVersions_ShouldApplyMigration(t *testing.T) {
```
AndrewSisley: todo: This test seems to pass in develop.  I don't think you have changed anything.

The production code change doesn't make sense to me either, can you please expand on what you think is, and was happening?

Calls to `l.store.Transform` with empty and/unknown IDs will result in the input slice being returned anyway.

AndrewSisley (changes requested):
Requesting changes, as I'm not sure you've changed anything.

tests/integration/collection_version/migrations/query/simple_test.go
AndrewSisley: Keenan encountered this issue when working on the LevelDB stuff.

We debugged it a bit together.  There is a change in the latest Lens, not the one that Defra currently consumes, that will return an error if an empty string is provided.

The only way that I could see where code in this PR fails on it's current Lens version, is if something somewhere is writing an empty string to Lens (allowing it's cached value to be picked up on this call, instead of not finding anything).  This would be a separate bug.

The change in this PR is necessary in order to update the Lens version, and should be kept.  I don't think it fixes the underlying issue, however I have no idea how it could be writing empties to Defra with the current way Defra uses Lens.

I still think this is a todo, as we should test the scenario that fails in `develop` branch, and this test doesn't seem to do that.

suggestion: I also suggest we take the time to figure out the underlying bug that caused this, as I'm not convinced that this code is it (it should only be hiding the write-bug).

nasdf (approved):
LGTM. Please merge when you get the chance. I am blocked by this.

tests/integration/collection_version/migrations/query/simple_test.go
AndrewSisley: :grin:  Islam spotted that there was an accidental index added  to the test schema - when removing the index the test fails in develop and passes here :grin:

AndrewSisley (dismissed):
LGTM, assuming you remove that drama-creating index from the test :rofl:   Thanks Islam :grin:

tests/integration/collection_version/migrations/query/simple_test.go
islamaliev: Correction: when removing index tests don't pass on the branch :/

AndrewSisley: RIP lol.

We still need the production code changes in order to use the latest (unpublished) Lens version, so maybe don't throw those out if you find a totally unrelated fix.

AndrewSisley: You could maybe even merge the prod. code without closing the ticket (change the PR/merge-commit title, and link to a new issue) - this should unblock Keenan, who is dependant on the latest Lens code (and not the bug fix, given that his tests are all passing on this branch).

EDIT: Although, it would probably make sense to incorporate this change in the issue/PR that bumps the Lens version - that feels nicer to me.

AndrewSisley (approved):
LGTM, thanks Islam!

PR #4263, opened 2025-12-09
feat: Serverless client side schema generation

### LINKED ISSUE #2982: Client side GraphQL schema generation CLI command
Add a CLI command to run the client side schema generation.

jsimnz:
## Relevant issue(s)

Resolves #2982 

## Description

Adds a new `sdl` command which contains a `generate` subcommand. This command takes as input one or more user defined schema types and outputs the fully generated defradb schema definition (SDL).

**Question**: I opted to include `include-searchable-encryption` as an user argument, since the `SchemManager.Generate` takes it as a parameter. We could just hardcode this to true, and it always output schema that supports searchable encryption types (even though the actual user type might not have any `@encryptedIndex` defined)

## How has this been tested?

Added embedded CLI tests, along with manual testing. Happy to add more if there's a preference.

- Ubuntu (WSL2)


internal/db/description/collection_cache.go
```diff
@@ -26,13 +26,21 @@ func InitCollectionCache(ctx context.Context) context.Context {
...
+// It is done to avoid an extra check to see if the cache exists or not when fetching
+// it from the context.
+func WithCollectionCache(ctx context.Context, cache *CollectionCache) context.Context {
+	return context.WithValue(ctx, collectionCacheKey{}, cache)
+}
+
 // getCollectionCache retrieves the collection short-id cache from the given context.
-func getCollectionCache(ctx context.Context) *collectionCache {
-	return ctx.Value(collectionCacheKey{}).(*collectionCache) //nolint:forcetypeassert
+func CollectionCacheFrom(ctx context.Context) *CollectionCache {
```
fredcarle: thought: We use different naming conventions for storing and retrieving from context across the codebase so it's not obvious when reading the name of this fonction what is happening.

cli/sdl_generate.go
```diff
@@ -0,0 +1,291 @@
...
+			} else {
+				// check if the file exists, if so check for the overwrite
+				// flag
+				ofinfo, err := os.Stat(outputFile)
+				if err != nil && !errors.Is(err, os.ErrNotExist) {
+					return err
+				}
+				if ofinfo != nil && !yesOverwrite {
+					fmt.Fprintln(os.Stderr, "output file path already exists. If you want to overwrite use -y")
+					os.Exit(1)
```
fredcarle: suggestion: Return an error instead of calling `os.Exit(1)`

cli/sdl_generate.go
```diff
@@ -0,0 +1,291 @@
...
+	cmd.PersistentFlags().BoolVarP(&yesOverwrite, "overwrite", "y", false,
+		"Overwrite any existing matching output file paths")
+
+	cmd.PersistentFlags().BoolVarP(&searchableEncryption, "include-searchable-encryption", "s",
+		false, "Include the schema type definitions to support Searchable Encryption")
+
+	return cmd
+}
+
+var introspectionQueryRequest = `
```
fredcarle: suggestion: There are about 1000 whitespace characters in this variable that will take space on the binary (about 1kB). Since this probably wont change over time, it might make sense to inline everything to reduce size. Everything counts if we want to minimize binary size :)

cli/sdl_generate.go
```diff
@@ -0,0 +1,291 @@
...
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var inputBuf io.Reader
+
+			// Either we use stdin or we concat all the file
+			// arguments
+			if len(args) == 1 && args[0] == "-" {
+				inputBuf = os.Stdin
+			} else {
+				fileInputBuf := bytes.NewBuffer(nil)
```
fredcarle: suggestion: `string.Builder` would have about half the allocations and almost twice as fast as using a buffer.

fredcarle (approved):
Approving with a few comments to resolve before merge.

fredcarle (approved):

cli/sdl_generate.go
fredcarle, line 1: thought: It would be nice to have a test for this in cli/test.

jsimnz: I had written up a test, but I wasn't happy with how to actually write the "expected" side of the test.

You can see here an example of the full generated SDL for *just* the single basic `User` type from the `examples/schema/user.graphql` file: https://gist.github.com/jsimnz/b9c42f4ba9550f3f7cd1b19652bdd4d9.

Its far too large/complex to write out some sort of test output by hand, we could do something similar to the introspection integration tests where we just test for certain subsections? Or we could include that file and reference it in the test, but that is somewhat circular without very tedious review of the file, since that file output comes from the code we want to test in the first place. 

Open to suggestions. As for other tests, the `EmbedExample` tests the structure/options/args already.

internal/db/description/collection_cache.go
jsimnz: Lmk if the changes make more sense

nasdf (approved):
LGTM. I'm excited to try out all of the gql codegen libraries!

internal/db/description/collection_cache.go
fredcarle: Better. Thanks.

cli/sdl_generate.go
fredcarle: > Or we could include that file and reference it in the test, but that is somewhat circular without very tedious review of the file, since that file output comes from the code we want to test in the first place.

Using a file would be my preferred approach. I know there is the circular aspect to it but if we agree that what it is currently producing is correct, then it will catch if anything changes inadvertently.

jsimnz: Following up here, turns out that the schema output will basically always be a semi-random in some way shape or form. Theres some ideas I have to even out this random-ness, but I dont think it needs to be a blocker at the moment for this.

I'll find some time to do it, low prio for v1

go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/valyala/fastjson v1.6.4
 	github.com/vito/go-sse v1.1.3
+	github.com/wundergraph/graphql-go-tools/v2 v2.0.0-rc.240
 	github.com/zalando/go-keyring v0.2.6
```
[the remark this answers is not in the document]
jsimnz: this is intentional

internal/request/graphql/schema/manager_npx_test.go
```diff
@@ -0,0 +1,119 @@
...
+	err = manager.WriteSDL(outBuf)
+	require.NoError(t, err)
+	sdlResult := outBuf.Bytes()
+
+	testFixturePath := getFullFixturePath(fixtureName)
+
+	cmd := exec.Command("npx", "-y", "@graphql-inspector/cli",
+		"diff",
+		string(sdlResult),
+		testFixturePath)
```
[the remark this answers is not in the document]
jsimnz: You are incorrect here, `@graphql-inspector/cli` does accept raw strings as input

cli/sdl_generate.go
jsimnz: @fredcarle, I decided to significantly increase the test coverage, both with unit tests, and a CLI tests (not part of the existing CLI integration tho).

PR #4264, opened 2025-12-09
feat: Add methods AddLens and ListLenses

### LINKED ISSUE #4262: Add AddLens method
We need a method for adding lens to defra that returns it's CID. That we can use later to pass `AddView`

Required for https://github.com/sourcenetwork/defradb/issues/4162

islamaliev:
## Relevant issue(s)

Resolves #4262

## Description

This change introduces `AddLens` and `ListLenses` methods that allows adding and retrieving lenses to/from defra. 

The CID returned by `AddLens` can be used for example for `PatchCollection`, but opens open possibility to improve other API methods that currently take entire lens modules like `AddView`. Instead they can be change to accept just lens id

### DIFF
--- a/http/handler_store.go
+++ b/http/handler_store.go
@@ -181,6 +181,22 @@ func (h *storeHandler) AddLens(rw http.ResponseWriter, req *http.Request) {
 	responseJSON(rw, http.StatusOK, &AddLensResponse{LensID: lensID})
 }
 
+type ListLensesResponse struct {
+	Lenses map[string]model.Lens `json:"lenses"`
+}
+
+func (h *storeHandler) ListLenses(rw http.ResponseWriter, req *http.Request) {
+	db := mustGetContextClientDB(req)
+
+	lenses, err := db.ListLenses(req.Context())
+	if err != nil {
+		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+		return
+	}
+
+	responseJSON(rw, http.StatusOK, &ListLensesResponse{Lenses: lenses})
+}
+
 func (h *storeHandler) GetCollection(rw http.ResponseWriter, req *http.Request) {
 	db := mustGetContextClientDB(req)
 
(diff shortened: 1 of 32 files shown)


http/handler_store.go
```diff
@@ -747,5 +831,7 @@ func (h *storeHandler) bindRoutes(router *Router) {
 	router.AddRoute("/debug/dump", http.MethodGet, debugDump, h.PrintDump)
 	router.AddRoute("/schema", http.MethodPost, addSchema, h.AddSchema)
 	router.AddRoute("/lens", http.MethodPost, setMigration, h.SetMigration)
```
AndrewSisley: suggestion:  Unsure whether this should be a todo or not...

We probably want to move the `SetMigration` http path before v1.  And it's current path is where `AddLens` should probably exist (bringing it into scope of this PR).  This way `Add` (POST) and `Get` (GET) live in the same location, then the 'special' `SetMigration` lives somewhere else, maybe even somewhere under `/collection`.

internal/db/lens.go
```diff
@@ -35,6 +36,27 @@ func (db *DB) getLensStore(ctx context.Context) store.Store {
 	return db.lensNode.Store
 }
 
+func (db *DB) addLens(ctx context.Context, lens model.Lens) (string, error) {
+	cid, err := db.getLensStore(ctx).Add(ctx, lens)
```
AndrewSisley, line 40: thought: The latest lens code (unreleased) simplifies both of these functions into one-liners, as the return strings instead of cids.

We need to bump the Defra Lens version before v1, and this could be the excuse we need to do it now.

Probably better to do the version-bump in another PR anyway, but I'd thought I'd note this here as removes basically all of the production logic within this PR.

tests/action/list_lenses.go
```diff
@@ -0,0 +1,95 @@
...
+	//
+	// If a value is not provided the lenses will be listed from all nodes.
+	NodeID immutable.Option[int]
+
+	// The identity of this request. Optional.
+	Identity immutable.Option[state.Identity]
+
+	// ExpectedCount is the expected number of lenses.
+	// If set, the action will verify the count matches.
+	ExpectedCount immutable.Option[int]
```
AndrewSisley: suggestion: I really think actions should take the same return type as the function under test.  I.e. `map[string]model.Lens`.

Not doing so bakes in testing limitations, hides the API under test, and creates very tedious work for us when we decide to change things to match the actual API (or creates a situation where we let the legacy props linger forever).

The implementation of `Execute` can allow for the omission or complex testing of certain properties without properties like this.

IMO this is a legacy mistake that can be found in our older actions, and should not be copied into our newer ones.

This includes the `int` in `map[int]model.Lens` - we have standardized `replace` syntax for handling exactly this kind of thing.

Might be worth a discord discussion thread if my opinion is controversial here.

AndrewSisley (approved):
LGTM, and thanks for adding these :)  Couple of suggestions to look at before merge though.

internal/db/lens.go
AndrewSisley: Quick warning on this, Keenan has just opened up a Lens PR, so maybe hold off on that release for now (no point doing it twice)

tests/action/list_lenses.go
islamaliev: done

http/handler_store.go
islamaliev: done

internal/db/lens.go
islamaliev: I don't see any related PRs in lens repo and defra is using the latest version v0.9.4

AndrewSisley: Keenan closed his PR on Friday, and if you have a look at the commits on `main` you'll see that the latest changes haven't been released yet.

https://github.com/sourcenetwork/lens/pull/151

AndrewSisley: Maybe have a quick chat with Keenan and find out what his plans are if you are thinking of releasing Lens, otherwise it might be wasted effort.

islamaliev: Keenan said he won't merge that changes and I'm free to merge this one

PR #4273, opened 2025-12-12
feat: Introduce setup wizard

### LINKED ISSUE #4189: Setup Wizard: First Stage
This is a concrete feature request from #3968.

Add a command-line based wizard setup script, which asks the following questions:

1. ...

    > DefraDB protects the storage and transmission of data with a keypair that will be generated now. You have the choice of where to store these generated keys.
    > Where do you want to store your keypair?
    >    - Filesystem (~/.defradb/keys)
    >    - OS (KeyChain, etc)

    - if "Filesystem" is chosen:

        > To protect these keys with encryption on the filesystem, please provide a secret (minimum X characters):
        > Where do you want to store your secret?
        >    - Filesystem (~/.env)
        >    - OS (KeyChain)

        - if "OS" is chosen

            (more questions?)

    - "OS" is chosen:

        (more questions?)

2. ...

    > Where do you want to store the configuration of these keypair/secret settings? (~/.defrab/config.yaml)

----

@shahzadlone suggests the following, for a polished interactive UI experience:

https://github.com/charmbracelet/bubbletea
https://github.com/charmbracelet/bubbles
https://github.com/charmbracelet/gum

### LINKED ISSUE #4190: Setup Wizard: Add "Import Keypair" question
"Do you want to import an existing keypair, or generate a new keypair?" (with some explanatory text about the implications/requirements of each choice)

per @jsimnz, also implies the existence of a protected "export keypair" operation (probably from the client).

ChrisBQu:
## Relevant issue(s)

Resolves #4189
Resolves #4190 

## Description

This PR introduces a new sub-package inside `cli`, `wizard`. The entire directory can be thought of as the code for a self-contained TUI application built using the Bubbletea framework (see: https://github.com/charmbracelet/bubbletea).

The Main function and program structure is defined in `cli/wizard/wizard.go`. This is the entry point of the application. It will be invoked after running `make install` and can be manually invoked through the CLI by running `defradb wizard`.

The structure of the program is such that the main Bubbletea model is defined inside `cli/wizard/model_main.go`, but it can be thought of as a state machine. It tracks sub-models as "steps" and, is updating the logic and view of one step at a time. Different submodels are defined in such a way as to be modular and re-usable. I have basic text models, multiple choice models, text input models, etc. 

Each sub-model can be assigned a next-step, or a series of next-steps. For example, there is `model_brancher` which takes in an evaluator function to know which step to branch to. Callbacks have also been implemented -- that is, functions can be assigned to each sub-model node, which will be invoked as the step finishes, and before we move to the next step.

This initial version of the wizard has the following functionality:

- Firstly, it confirms that the user wishes to continue with the setup wizard. On yes, the program continues, and on No, the program exits.

- Next, a config.yaml file will be generated, and the generation will be communicated to the user.

- Next, the user will be asked to choose where the keyring is stored, with the choices being filesystem or OS.

- If the user chooses Filesystem, the application will check for the existence of a DEFRA_KEYRING_SECRET environment variable. If it exists, the program will continue. If it doesn't, the program will ask the user if they'd like to create a .env file. Selecting no will report to the user how to manually set the environment value, and how to re-run the wizard, and it will then gracefully exit. Selecting yes will cause the program to ask the user for the secret value they'd like to use, and it will then create (and load/use) the .env value. If the user chooses OS, no secret value will be needed. In either case, the program flow will then continue.

- The user will be asked if they wish to import any existing keys. If they select yes, they will be given a sequence of prompts. Firstly, they will be asked to either generate or import the identity key, as it is required. For the other choices, peer, encryption and searchable encryption keys, importing is optional. For each choice, the user can input text (i.e. by pasting) to import the key. In the case of identity key, they will additionally be asked the key type. For each step, the application will confirm if the keys were generated successfully. 

- If the user selects not to import any keys, the keys will all be generated. For each key, the application will confirm the successful generation of the key.

**This needs to be tested on Mac before merge.**

This PR contains a set of unit tests, one for each of the callback and evaluator functions.

## Tasks

## How has this been tested?

- WSL


Makefile
```diff
@@ -83,6 +83,7 @@ default:
 .PHONY: install
 install:
 	@go install $(BUILD_FLAGS) ./cmd/defradb
+	@defradb wizard
```
shahzadlone, line 88: suggestion: Maybe we introduce another command called `make install:wizard` that calls the `make install` and then does this extra defradb wizard step (pretty much exactly what you have now). This is because I have a preference to leave the `make install` command to not be interactive and dependent on user input as a lot of scripts and tools can be utilizing this command to only install defradb

As it stands of course doing make install will give this:
image

shahzadlone (commented):
I think most of the implementation is straightforward and as I expected.

Just left a suggestion, and here are some questions I had.

question/thought: I see that the bubble tea specific stuff is mixed in within the same wizard package. I was wondering if it would make sense to have the independent bubble tea stuff in it's own package perhaps, if we want to use it else where too? For now I am happy with it as it is, but just wondering if for example all cli commands can take advantage of the interactive bubble tea ui to select different options etc. 

question: In my `make install` interactive flow that I tested on this PR there was no color. Wondering if that was intentional? IMO the bubble tea colors like the ones in their demos would jazz up our cli interaction significantly (not a blocker for the first PR in anyway).

wizard/callbacks.go
```diff
@@ -0,0 +1,79 @@
...
+	"errors"
+	"os"
+
+	"github.com/sourcenetwork/defradb/crypto"
+	"github.com/sourcenetwork/defradb/keyring"
+)
+
+// This callback will set keyring.backend to either "file" or "system"
+func callback_SetKeyringBackend(s step, ctx *WizardContext) error {
+	mm := s.(*modelMultipleChoice) //nolint:forcetypeassert
```
shahzadlone: question: why do we need the `nolint` here?

suggestion:
```suggestion
	mm, ok := s.(*modelMultipleChoice)
	if !ok {
		return errHandleHere
	}
```

wizard/callbacks.go
```diff
@@ -0,0 +1,79 @@
...
+
+// This callback will generate the keyring files
+func callback_GenerateKeyringFiles(_ step, ctx *WizardContext) error {
+	passwordStr, ok := os.LookupEnv("DEFRA_KEYRING_SECRET")
+	if !ok {
+		return errors.New(errDefraKeyringSecretNotSet)
+	}
+	cfgFile := getConfigFile()
+	keyringFilepath := getYAMLValueInFile(cfgFile, []string{"keyring", "path"})
+	fullKeyringFilepath := os.Getenv("HOME") + "/.defradb/" + keyringFilepath.(string) //nolint:forcetypeassert
```
shahzadlone: question/suggestion: Same question and suggestion as above.

wizard/callbacks.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// This callback will generate the keys in the system keyrind
+func callback_GenerateKeyringFilesInSystemKeyring(_ step, ctx *WizardContext) error {
+	cfgFile := getConfigFile()
+	keyringNamespace := getYAMLValueInFile(cfgFile, []string{"keyring", "namespace"}).(string) //nolint:forcetypeassert
```
shahzadlone: question/suggestion: Same question and suggestion as above.

cli/wizard/consts.go
```diff
@@ -0,0 +1,24 @@
...
+
+const (
+	// Keyboard keys
+	KEY_CONTROL_C = "ctrl+c"
+	KEY_ENTER     = "enter"
+	KEY_SPACE     = " "
+	KEY_UP        = "up"
+	KEY_DOWN      = "down"
+	KEY_K         = "k"
+	KEY_J         = "j"
```
shahzadlone, line 21: question: k is up and j is down as well right, it worked for me so I am guessing it's the vim key pattern.

ChrisBQu: Yes, that's exactly right. Technically, it doesn't have to be defined as a constant here, because it's not repeated in more than one place. But it felt like good practice, since we were defining the other keyboard keys as consts to put them all together. In the future, if different sub models want to use cursor, this is right here for it.

Makefile
ChrisBQu: Good call. I will be making this change.

ChrisBQu:
> question: In my make install interactive flow that I tested on this PR there was no color. Wondering if that was intentional? IMO the bubble tea colors like the ones in their demos would jazz up our cli interaction significantly (not a blocker for the first PR in anyway).

It was intentional, but it is not my intention to leave it that way. Today I'm going to be sharing the design at the dev ex meeting. If people like it overall, I will add color. It didn't seem like an early priority, but its easy to add in.

> question/thought: I see that the bubble tea specific stuff is mixed in within the same wizard package. I was wondering if it would make sense to have the independent bubble tea stuff in it's own package perhaps, if we want to use it else where too? For now I am happy with it as it is, but just wondering if for example all cli commands can take advantage of the interactive bubble tea ui to select different options etc.

Hmmm. I see the value in this. I'm not planning to touch the CLI in this specific series of PRs, but it might make sense to separate it out. Let's bring that one up in standup. You surely noticed that even though its in the wizard package, I tried to do my best to separate logic into ways that things are modular and re-usable. We can definitely separate those out into a separate package if we want.

wizard/callbacks.go
ChrisBQu: We must have these nolint directives in order to placate the linter. It will complain with "type assertion must be checked (forcetypeassert)" otherwise.

In the case of line 23, because we are creating this function specifically to assign to a submodel of that type, we know what type we can assert it to. This will be safe to do. But in theory, there would be a panic if we attached this callback to the wrong type of submodel.

In the other cases, its because `getYAMLValueInFile` returns `any`. Depending on what we request, we might not get a string back. But here... we are requesting a string, and after we get it, we can start treating it like a string. But theoretically, we could request the wrong type of value and it would panic.

In either of these cases, do you feel I am abusing the `forcetypeassert` directive, and do you desire we do it another way?

Edit: Sorry, I explained why I did it, but I missed part of your comment with the code suggestion. I guess I'm not super familiar with the linter's behavior. If catching the error is enough to satisfy the linter, than I like that much better and will do it. I'll leave the reply I wrote, but it might be obsolete now. :)

wizard/io.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+// getConfigFile returns the path to $HOME/.defradb/config.yaml
+func getConfigFile() string {
```
nasdf: suggestion: I think it'd be cleaner to use the existing config functions from the cli package. Viper can do most of what is in this file.

https://github.com/spf13/viper/tree/v1.21.0?tab=readme-ov-file#getting-values-from-viper

ChrisBQu: It does, but remember we cannot import the CLI package, because it creates a cyclic dependency.

We could import Viper in each of the two packages, but I think this approach would couplesthe wizard to the CLI code in a weird way, where we are relying on context provided by the CLI. We can't get arround that entirely. We need to be able to generate a config.yaml file. But I think it should be as minimal a connection thread as possible.

Ultimately, if you have strong feelings, I can do how you'd like. But you might have to talk me through some of the specifics of what you'd like, because right now I'm not sure it makes complete sense to me.

nasdf: If the base config logic (load, save, etc.) was moved to a sub-package of cli you should be able to avoid the cyclical dependency.

ChrisBQu: Right on, I get it. Let's do it.

wizard/callbacks.go
shahzadlone: > If catching the error is enough to satisfy the linter, than I like that much better 

Yes so this linter suppression should only be done where we want it to panic, all other cases should properly handle the error.

shahzadlone (commented):
question: Is there a reason why the `wizard` package is not under the `cli` package?

ChrisBQu:
> question: Is there a reason why the `wizard` package is not under the `cli` package?

Hmm. Until just now, I hadn't considered it. It seemed distinct because it doesn't utilize the same architecture, built of of Viper commands. But I guess... it still kind of is? I'm not against moving it to there. We should reach team consensus on this one I think.

getify:
> If they choose Filesystem, it will then check for the existence of a DEFRA_KEYRING_SECRET environment variable. If it is present, it will generate the keys. If it is not present, it will display an error message, and tell the user how to set the key and re-run the wizard.

This part seems unusual and contrary to the intended spirit of the wizard, unless I'm misunderstanding.

Someone wouldn't not have set this PRIOR TO `make install` would they? So then it's certain that the wizard will sort of stop/fail at this point under normal circumstances.

If I'm correct there, why wouldn't we just have the wizard prompt the user for the secret (if they haven't set it up) and set the ENV for them -- and of course, ask where they want that secret to persist?

ChrisBQu:
> > If they choose Filesystem, it will then check for the existence of a DEFRA_KEYRING_SECRET environment variable. If it is present, it will generate the keys. If it is not present, it will display an error message, and tell the user how to set the key and re-run the wizard.
> 
> This part seems unusual and contrary to the intended spirit of the wizard, unless I'm misunderstanding.
> 
> Someone wouldn't not have set this PRIOR TO `make install` would they? So then it's certain that the wizard will sort of stop/fail at this point under normal circumstances.
> 
> If I'm correct there, why wouldn't we just have the wizard prompt the user for the secret (if they haven't set it up) and set the ENV for them -- and of course, ask where they want that secret to persist?

I agree with this and it was actually my original intention. Its why the text input model has already been written. Originally instead of erroring out it would ask the user for the secret. But theres a problem.

There... isnt a good way for a Go program to set a system wide environment variable. The os.SetEnv function... does not do what you might think it does.

It was actually Keenan who suggested that it should error on the environment variable not being present. But if we have a better way to do it, I'd like to do it that way.

ChrisBQu:
A bit of follow up: After some discussion, to address the environment variable issue: we will be prompting for the generation of a .env file for use with the wizard, if the variable is not set. The user will be able to decline this operation, in which case they will be given the commands to set, on their own, the environment variable. And if they choose to create the file, the wizard will continue.

cli/wizard/callbacks.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	keyring, err := keyring.OpenFileKeyring(keyringFilepath, []byte(passwordStr))
+	if err != nil {
+		return err
+	}
+	encryptionKey, err := crypto.GenerateAES256()
+	if err != nil {
+		return err
+	}
+	err = keyring.Set("encryption-key", encryptionKey)
```
nasdf: todo: I think we should support the same keys that the start command does.

- `peerKey`
- `encryptionKey`
- `searchableEncryptionKey`
- `identityKey`

Some of them are optional and should have their own prompts and config values saved.

- `datastore.noencryption`
- `datastore.nosearchableencryption`

https://github.com/sourcenetwork/defradb/blob/develop/cli/start.go#L142

cli/wizard/utils_test.go
nasdf, line 1: todo: this file should end with `_test.go` otherwise it will be included in the normal build.

nasdf (changes requested):
The wizard works great on MacOS. I think we should add those additional keys though.

cli/wizard/utils_test.go
ChrisBQu: Done

cli/wizard/callbacks.go
```diff
@@ -44,6 +44,7 @@ func callback_GenerateConfigYAMLFile(_ step, ctx *WizardContext) error {
 
 // This callback will generate the keyring files
 func callback_GenerateKeyringFiles(_ step, ctx *WizardContext) error {
+	// Get the DEFRA_KEYRING_SECRET and keyring.path, and open the keyring
```
shahzadlone: suggestion: I personally think this comment is not needed. Is very easy to read the code and see what is happening. It confused me first because I didn't understand how these 3 things are all being done in the `os.LookupEnv` command as that is where it is documented

ChrisBQu: Agree -- removed it. That one seems superfluous even for me, and I definitely write more comments than the average dev. You have me wondering if maybe I should strip down comments across the wizard wholesale though.

shahzadlone: > Agree -- removed it. That one seems superfluous even for me, and I definitely write more comments than the average dev. You have me wondering if maybe I should strip down comments across the wizard wholesale though.

I do the same haha! I wouldn't bother for the rest unless you really feel a hard preference to remove some

cli/wizard/callbacks.go
```diff
@@ -0,0 +1,336 @@
...
+	if !ok {
+		return errors.New(errFailedToGetKeyringNamespace)
+	}
+	keyring := keyring.OpenSystemKeyring(keyringNamespace)
+	return generateKeysInKeyringFromStep(ctx, keyring, "stepSelectKeyTypes")
+}
+
+// This callback loads the environment variables from the .env file
+func callback_SetAndReloadDefraKeyringSecretEnvironmentVariable(_ step, ctx *WizardContext) error {
+	stepToRetrieveResultFrom := "stepGetDefraKeyringSecretInput"
```
nasdf: nitpick: I think the step names should be moved to constants to avoid typos.

cli/wizard/utils_test.go
nasdf: I don't see the change. Is this a GitHub issue?

nasdf (approved):
I tested on MacOS with the file and os ring types and it worked great. I left a comment about the step names that you may want to consider.

getify:
amazing work, @ChrisBQu! way to go!

PR #4274, opened 2025-12-16
feat: Rename foreign keys to _<rel>ID

### LINKED ISSUE #1751: Rename related/foreign id instances from `<related>_id` to `_<related>ID`
Part of https://github.com/sourcenetwork/defradb/issues/1750

Rename related/foreign id instances from `<related>_id` to `_<related>ID`

islamaliev:
## Relevant issue(s)

Resolves #1751

## Description

BREAKING CHANGE

Rename all relation id fields from `author_id` to `_authorID`.


shahzadlone:
@islamaliev thanks for picking this up as it's been something we have been wanting to rename for a while now. Pinging @ashutosh-src here to notify all the document related changes that would need to be done before v1 docs are finalized.

internal/planner/mapper/mapper.go
```diff
@@ -1261,7 +1261,10 @@ func resolveSecondaryRelationIDs(
 			continue
 		}
 
-		objectFieldName := strings.TrimSuffix(existingField.Name, request.RelatedObjectID)
+		objectFieldName, ok := request.ToRelatedObjectName(existingField.Name)
+		if !ok {
+			continue
+		}
```
shahzadlone, line 1267: question: Why do we need to do this?

islamaliev: actually no need, it should never happen. I just felt it was rude to ignore it :D 
I don't have any preferences here. Can ignore.

ghost:
> @islamaliev thanks for picking this up as it's been something we have been wanting to rename for a while now. Pinging @ashutosh-src here to notify all the document related changes that would need to be done before v1 docs are finalized.

Whats the action item from my end? @shahzadlone. I see some variable name format change in .md files.

islamaliev:
> Whats the action item from my end? @shahzadlone. I see some variable name format change in .md files.

@ashutosh-src we would need to update all documents and guidelines that user old name for related object id.
For example, with these collection:
```graphql
type Author {
  name: String
}

type Book {
  title: String
  author: Author
}
```
the `Book` collection in old way would reference `Author` with field `author_id`. The new way (what this PR changed) the field name is `_authorID`.

shahzadlone (approved):
Thank you so much for taking care of this, we been wanting this for a while.

LGTM

PR #4276, opened 2025-12-16
refactor: Introduce collection lock system

### LINKED ISSUE #4275: Introduce collection-lock system

AndrewSisley:
## Relevant issue(s)

Resolves #4275
Required for #4207

## Description

Introduces a new lock system, added here for collection-specific locks, allowing specific keys (i.e. collections) to be write locked on demand.

Read locks are held when the key/collection is accessed via the datastore.  All locks are unlocked on transaction close.

This system is generic, and should allow us to easily expand upon it, allowing locks of other scopes, such as around specific documents, as well as allowing users to perform atomic operations, and manually manage locks (if we ever want to permit that).

IMO most of this code needs to be written whether we lock at the datastore level using the `Keyedstore` interface or not, as most of the code protects against re-acquiring the same lock multiple times (which would cause a deadlock) - given that our public functions often calls other public functions, this is a problem regardless of whether we go with this automatic RLock approach or not.

The (WIP) truncate code making use of this can be found https://github.com/sourcenetwork/defradb/compare/develop...AndrewSisley:4207-truncate?expand=1


internal/db/lock/lock_set.go
jsimnz, line 1: question: This file (and a few others) have an unusually large amount of comments/descriptions/documentation describing the semantics of the implementation. Did you feel that the implementation didn't have clarity on its own (assuming of course a "standard" amount of comments)? Or were there explicit semantics you wanted to be sure were clear?

internal/db/lock/lock_set_race_test.go
```diff
@@ -0,0 +1,138 @@
...
+// bugs in production.
+func TestLockSet_ConcurrentRWLockRLockForSameTxnKey_ClosesCorrectly(t *testing.T) {
+	txn1 := newTxn(1)
+	txn2 := newTxn(2)
+	lockSet := newLockSet[int]()
+
+	wait := sync.RWMutex{}
+	// Lock the wait in order to make sure all concurrent locks start as close as possible to
+	// the same time.
+	wait.Lock()
```
jsimnz, line 97: praise: This is a cool trick/use of mutexes! At first I thought I should be using a `Waitgroup` but later realized what you doing. Neat.

jsimnz (approved):
I'm approving now.

With one `todo` (feel free to push back on this). I would love to get a very high level microbenchmark together to understand *any* perf impact this might have. I 100% recognize that acquiring a lock on its own isn't too expensive, but A) There are potentially 9 lock/unlock calls for every `collectionRLock` call, and `collectionRLock` is called on *every single* individual datastore operation. So in aggregate in larger and more complex queries that is hitting a lot of documents, there could start to be some level of notable perf implications.

Caveat: my approval here doesn't imply my *agreement* with this. I do strongly feel this is overengineered/overly complicated for the goal. This is exemplified by all the different lock types (`lockSet`, `abLock`, `txnLock`) each with their own set of (nested) maps of locks. This combined with the extensive documentation/commenting needed to convery the design (of which it took me quite a while to feel like I fully groked all the interwoven semantics and potential edge cases), and the number of warnings throughout is causing me notable concern, given some alternative simpler (naive) implementations.

However, you have done a *very* good job of scoping 99% of the semantics to just the `lock` package (only a single line minorly really leaked into the implementation, the version fetcher stuff, understandable). So I am confident in approving now and changing later without any downstream breaking changes or large refactors in the outer layers 🙏.

Happy to receive push back on my concerns here, but the complexity is undeniably self-evident, the question becomes is the complexity warranted.

All that being said, the fact that you kept all this in your head is impressive 💪, but also highlights my concerns 😂.

AndrewSisley:
> I would love to get a very high level microbenchmark together to understand any perf impact this might have. 

It might be pretty quick for me to add them, unit test style within the lock package, and it is sensible suggestion - let me know soon if you *dont* want me to add any before merge.

> There are potentially 9 lock/unlock calls for every

After the first read call within a transaction (per collection), there are only two (un)locks per `RLock` call, as `RLock` exits very early if a lock is already held, so it should be very cheap still (the map reads are probably more expensive than the locks).

> but the complexity is undeniably self-evident

100% and my belief that most of this complexity is required either way is why I put the effort in.  I think it will be pretty complicated and add a decent amount of thought-overhead to the rest of the codebase if we do not centralise and abstract this.

I think one of the problems with this discussion is that both of us are essentially crystal-balling how painful and error prone it would to do this *decentralized*, and without giving the decentralized route a go, and witnessing it's long term effects we cannot be sure about the cost-benefit of doing this.  I believe it would be a nightmare, and that locks are really not something I think we should be playing with day-to-day, and I do think the system in this PR will make it a lot easier to add related features (such as atomic operations), but I can't back that up with any kind of metrics and the two of us could probably talk this topic to death without reaching consensus :)

internal/db/lock/lock_set.go
AndrewSisley: A bit of both, this is highly concurrent code, and it is often not obvious IMO as to why some (most?) things are done by reading the code alone.

This is worsened by my opinion that most developers generally dont spend a whole lot of time playing with locks etc., and so I've also tried to accommodate a little for that too.

Additionally, a lot of bugs that can be introduced by lock misuse can be both very bad for users (and for business), and are relatively easy to sneak through testing (as our manually defined tests rely on fairly simple, and usually synchronous, operations that lack the chaos of a real world database - this makes them much easier to debug, but they will not flag a wide range of issues).  I'd love us to have a suite of chaos tests, and we might even be able to use a multiplier in the nightly CI for some of this, but we haven't had the time to build that yet.

internal/db/lock/lock_set_race_test.go
AndrewSisley: :grin: cheers, it feels a little dirty, but I was slightly pleased with myself when I realised I could just do this instead lol

internal/datastore/concurrent_txn_test.go
islamaliev, line 1: todo: typo in file name. `concurrent_txn_test.go`

internal/datastore/datastore.go
```diff
@@ -65,11 +80,26 @@ func (s *datastore) Iterator(ctx context.Context, opts IterOptions) (corekv.Iter
...
 
 func (s *datastore) Set(ctx context.Context, key Key, value []byte) error {
+	s.collectionRLock(ctx, key)
+
 	keyBytes := key.Bytes()
 	return s.underlying.Set(ctx, keyBytes, value)
 }
 
 func (s *datastore) Delete(ctx context.Context, key Key) error {
+	s.collectionRLock(ctx, key)
```
islamaliev, line 90: question: should be acquire write lock here?

internal/db/lock/ab_lock.go
```diff
@@ -0,0 +1,45 @@
...
+// other LockAs from being acquired, and vice versa.
+//
+// Acquired locks are released on transaction close.
+type abLock struct {
+	aGroup sync.WaitGroup
+	bGroup sync.WaitGroup
+}
+
+func (l *abLock) LockA(txn txn) {
+	l.aGroup.Add(1)
```
islamaliev, line 30: question: what is the chance of `LockA` and `LockB` to be called concurrently?

If that happens, it will deadlock.

internal/db/lock/ab_lock.go
```diff
@@ -0,0 +1,45 @@
...
+	txn.OnDiscard(func() { l.aGroup.Done() })
+	txn.OnError(func() { l.aGroup.Done() })
+	txn.OnSuccess(func() { l.aGroup.Done() })
```
islamaliev: question: is it 100% guaranteed that only one of them will be triggered?

If not maybe we could you use:
```go
var once sync.Once
done := func() { once.Do(func() { l.aGroup.Done() }) }
txn.OnDiscard(done)
txn.OnError(done)
txn.OnSuccess(done)
```

internal/db/lock/lock.go
```diff
@@ -0,0 +1,56 @@
...
+// This will prevent all other transactions from acquiring a read or write lock
+// to the given collection until the lock is released.  The lock will be released
+// when the transaction is either committed or discarded.
+//
+// The acquired lock will not block other threads operating within this transaction.
+func (l *LockSet) CollectionLock(txn txn, collectionShortID uint32) {
+	l.collectionLockSet.Lock(txn, collectionShortID)
+}
+
+// CollectionLock acquires a read lock for the given collection short id.
```
islamaliev: typo: `CollectionRLock`

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,381 @@
...
+// in keeping with other systems - it seems very unlikely that locks *should* ever be managed independently
+// from their transactions, as if a transaction desires a different lock scope, then it should probably be
+// doing whatever it is doing in multiple transactions, or without a transaction, instead.
+//
+// If a requested lock is already held, the requesting transaction will be blocked from acquiring any new
+// locks, and from unlocking held locks, until the requested lock becomes available.  For more info on this,
+// see the documentation for the `txnLocks` member of this type.
+//
+// Only public functions on this type should be called from outside of this type.  Calling private members
+// from anywhere else, internal or otherwise, removes any guarentees over its state's correctness.
```
islamaliev: typo: "guarantees"

internal/db/lock/lock_set_test.go
```diff
@@ -0,0 +1,310 @@
...
+func (t *dummyTxn) OnSuccess(fn func()) {
+	t.onSuccess = append(t.onSuccess, fn)
+}
+
+func (t *dummyTxn) OnError(fn func()) {
+	t.onError = append(t.onSuccess, fn)
+}
+
+func (t *dummyTxn) OnDiscard(fn func()) {
+	t.onDiscard = append(t.onSuccess, fn)
```
islamaliev: todo: Should be `append(t.onError, fn)` and `append(t.onDiscard, fn)`

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,381 @@
...
+//
+// If the transaction already has the write lock, this will no-op.
+//
+// If the transaction already has a read lock, it will promote the read lock into a write lock.
+// WARNING: It currently does this by unlocking the existing read lock, and then acquiring the write lock -
+// this can unblock other threads trying to acquire the lock, as well as the thread holding the original
+// read lock.
+//
+// If a lock was acquired, it will be unlocked on transaction commit/discard.
+func (l *lockSet[TKey]) Lock(txn txn, key TKey) {
```
islamaliev, line 119: todo: is it guaranteed that it's going to be called from a single thread?

If yes, please add documentation. If no, there is a lot of racing here which should be either fixed or documented explaining why we tolerate it.

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,381 @@
...
+		// until after the potential write without introducing a race condition.
+		l.locksByKeyLock.Lock()
+		var ok bool
+		lock, ok = l.locksByKey[key]
+
+		if !ok {
+			lock = &sync.RWMutex{}
+			// If the lock key does not exist yet, we must add it so
+			// that the read lock can be held if a write lock is attempted.
+			l.locksByKey[key] = lock
```
islamaliev, line 164: suggestion: locks are added to `locksByKey` but never removed. With many unique keys (e.g., document IDs), this map grows unbounded. Consider periodic cleanup of unused locks or something else

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,381 @@
...
+			lock.RUnlock()
+
+			// Because the implementation of `*lockSet[TKey].RLock` only actually locks the mutex if it isn't
+			// already held, we can safely delete it from `l.heldRLocksByTxnID` here.
+			//
+			// Because this `rUnlockAll` function is locked per transaction, we do not need to worry about the
+			// same transaction concurrently mutating (or reading) from `locks`.
+			delete(locks, key)
+		}
+	}
```
islamaliev, line 304: suggestion: `delete(locks, key)` seems unnecessary - the `locks` map was already removed from `heldRLocksByTxnID` and will be garbage collected. Same applies to `unlockAll`

internal/keys/datastore.go
```diff
@@ -13,3 +13,18 @@ package keys
...
+// concrete type.
+//
+// An important useage of this is to do with acquiring read locks whenever access to
+// collection-specific key-values in a store is requested.  Failure to implement this
+// interface when declaring Key types can cause parts of the system to bypass held
+// collection write locks, potentially resulting in data races and inconsistent action
+// results.
+type CollectionedKey interface {
+	GetCollectionShortID() uint32
+}
```
islamaliev, line 30: suggestion:
```go
type CollectionedKey interface {
	Walkable
	GetCollectionShortID() uint32
}
```

islamaliev (approved):
LGTM! I left some minor suggestion and not so minor questions.

In general I agree with John that if feels a bit over-engineered and is going to be hard to maintain without understanding all in intricacies and potential edge cases.

But it's a great work overall.

internal/datastore/concurrent_txn_test.go
AndrewSisley: :grin: Nice spot, took me a while to find it even after you flagged it

- [x] Fix file name

internal/datastore/datastore.go
AndrewSisley: No, everything besides very niche operations (like truncate, and future potential atomic operations) needs a read lock.

The mutex naming convention fails here a little.  A mutex read lock doesn't really have anything to do with reading, and a write lock doesn't really have anything to do with writing - that is just how they are commonly used.  A RWMutex is just a mutex that allows many of one arbitrary kind (labeled 'read'), and one of another arbitrary kind (labeled 'write').

The ABLock I've proposed breaks this convention, it could be called RWLock or similar - it as as much of a read-write thing as sync.RWMutex is (i.e. not at all).

internal/db/lock/ab_lock.go
AndrewSisley: `LockA` and `LockB` are very much designed to be called concurrently, and I can't spot the deadlock there.  Did you mean serially? e.g. the below?
```
ab := abLock{}

ab.LockA()
ab.LockB()
```
The abLock will deadlock in such a case, and this is tested/documented with `TestLockSet_RLockAllAndLockSameTxn_Deadlocks`.  I am fairly confident that this is deadcode though, as I don't think `ReadLockAll` and `LockB` is ever actually called at the moment - it can only ever be done if scanning the entire datastore (across all collections).

internal/db/lock/ab_lock.go
AndrewSisley: >  is it 100% guaranteed that only one of them will be triggered?

Yes, `OnError` is only ever called if commit fails, in which case `OnCommit` is not executed.  It is a fairly safe assumption, both because it would be wierd to change that (but not impossible), and also because unlocking a unlocked mutex will panic, meaning it would be impossible for any of our datstore-touching integration tests to pass were the `Txn` implementation to change in such a way without also changing this.

internal/db/lock/lock.go
AndrewSisley: Thanks!

- [x] fix typo

internal/db/lock/lock_set.go
AndrewSisley: Cheers :)

- [x] Fix 'typo'

internal/db/lock/lock_set_test.go
AndrewSisley: Nice spot, thanks

- [x] Fix `dummyTxn) OnDiscard`

internal/db/lock/lock_set.go
AndrewSisley: >  is it guaranteed that it's going to be called from a single thread?

If I understood you correctly, and you refer to this and zero-many other public functions on this type being called concurrently:

No, absolutely not, it is highly concurrent and locking it down to a single active thread would defeat it's purpose.

> If no, there is a lot of racing here which should be either fixed or documented explaining why we tolerate it.

Please flag any lines you are unsure about, I've tried fairly hard to be very heavy handed on the documentation here, and would like to improve anything that remains confusing.

internal/db/lock/lock_set.go
AndrewSisley: Correct, I deliberately did not clean them up in this PR as for collection short ids (the immediate, v1 usecase) it would be a waste of effort and would probably have a negative impact on performance.

Would you like this documented on `l.locksByKey`?  And/or (please specify) a github issue? I started adding a note about this at somepoint, but deleted it, deciding that it was diluting the more immediate information that the already large documentation is try to communicate.

internal/db/lock/lock_set.go
AndrewSisley: Nice spot! I had a stare at this, and can only assume it must have been left over from an older implementation (was a lot of rewriting in this file).  Is a nice complication to remove :)

- [x] Remove pointless delete in `rUnlockAll` and `unlockAll`

internal/keys/datastore.go
AndrewSisley: I disagree with this fairly strongly and will be leaving it as is.  `Walkable` and `CollectionedKey` have very different use cases and there is not enough to gain by coupling them together (and making any references to `CollectionedKey` now dependent on `Walkable`).

internal/db/lock/ab_lock.go
islamaliev: I meant calling it like this:
```go
ab := abLock{}

go func() {
    ab.LockA()
}
go func() {
    ab.LockB()
}
```
But maybe it's going to fine as after txn is complete both waitings will be released.

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,367 @@
...
+		if !ok {
+			lock = &sync.RWMutex{}
+			// If the lock key does not exist yet, we must add it so
+			// that the read lock can be held if a write lock is attempted.
+			l.locksByKey[key] = lock
+		}
+		l.locksByKeyLock.Unlock()
+	}
+
+	lock.Lock()
```
islamaliev, line 169: question: all lock acquisitions block indefinitely, right?

A transaction that never closes (due to a bug) will permanently hold locks with no detection or recovery mechanism. 

Maybe it makes sense to add some kind of timeout.

internal/db/lock/ab_lock.go
AndrewSisley: Ah, yeah, that should be fine - the second `LockFoo` will not complete until the first has been unlocked, and that is  by design.

internal/db/lock/lock_set.go
AndrewSisley: > all lock acquisitions block indefinitely, right?
>
> A transaction that never closes (due to a bug) will permanently hold locks with no detection or recovery mechanism.

Correct

> Maybe it makes sense to add some kind of timeout.

I fairly strongly believe that such a timeout should exist for the transaction, and that having two different timeouts for transaction-locks and the transaction would be very confusing and lead to a lot of easily avoidable application bugs.

Such a transaction timeout has either already been implemented, or is in the process of being implemented for v1 for other reasons (the primary, but not the only, being that network failures when creating a user-managed transaction over http can cause the created txn to be unreachable, this can cause problems besides lock retention).

internal/db/lock/lock_set.go
islamaliev: hm... looks like I made some hasty conclusions. I assumed some mutexes are dependent on one another. After more careful analysis the code looks legit.

I think my concern was more about the atomicity of operations here. But it also doesn't seem to cause any problems.

internal/db/lock/lock_set.go
islamaliev: a github issue would be nice

AndrewSisley:
@jsimnz /All I've added some benchmark tests, the results running on my laptop are:
```
Benchmark_InitialReadLock-20       	 1282075	       944.1 ns/op
Benchmark_SecondaryReadLock-20     	10831672	       110.9 ns/op
Benchmark_InitialWriteLock-20      	 1000000	      1137 ns/op
Benchmark_SecondaryWriteLock-20    	11214116	       106.8 ns/op
Benchmark_ReadLockPromotion-20     	 1693591	       720.2 ns/op
Benchmark_ReadLockUnlock-20        	11588942	       103.3 ns/op
Benchmark_WriteLockUnlock-20       	 7615803	       158.3 ns/op
```
I am satified with this.

internal/db/lock/lock_set.go
AndrewSisley: https://github.com/sourcenetwork/defradb/issues/4316

internal/db/lock/lock_set.go
```diff
@@ -0,0 +1,367 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package lock
+
+import (
+	"sync"
+)
+
+// lockSet manages Defra-wide locks, scoped to an arbitrary Defra-element deliminated by the given `TKey`
```
fredcarle: typo: `delimited`.

internal/db/lock/txn_locks.go
```diff
@@ -0,0 +1,95 @@
...
+// txnLocks manages a set of mutexes scoped to each transaction. Each transaction will have zero or one
+// mutexes associated with it.
+//
+// Acquired locks are not automatically unlocked, it is the responsibility of consumers of this type
+// to unlock any locks that they have acquired.
+//
+// This type deletes the mutex associated with each transaction when the transaction is commited/discarded.
+//
+// Only public functions on this type should be called from outside of this type.  Calling private members
+// from anywhere else, internal or otherwise, removes any guarentees over its state's correctness.
```
fredcarle: typo: `guarantees`.

fredcarle (commented):
I'll admit that I'm not a fan of this implementation. Outside of the `lock` package it looks really simple. But the actual complexity within `lock` is ***VERY*** worrying to me. The implementation is clever and well documented. But in general I'm not a fan of clever code and needing that level of documentation predicts code smell. This will be fine as long an no one ever needs to maintain it — which clearly is wishful thinking. 

I've said it before but I don't think the locking should be done at the datastore level. The fact that its done there is asking for that kind of complex lock system. 

You already have two approvals so I'm not going to argue for any changes at this time.

AndrewSisley:
> I've said it before but I don't think the locking should be done at the datastore level. The fact that its done there is asking for that kind of complex lock system.

None of the complexity within this the lock package is because of *where* the locks are acquired.  The lock package manages the complexity of ensuring that the correct locks are held (or not) regardless as to when or where they are acquired.

So long as internal code paths can call multiple locations where locks need to be acquired (and unlocked), this complexity needs to exist in some form.   

I see two alternatives:

1. We could have a wrapping system similar to the old transaction system where a `client` interface  implementation contains all the locks (and needs to figure out upfront which ones might need to be held), with the rest of the code base is entirely lock free and banned from calling into the locking layer.  This is more error prone and less performant than the approach in the PR, as it requires upfront work to determine which locks might be needed, and might be simpler, but it is not complexity free.
2. We can scatter locking code throughout the codebase along with enough logic and params to hopefully ensure that we never read (or write) lock a thread already holding a write lock, and that we unlock exactly the right number of times.  In my opinion this is very high effort, very error prone, and possibly just as complex (if more more) as the solution in this PR.  Unlike (1) and this PR it will also not facilitate future features.

internal/db/lock/lock_set.go
AndrewSisley: Cheers :)

- [x] Typo

internal/db/lock/txn_locks.go
AndrewSisley: Thanks

- [x] Fix 'typo'

internal/db/lock/ab_lock.go
AndrewSisley: :grin: I was incorrect here, and had tunnel-visioned in on the `OnError` call - sorry, and thanks Islam - we often call discard after committing (but I don't think it is required), and so do need `once` or something similar :) 

- [x] Correctly handle `abLock` `Done` calls

PR #4278, opened 2025-12-17
feat: Create view from cid

### LINKED ISSUE #4162: Create view from Lens ID
From build week feedback: https://discord.com/channels/1374835078781468722/1435311498665918557
 
> `AddView` takes a path to a Lens transform, a Lens CID cannot be provided.
> 
> There may also be a bug where if you provide the same lens to two different views it errors, as the lens already exists (needs confirmation).
> 
> I think it is possible to work around this by creating the view without a transform, and then patching it using the lens CID, but there is a chance that that is blocked off in the short-term.

islamaliev:
## Relevant issue(s)

Resolves #4162

## Description

Replace optional transform parameter for `AddView` to be transform's CID.
Now that there `AddLens` method on db we can use it's result CID to refer to lenses.

### DIFF
--- a/internal/db/view.go
+++ b/internal/db/view.go
@@ -14,6 +14,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/ipfs/go-cid"
+
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/acp/identity"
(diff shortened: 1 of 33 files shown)


internal/db/view.go
```diff
@@ -333,3 +335,18 @@ func (db *DB) generateMaximalSelectFromCollection(
...
+
+// lensCIDExists checks if a lens with the given CID exists in the lens store.
+func (db *DB) lensCIDExists(ctx context.Context, cidStr string) (bool, error) {
+	lenses, err := db.getLensStore(ctx).List(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	for storedCID := range lenses {
+		if storedCID.String() == cidStr {
```
AndrewSisley: suggestion: If instead of converting the List result to string, you could convert `cidStr` to a CID at the start of the function and compare cids.  As well as it being more performant, we can catch invalid cids provided by the user before performing the potentially very expensive `List` call.

AndrewSisley (approved):
LGTM, thanks Islam!

internal/db/view.go
islamaliev: done

PR #4283, opened 2025-12-18
fix: Remove duplicate filter on update/upsert operations

### LINKED ISSUE #4279: Mutation filters execute both before and after update
This makes it impossible to know which documents (if any) have been updated if you are updating the field in the filter to something that no longer matches the filter.

e.g. the following test *passes*, and `"Shahzad"` is updated:
```
func TestQueryInlineNotNullStringArray_WithAnyFilter_Succeeds(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type Users {
						name: String
						preferredStrings: [String!]
					}
				`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Shahzad",
					"preferredStrings": ["first", "second"]
				}`,
			},
			testUtils.CreateDoc{
				Doc: `{
					"name": "Fred",
					"preferredStrings": ["", "second"]
				}`,
			},
			testUtils.Request{
				Request: `mutation {
					update_Users(filter: {preferredStrings: {_all: {_in: ["first", "second"]}}}, input: {preferredStrings: ["a","b"]}) {
						name
					}
				}`,
				Results: map[string]any{
					"update_Users": []map[string]any{},
				},
			},
		},
	}

}
```

jsimnz:
## Relevant issue(s)

Resolves #4279

## Description

Previously, filters would be re-run during update and upsert (on update conditions) nodes. Producing empty result sets for queries that *shouldve* returned values (see linked issue for examples).

Fixed by caching all the docs that are being updated into an internal `valuesNode` which replaces the underlying `scanNode` when documents are being "refetched" via the plan after updates have been applied.

## Tasks

## How has this been tested?

Added/Updated appropriate integration tests for both update and upsert cases.

- Ubuntu (WSL2)


internal/planner/update.go
```diff
@@ -37,6 +38,9 @@ type updateNode struct {
 	results planNode
 
 	execInfo updateExecInfo
+
+	origScanNode *scanNode
+	valuesNode   *valuesNode
```
AndrewSisley: todo: I think it is unreasonable to consume this much memory for an unordered update and that we should do better.  I also think that swapping nodes within an executing node is particularly confusing.

IMO we should just not add the filter to the original parent scan.  This will save us from reading every updated document into memory, and will make a lot more sense to anyone debugging.

AndrewSisley (changes requested):
I think we can implement a better fix, and do not think we should be rushing this.

internal/planner/update.go
fredcarle: The list of returned documents should match the list of documents updated. Just like when we create documents we return the created documents and when we delete documents we return the deleted documents. 

If we don't use the filter, we need some other way to keep track of the updated documents. Keeping the whole document in memory can indeed be expensive. When looking at this initially I thought it would make sense to just keep the list of updated documents' docID and at the response step we would iterate over that list and return the corresponding docs.

AndrewSisley: We can just yield the documents as we are iterating through them.  The is no reason to collect them.

We can have a quick call if you are unsure about what I mean.

fredcarle: If you mean `update -> yield, update -> yield, update -> yield...` as we iterate over the store for a mutation, I don't think it's a good solution. It opens the doors for mid mutation failures especially for mutations over http.

Technically the concerns we have with the potential for a lot of documents in memory is not new. We have the same issue with get and delete (even create to a certain extent). I don't think we should block this PR for this. This fix can be added and we can subsequently change the behaviour for all query/mutation types.

AndrewSisley: > It opens the doors for mid mutation failures especially for mutations over http.

What do you mean? What kind of failures?

fredcarle: If anything causes the transaction to fail or even commit by accident mid iteration, either we end up with a corrupted state or we have to start over. That would be quite unpleasant.

AndrewSisley: That still doesn't make sense to me.  If the transaction fails the update is discarded.  And if the planner starts committing transactions part way through iteration then we have plenty of problems besides this one.

fredcarle: > If the transaction fails the update is discarded

That is horribly unpleasant. If I have a million documents to update and it fails for a reason unrelated to document update at the 999 999th document, I now have to redo all that work. 

> And if the planner starts committing transactions part way through iteration then we have plenty of problems besides this one.

It doesn't have to be the planner. What if it's a user managed transaction? Doing this we are opening the door to avoidable user errors.

My opinion is that a mutation request should have completely executed the mutation before returning anything.

AndrewSisley: > That is horribly unpleasant. If I have a million documents to update and it fails for a reason unrelated to document update at the 999 999th document, I now have to redo all that work.

This is how transactions work, nothing in this PR or my suggestion changes anything to do with this.

> It doesn't have to be the planner. What if it's a user managed transaction? Doing this we are opening the door to avoidable user errors.

Again, nothing in this PR or my suggestion changes anything to do with this.

> My opinion is that a mutation request should have completely executed the mutation before returning anything.

It would be in either case.  My mind would be completely blown if you haven't misunderstood what I was saying.

getify:
just a point of clarification: is the *intended* behavior after this PR lands that the filters are applied to select which documents are subject to an update (aka BEFORE) or that the filters are applied AFTER the update against the already updated doc(s) to determine what is returned?

internal/planner/update.go
fredcarle: Andy and I had a call. I realized I was using the issue that he opened shortly before #4286 as context in this conversation which turns out is completely unrelated. Now that we are on the same page, I agree with him that we shouldn't store the docs in memory again.

AndrewSisley:
> just a point of clarification: is the _intended_ behavior after this PR lands that the filters are applied to select which documents are subject to an update (aka BEFORE) or that the filters are applied AFTER the update against the already updated doc(s) to determine what is returned?

Filter is applied before the update.  With only and all updated documents being returned to the user (regardless as to whether they now pass the filter).

internal/planner/update.go
jsimnz: > IMO we should just not add the filter to the original parent scan. 

This was going to be my original plan, but I don't think that will work. There isn't two sets of filters that are run, the full planned query is basically run twice. It's first run to yield documents (for their IDs), then they are individually updated, then the full query is rerun. I think we can be smarter in the long term but this is *currently* how it works for correctness sake. But we want the new values to go through the full plan.

All this change is doing is caching the yielded docs and using that instead of the original scan node, it's exceedingly simple while maintaining every aspect of the current semantics.

It can be made to be equivalent memory usage on the future, but there's lots of places we can reduce memory in the planner. And the walkAndReplace can be made cheaper. These are easy, non breaking, and can be done anytime.

There's no rush to this, happy to take the time to review as necessary, but the implementation itself isn't rushed, just simple.

jsimnz: @AndrewSisley @fredcarle 

Reworked this based on your feedback/concerns. It is now even simpler than the original, and doesn't need to do any plan graph rewiring/reinitializing (after the initial creation/expansion).

However, I still use the original `valuesNode` approach for the `upsertNode` implementation, but that is bounded to only a single document anyway, so no perf/mem concerns there. Even within the `upsertNode` implementation, its a little clunky, and requires the rewiring/replacing the underlying `scanNode` during execution which Andy noted as a *less than ideal* design, specifically citing debugging. This is a fair characterization, but im going to leave it for now. Its a plausible change to make in the future with zero impact.

Lmk youre thougts :)

internal/planner/update.go
```diff
@@ -176,11 +161,17 @@ func (p *Planner) UpdateDocs(parsed *mapper.Mutation) (planNode, error) {
...
+	selectTopNode, err := p.SelectTopNode(&parsed.Select)
 	if err != nil {
 		return nil, err
 	}
-	update.results = resultsNode
 
-	return update, nil
+	// IMPORTANT: This assignment is a noop, but is left in place for
+	// documentation reasons. The undelying results plan is set during
+	// plan expansion. Specifically, the `expandSelectTopNode` function.
```
AndrewSisley: praise: Thank you for this documentation!

internal/planner/upsert.go
```diff
@@ -82,6 +94,14 @@ func (n *upsertNode) Next() (bool, error) {
 
 			n.source.Prefixes(prefixes)
 		}
+
+		if updater {
+			err := n.p.walkAndReplacePlan(n.source, n.origScanNode, n.valuesNode)
```
AndrewSisley, line 107: suggestion: Please consider adding a line or two of documentation here detailing why this is done - it is unexpected, and is probably fairly hard to figure out why when looking from the context of this node (outside of the PR).

internal/planner/update.go
```diff
@@ -102,14 +61,40 @@ func (n *updateNode) Next() (bool, error) {
 	}
 
 	n.currentValue = n.results.Value()
+
+	docID, err := client.NewDocIDFromString(n.currentValue.GetID())
+	if err != nil {
+		return false, err
+	}
+	doc, err := n.collection.Get(n.p.ctx, docID, false)
```
AndrewSisley: thought: The double fetch is a bit clunky here, but I'm sure how costly it actually would be - badger might not even perform a second io call depending on how it caches, and even if badger doesn't the OS might be pulling it from cache most of the time anyway.

I'm guessing we'll probably rework this sometime post v1, but plenty good enough for now IMO - thanks John :)

AndrewSisley (approved):
LGTM! Thanks for reworking this, and sorry about the hassle.

internal/planner/multi.go
```diff
@@ -108,6 +110,9 @@ func (p *parallelNode) Init() error {
 	p.children = newChildren
 	p.childIndexes = newChildIndexes
 
+	for _, c := range p.children {
+		fmt.Printf("child: %T %+v\n", c, c)
+	}
```
fredcarle: todo: Remove print. Here and all the one bellow too.

internal/planner/multi.go
```diff
@@ -144,11 +149,11 @@ func (p *parallelNode) Next() (bool, error) {
 		var err error
 		// isMerge := false
 		switch n := plan.(type) {
-		case *scanNode, *typeIndexJoin:
-			// isMerge = true
-			next, err = p.nextMerge(i, n)
 		case *dagScanNode:
 			next, err = p.nextAppend(i, n)
+		default: // anything else is a merge
+			// isMerge = true
```
fredcarle: todo: Remove `isMerge` comment here and on line 150.

fredcarle (approved):
LGTM. Just need to clean up a couple things before merge.

internal/core/doc.go
```diff
@@ -310,3 +310,28 @@ func (m *DocumentMapping) TryToFindIndexFromRenderKey(key string) (int, bool) {
 	}
 	return -1, false
 }
+
+// DocFromDocument creates a core.Doc from a client.Document using the provided DocumentMapping.
+//
+// The mapping is used to determine the correct field indices for each field in the document.
+// Fields that are not found in the mapping are skipped.
+func DocFromClient(clientDoc *client.Document, mapping *DocumentMapping) (Doc, error) {
```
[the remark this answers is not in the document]
fredcarle: todo: Make doc match function name.

internal/planner/multi.go
```diff
@@ -144,11 +144,11 @@ func (p *parallelNode) Next() (bool, error) {
...
```
fredcarle: todo: Remove commented out code. Here and on line 145 too.

internal/planner/multi.go
```diff
@@ -243,13 +243,31 @@ func (n *selectNode) addSubPlan(fieldIndex int, newPlan planNode) error {
...
+
+		// we need to replace the original "source" with the appropriate
+		// multiScanNode type. However, not all "source" types are equal.
+		// For query ops the target source is a `*scanNode`, for mutations
+		// the target source is either a `*createNode`, `*updateNode`, or
+		// a `*upsertNode`
+		//
+		// This is necessary since the `*typeIndexJoin` join will read
+		// from this source multiple times per iteration, so we need
+		// to make sure that we're chaching the necessary state. Eg during
```
fredcarle: typo: `caching`

internal/planner/update.go
```diff
@@ -182,12 +171,91 @@ func (p *Planner) UpdateDocs(parsed *mapper.Mutation) (planNode, error) {
...
+			continue
+		}
+		switch f := field.(type) {
+		case *mapper.Select:
+			if f.SkipResolve {
+				// Filter-only relation: include in inner select, remove from outer
+				preUpdateSelect.Fields = append(preUpdateSelect.Fields, field)
+				selectFieldsToDelete = append(selectFieldsToDelete, i)
+			}
+			// Render relations (SkipResolve=false) stay only on the outer select
```
fredcarle: question: Is this comment supposed to be here?

internal/planner/update.go
```diff
@@ -182,12 +171,91 @@ func (p *Planner) UpdateDocs(parsed *mapper.Mutation) (planNode, error) {
...
+	// relation sub-select creates a type join on the outer select.
+	if top, ok := selectNode.(*selectTopNode); ok {
+		top.planNode = top.selectNode
+	}
+
+	update.results = selectNode
+
+	// The outer select only renders the post-update results. The inner select already
+	// handles pre-update filtering (by filter and/or docIDs), so we clear both on the
+	// outer select to prevent re-evaluation against potentially changed values.
```
fredcarle, line 219: praise: Thanks for the good documentation in this function.

internal/planner/update.go
```diff
@@ -182,12 +171,91 @@ func (p *Planner) UpdateDocs(parsed *mapper.Mutation) (planNode, error) {
...
+3) Select without mutation 		(filter: NO 	| select: YES 	| RelationMutation: NO) => 	BEFORE: NO  | AFTER: YES
+4) filter & select without mut	(filter: YES	| select: YES	| RelationMutation: NO) => 	BEFORE: YES  | AFTER: NO
+
+MUTATION
+1) No type join interaction 	(filter: NO 	| select: NO 	| RelationMutation: YES) => BEFORE: NO  | AFTER: NO
+2) filter with mutation 		(filter: YES 	| select: NO 	| RelationMutation: YES) => BEFORE: YES | AFTER: NO
+3) Select with mutation 		(filter: NO 	| select: YES 	| RelationMutation: YES) => BEFORE: NO  | AFTER: YES
+4) filter & select without mut	(filter: YES	| select: YES	| RelationMutation: YES) => BEFORE: YES | AFTER: YES
+
+*/
```
fredcarle: question: Do we want to keep this? If so, is there a better location for it?

internal/planner/update.go
```diff
@@ -182,12 +171,91 @@ func (p *Planner) UpdateDocs(parsed *mapper.Mutation) (planNode, error) {
...
+	if len(idx) == 0 {
+		return s
+	}
+
+	sort.Ints(idx)
+
+	out := s[:0]
+	j := 0
+
+	for i := 0; i < len(s); i++ {
```
fredcarle: suggestion: `for i := range s`

fredcarle (approved):
LGTM. Just a few comments to resolve before merge.

internal/planner/update.go
jsimnz: I have no pref, was being used more as a scratch pad working through the various cases. Ill just delete

getify:
🎉🥳🚀

PR #4296, opened 2025-12-22
fix: On doc-sync wait for docs from all peers

### LINKED ISSUE #4265: Doc sync won't always result in latest doc version
Currently what doc sync does it request peers via pubsub to send their version of a specific document, but in response handler we just wait for the first peer to respond, then we stop waiting and return. Other peers from the network might respond with more recent version of the doc, but it will be ignored.

One of the ways to go about is to set timeout that the node needs to wait (at all times) before it stop waiting.
This is how sync-branchable-collection does it ([currently in PR](https://github.com/sourcenetwork/defradb/pull/4188))

Another thing: the node receiving responses doesn't check if the response doesn't check if the data received is actually for the docID that is requested. So peers can send anything and the node will take it.

islamaliev:
## Relevant issue(s)

Resolves #4265

## Description

Make doc-sync tracks responses from active peers and wait for all of them instead of just aborting further sync processing after the first response.

### DIFF
--- a/http/client_p2p.go
+++ b/http/client_p2p.go
@@ -240,7 +240,19 @@ func (c *Client) SyncDocuments(
 		return err
 	}
 
-	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, methodURL.String(), bytes.NewBuffer(body))
+	// Use a separate context for HTTP request with extra buffer time.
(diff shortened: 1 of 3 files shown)


fredcarle (approved):
LGTM.

PR #4298, opened 2025-12-22
feat: LevelDB store

### LINKED ISSUE #4297: WASM compatible datastore
DefraDB has no persistent store options for WASM builds. We need to add a compatible store to support persistence in the browser.

Both NutsDB and LevelDB work in the browser but with some caveats:
- NutsDB iterator needs to be fixed to support transactions
- Both LevelDB and NutsDB only support a single transaction at a time

nasdf:
## Relevant issue(s)

Resolves #4297

## Description

This PR adds support for a LevelDB backed datastore. Since LevelDB only supports one transaction at a time, I had to remove all non-required transactions from background processes such as p2p, searchable encryption, and kms.

- [x] TODO add test matrix config for leveldb

## Tasks

## How has this been tested?

Integration tests.

- MacOS


internal/db/id/collection.go
```diff
@@ -14,13 +14,33 @@ import (
...
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/sequence"
 	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
+// NewShortCollectionID returns the local, shortened, internal, collection id, which is used
+// only in locations where using the full CID would be a waste of storage space.
+//
+// GetShortCollectionID should be preferred over this method because it utilizes the cache.
+func NewShortCollectionID(ctx context.Context, collectionID string, systemStore corekv.ReaderWriter) (uint32, error) {
```
AndrewSisley: suggestion: The name `NewFoo` really threw me here, as I was expecting a new id - this function is just returning an existing ID.  I suggest renaming it to something else.

AndrewSisley: suggestion: I suggest this function takes a multistore, not a `corekv.ReaderWriter` - it is important that this function reads from the correct namespace, and short ids are only held in the system store - there is no need for flexibility here, and passing in `corekv.ReaderWriter` obfuscates the code.

`GetShortCollectionID` can be made to work with this if you add a `Multistore` function to `datastore.Txn`, but given that that public interface contains a private type for some legacy reason, I suggest not bothering for the sake of 2-3 lines of code duplication - just revert `GetShortCollectionID` instead.

go.mod
```diff
@@ -39,18 +39,19 @@ require (
...
+	github.com/sourcenetwork/corekv/chunk v0.3.1
+	github.com/sourcenetwork/corekv/leveldb v0.3.1
+	github.com/sourcenetwork/corekv/memory v0.3.1
+	github.com/sourcenetwork/corekv/namespace v0.3.1
 	github.com/sourcenetwork/corelog v0.0.8
 	github.com/sourcenetwork/go-p2p v0.1.7
 	github.com/sourcenetwork/goji v0.0.8
 	github.com/sourcenetwork/graphql-go v0.7.4-0.20251208194504-a831d1b53d40
 	github.com/sourcenetwork/immutable v0.3.0
-	github.com/sourcenetwork/lens/host-go v0.9.4
```
AndrewSisley, line 56: thought: If in the future you bump versions in their own PRs it will keep each PR, and each change-log item more focused, and will make things easier for reviewers. 

Bumping Lens here seems very out of place, and I didn't realise you were doing so until I got down into the changes to `lens.go`, at which point I had to scroll up to check the `go.mod` file to check this was the reason for the code changes.

IIRC there are other unrelated changes that have been made between these two Lens versions too.

No need for changes to this PR, but this would have been better if broken out.

internal/db/db.go
```diff
@@ -402,6 +397,11 @@ func (db *DB) Close() {
 		db.p2p.SECoordinator().Close()
 	}
 
+	err := db.rootstore.Close()
```
AndrewSisley, line 405: question: Why has this moved?

internal/db/p2p/sync_col_versions.go
```diff
@@ -218,3 +220,17 @@ func (p *P2P) syncCollectionVersion(
...
+
+type wrappedTxn struct {
+	datastore.Txn
+	corekv.ReaderWriter
+}
+
+func (p *P2P) getLensP2P(ctx context.Context) p2p.P2P {
+	clientTxn := datastore.CtxMustGetClientTxn(ctx)
+	dsTxn := datastore.MustGetFromClientTxn(clientTxn)
+	return p.lens.P2P.Value().WithTxn(&wrappedTxn{
```
AndrewSisley, line 232: todo: Please document why we have to do this, Lens using the older wrapper pattern instead of taking from the context is not obvious here, especially given that many readers of this code may not be super on top of the Lens repo state.

Consider creating an issue in the Lens repo noting this pattern divergence, and linking to it here.

internal/se/coordinator_protocol.go
```diff
@@ -77,7 +68,9 @@ func (coordinator *Coordinator) processPushSEArtifactsRequest(
 		}
 	}
 
-	if err := storeArtifacts(ctx, txn.Datastore(), artifacts); err != nil {
+	ds := coordinator.db.Multistore().Datastore()
+	ss := coordinator.db.Multistore().Systemstore()
+	if err := storeArtifacts(ctx, ds, ss, artifacts); err != nil {
```
AndrewSisley: todo: Please document very carefully why we do not need a transaction here - by not using a transaction you are permitting the partial writing of SE artifacts within a single SE request.

internal/se/coordinator_protocol.go
```diff
@@ -134,18 +127,11 @@ func (coordinator *Coordinator) querySEArtifactsFromDatastore(
...
-
 	queries := make([]fieldQuery, len(req.Queries))
 	for i, q := range req.Queries {
 		queries[i] = fieldQuery(q)
 	}
-
-	return fetchDocIDs(ctx, txn.Datastore(), req.CollectionID, queries)
+	ds := coordinator.db.Multistore().Datastore()
+	ss := coordinator.db.Multistore().Systemstore()
+	return fetchDocIDs(ctx, ds, ss, req.CollectionID, queries)
```
AndrewSisley: todo: I think this one needs to be discussed with the team - by not running the queries within a transaction, you are permitting each query to execute upon a different database state, and the results within each query is dependent on the store `Iterator` implementation in order to prevent the store-state from mutating within the query.

It there is already a gap here, in that the transaction is only scoped to the fetching of IDs, and so the doc can still mutate before the returned DocID can be acted upon - and there is no way for the user to manage the transactions themselves if they wanted to avoid this, but this new change increases the scope of the problem.

todo: Even if the code remains as-is, please document why we are okay with the lack of transactional protection here.

internal/se/coordinator_retry.go
```diff
@@ -122,23 +113,11 @@ func (coordinator *Coordinator) processSERetries(ctx context.Context) {
...
-			clientTxn, err := coordinator.db.NewTxn(false)
-			if err != nil {
-				log.ErrorContextE(ctx, "Failed to create transaction on retry", err)
-				return
-			}
-			defer clientTxn.Discard()
-			txn := datastore.MustGetFromClientTxn(clientTxn)
-
-			if err := txn.Peerstore().Set(ctx, iter.Key(), b); err != nil {
+			if err := coordinator.db.Multistore().Peerstore().Set(ctx, iter.Key(), b); err != nil {
```
AndrewSisley, line 116: todo: By removing the transaction I think you have introduced a race condition here - the checks on line `101` might be out of date by the time you come to overwrite them here, depending on how this function is scheduled.

If we are fine with this right now, please document why we are okay with it, and if it is dependent on how this function is scheduled and its concurrency, please add a warning to the function description noting the details to give us better odds of avoiding a breakage if we change the scheduling in the future.

EDIT: It looks like the same issue may now be present in the other functions below this line in this file - please address those in a similar way.

tests/integration/utils.go
```diff
@@ -922,6 +925,7 @@ func refreshTokens(
 // result-index will be nil.
 func refreshCollections(
 	s *state.State,
+	transactionID immutable.Option[int],
```
AndrewSisley, line 906: todo: Please document this parameter, and why me may or may not want to provide a transaction here. 

todo: It looks like this might be dead code, and it always has a `None` value - if so, please remove the parameter.

AndrewSisley (dismissed):
I think some more documentation, and perhaps a discussion around the changes to SE is required before merge.

internal/db/db.go
nasdf: This is the correct order and reduces the number of errors when closing.

internal/db/p2p/sync_col_versions.go
nasdf: It's the same pattern the db package uses for lens. I disagree that it needs documentation.

internal/se/coordinator_protocol.go
nasdf: Partial writes from networking code should be expected and handled by the consuming code or implementation.

internal/se/coordinator_protocol.go
nasdf: This shouldn't be a problem for a properly designed network system. Repeated or dropped actions should be handled and expected.

internal/se/coordinator_retry.go
nasdf: The `processSERetries` function will never run in parallel with the current design so this issue is irrelevant.

tests/integration/utils.go
nasdf: It is used. Please look again.

internal/se/coordinator_protocol.go
AndrewSisley: This isn't really networking code though, I'm talking about the result of a single, completely received, request being half written to disk in a way that allows for the half-result to be returned as a fully-successful appearing result.

internal/se/coordinator_retry.go
AndrewSisley: Okay nice, please document this then, as these lines are quite a long way from where that is visible.

internal/se/coordinator_protocol.go
AndrewSisley: This comment is not about repeated or dropped actions, it is about returning multiple conflicting database states from the same query.

internal/se/coordinator_protocol.go
nasdf: It's a network reliant system and should be designed to handle partial writes.

AndrewSisley: You could make the same argument about a GQL request made over http, but if it only returns half the documents along with a success code, people are going to be upset

AndrewSisley: I suggest discussing this on discord, and making sure that at least @islamaliev is tagged in that conversation

internal/se/coordinator_protocol.go
nasdf: The problem you are referencing would lead to duplicated or dropped actions due to differing database states. This is something that should be handled by the SE system and not require transactions.

AndrewSisley: > This is something that should be handled by the SE system and not require transactions.

Is it handled though? Or is this PR introducing a regression?

internal/se/coordinator_protocol.go
nasdf: > You could make the same argument about a GQL request made over http, but if it only returns half the documents along with a success code, people are going to be upset

That example doesn't apply here. If any of the writes fail then a response is not published.

internal/se/coordinator_protocol.go
nasdf: If not I would consider that a bug and potentially out of scope. If it is a regression then I'd expect a test to catch it. Without any solid evidence against my argument I don't see why this should be a todo.

AndrewSisley: IMO, anyone making any change should be mindful that the system was coded by imperfect beings, and that it might not be as well tested as they would like, and that it might not behave in the way that they believe it should.

The system may have been reliant on the transaction in order to behave correctly for users.  If you wish to remove the transaction I do believe that it is your responsibility to make sure that the system continues to correctly function without the transaction.

nasdf: If there is no documentation, tests, or evidence that something is wrong here, then I don't think you should block with a todo.

AndrewSisley: I recommend raising it with the team.  If you cannot tell me that a change you are proposing is correct from a user's perspective, I am very unlikely to approve it.

nasdf: I am telling you the changes I made are correct to the best of my knowledge. If you cannot point to an actual issue then your todo will be dismissed.

AndrewSisley: > If you cannot point to an actual issue then your todo will be dismissed.

That's totally okay, and that is what the `Dismiss` button is there for.  Just do not hide my disapproval from any other reviewer, and make sure that they are on board with the risk.

I do think it would be significantly more reasonable to ask your colleagues first though.  Islam might even know for sure that this is safe - could be a 5sec confirmation.

internal/db/merge.go
```diff
@@ -439,8 +446,18 @@ func (mp *mergeProcessor) initCRDTForType(ctx context.Context, crdtUnion crdt.CR
...
+			return nil, err
+		}
+		_, exists := mp.docIDs[docID]
+		if !exists {
+			doc, err := mp.col.Get(ctx, docID, false)
+			if err != nil && !errors.Is(err, client.ErrDocumentNotFoundOrNotAuthorized) {
+				return nil, err
+			}
+			mp.docIDs[docID] = doc
+		}
```
islamaliev, line 453: suggestion: this block is identical to `case crdtUnion.IsComposite():`. Would be nice to merge them or extract info a func

tests/integration/mutation/create/field_kinds/one_to_one_to_one/utils.go
```diff
@@ -21,6 +21,7 @@ func execute(t *testing.T, test testUtils.TestCase) {
 	testUtils.ExecuteTestCase(
 		t,
 		testUtils.TestCase{
+			SupportedDatabaseTypes: test.SupportedDatabaseTypes,
```
islamaliev, line 24: note: I don't like that sometime we re-create `TestCase` and copy stuff into it. I had some problems with it when developing secondary-index multiplier. Better approach would be just to change only what needs to be changed:
```go
test.Actions = append(
    []any{
        &action.AddSchema{
            Schema: `
                type Book {
                    name: String
                    rating: Float
                    author: Author
                    publisher: Publisher @primary
                }

                type Author {
                    name: String
                    age: Int
                    verified: Boolean
                    wrote: Book @primary
                }

                type Publisher {
                    name: String
                    address: String
                    published: Book
                }
            `,
        },
    },
    test.Actions...,
)
```

tests/integration/mutation/mix/with_txn_test.go
```diff
@@ -200,6 +207,12 @@ func TestMutationWithTxnDoesUpdateUserGivenSameTransactions(t *testing.T) {
 
 func TestMutationWithTxnDoesNotUpdateUserGivenDifferentTransactions(t *testing.T) {
 	test := testUtils.TestCase{
+		// LevelDB does not support concurrent transactions
+		SupportedDatabaseTypes: immutable.Some([]state.DatabaseType{
```
islamaliev, line 213: note: I'd say it's a rare case when we have to specify supported type. More often we need to specify unsupported type in terms of supported, which is very error-prone. When we add a new type there is a high chance that it will be skipped. We should have either both supported and unsupported or just unsupported.

internal/se/coordinator_protocol.go
islamaliev: I actually don't understand why we need to use them separately. I understand that LevelDB operates with only 1 transaction at a time. But this function is only executed once in isolation (on remote node). Why can't we just use 1 transaction for atomicity?

nasdf: Most of the transactions here have a potential to deadlock, but on top of that will degrade performance for the user. 

When a transaction is held by the P2P, SE, or KMS systems the user has to wait for them to commit or discard before they can perform any actions.

In my opinion none of these systems need to rely on transactions.

internal/se/coordinator_protocol.go
islamaliev: > If there is no documentation, tests, or evidence that something is wrong here, then I don't think you should block with a todo.

With the current state of testing framework it's almost impossible to cover all cases in multi-node environment. I raised this topic few time on stand up.

islamaliev: anyway for the current scenario the changes shouldn't cause any problem.

internal/db/view.go
```diff
@@ -351,7 +351,7 @@ func (db *DB) lensCIDExists(ctx context.Context, cidStr string) (bool, error) {
 	}
 
 	for storedCID := range lenses {
-		if storedCID.Equals(targetCID) {
+		if storedCID == targetCID.String() {
```
fredcarle: suggestion: Might as well use `cidStr` directly. Still we should keep the `cid.Decode` block to check that we have a valid CID.

tests/integration/collection_version/migrations/with_txn_test.go
```diff
@@ -19,10 +19,17 @@ import (
 	"github.com/sourcenetwork/defradb/client"
 	testUtils "github.com/sourcenetwork/defradb/tests/integration"
 	"github.com/sourcenetwork/defradb/tests/lenses"
+	"github.com/sourcenetwork/defradb/tests/state"
 )
 
 func TestSchemaMigrationGetMigrationsWithTxn(t *testing.T) {
 	test := testUtils.TestCase{
+		// LevelDB does not support concurrent transactions
```
fredcarle: question: I'm struggling to see how concurrent transactions are relevant both here and in other tests where you've added the same supported database restrictions. Can you shed some light for me please?

internal/db/id/collection.go
fredcarle: I'm ok with this taking a systemStore `ReaderWriter`. It's an internal function and the parameter name makes it clear what this needs to be. Using something else will clearly not work and would never be able to reach production.

fredcarle: I agree with trying to rename this.

internal/db/collection.go
```diff
@@ -211,27 +211,20 @@ func (c *collection) GetAllDocIDs(
 	if err := c.db.checkNodeAccess(ctx, acpTypes.NodeDocumentReadPerm); err != nil {
 		return nil, err
 	}
-
-	ctx, _, err := ensureContextTxn(ctx, c.db, true)
-	if err != nil {
-		return nil, err
-	}
 	return c.getAllDocIDsChan(ctx)
```
fredcarle, line 214: thought: It feels weird to have this call not use a transaction.

internal/se/se.go
```diff
@@ -30,9 +31,14 @@ import (
 )
 
 // storeArtifacts stores SE artifacts directly in the datastore.
-func storeArtifacts(ctx context.Context, ds datastore.Keyedstore, artifacts []secore.Artifact) error {
+func storeArtifacts(
+	ctx context.Context,
+	ds datastore.Keyedstore,
+	ss corekv.ReaderWriter,
```
fredcarle: thought: Might as well just receive a `datastore.Multistore`.

internal/db/collection.go
AndrewSisley: Why?

internal/se/se.go
```diff
@@ -57,12 +63,13 @@ func storeArtifacts(ctx context.Context, ds datastore.Keyedstore, artifacts []se
 func fetchDocIDs(
 	ctx context.Context,
 	ds datastore.Keyedstore,
+	ss corekv.ReaderWriter,
```
fredcarle: thought: Same as above.

fredcarle (approved):
Approving as with some somments to resolve before merge.

internal/db/collection.go
fredcarle: Because depending on the implementation of the store iterator, this breaks isolation.

AndrewSisley: I didn't think we cared about that for this function?

If we do, we could use a collection lock.  We can optimise it beyond just a write lock too here if we need (more effort), as here, concurrent reads are fine.

tests/integration/mutation/create/field_kinds/one_to_one_to_one/utils.go
AndrewSisley: I do see this (original code, not your comment) as a legacy pattern that everyone is moving away from.  I don't remember seeing anyone trying to introduce a new, similar func for a while now.

tests/integration/collection_version/migrations/with_txn_test.go
nasdf: Every public function on the database will implicitly create a transaction. Because the test framework creates and holds a transaction for the duration of the test, it will cause all other database calls to hang indefinitely.

AndrewSisley: Not to say that all such test changes are incorrect, but this particular test only appears to be using a single transaction, and so should work correctly with LevelDB.

nasdf: This specific test might be fixed with the changes I made to `RefreshCollections`. I will double check it before merging.

nasdf: I was able to revert this one.

internal/db/view.go
nasdf: Done

tests/integration/mutation/mix/with_txn_test.go
nasdf: I opened an issue for this and left TODOs https://github.com/sourcenetwork/defradb/issues/4442

internal/db/merge.go
nasdf: Done

internal/db/id/collection.go
nasdf: Renamed to `GetUncachedShortCollectionID`

internal/se/se.go
nasdf: Done

internal/se/se.go
nasdf: Done

PR #4300, opened 2025-12-23
fix: Patching of IsActive property

### LINKED ISSUE #4134: Patching IsActive to true when collection has single version errors
With `collections cannot be deleted. CollectionID: bafy...`.

Collections synced from other nodes are inactive by default, it should be possible to patch them to be so.  `SetActiveCollectionVersion` works fine, and this is what the tests used :(

islamaliev:
## Relevant issue(s)

Resolves #4134

## Description

Fix patching to allow changing of `IsActive` field value.

### DIFF
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -152,6 +152,8 @@ func (db *DB) patchCollection(
 	for _, col := range existingCols {
 		if col.IsActive {
 			existingColsByName[col.Name] = col
+		} else if _, exists := existingColsByName[col.Name]; !exists {
+			existingColsByName[col.Name] = col
 		}
 		existingColsByID[col.VersionID] = col
 	}
--- a/tests/integration/net/sync/collection_version/patch_test.go
+++ b/tests/integration/net/sync/collection_version/patch_test.go
@@ -108,6 +108,61 @@ func TestSyncColVersion_WithPatchVersionOfUnknownCollection(t *testing.T) {
 	testUtils.ExecuteTestCase(t, test)
 }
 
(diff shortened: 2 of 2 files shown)


internal/db/collection_define.go
```diff
@@ -152,6 +152,8 @@ func (db *DB) patchCollection(
 	for _, col := range existingCols {
 		if col.IsActive {
 			existingColsByName[col.Name] = col
+		} else if _, exists := existingColsByName[col.Name]; !exists {
```
AndrewSisley, line 155: todo: This only works if the user only intends to update the collection version with that name that has a version id that is lexicographically smaller than all other versions of that name.

Please change the production code, add test(s) that demonstrate this, and also check-for/add a test that shows it is possible to patch `IsActive` by specifying the `VersionID`, as well as the collection name (e.g. `[{"op": "replace", "path": "/baegsdg/IsActive", "value": true}]`

It also looks like it might cause issues elsewhere (if not now, then later), as it changes what `existingColsByName` currently is, to something else, and that something else is fairly complex.

AndrewSisley (changes requested):
I think you have only partially fixed this, it needs a little more work before merge.

internal/db/collection_define.go
AndrewSisley: I ran some tests locally - the issue has either been fixed, or there was a bug in my application code when I raised this (this was a build week issue).

I don't think we should be patching `IsActive` by Name - that seems very error prone.  I suggest closing both the PR and the issue.

islamaliev:
Closing because the issue can not be reproduced anymore

PR #4309, opened 2026-01-06
fix: prevent txn commit/discard during active operations

### LINKED ISSUE #4288: It appears possible to commit/discard manually managed transactions during action execution
Untested (also fix), but it appear possible to to commit and/or discard manually managed transactions during a database operation's execution - leaving the database in a corrupted state (e.g. collections and/or documents partially written).

This can happen if a user:

1. Creates manual transaction (concurrent or otherwise).
2. Submits an operation to Defra using transaction (1), but does not wait for it's completion.
3. Concurrent to the execution of (2), discards or commits transaction (1). 

This is possible with all clients.

SharanRP:
## Relevant issue(s)

Resolves #4288

## Description

This PR prevents users from committing or discarding manually-managed transactions while database operations are still in progress, which could lead to database corruption with partially-written data.

### The Problem

When users create manual transactions via `NewTxn()` and pass them to database operations, they could concurrently call `Commit()` or `Discard()` on the transaction before the operation completes. This could leave the database in a corrupted state (e.g., collections and/or documents partially written).

### The Solution

Added an atomic operation counter (`activeOps`) to track active operations on a transaction:

1. **`StartOp()`** - Called when an operation begins using the transaction (in `ensureContextTxn()`)
2. **`EndOp()`** - Called when an operation completes (in explicit txn's `Commit()`/`Discard()`)
3. **`Commit()`** - Returns `ErrTxnHasActiveOps` if `activeOps > 0`
4. **`Discard()`** - Logs error and skips discarding if `activeOps > 0`

### Changes

| File | Change |
|------|--------|
| `internal/datastore/errors.go` | Added `ErrTxnHasActiveOps` error |
| `internal/datastore/txn.go` | Added `activeOps` counter, `StartOp()`/`EndOp()` methods, and checks in `Commit()`/`Discard()` |
| `internal/db/txn.go` | Added `StartOp()` call in `ensureContextTxn()` and `EndOp()` calls in explicit txn wrappers |
| `internal/datastore/txn_ops_test.go` | Unit tests for operation tracking |
| `internal/db/txn_test.go` | Integration tests for transaction operation tracking |

## Tasks

### Limitations

- The `Discard()` method cannot return an error (interface constraint), so when called during active operations it logs an error and returns without discarding. This prevents corruption but may be unexpected to users.
- The atomic counter adds minimal overhead (increment/decrement) but is necessary for thread-safety.

## How has this been tested?

- Added unit tests in `internal/datastore/txn_ops_test.go` for the low-level `StartOp()`/`EndOp()` mechanism
- Added integration tests in `internal/db/txn_test.go` for full transaction operation tracking including concurrent scenarios
- All existing tests pass

### Specify the platform(s) on which this was tested

- Pop!_OS (Linux)

### DIFF
--- a/internal/datastore/txn.go
+++ b/internal/datastore/txn.go
@@ -12,13 +12,17 @@ package datastore
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corelog"
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/client"
 )
+
+var log = corelog.NewLogger("datastore")
 
 // Txn is a common interface to the BasicTxn struct.
 type Txn interface {
@@ -76,6 +80,14 @@ type Txn interface {
 
 	// OnDiscardAsync registers a function to be called asynchronously when the transaction is discarded.
 	OnDiscardAsync(fn func())
+
+	// StartOp marks the beginning of an operation on this transaction.
+	// Must be paired with EndOp. Used to prevent concurrent commit/discard while operations are in progress.
+	StartOp()
+
+	// EndOp marks the end of an operation on this transaction.
+	// Must be paired with StartOp.
+	EndOp()
 }
 
 type BasicTxn struct {
@@ -84,6 +96,10 @@ type BasicTxn struct {
 	txn corekv.Txn
 	id  uint64
 	ts  time.Time // timestamp
+
+	// activeOps tracks the number of operations currently in progress on this transaction.
+	// Used to prevent commit/discard while operations are executing.
+	activeOps atomic.Int64
 
 	successFns []func()
 	errorFns   []func()
@@ -117,6 +133,10 @@ func (t *BasicTxn) StartTS() time.Time {
 }
 
 func (t *BasicTxn) Commit() error {
+	if t.activeOps.Load() > 0 {
+		return ErrTxnHasActiveOps
+	}
+
 	var fns []func()
 	var asyncFns []func()
 
@@ -139,6 +159,11 @@ func (t *BasicTxn) Commit() error {
 }
 
 func (t *BasicTxn) Discard() {
+	if t.activeOps.Load() > 0 {
+		log.Error("Attempted to discard transaction while operations are in progress; ignoring discard")
+		return
+	}
+
 	t.txn.Discard()
 
 	for _, fn := range t.discardAsyncFns {
@@ -171,6 +196,14 @@ func (t *BasicTxn) OnErrorAsync(fn func()) {
 
 func (t *BasicTxn) OnDiscardAsync(fn func()) {
 	t.discardAsyncFns = append(t.discardAsyncFns, fn)
+}
+
+func (t *BasicTxn) StartOp() {
+	t.activeOps.Add(1)
+}
+
+func (t *BasicTxn) EndOp() {
+	t.activeOps.Add(-1)
 }
 
 type txnKey struct{}
--- a/internal/db/txn.go
+++ b/internal/db/txn.go
@@ -46,6 +46,8 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 			}
 			// If the txn has already been set on the context but it hasn't already been set as explicit,
 			// we create a copy of the txn and mark it as an explicit txn.
+			// Start an operation to prevent concurrent commit/discard.
+			txn.BasicTxn.StartOp()
 			explicitTxn := &Txn{
 				txn.BasicTxn,
 				txn.db,
@@ -59,6 +61,8 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 			//
 			// WARNING: This scenario creates a transaction where `*DB` is nil. Calling any method that requires this
 			// will result in a panic.
+			// Start an operation to prevent concurrent commit/discard.
+			txn.StartOp()
 			explicitTxn := &Txn{
 				txn,
 				nil,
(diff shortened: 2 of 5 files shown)


AndrewSisley (changes requested):
Hey Sharan, sorry about the bother, but I don't think this fully resolves the issue.  There appears to be nothing preventing the user from calling `EndOp` themselves, then committing/discarding mid operation and corrupting the database.

You've made it a little harder to do, but I think it probably needs a little more work.  Feel very free to join our discord to discuss this with the team :)

internal/db/txn.go
```diff
@@ -46,6 +46,8 @@ func ensureContextTxn(ctx context.Context, db transactionDB, readOnly bool) (con
 			}
 			// If the txn has already been set on the context but it hasn't already been set as explicit,
 			// we create a copy of the txn and mark it as an explicit txn.
+			// Start an operation to prevent concurrent commit/discard.
+			txn.BasicTxn.StartOp()
```
AndrewSisley, line 50: thought: The fix's reliance on `ensureContextTxn` makes me a bit nervous, as I'm not sure transactions should be required to pass through there, and technically we shouldn't be reliant on this.  

I also think there has been some talk from other team members around changing the way transactions are handled internally, and this function might not survive those changes.  

However your usage of it makes for a very easy fix, with a low surface area, making it's removal/replacement at a later date pretty cheap.

I am happy for this to be merged like this, but I wanted to flag this in case other team members have differing opinions here, or in case it helps inspire a quick alternative (none come to mind for me).

internal/db/txn.go
```diff
@@ -108,6 +114,8 @@ func (txn *Txn) Discard() {
 		// If the transaction has been explicitly defined, `Discard` should
 		// only be executed by the transaction creator. As such, a call to
 		// `Discard` on an explicit transaction should result in a no-op.
+		// End the operation to allow the creator to commit/discard.
+		txn.BasicTxn.EndOp()
```
AndrewSisley: todo: It is permitted to call `Discard` multiple times, but the current result of `txn.BasicTxn.EndOp()` will result in a negative `activeOps` count.  This means that if a user discards a transaction multiple times before re-using it, `activeOps` will have an incorrect value.

I cannot remember if all (or any) of our supported `corekv` stores support transaction reuse after calling `Discard`, but even if they don't, this scenario absolutely requires tests.

e.g.:
```
txn := newTxn()

... do stuff (not not - we also should have a test where nothing is done before the intial discard)

txn.Discard()
txn.Discard()

do exactly one operation

txn.Commit()
```

internal/datastore/txn.go
```diff
@@ -173,6 +198,14 @@ func (t *BasicTxn) OnDiscardAsync(fn func()) {
 	t.discardAsyncFns = append(t.discardAsyncFns, fn)
 }
 
+func (t *BasicTxn) StartOp() {
+	t.activeOps.Add(1)
```
AndrewSisley, line 202: todo: I do not think this will work in it's current form if multiple operations are performed by the user before Commit/Discard.  These scenarios must be tested before this PR can be merged.

For example:
```
txn := newTxn()

txn.DoSomething()
txn.DoSomething()

txn.Commit()

assert changes have been commited
```

internal/db/txn_test.go
```diff
@@ -0,0 +1,143 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package db
```
AndrewSisley, line 11: todo: These unit tests are nice to have to supplement short-term testing, however I think they are insufficient, particularly in the the long run, for such a sensitive part of the codebase.

The vast majority of the codebase is tested by the `tests/integration` directory, it is probably quite complex/daunting to get started there as a newcomer at the moment, but it is significantly more powerful than unit tests, and will allow for much easier testing of more complex scenarios.

If you are willing to spend the time to get to grips with it, that would be fantastic, and I encourage you to make an unguided attempt on your own before asking for additional help/examples (providing feedback to us on this experience would be very valuable) - but if needed do message us on discord and I'll try and give you a hand.

If that is too much for you right now that is not a problem, you are free to work on the rest of the PR as much as you like and if the team is happy enough with the rest of the solution, we'll add tests before merge (this is a fairly high priority issue, we will not always be able to offer this).

AndrewSisley (changes requested):
Hi Sharan!

Thank you very much for all your efforts on this, it is looking good and is easy to read (thank you in particular for the documentation).

Given the sensitivity of this part of the codebase, I feel like this needs more test coverage before I am comfortable with this being merged.  The existing unit tests are very nice to have, but in my opinion we need some fairly extensive integration tests for this.

No worries if you are unwilling to spend that level of effort here, a team member will pick this up later if so.

If you have not spotted the note in our `CONTRIBUTING.md`, we use subset of the https://conventionalcomments.org/ for PR feedback.  At the moment I have added several `todo` comments, and a `thought`, the `todo`s must be resolved before merge, the `thought` does not - please argue back against anything you disagree with - 'resolve' in this context often means convincing the reviewer that they are wrong in some way :) 

And please continue to discuss anything you are unsure about either here or on our discord - I'll try and make sure any conclusions relevant to the PR arrived at on discord are summarised here for tracking before merge. 

Have a great day,
Andy

SharanRP:
Hey! Thanks for the detailed feedback

I added an `opEnded` flag to the `Txn` struct to track whether `EndOp()` has already been invoked. This ensures that multiple calls to `Discard()` or `Commit()` do not decrement `activeOps` more than once, preventing scenarios where the value could go negative.

On the testing side, I’ve added unit tests that cover multiple discard calls, multiple sequential operations before a commit, discard-before-operation cases, concurrent commit scenarios, and immediate discard without any operations.

Next, I’ll be digging into the integration tests under `tests/integration` to add broader coverage. I’ll reach out on Discord if I need any guidance on the testing framework.

AndrewSisley:
Hey Sharan, I'm going to close this PR for now as it has gotten stale and the linked issue has been resolved via other means.

PR #4313, opened 2026-01-07
feat: Add optional health check to setup wizard

### LINKED ISSUE #4195: Setup Wizard: Ask to do a health-check query against defra
It also might be helpful to offer an option to send a *general introspection* query to the newly started node, such as this one:

```graphql
query Introspection {
  __schema { queryType { name } types { name } }
}
```

That seems like a useful test of "is it running properly?"

ChrisBQu:
## Relevant issue(s)

Resolves #4195 

## Description

This PR adds an optional health check as the final step to the setup wizard. The user will be asked if they wish to perform this check, and if they answer yes, the wizard will spin up a Defra instance, using the `defradb start` command, and will then ping the `health-check` endpoint. Success or failure will be communicated back to the user.

Because the health check callback function utilizes the actual binary, it doesn't play well with integration or unit tests. As such, to test it, a bash script has been written and included, which is invoked as a new Github workflow action. What this does, is build the Defra binary, then run the wizard. It then simulates all of the user's input to proceed through the steps of the wizard, and opts into performing the health check. Therefore, while this can be seen as a holistic test of the wizard (which is helpful/good), it also *explicitly* tests the health-check functionality.

Finally, several minor stylistic changes were made here. These were to address styling-related visual bugs or anomalies that manifested during the development of the health check. 

## Tasks

## How has this been tested?

- WSL


cli/wizard/wizard.go
```diff
@@ -220,15 +220,39 @@ func Main() {
...
+	stepQueryPerformingHealthCheck := initialModelMultipleChoice(
+		stepQueryPerformingHealthCheckID,
+		"Do you want to test that DefraDB is configured correctly by "+
+			"performing a health check?",
+		[]string{"Yes", "No"},
+	)
+
+	stepWillRunHealthcheck := initialModelText(
+		stepWillRunHealthcheckID,
+		"A health check will be performed. This may take up to 10 seconds to complete.",
```
AndrewSisley: suggestion: The `10` seconds here is somewhat tucked away in another file as a magic number - it might be worth linking these two by sharing a const in order to prevent them from becoming out of sync in the future, and to make it slightly easier for maintainers to understand where the number comes from.

cli/wizard/callbacks.go
```diff
@@ -330,3 +335,69 @@ func callback_GenerateSearchableEncryptionKey(_ step, ctx *WizardContext) error
...
+func callback_PerformHealthcheck(_ step, ctx *WizardContext) error {
+	printToTerminal(TerminalClearANSICode)
+	printToTerminal("Performing health check...")
+	defer printToTerminal(TerminalClearANSICode)
+
+	// Entire health check must finish within a finite amount of time
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Resolve the binary path to the one that launched the wizard
```
AndrewSisley, line 349: suggestion: Calling the binary like this is unusual, and I think it would be worth expanding this comment to explicitly detail why we want to do this this way instead of a more direct route.

AndrewSisley (approved):
LGTM, just a couple of (optional) suggestions from me before merge - thanks Chris :)

cli/wizard/wizard.go
ChrisBQu: I like this. I will be making this change.

cli/wizard/callbacks.go
ChrisBQu: This is a good idea. I have made this change.

.github/workflows/setup-wizard-health-check-test.yml
```diff
@@ -0,0 +1,35 @@
...
+      - reopened
+      - synchronize
+    branches:
+      - develop
+
+jobs:
+  test-wizard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
shahzadlone, line 27: todo: Why not `v4`? soon we will be changing to `v6` 

```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/setup-wizard-health-check-test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  pull_request:
+    types:
+      - edited
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - develop
```
shahzadlone, line 21: question: Why this instead of our normal triggers like:


```suggestion
on:
  pull_request:
    branches:
      - master
      - develop

  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
    branches:
      - master
      - develop

```

shahzadlone (commented):
@ChrisBQu Was just about to make this required but noticed somethings thought I would leave a comment on this merged PR

.github/workflows/setup-wizard-health-check-test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  test-wizard:
```
shahzadlone, line 24: todo: Without this the required action is hard to trace the name of the job (will be inconsistent) 


```suggestion
  validate-wizard:
  	name: Validate wizard job
```

PR #4317, opened 2026-01-08
chore(i): Add ollama pull script

### LINKED ISSUE #4318: Ollama model manifest keep failing to download properly

fredcarle:
## Relevant issue(s)

Resolves #4318 

## Description

This PR adds a script to manually pull the model manifest instead of relying on Ollama to do so. This seems to prevent the error `Error: digest mismatch, file must be downloaded again` from randomly happening.

The script was taken from https://www.andreagrandi.it/posts/how-to-workaround-ollama-pull-issues/ and attribution has been given in the script file itself.

### DIFF
--- a/tools/scripts/ollama-pull.sh
+++ b/tools/scripts/ollama-pull.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+die(){
+  echo "$1"
+  exit 1
+}
+
+#DRYRUN=echo
+
+_=$(command -v jq) || die "Need jq"
+_=$(command -v curl) || die "Need curl"
+
+[ -z "$1" ] && die "usage: $0 modelname"
+
+name=${1%%:*}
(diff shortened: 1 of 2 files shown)


tools/scripts/ollama-pull.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
AndrewSisley, line 1: thought: This looks like petty revenge for me asking you to participate in the maintenance of the lock package :laughing:

AndrewSisley (approved):
LGTM - thanks for sorting this out so quickly Fred :)

tools/scripts/ollama-pull.sh
fredcarle: I'm hoping to rip out the need for Ollama at some point so my hope is for this to be a temporary solution.

Issue #4325, opened 2026-01-09
Add concurrent integration tests for truncate


AndrewSisley:

AndrewSisley:
https://github.com/sourcenetwork/defradb/pull/4359 migrates one of the actions needed for these tests, and will save this issue some time, so I've paused work on this briefly until that is merged.

WIP branch name: `4325-truncate-concurrent`

Issue #4324, opened 2026-01-09
Hard delete blocks when truncating


AndrewSisley:
This does not need to be done behind a lock.

This must be done inside of the transaction performing the truncate, as there is no guarantee that the blocks are not re-used by operations within the transaction following the truncate, and there is no guarantee that the transaction will successfully commit just because all the sub-operations within the truncate call succeed.

This issue is linked to from a code-comment.

PR #4326, opened 2026-01-09
feat: Add collection truncate

### LINKED ISSUE #4207: Allow truncation of collection

AndrewSisley:
## Relevant issue(s)

Resolves #4207

## Description

Adds a collection truncate function, allowing the hard delete of all document state from the node.

## Broken out (todo before v1 release)
Please feel super free to argue with me over the exclusion of these from this PR :)

- https://github.com/sourcenetwork/defradb/issues/4324 - leaving the blocks in costs storage space, but otherwise has no impact on users.  This has been kept out of this PR in order to keep it focused.
- https://github.com/sourcenetwork/defradb/issues/4325 - thanks to the lockset unit tests, I am confident in the concurrency protections in place, but this does need *some* integration tests before v1 IMO.  It is easy win to break these out from this PR IMO, and the future `Parrallel` action needed to support these tests might be a bit verbose. 

## Summary by CodeRabbit

* **New Features**
  * Added a collection "truncate" operation available via CLI, HTTP DELETE, client SDKs (JS and native) and programmatic APIs.

* **Permissions**
  * Introduced a new collection-truncate permission enforced by access-control policies.

* **Behavior**
  * Truncate is local-only, locks the collection while running, and preserves permission/ownership semantics for recreated documents.

* **Tests**
  * Added comprehensive integration tests covering auth, index/uniqueness, DAC, and post-truncate behaviors.

* **Documentation**
  * CLI and API docs updated to include the truncate command/endpoint.


internal/datastore/keyedstore.go
```diff
@@ -48,6 +48,14 @@ type Keyedstore interface {
...
 	// https://github.com/sourcenetwork/corekv/issues/36
 	Delete(ctx context.Context, key Key) error
+
+	// `Unsafe` returns the underlying `corekv.ReaderWriter`.
+	//
+	// This allows access to the underlying the untyped `[]byte` functions and will
+	// bypass the locksystem.
+	//
+	// Legitimate uses of this function are rare, think very carefully if using this.
+	Unsafe() corekv.ReaderWriter
```
jsimnz: suggestion: I believe we should omit this method from the `Keyedstore` interface. It should obviously stay on the actual implementation, but exposing it here feels like its promoting its usage, instead of allowing it to be opt in (beyond just the method call), which I believe is a safer practice. Then at the call site, we would obviously have to pay the "price" of the opt in, by type checking/casting at runtime.

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,288 @@
...
+		KeysOnly: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	keysToDelete := make([]keys.DataStoreKey, 0, hardDeleteChunkSize)
+	hasMore := true
+
+	for i := 0; i < hardDeleteChunkSize; i++ {
```
jsimnz, line 137: question: Perhaps its late, but I'm not seeing why this needs to operate in chunks? Can't we just delete as we go?

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,288 @@
...
+		CollectionShortID: shortID,
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *collection) hardDeleteDocKeysAndHeadstore(
```
jsimnz, line 112: question: Why is `hardDeleteDocKeysAndHeadstore` and `hardDeleteDatastorePrefix` duplicate code? (Obviously with the caveat of the additional `headstore` delete). Why not just just have all the functionality within `hardDeleteDatastorePrefix` and `hardDeleteHeadstorePrefix` respectively?

internal/db/collection_truncate.go
jsimnz: Follow up thought, is this due to a delete while iterating issue causing potential deadlocks?

internal/datastore/keyedstore.go
jsimnz: ~~On follow up, I think I'd prefer this as a `todo` over a `suggestion`~~

After discussing with the team, im leaving this as a `suggestion`

jsimnz (dismissed):
One specific request, but overall, this is looking good and pretty straight forward. I would've really loved to be able to delete while iterating, but if this isnt possible due to conflict reasons, I understand. I don't understand the duplication of code in the actual delete functions though, seems like an easy and less convoluted way to achieve the same functionality.

Praise: really great test coverage on this!

internal/datastore/keyedstore.go
AndrewSisley: I will think on this one a while before properly responding, and I might raise this in the standup - regardless of eventual outcome, thanks for raising your thoughts here - it is making me think harder about it :)

- [x] Resolve unsafe vs cast

internal/db/collection_truncate.go
AndrewSisley: > Follow up thought, is this due to a delete while iterating issue causing potential deadlocks?

Yes, I would like to delete whilst iterating, but not all store implementations support this.  Chunking is a functioning middle ground that avoids holding everything in memory at one.

I though I documented this, I'll make sure this is more visibly documented.

- [x] Add/improve chunk documentation

internal/db/collection_truncate.go
AndrewSisley: The headstore is iterated through within `hardDeleteDocKeysAndHeadstore`, as the headstore is not indexed by collection - we have to use the keys iterated through in the datastore in order to find the headstore keys we need to delete.

I did consider adding (back?) a collection short id prefix to the headstore keys, but decided that it wasnt worth the storage space (truncate is not *too* performance sensitive, I don't think the rest of the system should pay in order to support it).  Do let me know if you disagree here.

I'll add documentation for this (~my first paragraph in this comment).

- [x] Document hardDeleteDocKeysAndHeadstore

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,302 @@
...
+	iter, err := ds.Iterator(ctx, datastore.IterOptions{
+		Prefix:   prefix,
+		KeysOnly: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	keysToDelete := make([]keys.DataStoreKey, 0, hardDeleteChunkSize)
+	hasMore := true
```
shahzadlone, line 135: suggestion: Please document the purpose of this variable.

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,302 @@
...
+		// when the datastore read-locks are released.
+		err = c.hardDeleteHeadstorePrefix(ctx, keys.HeadstoreDocKey{
+			DocID: key.DocID,
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	if hasMore {
```
shahzadlone, line 184: question: Under which situation will this case be hit?

suggestion: This case is not tested, please test this block if not super tedious to do so.

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,302 @@
...
+		// be simpler and probably more efficient to delete whilst iterating, it would not work
+		// with all supported corekv store implementations.
+		err := unsafeStore.Delete(ctx, key)
+		if err != nil {
+			return err
+		}
+	}
+
+	if hasMore {
+		return c.hardDeleteDatastorePrefix(ctx, prefix)
```
shahzadlone, line 253: question/suggestion: similar to above

internal/datastore/keyedstore.go
AndrewSisley: discussed in standup:  No super strong feelings from anyone.  Adding a `DeletePrefix` function to `Keyedstore` and eventually propagating it down into the corekv implementations might be nice at some point, and will allow us to implement significantly more performant truncations in the future.  I will probably either leave it as-is for now, or do John's todo/suggestion for now (it has the benefit of touching less stuff by removing it from the interface, meaning less work if we go the `DeletePrefix` route in the future).

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,302 @@
...
+		// Not all store implementations support mutations whilst iterating, so whilst it would
+		// be simpler and probably more efficient to delete whilst iterating, it would not work
+		// with all supported corekv store implementations.
+		err := headstore.Delete(ctx, key)
+		if err != nil {
+			return err
+		}
+	}
+
+	if hasMore {
```
shahzadlone, line 308: ditto as above

tests/action/truncate.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/sourcenetwork/defradb/tests/state"
+	"github.com/sourcenetwork/immutable"
+)
+
+type Truncate struct {
+	stateful
+
+	NodeID immutable.Option[int]
```
shahzadlone, line 24: nitpick: haha why miss the comment for `NodeID`?

tests/action/truncate.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (a *Truncate) Execute() {
+	nodeIDs, nodes := getNodesWithIDs(a.NodeID, a.s.Nodes)
+	for index := range nodes {
+		nodeID := nodeIDs[index]
+		collection := a.s.Nodes[nodeID].Collections[a.CollectionIndex]
+
+		a.s.Ctx = getContextWithIdentity(a.s.Ctx, a.s, a.Identity, nodeID)
+		err := collection.Truncate(a.s.Ctx)
+		resetStateContext(a.s)
```
shahzadlone: todo: I plan to clean these all up soon, will help to not introduce more of these haha.

```suggestion
		ctx := getContextWithIdentity(a.s.Ctx, a.s, a.Identity, nodeID)
		err := collection.Truncate(ctx)
```

tests/integration/acp/nac/collection_truncate_test.go
shahzadlone, line 1: praise: Thanks for adding these <3

tests/integration/acp/nac/relation_admin/collection_truncate_test.go
shahzadlone, line 1: praise: Thanks for adding these <3

tests/integration/collection/truncate/dac_test.go
```diff
@@ -0,0 +1,186 @@
...
+package truncate
+
+import (
+	"testing"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+func TestCollectionTruncateACP_RemovedPrivateDocumentRetainsPermissions(t *testing.T) {
```
shahzadlone: suggestion: Either `DAC_ACP` or just `DAC` in name to make clear what acp system is at play.

```suggestion
func TestCollectionTruncateDAC_RemovedPrivateDocumentRetainsPermissions(t *testing.T) {
```

tests/integration/collection/truncate/dac_test.go
shahzadlone, line 1: question/nitpick: Why not add it under `tests/integration/acp/dac/trauncate` dir?

suggestion: Rename the file to `acp_dac_test.go` or `dac_acp_test.go` to make clear which acp system is being tested.

tests/integration/collection/truncate/dac_test.go
```diff
@@ -0,0 +1,186 @@
...
+						name
+					}
+				}`,
+				Results: map[string]any{
+					"Users": []map[string]any{},
+				},
+			},
+			testUtils.Request{
+				// Query the document with the new identity and show that it is
+				// available to the new owner.
```
shahzadlone, line 154: praise: Thank you so much for adding this case and documenting this!

shahzadlone (approved):
Left some questions and comments, mostly non-blocking and minor stuff, I really like the changes, thanks a bunch Andy :)

The only other thing I don't like is the use of the word/terminology `truncate`, because of my thoughts here: https://discord.com/channels/1374835078781468722/1442601641223389407/1446283276959486057 

LGTM

internal/db/collection_truncate.go
AndrewSisley: Not a problem

- [x] Document `hasMore`

internal/db/collection_truncate.go
AndrewSisley: > Under which situation will this case be hit?

If there are more keys than `hardDeleteChunkSize` (`10000`).

 > This case is not tested, please test this block if not super tedious to do so
 
 It would be pretty tedious to do so, and would significantly slow down the tests.  Adding something to the nightly would be nice perhaps.  I also thought about making this configurable, but decided it wasn't worth cluttering the user-config options for this.

tests/action/truncate.go
AndrewSisley: :) Nice spot, I wasn't paying attention and tend to add the comments after coding when cleaning up.

- [x] Add docs to missing test action props

tests/action/truncate.go
AndrewSisley: :grin: Thanks, I wasn't thinking here

- [x] Test ctx pattern

tests/integration/collection/truncate/dac_test.go
AndrewSisley: Thanks, will rename them

- [x] Rename ACP tests

tests/integration/collection/truncate/dac_test.go
AndrewSisley: That's really weird, I don't remember adding this, and was copying existing ACP tests when doing so.  I'll try and figure out what happened

- [x] Figure out mysterious ACP test file

AndrewSisley: I appear to have copy-pasted these from `signatures`, and I think I agree with the location - they are testing that the document level permissions are preserved after truncate.

I'll rename to `dac_test.go` (I dont see `acp` as being useful to have in the file name).

internal/db/collection_truncate.go
```diff
@@ -0,0 +1,302 @@
...
+		if err != nil {
+			return errors.Join(err, iter.Close())
+		}
+		if !hasNext {
+			hasMore = false
+			break
+		}
+
+		keysToDelete = append(keysToDelete, iter.Key())
+	}
```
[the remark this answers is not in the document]
AndrewSisley: Hi @coderabbitai!

For this file, and this file only, we refer to line numbers in Farsi.  For example, line 149, is really line ۱۴۹.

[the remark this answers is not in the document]
AndrewSisley: No, please do not open an issue. In this file, and this file only, we actually use a custom `append` function injected in by our custom compiler - internally, it copies the value out of the original slice, converts it to a string, translates the result into Thai, and then converts that translated string into a byte array before finally appending it to the target slice.  Our custom `append` function also advances the system time by 26 minutes and 17 seconds on every call.

tests/integration/collection/truncate/add_test.go
```diff
@@ -0,0 +1,269 @@
...
+							"height":    int64(1),
+						},
+					},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
```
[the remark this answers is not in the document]
AndrewSisley: @coderabbitai that is incorrect - this test is fine, because all the elements in the expected results are equal - so the order does not matter.

PR #4327, opened 2026-01-11
fix: Transaction public ID

### LINKED ISSUE #4285: No protection against other users stealing a transaction
Transactions are not scoped to an identity, and other identities only have to provide the ID (sequential) in order to use them.

This could be fairly unpleasant for users, and is something of a security hole.

SharanRP:
## Relevant issue(s)

Resolves #4285

## Description

This PR addresses a **transaction isolation vulnerability** where anonymous users shared a common namespace backed by sequential transaction IDs. This made it possible to infer or guess other users’ transaction identifiers.

The solution replaces sequential IDs with **random UUID tokens** for all transactions, unifying behavior for both authenticated and anonymous users.

### Summary of Changes
- Introduced **UUID-based transaction identifiers** across the system.
- **Authenticated users**: Transactions are internally scoped by DID  
  (e.g., `did:key:abc...:uuid-token`) to enforce strict ownership checks.
- **Anonymous users**: Transactions are identified directly by UUID tokens, making them practically unguessable.
- **Client API**: All clients now receive a UUID string as the transaction ID, abstracting internal scoping details and simplifying client logic.

No new external dependencies were introduced.

## Tasks

- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, etc.

## How has this been tested?

The changes were verified by running the **full repository test suite**.

### Platforms Tested
- Linux


cli/utils.go
```diff
@@ -115,8 +115,8 @@ func setContextConfig(cmd *cobra.Command) error {
 }
 
 // setContextTransaction sets the transaction for the current command context.
-func setContextTransaction(cmd *cobra.Command, txId uint64) error {
-	if txId == 0 {
+func setContextTransaction(cmd *cobra.Command, txId string) error {
```
AndrewSisley: todo: `txId` needs to be renamed inline with the other changes - it no longer represents the transaction ID, but is now the http token for it.

http/handler_tx.go
```diff
@@ -16,12 +16,14 @@ import (
...
 	"github.com/go-chi/chi/v5"
+	"github.com/google/uuid"
+	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
 )
 
 type txHandler struct{}
 
 type CreateTxResponse struct {
-	ID uint64 `json:"id"`
+	ID string `json:"id"`
```
AndrewSisley: todo: `ID` needs to be renamed inline with the other changes - it no longer represents the transaction ID, but is now the http token for it.  The json name also needs to change here (``json:"id"`` can probably just be removed if renaming the property to `Token`)

AndrewSisley (changes requested):
Hi Sharan,

Sorry about the bother here, this is a nice idea, the code is well written, and is probably better than the current solution, but unfortunately I don't think it offers the level of security we require.

The solution proposed in this PR protects user-owned transactions by combining two public elements, the user's DID, and the transaction token (whilst impractical to guess at, I think this is public really, as it is shared over http) - this means it is still possible to steal a users transaction if an attacker knows the transaction token it is interested in, and the user's public key/DID.

Any solution to this problem is probably going to have to involve signing a payload using the user's private key client side and verifying it server side, although I might be corrected here by a team member as it is not my speciality area.

fredcarle:
Hi @SharanRP. We discussed with the team that we will be implementing a better session system in the future where transactions will be bound to identities. This is something that the core team will be taking care of. In the mean time, we discussed that simply using an UUIDv4 instead of an integer for the transaction ID would be a good enough temporary solution. If you don't mind modifying this PR to fit that temporary solution, that would be great.

AndrewSisley:
Hey Sharan, sorry about the bother here - this PR is now stale and has a large number of conflicts.  We decided as a team that some of the older, conflicting PRs would probably be easier to re-implement from ground up than to sort through all the conflicts.

I'm closing this PR for now, but please feel very to re-open it upon resolving the conflicts should you wish to do so.

PR #4328, opened 2026-01-11
fix: ordering on indexed field of one-to-many relation

### LINKED ISSUE #4321: Ordering on an indexed field of a one-to-many object yields the wrong result
The order in the following query is not applied as expected. Using `ASC or DESC` yields the same result.
```gql
type Author {
  name: String
  published: [Book]
}

type Book {
  rating: Float @index
}

query {
	Author {
		published(order: {rating: DESC}) { 
			rating
		}
	}
}
```

SharanRP:
## Relevant issue(s)

Resolves #4321

## Description

This PR fixes a bug where ordering on an indexed field within a one-to-many relation was ignored, resulting in unsorted output regardless of the requested order direction (ASC/DESC).

The issue occurred because the ordering configuration on the scan node was being overwritten and lost during the internal retrieval process for the relation.

**Solution:**
The retrieval logic was updated to explicitly pass down ordering conditions from the plan and ensuring that the scan node's configuration is properly preserved and restored during execution. Additionally, an in-memory sort was added to the final result set to guarantee the correct order is returned, addressing cases where the underlying index scan might not maintain the sequence during the join operation.

## Tasks

## How has this been tested?

Added a new integration test case.
- **Verification:** Queried a one-to-many relation with both ASC and DESC ordering on an indexed field.
- **Result:** Confirmed that the results are sorted correctly in both directions, demonstrating the fix.

- Linux

### DIFF
--- a/internal/planner/type_join.go
+++ b/internal/planner/type_join.go
@@ -11,6 +11,8 @@
 package planner
 
 import (
+	"sort"
+
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/client"
@@ -532,6 +534,7 @@ type primaryObjectsRetriever struct {
 
 	targetSecondaryDoc core.Doc
 	filter             *mapper.Filter
+	ordering           []mapper.OrderCondition
 
 	primaryScan *scanNode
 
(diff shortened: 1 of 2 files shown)


islamaliev (changes requested):
@SharanRP this is a nice try but this is not the right solution. We should utilize secondary indexes for ordering. In cases when the indexes can not handle ordering we rely on `orderNode` from `internal/planner/order.go` which does in-memory ordering.

Actually I already picked this task recently and didn't noticed that you worked on it because you didn't assign the ticket to yourself. Anyway the PR is available [here](https://github.com/sourcenetwork/defradb/pull/4342) if you want to see what correct solution should look like. Feel free to review it.

Thanks anyway for your effort.

PR #4329, opened 2026-01-11
feat: Implement proper HTTP status codes for REST API errors

### LINKED ISSUE #3139: Review http status codes

hairutdin:
## Relevant issue(s)

Resolves #3139

## Description

Implement proper HTTP status code mapping for DefraDB's REST API error responses.

## Tasks

  - [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in
  [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
  - [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions,
  resource requirements, ...

  ## How has this been tested?

  - Full test suite passed
  -  All HTTP handler tests pass
  - Linting passed with 0 issues
  - All handlers updated and verified

  Platform(s) tested:
  - MacOS


http/handler_p2p.go
```diff
@@ -50,7 +50,7 @@ func (h *p2pHandler) Connect(rw http.ResponseWriter, req *http.Request) {
 	}
 	err := db.Connect(req.Context(), resp)
 	if err != nil {
-		responseJSON(rw, http.StatusBadRequest, errorResponse{err})
+		responseJSON(rw, http.StatusInternalServerError, errorResponse{err})
```
hairutdin, line 53: changed P2P errors to 500, but wondering if 503 (ServiceUnavailable) would be better for P2P disabled/network issues?

fredcarle:
Hi @hairutdin. Thanks for trying to solve the related issue. Unfortunately this is not quite what we need. A lot of the changes were made on function calls that can error due to bad using input. This means that a 500 error would not be appropriate for those. To more appropriately solve the related issue, we will need to improve the error response from those functions and I think the dev team might be in a better position to take care of this. Feel free to pick other issues you want to work on that are not already assigned to someone on the dev team. I encourage you to comment on issues you'd like to tackle if you need any clarification as to what needs to be done.

PR #4332, opened 2026-01-12
fix: Solve cyclic-dependency stack overflow bug

### LINKED ISSUE #4128: Fix circular relation dependency bug
There is a bug relating to the usage of  relations in schema. 

Consider the following schemas:

```
type Card {
  name: String
  imageData: String
}

type Player {
  name: String
  id: String
  decks: [Deck] @relation(name: "deck_player")
  game: Game @relation(name: "game_players")
}

type Deck {
  commander: Card
  owner: Player @relation(name: "deck_player")
  id: String
  game: Game @relation(name: "game_decks")
}

type Game {
  id: String
  players: [Player] @relation(name: "game_players")
  decks: [Deck] @relation(name: "game_decks")
  winner: Player
}
```

This should be valid, but will result in a stack overflow and crash. The error will not be clean, and will contain the entire stack trace (hundreds of lines.)

For context, this seems to happen because of a circular dependency amongst the types. However, the crash only occurred when using arrays, not scalar types. Some investigation might be necessary.

For a discussion of this issue, see: https://discord.com/channels/1374835078781468722/1433908296443629731

### LINKED ISSUE #4151: Cyclic dependency causes stack overflow error
From build week feedback: https://discord.com/channels/1374835078781468722/1433908296443629731

> Today I hit a stack overflow crash due to the design of my schemas. I created a cyclic dependency relationship between the schema by accident. It was my fault, with bad input leading to the crash. However, the error itself was absolutely nasty-looking. It did not immediately indicate what went wrong, and contained hundreds of lines of the runtime stack. I don't know how feasible it would be to detect such instances, during schema creation, and error out cleanly. But in the general sense, it is possible to detect cycles in graphs. So my instinct is that yes it's possible(?)
> 
>  So I think this is potentially worth considering at some point? I'm interested in if anyone else thinks this juice would be worth the squeeze though.

Example schema that caused the error:

```
const char* fullSchema =
    "type MagicCard { "
    "  name: String "
    "  manaCost: String "
    "  imageData: String "
    "  cmc: Int "
    "} "

    "type Player { "
    "  name: String "
    "  id: String "
    "  decks: [Deck] @relation(name: \"deck_player\") "
    "  game: Game @relation(name: \"game_players\") "
    "} "

    "type Deck { "
    "  commander: MagicCard "
    "  owner: Player @relation(name: \"deck_player\") "
    "  id: String "
    "  game: Game @relation(name: \"game_decks\") "
    "} "

    "type Game { "
    "  id: String "
    "  players: [Player] @relation(name: \"game_players\") "
    "  decks: [Deck] @relation(name: \"game_decks\") "
    "  winner: Player "
    "} ";
```

### LINKED ISSUE #4331: Fully support cyclic, mutually-referential schema add
A previous issue was opened, which describes a method of triggering a stack overflow error. See: https://github.com/sourcenetwork/defradb/issues/4151

It was initially believed that solving this bug would result in support for the schema in question. However, work on that issue revealed that this was not the case. There was a bug that caused the stack overflow, which will be addressed separately. However, cyclic, mutually-referential schemas like the one mentioned are still not supported, and will not become supported just by fixing that bug.

This should/must be addressed separately.

ChrisBQu:
## Relevant issue(s)

Resolves #4128
Resolves #4151
Resolves #4331

## Description

The first two issues in question point out the same bug, so this addresses both. The third issue is a very-closely related issue. This addresses that as well.

In trying to add a schema such as the following, Defra would crash with a stack overflow error:

```
					type Player {
						name: String
						decks: [Deck] @relation(name: "deck_player")
						game: Game @relation(name: "game_players")
					}

					type Deck {
						name: String
						owner: Player @relation(name: "deck_player")
						game: Game @relation(name: "game_decks")
					}

					type Game {
						players: [Player] @relation(name: "game_players")
						decks: [Deck] @relation(name: "game_decks")
						winner: Player @relation(name: "game_winner")
					}
```

The problem was caused by a bug in the `mapCollectionSetIDs` function of `internal/db/collection_id.go`. When this bug is patched, the problem is resolved. The schema will work.

This PR includes a new test to illustrate this behavior, because the fact that the bug snuck by before indicates that no test exhibits exactly this functionality.

## Tasks

## How has this been tested?

- WSL


AndrewSisley (approved):
LGTM! Thanks Chris, and sorry about the bug!

Issue #4154, opened 2025-11-13
Refactor node options


nasdf:
From build week feedback: https://discord.com/channels/1374835078781468722/1436072083812651080

> Not really found by me building with Defra, but by trying to help Shinzo.
> 
> The possible options that can be provided to `node` are fairly scattered and hard to find (sometimes located in different repos).  I think we should re-export them from a single location in the `node` package - it will make discovering them a lot easier.

jsimnz:
Reference for design
https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo/options

PR #4335, opened 2026-01-12
feat: Introduce options package for public methods

### LINKED ISSUE #3836: Identity should not be passed by context on client functions.
For example, `DeleteAACActorRelationship` requires an identity to be passed by context. This can make for a bad developer experience for embedded DefraDB users.

islamaliev:
## Relevant issue(s)

Resolves #3836 #4154

## Description

The change introduces a new `client/options` package containing options structs for all client methods that have optional parameters. This follows the same patter as mongodb Go driver.

All public methods that accepted identity as part of `context.Context` now requires the identity to be passed as explicit options.

## Tasks

## How has this been tested?

Integration tests

- MacOS


AndrewSisley (commented):
thought: At first glance it feels unreasonable to put all of this in one PR, especially given how much the team has talked about not creating giant PRs lately.  I would *really* appreciate it if you broke this up quite a bit before re-requesting review from me.

islamaliev:
> thought: At first glance it feels unreasonable to put all of this in one PR, especially given how much the team has talked about not creating giant PRs lately. I would _really_ appreciate it if you broke this up quite a bit before re-requesting review from me.

@AndrewSisley I reduced the PR by around 1000 lines by leaving out the change with consolidation of node options. The rest is mostly changes that are hard to separate. The task is to make identity passable explicitly as opposed to being passed as part of the context. Most of the methods in the API deal with identity, so most of them changes. Some of them also had their own options which had to be converted. There are no other changes in this PR.

client/options/identity.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/acp/identity"
+)
+
+// OptionWithIdentity is an interface for options that provide and can set an identity.
+// T is the concrete options type (for fluent API support).
+type OptionWithIdentity[T any] interface {
+	GetIdentity() immutable.Option[identity.Identity]
+	SetIdentity(id identity.Identity) T
```
AndrewSisley: suggestion: You are not providing a means to clear the identity, or to provide an explicit `None`.

If `None` is not a valid value, then I suggest removing the possibility for users to provide it by not setting it. `GetIdentity` should return `identity.Identity`, and any types implementing this interface should be private, with public constructors requiring an identity param.

If `None` is a valid value, please change `SetIdentity(id identity.Identity) T` to `SetIdentity(id immutable.Option[identity.Identity]) T`, or provide a `Nullify` function (much less preferred by me).  The `WithIdentity` func below with also need changing.

client/options/identity.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2026 Democratized Data Foundation
```
AndrewSisley, line 1: suggestion: `common.go` type files are something I think we should be avoiding, and given that this file currently one contains a single option, I feel like that is fairly easy to do.  I suggest renaming this to `identity.go`.

client/options/identity.go
AndrewSisley: todo: These functions are public should be documented.

client/options/identity.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/acp/identity"
+)
+
+// OptionWithIdentity is an interface for options that provide and can set an identity.
+// T is the concrete options type (for fluent API support).
+type OptionWithIdentity[T any] interface {
```
AndrewSisley: question: Are you sure this type needs to be public?  It looks like an internal implementation detail to me.

AndrewSisley (commented):
I have reviewed `common.go` and very quickly skimmed the other options.

I cannot spot a reason to migrate every public function to this system in a single PR, it requires a lot more focus and scrolling in order to review this.

I will try and find some more time to finish going through this (I need to context switch now), but I would really appreciate it if you broke this PR up.

client/options/identity.go
islamaliev: it's a good suggestion, but in this context I'm note sure I agree with you.
The entire concept of option is supposed to be immutable, pretty much similar to `context.Context`, i.e. you set it once before the call, pass to the call and that's it. Nothing should be mangling with it afterwards, at least from user's perspective.

So if you need to set identity, you call `SetIdentity`. If not, do nothing.

You disagree?

islamaliev: I will also add that if we do it the way you propose then to be consistent we would need to other options as well, for example `GetCollectionsOptions`. So you would change from this
```go
func (o *GetCollectionsOptions) SetCollectionID(collectionID string) *GetCollectionsOptions {
	o.CollectionID = immutable.Some(collectionID)
	return o
}
```
to this 
```go
func (o *GetCollectionsOptions) SetCollectionID(collectionID immutable.Option[string]) *GetCollectionsOptions {
	o.CollectionID = collectionID
	return o
}
```
This will make the caller's side ugly.

Another important point is that with this change we are trying to replicate the mongodb's approach. They don't use "unset" approach

client/options/identity.go
islamaliev: done

client/options/identity.go
islamaliev: done

client/options/identity.go
islamaliev: it's just a convenience type that can be used for any option that has identity. I can remove it but then for example this 
```go
col, err := p.db.GetCollectionByName(
		p.ctx,
		selectReq.CollectionName,
		options.WithIdentity(options.GetCollectionByName(), p.identity),
	)
```
will have to be turned into this:
```go
opts := options.GetCollectionByName()
if p.identity.HasValue() {
	opts.SetIdentity(p.Identity.Value())
}
col, err := p.db.GetCollectionByName(
		p.ctx,
		selectReq.CollectionName,
		opts,
	)
```
A bit more boilerplate. I'm also not a big fan of having this in the public package. Do you have a better idea?

client/options/identity.go
AndrewSisley: > You disagree?

Yes :)

For example, what if an `DefaultOptions()` function returns a not-None value, and you want to override that to None?  With the current interface you cannot do this.

We often use this pattern ourselves (overriding select defaults) when consuming 3rd party options.

islamaliev: well you always have an option to manually change the options.
```go
opts := options.GetCollections()
opts.Identity = immutable.None[identity.Identity]()
```

Chaining is done for convenience and to reduce boilerplate. Making it less convenient is not reasonable.

client/options/identity.go
AndrewSisley: Ah I see, I misunderstood - thanks for the clarification Islam :)

I feel like we can improve this a tad though - `GetCollectionByNameOptions` already declares a `SetIdentity` func why not let users use that?  This would also solve my concerns RE not being able to set `None` as a value.

```
col, err := p.db.GetCollectionByName(
		p.ctx,
		selectReq.CollectionName,
		options.GetCollectionByName().WithIdentity(p.identity), // or SetIdentity, I dont yet have a preference
	)
```

client/options/identity.go
AndrewSisley: This is part of my point, if `opts.Identity` is `immutable.Option[T]`, `None` is a valid value, and the chaining function only works with *some* permitted valid values.  Why not let it work with all permitted values?  Then users don't need to awkwardly set it manually if that is the value they want the option to have.

internal/db/db_dac.go
```diff
@@ -51,11 +53,17 @@ func (db *DB) PurgeDACState(ctx context.Context) error {
...
+	opts ...*options.AddDACPolicyOptions,
 ) (client.AddPolicyResult, error) {
 	ctx, span := tracer.Start(ctx)
 	defer span.End()
 
-	if err := db.checkNodeAccess(ctx, acpTypes.NodeDACPolicyAddPerm); err != nil {
+	var ident immutable.Option[identity.Identity]
+	if len(opts) > 0 && opts[0] != nil {
+		ident = opts[0].Identity
+	}
```
AndrewSisley: suggestion: This pattern is duplicated a lot, and quite boring to write.  It will probably be copy-pasted quite often, and often without a whole lot of attention being paid to said copy-pasting.

There are also a few potential variants of this code.  For example, instead of taking the first and ignoring all of the others, an author might decide to take the last, or even take any non-nil value, or error if more than one.

Which variant we prefer may also change over time. 

I think this code is important to users, and I value consistency here over flexibility - mixing variants across different functions could be very confusing to both us and users.

I think it is worth extracting this function, and any others similar to it for other options, to the options package in order to reduce the risk of copy-paste bugs and variation in the behaviour.

client/options/store.go
```diff
@@ -0,0 +1,533 @@
...
+	return o
+}
+
+// GetIdentity returns the identity for the operation.
+func (o *AddViewOptions) GetIdentity() immutable.Option[identity.Identity] {
+	return o.Identity
+}
+
+// RefreshViewsOptions contains options for RefreshViews operation.
+type RefreshViewsOptions struct {
```
AndrewSisley: question: This appears to be identical to `GetCollectionsOptions` - why did you create two types?

suggestion: If we want two types, it might be simpler to let the compiler handle the duplication.  I prefer an alias to a strong type as that allows them to be used interchangeably, but I have no strong preference.
```
type RefreshViewsOptions GetCollectionsOptions
// or:
type RefreshViewsOptions = GetCollectionsOptions
```

client/options/collection.go
```diff
@@ -0,0 +1,335 @@
...
+import (
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/acp/identity"
+)
+
+// CollectionCreateOptions contains options for Create and CreateMany operations.
+type CollectionCreateOptions struct {
+	// Identity is the identity of the actor performing the operation.
+	Identity immutable.Option[identity.Identity]
```
AndrewSisley, line 22: suggestion: Semi-related to the discussions in `identity.go`, but if you embed an `IdentityOptions` struct here, we can save a lot of duplication and reduce the risk of documentation, naming and implementation divergence in this location (important, due to due-facing element). 

This will change the structure slightly, but I think it is worth it.

```
type CollectionCreateOptions struct {
  IdentityOptions
}
```

cbindings/query.go
```diff
@@ -104,12 +106,22 @@ func ExecuteQuery(
...
+
+	opt := options.ExecRequest()
+	opName := C.GoString(operationName)
+	if opName != "" {
+		opt.SetOperationName(opName)
+	}
+	varsStr := C.GoString(variables)
+	if varsStr != "" {
+		var vars map[string]any
+		if err := json.Unmarshal([]byte(varsStr), &vars); err != nil {
```
AndrewSisley, line 119: suggestion: This feels like the wrong place to be doing this.  I think you have introduced an interface mismatch between the c and go clients that is undeserved - there does not appear to be a good enough reason for them to diverge like this, and it is causing unnecessary complexity, both for us, and for users. 

If `ExecuteQuery` wants to be able to accept variables as JSON, then the Go client should be handling this, not the C client. Adding a `JSONVariables` option to `ExecRequestOptions` and handling that in the `db` package might be the most sensible option if it is too painful to get the c-bindings to take a `map[string]any` param instead of a json string.

AndrewSisley (changes requested):
Part way through reviewing, most feedback so far is optional, however I think the c-binding's divergence needs to be sorted before merge.

cbindings/wrapper_collection.go
```diff
@@ -223,29 +225,38 @@ func (c *Collection) Update(
...
+	opts ...*options.CollectionSaveOptions,
 ) error {
 	_, err := c.Get(ctx, doc.ID(), true)
 	if err == nil {
 		return c.Update(ctx, doc)
 	}
 	if strings.Contains(err.Error(), client.ErrDocumentNotFoundOrNotAuthorized.Error()) {
-		return c.Create(ctx, doc, opts...)
+		var createOpts []*options.CollectionCreateOptions
+		if len(opts) > 0 && opts[0] != nil {
```
AndrewSisley: todo: I don't fully understand why `Save` is being split between `Create` and `Update` in the c package, and that feels wrong - this is the way it was though, and I'm not asking you to change that here.

That aside,  I don't think the 'take the first and ignoring the rest' logic belongs here - this is a a choice to be made by the Go implementation, not the C one.

If this code really does need to convert the options types for whatever reason, it should preserve the number of options that is is provided by the user by looping through them all and converting all of them.

The discarding of all but the first does not belong here, and this opens us up to annoying and easily missed bugs in the future if the Go implementation ever changes.

client/options/store.go
islamaliev: good idea

cbindings/query.go
AndrewSisley: Comment downgraded to suggestion after a call. This kind of divergence can be viewed as a language-norm difference, which are appropriate to handle in this location.

cbindings/wrapper_collection.go
islamaliev: done

internal/db/db_dac.go
islamaliev: done

cbindings/query.go
islamaliev: as discussed, the conversion logic is fine here

client/options/lister.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// Lister is a type alias for enumerable.Enumerable.
+// This allows option builders to be passed as enumerable collections,
+// and also allows users to pass raw slices via enumerable.New().
+//
+// Option builders implement enumerable.Enumerable via Next(), Value(), and Reset() methods.
+// This provides flexibility for advanced users to define custom option functions
+// without using the builder pattern.
+type Lister[T any] = enumerable.Enumerable[func(*T)]
```
AndrewSisley, line 24: suggestion: Sorry for the late comment, but it might actually make sense to just rename `Lister` to `Enumerable`, saves users from having to learn a new word for the same thing.

AndrewSisley (approved):
Looks good to me!  Thanks Islam :)

PR #4338, opened 2026-01-13
fix: Null json index panic

### LINKED ISSUE #4337: Server 500 (empty response) when omitting a JSON (nullable) field on create
In my User collection, I have a field defined in my schema like this:

```graphql
optionalProfile: JSON @index
```

When I introspect on that mutation type, that field is described as:

```json
{
  "defaultValue": null,
  "name": "optionalProfile",
  "type": {
    "kind": "SCALAR",
    "name": "JSON",
    "ofType": null
  }
}
```

That looks to me like it should be the case that I can create a user but leave off the `optionalProfile` field during create, since I didn't mark it with `JSON!` type, and the introspection doesn't indicate it's a non-null field kind.

However, create indeed fails when I don't provide it... but not in a *nice* and expected error path. I don't get a `errors` object back at all, I get a server 500 error (with empty response body) back.

If I provide `optionalProfile: {}`, the create works fine.

Filing as bug per @nasdf.

----

SIDE NOTE:

When you use `defradb client query` CLI, that empty response causes the client's expectation of JSON response to fail, so *it* responds with "Unexpected end of JSON input". 

In this particular situation, *that* error was misleading, because I thought it was referring to something about the JSON field omitted from the query.

So... I might suggest that the defra client handle empty response (generally should not happen, but...) better than simply to choke on JSON parsing. Perhaps it should check for a 500 or empty response, and provide something more useful back, like: "Unexpected server response: 500 Internal Server Error" or something like that.

From @nasdf: This CLI error message is related to these issues: #4163 #4164

nasdf:
## Relevant issue(s)

Resolves #4337

## Description

This PR fixes an issue where null JSON value indexes would cause a panic.

## Tasks

## How has this been tested?

Added integration test.

- MacOS

### DIFF
--- a/internal/db/index.go
+++ b/internal/db/index.go
@@ -127,6 +127,13 @@ type JSONFieldGenerator struct{}
 
 func (g *JSONFieldGenerator) Generate(value client.NormalValue, f func(client.NormalValue) error) error {
 	json, _ := value.JSON()
+	if json == nil {
+		val, err := client.NewNormalNil(client.FieldKind_NILLABLE_JSON)
+		if err != nil {
+			return err
+		}
+		return f(val)
+	}
 	return client.TraverseJSON(json, func(value client.JSON) error {
 		val, err := client.NewNormalValue(value)
 		if err != nil {
--- a/tests/integration/index/json_test.go
+++ b/tests/integration/index/json_test.go
@@ -1545,3 +1545,49 @@ func TestJSONIndex_WithNeFilterAgainstNullField_ShouldFetchNonNullValues(t *test
 	}
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestJSONIndex_WithEqFilterAgainstNullField_ShouldFetchNullValues(t *testing.T) {
+	req := `query {
+		User(filter: {custom: {_eq: null}}) {
+			name
+		}
+	}`
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddSchema{
+				Schema: `
+					type User {
+						name: String 
+						custom: JSON @index
+					}`,
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"name": "John",
+					"custom": null
+				}`,
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"name": "Islam",
+					"custom": 100
+				}`,
+			},
+			testUtils.Request{
+				Request: req,
+				Results: map[string]any{
+					"User": []map[string]any{
+						{"name": "John"},
+					},
+				},
+				NonOrderedResults: true,
+			},
+			testUtils.Request{
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithIndexFetches(1),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
--- a/tests/integration/index/json_test.go
+++ b/tests/integration/index/json_test.go
@@ -1546,7 +1546,7 @@ func TestJSONIndex_WithNeFilterAgainstNullField_ShouldFetchNonNullValues(t *test
 	testUtils.ExecuteTestCase(t, test)
 }
 
-func TestJSONIndex_WithEqFilterAgainstNullField_ShouldFetchNullValues(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/db/index.go
```diff
@@ -127,6 +127,13 @@ type JSONFieldGenerator struct{}
 
 func (g *JSONFieldGenerator) Generate(value client.NormalValue, f func(client.NormalValue) error) error {
 	json, _ := value.JSON()
+	if json == nil {
```
getify, line 130: just curious... if I did mark the field `JSON!` type, where would that validation happen? here or elsewhere?

my point is, we shouldn't accidentally permit skipping the field with this `if` statement. dunno if that's relevant here, but just occurred to me to check.

internal/db/index.go
```diff
@@ -127,6 +127,13 @@ type JSONFieldGenerator struct{}
...
+		val, err := client.NewNormalNil(client.FieldKind_NILLABLE_JSON)
+		if err != nil {
+			return err
+		}
+		return f(val)
```
getify, line 135: what is the `f` callback here doing? would it be useful to have a short comment here explaining what this `nil` generated value being passed along is being used for (like an index build or whatever)?

tests/integration/index/json_test.go
```diff
@@ -1545,3 +1545,49 @@ func TestJSONIndex_WithNeFilterAgainstNullField_ShouldFetchNonNullValues(t *test
...
+				Schema: `
+					type User {
+						name: String 
+						custom: JSON @index
+					}`,
+			},
+			testUtils.CreateDoc{
+				Doc: `{
+					"name": "John",
+					"custom": null
```
getify, line 1567: what happens if you just don't even specify the `custom` field -- like literally omit it! -- versus setting it to `null`? that's the case I was hitting.

should we have separate steps in this test for both "set to null explicitly" and "omit entirely"?

tests/integration/index/json_test.go
```diff
@@ -1545,3 +1545,49 @@ func TestJSONIndex_WithNeFilterAgainstNullField_ShouldFetchNonNullValues(t *test
 	}
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestJSONIndex_WithEqFilterAgainstNullField_ShouldFetchNullValues(t *testing.T) {
```
getify: will this test check both the path where you've done a `null` (or omission) of the optional field via inline, versus via external variables?

I know I've seen other cases where there's differences in behavior between inline and variables-based querying. Just want to make sure we're not diverging in this particular bug respect, and that we're testing both affirmatively, not just assuming.

tests/integration/index/json_test.go
AndrewSisley: I think this is the only `todo` from me, although I would prefer to have a second test for implicit-by-omission instead of including it in this test (suggestion, non-blocking).

AndrewSisley (approved):
LGTM, thanks Keenan - just the one testing todo from me - documentation might be handy if you are confident enough in your understanding of the original code to provide it, but IMO you shouldn't go out of your way to track all of that down if you didn't need to in order to fix the bug.

internal/db/index.go
nasdf: That check happens somewhere else. As far as I'm aware we don't currently support a `JSON!` field. I believe the decision to not include it was made because `null` is technically valid JSON.

internal/db/index.go
nasdf: The callback is used to generate the index values for the object. The code is fairly straight forward imo, but I can add a comment if you or others think it's needed.

tests/integration/index/json_test.go
nasdf: I've added another omitted version of the test

tests/integration/index/json_test.go
nasdf: I think it would be better to test that behavior with a test matrix option. Then we can run every test with variables and get more coverage with less changes.

nasdf: https://github.com/sourcenetwork/defradb/issues/4339

internal/db/index.go
getify: I suppose it's fine if others find it self-explanatory. There's a little less obvious context when reviewing PRs here versus being in the weeds of the code in your IDE. ;-)

PR #4342, opened 2026-01-14
fix: Index-based order for nested queries

### LINKED ISSUE #4321: Ordering on an indexed field of a one-to-many object yields the wrong result

islamaliev:
## Relevant issue(s)

Resolves #4321

## Description

Now secondary indexes are used for sub queries with `order` command.

## Tasks

## How has this been tested?

Integration tests

- MacOS

### DIFF
--- a/internal/planner/type_join.go
+++ b/internal/planner/type_join.go
@@ -612,13 +612,12 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
 
 	oldFetcher := r.primaryScan.fetcher
 	oldIndex := r.primaryScan.index
-	oldOrdering := r.primaryScan.ordering
 
 	r.primaryScan.index = findIndexByFieldName(r.primaryScan.col, r.relIDFieldDef.Name)
-	r.primaryScan.ordering = nil
 
 	canOrderByIndex := false
 
+	// if there is not index for filter, we try to find one for ordering
 	if !r.primaryScan.index.HasValue() {
 		var orderIndex immutable.Option[client.IndexDescription]
 		orderIndex, canOrderByIndex = r.findOrderingIndex()
@@ -642,7 +641,6 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
 
 	r.primaryScan.fetcher = oldFetcher
 	r.primaryScan.index = oldIndex
-	r.primaryScan.ordering = oldOrdering
 
 	return docs, nil
 }
--- a/internal/planner/type_join.go
+++ b/internal/planner/type_join.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sourcenetwork/defradb/client/request"
 	"github.com/sourcenetwork/defradb/internal/connor"
 	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/db/fetcher"
 	"github.com/sourcenetwork/defradb/internal/db/id"
 	"github.com/sourcenetwork/defradb/internal/keys"
 	"github.com/sourcenetwork/defradb/internal/planner/filter"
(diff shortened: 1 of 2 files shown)


tests/integration/index/query_with_relation_sub_order_test.go
```diff
@@ -0,0 +1,331 @@
...
+								{"title": "Book B1", "rating": 4.0},
+								{"title": "Book B2", "rating": 2.5},
+							},
+						},
+					},
+				},
+			},
+			testUtils.Request{
+				Request:  makeExplainQuery(req),
+				Asserter: testUtils.NewExplainAsserter().WithOrder().WithIndexFetches(8),
```
AndrewSisley, line 326: todo: Please document why the index is being fetched 8 times when returning 4 indexed documents.

internal/planner/type_join.go
```diff
@@ -604,8 +612,22 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
 
 	oldFetcher := r.primaryScan.fetcher
 	oldIndex := r.primaryScan.index
+	oldOrdering := r.primaryScan.ordering
 
 	r.primaryScan.index = findIndexByFieldName(r.primaryScan.col, r.relIDFieldDef.Name)
+	r.primaryScan.ordering = nil
```
AndrewSisley: todo: Please document why this is being explicitly set to nil

internal/planner/type_join.go
```diff
@@ -604,8 +612,22 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
...
+
+	canOrderByIndex := false
+
+	if !r.primaryScan.index.HasValue() {
```
AndrewSisley, line 621: todo: Please document why we are only (re)setting `ordering` if the original index is None.

internal/planner/type_join.go
```diff
@@ -604,8 +612,22 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
...
+		var orderIndex immutable.Option[client.IndexDescription]
+		orderIndex, canOrderByIndex = r.findOrderingIndex()
+		if canOrderByIndex {
+			r.primaryScan.index = orderIndex
+			r.primaryScan.ordering = r.ordering
```
AndrewSisley, line 626: todo: Please document (somewhere) why we wish to use the result of `r.findOrderingIndex` *only if* the `r.primaryScan.index` is None.

internal/planner/type_join.go
```diff
@@ -620,6 +642,7 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
 
 	r.primaryScan.fetcher = oldFetcher
 	r.primaryScan.index = oldIndex
+	r.primaryScan.ordering = oldOrdering
```
AndrewSisley: todo: Please document why we are overwriting all the effort that has been done in the new `if !r.primaryScan.index.HasValue() {` block here before returning from the function.

If the result of that block is to only be consumed by `primaryScan.initFetcher`, please also explain why!

AndrewSisley (approved):
LGTM, thanks for the fix Islam!

internal/planner/type_join.go
islamaliev: reassigninng of `oldOrdering` was just a precaution. It's not necessary. I removed it.

internal/planner/type_join.go
islamaliev: This wan just a precaution, but it's not necessary. I removed it.

internal/planner/type_join.go
islamaliev: already commented

internal/planner/type_join.go
islamaliev: done

tests/integration/index/query_with_relation_sub_order_test.go
islamaliev: done

PR #4350, opened 2026-01-15
fix: Prevent self-deadlock in abLock

### LINKED ISSUE #4311: Lock system can deadlock if write lock and RLockAll are held at the same time by same txn
It is impossible for users to do this at the moment, as RLockAll is dead code (I think), and write locks are only acquired when truncating.

It can however be introduced to users in the future if we are not paying attention.

This issue is linked to by a code-comment on a test, Ctrl+f to find it in the codebase.

Note: If something holding a `b` lock tries to acquire an `a` lock, we need it to actually hold both an `a` and a `b` lock, otherwise competing threads would remain free to acquire `b` locks whilst the original thread thinks it holds the `a` lock. Similarly, the original thread should not be able to acquire the `b` lock, until after all other threads holding their `a` locks have completed.

SharanRP:
## Relevant issue(s)

Resolves #4311

## Description

This PR fixes a potential deadlock in the lock system when the same transaction holds both [RLockAll](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock.go:53:0-55:1) and a write [Lock](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock_set.go:108:0-198:1) at the same time.

**The Problem:**
The [abLock](cci:2://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/ab_lock.go:26:0-36:1) mechanism uses two WaitGroups (A-group and B-group) where:
- [Lock](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock_set.go:108:0-198:1) (write) acquires an A-lock and waits for all B-locks to complete
- [RLockAll](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock.go:53:0-55:1) acquires a B-lock and waits for all A-locks to complete

When the same transaction holds a B-lock (via [RLockAll](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock.go:53:0-55:1)) and then tries to acquire an A-lock (via [Lock](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock_set.go:108:0-198:1)), it waits for its own B-lock to release → **deadlock**.

**The Fix:**
Added transaction ownership tracking to [abLock](cci:2://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/ab_lock.go:26:0-36:1). When a transaction already holds a B-lock and tries to acquire an A-lock (or vice versa), the blocking wait is skipped since the transaction owns both locks.

This is currently "dead code" as noted in the issue, but prevents future bugs if code changes are made without awareness of this constraint.

## Tasks

## How has this been tested?

- Updated existing test `TestLockSet_RLockAllAndLockSameTxn_Deadlocks` → [TestLockSet_RLockAllAndLockSameTxn_DoNotDeadlock](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock_set_test.go:239:0-248:1) to verify the fix
- Added new test [TestLockSet_LockAndRLockAllSameTxn_DoNotDeadlock](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/lock/lock_set_test.go:250:0-259:1) for reverse lock order
- Ran full lock package test suite: `go test -v ./internal/db/lock/...` (19 tests pass)
- Ran with race detector: `go test -race -v ./internal/db/lock/...` (all pass)

-  Linux

### DIFF
--- a/internal/db/lock/ab_lock.go
+++ b/internal/db/lock/ab_lock.go
@@ -17,39 +17,134 @@ import (
 // abLock provides a mechanism that allows two competing action sets to block
 // the other without blocking other members of its set.
 //
-// LockA will prevent LockB from being acquired, but it will not prevent
+// LockA will prevent LockB from being acquired by other transactions, but it will not prevent
 // other LockAs from being acquired, and vice versa.
+//
+// When a transaction already holds a lock of one type and tries to acquire the opposite type,
+// it will wait for OTHER transactions' locks to be released, but not its own. This prevents
+// self-deadlock while maintaining proper lock semantics - the transaction will hold both
+// lock types and properly block other transactions.
 //
 // Acquired locks are released on transaction close.
 type abLock struct {
-	aGroup sync.WaitGroup
-	bGroup sync.WaitGroup
+	mu       sync.Mutex
+	initOnce sync.Once
+
+	// cond is used to signal when lock counts change
+	cond *sync.Cond
+
+	// aCount tracks the total number of A-locks held (across all transactions)
+	aCount int
+	// bCount tracks the total number of B-locks held (across all transactions)
+	bCount int
+
+	// aHolders tracks how many A-locks each transaction holds (by txn ID)
+	aHolders map[uint64]int
+	// bHolders tracks how many B-locks each transaction holds (by txn ID)
+	bHolders map[uint64]int
 }
 
+// init initializes the abLock if needed (lazy initialization)
+func (l *abLock) init() {
+	l.initOnce.Do(func() {
+		l.cond = sync.NewCond(&l.mu)
+		l.aHolders = make(map[uint64]int)
+		l.bHolders = make(map[uint64]int)
+	})
+}
+
+// LockA acquires an A-lock for the given transaction.
+//
+// This will block until all B-locks held by OTHER transactions are released.
+// If this transaction already holds B-locks, those are excluded from the wait.
 func (l *abLock) LockA(txn txn) {
-	l.aGroup.Add(1)
+	l.init()
+	l.mu.Lock()
 
-	// We need to make sure this is only ever called once!  It is permitted to discard
-	// a transaction after it is commited whether it errors or not.
-	var once sync.Once
-	done := func() { once.Do(func() { l.aGroup.Done() }) }
-	txn.OnDiscard(done)
-	txn.OnError(done)
-	txn.OnSuccess(done)
+	txnID := txn.ID()
 
-	l.bGroup.Wait()
+	// Count of B-locks held by OTHER transactions
+	otherBCount := l.bCount - l.bHolders[txnID]
+
+	// Wait until all OTHER transactions' B-locks are released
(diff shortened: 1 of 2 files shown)


AndrewSisley (changes requested):
Hi Sharan, sorry about the bother but I do not think this solution does what it needs to do.

It does avoid a deadlock within `abLock`, but it does not offer the functionality that we really need it to provide.

If something holding a `b` lock tries to acquire an `a` lock, we need it to actually hold both an `a` and `b` lock, otherwise competing threads would remain free to acquire `b` locks whilst the original thread thinks it holds the `a` lock.  Similarly, the original thread should not be able to acquire the `b` lock, until after all other threads holding their `a` locks have completed.

I'll update the issue to make this clearer.

SharanRP:
Thanks for the feedback ,I misunderstood the requirement initially.

internal/db/lock/ab_lock.go
```diff
@@ -17,39 +17,132 @@ import (
...
+	if l.cond == nil {
+		l.cond = sync.NewCond(&l.mu)
+	}
+	if l.aHolders == nil {
+		l.aHolders = make(map[uint64]int)
+	}
+	if l.bHolders == nil {
+		l.bHolders = make(map[uint64]int)
+	}
 }
```
fredcarle: suggestion(non-blocking): You could optimize this slightly with a `sync.Once`:
```go
type abLock struct {
    mu       sync.Mutex
    initOnce sync.Once  // Add this

    cond *sync.Cond

    aCount int
    bCount int

    aHolders map[uint64]int
    bHolders map[uint64]int
}

func (l *abLock) init() {
    l.initOnce.Do(func() {
        l.cond = sync.NewCond(&l.mu)
        l.aHolders = make(map[uint64]int)
        l.bHolders = make(map[uint64]int)
    })
}
```
This avoids the repeated nil checks on every lock call. The sync.Once guarantees the initialization runs exactly once, even with concurrent callers.

However, there's a subtle issue: sync.Once.Do() has its own internal synchronization, so calling it while already holding l.mu works but is slightly redundant. A cleaner variant would be to
call init() before acquiring the mutex:

```go
func (l *abLock) LockA(txn txn) {
    l.init()      // Initialize once if needed (lock-free after first call)
    l.mu.Lock()
    // ... rest of the logic, no init() call needed here
}
```
This way, after the first call, sync.Once.Do becomes a fast no-op (just an atomic read) without holding the mutex.

fredcarle (commented):
Thanks for this contribution Sharan and I appreciate the fact that you took the feedback given by Andy and refactored accordingly. The changes look good from my point of view but I'll let @AndrewSisley give the final approval on this one.

AndrewSisley:
Hi Sharan, I'm closing this issue as we have decided to go with another solution, thanks for having a look at it :)

PR #4356, opened 2026-01-16
fix: Add missing return after successful merge execution

### LINKED ISSUE #4355: Merge retry loop missing return on success causes unnecessary repeated execution
The `Merge` function in `internal/db/merge.go`  is missing a `return nil` statement after successful execution. This causes the loop to continue iterating even when `executeMerge` succeeds, resulting in every merge operation running up to `MaxTxnRetries` times instead of once.

SharanRP:
## Relevant issue(s)

Resolves #4355

## Description

This PR fixes a bug in the merge retry loop where the function was missing a `return nil` statement after successful execution.

**The problem:** In `internal/db/merge.go` the `Merge()` function's retry loop  didn't return on success, causing the loop to continue iterating. This resulted in every merge operation running up to `MaxTxnRetries` times instead of exiting immediately after the first successful execution.

**Why it went unnoticed:**
- Merging is idempotent, so repeated execution produces the same result
- Existing tests call `executeMerge()` directly, bypassing the retry loop

**The fix:** Added `return nil` after the successful execution case to exit the loop immediately.

## Tasks

## How has this been tested?

Ran existing merge unit tests to verify the fix doesn't break current functionality

## Release Notes

* **Bug Fixes**
  * Merge operation now stops retrying once a merge completes successfully, reducing unnecessary retry cycles and improving merge completion time.
  * Behavior for conflict and error handling remains unchanged to preserve expected failure semantics.

### DIFF
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -72,6 +72,7 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 		if err != nil {
 			return err
 		}
+		return nil
 	}
 	return nil
 }


internal/db/merge.go
```diff
@@ -72,6 +72,7 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 		if err != nil {
 			return err
 		}
+		return nil
```
fredcarle, line 75: praise: Nice catch! I wonder I'd be curious to know quantitatively the wasted overhead that this caused.

fredcarle (approved):
Thanks for this Sharan.

PR #4365, opened 2026-01-19
feat(i): Remove blocks when truncating

### LINKED ISSUE #4324: Hard delete blocks when truncating

AndrewSisley:
## Relevant issue(s)

Resolves #4324

## Description

Removes the relevant document and collection blocks when truncating.

They are removed within the truncate transaction, as there is no guarantee that the transaction will successfully commit just because all the sub-operations within the truncate call succeed.

## How has this been tested?

This change is untested, beyond knowing that it doesn't error.

The removal of the blocks having already removed the heads is internal to Defra, if, we do not consider the blockstore public.  This PR is really just a storage optimisation IMO, and one that is too much of a pain to bother testing at the moment - for stores like badger, it's probably not even going to shrink the reserved disk space (at least for smaller data sets).

Please do let me know if you disagree with this!

## Summary by CodeRabbit

* **Refactor**
  * Reworked internal deletion flows for collections and documents to use dedicated, consistent deletion paths.
  * Switched to concrete key handling and construction instead of raw byte manipulation for more robust processing.
  * Deletions now run in chunked, recursive batches and remove both metadata and block data per chunk.
  * Preserves transactional locking semantics while improving reliability and efficiency.

### DIFF
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -272,7 +272,7 @@ func (c *collection) hardDeleteDocumentBlocks(
 		return err
 	}
 
-	keysToDelete := make([][]byte, 0, hardDeleteChunkSize)
+	keysToDelete := make([]keys.HeadstoreDocKey, 0, hardDeleteChunkSize)
 	// If there are more keys than we wish to load into memory at once, this will be set to
 	// true, and we'll continue the delete in another pass.
 	hasMore := true
@@ -287,7 +287,12 @@ func (c *collection) hardDeleteDocumentBlocks(
 			break
 		}
 
-		keysToDelete = append(keysToDelete, iter.Key())
+		key, err := keys.NewHeadstoreDocKey(string(iter.Key()))
+		if err != nil {
+			return errors.Join(err, iter.Close())
+		}
+
+		keysToDelete = append(keysToDelete, key)
 	}
 
 	err = iter.Close()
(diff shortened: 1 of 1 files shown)


fredcarle (changes requested):
Thanks for taking care of the blocks. 

todo: I would really like to see an integration test that queries the commits for a given collection and results in an empty set. I care more about those than the collection blocks.

AndrewSisley:
> todo: I would really like to see an integration test that queries the commits for a given collection and results in an empty set. I care more about those than the collection blocks.

This is already tested, our commit API relies on the headstore, it doesnt directly scan the blockstore, so when the heads were deleted (previous PR), the commits stopped being returned, e.g. `TestCollectionTruncateAdd_RemovesBlocks`

fredcarle:
> it doesnt directly scan the blockstore

It does scan the blockstore though. Thanks for sharing that test. It does exactly what I was looking for.

fredcarle (approved):
LGTM

internal/db/collection_truncate.go
```diff
@@ -287,26 +284,109 @@ func (c *collection) hardDeleteHeadstorePrefix(
...
+	err = iter.Close()
+	if err != nil {
+		return err
+	}
+
+	blockstore := txn.Blockstore()
+
+	for _, key := range keysToDelete {
+		// Not all store implementations support mutations whilst iterating, so whilst it would
+		// be simpler and probably more efficient to delete whilst iterating, it would not work
```
shahzadlone, line 304: Humor: That's a lot of `whilst`

internal/db/collection_truncate.go
```diff
@@ -287,26 +284,109 @@ func (c *collection) hardDeleteHeadstorePrefix(
...
+	iter, err := headstore.Iterator(ctx, corekv.IterOptions{
+		Prefix:   prefix.Bytes(),
+		KeysOnly: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	keysToDelete := make([]keys.HeadstoreColKey, 0, hardDeleteChunkSize)
+	// If there are more keys than we wish to load into memory at once, this will be set to
```
shahzadlone, line 344: Praise: thanks for this comment

shahzadlone (approved):
LGTM

Issue #4351, opened 2026-01-16
Performance Issue: Slow `_count` on query


dvncan:
**Describe the problem**
Count is very slow >100k docs

**To Reproduce**
http://35.239.160.177:444/ 

Metrics at time of test:
```json
  "metrics": {
    "attestations_created": 1547828,
    "attestation_errors": 0,
    "signature_verifications": 0,
    "signature_failures": 0,
    "documents_received": 1548224,
    "documents_processed": 1547828,
    "documents_dropped": 0,
    "blocks_processed": 5161,
    "transactions_processed": 444196,
    "logs_processed": 1058915,
    "access_lists_processed": 39952,
```

100k docs : `9s`
```
{
  Ethereum__Mainnet__Transaction(limit:100000){
    _docID
  }
}
```

Control - all transaction doc ids: `3m`
```
 {
    Ethereum__Mainnet__Transaction(){
        _docID
      }
}
```

count 100k docs: 1.6min
```graphql
{
  _count(Ethereum__Mainnet__AttestationRecord:{
    limit:100000
		filter:{doc_type:{_eq:"Ethereum__Mainnet__Transaction"}}    
  })
}

```

Total count: 444196
**Basic Count Response**: `5min`
```graphql
{
  _count(Ethereum__Mainnet__Transaction:{})
}
```

**Expected behavior**
Faster count

**Platform**
### Basic information
Ubuntu 22.04
e2-standard-16 (16 vCPUs, 64 GB Memory)

PR #4371, opened 2026-01-20
fix: Prevent wrapper from defaulting to all fields

### LINKED ISSUE #4380: Prevent fetcher from defaulting to all fields if none are given.

SharanRP:
## Relevant issue(s)

Resolves #4380 
Part of #4351

## Description

This PR fixes a performance regression where the `wrappingFetcher` would default to fetching *all* fields from the schema if the requested field list was empty. This behavior negatively impacted aggregate queries like `_count`, which request zero fields but were incurring the cost of a full document fetch.

Additionally, this PR updates `scanNode` to skip the `DecodeToDoc` step when no fields are requested.

## Tasks

## How has this been tested?

Existing integration tests related to `_count` passed successfully.

- Linux


internal/db/fetcher/wrapper.go
```diff
@@ -124,10 +124,6 @@ func (f *wrappingFetcher) Start(ctx context.Context, prefixes ...keys.Walkable)
 		return err
 	}
 
-	if len(f.fields) == 0 {
-		f.fields = f.col.Version().Fields
-	}
```
fredcarle, line 129: todo: This is a good change but it requires a small tweak for it to not also bring in a regression. There are a number of internal functions that use `collection.Get` to retrieve a single document and when one uses that function, they expect the full document to be returned. Same goes for a user that wants to use this public function. I think that the only thing that needs to change for this to work as expected is `./internal/db/collection_get.go#ln58` replacing nil for `c.Version().Fields`.

fredcarle (changes requested):
Thanks again for another nice PR. I have one todo for you before I can approve the PR. Let me know if you have any questions.

internal/db/fetcher/wrapper.go
SharanRP: Good catch! You're right, `collection.Get` was relying on that default behavior. I've updated `collection.Get` to explicitly request all fields `(c.Version().Fields)` instead of passing `nil`, so it works correctly now without regressing. Pushed the fix!

AndrewSisley:
Hey Sharan,

Sorry again about the bother here, but we agreed to close this for now as the conflicts are significant in comparison to the complexity of the fix.  Please feel very free to re-open this should you wish to resolve the conflicts yourself.

PR #4373, opened 2026-01-20
feat: Improve NotAuthorizedToPerformOperation errors

### LINKED ISSUE #4108: Improve NotAuthorizedToPerformOperation error by appending the missing permission.
TLDR: We want to have in addition to just "not authorized to perform operation" to also append the missing permission to the user.

From Discord Discussion:
> We discussed a short-term (pre v1) vs long-term (post v1) approach in tackling operations that might be dependent on other operations that might require more permissions. Here we discussed the following approached:
  `Short Term`:
    - `simple`: Fail at the call that requires that additional permission with the `no auth` error (perhaps improve the log/error by appending that missing permission that is needed by this identity).

Discord Discussion Summary: https://discord.com/channels/1374835078781468722/1422306930348523560/1423424575101730979

ChrisBQu:
## Relevant issue(s)

Resolves #4108 

## Description

The `NotAuthorizedToPerformOperation` error is not especially descriptive or helpful to users. It doesn't fully explain what was why their action failed, because it notably doesn't tell them what permissions they lack. This PR addresses that. Here is an example of the before and after for an error:

Before:

_not authorized to perform operation_

After:

_not authorized to perform operation: missing permission collection-get_

The larger crux of the PR comes from changing all associated tests to check for a richer error message. Sometimes, there is a difference in behavior between clients and the other clients, and sometimes, the error may not be what a user might expect. This is something that was discussed, and which will be addressed when granular NAC permissions are implemented. For now, any such case is marked with a todo comment, and a link to the associated Git issue.

## Tasks

## How has this been tested?

- WSL


client/errors.go
```diff
@@ -269,3 +269,25 @@ func NewErrEmbeddingFieldEmbedding(fieldName string) error {
 func NewErrNotFound(kv errors.KV) error {
 	return errors.New(errNotFound, kv)
 }
+
+type ErrNotAuthorizedMissingPermission struct {
```
AndrewSisley: question: Why have you done this, instead of using `errors.KV`?

client/errors.go
```diff
@@ -269,3 +269,25 @@ func NewErrEmbeddingFieldEmbedding(fieldName string) error {
...
+func (e *ErrNotAuthorizedMissingPermission) Error() string {
+	return fmt.Sprintf(
+		"not authorized to perform operation: missing permission %q",
+		e.Permission,
+	)
+}
+
+// The following allows errors.Is(err, ErrNotAuthorizedToPerformOperation)
+func (e *ErrNotAuthorizedMissingPermission) Is(target error) bool {
+	return target == ErrNotAuthorizedToPerformOperation
```
AndrewSisley: question: Our tests only assert that the expected sub-string is returned, they never actually test `errors.Is`.

Are you sure this works?  To me it looks like it is just checking the memory ref, and will not return true for `NewErrNotAuthorizedMissingPermission().Is(NewErrNotAuthorizedMissingPermission())`

ChrisBQu: Yeah, I did notice that we didn't actually use errors.Is in the tests. Which was convenient here, I guess. But you may be right this doesn't do what I wanted it to do. I put it here as a convenience for the future but... I think I'm leaning towards just removing it. If we need such functionality later, we can introduce it again and ensure that it works.

client/errors.go
ChrisBQu: I don't have a good answer -- I wrapped it the first way I thought to do so, to confirm that it worked. But I think you're right, this is the way to go. I've made the change.

I don't know why I didn't think to do it using errors.KV to begin with -- that's what I used for the Cbindings...

internal/db/acp/check.go
```diff
@@ -236,5 +236,5 @@ func CheckNodeOperationAccess(
 		return nil
 	}
 
-	return client.ErrNotAuthorizedToPerformOperation
+	return client.NewErrNotAuthorizedMissingPermission(permission.String())
```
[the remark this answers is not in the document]
ChrisBQu: Good catch. Thank you, coderabbitai.

client/errors.go
```diff
@@ -269,3 +269,7 @@ func NewErrEmbeddingFieldEmbedding(fieldName string) error {
 func NewErrNotFound(kv errors.KV) error {
 	return errors.New(errNotFound, kv)
 }
+
+func NewErrNotAuthorizedMissingPermission(permission string) error {
+	return errors.New(ErrNotAuthorizedToPerformOperation.Error(), errors.NewKV("Permission", permission))
```
fredcarle: suggestion: We have the bad habit of using strings with `errors.New` in our error constructors. This has drawbacks when it come to use `errors.Is` with the constructor created errors. I'll open a ticket to track that this needs to be resolved globally but in this specific case, I suggest you change to:
```go
errors.WithStack(ErrNotAuthorizedToPerformOperation, errors.NewKV("Permission", permission))
```

fredcarle (approved):
LGTM with one suggestion

client/errors.go
AndrewSisley: What are the drawbacks?  Are you worried about performance?

ChrisBQu: I am okay to make this change, but I am also interested in what the drawbacks are.

fredcarle: I tried to back up my statement with facts and I proved myself wrong. `errors.WithStack` actually just does `withStack(err.Error())` internally which is the equivalent of what you currently have. I think this came from a previous mental model I had of our errors package. Sorry for the bother.

shahzadlone (changes requested):
Thanks for doing the suggestion we discussed on discord.

One thing I am still not seeing here is that no tests were updated (as they are still passing by matching the subset of the string you just changed.)

Some other reasons I wanted to do this in addition to improve errors for users (as marked in issue and linked discussion):
1) Ensure the permission that causes the auth error is the exact one we expect (asserted in the test).
2) Would eventually give us a better visualization of the NAC permission dependency graph.

todo: Actually assert the target missing permission in all the tests that use this. I suggest to use a helper util asserting function that takes a `NodeResourcePermission` type from `acp/types/types.go` and formats the string to assert like we do for some other tests.

internal/db/acp/check.go
```diff
@@ -155,7 +155,7 @@ func CheckDocAccessWithIdentityFunc(
 	return hasAccess, nil
 }
 
-// CheckNodeOperationAccess returns an [client.ErrNotAuthorizedToPerformOperation]
+// CheckNodeOperationAccess returns an [client.NewErrNotAuthorizedMissingPermission]
```
shahzadlone: Todo: i don't understand why the return type in documentation is changed to a new function you made

tests/integration/acp/nac/collection_patch_test.go
```diff
@@ -86,6 +138,56 @@ func TestNAC_GatesCollectionPatch_NoIdentity_NotAuthorizedError(t *testing.T) {
 
 func TestNAC_GatesCollectionPatch_WrongIdentity_NotAuthorizedError(t *testing.T) {
 	test := testUtils.TestCase{
+		// TO DO: Investigate and test this behavior across all client types when implementing granular NAC permissions.
```
shahzadlone: nitpick: I do prefer `todo` or `TODO` OVER `TO DO`

tests/integration/acp/nac/collection_patch_test.go
```diff
@@ -86,6 +138,56 @@ func TestNAC_GatesCollectionPatch_NoIdentity_NotAuthorizedError(t *testing.T) {
...
+				`,
+				ExpectedError: testUtils.FormatExpectedErrorWithPermission("collection-patch"),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestNAC_GatesCollectionPatch_WrongIdentity_NotAuthorizedError_CLIClient(t *testing.T) {
```
shahzadlone: Todo: according to our styling rules the CLIClient should come before `_NotAuthorizedError` as the last part should be the expected behavior

tests/integration/acp/nac/collection_patch_test.go
```diff
@@ -110,7 +212,7 @@ func TestNAC_GatesCollectionPatch_WrongIdentity_NotAuthorizedError(t *testing.T)
 						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} }
 					]
 				`,
-				ExpectedError: "not authorized to perform operation",
+				ExpectedError: testUtils.FormatExpectedErrorWithPermission("collection-get"),
```
shahzadlone: Suggestion: why hard coding i thought we discussed to use the acp types defined

internal/db/acp/check.go
ChrisBQu: Code rabbit pointed that one out, and I thought it made sense, so I changed it. We are not returning the other error type anymore. Do you think the intention of this comment is unclear now?

fredcarle: I think `client. ErrNotAuthorizedToPerformOperation` is clearer and more accurate.

tests/integration/acp/nac/collection_patch_test.go
ChrisBQu: I am beyond confused by this. I do know we discussed it, but I thought we came to the *opposite* conclusion. Please be very clear about how you think it should be written, because I feel so far we may have been somehow talking past one another.

I will take one new guess(?)

In `acp/types/types.go` there is an array of strings defined as `RequiredResourcePermissionsForNode`. And there is a list of constants, such as `NodeCollectionGetPerm`. Do you want me to use `RequiredResourcePermissionsForNode[NodeCollectionGetPerm]` to refer to it?

tests/integration/acp/nac/collection_patch_test.go
ChrisBQu: Will change!

tests/integration/acp/nac/collection_patch_test.go
ChrisBQu: Will change!

internal/db/acp/check.go
ChrisBQu: I will change it back.

tests/integration/acp/nac/collection_patch_test.go
shahzadlone: You just call the string method on the types and it will give you the string i believe.

ChrisBQu: Shahzad and I just had a short discussion on Discord about this, and it was all just a misunderstanding. Will be fixing this up. :)

client/errors.go
```diff
@@ -269,3 +269,7 @@ func NewErrEmbeddingFieldEmbedding(fieldName string) error {
 func NewErrNotFound(kv errors.KV) error {
 	return errors.New(errNotFound, kv)
 }
+
+func NewErrNotAuthorizedMissingPermission(permission string) error {
+	return errors.WithStack(ErrNotAuthorizedToPerformOperation, errors.NewKV("Permission", permission))
+}
```
shahzadlone: question: What is the benefit of using `WithStack` here rather than following the same pattern like `errors.New` that other errors use?

todo: The convention we use (can look at others in this file) is to just add `New` before the error name. Also think that the original `ErrNotAuthorizedToPerformOperation` name is more clearer.

So it should be:

```suggestion
func NewErrNotAuthorizedToPerformOperation(permission string) error {
	return errors.WithStack(ErrNotAuthorizedToPerformOperation, errors.NewKV("Permission", permission))
}
```

client/errors.go
```diff
@@ -269,3 +269,7 @@ func NewErrEmbeddingFieldEmbedding(fieldName string) error {
 func NewErrNotFound(kv errors.KV) error {
 	return errors.New(errNotFound, kv)
 }
+
+func NewErrNotAuthorizedMissingPermission(permission string) error {
```
shahzadlone: suggestion: Why not make this function take in the direct permission type like the `NewErrInvalidTypeForEmbedding` helper function this way the caller wouldn't have to worry about putting it in a string format by doing this `permission.String()`, that can be done within this function body.

internal/db/acp/check.go
shahzadlone: suggestion: If you do the other suggestion in `errors.go`

This becomes:

```suggestion
	return client.NewErrNotAuthorizedMissingPermission(permission)
```

tests/integration/acp/nac/relation_admin/collection_get_by_name_test.go
```diff
@@ -45,7 +46,7 @@ func TestNAC_AdminRelation_CanCollectionGetByName(t *testing.T) {
 					Name:            immutable.Some("Users"),
 					IncludeInactive: immutable.Some(false),
 				},
-				ExpectedError: "not authorized to perform operation",
+				ExpectedError: testUtils.FormatExpectedErrorWithPermission(acpTypes.NodeCollectionGetPerm.String()),
```
shahzadlone: todo: Why doesn't the formatting helper function do all the formatting? It's odd to have to do the `.String()` separately out of the format function. Please do the `.String()` inside the `testUtils.FormatExpectedErrorWithPermission` function.

Then these will become:

```suggestion
				ExpectedError: testUtils.FormatExpectedErrorWithPermission(acpTypes.NodeCollectionGetPerm),
```

Should be quick and easy to change all using a regex replace.

tests/integration/acp/nac/collection_patch_test.go
```diff
@@ -52,6 +55,56 @@ func TestNAC_GatesCollectionPatch_AuthorizedIdentity_AllowAccess(t *testing.T) {
 
 func TestNAC_GatesCollectionPatch_NoIdentity_NotAuthorizedError(t *testing.T) {
 	test := testUtils.TestCase{
+		// TO DO: Investigate and test this behavior across all client types when implementing granular NAC permissions.
+		// See: https://github.com/sourcenetwork/defradb/issues/4383
```
shahzadlone: todo:

```suggestion
		// todo: Investigate and test this behavior across all client types when implementing granular NAC permissions.
		// See: https://github.com/sourcenetwork/defradb/issues/4383
```

tests/integration/acp/nac/collection_truncate_test.go
```diff
@@ -46,7 +49,55 @@ func TestNAC_GatesCollectionTruncate_AuthorizedIdentity_AllowAccess(t *testing.T
...
+				CollectionIndex: 0,
+				ExpectedError:   testUtils.FormatExpectedErrorWithPermission(acpTypes.NodeCollectionTruncatePerm.String()),
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestNAC_GatesCollectionTruncate_NoIdentity_CLIandHTTPClient_NotAuthorizedError(t *testing.T) {
```
shahzadlone: suggestion: Also add the `C` in the test name as it's missing.
```suggestion
func TestNAC_GatesCollectionTruncate_NoIdentity_CLIandCandHTTPClient_NotAuthorizedError(t *testing.T) {
```

tests/integration/acp/nac/document_read_test.go
```diff
@@ -53,6 +54,7 @@ func TestNAC_GatesDocumentRead_AuthorizedIdentity_AllowAccess(t *testing.T) {
 }
 
 func TestNAC_GatesDocumentRead_NoIdentity_NotAuthorizedError(t *testing.T) {
+	// todo: Investigate and test this behavior across all client types when implementing granular NAC permissions.
```
shahzadlone, line 57: question/todo: Missed to mark the issue?

tests/integration/acp/nac/p2p_document_delete_test.go
```diff
@@ -27,9 +28,7 @@ func TestNAC_GatesP2PDocumentDelete_AuthorizedIdentity_AllowAccess(t *testing.T)
 				state.HTTPClientType,
 				state.CLIClientType,
 				state.GoClientType,
-				// TODO: https://github.com/sourcenetwork/defradb/issues/4091
-				// We have to fix the c-binding identity passing issue to support c-client.
-				// state.CClientType,
```
shahzadlone, line 32: praise: Thanks for enabling these and removing the resolved issue

tests/integration/utils.go
```diff
@@ -1940,3 +1940,7 @@ func performVerifySignatureAction(s *state.State, action VerifyBlockSignature) {
 		}
 	}
 }
+
+func FormatExpectedErrorWithPermission(permission string) string {
+	return fmt.Sprintf("%s. Permission: %s", client.ErrNotAuthorizedToPerformOperation, permission)
```
shahzadlone: suggestion: Like the other suggestion, take in the type not the `string`, and then call the `permission.String()`.


```suggestion
	return fmt.Sprintf("%s. Permission: %s", client.ErrNotAuthorizedToPerformOperation, permission.String())
```

shahzadlone (approved):
Thanks a bunch for spending the time to do all the requested changes.

Left some minor todos and suggestions.

Leaving an LGTM but please resolve the comments before merge.

client/errors.go
fredcarle: The benefit is that we don't have to declare a string `errNotAuthorizedToPerformOperation` and the error `ErrNotAuthorizedToPerformOperation` plus `NewErrNotAuthorizedToPerformOperation`. We can use the error directly. I think it's much cleaner and simplifies the `errors.go` files.

tests/integration/utils.go
ChrisBQu: I have made this change.

internal/db/acp/check.go
ChrisBQu: Done.

client/errors.go
ChrisBQu: Done.

PR #4379, opened 2026-01-21
feat: Refresh materialized views on create

### LINKED ISSUE #4161: Refresh materialized views on create
From build week feedback: https://discord.com/channels/1374835078781468722/1435676951988666378

> Creating views as materialized by default and not immediately refreshing them is very confusing

AndrewSisley:
## Relevant issue(s)

Resolves #4161

## Description

Refresh materialized views on create.

Also reworks the view-type multiplier, as it was only working when the test did not already declare the `@materialized` directive - and thanks to a change a while ago which flipped the defaults, most of our tests were explicitly declaring the directive, rendering the multiplier useless.

### DIFF
--- a/internal/db/view.go
+++ b/internal/db/view.go
@@ -95,6 +95,17 @@ func (db *DB) addView(
 		return nil, err
 	}
 
+	for _, view := range returnDescriptions {
+		if view.Query.HasValue() && view.IsMaterialized {
+			err := db.refreshViews(ctx, client.CollectionFetchOptions{
+				VersionID: immutable.Some(view.VersionID),
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	return returnDescriptions, nil
 }
 
--- a/tests/integration/view/simple/materialized_test.go
+++ b/tests/integration/view/simple/materialized_test.go
@@ -97,7 +97,6 @@ func TestView_SimpleMaterialized_RefreshesAfterEarlierRefresh(t *testing.T) {
 					}
 				`,
 			},
-			&action.RefreshViews{},
 			testUtils.CreateDoc{
 				Doc: `{
 					"name":	"Fred"
--- a/tests/integration/view/simple/materialized_test.go
+++ b/tests/integration/view/simple/materialized_test.go
@@ -73,10 +73,7 @@ func TestView_SimpleMaterialized_DoesNotAutoUpdateOnViewCreate(t *testing.T) {
 func TestView_SimpleMaterialized_DoesNotAutoUpdate(t *testing.T) {
 	test := testUtils.TestCase{
 		SupportedViewTypes: immutable.Some([]testUtils.ViewType{
-			// As the MaterializedViewType will auto refresh views immediately prior
-			// to executing requests, this test of materialized views actually only
-			// supports running with the CachelessViewType flag.
-			testUtils.CachelessViewType,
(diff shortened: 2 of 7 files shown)


internal/db/view.go
```diff
@@ -95,6 +95,17 @@ func (db *DB) addView(
 		return nil, err
 	}
 
+	for _, view := range returnDescriptions {
+		if view.Query.HasValue() && view.IsMaterialized {
+			err := db.refreshViews(ctx, client.CollectionFetchOptions{
```
fredcarle, line 100: question: This could be a long running function. Do we really want to call this synchronously on view add?

thought: It might be worth starting a background process for this.

tests/integration/view/simple/materialized_test.go
```diff
@@ -52,16 +97,34 @@ func TestView_SimpleMaterialized_DoesNotAutoUpdateOnViewCreate(t *testing.T) {
 					}
 				`,
 			},
+			&action.RefreshViews{},
```
fredcarle: question: If we make it so the view is refreshed on create, this call is a little redondant. Why is it important to have it in this test?

AndrewSisley: I agree, nice spot :)

- [x] Remove extra refresh

internal/db/view.go
AndrewSisley: I think that aysnc is the user's responsibility, not ours.

I thought about adding a param that allows them to *not* refresh the view, but decided against including that in v1.

fredcarle: This is also the behaviour of other databases but it might be worth documenting publicly.

tests/integration/view/simple/materialized_test.go
```diff
@@ -19,13 +19,58 @@ import (
...
+				Results: map[string]any{
+					// Even though UserView was created after the document was created, the results are
+					// present because the view will automatically referesh upon its creation.
+					"UserView": []map[string]any{
+						{
+							"name": "John",
+						},
+					},
+				},
+			},
```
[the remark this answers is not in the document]
fredcarle: This seems like a logical suggestion. What made you decide to just resolve it?

tests/integration/view/simple/materialized_test.go
```diff
@@ -19,13 +19,58 @@ import (
...
+						name
+					}
+				`,
+				SDL: `
+					type UserView {
+						name: String
+					}
+				`,
+			},
+			testUtils.Request{
```
fredcarle, line 49: todo: As coderabbit suggested, please make sure that `DoNotRefreshViews` is set to true.

fredcarle (approved):
LGTM with one todo to resolve before merge.

tests/integration/view/simple/materialized_test.go
AndrewSisley: > This seems like a logical suggestion. What made you decide to just resolve it?

Replying here to avoid coderabbit spam.

I misread coderabbit.  I don't pay it a whole lot of attention, as it seems to be wrong more frequently than it is right, and is very spammy.  In a sense, it has already begun training me to ignore whatever it says.

This time it was right :)  Thanks (Fred) for flagging.

- [x] Add DoNotRefreshViews with documentation

internal/db/view.go
AndrewSisley: Documentation updated

PR #4293, opened 2025-12-19
feat: Add support for `_eq` and  `_neq` to inline scalar arrays

jsimnz:
## Relevant issue(s)

Resolves #4280 

## Description

Adds support for `_eq` and `_ne` to inline arrays and nonnull inline arrays.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

Notable expansion of the inline_array integration tests to cover all current inline array scalar types with both `_eq` and `_ne`.

Specify the platform(s) on which this was tested:
- Ubuntu (WSL2)

PR #4385, opened 2026-01-21
feat: Add DateTime inline array support

### LINKED ISSUE #4375: DateTime inline scalar arrays
Support both nillable and non-nillable DateTime arrays.

I don't believe there is any blockers to this feature, was just a potential oversight when implementing either DateTime scalars or inline arrays.

Was surfaced in this comment here: https://github.com/sourcenetwork/defradb/pull/4293#discussion_r2707006476

SharanRP:
## Relevant issue(s)

Resolves #4375 

## Description

Add support for DateTime inline arrays (`[DateTime!]` and `[DateTime]`) as noted in PR #4293.

## Limitation

Nillable DateTime arrays ([DateTime]) have known CBOR serialization complexities with immutable.Option[time.Time] values that may require additional handling in certain edge cases.

## Tasks

## How has this been tested?

Added integration tests for DateTime arrays. All existing tests pass.

-  Linux


tests/integration/query/inline_array/with_datetime_test.go
```diff
@@ -0,0 +1,148 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	executeTestCaseDateTime(t, test)
+}
+
+func mustParseTime(s string) time.Time {
```
fredcarle: todo: Please use `testUtils.MustParseTime` instead of this function.

internal/core/encoding.go
```diff
@@ -241,6 +258,34 @@ func convertToFloat32(propertyName string, untypedValue any) (float32, error) {
...
+					return t, nil
+				}
+				if t, err := time.Parse(time.RFC3339Nano, s); err == nil {
+					return t, nil
+				}
+			}
+			if t, ok := v.(time.Time); ok {
+				return t, nil
+			}
+		}
```
fredcarle: question: Have you tested that this is can actually happen? I'm asking because we specify the `TimeRFC3339Nano` option for CBOR encoding and thus the map for should never happen. As such, I'm fairly certain that this switch case is dead code.

fredcarle (changes requested):
Hi Sharan. Thanks for this PR.

I've have a couple comments that need to be resolved before I can approve plus I would like to see a few more tests written.
- Tests for nillable arrays with null elements.
- Filter tests (_any, _all, _none) like other array types
- Mutation/update tests
- Index query tests (indexing is enabled but not tested)
- Error case tests (malformed datetime strings)

Let me know if you have any questions :)

internal/core/encoding.go
SharanRP: Ah nice catch,I added that case speculatively while debugging, but it was indeed dead code. I've removed it

SharanRP:
@fredcarle  Thanks for the feedback, I've addressed all the requested changes.

fredcarle:
Covered by #4915

PR #4391, opened 2026-01-22
fix(i): Delete entire block dag on truncate

### LINKED ISSUE #4377: Truncate only deletes the head blocks
I forgot to walk them back.  We are currently orphaning a lot of blocks.

AndrewSisley:
## Relevant issue(s)

Resolves #4377

## Description

Deletes the entire block dag on truncate, instead of just the head blocks.

## How has this been tested?

As was in the PR that introduced this bug and deleted the head blocks, this is a performance change with no other user-visible behaviour and is untested besides a couple of sanity-checking in-development printlns, and the absence of errors being thrown whilst this code is being executed.

### DIFF
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -438,13 +438,10 @@ func deleteBlocks(ctx context.Context, head cid.Cid) error {
 
 		switch {
 		case decodedBlock.Delta.IsField():
-			// At the time of writing, field blocks do not have any heads that will not already be linked
-			// to by other DAGs being deleted, so we have decided that the compute that we will save by not
-			// trying to `Get` them is worth the risk of potentially missing blocks in the future should
-			// this change.
-			for _, link := range decodedBlock.Links {
-				toDelete[link.Cid] = struct{}{}
-			}
+			// At the time of writing, field blocks do not have any links besides Encryption and Signature,
+			// that will not already be linked to by other DAGs being deleted, so we have decided that the
+			// compute that we will save by not trying to `Get` them is worth the risk of potentially missing
+			// blocks in the future should this change.
 
 		default:
 			for _, link := range decodedBlock.AllLinks() {
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -436,8 +436,20 @@ func deleteBlocks(ctx context.Context, head cid.Cid) error {
 			return err
 		}
 
-		for _, link := range decodedBlock.AllLinks() {
-			toDelete[link.Cid] = struct{}{}
+		switch {
+		case decodedBlock.Delta.IsField():
+			// At the time of writing, field blocks do not have any heads that will not already be linked
+			// to by other DAGs being deleted, so we have decided that the compute that we will save by not
+			// trying to `Get` them is worth the risk of potentially missing blocks in the future should
+			// this change.
+			for _, link := range decodedBlock.Links {
+				toDelete[link.Cid] = struct{}{}
+			}
+
+		default:
+			for _, link := range decodedBlock.AllLinks() {
+				toDelete[link.Cid] = struct{}{}
+			}
 		}
 	}
 
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -13,9 +13,13 @@ package db
 import (
 	"context"
 
+	"github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
+
 	"github.com/sourcenetwork/corekv"
 	acpTypes "github.com/sourcenetwork/defradb/acp/types"
 	"github.com/sourcenetwork/defradb/errors"
+	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/id"
 	"github.com/sourcenetwork/defradb/internal/keys"
@@ -297,8 +301,6 @@ func (c *collection) hardDeleteDocumentBlocks(
 		return err
 	}
 
-	blockstore := txn.Blockstore()
-
 	for _, key := range keysToDelete {
 		// Not all store implementations support mutations whilst iterating, so whilst it would
 		// be simpler and probably more efficient to delete whilst iterating, it would not work
@@ -308,7 +310,7 @@ func (c *collection) hardDeleteDocumentBlocks(
 			return err
 		}
 
-		err = blockstore.DeleteBlock(ctx, key.Cid)
+		err = deleteBlocks(ctx, key.Cid)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


internal/db/collection_truncate.go
```diff
@@ -391,3 +391,55 @@ func (c *collection) hardDeleteCollectionBlocks(
 
 	return nil
 }
+
+// deleteBlocks deletes the block of the given cid and all the blocks it links to, if
+// a block with this cid is found.
+//
+// If the block is not found, it will not error.
+func deleteBlocks(ctx context.Context, head cid.Cid) error {
```
fredcarle, line 399: todo: There is a way to make this more efficient. You should only fetch the heads and directly delete the links. This way you're only walking the composites and saving a lot of work. For branchable you would have to walk both heads (other collection blocks) and links (document composites) though. But still overall much less work.

fredcarle (changes requested):
I'd like to reduce the amount of work done here before approval.

internal/db/collection_truncate.go
```diff
@@ -391,3 +391,55 @@ func (c *collection) hardDeleteCollectionBlocks(
...
+
+		currentBlock, err := blockstore.Get(ctx, currentBlockCid)
+		if errors.Is(err, ipld.ErrNotFound{}) {
+			// We are looping through the links in a simple way that may result in us
+			// attempting to delete blocks we have already deleted, this can include
+			// blocks deleted by walking the dag pointed-to from another headstore key
+			// (another call to `deleteBlocks`).
+			//
+			// If we encounter such a block, we can skip over the error and continue.
+			continue
```
shahzadlone, line 423: question/thought: Is that the only case we have to worry about? Just think out loud if there is an edge case for skipping on a ErrNotFound

internal/db/collection_truncate.go
AndrewSisley: That would only delete the top level blocks no?

If a doc is updated the previous head is removed from the headstore.  We need to walk the DAG in order to delete all the blocks.

internal/db/collection_truncate.go
AndrewSisley: I don't think so, and if there is, I don't think we really lose anything by not knowing about it - we try to delete a block that wasn't there and continue.

internal/db/collection_truncate.go
fredcarle: The blocks have mainly 2 link fields
```go
type Block struct {
  Heads []cidlink.Link
  Links []DAGLink
  ...
}
```
Walking the DAG (from a document composite) through the `Heads` is enough to delete all the document blocks. However, I realize we still need to load the link blocks to make sure we delete the `Encryption` and `Signature` blocks if they are different than that of the composite. But we don't need to load the `Encryption` and `Signature` blocks and neither de we need to walk from the link blocks of the fields. 

image

image

AndrewSisley: Clarified on discord, resuming conversation here.

I am a bit uncomfortable about adding too much logic here as it risks us missing blocks in the future if the linking ever changes.  This is particularly risky with our current lack of tests here (we don't test that specific blocks have been removed).

I'm also slightly hesitant because we don't have any performance tests for this, so whilst skipping over some links will almost certainly be faster, we haven't actually measured the relative impact of that on the `Truncate` call.

I think I'm going to avoid whitelisting block kinds and their links, and instead blacklist field level blocks, this should hopefully reduce the risk of not deleting stuff in the future (there will still remain some risk to field links though, I'll document).

- [x] Don't delete field head links

fredcarle: > and instead blacklist field level blocks

I'm ok with that 👍

fredcarle: Just because you didn't mention it in you last comment and I want to make sure we don't forget them, I'm gonna add a todo.

todo: Make sure you delete the Encryption and Signature blocks for both composites and fields.

AndrewSisley: Cheers, I thought they were in `Links`

- [x] Delete enc and sig blocks

internal/db/collection_truncate.go
```diff
@@ -379,7 +379,7 @@ func (c *collection) hardDeleteCollectionBlocks(
 			return err
 		}
 
-		err = blockstore.DeleteBlock(ctx, key.Cid)
+		err = deleteBlocks(ctx, key.Cid)
```
shahzadlone, line 382: question: This line is never hit by tests?

internal/db/collection_truncate.go
```diff
@@ -391,3 +391,75 @@ func (c *collection) hardDeleteCollectionBlocks(
...
+		}
+
+		switch {
+		case decodedBlock.Delta.IsField():
+			// At the time of writing, field blocks do not have any heads that will not already be linked
+			// to by other DAGs being deleted, so we have decided that the compute that we will save by not
+			// trying to `Get` them is worth the risk of potentially missing blocks in the future should
+			// this change.
+			for _, link := range decodedBlock.Links {
+				toDelete[link.Cid] = struct{}{}
```
shahzadlone: nitpick: Is this dead/dormant code right now?

internal/db/collection_truncate.go
AndrewSisley: No, I think there is a test gap for branchable collections.  They don't really do anything special that would break, but we should have a sanity check at least.

- [x] Close branchable test gap

internal/db/collection_truncate.go
AndrewSisley: No, why do you ask?

AndrewSisley:
@coderabbitai configuration

internal/db/collection_truncate.go
fredcarle: This is indeed dead code as `Field` blocks never have links to other blocks within the `Links` field

internal/db/collection_truncate.go
```diff
@@ -391,3 +391,75 @@ func (c *collection) hardDeleteCollectionBlocks(
...
+			}
+
+		default:
+			for _, link := range decodedBlock.AllLinks() {
+				toDelete[link.Cid] = struct{}{}
+			}
+		}
+
+		if decodedBlock.Encryption != nil {
+			toDelete[decodedBlock.Encryption.Cid] = struct{}{}
```
fredcarle, line 453: todo: Please add code test coverage for truncating encrypted block.

internal/db/collection_truncate.go
```diff
@@ -391,3 +391,75 @@ func (c *collection) hardDeleteCollectionBlocks(
...
+		}
+
+		if decodedBlock.Signature != nil {
+			toDelete[decodedBlock.Signature.Cid] = struct{}{}
```
fredcarle, line 457: todo: Please add test coverage for truncating signed blocks.

fredcarle (approved):
Approving with 2 todos to resolve before merge. It would be good to take care of the dead code as well.

internal/db/collection_truncate.go
AndrewSisley: Ugh, I fell back into my prior/outdated thinking here.  I'll remove this, it should be a no-op.

- [x] Rm deadcode

internal/db/collection_truncate.go
AndrewSisley: I think this should be covered automatically.  I've opened https://github.com/sourcenetwork/defradb/issues/4452

internal/db/collection_truncate.go
AndrewSisley: I think this should be covered automatically. I've opened https://github.com/sourcenetwork/defradb/issues/4453

PR #4406, opened 2026-01-23
fix: Filter on top-level indexed JSON field

### LINKED ISSUE #4396: Fix filtering on json field with top-level condition

islamaliev:
## Relevant issue(s)

Resolves #4396

## Description

Fix filtering on top-level indexed JSON field

### DIFF
--- a/internal/db/fetcher/indexer_iterators.go
+++ b/internal/db/fetcher/indexer_iterators.go
@@ -800,8 +800,10 @@ func (f *indexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
 
 // makeFieldFilterCondition creates a fieldFilterCond based on the given operator and filter value on
 // the given indexed field.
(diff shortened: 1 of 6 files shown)


internal/db/fetcher/indexer_iterators.go
```diff
@@ -815,7 +817,7 @@ func makeFieldFilterCondition(
 	}
 
 	var err error
-	if len(jsonPath) > 0 {
+	if indexedField.Kind == client.FieldKind_NILLABLE_JSON && (len(jsonPath) > 0 || filterVal != nil) {
```
AndrewSisley: suggestion: Documentation here would be really handy.  I am struggling to figure out why this if is so nuanced.

AndrewSisley (approved):
LGTM :)

internal/db/fetcher/indexer_iterators.go
islamaliev: done

PR #4410, opened 2026-01-24
fix: Bundle libwasmer.so to avoid modcache dependency

### LINKED ISSUE #4384: libwasmer.so linked in go-modcache is unexpected
Just re-compiled DefraDB and everything worked fine, but I noticed the compilation had left about 11GB worth of stuff in my various go-cache and go-modcache directories. These seemed perfectly safe to blow away after compilation/installation was complete.

However, once I did, I inadvertently removed the libwasmer.so that was in the go-modcache, and defra wouldn't run.

I believe our installer should move that libwasmer.so to a more stable location (`/usr/local/lib..` or whatever), and we should have the binary linked there instead of into a deep location in the go-modcache.

SharanRP:
## Relevant issue(s)

Resolves #4384

## Description

This PR fixes an issue where the compiled `defradb` binary was dynamically linked to `libwasmer.so` residing inside the Go module cache (`$GOPATH/pkg/mod/...`). This caused the binary to fail at runtime if the module cache was cleared (e.g., via `go clean -modcache`) or if the binary was moved to a machine without the cache populated.

**Key Changes:**
1.  **New Helper Script ([tools/scripts/install_libwasmer.sh](cci:7://file:///home/sharanrp/Desktop/projects/defradb/tools/scripts/install_libwasmer.sh:0:0-0:0)):** 
    - Automatically locates the correct `libwasmer` shared object (handling Linux/macOS and AMD64/ARM64) within the installed `wasmer-go` module.
    - Copies the library to a specified target directory, handling file permissions gracefully (e.g., overwriting read-only cache files).

2.  **Updated [Makefile](cci:7://file:///home/sharanrp/Desktop/projects/defradb/Makefile:0:0-0:0):**
    - **`make build` (Local Dev):** Now bundles `libwasmer.so` directly into the `build/` directory alongside the executable. The binary is compiled with `CGO_LDFLAGS="-Wl,-rpath,$ORIGIN"`, ensuring it prefers the bundled library. This makes the build directory self-contained and portable.
    - **`make install` (System Install):** Installs `libwasmer.so` to a stable system path (default: `/usr/local/lib`) and links the binary to that absolute location.

## Tasks

## How has this been tested?

The fix was verified on Linux by confirming the binary loads the local library instead of the cache version:

1.  **Clean Build:** Ran `make clean` and removed any existing `build/libwasmer.so`.
2.  **Compilation:** Ran `make build` and verified that `libwasmer.so` was successfully copied to the `build/` directory.
3.  **Linkage Verification:**
    - Ran `readelf -d build/defradb | grep RUNPATH` and confirmed the output included `Library runpath: [$ORIGIN:...]`.
    - Ran `LD_DEBUG=libs ./build/defradb version` and confirmed specifically that the loader used the local file:
      ```
      trying file=/home/user/projects/defradb/build/libwasmer.so
      calling init: /home/user/projects/defradb/build/libwasmer.so
      ```

- Linux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,9 +80,12 @@ DEFAULT_TEST_DIRECTORIES=./...
 default:
 	@go run $(BUILD_FLAGS) cmd/defradb/main.go
 
+WASMER_INSTALL_PATH ?= /usr/local/lib
+
 .PHONY: install
 install:
-	@go install $(BUILD_FLAGS) ./cmd/defradb
+	@bash tools/scripts/install_libwasmer.sh $(WASMER_INSTALL_PATH)
+	@CGO_LDFLAGS="-Wl,-rpath,$(WASMER_INSTALL_PATH)" go install $(BUILD_FLAGS) ./cmd/defradb
 
 install-wizard: install
 	@defradb wizard
@@ -102,9 +105,13 @@ endif
 .PHONY: build
 build:
 ifeq ($(path),)
-	@go build $(BUILD_FLAGS) -o build/defradb cmd/defradb/main.go
+	@mkdir -p build
+	@bash tools/scripts/install_libwasmer.sh $(PWD)/build
+	@CGO_LDFLAGS="-Wl,-rpath,\$$ORIGIN" go build $(BUILD_FLAGS) -o build/defradb cmd/defradb/main.go
 else
-	@go build $(BUILD_FLAGS) -o $(path) cmd/defradb/main.go
+	@dir=$$(dirname $(path)); \
+	bash tools/scripts/install_libwasmer.sh $$dir; \
+	CGO_LDFLAGS="-Wl,-rpath,\$$ORIGIN" go build $(BUILD_FLAGS) -o $(path) cmd/defradb/main.go
 endif
 
 # Usage: make cross-build platforms="{platforms}"
--- a/Makefile
+++ b/Makefile
@@ -99,13 +99,16 @@ ifneq ($(OS_GENERAL),Linux)
 	@echo "Direct installation of Defradb's man pages is not supported on your system."
 endif
 
-# Usage:
-# 	- make build
-# 	- make build path="path/to/defradb-binary"
+# Target OS for cross-platform support
+GOOS ?= $(shell go env GOOS)
+
 BUILD_PATH = $(if $(path),$(path),build/defradb)
 WASMER_LIB_NAME = libwasmer.so
-ifeq ($(OS_GENERAL),Darwin)
+RPATH_TOKEN = \$$ORIGIN
+
+ifeq ($(GOOS),darwin)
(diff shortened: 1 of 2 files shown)


fredcarle (commented):
This seems ok to me but I'll let @shahzadlone who is much better than me with Makefiles review this.

Makefile
```diff
@@ -102,9 +105,13 @@ endif
...
 ifeq ($(path),)
-	@go build $(BUILD_FLAGS) -o build/defradb cmd/defradb/main.go
+	@mkdir -p build
+	@bash tools/scripts/install_libwasmer.sh $(PWD)/build
+	@CGO_LDFLAGS="-Wl,-rpath,\$$ORIGIN" go build $(BUILD_FLAGS) -o build/defradb cmd/defradb/main.go
 else
-	@go build $(BUILD_FLAGS) -o $(path) cmd/defradb/main.go
+	@dir=$$(dirname $(path)); \
+	bash tools/scripts/install_libwasmer.sh $$dir; \
+	CGO_LDFLAGS="-Wl,-rpath,\$$ORIGIN" go build $(BUILD_FLAGS) -o $(path) cmd/defradb/main.go
```
shahzadlone: suggestion/idea: This is getting more complex and duplicate code, what if we do the setting of directory/path in the ifeq/else clauses and then do the calling of the common steps outside (you can keep `@mkdir -p build` specific to `build/defradb` / default case).

Kind of can do like we do `OS_GENERAL` above (but do it here/local to this rule), by introducing a local `BUILD_PATH` variable perhaps (you can assume final binary name to be `defradb` in that path).

Once that is done (you have the `BUILD_PATH` deduced) then outside the clauses (but still local to the make target rule) can do:

1) `@bash tools/scripts/install_libwasmer.sh $$(dirname $(BUILD_PATH)); \` 
2) `@CGO_LDFLAGS="-Wl,-rpath,\$$ORIGIN" go build $(BUILD_FLAGS) -o $(BUILD_PATH) cmd/defradb/main.go`

NOTE: I have typed this on my cell phone so I probably have lots of mistakes in the syntax but just wanted to demonstrate the idea.

shahzadlone: question: So we will run this install script for wasmer every time we run the build command? I am not sure if that is desirable?

SharanRP:
@shahzadlone Thanks for the feedback! I've refactored the `Makefile` to remove the duplication and added a smart check so the install script only runs when the library is missing or `go.sum` has changed.

fredcarle:
Hi @SharanRP! It had slipped my mind but a few weeks ago we had made the decision to remove the dependency on libwasmer for the v1 release. Sorry for the time you spent on this PR. 

This is the issue that actually needs to be resolved: https://github.com/sourcenetwork/defradb/issues/4424

PR #4415, opened 2026-01-26
fix: Index on JSON _geq and _leq null filter ops

### LINKED ISSUE #4398: JSON nested null filter returns wrong results
Comparison operators with `null` value on nested JSON paths return fewer results than expected

islamaliev:
## Relevant issue(s)

Resolves #4398

## Description

Use index only for top-level `_leq: null` filter. Other cases like json nested `_leq` or any `_geq` are ignore by index as there is no benefit of using index there.

## Tasks

## How has this been tested?

Integration tests

- MacOS


internal/db/fetcher/indexer_iterators.go
```diff
@@ -762,6 +762,15 @@ func (f *indexFetcher) determineFieldFilterConditions() ([]fieldFilterCond, erro
 						return true
 					}
 
+					// For _geq: null, every value is >= null, so the index provides no benefit -
```
fredcarle, line 765: praise: Thanks for the documentation.

fredcarle (approved):
LGTM

PR #4416, opened 2026-01-26
fix: Flaky SourceHub tests

Lodek:
## Relevant issue(s)

#4441

## Description

Fixes the flaky SourceHub CI tests. This PR adds two primary changes:
1. Use a readiness probe which tests the actual running SourceHub process
2. Replace the setup logic which would spawn processes with `testcontainers` to create instances of sourcehub.

## Tasks

## How has this been tested?

make test and CI

- Arch Linux


.github/composites/get-sourcehub-image/action.yml
```diff
@@ -0,0 +1,55 @@
+# Copyright 2024 Democratized Data Foundation
```
shahzadlone: Todo: 2026

.github/composites/get-sourcehub-image/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    # Fetch sourcehub repo to get the full commit sha.
+    # note: go.mod's automatically generated placeholder version only stores the short commit.
+    - name: Checkout sourcehub code into the directory
+      uses: actions/checkout@v6
+      with:
+        repository: sourcenetwork/sourcehub
+        path: _sourcehub
+        fetch-depth: 0
+
+    - name: Build SourceHub image name
```
fredcarle, line 42: thought: It would be really nice if there was prebuilt containers for every version that were available on the sourcehub repo.

fredcarle (approved):
LGTM. There seems to be about 30-40 seconds added to the jobs that use this. We might be able to shave that a little with using pre-built containers but overall if we don't have to rerun the CI, it's a much better setup as it is.

.github/composites/get-sourcehub-image/action.yml
Lodek: I'm not sure I follow here Fred, the containers already exists in the SH repo.
Maybe Build is not the best name for the step, but we are just generating the image name.

PR #4420, opened 2026-01-26
test: Add truncate parallel test

### LINKED ISSUE #4325: Add concurrent integration tests for truncate

AndrewSisley:
## Relevant issue(s)

Resolves #4325

## Description

Adds a truncate parallel test, making sure that all previously existing documents are deleted.

## How has this been tested?

Executed locally (Linux), and re-running the CI a couple of times.

### DIFF
--- a/tests/action/parallel.go
+++ b/tests/action/parallel.go
@@ -44,25 +44,35 @@ func (a *Parallel) SetState(s *state.State) {
 func (a *Parallel) Execute() {
 	// startLock is responsible for ensuring that all child actions are scheduled/ready before
 	// any of them begin execution.
-	startWG := sync.WaitGroup{}
-	startWG.Add(len(a.Children))
+	startLock := sync.RWMutex{}
+	startLock.Lock()
 
 	// finishedWG is responsible for ensuring that all child actions complete their execution
 	// before this action-function unblocks/completes.
 	finishedWG := sync.WaitGroup{}
 
+	// childrenReady is responsible for ensuring that all child routines have been set up and are
+	// now waiting for the start lock to be unlocked.
+	childrenReady := sync.WaitGroup{}
 	for _, childAction := range a.Children {
+		childrenReady.Add(1)
 		finishedWG.Go(func() {
+			childrenReady.Done()
+
 			// Block behind the startWG until all child actions are ready to proceed
-			startWG.Done()
+			startLock.RLock()
+			defer startLock.RUnlock()
 
 			childAction.Execute()
 		})
 	}
 
+	// Wait for all the children to be ready before allowing them to start.
+	childrenReady.Wait()
+
 	// Release the start lock once all child actions are queued so that they may all begin at
 	// as close a point in time as the runtime/machine allows.
-	startWG.Wait()
+	startLock.Unlock()
 
 	finishedWG.Wait()
 }
--- a/tests/action/parallel.go
+++ b/tests/action/parallel.go
@@ -1,0 +1,68 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package action
+
+import (
+	"sync"
+
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+// Parallel executes all child actions in their own child go-routine, scheduled to start as
+// close in time as the runtime will allow.
+//
+// This action will not complete it's execution until all child actions have completed their
+// execution.
+type Parallel struct {
+	s *state.State
+
+	// The child actions that should be executed in parallel.
+	Children []Action
+}
+
+var _ Action = (*Parallel)(nil)
+var _ Stateful = (*Parallel)(nil)
+
+func (a *Parallel) SetState(s *state.State) {
+	a.s = s
+
+	for _, child := range a.Children {
+		if stateful, ok := child.(Stateful); ok {
+			stateful.SetState(s)
+		}
+	}
+}
+
+func (a *Parallel) Execute() {
+	// startLock is responsible for ensuring that all child actions are scheduled/ready before
+	// any of them begin execution.
+	startWG := sync.WaitGroup{}
+	startWG.Add(len(a.Children))
+
+	// finishedWG is responsible for ensuring that all child actions complete their execution
+	// before this action-function unblocks/completes.
+	finishedWG := sync.WaitGroup{}
+
+	for _, childAction := range a.Children {
+		finishedWG.Go(func() {
+			// Block behind the startWG until all child actions are ready to proceed
+			startWG.Done()
+
+			childAction.Execute()
+		})
+	}
+
+	// Release the start lock once all child actions are queued so that they may all begin at
+	// as close a point in time as the runtime/machine allows.
+	startWG.Wait()
+
+	finishedWG.Wait()
+}
--- a/tests/integration/collection/truncate/parallel_test.go
+++ b/tests/integration/collection/truncate/parallel_test.go
@@ -41,16 +41,19 @@ func TestCollectionTruncateParallel_DeletesAllPreviouslyExistingDocuments(t *tes
 			&action.Parallel{
 				Children: []action.Action{
 					&action.CreateDoc{
+						DoNotWaitForEvent: true,
 						DocMap: map[string]any{
 							"name": "Fred",
 						},
 					},
 					&action.CreateDoc{
+						DoNotWaitForEvent: true,
 						DocMap: map[string]any{
 							"name": "Shahzad",
 						},
 					},
 					&action.CreateDoc{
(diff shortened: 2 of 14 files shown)


tests/integration/collection/truncate/parallel_test.go
```diff
@@ -0,0 +1,94 @@
...
+							}
+						}`,
+						Results: map[string]any{
+							"Users": []map[string]any{},
+						},
+					},
+					&action.Truncate{},
+				},
+			},
+			// After the concurrent action has completed, make sure that `John` and `Islam` dont not exist.
```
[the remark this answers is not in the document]
AndrewSisley: :grin: Thanks coderabbit

tests/action/parallel.go
```diff
@@ -0,0 +1,67 @@
...
+			stateful.SetState(s)
+		}
+	}
+}
+
+func (a *Parallel) Execute() {
+	startLock := sync.RWMutex{}
+	startLock.Lock()
+
+	finishedWG := sync.WaitGroup{}
```
fredcarle, line 52: suggestion: calling it `finishedWG` made me look for meaning where there was none. Just `wg` would be fine.

tests/action/parallel.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for _, childAction := range a.Children {
+
+		finishedWG.Go(func() {
+			// Block behind the startLock until all child actions are ready to proceed
```
fredcarle: thought: This statement is misleading. it only blocks until `startLock.Unlock` is reached. 

suggestion: If you want to guarantee that the go routine is ready before reaching `startLock.Unlock`, you'd need an extra WaitGroup within the for loop. 
```go
childWG := &sync.WaitGroup{}
child.WG.Add(1)
finishedWG.Go(func() {
  child.WG.Done()
  ...
}
childWG.Wait()
```

AndrewSisley: Is there any benefit in doing so?  I assumed calling `go func(){}()` would schedule the routine (and block until scheduled), by waiting within the child func execution, I thought you'd be removing it from the scheduler again until it is unblocked.

I might have a search for this, but if you know, and/or have any handy links before I get round to looking that would be really handy :)

tests/action/parallel.go
AndrewSisley: There is meaning, but it looks like the name isn't good enough `finished[WG]` distinguishes it from `start[Lock]`, the wg controls the unblocking upon completion, the lock controls when the child routines begin.

I'll have another look at the name, but it is unlikely that it will be renamed to `wg`.

- [x] Have another look at `finishedWG`'s name, consider documenting

tests/action/parallel.go
AndrewSisley: I did some testing because I was interested :) 

TLDR is that it doesnt really matter.  

I believe that you are wrong here (if I understood you correctly) and that `go func(){}()` blocks until the child routine is scheduled.  Interestingly, the wait group appears to be faster than a rwlock, but both options allow the child routines to progress in roughly the same time frame (a range of about 3 microseconds).  Lock speed also seems to vary quite a lot per test execution (variance stays roughly the same though, and so does the wg).

Test code:
```
finishedWG := sync.WaitGroup{}
	var wg sync.WaitGroup
	wg.Add(1)
	a := time.Now()
	n := 20
	r := make([]int64, n)
	for i := 0; i < n; i++ {
		finishedWG.Go(func() {
			ii := i
			wg.Wait()

			r[ii] = time.Since(a).Nanoseconds()
		})
	}
	wg.Done()
	finishedWG.Wait()

	println("wg:")
	println(fmt.Sprint(r))
	var min int64 = 1000000
	var max int64 = 0
	var sum int64 = 0
	for _, v := range r {
		if v < min {
			min = v
		}
		if v > max {
			max = v
		}
		sum = sum + v
	}
	avg := float64(sum) / float64(n)
	var variance float64 = 0
	for _, v := range r {
		variance = variance + math.Abs(float64(avg)-float64(v))
	}
	println("min: " + fmt.Sprint(min))
	println("max: " + fmt.Sprint(max))
	println("range: " + fmt.Sprint(max-min))
	println("sum: " + fmt.Sprint(sum))
	println("avg: " + fmt.Sprint(avg))
	println("variance: " + fmt.Sprint(variance))

	time.Sleep(time.Millisecond)

	startLock := sync.RWMutex{}
	startLock.Lock()
	a = time.Now()
	r = make([]int64, n)
	for i := 0; i < n; i++ {
		finishedWG.Go(func() {
			ii := i
			startLock.RLock()
			defer startLock.RUnlock()

			r[ii] = time.Since(a).Nanoseconds()
		})
	}
	startLock.Unlock()
	finishedWG.Wait()
	println("lock:")
	println(fmt.Sprint(r))
	min = 1000000
	max = 0
	sum = 0
	for _, v := range r {
		if v < min {
			min = v
		}
		if v > max {
			max = v
		}
		sum = sum + v
	}
	avg = float64(sum) / float64(n)
	variance = 0
	for _, v := range r {
		variance = variance + math.Abs(float64(avg)-float64(v))
	}
	println("min: " + fmt.Sprint(min))
	println("max: " + fmt.Sprint(max))
	println("range: " + fmt.Sprint(max-min))
	println("sum: " + fmt.Sprint(sum))
	println("avg: " + fmt.Sprint(avg))
	println("variance: " + fmt.Sprint(variance))
	time.Sleep(time.Millisecond)

	panic("dfsfa")
```

Results:
```
wg:
[11271 11425 11610 11801 12008 12264 12380 12559 12780 12986 13173 13364 13546 13727 13974 14071 14237 14462 14636 10642]
min: 10642
max: 14636
range: 3994
sum: 256916
avg: 12845.8
variance: 19436

lock:
[28656 28798 28990 29165 29364 29618 29814 30013 30125 30246 30353 30584 30784 30917 31077 31213 31330 31459 31571 28157]
min: 28157
max: 31571
range: 3414
sum: 602234
avg: 30111.7
variance: 16860.6
```

AndrewSisley: I might switch to the wg as its slightly faster, but IMO it really doesn't matter here and whether or not I bother to change it probably depends on how much enthusiasm I have tomorrow :grin:

- [x] Consider changing to wg

fredcarle: There was a small misunderstanding. You would keep the start lock.

The issue before is that the write lock was almost guaranteed to be freed before all the goroutine were ready to operate. The idea behind the second wait group is to allow all goroutine to be instantiated and paused with the read lock before sending the start signal by unlocking the write lock.

AndrewSisley: That's what I thought you were saying, and what I think is incorrect :)

If it is technically possible due to the Go implementation, in reality it looks like the time delay between calling `go func(){}()` and setting up the last routine is tiny compared to the freeing  of the wg/lock.

The last routine in the loop does regularly appear to be the first to complete, but only by a very tiny about, and it might just be a quirk of the scheduler operating with a LIFO bias.

fredcarle: You're seeing that locally right? Have you tested with GOMAXPROCS=1? This would be more representative of a low resource machine.

AndrewSisley: I don't think that would add any relevant information, as the timings will be based on when the child jobs actually execute - it will just add more noise into the results.  None of the child jobs can begin until all of them have been constructed, so limiting GOMAXPROCS would just result in a queue of child jobs making any timing differences between the last constructed child and it's execution even larger - obscuring any time-delta not caused by the cost of unlocking either wg or mutex.

tests/action/parallel.go
```diff
@@ -0,0 +1,68 @@
...
+	startWG.Add(len(a.Children))
+
+	// finishedWG is responsible for ensuring that all child actions complete their execution
+	// before this action-function unblocks/completes.
+	finishedWG := sync.WaitGroup{}
+
+	for _, childAction := range a.Children {
+		finishedWG.Go(func() {
+			// Block behind the startWG until all child actions are ready to proceed
+			startWG.Done()
```
fredcarle: todo: This comment is still wrong. Nothing is blocking this goroutine from executing. Please remove the startWG as it's not doing anything for use here.

AndrewSisley: Nice spot lol, I switch to the wg but flipped the done/wait calls.  I think I'm going to revert to the lock, it is slower, but reads much better than a wg with only ever a 1 or zero value. 

- [x] Fix start wg

tests/action/parallel.go
AndrewSisley: Additional testing done, and I am now convinced that Fred is right :)  `go func(){}()` must be doing a whole bunch of async work before entering the app code.

Additionally original numbers posted were skewed - the difference between wg and mutex was caused by the reuse of `a` (possibly the inline disposal of the original value?).    Oddly, Fred's and my machines differed significantly here, with mine being consistently as originally posted, and Fred's sometimes reversing (wg being faster).

I will adjust the code before merging, and will probably revert back to the mutex for the unlock, as it is clearer and just as performant.

- [x] Wait for child routine setup before unlocking

tests/action/parallel.go
AndrewSisley: Resolving this conversation as it is covered by the original

fredcarle (approved):
LGTM

PR #4429, opened 2026-01-27
fix: Test matcher subscription bug

### LINKED ISSUE #4428: Test matcher subscription bug
I discovered a problem when debugging this test: `TestCommitSubscription_WithDocFilterAndMultipleMutations_FilteredDoc`

Test matchers are reset before subscriptions are finished, which means any matchers used within the subscription will give incorrect results.

nasdf:
## Relevant issue(s)

Resolves #4428

## Description

This PR fixes a bug with test matchers in subscriptions.

## Tasks

## How has this been tested?

Integration tests.

- MacOS

### DIFF
--- a/tests/integration/subscription/with_commit_test.go
+++ b/tests/integration/subscription/with_commit_test.go
@@ -174,8 +174,6 @@ func TestCommitSubscription_WithCommitLinksCreateMutations_ValidLinks(t *testing
 }
 
 func TestCommitSubscription_WithDocFilterAndMultipleMutations_FilteredDoc(t *testing.T) {
-	updateCid := testUtils.NewSameValue()
-
 	docID := "bae-45e90427-d499-598b-902a-6a3c65d0b504"
 	test := testUtils.TestCase{
 		Actions: []any{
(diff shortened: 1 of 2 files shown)


tests/integration/subscription/with_commit_test.go
```diff
@@ -194,6 +194,14 @@ func TestCommitSubscription_WithDocFilterAndMultipleMutations_FilteredDoc(t *tes
...
 				}`,
 				Results: []map[string]any{
+					{
+						"_commits": []map[string]any{
+							{
+								"cid":   "bafyreib5dvg3wkm722kietpvx5gmfueilyvywyiz2tl44q6xnhv4bedcpq",
+								"docID": docID,
+							},
+						},
+					},
```
fredcarle, line 206: question: I'm struggling to understand how we get 2 separate commits results. Why do we now have a second commit in the result set?

nasdf: One for the create and one for the update

fredcarle: We shouldn't be getting the one for the create since the subscription was initiated after the create call. That's why I'm puzzled by this.

nasdf: My guess would be that the create action just after the subscription triggers it to refresh.

nasdf: I just verified that is what's happening. We don't filter the select based on the DocID.

https://github.com/sourcenetwork/defradb/blob/38d0766692633817e0658f8548bd8fd13f4b71e2/client/request/commit.go#L61-L76

fredcarle: Thanks for testing. I was about to do that myself. 

I think we are filtering based on the DocID but not doing the `&` operation with the CID that would filter out the result.

fredcarle: Either way it's a bug. Thanks for finding it :)

fredcarle (approved):
Approving but before merge please open an issue explaining the bug and link the issue to a `TODO` above the test result item.

PR #4432, opened 2026-01-27
test: Add tests for truncation of views

### LINKED ISSUE #4390: No tests for truncating materialized views
It might not do anything either, as I'm not sure the view datastore keys are deleted by truncate atm.  It probably doesn't error.

AndrewSisley:
## Relevant issue(s)

Resolves #4390

## Description

Adds tests for the truncation of views.

The view-type multiplier ensures that this covers both materialized and non-materialized views.  Truncating a non-materialized view is a no-op.

### DIFF
--- a/tests/action/create_view.go
+++ b/tests/action/create_view.go
@@ -65,31 +65,29 @@ func (a *CreateView) Execute() {
 		}
 
 	default:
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
LGTM

Issue #4430, opened 2026-01-27
SE race condition


AndrewSisley:
`TestDocEncryptionPeer_WithSimpleRequest_ShouldFetchSuccessfully` is flaky and sometimes fails due to a detected datarace.  This appears to be a production issue, but only on database close.
```
=== FAIL: tests/integration/searchable_encryption TestDocEncryptionPeer_WithSimpleRequest_ShouldFetchSuccessfully (unknown)
Jan 27 18:18:55.284 INF tests.integration TestDocEncryptionPeer_WithSimpleRequest_ShouldFetchSuccessfully database=badger-in-memory client=cli mutationType=collection-named databaseDir="" badgerEncryption=false skipNetworkTests=false changeDetector.Enabled=false changeDetector.SetupOnly=false changeDetector.SourceBranch=develop changeDetector.TargetBranch="" changeDetector.Repository=https://github.com/sourcenetwork/defradb.git KMS=pubsub
Jan 27 18:18:55.344 INF p2p Created LibP2P host PeerId=12D3KooWMBKL2MHwVDnWLEJjRhwJFWdxgYcvTtPdgcqxfjGeXRCE Address=[/ip4/10.2.2.108/tcp/0]
Jan 27 18:18:55.349 INF p2p Adding pubsub topic PeerID=12D3KooWMBKL2MHwVDnWLEJjRhwJFWdxgYcvTtPdgcqxfjGeXRCE Topic=doc-sync
Jan 27 18:18:55.350 INF p2p Adding pubsub topic PeerID=12D3KooWMBKL2MHwVDnWLEJjRhwJFWdxgYcvTtPdgcqxfjGeXRCE Topic=sync-branchable
Jan 27 18:18:55.351 INF p2p Adding pubsub topic PeerID=12D3KooWMBKL2MHwVDnWLEJjRhwJFWdxgYcvTtPdgcqxfjGeXRCE Topic=encryption
Jan 27 18:18:55.351 INF db Starting without nac (not configured/setup)
Jan 27 18:18:55.402 INF p2p Created LibP2P host PeerId=12D3KooWHoACqyUrQSMSS5Ucs5Uichd3BgfmQAq1GhscqUEC3pB2 Address=[/ip4/10.2.2.108/tcp/0]
[...]

PR #4435, opened 2026-01-28
fix: Prevent database operations after close

### LINKED ISSUE #4085: Possible panic caused by race condition on close
There is a possible scenario (real world low probability) demonstrated by our integration tests where a panic can occur when closing a defra node while a subscription or replicator retry is active. 

With the subscription, this happens if a mutation is received by the subscription handler at the same time that defra is being closed. It results in a null pointer when the datastore used is badger (memory or file).

SharanRP:
## Relevant issue(s)

Resolves #4085 (and addresses symptoms in #4430)

## Description

This PR fixes a critical shutdown race condition where background tasks (subscriptions and replicator retries) would attempt to perform database operations after the node had been closed.

When use with Badger, this resulted in "null pointer" panics or "send on closed channel" errors because the store was closed while background goroutines were still processing the last few events or failures.

Changes:
- Added `ctx context.Context` to the [DB](cci:2://file:///home/sharanrp/Desktop/projects/defradb/internal/db/p2p/p2p.go:66:0-96:1) struct to track node lifetime.
- Prevented new transactions from being created during shutdown in [NewTxn](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/db.go:456:0-468:1) and [NewConcurrentTxn](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/db.go:221:0-226:1).
- Added context checks to the subscription event loop in [handleSubscription](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/subscriptions.go:27:0-138:1).
- Added context and mutex safety checks to replicator failure handlers.

## Tasks

## How has this been tested?

Tests were performed with the Go race detector enabled:
- `go test -race ./tests/integration/searchable_encryption/... -count=20`
- `go test -race ./tests/integration/subscription/... -count=10`
- Verified logs show graceful early exits: `ERR p2p Failed to handle replicator failure. $err="context canceled"`

Tested on:
-  Linux


internal/db/subscriptions.go
```diff
@@ -63,6 +63,11 @@ func (db *DB) handleSubscription(ctx context.Context, r *request.Request) (<-cha
 					continue // invalid event value
 				}
 			}
+
+			if ctx.Err() != nil {
+				return
+			}
+
```
fredcarle: suggestion: I don't think this check is necessary since the select statement already handles checking for context done.

internal/db/p2p/replicator.go
```diff
@@ -389,12 +389,25 @@ func (p *P2P) handleReplicatorRetries(ctx context.Context) {
 }
 
 func (p *P2P) handleReplicatorFailure(ctx context.Context, peerID, docID string) error {
+	// Check if context is cancelled before attempting database operations.
+	// This prevents attempts to write to a closed database during shutdown,
+	// which can cause a panic ("send on closed channel").
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
```
fredcarle: suggestion: If you want to check after the lock is acquired that's fine but checking before and after is a bit overkill. I suggest removing the first one and just checking after the lock is aquired.

internal/se/coordinator.go
```diff
@@ -217,6 +217,12 @@ func (coordinator *Coordinator) handleReplicationFailure(
 	docID, collectionID, peerID string,
 	fieldNames []string,
 ) error {
+	// Check if context is cancelled before attempting database operations.
+	// This prevents attempts to write to a closed database during shutdown.
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
```
fredcarle: suggestion: This one isn't needed as the check is taken care of with the `NewTxn` call bellow.

fredcarle (approved):
LGTM

PR #4436, opened 2026-01-28
fix: Ensure commit subscriptions filter by docID

### LINKED ISSUE #4434: Commit subscriptions trigger on all updates
Commit subscriptions do not filter updated events based on DocID.

This causes previously made commits to appear in subscriptions.

Update this test when fixed: `TestCommitSubscription_WithDocFilterAndMultipleMutations_FilteredDoc`

https://github.com/sourcenetwork/defradb/blob/38d0766692633817e0658f8548bd8fd13f4b71e2/client/request/commit.go#L61-L76

SharanRP:
## Relevant issue(s)

Resolves #4434

## Description

This PR fixes a performance bug where commit subscriptions would wake up for every single update in the database, even if it wasn't for the document being watched.

I updated the code to check the `docID` early. If the update doesn't match the subscribed document, it now stops immediately and discards the transaction.

## Tasks

## How has this been tested?

Verified using the dedicated integration test:
- `TestCommitSubscription_WithDocFilterAndMultipleMutations_FilteredDoc`


fredcarle:
Hi Sharan. As you'll notice I've modified the code a bit to not have the check in the SubscriptionSelect constructor. The constructor should be left to do only one thing and that is constructing the object.

I've taken advantage of this change to include a check on the CID as well. Plus modified the problematic test to be representative of the title/description. It's now testing what it had intended to test.

Thanks for opening the PR and initiating this change.

jsimnz (approved):
Thanks @SharanRP for jumping on this so quickly. Although it was slightly updated by Fred, it gave a great starting point, and highlighted an optimization we could take advantage of.

PR #4438, opened 2026-01-28
fix: Aggregates on indexed relations

### LINKED ISSUE #4401: Incorrect results with deep filter/order with sum aggregation
Queries combining deep filters on relations, ordering, and sum aggregation return incorrect values when using index

islamaliev:
## Relevant issue(s)

Resolves #4401

## Description

When computing aggregates with deep filters on related types using secondary indexes, the `invertibleTypeJoin` was not clearing its state between iterations for different parent documents.

The `encounteredDocIDs` and `docsToYield` fields retained values from previous iterations, causing the join to skip documents it incorrectly thought it had already processed. This resulted in aggregate functions returning 0 for all results except the first parent.
                   
The fix clears these fields in `Init()` so each parent document gets a fresh iteration through its related documents.

## Tasks

## How has this been tested?

Integration tests

- MacOS


fredcarle (approved):
LGTM

PR #4451, opened 2026-01-29
fix: Resolve storage leak upon document deletion

### LINKED ISSUE #4419: Test for storage leaks
We currently do not test that space is actually freed in the store upon delete, we only test that the data is no longer accessible via our interfaces.

SharanRP:
## Relevant issue(s)

Resolves #4419

## Description

Fixes a storage leak where document data (DataStore keys and secondary indexes) persisted after deletion. Updated [applyDelete](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/db/collection_delete.go:127:0-272:1) to explicitly purge Primary, Value, and Field keys from the projected state while preserving DAG history for P2P sync.

The fix targets the local projected state (DataStore). BlockStore blocks (DAG history) are intentionally preserved to allow P2P sync of the deletion event, which is standard CRDT behavior but may result in non-zero growth of the BlockStore.

## Tasks

## How has this been tested?

Added [TestStorageFreedUponDelete](cci:1://file:///home/sharanrp/Desktop/projects/defradb/tests/integration/mutation/delete/storage_leak_test.go:96:0-123:1) regression test. It performs a raw iteration of the [Rootstore](cci:1://file:///home/sharanrp/Desktop/projects/defradb/node/node.go:39:1-39:28) to verify that no active data keys remain for a deleted document.

Run: `go test ./tests/integration/mutation/delete/ -v -run TestStorageFreedUponDelete`

- Linux


AndrewSisley:
Hey Sharan, I think you have misunderstood the ticket and the function you have changed here.  I'm going to close this PR as it is quite stale now.

PR #4454, opened 2026-01-30
feat: Fetch orphan parent docs with exhaustive

### LINKED ISSUE #4551: Queries ordering by indexed relation fields exclude documents without the related document
When a query orders by a field on a related type and that field has a secondary index, the results may differ from the same query without the index. Specifically, documents that don't have the related document are silently excluded.

For example, given Book with an optional publisher relation and an index on Publisher.yearOpened:

```graphql
query {
  Book(order: {publisher: {yearOpened: ASC}}) { title }
}
```

Without the index, this returns all books — books without a publisher appear with null values. With the index, the planner inverts the join to use the index for ordering, which means it iterates publishers and fetches their books. Books without a publisher are never visited and don't appear in results.

This inconsistency is by design for performance, but users need a way to get complete results when they need them.

## Solution:

Add a @exhaustive query directive that opts into including documents without the related document at the correct sort position (nulls first for ASC, nulls last for DESC):

```graphql
query @exhaustive {
  Book(order: {publisher: {yearOpened: ASC}}) { title }
}
```

islamaliev:
Resolves #4551

When ordering by a relation field that uses a secondary index, the planner inverts the join for efficiency. This means parents without the related document (orphans) are excluded from results. This PR adds a `@exhaustive` query directive to opt into including them:

```graphql
query @exhaustive {
  Book(order: {publisher: {yearOpened: ASC}}) { title }
}
```

Orphan handling works at both top-level and nested (subquery) ordering, is limit-aware, and uses different strategies depending on which side of the relation stores the foreign key.

Other changes: extracted index selection into a single function with clear priority ordering, fixed the sub-query filter builder to be non-mutating, and updated pre-existing tests to use `@exhaustive` where they expect orphans in results.

## Documentation note

@ashutosh-src  The secondary index documentation should be updated to cover the following.

When a query orders by a field on a related type and that field has a secondary index, the planner uses the index to avoid sorting. It does this by inverting the join: instead of scanning the parent collection and looking up the child for each, it walks the child's index in order and fetches the matching parents. For example, `Book(order: {publisher: {yearOpened: ASC}})` with an index on `Publisher.yearOpened` will iterate publishers by year and fetch each publisher's book, rather than scanning all books and sorting. This is faster but means books that have no publisher are never visited and won't appear in results.

The `@exhaustive` directive restores complete results. It is placed at the query level (same position as `@explain`) and takes no arguments. When present, the system performs additional fetches to find parent documents that have no related child and inserts them at the correct sort position: nulls first for ASC, nulls last for DESC. This applies both to top-level ordering and to ordering inside nested selections (e.g. ordering books within an author query). Using `@exhaustive` adds overhead: extra queries and potentially more memory, so it should only be used when the caller needs to see all documents regardless of whether the relation exists.

## How this was tested

With integration tests


internal/planner/orphan.go
```diff
@@ -0,0 +1,303 @@
...
+			excludeIDs = append(excludeIDs, id)
+		}
+		orphans, err = fetcher.fetchAllExcluding(n.join.subFilter, excludeIDs)
+	}
+
+	if err != nil {
+		return err
+	}
+
+	n.orphanDocs = orphans
```
AndrewSisley: todo: `n.orphanDocs` is only ever read immediately after a call to `n.fetchOrphans()`.  After reading it is set to nil.

Passing it around like this makes the code far more complex than it needs to be, please return it from `fetchOrphans()` instead of doing this.

client/request/consts.go
```diff
@@ -57,7 +57,8 @@ const (
 	// which might have a different _docID originally.
```
AndrewSisley, line 57: question: Why did you not use the PR description template?  Do you dislike it?  Would you like to change/drop it?

AndrewSisley, line 57: todo: Please fix the typo in the PR title, it will show up in the change log

internal/planner/orphan.go
```diff
@@ -0,0 +1,303 @@
...
+	for {
+		hasNext, err := n.source.Next()
+		if err != nil {
+			return false, err
+		}
+		if !hasNext {
+			n.sourceExhausted = true
+			return n.nextASCSecondaryParent()
+		}
+		n.bufferedDocs = append(n.bufferedDocs, n.source.Value())
```
AndrewSisley: suggestion: `n.bufferedDocs` is a confusing property to me.

Maybe it just needs documentation, however it looks like it should be removed from the node state:

It is only referenced in this function, and looks like it is only ever set to not-nil once per `Init()`.

The more I look at `nextASCSecondaryParent` the more confused it seems - the first half of the function only ever seems to be called from the second half of the function via the fake-recursion call, and the second half of function can only be reached once.

I think this function needs to be rewritten, and `n.bufferedDocs` removed.

internal/planner/orphan.go
```diff
@@ -0,0 +1,303 @@
...
+	if len(n.docsToYield) == 0 {
+		return core.Doc{}
+	}
+	return n.docsToYield[0]
+}
+
+func (n *orphanNode) Next() (bool, error) {
+	n.execInfo.iterations++
+
+	if len(n.docsToYield) > 0 {
```
AndrewSisley: todo: Please document this block

internal/planner/planner.go
```diff
@@ -104,6 +104,16 @@ type Planner struct {
...
+
+	// exhaustive is set when the @exhaustive directive is present on the query.
+	// When true, orphan parent documents will be included when ordering by relation
+	// fields with indexes. When false (default), orphans are excluded for performance.
+	exhaustive bool
+
+	// inNestedJoin tracks whether we're expanding a join that is nested inside another join.
+	// When true, orphanNode should not be added because nested joins are iterated via
+	// retrievePrimaryDocs which handles orphans correctly with parent context.
+	inNestedJoin bool
```
AndrewSisley: todo: This feels like the wrong place to store `inNestedJoin` and `exhaustive`, please remove them from here.  This type will get very busy very quickly if we were to do this for everything.

AndrewSisley (changes requested):
Still going through this, but have some todos that I'm submitting now.

internal/planner/subquery_fetcher.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/sourcenetwork/defradb/internal/lens"
+	"github.com/sourcenetwork/defradb/internal/planner/filter"
+	"github.com/sourcenetwork/defradb/internal/planner/mapper"
+	lensStore "github.com/sourcenetwork/lens/host-go/store"
+)
+
+// subQueryFetcher executes sub-queries independently without mutating scan nodes.
+// It creates its own fetcher instance for each query, avoiding shared state issues
+// that can arise from the save-restore pattern.
+type subQueryFetcher struct {
```
AndrewSisley: question: What drove you to introduce this type?

todo: If it was frustration with the walkAndReplacePlan stuff, please note in the documentation for this type that this is a temporary workaround, and link this type to a github issue so that we can remove it later when the walkAndReplacePlan stuff gets sorted.

internal/planner/type_join.go
```diff
@@ -172,9 +167,12 @@ func (n *typeIndexJoin) simpleExplain() (map[string]any, error) {
 	}
 
 	var err error
-	switch joinType := n.joinPlan.(type) {
+	joinPlan := n.joinPlan
+	if orphan, ok := joinPlan.(*orphanNode); ok {
+		joinPlan = orphan.source
```
AndrewSisley: question: This appears to be bypassing explain in an unusual way - why are you doing this?

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+
+	result := append(joinDocs, orphans...)
+	if limit > 0 && uint64(len(result)) > limit {
+		result = result[:limit]
+	}
+	return result, nil
+}
+
+// getLimit returns the limit from the primary side's plan, or 0 if no limit.
+func (r *primaryObjectsRetriever) getLimit() uint64 {
```
AndrewSisley: question: I think that this code touching limit is quite a large code smell - why are you doing this instead of using the `limitNode`?

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+// a correct exclusion set, then identifies orphans as docs not in the join results.
+// The limit is temporarily removed and re-applied after merging.
+// Used when the primary doc does not store the FK for the ordering relation.
+func (r *primaryObjectsRetriever) collectDocsWithOrphansByExclusion(
+	direction mapper.SortDirection,
+	limit uint64,
+) ([]core.Doc, error) {
+	// Must remove limit to get all join docs for exclusion set.
+	if limit > 0 {
+		r.setLimit(0)
```
AndrewSisley: question: What is the significance of this?  Is it going to fetch documents from the store ignoring the user provided limit, resulting in a query that works significantly harder than the user expects?

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+		&r.primaryScan.execInfo.fetches,
+	)
+
+	return fetcher.fetchOrphans(relIDFieldName, relIDFieldMapIndex, parentFilter)
+}
+
+// fetchOrphanDocsByExclusion fetches all docs for the current parent and excludes
+// those present in joinDocs. Used when the primary doc does not store the FK for the
+// ordering relation, so orphans cannot be identified from the doc data alone.
+func (r *primaryObjectsRetriever) fetchOrphanDocsByExclusion(joinDocs []core.Doc) ([]core.Doc, error) {
```
AndrewSisley: question: Fetching by an excludes list feels counter intuitive to me, why can we not fetch orphans and non-orphans in a single pass?

question: What is in `joinDocs`?  If there is a user-provided limit, and orphans are to be returned first, will the `joinDocs` not potentially be very large compared to the returned set of orphans?

suggestion: Please document why we are fetching using an exclusion list.

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+//
+//	Orphans are self-identifying via FK IS NULL.
+//	ASC: fetch orphans first, fill remaining from join.
+//	DESC: fetch from join with limit, fill remaining with orphans.
+//
+// Secondary (no FK, e.g. Publisher stores book_id, not Book):
+//
+//	Orphans can only be identified by exclusion after seeing join results.
+//	Both ASC and DESC: collect join docs first, then fetch all docs for parent,
+//	exclude join doc IDs to find orphans.
```
AndrewSisley: suggestion/question: The code in this file appears to be duplicating a lot of the ordering work performed by the `orphanNode`.

Can you please clarify the differences between what is going on in the two files, and why it is needed?

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+	result := append(joinDocs, orphans...)
+	if limit > 0 && uint64(len(result)) > limit {
+		result = result[:limit]
+	}
+	return result, nil
+}
+
+// getLimit returns the limit from the primary side's plan, or 0 if no limit.
+func (r *primaryObjectsRetriever) getLimit() uint64 {
+	if selectTop, ok := r.primarySide.plan.(*selectTopNode); ok {
```
AndrewSisley: todo: Please document under what conditions the cast will fail, and why we want silently return `0` in such a situation instead of erroring.

AndrewSisley (changes requested):

internal/planner/type_join_orphans.go
```diff
@@ -0,0 +1,256 @@
...
+
+	orphans, err := r.fetchOrphanDocsByFK()
+	if err != nil {
+		return nil, err
+	}
+
+	if limit > 0 {
+		remaining := limit - uint64(len(joinDocs))
+		if uint64(len(orphans)) > remaining {
+			orphans = orphans[:remaining]
```
AndrewSisley: todo: This appears to be fetching everything, ignoring the limit, and then trimming the result - this largely defeats the the purpose of providing a limit as the bulk of the work is being done anyway.

If this is how the code is behaving, please rework it.

If it is not how the code is behaving, please add documentation explaining what and why it is doing what it is doing.

internal/planner/type_join_orphans.go
islamaliev: The exclusion path is only used when we fetch secondary docs that do not store the FK for the ordering relation. They don't have information about orphan relations. The only way to find orphans is to first see which docs the join produced, then fetch all docs and subtract.

When we fetch primary docs we use `fetchOrphanDocsByFK` with a `FK IS NULL` filter instead, which avoids the exclusion set entirely.

Regarding `joinDocs` size: yes, for the secondary side the limit must be temporarily removed to get all join docs, so `joinDocs` can be larger than the final result. If we kept the limit, docs clipped by it would incorrectly appear as orphans in the exclusion. The primary-side path does not have this problem.

internal/planner/type_join_orphans.go
islamaliev: hm... turns out it wasn't necessary. Added a comment

internal/planner/type_join_orphans.go
AndrewSisley: This doesn't yet make sense to me:

- If there is a filter on the parent, then only non-orphans could ever be returned no?
- If there no filter on the parent, then the parent is irrelevant to which docs get returned, and the child set can be fetched as if it was not a join (a single scan, no need for orphan magic).  
  - If there is a limit, then a single child scan will be more efficient and predictable than fetching all the parents and building a potentially very large exclusion set
  - If there is no limit, then we don't need to worry about the limit related stuff, or an exclusion set (no parent filter)

client/request/consts.go
islamaliev: done

islamaliev: I think we can simplify it, but I'm advocating for dropping.

internal/planner/type_join_orphans.go
islamaliev: I refactored a bit, this file is gone now

internal/planner/orphan.go
islamaliev: done: refactored orphanNode a bit and added more comments to functions

internal/planner/type_join_orphans.go
islamaliev: The query is e.g. `Book(order: {publisher: {establishedYear: ASC}})`. The inverted join scans Publisher's `establishedYear` index to produce Books in the correct order. An orphan Book (no linked Publisher) never appears in that index scan, it has no Publisher entry to index.

> If there is a filter on the parent

the parent here is Book (the queried collection). A filter on Book doesn't help identify orphans. Even `Book(filter: {title: {_eq: "Foo"}}, order: {publisher: {establishedYear: ASC}})` can match a Book that has no Publisher. The filter constrains which Books we return, not whether they have a Publisher.

> If there is no filter on the parent, then the parent is irrelevant and the child set can be fetched as a single scan

the child here is Publisher (the relation target being ordered). We can't just scan Books directly because we need them ordered by `publisher.establishedYear`, which lives on Publisher. The whole point of the inverted join is to use Publisher's index for ordering. A single Book scan would give us Books in docID order, not in publisher's establishedYear order. We'd then need a sort node, defeating the index optimization.

The exclusion set is specifically needed because the secondary side (Book) has no FK column — the FK (_bookID) lives on Publisher. So there's no "bookPublisherID IS NULL" field to filter on. The only way to know a Book has no Publisher is: we scanned all Publishers via the index, resolved their Books, and any Book not encountered is an orphan.

That said, I agree this path is expensive — it requires seeing all join results before identifying orphans. The primary-side path avoids this entirely with FK IS NULL. The exclusion path is the fallback for schemas where the FK is on the other side.

internal/planner/orphan.go
islamaliev: Done. bufferedDocs is removed, nextASCSecondaryParent is gone. The rewritten nextASC is a straightforward linear flow

internal/planner/type_join_orphans.go
AndrewSisley: > the parent here is Book (the queried collection)

We are talking about parents in two different ways.  I am referring to the relationship structure, you appear to be referring to the query structure.

When I say 'If there is a filter on the parent', I am talking about one on `publisher`.

---

I see the mistake I made there:
>  then the parent is irrelevant to which docs get returned

This is not true ignoring orphans.  As the order combined with the limit does affect which records are returned...

> The exclusion set is specifically needed because the secondary side (Book) has no FK column — the FK (_bookID) lives on Publisher. So there's no "bookPublisherID IS NULL" field to filter on.

Yes, but what you could do is read the child (book), then point look up to the publisher index to see if the book-doc exists within it - this avoids the need to build an exclusion set, as the index on the primary already *is* the exclusion set. 

You should be able to rework this without ever fetching the parents (publisher).  You can sort the scanned children (books) very efficiently (including determining whether they are orphans or not) as they are read into memory using the index on the primary, without actually having to read the parents into memory (just the index key-values required to fulfil the limit).

internal/planner/orphan.go
islamaliev: done

internal/planner/planner.go
islamaliev: They're now grouped into a `joinExpandState` struct, but you're right, it still lives on Planner. The reason is that these values need to flow across expandPlan, expandTypeIndexJoinPlan, expandTypeJoin back to expandSelectTopNodePlan, and expandPlan is a generic recursive dispatch that can't carry feature-specific parameters. Planner is the only shared context across that call chain. Open to suggestions if you see a cleaner way to thread this state through.

internal/planner/subquery_fetcher.go
islamaliev: It's not a workaround for walkAndReplacePlan. The orphan detection needs to run independent fetches mid-iteration, e.g. "fetch all parents where FK IS NULL" or "fetch all parents except these IDs." 
These can't go through the existing scanNode because it's already being iterated by the join. Mutating its filter or fetcher state mid-iteration would corrupt it. subQueryFetcher creates its own fetcher instance per query with proper cleanup, so it's independent by design, not by limitation.

internal/planner/type_join.go
islamaliev: I rewrote it

internal/planner/type_join_orphans.go
islamaliev: yeah, it was a tricky one, but I refactored it so that now limitNode does the job

internal/planner/type_join_orphans.go
islamaliev: This is refactored already

internal/planner/type_join_orphans.go
islamaliev: This file (type_join_orphans.go) has been removed in the refactor. The old collectDocsWithOrphansByExclusion that removed the limit, fetched everything, then trimmed is gone

The current approach splits into two cases: 

Primary parent (stores FK): Uses a sequenceNode pipeline. The limitNode sits above and stops pulling after N docs naturally. No limit manipulation needed.

Secondary parent (no FK, exclusion path): This one still has to see all join results before identifying orphans, there's no way around it. If you have a limit of 5 but 1000 joined docs, we need all 1000 to know which parents are missing. But the limit is now enforced naturally by the pipeline (limitNode above the orphanNode wrapper) rather than by fetch-all-then-trim.

internal/planner/type_join_orphans.go
islamaliev: Yes, the unique index on _bookID means a point lookup per Book is cheap. But it only tells you whether a Publisher exists — not what its establishedYear is. You'd still need to fetch the Publisher or do another lookup to get the sort value. And once you have all the sort values, you need to sort in memory — which is exactly what orderNode does, buffering all docs into allSortStrategy before sorting.

The whole point of the inverted join is to avoid that in-memory sort by walking the establishedYear index in order. Switching to per-Book lookups would reintroduce it, making it strictly worse: N point lookups + N Publisher fetches + full in-memory sort, vs one index scan (already sorted) + an exclusion set of string IDs.

internal/planner/planner.go
AndrewSisley: `exhaustive` is a directive property on a GQL request, and should be able to live quite happily on the mapper/request objects along with other directives such as explain no?  It is odd if it is not represented there I think.

`inNestedJoin` sounds join specific, and is always set to true for any plan graph with a type join.  It only looks like it is read in a location where it is always true, and looks like it should be deleted (it is read in `expandTypeIndexJoinPlan`, after calling `expandJoin` which always sets it to true).

internal/planner/subquery_fetcher.go
AndrewSisley: Why can't you just use another scan node instance instead of this?  You can have more than one instance if you want one lot live a lot longer without mutation than the other.

internal/planner/type_join_orphans.go
AndrewSisley: You can still use the index to sort without holding all of it in memory.  And you can sort as you read:
```
var orderedSetOfChildren
for fetcher.HasChild() {
   child :=  fetcher.Value()
   
   isOrphaned := fkIndexLookup()
   sortValue := orderIndexLookup()
   
   orderedSetOfChildren.InsertChildInCorrectOrder(sortValue, child)
   
   if orderedSetOfChildren == limit {
      return orderedSetOfChildren
   }
}
return orderedSetOfChildren
```
This is makes it an O(N) operation that respects the limit and performs in a predictable manner, without holding unwanted stuff in memory.

internal/planner/subquery_fetcher.go
islamaliev: scanNode is built via p.Scan(mapperSelect) which requires a *mapper.Select, the orphan fetch doesn't come from a parsed select, it's constructed at runtime with a programmatic filter. I'd need to either synthesize a mapper.Select to satisfy p.Scan(), or build the scanNode manually by duplicating the same field setup that subQueryFetcher already does.

Current approach seems to be simpler.

That said, I can see the argument that having fetch logic outside of scanNode is duplication. I will check if it's easy/reasonable to switch to scanNode

internal/planner/type_join_orphans.go
islamaliev: The pseudocode iterates Books in docID order, not in publisher.establishedYear order. With limit: 3 over 10,000 Books, you can't stop at 3 — the next Book might have a lower establishedYear than anything seen so far. You'd scan all 10,000 Books, do 10,000 point lookups, and maintain a sorted set, vs the inverted join which walks the establishedYear index (already sorted), takes the first 3 entries, and stops.

The `if orderedSet == limit { return }` early exit only works when the scan order matches the sort order, which it doesn't here.

You're right that the exclusion path is expensive though. The proper fix is cross-collection indexes (#4585) — with ability to index on Book._publisherID (secondary side), orphan Books would have NULL entries in the index, so you could seek directly to the NULL range. No exclusion set, limit works naturally, no need to fetch everything first. The exclusion path is the correct fallback until that lands.

AndrewSisley: > You'd scan all 10,000 Books

Yes, this is also true for the parent side though - there could be 10,000 publishers, and you are currently fetching all of them in order to return none of them no?

This is what I mean by my proposal being more predictable - books is not indexed, so it is somewhat expected to be scanning through all of the books - it feels much less expected to me to scan through all of the publishers to do this.

Thanks for adding #4584 - this makes sense, and is the correct way to index for the situation in this PR, users should be deliberately construct and consume the index, instead of either half-using it in my suggestion, or doing your proposal, which I see as completely unexpected.

Might be worth chatting about this on discord - I haven't yet marked this as 'todo' level for me, but I feel like fetching all parent/publisher records in order to build an exclusion set for the child (books) is something that users will not expect, in an area where I think predictability is important.  

It will also be detrimental to performance in a lot of cases, and feels like we are trying to be too clever, and will be getting it wrong for some queries.

internal/planner/filter/split.go
```diff
@@ -41,7 +41,11 @@ func SplitByFields(filter *mapper.Filter, fields ...mapper.Field) (*mapper.Filte
 		if newSplitF == nil {
 			continue
 		}
-		splitF.Conditions = MergeConditions(splitF.Conditions, newSplitF.Conditions)
+		if splitF == nil {
+			splitF = newSplitF
```
AndrewSisley: todo: Please document when this can happen, codecov says it is dead code.

internal/planner/filter/split.go
```diff
@@ -41,7 +41,11 @@ func SplitByFields(filter *mapper.Filter, fields ...mapper.Field) (*mapper.Filte
 		if newSplitF == nil {
 			continue
 		}
-		splitF.Conditions = MergeConditions(splitF.Conditions, newSplitF.Conditions)
+		if splitF == nil {
```
AndrewSisley: question: Why did you change this now?  It looks like an optimization.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+
+// fetchOrphans fetches and returns parent documents that have no related children.
+// Used only in standalone mode (FK IS NULL path for primary-side parents) and
+// subquery mode (nested joins with primary ordering).
+func (n *orphanNode) fetchOrphans() (_ []core.Doc, err error) {
+	n.fetched = true
+
+	parentScan := getNode[*scanNode](n.join.parentSide.plan)
+	if parentScan == nil {
+		return nil, nil
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	if !n.join.parentSide.relFieldDef.HasValue() {
+		return nil, nil
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+
+	var orphanFilter *mapper.Filter
+	var relationIDFieldName string
+
+	if n.isSubQuery {
+		orphanFilter = addNullFilterOnField(n.subQueryFilter, n.subQueryRelIDFieldMapIdx)
+		relationIDFieldName = n.subQueryRelIDFieldName
+	} else if n.join.parentSide.isPrimary() {
+		if !n.join.parentSide.relIDFieldMapIndex.HasValue() {
+			return nil, nil
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+	if n.source != nil {
+		return n.nextWrapped()
+	}
+	return n.nextStandalone()
+}
+
+// nextStandalone fetches all orphans on first call, then yields them sequentially.
+func (n *orphanNode) nextStandalone() (bool, error) {
+	if !n.fetched {
+		orphans, err := n.fetchOrphans()
```
AndrewSisley: question: Why are you collecting all the orphans up front, storing them in a slice, and then yielding them one by one?

Why not yield them as they are read?

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+}
+
+func (n *orphanNode) Init() error {
+	n.docs = nil
+	n.current = 0
+	n.fetched = false
+	n.phases = nil
+	n.pointLookupDone = false
+	if n.parentClone != nil {
+		_ = n.parentClone.Close()
```
AndrewSisley: todo: Please document why you think discarding this error is appropriate.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+	if n.source != nil {
+		return errors.Join(n.source.Close(), cloneErr)
+	}
+	return cloneErr
+}
+
+func (n *orphanNode) Next() (bool, error) {
+	n.execInfo.iterations++
+
+	if n.source != nil {
```
AndrewSisley: suggestion: This very much looks like this is two node-types posing as one - please split them, it should make reading them a bit easier.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+// nextOrphanByPointLookup returns the next orphan parent by iterating parents one at a time
+// and checking each via a point lookup on the child's unique FK index.
+// Returns (doc, true, nil) for an orphan, (empty, false, nil) when exhausted.
+func (n *orphanNode) nextOrphanByPointLookup() (_ core.Doc, _ bool, err error) {
+	if n.pointLookupDone {
+		return core.Doc{}, false, nil
+	}
+
+	if n.parentClone == nil {
+		if err := n.initPointLookupState(); err != nil {
```
AndrewSisley: question: Why are you making this call here, instead of in `n.init()`?

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+		n.pointLookupDone = true
+		return nil
+	}
+
+	childRelFieldName := n.join.childSide.relFieldDef.Value().Name
+	n.childFKFieldName = request.ToFieldID(childRelFieldName)
+	n.childFKFieldMapIdx = n.join.childSide.relIDFieldMapIndex.Value()
+
+	parentScan := getNode[*scanNode](n.join.parentSide.plan)
+	if parentScan == nil {
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+	n.execInfo.fetches.Add(scan.execInfo.fetches)
+
+	return docs, nil
+}
+
+// initPointLookupState initializes the parent iterator clone and child lookup info
+// for streaming orphan detection via point lookups. Called once on first need.
+func (n *orphanNode) initPointLookupState() error {
+	if !n.join.childSide.relFieldDef.HasValue() || !n.join.childSide.relIDFieldMapIndex.HasValue() {
+		n.pointLookupDone = true
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+			return core.Doc{}, false, err
+		}
+		if n.pointLookupDone {
+			return core.Doc{}, false, nil
+		}
+	}
+
+	childScan := getNode[*scanNode](n.join.childSide.plan)
+	if childScan == nil {
+		n.pointLookupDone = true
```
AndrewSisley: todo: Please document when and why this can happen.

internal/planner/orphan.go
```diff
@@ -0,0 +1,498 @@
...
+			docMapping:          childScan.documentMapping,
+		})
+
+		lookupClone := childScan.cloneWithFilter(lookupFilter, childResult.index)
+		initErr := lookupClone.Init()
+		if initErr != nil {
+			return core.Doc{}, false, initErr
+		}
+
+		hasChild, nextErr := lookupClone.Next()
```
AndrewSisley: todo: I do not understand this - why are you trying to fetch the entire child via a full scan-node, instead of just performing a `datastore.Has()` call on the index?

Please document, and reply with the answer to this comment.

internal/planner/filter/split.go
islamaliev: It's not an optimization. It's a bug fix. Before this change, SplitByFields was always called with a single field, so splitF was always non-nil when reaching the loop. But getFieldsToSplitForTypeJoin now passes 2 fields: the relation field AND the secondary FK field. That's why it's here

internal/planner/filter/split.go
islamaliev: if was just a defensive fix. But you are right, we don't need it. I removed

internal/planner/orphan.go
islamaliev: I return it now

internal/planner/orphan.go
islamaliev: you are right. This is not efficient. I think I copy&pasted if from other place. Thanks for spotting this

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: They are just defensive mechanisms. There is no case that I can think of. I don't have preference, I guess it's just a habit. I will remove them

internal/planner/orphan.go
islamaliev: Good catch. Adjusted

internal/planner/orphan.go
islamaliev: I had them split already once that that felt awkward, so I merged them.
Let me have a look at this again. Maybe after other changes I did it would make more sense now

internal/planner/orphan.go
islamaliev: I made them stream in one place and forgot about this one. Thanks for spotting it.

internal/planner/orphan.go
islamaliev: I checked at the state again and now it looks much more suitable for split than it wa before. So I did it.

internal/planner/planner.go
islamaliev: inNestedJoin isn't always true where it's read. expandTypeJoin saves and restores the old value, so by the time expandJoin checks it after the call returns, it's back to whatever it was before. For a top-level join it's false, for a nested one it's true, that's how we decide whether to wire orphan nodes directly or defer via pendingOrphanWiring.

internal/planner/planner.go
```diff
@@ -286,15 +302,93 @@ func (p *Planner) expandMultiNode(multiNode MultiNode, parentPlan *selectTopNode
...
+					// Wrap join with orphanNode that handles ordering internally.
+					plan.joinPlan = newOrphanNodeWithSource(join, node, orderDir.Value())
+				}
+			} else if parentPlan != nil {
+				p.joinExpand.pendingOrphanWiring = &orphanWiringRequest{
+					join:         join,
+					direction:    orderDir.Value(),
+					useExclusion: !join.parentSide.isPrimary(),
+				}
+			}
```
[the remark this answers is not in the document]
AndrewSisley: @islamaliev I'm a bit tired and inattentive atm, but it sounds like the rabbit has found something - please resolve this before merge, either by changing the code, or documenting why it is desirable.  Unless it's talking nonsense of course :)

AndrewSisley (approved):
LGTM, thanks Islam!

Please have a look at the code-rabbit comment in planner that I've tagged you in before merge.

internal/planner/planner.go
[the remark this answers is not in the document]
islamaliev: done

PR #4457, opened 2026-02-01
fix: Field short ID retrieval

### LINKED ISSUE #4456: Field short ID retrieval broken when more than 9 collections exist
There is a bug where having more than 9 collections causes field short ID assignment to break, resulting in potentially the wrong value being assigned to certain fields and some fields bing missed entirely. This is due to fieldID keys using strings for integer values.

To explain this further, let's use the following schemas:
```gql
type Foo1 {
  about: String
  name: String
}
type Foo2 {
  name: String
}
...
type Foo10 {
  name: String
}
```
This will result in the following system store field short ID KVs:
```
# notice the key order
/s/field/shortID/1/about_field_CID: 2
/s/field/shortID/1/name_field_CID: 3
/s/field/shortID/10/name_field_CID: 2
/s/field/shortID/2/name_field_CID: 2
...
```
When getting the shortIDs for a given collection, the production code used an iterator to fetch all the collection short field IDs at once and then caches the result. This is an optimization to avoid retrieving the short IDs multiple times for a given transactions. 

This behaviour results in the following outcome when fetching field short IDs for collection `Foo1`:
```
# Iterate on prefix "/s/field/shortID/1"
/s/field/shortID/1/about_field_CID: 2 --cache--> about_field_CID: 2
/s/field/shortID/1/name_field_CID: 3 --cache--> name_field_CID: 3
/s/field/shortID/10/name_field_CID: 3 --cache--> name_field_CID: 2
# Hit "/s/field/shortID/2" so stop iterating
```
Resulting cache:
```
# Notice that "about_field_CID" is not referenced. 
name_field_CID: 3
name_field_CID: 2
```
When mutating a `Foo1` document, the input values get parsed into a map. This map is then iterated over to extract the values and store them into the database. The nature of maps make the iteration give the key-values in a random order. So depending on when each of the key-values gets handled, the value stored to the datastore will be different. The last one will win in this case. 

The outcome of this is demonstrated by the following integration test:
```go
func TestMutationCreate_With10Collections(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type Foo1 {
						# The name used for the fields is important as the field shortID
						# is serially assigned based on alphabetical order of field names.
						about: String
						name: String
					}
					type Foo2 {
						name: String
					}
					type Foo3 {
						name: String
					}
					type Foo4 {
						name: String
					}
					type Foo5 {
						name: String
					}
					type Foo6 {
						name: String
					}
					type Foo7 {
						name: String
					}
					type Foo8 {
						name: String
					}
					type Foo9 {
						name: String
					}
					type Foo10 {
						name: String
					}
				`,
			},
			&action.Request{
				Request: `mutation {
					create_Foo1(input: {about: "something", name: "John"}) {
						about
						name
					}
				}`,
				Results: map[string]any{
					"create_Foo1": []map[string]any{
						{
							"about": nil,
							"name":  testUtils.AnyOf("John", "something"),
						},
					},
				},
			},
		},
	}

}
```

To solve this we need to encode the integer values in the key.

fredcarle:
## Relevant issue(s)

Resolves #4456 

## Description

This PR fixes an bug that occurred when more than 9 collections were added to DefraDB. Having 10+ collections caused the field short ID retrieval to mishandle the fieldID assignments which then caused some fields to be ignore even when present in the input object and some fields to be randomly assigned a value depending on the input values.

## Tasks

## How has this been tested?

Added an integration test with more than 9 collections

- MacOS

### DIFF
--- a/internal/encoding/int.go
+++ b/internal/encoding/int.go
@@ -234,7 +234,7 @@ func EncodeUvarintDescending(b []byte, v uint64) []byte {
 // buffer. The remainder of the input buffer and the decoded uint64
 // are returned.
 func DecodeUvarintAscending(b []byte) ([]byte, uint64, error) {
-	if len(b) == 0 {
+	if len(b) == 0 || b[0] == byte('/') {
 		return nil, 0, NewErrInsufficientBytesToDecode(b, "uvarint")
 	}
 	length := int(b[0]) - intZero
--- a/docs/data_format_changes/i4456-field-short-id-retrieval.md
+++ b/docs/data_format_changes/i4456-field-short-id-retrieval.md
@@ -1,3 +1,3 @@
 # Field Short ID Retrieval Bug Fix
 
-Fixed an issue where retrieving field short IDs could result in a corrupted datastore state. The bug occured when more than 9 collections were present as a result of the collection short id being written in the key as a string. Encoding the collection short id as a variable length integer resolves the issue. This however as is changes the way field short ids are stored in the system store.
+Fixed an issue where retrieving field short IDs could result in a corrupted datastore state. The bug occurred when more than 9 collections were present because the collection short ID was written in the key as a string. Encoding the collection short ID as a variable-length integer resolves the issue. This, however, changes how field short IDs are stored in the system store..
--- a/docs/data_format_changes/i4456-field-short-id-retrieval.md
+++ b/docs/data_format_changes/i4456-field-short-id-retrieval.md
@@ -1,0 +1,3 @@
+# Field Short ID Retrieval Bug Fix
+
+Fixed an issue where retrieving field short IDs could result in a corrupted datastore state. The bug occured when more than 9 collections were present as a result of the collection short id being written in the key as a string. Encoding the collection short id as a variable length integer resolves the issue. This however as is changes the way field short ids are stored in the system store.
--- a/internal/db/id/field.go
+++ b/internal/db/id/field.go
@@ -58,7 +58,7 @@ func GetShortFieldID(
 			break
 		}
 
-		key, err := keys.NewFieldIDFromString(string(iter.Key()))
+		key, err := keys.NewFieldIDFromBytes(iter.Key())
 		if err != nil {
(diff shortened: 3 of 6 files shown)


internal/encoding/int.go
```diff
@@ -234,7 +234,7 @@ func EncodeUvarintDescending(b []byte, v uint64) []byte {
 // buffer. The remainder of the input buffer and the decoded uint64
 // are returned.
 func DecodeUvarintAscending(b []byte) ([]byte, uint64, error) {
-	if len(b) == 0 {
+	if len(b) == 0 || b[0] == byte('/') {
```
jsimnz, line 237: question: What is this additional case catching?

jsimnz (approved):
LGTM, ezpz. Was this annoying to debug/track down?

internal/encoding/int.go
fredcarle: It's catching the case where you would expect the key to be something like `/prefix/\x89/something` but receive something like `/prefix//`. Otherwise `DecodeUvarintAscending` returns `0xffffffa7` for the uint64

fredcarle:
> Was this annoying to debug/track down?

@jsimnz, it was quite annoying to track down because the error was returned from deep into the call stack with no stack trace. But that was only part of the troubleshooting. I then had to figure out how defra got to the point of trying to validate a docID from a datetime value (this was happening with Kyle's scenario). Looking at the field descriptions and short IDs when comparing the working schemas with the broken schemas, I notice that the short IDs had changed for some of the fields. This is what led me to believe there was something to do with the short ID storage/retrieval.

internal/encoding/int.go
jsimnz: Gotcha, thanks for the heads up. Based on this I wouldve suggested a brief doc/commenting describing the fact, since the context of why/how we decode something and its respective edge cases can get a bit lost/hazy in the weeds of the actual implementation. 

But, you've already merged it 🤡, so its all good. 👍

PR #4473, opened 2026-02-04
test: Add concurrent transaction mutation tests

### LINKED ISSUE #4284: No tests for concurrent transaction mutations to same document

SharanRP:
## Relevant issue(s)

Resolves #4284

## Description

Added integration tests to verify transaction conflicts when concurrent transactions mutate the same document. While `Update vs Update` was already covered in existing tests, this PR adds missing coverage for:
- Update vs Delete
- Delete vs Update
- Delete vs Delete
- Create vs Create (identical content)

These tests ensure database consistency and correct conflict detection across all mutation combinations.

## Tasks

## How has this been tested?

The changes were verified by running the newly created integration tests specifically targeting transaction conflicts.

To reproduce:
```bash
go test -v ./tests/integration/mutation/conflict/...
```

- Linux

### DIFF
--- a/tests/integration/mutation/conflict/txn_conflict_test.go
+++ b/tests/integration/mutation/conflict/txn_conflict_test.go
@@ -20,11 +20,17 @@ import (
 	"github.com/sourcenetwork/defradb/tests/state"
 )
 
+// TestTxnConflict_UpdateVsDelete tests that when two concurrent transactions
+// target the same document — one updating and one deleting — the second
+// transaction to commit fails with a transaction conflict.
 func TestTxnConflict_UpdateVsDelete(t *testing.T) {
 	test := testUtils.TestCase{
+		// LevelDB does not support concurrent transactions.
+		// TODO https://github.com/sourcenetwork/defradb/issues/4442
 		SupportedDatabaseTypes: immutable.Some([]state.DatabaseType{
 			testUtils.BadgerFileType,
 			testUtils.BadgerIMType,
+			testUtils.DefraIMType,
 		}),
 		Actions: []any{
 			&action.AddSchema{
@@ -36,6 +42,7 @@ func TestTxnConflict_UpdateVsDelete(t *testing.T) {
 				`,
 			},
 			&action.CreateDoc{
+				CollectionID: 0,
 				Doc: `{
 					"name": "John",
 					"age": 27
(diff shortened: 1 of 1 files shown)


tests/integration/mutation/conflict/txn_conflict_test.go
```diff
@@ -0,0 +1,275 @@
...
+			&action.CreateDoc{
+				Doc: `{
+					"name": "John",
+					"age": 27
+				}`,
+			},
+			&action.Request{
+				TransactionID: immutable.Some(0),
+				Request: `mutation {
+					delete_User(docID: "bae-bb8ed746-4570-5651-ac69-39a21f733211") {
```
AndrewSisley, line 148: todo: Please do not use the `Request` action for create/update/delete actions - doing so prevents the test framework from automatically testing this via different pathways

AndrewSisley (changes requested):
Hi Sharan,

I think you have misunderstood the ticket (or I have :)) - this is asking for testing that these mutations function correctly when performed concurrently on different threads.

We have some test gaps with single threaded transactions - I think create is missing them, but IIRC delete and update already have these tests (I might be wrong).

SharanRP:
Hi @AndrewSisley , thanks for the feedback, I want to make sure I get this right.

I looked into using `action.Parallel` for true goroutine-based concurrency, but I ran into a few framework constraints. `UpdateDoc` and `DeleteDoc` are plain structs in the integration package, they don't implement the Action interface (no Execute() method), so they can't be used as children of Parallel. `CreateDoc` also doesn't have a `TransactionID` field, so there's no way to bind it to a specific transaction. For the same reason, the only way to perform a mutation within an explicit transaction is via `&action.Request{}` with `TransactionID` , which is the approach used in the existing `TestMutationWithTxnDoesNotAllowUpdateInSecondTransactionUser`  test in `with_txn_test.go`.

 Should I  keep the current approach (two open transactions, sequential mutations, conflict on commit) since it does test concurrent transaction conflict detection,or should I look into upgrading `UpdateDoc/DeleteDoc` to implement Action so they can run inside Parallel? Happy to go either route, just want to align on the right approach before proceeding.

AndrewSisley:
Hey Sharan, sorry about the bother (again), but this issue has been resolved via other means and I am closing this PR.

PR #4474, opened 2026-02-04
feat(p2p): Add encryption hint identification for faster block sync

### LINKED ISSUE #4340: Add key identification hint for encrypted blocks
When we receive encrypted blocks from P2P, there's no quick way to know if we can decrypt them. We either look up the `Encryption` block from IPLD Blockstore or request keys from peers via KMS - both are slow when processing lots of blocks.

Idea: generate a hint during encryption and store it in the `Block` struct:

```go
type Block struct {
    // ... existing fields ...
    EncryptionHint []byte  // HMAC(key, collectionVersionID || docID || fieldName)[:16]
}
```

The hint needs to be accessible without fetching the `Encryption` block.

On the receiving side, compute the expected hint for keys we have and compare. Match = probably can decrypt, fetch the `Encryption` block. No match = skip.

This came up in a discussion with the Mizu team - they have a problem with Orbis threshold decryption.

SharanRP:
## Relevant issue(s)

Resolves #4340

## Description

This PR implements an encryption hint system to speed up P2P syncing by avoiding unnecessary network requests. I added a 16-byte HMAC hint to the block structure so nodes can check if they have a matching key locally before asking other peers for help. This new logic allows the P2P sync flow to skip slow network round-trips whenever a key is already available in local storage. I updated the block schema, the KMS service interface, and the P2P synchronization layer to support this optimization.

## Tasks

- [x] I made sure to discuss its limitations such as changes to the block CIDs.

## How has this been tested?

I performed the following tests to verify the changes:
1. **Unit Tests**: Added 18 new tests in `internal/core/block/hint_test.go` covering hint generation, matching, and edge cases.
2. **Integration Tests**: Updated and verified all tests in `tests/integration/encryption/` to ensure the new block format (and updated CIDs) work correctly across the network.

- Linux

### DIFF
--- a/internal/core/block/hint_test.go
+++ b/internal/core/block/hint_test.go
@@ -1,0 +1,278 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package coreblock
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestGenerateEncryptionHint_WithValidInputs_ReturnsHint(t *testing.T) {
+	key := []byte("test-encryption-key-32-bytes!!")
+	docID := []byte("bafyreib...")
+	fieldName := []byte("name")
+
+	hint := GenerateEncryptionHint(key, docID, fieldName)
+
+	if hint == nil {
+		t.Fatal("expected hint to be non-nil")
+	}
+	if len(hint) != EncryptionHintLength {
+		t.Fatalf("expected hint length %d, got %d", EncryptionHintLength, len(hint))
+	}
+}
+
+func TestGenerateEncryptionHint_WithEmptyKey_ReturnsNil(t *testing.T) {
+	docID := []byte("bafyreib...")
+	fieldName := []byte("name")
+
+	hint := GenerateEncryptionHint(nil, docID, fieldName)
+
+	if hint != nil {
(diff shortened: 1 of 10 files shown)


internal/core/block/store.go
```diff
@@ -83,6 +83,13 @@ func AddDelta(
 			return cidlink.Link{}, nil, err
 		}
 		dagBlock.Encryption = &encLink
+
+		// Generate encryption hint for quick key matching during P2P sync
+		fieldNameBytes := []byte{}
+		if encBlock.FieldName != nil {
+			fieldNameBytes = []byte(*encBlock.FieldName)
+		}
+		dagBlock.EncryptionHint = GenerateEncryptionHint(encBlock.Key, encBlock.DocID, fieldNameBytes)
```
islamaliev: thought: I'm not sure if doing this unconditionally is the best solution. I think we can make it add hint by default, but add a config option for node to opt-out.

This is additional 16 bytes per block after all that users might not want to "pay" for.

internal/core/block/hint_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if hint == nil {
+		t.Fatal("expected hint to be non-nil")
+	}
+	if len(hint) != EncryptionHintLength {
+		t.Fatalf("expected hint length %d, got %d", EncryptionHintLength, len(hint))
+	}
+}
+
+func TestEncryption_MatchesHint_WithMatchingHint_ReturnsTrue(t *testing.T) {
```
islamaliev: nitpick: we try to stick to naming format `<prefix>_<condition>_<expectation>`

AndrewSisley:
Hi Sharan, sorry about the bother here but we have decided to go with another solution.  Thanks for giving it a crack!

PR #4478, opened 2026-02-05
fix(cli): Standardize command use strings

### LINKED ISSUE #1651: Make the `Use` string standard across the CLI commands
We have a few different ways of representing args and flags in the CLI. We should standardize.

```
create -c --collection <collection> --fields <fields> [-n --name <name>]
delete [-f, --full | -c, --collection] <peer>
set [src] [dst] [cfg]
```

SharanRP:
## Relevant issue(s)

Resolves #1651

## Description

Standardized the `Use` string across CLI commands by removing redundant flag definitions. The `Use` string now only contains the command name and required positional arguments. Optional flags are automatically listed in the "Flags" section generated by Cobra, resulting in cleaner and more consistent help output.

## Tasks

## How has this been tested?

Verified by running help commands for various modified CLI tools (e.g., `defradb client collection create --help`, `defradb client index create --help`) and confirming the `Usage:` line is clean and flags are correctly listed in the "Flags" section.

- Linux

### DIFF
--- a/cli/acp_dac_policy_add.go
+++ b/cli/acp_dac_policy_add.go
@@ -22,7 +22,7 @@ func MakeDocumentACPPolicyAddCommand(ctx context.Context) *cobra.Command {
 	var policyFile string
 
 	var cmd = &cobra.Command{
-		Use:   "add <policy>",
+		Use:   "add [policy]",
 		Short: "Add new policy",
 		Long: `Add new policy
 
--- a/cli/acp_dac_policy_add.go
+++ b/cli/acp_dac_policy_add.go
@@ -22,7 +22,7 @@ func MakeDocumentACPPolicyAddCommand(ctx context.Context) *cobra.Command {
 	var policyFile string
 
 	var cmd = &cobra.Command{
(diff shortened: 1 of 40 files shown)


ChrisBQu:
Thank you for picking up this issue. We actually discussed it in the team standup, and we mostly agree with the approach you took by allowing Cobra to provide the flags. However, we also wanted some additional work to be done in this area. Specifically, we want to make sure all of the parameters listed use the appropriate bracket types (see: http://docopt.org/).

For example, we would like to use ( ) in place of [ ] in cases of non-optional elements, per the standard's description.

Additionally, we would like to see the description of the flags mention that they are required, in those cases.

This does, understandably, increase the scope of this issue. If you would like to expand your work to fit the increased scope, that'd be great. Otherwise, someone on our team will eventually be able to get to it. In either case, we appreciate your contribution and effort.

SharanRP:
I’ve made the requested changes to align the CLI `Use` strings and required flag descriptions with docopt. Please let me know if I missed anything.

PR #4480, opened 2026-02-05
fix(http): Standardize error status codes and centralize logic

### LINKED ISSUE #4101: We have http.StatusNotFound(404) that shouldn't be 404s
We have http.StatusNotFound(404) that shouldn't be 404s

For example in `func CollectionMiddleware` in `http/middleware.go`.

We have `rw.WriteHeader(http.StatusNotFound)`

There are a few other places as well. Perhaps we should also check for errors == `client.ErrNotAuthorizedToPerformOperation` (`"not authorized to perform operation"`) and make the status 401 (unauthorized for those cases everywhere)

SharanRP:
## Relevant issue(s)

Resolves #4101

## Description

This PR fixes a bug where the API would sometimes return the wrong HTTP status codes, like showing a 404 Not Found when the real issue was a security or server error.

Instead of hardcoding status codes in every file, I created a central helper in `http/errors.go`. This helper looks at the error type and picks the right code automatically (like 401 for unauthorized or 400 for bad data). This makes the API more accurate and much easier to maintain. I also added a small helper to our `errors` package to help identify standard library errors like bad JSON or hex formatting.

## Tasks

## How has this been tested?

I verified the changes by running the full HTTP test suite with `go test ./http/...` and confirmed that all status codes are correctly returned. 

- Linux

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -82,6 +82,21 @@ func Is(err, target error) bool {
 	return errors.Is(err, target)
 }
 
+// As finds the first error in err's tree that matches target, and if one is found,
+// sets target to that error value and returns true. Otherwise, it returns false.
+//
+// The tree consists of err itself, followed by the errors obtained by repeatedly
+// calling its Unwrap() error or Unwrap() []error method. When err wraps multiple
+// errors, As examines err followed by a depth-first traversal of its children.
+//
+// An error matches target if the error's concrete value is assignable to the value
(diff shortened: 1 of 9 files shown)


AndrewSisley:
Hey Sharan, sorry about the bother here - it looks like we messed up and created a duplicate ticket that was resolved independently of this PR.

praise: The solution that we went for is very similar to what you proposed, and there would have been a high chance that we would have merged this were someone to have reviewed your PR in a more timely manner.

PR #4512, opened 2026-02-12
fix: Only scan relevant doc-commits in commit queries

### LINKED ISSUE #3846: Only scan relevant doc-commits in commit queries
The DAG fetcher currently scans everything if not given a specific CID/prefix.  This is messy and inefficient.  There is a todo in `fetcher/dag.go` for this.  The returned results is correct at the moment, but we are reading more key-values from the store than we need to.

We likely want to allow users to fetch the collection-commits via commit queries, but we may not want them bundled in with doc commits.

SharanRP:
## Relevant issue(s)

Resolves #3846

## Description

Limit the DAG head fetcher to only scan collection and document commit prefixes when no prefix is provided, so non‑commit headstore keys are not scanned. No new dependencies.

## Tasks

## How has this been tested?

- `go test ./tests/integration/query/commits/... `
- `go test ./internal/db/fetcher/ ./internal/keys/ `

-  Linux


internal/db/fetcher/dag.go
```diff
@@ -24,69 +25,111 @@ import (
...
-	// https://github.com/sourcenetwork/defradb/issues/3846
-	switch headStoreKey.(type) {
-	case keys.HeadstoreFieldDefinition, keys.HeadstoreCollectionDefinition, keys.HeadstoreCollectionSetDefinition:
-		return hf.FetchNext()
+		// Skip non-commit keys (e.g. schema definitions) if they happen to be in the scan range.
+		// Normally the caller restricts the scan range to collection/doc commits,
+		// so this is just a defensive check.
+		switch headStoreKey.(type) {
+		case keys.HeadstoreFieldDefinition, keys.HeadstoreCollectionDefinition, keys.HeadstoreCollectionSetDefinition:
+			return hf.FetchNext()
```
AndrewSisley: todo: It should be impossible to hit this line, unless the consuming code is changed with the intent of yielding these commits.

Please verify this quickly (maybe by adding temporary a `panic("test")` immediately before this line before running `make test`), and then delete if it is confirmed to be dead code.

If it is not, please describe how it is hit, both in reply to this comment, and with a code-comment.

internal/db/fetcher/dag.go
```diff
@@ -24,69 +25,111 @@ import (
...
 	}
 
-	cid := headStoreKey.GetCid()
-	return &cid, nil
+	return nil, nil
 }
 
 func (hf *HeadFetcher) Close() error {
-	if hf.kvIter == nil {
+	if len(hf.kvIters) == 0 {
```
AndrewSisley: todo: This if-early-return block is no longer useful, if len==0, then the code below will return nil anyway.  The performance difference does not matter here (if any).

AndrewSisley (approved):
Looks good to me! Sorry I took so long to review this, I saw it when really busy and then forgot about it.

Thank you very much for this Sharan!

AndrewSisley:
sorry one more thing to do before merge, the formatting (whitespace, by the looks of it) is off - I think running `make lint:fix` will sort it out for you if you dont want to do it manually. I think the Go plugin for visual studio code will also sort it out for you if you edit and re-save the file.

internal/db/fetcher/dag.go
SharanRP: I followed your suggestion and verified it by adding a temporary panic, it was never hit during the tests....so I’ve gone ahead and removed the unreachable code. Thanks for the heads-up!!

PR #4528, opened 2026-02-14
feat: Gate sync documents op with NAC

### LINKED ISSUE #4525: NAC gate SyncDocuments
NAC gate SyncDocuments

shahzadlone:
## Relevant issue(s)
Resolves #4525

## Description
- Gate sync documents op with NAC

### DIFF
--- a/internal/db/p2p.go
+++ b/internal/db/p2p.go
@@ -339,7 +339,20 @@ func (db *DB) ListP2PDocuments(
 // context.WithTimeout can be used to set a timeout for the operation.
 //
 // WARNING: This function does not respect transactions.
-func (db *DB) SyncDocuments(ctx context.Context, collectionName string, docIDs []string) error {
+func (db *DB) SyncDocuments(
+	ctx context.Context,
+	collectionName string,
+	docIDs []string,
+	opts ...options.Enumerable[options.SyncDocumentsOptions],
+) error {
+	opt := utils.NewOptions(opts...)
+
+	if err := db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeP2PSyncDocumentsPerm); err != nil {
+		return err
+	}
+
+	ctx = identity.WithContext(ctx, opt.Identity)
+
 	if db.p2p == nil {
 		return ErrNoP2P
 	}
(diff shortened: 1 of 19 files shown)


internal/db/p2p/sync_doc.go
```diff
@@ -86,7 +90,7 @@ func (p *P2P) syncDocuments(
 	}
 
 	if len(activePeers) == 0 {
-		return nil, ErrTimeoutDocSync
```
shahzadlone, line 89: @sourcenetwork/database-team Was this a bug or intentional? When node restarts there would be 0 active peers and tests otherwise would keep timing out. Should I keep it as it was an instead be reconnecting?

internal/db/p2p.go
```diff
@@ -335,7 +335,20 @@ func (db *DB) ListP2PDocuments(
...
+	docIDs []string,
+	opts ...options.Lister[options.SyncDocumentsOptions],
+) error {
+	opt := utils.NewOptions(opts...)
+
+	if err := db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeP2PSyncDocumentsPerm); err != nil {
+		return err
+	}
+
+	ctx = identity.WithContext(ctx, opt.Identity)
```
shahzadlone, line 354: question: I am stil not 100% sure why we need to do this after the `checkNodeAccess`. Is there not a better way to do one approach that can then use `identity` for all cases, seems tedious to use `opt.Identity` in some places (even though they also have `ctx`) and then in some other places have to again like before still inject the identity in the `ctx`. Doing here because I saw all other places following this pattern

shahzadlone: @islamaliev LMK if any insights

AndrewSisley: This line will also overwrite any previously set ctx-identity.  Having two (public) places to store and track what *should* be the same information is IMO a code smell, and an easy way for us (and users) to introduce bugs.

AndrewSisley: Raised on discord

AndrewSisley (approved):
LGTM

internal/db/p2p/sync_doc.go
AndrewSisley: It looks like you've reverted the changes you mentioned in your comment - does this question still need answering?

shahzadlone: Yes I have. You can still answer if you'd like :)

Issue #4268, opened 2025-12-11
Concurrent writes during a collection version delete succeed and break commits api


AndrewSisley:
I was 'fairly confident' that the transaction system would prevent this, however it does not (at least with badger stores).

Instead of the create-doc calls erroring out, they are allowed to complete after the last collection version has been deleted.

This orphans blocks, and cause any subsequent calls to the commits API (without collection filters at least) to fail with a `key not found` error as the planner tries to populate the orphaned doc commits to a collection that no longer exists.

This issue probably also affects document writes received over P2P and other means (import?), but that has not been tested.

This issue is documented by `TestColVersionUpdateRemoveCollections_ConcurrentWrite` and there is a code-comment referencing this issue in the WIP branch.

AndrewSisley:
WIP branch `4268-rm-col-v-concurrent-write`.

I believe the problem is bigger than just this bug, but it is taking me a long time to fully understand it.  I'll update this comment with more info in a bit.  If we really need to, we can patch over the most visible symptoms without resolving the underlying problems (planner/commits.go, errors fetching collection that no longer exists).

Known issues:
- Ctx-caches can be constructed before the col-delete, and then used to insert docs after the col-delete (different txn)
- The store based cache/sids are cleared, but the in-memory caches are not, as they are isolated/scoped to ctx-txn

WIP thoughts:
> short ids are unrecoverable, and cannot be deleted unless you can be 100% sure there are no references to them anywhere
    - datastore references collection short ids, so the collection sid cannot be deleted, unless we lock the col-datastore
    - ~~if we do not ever delete short ids, then we never need to lock the references to them, even if the full definition of what the sid represents is deleted~~-- this would not help when deleting refs to the sids...
    - OR, we can allow stuff to have multiple short ids, indexed in the key-part: [cid]/[sid] to avoid mutating a single key-value (if doing [cid] => [,]) - a similar structure could be used in the in-memory caches 
>
> this problem probably also affects secondary indexes, and maybe (less certain) the headstore, as stuff that they reference is mutable.
> 
> Solution should ideally be able to handle this issue at a document level (d-sid) and others as well as col level
> 
> Lock system remains desirable, otherwise certain combinations become really painful (truncate+col-delete), unless... we decouple them, and allow the truncation of unknown collection versions!
>
[...]

PR #4546, opened 2026-02-17
fix(i): Lock collection when deleting collection versions

### LINKED ISSUE #4511: Deleting the last remaining collection version should lock collection
Deleting the last remaining collection version should lock the collection, and cause competing actions to fail with a transaction conflict error.

Related to https://github.com/sourcenetwork/defradb/issues/4268 and https://github.com/sourcenetwork/defradb/issues/4510

Deleting fields also needs to lock, as their sids are being deleted.

AndrewSisley:
## Relevant issue(s)

Resolves #4511

## Description

Locks the collection when deleting collection and field short ids.

Also fixes a bug where collection short ids weren't actually getting deleted when the last collection version was deleted.

This is currently a pain to test due to the caching issue described in https://github.com/sourcenetwork/defradb/issues/4268 - the CI will make sure that no regressions are introduced, but they do not (yet) validate that the bug has been fixed.  Tests will be added during #4268.

### DIFF
--- a/internal/db/description/collection.go
+++ b/internal/db/description/collection.go
@@ -504,7 +504,7 @@ func DeleteCollection(
 		return err
 	}
 
-	if len(versions) == 0 {
+	if len(versions) == 1 {
 		// Only delete the collection short ID if this was the last local version
 		err = id.DeleteShortCollectionID(ctx, version.CollectionID)
 		if err != nil {
--- a/internal/db/description/collection.go
+++ b/internal/db/description/collection.go
@@ -22,6 +22,7 @@ import (
 	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/id"
+	"github.com/sourcenetwork/defradb/internal/db/lock"
 	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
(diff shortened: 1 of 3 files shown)


internal/db/description/collection.go
```diff
@@ -499,12 +505,16 @@ func DeleteCollection(
...
+	err = id.DeleteShortFieldIDs(ctx, lockSet, version, versions)
 	if err != nil {
 		return err
 	}
 
-	if len(versions) == 0 {
+	if len(versions) == 1 {
+		// It is impossible to recreate the collection short ID once it is deleted, so we must lock the collection
+		// whilst we finalize this operation, otherwise other threads/operations may try and make use of it.
+		lockSet.CollectionLock(txn, shortID)
```
nasdf, line 516: question: Could this lock be acquired before `DeleteShortFieldIDs` so that we don't need two different locks?

internal/db/description/collection.go
```diff
@@ -499,12 +505,16 @@ func DeleteCollection(
 
 	// WARNING - DeleteShortFieldIDs is dependent on the collection short id still existing, it should be called
 	// before deleting the collection short id.
-	err = id.DeleteShortFieldIDs(ctx, version, versions)
+	err = id.DeleteShortFieldIDs(ctx, lockSet, version, versions)
 	if err != nil {
 		return err
 	}
 
-	if len(versions) == 0 {
```
nasdf, line 507: question: Was this an existing bug?

nasdf (approved):
LGTM

internal/db/description/collection.go
AndrewSisley: > two different locks

I'm not completely sure if you mean two different locks, or two different calls to acquire (a) lock(s).

Field short ids get deleted if the last collection that contains them gets deleted, so they can be deleted, and the lock required, even if there remain collection versions after this collection version is deleted.  So two calls to acquire (a) lock(s) are required.

Two (or more) locks will never actually be acquired.  The lock system ensures that each transaction can only ever hold a single lock, so if both calls to `CollectionLock` are made, only one lock will be held.  This is important to prevent deadlocks, such as the one that would occur here if two locks for the same collection were requested by the same thread.

internal/db/description/collection.go
AndrewSisley: Yes, sorry, I should have noted that in the PR description.  It was documented in the commit message, but *that* documentation was not noted in the PR description.

I'll edit the PR description before merge for tracking - thanks for flagging :)

PR #4549, opened 2026-02-18
refactor: Move identity context funcs to internal

### LINKED ISSUE #4548: Make identity.WithContext internal
It is no longer needed by users and might suggest to them that they should be using it (esp. if they scan through our code)

AndrewSisley:
## Relevant issue(s)

Resolves #4548

## Description

Moves the identity context funcs to internal.

Nothing really coded here, just moved the file from `/acp/identity` to `/internal/identity`.  Most locations that import both import  the internal dependency as `iIdentity "github.com/sourcenetwork/defradb/internal/identity"`, a handful omit the alias (in keeping with whatever it was doing before, most places aliased `/acp/identity` to `acpIdentity` to avoid local variable naming collisions.

### DIFF
--- a/cbindings/acp.go
+++ b/cbindings/acp.go
@@ -19,9 +19,9 @@ import "C"
 import (
 	"context"
 
-	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
 	"github.com/sourcenetwork/defradb/client"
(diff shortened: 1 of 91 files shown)


islamaliev (approved):
LGTM!

PR #4566, opened 2026-02-20
fix: Serialize schema mutations to prevent race conditions

### LINKED ISSUE #4545: No protection against making multiple `AddSchema` calls concurrently
This might cause the GQL types to forget types, and risks incompatible types being added (duplicate names might have txn protect due to an index).

Needs to share some protection with PatchCollection calls.

SharanRP:
## Relevant issue(s)

Resolves #4545

## Description

Concurrent `AddSchema` calls (and related schema-mutating operations) had no concurrency protection, causing races on the `parser.schemaManager` pointer swap inside `txn.OnSuccess`. This could silently lose registered GQL types or produce incompatible duplicates.

A `gqlLock sync.RWMutex` is added to `DB`. Schema-mutating operations acquire a write lock held through `txn.Commit()`, ensuring the `OnSuccess` pointer swap is fully serialized. Concurrent read paths (`execRequest`, `NewFilterFromString`) acquire a read lock to prevent racing against an in-progress mutation.

## Tasks

## How has this been tested?

Two new tests in `internal/db/schema_concurrent_test.go` run under `-race`: one fires 10 concurrent `AddSchema` writers, another combines 10 writers with 10 concurrent `ExecRequest` readers.

```shell
go test -race -run "TestAddSchema_Concurrent" ./internal/db/ -v

### DIFF
--- a/internal/db/collection_update.go
+++ b/internal/db/collection_update.go
@@ -164,11 +164,10 @@ func (c *collection) makeSelectionPlan(
 			return nil, ErrInvalidFilter
 		}
 
-		c.db.gqlLock.RLock()
-		f, err = c.db.parser.NewFilterFromString(c.Name(), fval)
-		c.db.gqlLock.RUnlock()
-		if err != nil {
-			return nil, err
+		var err2 error
+		f, err2 = c.db.parser.NewFilterFromString(c.Name(), fval)
+		if err2 != nil {
+			return nil, err2
 		}
 	case immutable.Option[request.Filter]:
 		f = fval
--- a/internal/db/collection_update.go
+++ b/internal/db/collection_update.go
@@ -164,7 +164,9 @@ func (c *collection) makeSelectionPlan(
 			return nil, ErrInvalidFilter
 		}
 
+		c.db.gqlLock.RLock()
 		f, err = c.db.parser.NewFilterFromString(c.Name(), fval)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: internal/db/collection_update.go [hunk 1]
```diff
@@ -164,7 +164,9 @@ func (c *collection) makeSelectionPlan(
 			return nil, ErrInvalidFilter
 		}
 
+		c.db.gqlLock.RLock()
```
AndrewSisley: todo: I dont think this, and request.go should be protected by the lock.  Preventing the execution of all GQL requests when a collection is being updated is a huge throughput hit in my opinion.  It will also block all collection updates while GQL requests are being processed (scheduling is up to the runtime, so it could be quite a while).

The switch over from old to new GQL types is nearly atomic, and it can be made so fairly easily (or at least protected by a mutex with a very limited scope) - see line 139 in `internal/request/graphql/parser.go`.

The requests don't need to worry about using a stale GQL type system, so long as the switch-over is atomic and can't be read mid-update.

AndrewSisley (changes requested):
I think it is a great start, but I don't think we can tolerate the locking of the GQL requests whilst the collections are updating.

AndrewSisley:
Hey Sharan, sorry about the bother here, but we are going to proceed with another solution.  The locks in this PR are too long lived for us to accept.

PR #4571, opened 2026-02-23
feat: Accept arrays of CIDs for GQL queries

### LINKED ISSUE #4565: Change type signature of `cid` on get query to be a list
This is a *first step* for #4304, which only changes the type signature for `cid` to be a list. This needs to happen for RC1, so that we don't have "breaking changes" from a type signature perspective.

Importantly, since this issue is not implementing the underlying behavior yet, this task should restrict the list to a single value and throw an error if more values are passed.

### LINKED ISSUE #4568: Change type signature of cid on _commit query to be a list
This is a *first step* for #4303, which only changes the type signature for `cid` on `_commit` to be a list. This needs to happen for RC1, so that we don't have "breaking changes" from a type signature perspective.

Importantly, since this issue is not implementing the underlying behavior yet, this task should restrict the list to a single value and throw an error if more values are passed.

AndrewSisley:
## Relevant issue(s)

Resolves #4565
Resolves #4568

## Description

Accepts arrays of CIDs for GQL queries - both for normal collection requests, and for commit queries.

It does not change the behaviour where an error will be returned if no matching documents are found for a cid.  It does not implement the feature, and an error will be returned if more than one cid is provided.

### DIFF
--- a/client/request/cid.go
+++ b/client/request/cid.go
@@ -15,11 +15,11 @@ import "github.com/sourcenetwork/immutable"
 // CIDFilter is an embeddable struct that hosts a consistent set of properties
 // for filtering an aspect of a request by commit CID.
 type CIDFilter struct {
-	// CID is an optional value that selects a single document at the given commit CID
+	// CIDs is an optional value that selects a single document at each given commit's CID
 	// for processing by the request.
 	//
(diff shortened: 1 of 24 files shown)


fredcarle (approved):
LGTM

PR #4578, opened 2026-02-24
feat: Change GQL commits.docId param to list

### LINKED ISSUE #4567: Change type signature of docID on _commit query to be a list
This is a *first step* for #4302, which only changes the type signature for `docID` on `_commit` to be a list. This needs to happen for RC1, so that we don't have "breaking changes" from a type signature perspective.

Importantly, since this issue is not implementing the underlying behavior yet, this task should restrict the list to a single value and throw an error if more values are passed.

AndrewSisley:
## Relevant issue(s)

Resolves #4567

## Description

Change GQL commits.docId param to list type.  Does not implement the handling of more than one.

### DIFF
--- a/client/request/commit.go
+++ b/client/request/commit.go
@@ -49,6 +49,7 @@ func (c CommitSelect) ToSelect() *Select {
 			Alias: c.Alias,
 		},
 		DocIDsFilter: c.DocIDsFilter,
+		CIDFilter:    c.CIDFilter,
 		Limitable:    c.Limitable,
 		Offsetable:   c.Offsetable,
 		Orderable:    c.Orderable,
--- a/client/request/commit.go
+++ b/client/request/commit.go
@@ -26,16 +26,13 @@ type CommitSelect struct {
 	ChildSelect
 
 	CIDFilter
+	DocIDsFilter
 
 	Limitable
 	Offsetable
 	Orderable
 	Groupable
 	Filterable
-
-	// DocID is an optional filter which when provided will limit commits to those
-	// belonging to the given document.
-	DocID immutable.Option[string]
 
 	// Depth limits the returned commits to being X places in the history away from the
 	// most current.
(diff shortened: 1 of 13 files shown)


client/request/commit.go
```diff
@@ -51,12 +48,14 @@ func (c CommitSelect) ToSelect() *Select {
...
 			Alias: c.Alias,
 		},
-		Limitable:   c.Limitable,
-		Offsetable:  c.Offsetable,
-		Orderable:   c.Orderable,
-		Groupable:   c.Groupable,
-		Filterable:  c.Filterable,
-		ChildSelect: c.ChildSelect,
+		DocIDsFilter: c.DocIDsFilter,
+		CIDFilter:    c.CIDFilter,
```
nasdf, line 52: question: was this a bug? It seems like it was never set before this change.

internal/planner/commit.go
```diff
@@ -255,8 +258,12 @@ func (n *dagScanNode) Next() (bool, error) {
...
 	if n.commitSelect.Cids.HasValue() &&
 		len(n.visitedNodes) == 0 &&
-		n.commitSelect.DocID.HasValue() &&
-		currentDocID != n.commitSelect.DocID.Value() {
+		n.commitSelect.DocIDs.HasValue() &&
+		len(n.commitSelect.DocIDs.Value()) > 0 &&
+		// todo - for now we just take the first docID and ignore the rest, an error
+		// should be thrown in the parser anyway if the user provides more than one.
+		// https://github.com/sourcenetwork/defradb/issues/4302
+		currentDocID != n.commitSelect.DocIDs.Value()[0] {
```
nasdf, line 266: thought: I think the correct behavior for an empty array (as opposed to a null array) should be to match nothing.

fredcarle (approved):
LGTM

tests/integration/query/commits/with_doc_id_test.go
nasdf, line 1: thought: A test for an empty list (as opposed to null) might be useful

nasdf (changes requested):
I'd like to make sure the behavior of an empty list is consistent with what we expect before approving.

client/request/commit.go
AndrewSisley: Not a bug, there was another property used that shadowed this (see 2nd commit).

internal/planner/commit.go
AndrewSisley: I agree, and this was noted on discord.

I've kicked implementing this (and the same for cids) to post rc1.

tests/integration/query/commits/with_doc_id_test.go
AndrewSisley: Not a problem, I've added one.

nasdf (approved):

PR #4579, opened 2026-02-24
refactor: Remove GetAllDocIDs/docIDs from collection api

### LINKED ISSUE #4514: Remove the get docIDs/list api
Remove the get docIDs/list api  

https://discord.com/channels/1374835078781468722/1390656042014343261/1471623400354087066

shahzadlone:
## Relevant issue(s)
Resolves #4514 

## Description
Remove the `client collection docIDs` command and `GetAllDocIDs` method from the Collection interface and all client implementations as the first step toward reorganizing collection vs document-level operations.

Note(s):
- GetAllDocIDs wasn't fully deleted it was made private as getAllDocIDsChan returning a private docIDResult struct. This is because internal/db/backup.go still needs to iterate doc IDs for export.
- So the public API surface was removed but the internal functionality was preserved.
- Zero test files changed (no tests depended on GetAllDocIDs directly).


nasdf (approved):
Changes look good to me. 

Unrelated question: should we also consider removing the document get / create APIs? Both can be achieved with the GQL query interface.

AndrewSisley (approved):
LGTM, the unchecked `(*collection)` cast makes me a little uncomfortable, but we are pushed for time.

shahzadlone:
> Unrelated question: should we also consider removing the document get / create APIs? Both can be achieved with the GQL query interface.

I thought we had consensus and did want to cleanup/remove most of the API anyways because they can all be achieved with GQL soon? 

Or are you suggesting pre-v1?

AndrewSisley:
> Unrelated question: should we also consider removing the document get / create APIs? Both can be achieved with the GQL query interface.

I think that discussion is something easily saved for v2.

shahzadlone:
> the unchecked `(*collection)` cast makes me a little uncomfortable, but we are pushed for time.

- [x] Fixed in the 2nd commit

internal/db/backup.go
```diff
@@ -187,7 +187,11 @@ func (db *DB) basicExport(ctx context.Context, config *client.BackupConfig) (err
 		if err != nil {
 			return err
 		}
-		docIDsCh, err := col.GetAllDocIDs(ctx)
+		internalCol, ok := col.(*collection)
+		if !ok {
+			return NewErrInvalidCollectionType(col.Name())
+		}
```
fredcarle: thought: I understand that @AndrewSisley was nervous about the unchecked casting but this is effectively dead code. There isn't much of a benefit to having it here.

shahzadlone: I think you are right, basicExport is on *DB which always creates *collection internally, so the cast can never fail.

I will just revert the change

AndrewSisley: I was a bit conflicted over this, and didn't add a comment to this line partly because of this.  I am more nervous because Chris might be changing something here, and it is not implausible for us to have more than one implementation of the interface in the future.

I figured it was not worth thinking too hard about it right now, so I have no objection to this being reverted if people prefer.

Issue #2413, opened 2024-03-14
Implement the Collection Exists Function


shahzadlone:
Implement the cli and http Collection Exists Function that already has a public method on `client.Collection` (in client/collection.go).

It should respect ACP logic and only show true for documents that exist and are accessible.

Here is the public function we have so far (after acp is merged):
```
// Exists checks if a given document exists with supplied DocID.
//
// Will return true if a matching document exists, otherwise will return false.
Exists(ctx context.Context, identity immutable.Option[string], docID DocID) (bool, error)
```

PR #4581, opened 2026-02-24
refactor: Rename document-level cmds under collection api

### LINKED ISSUE #3993: Rename document specific client collection commands to be more consistent

### LINKED ISSUE #4577: Deleting index without name hits wrong http api and gives bad error
Deleting index without name hits wrong http api and gives bad error

Current: "malformed document ID"
Should be: "can not delete index without index name" (or something like that)

```
func TestIndexDelete_WithoutName_ShouldReturnError(t *testing.T) {
    test := &integration.Test{
        Actions: []action.Action{
            &action.SchemaAdd{
                InlineSchema: `
                    type User {
                        name: String
                    }
                `,
            },
            &action.IndexDelete{
                Collection: "User",
                // Name is empty
                ExpectError: "malformed document ID",
            },
        },
    }

    test.Execute(t)
}
```

shahzadlone:
## Relevant issue(s)
- Resolves #3993 
- Resolves #4577

## Description

- Rename `client collection {add,delete,get,update}` to `client document {add,delete,get,update}`

- Some re-organizing to move document-level stuff under separate files.

- Also fixed an http route clashing bug when index delete command would get empty name.


cli/test/integration/index/delete_test.go
```diff
@@ -102,7 +102,7 @@ func TestIndexDelete_WithoutName_ShouldReturnError(t *testing.T) {
 			&action.IndexDelete{
 				Collection: "User",
 				// Name is empty
-				ExpectError: "malformed document ID",
+				ExpectError: "index name is required",
```
shahzadlone, line 105: for reviewers: #4577 was fixed in this test here.

http/handler_collection.go
```diff
@@ -607,17 +611,27 @@ func (h *collectionHandler) bindRoutes(router *Router) {
 	router.AddRoute("/collections/{name}", http.MethodPost, collectionAdd, h.Add)
 	router.AddRoute("/collections/{name}", http.MethodPatch, collectionUpdateWith, h.UpdateWithFilter)
 	router.AddRoute("/collections/{name}", http.MethodDelete, collectionDeleteWith, h.DeleteWithFilter)
+	deleteIndexWithoutName := openapi3.NewOperation()
+	deleteIndexWithoutName.OperationID = "index_delete_without_name"
+	deleteIndexWithoutName.Description = "Returns an error when no index name is provided"
+	deleteIndexWithoutName.Tags = []string{"index"}
+	deleteIndexWithoutName.AddParameter(collectionNamePathParam)
+	deleteIndexWithoutName.Responses = openapi3.NewResponses()
```
shahzadlone: for reviewers: #4577 related change

http/handler_collection.go
```diff
@@ -280,6 +280,10 @@ func (h *collectionHandler) ListIndexes(rw http.ResponseWriter, req *http.Reques
 	responseJSON(rw, http.StatusOK, indexes)
 }
 
+func (h *collectionHandler) DeleteIndexNoName(rw http.ResponseWriter, req *http.Request) {
+	responseJSON(rw, http.StatusBadRequest, errorResponse{fmt.Errorf("index name is required")})
+}
```
shahzadlone: for reviewers: related to #4577

cli/document.go
```diff
@@ -0,0 +1,90 @@
...
+
+func MakeDocumentCommand(ctx context.Context) *cobra.Command {
+	var txID uint64
+	var identity string
+	var collection string
+	var collectionID string
+	var versionID string
+	var getInactive bool
+	var cmd = &cobra.Command{
+		Use:   "document [--collection <name> --collection-id <collectionID> --version-id <versionID>]",
```
AndrewSisley: thought: Seeing all the params listed like this makes me want to ask for `collection-name` instead - the PR change feels inconsistent with the other params atm.

http/handler_collection.go
AndrewSisley: @nasdf @ChrisBQu My understanding is that this is going to show up in the openapi output and playground as a valid api and will actually encourage users to call it.

Are you 3 (and anyone else) sure we want this fix?

internal/db/collection_delete.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // DeleteWithFilter deletes using a filter to target documents for delete.
```
AndrewSisley, line 30: todo: Please update the documentation

internal/db/collection.go
AndrewSisley: todo: Please update the documentation (all funcs)

internal/db/collection.go
```diff
@@ -811,7 +811,7 @@ func (c *collection) save(
 // otherwise will return false, along with an error, if it cannot.
 // If the document doesn't exist, then it will return false, and a ErrDocumentNotFound error.
 // This operation will all state relating to the given DocID. This includes data, block, and head storage.
-func (c *collection) Delete(
+func (c *collection) DocumentDelete(
```
AndrewSisley: discussion: @sourcenetwork/database-team I'm not sure we want these Go/JS client renames. The 'Document' name prefix doesn't add any value here in my opinion.

Please raise this on Discord if we don't get a rapid consensus here.

cbindings/wrapper_document.go
```diff
@@ -0,0 +1,460 @@
...
+		cShowDeleted,
+		copts,
+		cIdentity,
+	))
+
+	if res.Status != 0 {
+		return false, errors.New(res.Error)
+	}
+	return true, nil
+}
```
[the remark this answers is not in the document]
shahzadlone: @ChrisBQu Is DocumentExists command broken? I think we have a huge test gap for it. We made this a while back from what i remember: https://github.com/sourcenetwork/defradb/issues/2413 possibly should not do that now as we will soon remove the apis (right @nasdf ?)

AndrewSisley (approved):
I disagree with the renaming of the Go/JS client functions, and would like to check in with the dev-ex guys RE the new endpoint, approving assuming those discussions (and a couple of minor things) are resolved before merge.

cli/document.go
shahzadlone: I also thought about this same point, happy to adjust.

internal/db/collection_delete.go
shahzadlone: - [x] already done haha you are too speedy with the reviews (thanks <3), will push soon.

internal/db/collection.go
shahzadlone: Will raise in discord.

 Raised here: https://discord.com/channels/1374835078781468722/1466897550371061791/1475997703375687832

http/handler_collection.go
shahzadlone: I noted in the PR description, that this is hacky and likely will be moved out of this PR. Still happy to hear where people think I should do this fix.

internal/db/collection.go
shahzadlone: - [x] already done haha you are too speedy with the reviews (thanks <3), will push soon.

cli/document.go
shahzadlone: @AndrewSisley Did some changes here: https://github.com/sourcenetwork/defradb/pull/4581/changes/a075f538f2b7da4447ea04848c58a23a5cc01fd1

LMK if happy or have other suggestions.

http/handler_collection.go
nasdf: I agree this is an odd way to fix the issue. If we renamed these routes I believe the issue would be fixed.

```go
router.AddRoute("/collections/{name}/document/{docID}", http.MethodGet, collectionGet, h.Get)
router.AddRoute("/collections/{name}/document/{docID}", http.MethodPatch, collectionUpdate, h.Update)
router.AddRoute("/collections/{name}/document/{docID}", http.MethodDelete, collectionDelete, h.Delete)
```

shahzadlone: Yes @nasdf but would you be happy with a `405 Method Not Allowed` error? That is what it will give then I believe because the rename to /collections/{name}/document/{docID} prevents the wildcard collision, but the handler gives a proper 400 Bad Request with "index name is required" instead of a generic 405 Method Not Allowed. Both pieces work together.

cli/document.go
AndrewSisley: The linked commit looks very suspicious, you seem to have deleted a lot of code/functionality.  Why have you done this?

http/handler_collection.go
nasdf: What if we returned a client side error if docID is invalid or empty?

shahzadlone: I think you meant if index name is empty? Yea that should work, I did this change.

- [x] Now client checks nameArg == "" and returns client.ErrIndexNameRequired ("index name is required")

cbindings/wrapper_document.go
```diff
@@ -0,0 +1,461 @@
...
+) error {
+	addOpts := utils.NewOptions(opts...)
+	isEncrypted := 0
+	if addOpts.EncryptDoc {
+		isEncrypted = 1
+	}
+	encryptedFields := C.CString("")
+	if len(addOpts.EncryptedFields) > 0 {
+		encryptedFields = C.CString(strings.Join(addOpts.EncryptedFields, ","))
+	}
```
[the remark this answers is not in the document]
shahzadlone: @ChrisBQu What do you think about this?

internal/db/collection.go
shahzadlone: - [x] Reverted names back

cli/document.go
shahzadlone: - [x] Renamed flag to `--collection-name`

shahzadlone:
@coderabbitai resolve

cli/document.go
shahzadlone: Discussed on Discord. Andy seems happy to merge :)

Issue #4308, opened 2026-01-05
Fix and reintroduce flaky tests removed in #4307


AndrewSisley:
#4307

As a quick work-around to this problem we could just extend our testing framework to allow marking test cases as flaky. 
If the test fails the testing framework will retry multiple time (can be configured) until it passes.

Frameworks like [ginkgo](https://onsi.github.io/ginkgo/#repeating-spec-runs-and-managing-flaky-specs) actually allow marking tests as flaky.

islamaliev:
done long ago

PR #4582, opened 2026-02-25
test: Enable marking tests as flaky

islamaliev:
Resoves #4308 

Add `FlakeRetries` field to `TestCase` that signals to testing framework should retry the test multiple times.

Also uncomment current branchable-collection tests and mark the mas flaky.

## Tasks

## How has this been tested?

with integration tests

- MacOS


tests/integration/flake_retry.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (f *flakeTestCapture) Fatalf(format string, args ...any) {
+	f.hasFailed = true
+	f.Logf(format, args...)
+	runtime.Goexit()
+}
+
+func (f *flakeTestCapture) Failed() bool {
+	return f.hasFailed
+}
```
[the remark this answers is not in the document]
islamaliev: Retries are sequential — each goroutine completes (`<-done`) before the next starts, and `flakeTestCapture` is local to each attempt. No concurrent access to `hasFailed`.

tests/integration/utils.go
```diff
@@ -211,21 +211,35 @@ func ExecuteTestCase(
...
+						st,
+						collectionNames,
+						testCase,
+						kms,
+						dbt,
+						ct,
+						documentACPType,
+					)
+					cancel()
+				}
```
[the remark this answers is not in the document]
islamaliev: Fixed — `cancel()` is now deferred.

tests/integration/flake_retry.go
```diff
@@ -0,0 +1,100 @@
...
+		ft := &flakeTestCapture{TB: t}
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			run(ft)
+		}()
+		<-done
+		if !ft.hasFailed {
+			if attempt > 0 {
+				log.InfoContext(context.Background(),
```
AndrewSisley: suggestion: `t.Log()` might be slightly more appropriate in this file.

tests/integration/flake_retry.go
```diff
@@ -0,0 +1,100 @@
...
+	hasFailed bool
+}
+
+func (f *flakeTestCapture) Fail() {
+	f.hasFailed = true
+}
+
+func (f *flakeTestCapture) FailNow() {
+	f.hasFailed = true
+	runtime.Goexit()
```
AndrewSisley: todo: I can't spot a good reason to be implementing the later halves of these functions ourselves, and it creates the impression that we are deliberately diverging from the `testing.TB` implementation without describing why.

Please replace the `runtime.Goexit()`, `f.Log(args...)`, etc calls with `f.TD.FailNow()`, `f.TD.Error(args)`, etc - or document each detailing why we are implementing them ourselves.

tests/integration/flake_retry.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"fmt"
+	"runtime"
+	"testing"
+)
+
+// runTestWithRetry runs the given test function up to maxRetries+1 times.
+// It returns true if any attempt passes, false if all attempts fail.
+//
+// Each retry attempt runs in a separate goroutine with a flakeTestCapture
```
AndrewSisley: suggestion: This also impacts `t.Skip` calls, and probably a few others (Cleanup?), which can cause frustration as tests that say the should be skipped, are not - I suggest noting this in this documentation block in the hope that it saves us time in the future.

A better approach might be to take advantage of us using `flakeTestCapture`, and actually not call `runtime.Goexit()` and instead signal to `runTestWithRetry` that the test flaked out without having to spin up child routines.

AndrewSisley (approved):
LGTM, but please either remove the custom `TB` implementations, or document them and how they differ.

fredcarle (approved):
Approving assuming Andy's comments are resolved. I have pretty much the same concerns.

tests/integration/flake_retry.go
islamaliev: done

tests/integration/flake_retry.go
islamaliev: done

tests/integration/flake_retry.go
islamaliev: when a try fails, the parent test should not fail. That's why we don't call f.TD.FailNow()

AndrewSisley: `f.TD.FailNow()` shouldn't fail the parent because of the child goroutine (if the child routine is kept, if not `runtime.Goexit` wont help here either and the new implementation would need to be documented too)

islamaliev: Since we moved to panic/recover (per your suggestion in the other thread), we can no longer delegate FailNow to f.TB, that would call runtime.Goexit() on the main goroutine and kill the retry loop.                                                                                                                                                                                              
                                                                                                                                                                                                         
We do delegate logging though. It's only the abort mechanism that's ours. Documente in the flakeTestCapture type comment.

PR #4585, opened 2026-02-25
fix: Use testing.Testing() for test mode detection

### LINKED ISSUE #2801: Detect if defradb is being run in test mode
We currently employ a method for detecting test mode execution that may not be optimal or align with conventional practices:
```go
func init() {
	arg := os.Args[0]
	// If the binary is a test binary, use a deterministic nonce.
	if strings.HasSuffix(arg, ".test") || strings.Contains(arg, "/defradb/tests/") {
		generateNonceFunc = generateTestNonce
		generateEncryptionKeyFunc = generateTestEncryptionKey
	}
}
```
The condition `strings.Contains(arg, "/defradb/tests/")` is specifically included to accommodate debug mode execution.

While we have explored using build tags (`//go:build test`) for conditional compilation, which would be the ideal solution, we encountered limitations. The build tag only takes effect when explicitly passed (`-tags=test`). We expected Go to automatically set a test-specific build tag during test runs, but it doesn't seem to be the case. Manually passing this flag is not a viable option as it would impede our ability to execute tests seamlessly from IDEs.

We need to implement a more robust and conventional method for test mode detection that:
- Doesn't rely on string parsing of binary names or file paths
- Works consistently across different execution environments (standard test runs, debug mode, IDE integrations)
- Aligns with Go best practices for test configuration

SharanRP:
## Relevant issue(s)

Resolves #2801 

## Description

Replace unreliable `os.Args[0]` string parsing with Go's official `testing.Testing()` function to detect test context. This eliminates false positives from debug binaries and path-dependent detection.

## Tasks

## How has this been tested?

The changes were verified by adding specific test cases that check the behavior of `generateNonceFunc` and `generateEncryptionKeyFunc`

-  Linux


crypto/nonce.go
```diff
@@ -43,10 +42,7 @@ func generateTestNonce() ([]byte, error) {
 }
 
 func init() {
-	arg := os.Args[0]
-	// If the binary is a test binary, use a deterministic nonce.
```
fredcarle, line 47: todo: Please keep this documentation in.

internal/encryption/encryptor.go
```diff
@@ -181,12 +180,7 @@ func ShouldEncryptIndividualField(ctx context.Context, fieldName immutable.Optio
 }
 
 func init() {
-	arg := os.Args[0]
-	// If the binary is a test binary, use a deterministic nonce.
```
fredcarle, line 185: todo: Please keep this documentation in.

fredcarle (approved):
Thanks for this change Sharan. Just 2 small things to do to before I can merge this. You removed documentation that let the reader know why we have a specific behaviour for testing and I would like to keep it in.

PR #4587, opened 2026-02-25
fix: Fix broken transaction behavior

### LINKED ISSUE #3964: Make integration tests use Transaction wrappers
All the clients we have for integration testing, that is, the C, CLI, HTTP, and JS clients, contain wrappers for client.Txn. However, our integration test harness actually only calls into the Commit and Discard functions of these wrappers. 

For example, the test `TestMutationWithTxnDoesNotAllowUpdateInSecondTransactionUser` contains three calls to `testUtils.Request`, two of which do not have a TransactionID attached, and one of which does.  However, if you call this test, all three calls will be made directly to the main wrapper's `ExecRequest` function. The transaction wrapper's `ExecRequest` function will never be invoked. 

This does not cause any tests to fail, but it is unintended behavior, and in the case of the C client, requires that the transactions be managed explicitly in some way, in order to be passed back into the backend. This, I think, defeats the purpose of having transaction wrappers in the first place.

### LINKED ISSUE #4576: Transactions do not behave correctly in C bindings
Transactions are not respecting isolation in the C bindings. To illustrate this, there is the following C program:

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include "libdefradb.h"
#include "defra_structs.h"

int main() {

    // ---------------------------------------------------
    // Start the node
    // ---------------------------------------------------
	
    NodeInitOptions nodeOpts = {0};
    nodeOpts.inMemory = 1;
    nodeOpts.disableP2P = 1;
    nodeOpts.disableAPI = 1;
    nodeOpts.enableNodeACP = 0;

    NewNodeResult node = NewNode(nodeOpts);
    uintptr_t node_ptr = node.nodePtr;

    if (node.status != 0) {
        printf("Node error: %s\n", node.error);
        return 1;
    } else {
		printf("Node created. \n");
	}

    // ---------------------------------------------------
    // Add schema (without transaction)
    // ---------------------------------------------------

	   const char* userSchema =
        "type Users {\n"
        "  name: String\n"
        "  age: Int\n"
        "}";
		
	Result schemaRes = AddSchema(node.nodePtr, (char *)userSchema, 0);
	if (schemaRes.status != 0) {
		printf("Error adding schema: %s\n", schemaRes.error);
		return 1;
	} else {
		printf("Added schema.\n");
	}
	
	// ---------------------------------------------------
    // Create a transaction
    // ---------------------------------------------------
	
	uintptr_t txnPtr = 0;
	NewTxnResult txnRes = TransactionCreate(node.nodePtr, 0, 0);
	if (txnRes.status != 0) {
		printf("Transaction create error: %s\n", txnRes.error);
		return 1;
	} else {
		txnPtr = txnRes.txnPtr;
		printf("Transaction created with handle: %lu\n", txnPtr);
	}
	
    // ---------------------------------------------------
    // Create document (with transaction)
    // ---------------------------------------------------

	CollectionOptions copts = {0};
	copts.name = "Users";
	
	char* doc = (char *)"[{\"name\": \"Alica\", \"age\": 16}, {\"name\": \"Verso\", \"age\": 26}]";
	Result createRes = CollectionCreate(txnPtr, doc, 0, (char*)"", copts, 0);
	
	if (createRes.status != 0) {
		printf("Error creating document: %s\n", createRes.error);
		return 1;
	} else {
		printf("Called CollectionCreate.\n");
	}
	
    // ---------------------------------------------------
    // Commit the transaction
    // ---------------------------------------------------
	
	// There is no difference in behavior if this line is commented out or not
	// TransactionCommit(txnPtr);

    // ---------------------------------------------------
    // Make the query
    // ---------------------------------------------------

    Result readRes = ExecuteQuery(
        node_ptr,
        "query {\n"
        "  Users {\n"
        "    name\n"
        "    age\n"
        "  }\n"
        "}\n",
        0,
        "",
        ""
    );

    printf("Read Result:\n%s\n%s\n", readRes.value, readRes.error);

    // ---------------------------------------------------
    // Cleanup
    // ---------------------------------------------------

    NodeClose(node_ptr);
    return 0;
}

```

### LINKED ISSUE #4628: GraphQL Parser does not respect transactions
If you look inside `internal/request/graphql/parser.go` and investigate what it's doing, you will see that it actually does not respect transactions. If the user creates a transaction, and then inside that transaction they create a new collection, they will not be able to reference it inside a graphql mutation, because it will fail to be validated.

ChrisBQu:
## Relevant issue(s)

Resolves #4576
Resolves #3964
Resolves #4628

## Description

**Changes to C Bindings and C Bindings Wrapper**
A new helper function was written, `attachTxnFromPointer` which takes in a node/txn pointer and a context. If the pointer was a node handle, nothing will change. If it was a transaction handle, the transaction will be recovered, and attached to the context. In this way, we bring transactional behavior to the C bindings.

The wrapper has also been updated. The Transaction structure now contains a field for client.Txn. In functions throughout `wrapper_tx.go`, the context is modified to attach this value to the context before calling into the main CWrapper function. The functions of the CWrapper will call a helper function `setCtxTxnFromCollection` to see if a transaction is present. They will then make the appropriate call to the exported functions, using either the node handle or the transaction handle. There were places where this was not being done correctly before, but it is now uniformly done. Similarly, the Collection wrapper now contains an immutable.Option[client.Txn]. The function calls in the wrapper use a new helper function `setCtxFromCollection` to modify a context, attaching the Txn that belongs to the collection. This value is set inside the GetCollections function.

**Changes to HTTP Client**

The collection object, like the C client wrapper, is modified to have a transaction attached to it. Other changes are also analogous to the C wrapper.

**Changes to Go Client**

There are architectural changes made here to get things to work. Again, a transaction can be attached to a collection. And across the functions, the typical pattern is that the function examines the context and collection, and keeps note of whether or not there was a transaction. In the case that there wasn't, the Go client tends to create a transaction for the purpose of the operation which is committed at the end of function execution. Now, we will only do that if the transaction was ephemerally created. If one was passed into the function, we will *not* commit or discard it, leaving that to the user to do so.

**Changes to the CLI Wrapper**

The same change is made to the Collection struct, adding a transaction field. And now, we call into the functions by appending the transaction to the command string, using a new helper function, `appendTxnArgs`. The crux of the work is offloaded to the HTTP client.

**Changes to the Test Harness**

Many test actions now contain a new field, "TransactionID" that simply didn't exist before. Beyond that, there is a new utility function, `GetCanonicallyOrderedCollections` which takes a node and a transaction, and returns a list of collections sorted the same way as what would be produced by the `RefreshCollections` function. This is to enable a new pattern used throughout the actions that use collections: if the transaction ID is present, we will get the list of collections _on the transaction_, instead of on the Node. Then, we make the appropriate call to the backend. For actions that do not utilize collection, the change is smaller. We check if the transaction is present, and if so, we call the function on it, otherwise we call the function on the node.

**New Tests**

See that there is an entire directory of new tests for transactional behavior located at `tests/integration/txn`

**Transaction Fix in Parser**

This PR also addresses a bug that existed in the parser. See `internal/request/graphql/parser.go` for the changes. But what was happening, was that the GQL was always being validated against a schemaManager stored in the parser object. This was only being updated when transactions committed, but this meant that GQL mutations inside of the same transaction that created a new collection could not see all of the schemas they should have been able to see, and could not be validated. This addresses that by attaching a map of schemaManagers to the parser, and mapping transaction ID to a schemaManager object. If a transaction is present, the GQL will be validated against the appropriate manager. 

Note: This was an issue that came up while investigating tests. It was necessary to attach this fix to this PR so that the tests would continue passing, with the changes made to the test harness.

## Tasks

## How has this been tested?

- WSL


cbindings/wrapper.go
```diff
@@ -875,6 +898,11 @@ func (w *CWrapper) GetCollections(
 	ctx context.Context,
 	opts ...options.Enumerable[options.GetCollectionsOptions],
 ) ([]client.Collection, error) {
+	txn, hadTxn := datastore.CtxTryGetClientTxn(ctx)
+	if hadTxn {
+		ctx = datastore.CtxSetFromClientTxn(ctx, txn)
```
nasdf, line 903: question: Why does the transaction need to be added to the context if it already exists?

client/json.go
```diff
@@ -333,7 +333,7 @@ func (arr jsonArray) MarshalJSON() ([]byte, error) {
 }
 
 func (arr jsonArray) Unwrap() any {
-	result := make([]any, len(arr.jsonBase.val))
```
nasdf, line 336: question: Why was this changed?

internal/core/block/store.go
```diff
@@ -155,7 +155,7 @@ func determineBlockEncryption(
 			return nil, cidlink.Link{}, err
 		}
 		if prevBlock.Encryption != nil {
-			prevBlockEncBytes, err := blockstore.NewIPLDStore(txn.Encstore()).Get(ctx, prevBlock.Encryption.Cid.KeyString())
```
nasdf, line 158: question: Why was this changed?

tests/integration/utils.go
```diff
@@ -1200,35 +1235,56 @@ func updateDoc(
...
 			func() error {
-				return mutation(
+				err := mutation(
 					s,
-					action,
+					a,
 					node,
 					nodeID,
 					collection,
+					txnOption,
```
nasdf: todo: This function appears to be using a transaction created from a different node.

tests/integration/lens.go
```diff
@@ -76,10 +75,25 @@ func configureMigration(
...
 
-		txn := getTransaction(s, node.Client, action.TransactionID, action.ExpectedError)
+		// Check if a transaction is attached to this action. If so, we will be using it.
+		var txn client.Txn
+		hadTxn := false
+		if action.TransactionID.HasValue() {
+			hadTxn = true
+			var err error
+			txn, err = s.GetTransaction(s.Nodes[action.NodeID.Value()], action.TransactionID)
+			if err != nil {
```
nasdf: todo: This should call `require.NoError(s.T, err)` instead of ignoring.

tests/action/refresh_view.go
```diff
@@ -41,13 +42,28 @@ type RefreshViews struct {
...
 // Execute executes the refresh views action.
 func (a *RefreshViews) Execute() {
+	// Check if a transaction is attached to this action. If so, we will be using it.
+	var txn client.Txn
+	hadTxn := false
+	if a.TransactionID.HasValue() {
+		hadTxn = true
+		var err error
+		txn, err = a.s.GetTransaction(a.s.Nodes[a.NodeID.Value()], a.TransactionID)
+		if err != nil {
```
nasdf: todo: Handle this error or update `GetTransaction` to handle it internally with `require.NoError` calls.

tests/action/patch_collection.go
```diff
@@ -45,13 +46,28 @@ type PatchCollection struct {
...
 // Execute executes the patch collection action.
 func (a *PatchCollection) Execute() {
+	// Check if a transaction is attached to this action. If so, we will be using it.
+	var txn client.Txn
+	hadTxn := false
+	if a.TransactionID.HasValue() {
+		hadTxn = true
+		var err error
+		txn, err = a.s.GetTransaction(a.s.Nodes[a.NodeID.Value()], a.TransactionID)
+		if err != nil {
```
nasdf: todo: Handle this error

nasdf (changes requested):
I left a few notes on specific changes. Can you also make sure that none of the new errors in the test framework are discarded? Every error should be handled by `require.NoError` or other test utilities.

cbindings/wrapper.go
ChrisBQu: Good catch. I think this was a change that got partially made, and there was some residue left over.

internal/core/block/store.go
ChrisBQu: This one, and the above one, I believe, were the result of me letting the automatic linter tool fix formatting. And it overstepped in what it was doing. I thought I went back and caught all of them, but a couple slipped through. I'm just going to revert these.

client/json.go
ChrisBQu: Actually, despite my comment below, I don't see any change here. I compare what I have locally to what is in the main branch, and I don't see any difference(?)

internal/db/collection_index.go
```diff
@@ -179,13 +181,19 @@ func (c *collection) NewIndex(
 	if err != nil {
 		return client.IndexDescription{}, err
 	}
-	defer txn.Discard()
+
+	if !hadTxn {
```
nasdf: todo: I don't think these `hadTxn` checks are needed. If you look at the `ensureContextTxn` function it returns a transaction struct that handles explicit and implicit transactions.

tests/action/utils.go
```diff
@@ -57,9 +63,11 @@ func refreshCollections(
 		if identOption.HasValue() {
 			opts.SetIdentity(identOption.Value())
 		}
-		allCollections, err := node.GetCollections(s.Ctx, opts)
+		allCollections, err := txn.GetCollections(ctx, opts)
 		require.Nil(s.T, err)
 
+		txn.Discard()
```
nasdf: nitpick: I prefer `defer txn.Discard()` near the `NewTxn` call

nasdf (changes requested):

internal/db/collection_index.go
ChrisBQu: After discussing this, and then confirming it, I have made this change across the files. Good call.

internal/db/document_update.go
```diff
@@ -50,12 +52,13 @@ func (c *collection) UpdateDocumentsWithFilter(
...
 		return nil, err
 	}
-	defer txn.Discard()
 
 	res, err := c.updateWithFilter(ctx, filter, updater)
 	if err != nil {
 		return nil, err
 	}
+
+	defer txn.Discard()
```
nasdf: question: why was this moved? It seems like it won't be properly discarded if the `updateWithFilter` call fails.

internal/db/store.go
```diff
@@ -224,14 +236,19 @@ func (db *DB) PatchCollection(
 	if err != nil {
 		return err
 	}
-	defer txn.Discard()
+	if !hadTxn {
```
nasdf: todo: This check should not be needed

internal/db/store.go
```diff
@@ -224,14 +236,19 @@ func (db *DB) PatchCollection(
...
+		defer txn.Discard()
+	}
 
 	err = db.patchCollection(ctx, patchString, migration)
 	if err != nil {
 		return err
 	}
 
-	return txn.Commit()
+	if !hadTxn {
```
nasdf: todo: same as above.

nasdf (changes requested):
Just a few small things to fix then I think this PR should be ready.

internal/db/document_update.go
ChrisBQu: Not sure how that one got moved. I might have been tinkering with something and just made a mistake. Good catch. Moving it back.

tests/integration/txn/update_doc_test.go
```diff
@@ -0,0 +1,353 @@
+// Copyright 2025 Democratized Data Foundation
```
nasdf: todo: These should be 2026

nasdf (approved):
I'm happy with the changes now. One minor note you should update the header year.

PR #4592, opened 2026-02-27
fix: Return error from CLI when collection not found

### LINKED ISSUE #4590: Return error from CLI.collection commands when collection not found

AndrewSisley:
## Relevant issue(s)

Resolves #4590

## Description

Returns an error from CLI when a collection is not found, instead of dumping the help text with no explanation.  Also brings it inline with the C and Go clients.  Http returns a 404, which I think is appropriate, but I'm not too confident about it.

Also unifies the error messages returned from the C and Go clients when this happens.

### DIFF
--- a/cbindings/collection.go
+++ b/cbindings/collection.go
@@ -68,7 +68,7 @@ func getCollection(
 
 	// Only one collection should match the criteria
 	if len(cols) == 0 {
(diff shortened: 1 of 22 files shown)


ChrisBQu (approved):
LGTM. Appreciate you checking in about potential conflicts with the work I've been doing on this one, by the way.

PR #4603, opened 2026-03-02
test: Extend CRDT tests

islamaliev:
## Relevant issue(s)

Resolves #3887

## Description

Extends test coverage for complex Merkle CRDT merge scenarios.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

Specify the platform(s) on which this was tested:
- MacOS

Issue #4508, opened 2026-02-12
EncryptedIndexes patch collection tests do not test what they try to


AndrewSisley:
They all seem to forget to specify which collection they are trying to patch, for example:
```
func TestPatchCollection_RemoveEncryptedIndex_ShouldError(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddSchema{
				Schema: `
					type Users {
						name: String
						email: String @encryptedIndex
					}
				`,
			},
			&action.PatchCollection{
				Patch: `
					[
						{
							"op": "remove",
							"path": "/EncryptedIndexes"
						}
					]
				`,
				ExpectedError: "unable to remove nonexistent key",
			},
		},
	}

	testUtils.ExecuteTestCase(t, test)
}
```

PR #4609, opened 2026-03-04
test: Track multiple expected DAG heads per doc

### LINKED ISSUE #4608: Test framework: WaitForSync returns too early with 3+ node concurrent updates to same document
`ExpectedDAGHeads` in the P2P test state uses `map[string]cid.Cid`, which stores only one expected CID per document key. When 3 or more nodes concurrently update the same document (without `WaitForSync` between updates), each call to `updateConnectedNodes` overwrites the previously recorded expected CID for that document key. As a result, `waitForMergeEvents` only waits for the last-written CID to be merged rather than all of them. The test proceeds to query before all merges are complete, producing inconsistent results.

**Example:** 3 nodes in a full mesh with document subscriptions, each does an update producing CID_A, CID_B, CID_C respectively. The correct expected state would be:
- Node 0 expects: CID_B and CID_C
- Node 1 expects: CID_A and CID_C
- Node 2 expects: CID_A and CID_B

But due to the overwrite, only one CID survives per node. `WaitForSync` returns as soon as that single CID arrives. If the other merge hasn't completed yet, the query sees partial results.

This doesn't affect existing tests because all current multi-node tests either use only 2 nodes (where each node gets exactly one expected CID, so there's no overwrite) or place `WaitForSync` between updates.

islamaliev:
## Relevant issue(s)

Resolves #4608

## Description

Fixes the test framework's `WaitForSync` to correctly wait for all pending merges when multiple nodes concurrently update the same document.

`ExpectedDAGHeads` was `map[string]cid.Cid` which could only store one expected CID per document key. With 3+ nodes doing concurrent updates, each call to `updateConnectedNodes` overwrote the previous expected CID, causing `waitForMergeEvents` to return before all merges completed. Changed the type to `map[string][]cid.Cid` so all expected CIDs are tracked and waited on.

This PR is prerequisite for #4603

## How has this been tested?

Integration tests

- MacOS


AndrewSisley (changes requested):
todo: I think you have linked this to the wrong issue - this change appears to be a prerequisite for #4508

todo: Same comment as on https://github.com/sourcenetwork/defradb/pull/4611 - this looks like it has increased flakiness for the CI jobs - please have a look if you haven't already, and make sure that you are sure that this PR improves the flaky situation instead of making it worse.

tests/integration/events.go
```diff
@@ -212,23 +214,35 @@ func waitForMergeEvents(s *state.State, action WaitForSync) {
...
+		// have appended a CID, and we must wait for all of them.
+		pending := make(map[string]map[cid.Cid]struct{})
+		for key, cids := range node.P2P.ExpectedDAGHeads {
+			cidSet := make(map[cid.Cid]struct{}, len(cids))
+			for _, c := range cids {
+				cidSet[c] = struct{}{}
+			}
+			// remove any CIDs that are already merged
+			if actual, ok := node.P2P.ActualDAGHeads[key]; ok {
+				delete(cidSet, actual.CID)
```
nasdf: thought: I believe you need to walk the entire DAG history and remove all previous version CIDs to make this work correctly. 

There is no guarantee that all of the CIDs will make it to each node and they may arrive out of order.

Consider the update: `A` -> `B` -> `C` of a document.

All three CIDs will be added to the wait set, but if `C` arrives to the node before `A` and `B` there will be no update event triggered for `A` or `B`.

islamaliev: Good catch, this is exactly the problem I ran into. The updated implementation now only waits for the latest CID per document key (`cids[len(cids)-1]`). Since merging C processes its full DAG including A and B, we don't need individual events for ancestors.

Walking the DAG history would also work but needs blockstore access from the test framework. Waiting for the latest CID is simpler and correct since DAG subsumption guarantees all ancestors are included.

islamaliev:
@AndrewSisley Thanks for the review.

Re issue link: you're right, this is a prerequisite for #4508. I'll fix the link.

Re CI flakiness: I've investigated and the flakiness was caused by the initial implementation that waited for ALL accumulated CIDs individually. That approach fails when DAG subsumption occurs (later CID's merge processes all ancestors but fires only one event). I've pushed a fix that waits only for the latest CID per document key, which handles both cases correctly. Ran the flaky test 10 times locally with no failures.

tests/integration/events.go
```diff
@@ -213,23 +215,39 @@ func waitForMergeEvents(s *state.State, action WaitForSync) {
...
-
-		// remove any heads that are already merged
-		// up to the expected head
-		for key, val := range node.P2P.ActualDAGHeads {
-			if head, ok := expect[key]; ok && head.String() == val.CID.String() {
-				delete(expect, key)
+		// Build the set of doc keys with pending merges.
+		// We only need to wait for the latest expected CID per key,
+		// because a later CID's DAG always includes all earlier CIDs.
+		// When the latest CID is merged, all its ancestors are merged too.
```
AndrewSisley: todo: This looks incorrect for stuff with multiple heads.  I do not know if we have any tests that have multiple heads, and use the `WaitForSync` action (that calls this function) at the moment, but we probably should in the near future.

I assume we don't yet given your testing, so I'm fine with the logic remaining as is for now, but please create a ticket, and link to it from the `WaitForSync` action's documentation, and from the documentation for `waitForMergeEvents` - unless you are certain (please explain why) that I am incorrect here.

AndrewSisley (approved):
Looks good to me, thanks Islam!

I've added a todo documentation request, please resolve that before merge.

tests/integration/events.go
islamaliev: You were spot on — the "latest CID only" approach breaks when different nodes update the same doc concurrently. Fixed it.

Each expected CID is now tagged with its source node ID. At build time we keep only the latest CID per (key, source) pair — same-source CIDs form a linear chain where the latest subsumes ancestors, but different-source CIDs are concurrent branches that each need their own merge event.

The concurrent update tests (LWW different fields, pcounter, pncounter simultaneous updates) live on a related branch and exercise this multi-head scenario with WaitForSync.

AndrewSisley: > Each expected CID is now tagged with its source node ID. At build time we keep only the latest CID per (key, source) pair — same-source CIDs form a linear chain where the latest subsumes ancestors, but different-source CIDs are concurrent branches that each need their own merge event.

I know you've merged now, but I think this approach can still fall down for concurrent updates to the same document on the same node.

I think this is broken in production at the moment at the datastore level, but not at the blockstore, so in the future when we fix the datastore, we will likely have tests tripping up on this.

PR #4617, opened 2026-03-06
feat: Cursor based query pagination

jsimnz:
## Description

Implements a cursor based pagination query. Cursor pagination is much more effecient than limit/offset pagination, as its lends itself to indexes much more naturally.

This implementation exposes a new `_cursor` query type, which replicates a given developer defined type query, but replaces the `limit` and `offset` syntax with `first`/`after` and `last`/`before`, which take in numeric limit and cursor tokens respectively.

An intentional limitation of this cursor pagination is that all `order` fields *must* have an appropriately defined index on that field (or a member of a composite index). It will return an error if it can't find a matching index. This is an intentional design consideration for performance reasons.

## Tasks

## How has this been tested?

Additions to the integration test suite

- Ubuntu


client/request/after.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2024 Democratized Data Foundation
```
fredcarle: todo: 2026 for all the new files at least :)

jsimnz: Oops, wrong copy paste

internal/planner/planner.go
```diff
@@ -258,6 +265,49 @@ type aggregateNode interface {
 	SetPlan(plan planNode)
 }
 
+func (p *Planner) expandCusrorPlan(plan *selectTopNode) error {
```
islamaliev, line 268: typo: `expandCursorPlan`

client/request/cursor.go
```diff
@@ -0,0 +1,63 @@
...
+type CursorSelect struct {
+	Field
+
+	// Select is the inner collection query (e.g., User { name }).
+	Select *Select
+
+	Firstable
+	Afterable
+	Lastable
+	Beforeable
```
islamaliev, line 31: praise: I like the names :D

client/request/last.go
```diff
@@ -0,0 +1,18 @@
...
+// licenses/APL.txt.
+
+package request
+
+import "github.com/sourcenetwork/immutable"
+
+// Lastable is an embeddable struct for backward cursor-based pagination.
+type Lastable struct {
+	Last immutable.Option[uint64]
+}
```
islamaliev, line 18: suggestion: maybe it makes sense to put them all in one file as they are highly cohesive

internal/db/fetcher/indexer.go
```diff
@@ -90,6 +92,32 @@ func newIndexFetcher(
 		return nil, err
 	}
 
+	if startKey.HasValue() {
+		if matchIter, ok := iter.(*indexMatchIterator); ok {
```
islamaliev, line 96: todo: I guess this covers most of the cases but not all. This excludes all `_in`, `_or` (`multiIndexIterator`) ops as well as arrays and JSONs (`memorizingIndexIterator`). 
They all use underneath either `indexMatchIterator` or `eqSingleIndexIterator`. I think we should cover all the cases. And make sure they are tested.

internal/db/fetcher/wrapper.go
```diff
@@ -90,13 +90,14 @@ func (f *wrappingFetcher) Start(ctx context.Context, prefixes ...keys.Walkable)
...
 	dsPrefixes := make([]keys.DataStoreKey, 0, len(prefixes))
+	var indexStartKey immutable.Option[keys.Walkable]
 	for _, prefix := range prefixes {
-		dsPrefix, ok := prefix.(keys.DataStoreKey)
-		if !ok {
-			continue
+		switch p := prefix.(type) {
+		case keys.DataStoreKey:
+			dsPrefixes = append(dsPrefixes, p)
+		case *keys.IndexDataStoreKey:
```
islamaliev, line 98: note: hm... I wonder why it's a pointer

internal/planner/planner.go
```diff
@@ -751,7 +801,32 @@ func (p *Planner) MakeSelectionPlan(selection *request.Select) (planNode, error)
 //
 // Note: Caller is responsible to call the `Close()` method to free the allocated
 // resources of the returned plan.
-//
+
+// validateCursorIndex checks that a cursor query has a compatible index for ordering.
+// Returns (reversed, error) where reversed indicates if iteration should be in reverse direction.
+func (p *Planner) validateCursorIndex(plan *selectTopNode) (bool, error) {
+	scan := getNode[*scanNode](plan.selectNode)
```
islamaliev, line 808: todo: you injected this function between comments for `MakePlan`

internal/planner/planner.go
```diff
@@ -751,7 +801,32 @@ func (p *Planner) MakeSelectionPlan(selection *request.Select) (planNode, error)
...
+	if scan == nil {
+		return false, ErrNoSupportingIndexForCursor
+	}
+
+	if len(scan.ordering) == 0 {
+		return false, nil
+	}
```
islamaliev, line 815: suggestion: this is redundant as `CanBeOrderedByIndex` checks it already

internal/planner/select.go
```diff
@@ -408,6 +415,62 @@ func findIndexByOrderingField(scanNode *scanNode) immutable.Option[client.IndexD
 	return immutable.None[client.IndexDescription]()
 }
 
+// validateCursorOrderIndex ensures that ordering fields have a supporting index.
+// Returns an error if cursor pagination is used with non-indexed ordering fields.
+func validateCursorOrderIndex(scanNode *scanNode) error {
```
islamaliev, line 420: suggestion: I think this func and `validateCursorIndex` are doing very similar job (not 100% sure). Can't we merge them?

internal/cursor/cursor.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"encoding/base64"
+	"encoding/json"
+)
+
+// CursorPayload holds the data encoded into a cursor token.
+type CursorPayload struct {
+	DocID     string         `json:"d"`
+	Keys      map[string]any `json:"k,omitempty"`
```
islamaliev, line 21: question: can't we turn this into `map[string]client.NormalValue`?

internal/cursor/cursor.go
```diff
@@ -0,0 +1,55 @@
...
+	Direction string         `json:"o"`
```
islamaliev, line 22: question: I didn't find in the code (most likely I missed) usage/validation of `Direction` fields. Would you tell me where and how it's used?

tests/integration/results.go
```diff
@@ -235,6 +235,210 @@ func (matcher *docIDAt) String() string {
...
+		for _, field := range m.expectedKeyFields {
+			if _, exists := payload.Keys[field]; !exists {
+				m.lastError = fmt.Sprintf("missing expected key field %q", field)
+				return false, nil
+			}
+		}
+	}
+
+	// Validate expected key-value pairs if specified
+	if len(m.expectedKeys) > 0 {
```
islamaliev, line 325: nitpick: `len` checks are redundant here

tests/integration/query/cursor/utils.go
```diff
@@ -0,0 +1,72 @@
...
+
+// userCollectionGQLSchema defines a User type with indexed age field for cursor pagination tests.
+var userCollectionGQLSchema = `
+	type User {
+		name: String
+		age: Int @index
+	}
+`
+
+func executeTestCase(t *testing.T, test testUtils.TestCase) {
```
islamaliev, line 29: todo: I don't like this anti-pattern that we use in some places in our code where we have this "util" method doing some additional stuff.

It bit me several times already when I implemented secondary-index multiplier because these methods selectively copy some values from the given `test`.

If you want to append something to the test, do 
```go
test.Actions = append(
    &action.AddSchema{
		Schema: userCollectionGQLSchema,
	}, test.Actions...)
```

Unless you have a good reason for doing so which would require a comment.

tests/integration/query/cursor/with_first_test.go
```diff
@@ -0,0 +1,185 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	executeTestCase(t, test)
+}
+
+func TestCursorWithFirst_ReturnsPageInfoWithHasNextTrue(t *testing.T) {
```
islamaliev, line 104: suggestion: This one is redundant. `TestCursorWithFirst_ReturnsExactlyNResults` already does it

tests/integration/query/cursor/with_first_test.go
```diff
@@ -0,0 +1,185 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	executeTestCase(t, test)
+}
+
+func TestCursorWithFirst_ReturnsPageInfoWithHasNextFalse(t *testing.T) {
```
islamaliev, line 147: suggestion: also redundant

tests/integration/query/cursor/with_float_field_test.go
```diff
@@ -0,0 +1,267 @@
...
+							endCursor
+						}
+					}
+				}`,
+				Asserter: testUtils.ResultAsserterFunc(func(t testing.TB, result map[string]any) (bool, string) {
+					cursor := result["_cursor"].(map[string]any)
+					users := extractUsers(cursor["User"])
+					allDocs = append(allDocs, users...)
+					return true, ""
+				}),
```
islamaliev, line 112: note: this is technically a misuse of `Asserter`

tests/state/state.go
```diff
@@ -45,6 +45,10 @@ type TestState interface {
 	GetIdentity(Identity) acpIdentity.Identity
 	// GetDocID returns the document ID for the given collection index and document index.
 	GetDocID(collectionIndex, docIndex int) client.DocID
+	// SetCapturedVariable stores a named value for later retrieval.
+	SetCapturedVariable(name string, value any)
```
islamaliev, line 49: todo: I think you missed the part in the comment of `TestState` that says "read-only interface". I was meant for matchers to be used. If we allow matchers to mutate the state this might lead to unpleasant behaviour in the future if we are not careful with what we change.

I think we either have to reconsider the purpose of this interface or think about more appropriate solution long-term. Maybe we need another interface `MutableTestState` that is passed around to matchers (or special matchers).

Maybe a clearer approach for now would be to use an approach similar to `UniqueValue` matcher where you instantiate a matcher before actions' descriptions and use/reference it however you want in those actions.

tests/integration/query/cursor/with_composite_index_test.go
```diff
@@ -0,0 +1,317 @@
...
+							endCursor
+						}
+					}
+				}`,
+				Asserter: testUtils.ResultAsserterFunc(func(t testing.TB, result map[string]any) (bool, string) {
+					cursor := result["_cursor"].(map[string]any)
+					users := extractUsers(cursor["User"])
+					allDocs = append(allDocs, users...)
+					return true, ""
+				}),
```
islamaliev, line 117: suggestion: you do this action quite often. I think it makes sense to create a dedicated action/asserter (whatever it is) for it.

islamaliev (changes requested):
Gread job overall. I left some suggestions/concerns that I'd like to clarify before merging.

Am I understanding this correctly that in order for cursor to work we need an index on ordered-by field? I understans the requirement when we want to have a specific order, but if we just request docs in ther natural order why can't we use cursor? We can also specify prefix for such queries after all.

Another point: seems like gql supports parameters like `filter` and `groupBy` but I didn't see any test covering/demonstrating it.
Maybe we could write a test multiplier for this similar to secondary-index that would just run all queries wrapped in cursor.

client/request/last.go
jsimnz: good point

internal/planner/planner.go
jsimnz: :clown_face:

internal/db/fetcher/wrapper.go
jsimnz: I'll poke, there was some type error at one point and this resolved it, ill investigate further,

tests/integration/query/cursor/with_composite_index_test.go
jsimnz: I dont want to hide it away in an opaque utils func more than I have to. Its only a few lines and easy to read/reason about/change in the future.

tests/integration/query/cursor/with_float_field_test.go
jsimnz: I consider this a nitpick. This isnt violating any semantic, contract, or side effect. Feel free to push back

tests/integration/query/cursor/with_first_test.go
jsimnz: agreed

tests/integration/query/cursor/with_first_test.go
jsimnz: agreed

tests/state/state.go
jsimnz: refactored entirely. Move to the matcher design as suggested

tests/integration/query/cursor/utils.go
jsimnz: agreed

internal/planner/select.go
jsimnz: simplified them

internal/cursor/cursor.go
jsimnz: This is a fairly straight forward and strongly (enough) typed, dont feel its necessary

internal/cursor/cursor.go
jsimnz: Was outdated, removed

PR #4629, opened 2026-03-12
feat: P2P Resource manager with predefined profiles via CLI option

### LINKED ISSUE #421: Leverage libp2p's Network Resource Manager for net-node
The feature was introduced in v0.18 of libp2p. Documentation: https://github.com/libp2p/go-libp2p-resource-manager/blob/master/README.md. Why? Be aware of & handle resource accounting and limits.

alexmylonas:
## Relevant issue(s)

Resolves #421
Depends on https://github.com/sourcenetwork/go-p2p/pull/22

## Description
This PR integrates the libp2p resource manager into DefraDB's P2P networking stack, by providing two distinct profile (`limited` & `server`) options via `--resource-profile` CLI flag. 

### `limited` profile
Targeting hardware constraint nodes by pinning limits to half of the baseline, for system, transient and peer connections and streams.

### `server` profile
Adds reasonable limits only to `Peer` connections to ensure fairness and avoid peer monopolization of resources 

### no profile
Fallbacks to existing autoscale resource manager as provided by libp2p.

## Tasks

## How has this been tested?

Added unit tests for profile setting and showcasing limit hits for connections and peers

- MacOS


node/resource_manager.go
```diff
@@ -0,0 +1,96 @@
...
+	ResourceProfileLimited = "limited"
+	// ResourceProfileServer only applies limits per-peer, rest is autoscaled
+	ResourceProfileServer = "server"
+)
+
+// resourceProfiles maps profile names to their PartialLimitConfig overrides.
+var resourceProfiles = map[string]rcmgr.PartialLimitConfig{
+	ResourceProfileLimited: {
+		System: rcmgr.ResourceLimits{
+			ConnsInbound:    32,
```
alexmylonas: The limits from `limited` are practically half of the baseline that the underlying p2p defaults to.

While i used my best judgement, I leave to maintainers to propose limits that might make more sense for use cases.

We can also entertain the idea of even more stricter profile for edges nodes that might connect to 1-2 peers.

node/resource_manager.go
```diff
@@ -0,0 +1,96 @@
...
+			ConnsOutbound:   4,
+			Conns:           4,
+			StreamsInbound:  64,
+			StreamsOutbound: 128,
+			Streams:         128,
+			Memory:          64 << 20,
+		},
+	},
+	ResourceProfileServer: {
+		PeerDefault: rcmgr.ResourceLimits{
```
alexmylonas: Imo, `server` profile settings should be uncapped to leverage the autoscale depending on the hardware but pinning reasonable (and generous) limits at the peer level to provide peer fairness and protect against monopolization.

go.mod
```diff
@@ -466,3 +466,5 @@ require (
 )
 
 replace github.com/wlynxg/anet => github.com/sourcenetwork/anet v0.0.0-20250417190629-7c87cba7799e
+
+replace github.com/sourcenetwork/go-p2p => github.com/alexmylonas/go-p2p v0.0.0-20260310235021-0e7341cfd30b
```
alexmylonas: TODO: remove after https://github.com/sourcenetwork/go-p2p/pull/22 has been merged and tagged.

node/resource_manager.go
```diff
@@ -0,0 +1,96 @@
...
+			Streams:         1024,
+			Memory:          128 << 20,
+		},
+	},
+}
+
+// buildResourceManager constructs a resource manager from the given profile name.
+// If profile is empty, nil is returned and go-p2p will use libp2p's autoscaled defaults.
+func buildResourceManager(profile string) (network.ResourceManager, error) {
+	if profile == "" {
```
reecepbcups: This is already guarded for via `if ResourceProfile != ""` right?

cli/config/config.go
```diff
@@ -62,6 +62,7 @@ var ConfigFlags = map[string]string{
 	"peers":                      "net.peers",
 	"p2paddr":                    "net.p2paddresses",
 	"no-p2p":                     "net.p2pdisabled",
+	"resource-profile":           "net.resourceprofile",
```
reecepbcups, line 65: UX nit: I do not see any input validation at the CLI level (i.e. --resource-profile=typo would pass al the way to buildResourceManager at startup before erroring.)

node/resource_manager.go
reecepbcups: With inbound=32 and outbound=64, setting Conns=64 means if all 64 slots are taken by outbound, zero inbound can connect right? This would mean Conns needs to be 96 (same pattern for streams)

alexmylonas: Yes, this is accurate, the intent behind it is outbound prioritisation over inbound, the libp2p follows the same pattern for their [defaults](https://github.com/libp2p/go-libp2p/blob/67f4f585fbdb16dd801d6c6c9010b8fa2917cfc0/p2p/host/resource-manager/limit_defaults.go#L663).

**For connections:**
If a node is so busy that requires 64 outbound connections, rejecting new inbound conns could be a reasonable back pressure, this depends of course on the use case. 

In addition, the `connectionManager` should be the primarily enforcement mechanism for connections, as is responsible for trimming connections when they reach above the high watermark, however I mentioned [here](https://github.com/sourcenetwork/go-p2p/pull/22#discussion_r2915234158) that if the limits are higher than connectionManager's lo/hi, it wont be in effect. 

It would also interesting to allow customization of those lo/hi watermarks as well but this could be a followup.

**For streams:** 
While the logic is similar for streams the limits seem generous with an average 16streams/connection. 

**Memory**
Finally, an important consideration is how much memory each connection consumes with limit of 128MB: 
The averages: 
- `Conns: 64`:  ~2MB per connection
- `Conns: 96`:  ~1.34MB per connection
Note: I ran some over the counter memory profiling with 6 GossipSub topics averaging around  ~300KB per connection (only accounting Go's heap though) so 96 connection can be well below the limit.

In summary, I am ok with with increasing the total connections so that inbound+outbound = connections (and the streams more for consistency though). I am updating the rest of field to match this pattern.

nitpick: there is a convention around network configuration numbers to be powers of 2 but it doesn't seem to have have any practical impact.

Let me know if you think differently.

node/resource_manager.go
alexmylonas: Yes, and I think I agree that this function shouldn't enforce business logic.

You either pass a valid predefined profile or you get an error.

cli/config/config.go
alexmylonas: Yes, that's correct.

In general, I can see there is limited content validation for input namely, `peers` or `document-acp-type` which is closer to the input format, both fail downstream. 

I would be happy to add it if there is strong preference.

go.mod
reecepbcups: v0.10.0 is now tagged

fredcarle: Sorry for the mistake on the tag. v0.1.10 is now ready :)

cli/start.go
```diff
@@ -284,6 +285,11 @@ func MakeStartCommand(ctx context.Context) *cobra.Command {
 		cfg.GetBool(config.ConfigFlags["no-p2p"]),
 		"Disable the peer-to-peer network synchronization system",
 	)
+	cmd.PersistentFlags().String(
+		"resource-profile",
+		cfg.GetString(config.ConfigFlags["resource-profile"]),
+		"Set the resource manager profile for the p2p network (supported: limited, server) (default \"\")",
```
fredcarle: suggestion: No need to specify the default here as you've already done that in `config/config.go` and the docs will pick up on that automatically.

cli/config/config.go
fredcarle: I think what you did is fine given the state of the `start` function. Although I don't think you should try to add validation specifically for the `resource-profile` flag, I would be curious to know what approach you would take to validate all the possible inputs to the start function generally.

node/resource_manager.go
```diff
@@ -0,0 +1,92 @@
...
+			Memory:          128 << 20,
+		},
+	},
+}
+
+// buildResourceManager constructs a resource manager from the given profile name.
+func buildResourceManager(profile string) (network.ResourceManager, error) {
+	partial, ok := resourceProfiles[profile]
+	if !ok {
+		return nil, fmt.Errorf("unknown resource profile %q: valid values are %q, %q",
```
fredcarle: nitpik: If you look at the `node/store.go` you'll see that we have a specific way of returning new errors. Errors are defined in the `errors.go` file (we do this in almost all packages) and dynamic values associated with the errors are displayed as kv pairs. This allows the errors to be consistantly recognizable regardless of the dynamic nature of the message that can be provided to users.

node/resource_manager.go
fredcarle, line 1: thougth: Now that I see it I regret my choice of option A2 😅. To be fair the code here is correct and I think the different limits are reasonable.

I didn't consider that `rcmgr` was going to be the resource manager from `libp2p` itself. The reason for `go-p2p` to exist is to try to isolate Defra from the specifics of `libp2p` itself and simply require that the p2p stack used implements the `client/Host` interface. This would allow users to provide their own p2p implementation (something other than `libp2p`) if they wanted to. This means that the code in the file should be moved to `go-p2p`.

I think it's fine for `go-p2p` to define these different resource profiles even though they are technically made for Defra. They are optional though and sensible so it makes sense, in my view, for them to exists over there. Also, given that `go-p2p` is made explicitly with `libp2p`, it's totally ok for the `WithResourceManager` option to continue existing in its current form. We would also have a `WithResourceProfile` option.

alexmylonas: Yes, I agree that importing a specific library of `libp2p` could create a lock-in, I should have pushed harder for A1 myself :)

So what we are thinking is leave the `WithResourceManager` in `go-p2p` if some1 needs something custom they can add it via defradb code, by adding a line and building a `ResourceManager` which would allow customization.

Add `WithResourceProfile` to `go-p2p` and move the code from `resource-manager.go` there and just a pass a profile `string` downstream. 

---

There is also a third alternative but with less direct control over specifics limits. Instead of using a profile we can define a memory limit for p2p like `MaxMemory`, then we can derive specific limits from that i.e. 1 connection per 1 MB. 

Very similarly to how kubo handles it by defining the user inputs [here](https://github.com/ipfs/kubo/blob/9c4e2683e00995533eb374771fd574ec5b4a002f/config/swarm.go#L116) and deriving some limits based on memory and FD [here](https://github.com/ipfs/kubo/blob/9c4e2683e00995533eb374771fd574ec5b4a002f/core/node/libp2p/rcmgr_defaults.go#L37)

So `defradb` will pass options via `WithResourceLimits` and go-p2p will translate to libp2p specifics.

What do you think?

fredcarle: I like that third option. It makes it nicely applicable to any other p2p implementation also.

cli/config/config.go
alexmylonas: Imo, letting variables failing downstream at the package level instead of having a central validation isn't a bad direction and provides separation of concerns. 

However, if we were to pursue that, I would add a `Validate` method to the `NodeOptions` which would run inside `New()` to do validation for all values related to `node`. 

In addition, there are some `cfg` values not related to `node` like `log.level` which right now are not validated, so probably a `Validate(cfg)` for those to be run at the end of `PersistentPreRunE`

node/resource_manager.go
alexmylonas: Sounds good, so the plan changes: 

- We are creating `WithResourceLimits` in `go-p2p`, accepting a `ResourceLimits{MaxMemory, MaxFileDescriptors}` struct
- `go-p2p` is translating/deriving connections limits from input and building the libp2p resource manager
- In `defradb`, we are calling `WithResourceLimits` instead of `WithResourceManager` 

Some open questions related to `defradb` inputs:
- [ ] Do we want to allow users to also specify `MaxMemoryPerPeer` (and FD) ? 
- [ ] Do we still want to use the profiles instead of something like `--p2p-max-memory` ? I think at this point it only make sense if we allow `MaxMemoryPerPeer` (server profile)
- [ ] If we use `--p2p-max-memory` flag are we going accept raw bytes or humanize form like 4GiB

fredcarle: I think we can keep it simple for now and see what feedback we get from users. If it turns out that fine tuning is necessary then we can revisit this at that time.

cli/start.go
alexmylonas: I removed as it doesn't have a significant contextual meaning but it's not picked but it's not picked up automatically cause it's the zero value. If the default value was `limited` it would have been picked up as expected.

node/resource_manager.go
alexmylonas: good callout, updated!

node/resource_manager.go
alexmylonas: Hi @fredcarle, I have pushed a new PR https://github.com/sourcenetwork/go-p2p/pull/25 in `go-p2p` that moves the Resource Manager there.

DefraDB now only translates resource profile to `ResourceLimits` and passes them downstream.

Let me know what you think!

PR #4630, opened 2026-03-13
fix: Handle trailing bytes in dump keys

### LINKED ISSUE #4573: Dump endpoint fails with 'invalid cid: trailing bytes' on blockstore keys
## Summary

The `/debug/dump` endpoint fails with `invalid cid: trailing bytes in data buffer passed to cid Cast` when iterating blockstore keys that contain extra bytes beyond the CID.

## Root Cause

In `internal/datastore/multi.go`, `HumanReadableKey()` calls `cid.Cast()` directly on key slices:

```go
func HumanReadableKey(key []byte) (string, error) {
    switch key[0] {
    case blockStoreKey:
        if bytes.HasPrefix(key[1:], []byte{toMergeIndexPrefix}) {
            cid, err := cid.Cast(key[2:])  // fails if trailing bytes exist
            ...
        }
        cid, err := cid.Cast(key[1:])      // fails if trailing bytes exist
        ...
    case encStoreKey:
        cid, err := cid.Cast(key[1:])      // fails if trailing bytes exist
        ...
    }
}
```

`cid.Cast()` is strict — it rejects any trailing bytes after the CID. When blockstore keys contain chunk index or other metadata bytes appended after the CID, the parse fails.

**Additional bug**: There's a duplicate `case encStoreKey` that shadows `case systemStoreKey` (lines ~147-148).

## Fix

Use `cid.Decode()` or parse the CID prefix length and slice accordingly, rather than passing the full remaining key bytes to `cid.Cast()`.

## Reproduction

```bash
# Create schema + document, then dump
defradb client schema add 'type User { name: String  age: Int }'
defradb client query 'mutation { create_User(input: {name: "Alice", age: 30}) { _docID } }'
curl http://localhost:9181/api/v0/debug/dump
# Returns: "invalid cid: trailing bytes in data buffer passed to cid Cast"
```

Also reproducible via integration tests in `defradb.rs`:
```bash
cargo test -p integration-test --test backup -- go_dump
```

SharanRP:
## Relevant issue(s)

Resolves #4573

## Description

Fixes the `/debug/dump` endpoint failure when chunk storage is enabled.

- Replaced `cid.Cast` with `cid.CidFromBytes` in [HumanReadableKey](cci:1://file:///home/sharanrp/Desktop/projects/defradb/internal/datastore/multi.go:122:0-154:1) to tolerate trailing chunk metadata bytes in raw keys.
- Corrected a copy-paste bug where `case encStoreKey` shadowed `systemStoreKey`.

## Tasks

- [x] I made sure the code is well commented.

## How has this been tested?

Verified with a reproduction script using `cid.CidFromBytes` on keys with trailing metadata. Ran internal datastore tests to ensure no regressions.

Tested on:
- Linux


AndrewSisley:
Hey Sharan, sorry that we are only just getting round to review now.  

The fix looks probable, but it cannot be merged without integration tests.  Please add some before re-opening this if you wish to resolve the ticket.

Thanks for having a look!

PR #4633, opened 2026-03-18
fix: Fallback to scanNode if index can't handle

### LINKED ISSUE #4632: Fix secondary index fallback for unsupported filter conditions
Secondary index doesn't fall back to full scan for several filter conditions it can't handle, returning wrong results instead:

- JSON ordering operators (`_gt`, `_lt`, `_ge`, `_le`) with non-numeric values
- JSON `_like`/`_nlike` on root level (index only stores leaf values)
- Array field `_eq`/`_ne` with literal array values (index stores elements, not whole arrays)

Also, the original scan node leaks its iterator when replaced by a valuesNode during upsert.

Fixing these removes the need for ~30 `SecondaryIndex` test exclusions.

Parent issue: #4353

islamaliev:
## Relevant issue(s)

Resolves #4632

## Description

Fix cases where the secondary index tries to handle filter conditions it can't actually process (JSON ordering on non-numeric values, like/nlike on root-level JSON, array equality with literal arrays). Instead of returning wrong results, it now falls back to a full scan. 

Also fixes a scan node leak in the upsert path — when the original scan node gets replaced by a valuesNode, it wasn't being closed, leaking the iterator. 

Removes the SecondaryIndex test exclusions (~30 tests) that were working around these issues. 

## Tasks

## How has this been tested?

Integration tests

- MacOS


AndrewSisley (approved):
LGTM, and thanks for the documentation, it explained things really well IMO

PR #4642, opened 2026-03-19
fix: Bump api version to v1

### LINKED ISSUE #4640: Bump API to v1
The HTTP API is currently labeled v0 but should be updated to v1

### LINKED ISSUE #4641: Add API latest version redirect
We currently support an API versioning scheme `/api/v1/` to denote major versions. We should also support a base `/api/` route that redirects to the latest version to simplify docs examples.

nasdf:
## Relevant issue(s)

Resolves #4640
Resolves #4641

## Description

This PR bumps the API version to v1. It also adds a default version redirect so that `/api/` can be used in place of `/api/v1` for the latest version.

## Tasks

## How has this been tested?

Existing tests.

- MacOS


http/http_client.go
```diff
@@ -40,7 +40,7 @@ func newHttpClient(rawURL string) (*httpClient, error) {
 	return &httpClient{
 		client:  http.DefaultClient,
 		baseURL: baseURL,
-		apiURL:  baseURL.JoinPath("/api/v0"),
+		apiURL:  baseURL.JoinPath("/api"),
```
fredcarle: thought: I'm not sure if the client api version should be floating. This seems like an ok approach only if we agree that no pre v1 nodes will ever be used.

http/handler.go
```diff
@@ -25,8 +25,12 @@ import (
 // This is checked by the http/handler_extras.go/Purge function to determine which response to send
 var IsDevMode bool = false
 
-// Version is the identifier for the current API version.
-var Version string = "v0"
+var (
```
fredcarle: todo: Please make these constants.

http/handler.go
```diff
@@ -87,12 +91,14 @@ func NewHandler(db DB) (*Handler, error) {
...
 	mux := chi.NewMux()
-	mux.Route("/api/"+Version, func(r chi.Router) {
+	mux.Route("/api", func(r chi.Router) {
 		r.Use(
 			ApiMiddleware(db, txs),
 			TransactionMiddleware,
 			AuthMiddleware,
 		)
+		r.Mount("/"+VersionV0, router)
+		r.Mount("/"+VersionV1, router)
```
fredcarle: thougth: There is a risk here of changing router and it being applied to both v0 and v1 when users expect to deal with explicitly v0 behaviour. It might be better to decouple them.

docs/website/getting-started.md
```diff
@@ -38,7 +38,7 @@ In this document, we use the default configuration, which has the following beha
 
 - `~/.defradb/` is DefraDB's configuration and data directory
 - `client` command interacts with the locally running node
-- The GraphQL endpoint is provided at <http://localhost:9181/api/v0/graphql>
+- The GraphQL endpoint is provided at <http://localhost:9181/api/graphql>
```
fredcarle: suggestion: The versionned endpoint should still be mentionned so that users know that they can pin to a given version.

docs/website/references/http/openapi.json
```diff
@@ -642,7 +642,7 @@
     },
     "info": {
         "title": "DefraDB API",
-        "version": "0"
+        "version": "v1"
```
fredcarle: question: Is the `v` necessary? To me it seems redundant since the the field is `version`. If we want it to be based on a variable, maybe we can define the version in `openapi.go`.

http/handler.go
nasdf: I don't think we're planning on supporting multiple API versions within a single instance though right? I think it'd be better to remove both v1 and v0 versions when we bump to v2.

docs/website/getting-started.md
nasdf: updated here and in the readme

http/handler.go
nasdf: done

http/http_client.go
nasdf: good catch, I changed it to be locked to v1

docs/website/references/http/openapi.json
nasdf: should be fixed now

http/handler.go
fredcarle: True. All good then.

fredcarle (approved):
LGTM

shahzadlone (approved):
Looks good

PR #4649, opened 2026-03-23
fix: Make change detector skip transactional isolation tests

### LINKED ISSUE #4648: Make Change Detector skip tests exhibiting transactional isolation
Transactions are not and cannot be kept alive between two different DefraDB instances. This means tests that display transactional isolation, such as `TestTxn_DeleteIndex_ExhibitsTransactionalIsolation_Succeeds` cannot work with the change detector. Therefore, the detector should be adjusted to detect and skip these tests.

ChrisBQu:
## Relevant issue(s)

Resolves #4648 

## Description

Some tests do not work with the change detector. Specifically, tests that do things inside transactions without committing them, displaying transactional isolation, cannot work with the change detector. This adjusts the change detector logic to skip these tests. It does so by creating a set of transaction IDs that are created prior to the first action of a test. Upon seeing a CommitTransaction action, it removes the appropriate item from the set. Then, when this initial scan is done, we know that if any IDs are left in the set, it means the test uses uncommitted transactional behavior and must be skipped.

## Tasks

## How has this been tested?

-WSL

### DIFF
--- a/tests/integration/utils.go
+++ b/tests/integration/utils.go
@@ -769,28 +769,25 @@ func getActionRange(t testing.TB, testCase TestCase) (int, int) {
 
 ActionLoop:
 	for i := range testCase.Actions {
-		switch testCase.Actions[i].(type) {
+		switch concreteAction := testCase.Actions[i].(type) {
 		case SetupComplete:
 			setupCompleteIndex = i
 			// We don't care about anything else if this has been explicitly provided
 			break ActionLoop
 
 		case *action.AddCollection:
-			concreteAction := testCase.Actions[i].(*action.AddCollection)
 			if concreteAction.TransactionID.HasValue() {
 				transactionIDset[concreteAction.TransactionID.Value()] = struct{}{}
 			}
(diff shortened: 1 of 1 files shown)


tests/integration/utils.go
```diff
@@ -772,7 +775,34 @@ ActionLoop:
 			// We don't care about anything else if this has been explicitly provided
 			break ActionLoop
 
-		case *action.AddCollection, *action.AddDoc, UpdateDoc, Restart, CommitTransaction:
+		case *action.AddCollection:
+			concreteAction := testCase.Actions[i].(*action.AddCollection)
```
AndrewSisley: todo: It looks like you can remove the casts here by declaring a variable on line 772:
```
switch concreteAction := testCase.Actions[i].(type) {
```

AndrewSisley (approved):
I feel like there should be a cleaner way (not thought much on it though), but this unblocks the build :)

Please have a quick look at the rabbit comment before merge.

Issue #4626, opened 2026-03-12
ConcurrentTransaction is a lie, and will break many things if concurrent calls are made to it


AndrewSisley:
It only protects locks concurrent writes to the underlying store.  It does not prevent concurrent writes that spans multiple store operations, or caches (i.e. most/all public calls).

AndrewSisley:
see also https://github.com/sourcenetwork/defradb/issues/4627 https://github.com/sourcenetwork/defradb/pull/4651#discussion_r3011837339

AndrewSisley:
Some solution overlap with https://github.com/sourcenetwork/defradb/issues/4626

AndrewSisley:
WIP branch: ~~`4626-concurrent-txn`~~ `4626-txn-concurrency-limit`

PR #4651, opened 2026-03-23
refactor: Remove collection cache from context

### LINKED ISSUE #4583: Move ctx cache onto db
Broken out from https://github.com/sourcenetwork/defradb/issues/4268 - we need to manage the transaction-cache anyway due to the bug noted in that ticket.  Removing the cache from context makes it a little more intuitive, and increases it's scope to cover the whole transaction, across multiple contexts/calls.

AndrewSisley:
## Relevant issue(s)

Resolves #4583

## Description

Removes the collection cache from context.

It is replaced with a new cache/repository implementation, held on the `db.DB` instance.  There is a shared element to this, that can be used to cache other stuff in the future (short ids in the immediate future).  The solution has been designed around the future implementation of a global cache, noted in the readme, the implementation of this is not IMO required for the bug fix and v1, so it has been left unimplemented for now.

It would have been possible to host this in the context, but I thought that was ugly and a little abusive, especially given that the cache now needs to cascade down to the store, and the previous location hid IO operations, causing them to leak into places that shouldn't have been (e.g. generate.go).  The new design also allows the cache to be maintained across contexts, which it couldn't do before.

The new implementation solves #4268 even though collection short ids are still not forbidden, the testing is a bit involved and has been broken out to #4657.  There are a handful of todos in the production code noting important, non-obvious areas to test as part of that that are tested and removed in #4657.

Is probably worth reviewing commit by commit, and on the 4th commit, start by reading `cache/README.md`.

### DIFF
--- a/internal/db/cache/README.md
+++ b/internal/db/cache/README.md
@@ -1,0 +1,35 @@
+# Cache
+
+The `cache` package provides the means to cache anything in memory in Defra.
+
+This document details how and why this *should* work, it does not necessarily detail how it is currently implemented.
+
+## Details
+
+The cache is split into several layers - `transaction`, `global` and `store`.  Each layer is free to perform additional cache layering should it chose, for example most on-disk `store` implementations will have some form of in-memory key-value caching.
+
+## Transaction layer caching
+
+The `transaction` cache layer sits on top of the `global` layer.  When it misses its cache, it will attempt to fetch the requested value from the `global` layer.
+
+The `transaction` cache populates upon read from the `global` layer to ensure that transaction isolation is maintained, and that concurrent changes made to the `global` cache to not affect ongoing transactions.  The transaction cache needs to hold onto a version number that it can pass along to the `global` cache when reading, to ensure it only reads `global` values that existed when the transaction was created - preserving transaction isolation.
+
+Each transaction-specific cache must be disposed of when the transaction is either discarded or committed.
+
+Because the `transaction` cache only needs to hold on to a smaller subset of values held in the global cache, and is relatively short-lived, it is currently believed that it is not necessary to actively manage its size.
+
+The `transaction` cache layer updates its values upon write to the transaction, there is little benefit in expiring them and requiring their re-fetching from lower layers, and this avoids us from having to worry about reading old, transaction-overwritten, values from the `global` and `store` layers.
+
+## Global layer caching
+
+The `global` cache layer sits on top of the `store` layer.  When it misses its cache, it will attempt to fetch the requested value from the `store` layer.
+
+The `global` cache populates upon read from the `store` layer.  Each cached value is versioned, and will only be yielded to transactions with a higher version.
+
+The `global` cache needs to have its size limited, this limit should be configurable.
+
+## Store layer caching
+
+The `store` layer sits at the bottom of the stack, and is responsible for fetching from the corekv store. The corekv store implementations may perform caching of their own.
+
+Access to the underlying store is driven by the corekv transaction held on the context parameter.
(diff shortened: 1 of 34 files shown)


internal/db/description/collection_cache.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Democratized Data Foundation
```
AndrewSisley, line 1: this file was deleted and rebuilt ground up, I suggest that reviewers ignore the diff and just review the new file.

internal/db/cache/txn.go
```diff
@@ -0,0 +1,147 @@
...
+
+	value, ok, err := l.lower.TryGet(ctx, key)
+	if err != nil || !ok {
+		var d TV
+		return d, false, err
+	}
+
+	txnCache.Cache(value)
+
+	return value, ok, err
```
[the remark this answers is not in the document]
AndrewSisley: Such a use case is problematic for many reasons besides this one as noted by https://github.com/sourcenetwork/defradb/issues/4626 I'm resolving this conversation and will make a note of it in the linked ticket.

internal/db/cache/txn.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return nil
+}
+
+func (l *TxnRepository[T, TV]) Forbid(value TV) {
+	l.lower.Forbid(value)
+
+	l.txnLock.RLock()
+	for _, cache := range l.cache {
+		cache.Forbid(value)
```
shahzadlone, line 144: question: This is never tested?

tests/bench/query/planner/utils.go
shahzadlone, line 1: question: What caused you to remove these? these are effected by collection cache removal?

internal/db/cache/README.md
shahzadlone, line 1: praise: I like this document, thanks

internal/db/description/collection_store.go
```diff
@@ -0,0 +1,406 @@
...
+		err = txn.Systemstore().Set(ctx, collectionVersionKey.Bytes(), []byte{})
+		if err != nil {
+			return err
+		}
+	}
+
+	i.forbiddenLock.Lock()
+	// If this transaction writes a collection version that was previously forbidden, we must unforbid it
+	// within the context of this transaction.
+	// todo - make sure the unforbidding is tested!
```
shahzadlone, line 112: todo: There is a unlinked todo here. Guessing it's not resolved yet, is there a test that tests this unforbidding case yet?

internal/db/description/collection_store.go
```diff
@@ -0,0 +1,406 @@
...
+	delete(i.forbiddenCollectionIDs, value.CollectionID)
```
shahzadlone, line 113: question: why not use txn.OnDiscard to forbid again if txn fails?

shahzadlone (commented):
I like that the cache can be maintained across contexts with this change, left some comments

internal/db/cache/txn.go
AndrewSisley: Not in this PR, see the one linked in the description for tests

tests/bench/query/planner/utils.go
AndrewSisley: Documented in the commit message, they are very invasive, costing us whenever making changes to anything they touch, they also have very limited value now that we can write benches using the integration test actions.

internal/db/description/collection_store.go
AndrewSisley: My bad I forgot to link this one, it should be covered in the follow up PR with the tests, I've added a PR todo there https://github.com/sourcenetwork/defradb/pull/4657/changes#r3023326856

Going to leave this line unlinked as there is already a PR open that addresses it and linking it isn't worth the bother.

internal/db/description/collection_store.go
AndrewSisley: What do you think we would gain by doing so?

If the txn fails, this cache is dropped anyway.

shahzadlone (approved):
LGTM

PR #4657, opened 2026-03-25
fix: Prevent doc writes after collection deleted

### LINKED ISSUE #4268: Concurrent writes during a collection version delete succeed and break commits api

### LINKED ISSUE #4663: Cannot recreate collection after deleting it
The following test errors on the last action.  The error differs if using a transaction or not.  It should not error at all:
```
	test := testUtils.TestCase{
		Actions: []any{
			&action.AddCollection{
				SDL: `
					type Users {
						name: String
					}
				`,
			},
			&action.AddCollection{
				// We don't yet have a create transaction action, so we need to create
				// one by performing an operation on anything but the `Users` collection
				// as doing anything on `Users` will acquire a read lock on the collection.
				TransactionID: immutable.Some(0),
				SDL: `
					type Books {
						name: String
					}
				`,
			},
			&action.PatchCollection{
				Patch: `
						[
							{
								"op": "remove",
								"path": "/Users"
							}
						]
					`,
			},
			&action.AddCollection{
				TransactionID: immutable.Some(0),
				SDL: `
					type Users {
						name: String
					}
				`,
			},
		},
	}

```
`error checking for known block` without a txn, `collection already exists` with a transaction.

AndrewSisley:
## Relevant issue(s)

Resolves #4268
Partially resolves #4663

## Description

Prevents doc writes after last collection version deleted.

### DIFF
--- a/internal/db/description/collection_store.go
+++ b/internal/db/description/collection_store.go
@@ -111,7 +111,6 @@ func (i *collectionStore) Write(ctx context.Context, value client.CollectionVers
 	i.forbiddenLock.Lock()
 	// If this transaction writes a collection version that was previously forbidden, we must unforbid it
 	// within the context of this transaction.
-	// todo - make sure the unforbidding is tested!
 	delete(i.forbiddenCollectionIDs, value.CollectionID)
 	i.forbiddenLock.Unlock()
 
--- a/internal/db/description/collection_store.go
+++ b/internal/db/description/collection_store.go
@@ -35,14 +35,16 @@ type collectionStore struct {
 	forbiddenCollectionIDs map[string]struct{}
 
 	lockSet              *lock.LockSet
+	txnFreeDatastore     corekv.Reader
 	collectionRepository *cache.TxnRepository[CollectionIndex, client.CollectionVersion]
 }
 
 var _ cache.Repository[CollectionIndex, client.CollectionVersion] = (*collectionStore)(nil)
 
-func newCollectionStore(lockSet *lock.LockSet) *collectionStore {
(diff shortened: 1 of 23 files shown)


internal/db/description/collection_store.go
```diff
@@ -0,0 +1,512 @@
...
+		err = txn.Systemstore().Set(ctx, collectionVersionKey.Bytes(), []byte{})
+		if err != nil {
+			return err
+		}
+	}
+
+	i.forbiddenLock.Lock()
+	// If this transaction writes a collection version that was previously forbidden, we must unforbid it
+	// within the context of this transaction.
+	// todo - make sure the unforbidding is tested!
```
AndrewSisley: todo: I missed this todo - I'm fairly sure this is tested now, but double check and then delete the todo.

- [x] Remove todo

AndrewSisley: This is covered by TestColVersionUpdateRemoveCollection_CollectionMayBeRedeclaredAndUsedByTxn, removing todo.

shahzadlone (approved):
LGTM

PR #4662, opened 2026-03-25
fix: Orphan scan @exhaustive + relation ordering

### LINKED ISSUE #4660: Orphan scan ignores scalar filters with @exhaustive + relation ordering
## Description

When using `@exhaustive` with ordering by a relation field, scalar filter conditions on the parent (primary) side are not applied to the orphan scan. This causes orphan documents that should be filtered out to incorrectly appear in query results.

## Steps to Reproduce

Given the schema:
```graphql
type Book {
    title: String
    rating: Int @index
    publisher: Publisher
}
type Publisher {
    name: String
    book: Book @primary
}
```

And data:
- One Book ("Book1", rating: 5)
- One Publisher linked to Book1 ("LinkedPublisher")
- One orphan Publisher ("IncludedOrphan", no book)
- One orphan Publisher ("ExcludedOrphan", no book)

The query:
```graphql
query @exhaustive {
    Publisher(
        filter: {name: {_neq: "ExcludedOrphan"}},
        order: {book: {rating: ASC}}
    ) {
        name
    }
}
```

**Expected:** `[IncludedOrphan, LinkedPublisher]`
**Actual:** `[ExcludedOrphan, IncludedOrphan, LinkedPublisher]`

## Root Cause

In `orphan.go:initStandaloneScan`, the top-level orphan scan builds its filter from `n.join.subFilter` (the child's scan filter). However, `prepareScanNodeFilterForTypeJoin` leaves scalar conditions (e.g., `name _neq "ExcludedOrphan"`) on the **parent's** scan — not in `subFilter`. The orphan node never sees these conditions, so its standalone scan returns unfiltered orphan documents.

## Related

Additionally, in `type_join.go:retrievePrimaryDocs`, the sub-query orphan path (`setSubQueryContext` / `setSubQueryFilter`) wraps `r.filter` with a redundant `addFilterOnField` call that adds a `relID = docID` condition, only for `addNullFilterOnField` to immediately strip and replace it with `relID IS NULL`.

fredcarle:
## Relevant issue(s)

Resolves #4660

## Description

When using `@exhaustive` with ordering by a relation field, scalar filter conditions on the parent (primary) side were not applied to the orphan scan. This caused orphan documents that should have been excluded by the filter to incorrectly appear in query results.

The problem was in `orphan.go:initStandaloneScan`. The orphan node built its filter from `join.subFilter` (the child's scan filter), but scalar conditions on the parent (e.g. `name _neq "X"`) are left on the parent's scan by `prepareScanNodeFilterForTypeJoin` and were never captured in `subFilter`. The fix merges the parent scan's filter into the orphan's base filter so the standalone scan respects all filter conditions.

Additionally, in `type_join.go:retrievePrimaryDocs`, the orphan sub-query setup wrapped `r.filter` with a redundant `addFilterOnField` call that added a `relID = docID` condition, only for `addNullFilterOnField` to immediately strip and replace it with `relID IS NULL`. This has been simplified to pass `r.filter` directly.

## Tasks

## How has this been tested?

Added `TestQueryWithIndex_ExhaustiveOrderByRelationWithScalarFilter_ShouldFilterOrphans` which sets up publishers with and without related books, applies a scalar filter excluding one orphan, and verifies only the non-excluded orphan and linked publisher appear in results. All existing `@exhaustive` ordering tests and the full index integration suite pass.

- MacOS


internal/planner/orphan.go
```diff
@@ -143,7 +143,16 @@ func (n *orphanNode) initStandaloneScan() error {
 		relationIDFieldName = n.subQueryRelIDFieldName
 	} else if n.join.parentSide.isPrimary() {
 		relIDFieldMapIndex := n.join.parentSide.relIDFieldMapIndex.Value()
-		orphanFilter = addNullFilterOnField(n.join.subFilter, relIDFieldMapIndex)
+		baseFilter := n.join.subFilter
+		// The parent scan may hold scalar filter conditions (e.g. name _neq "X") that
+		// prepareScanNodeFilterForTypeJoin left there. These are not in subFilter (which
```
AndrewSisley, line 148: todo: Discussed in person, it looks like we are fixing the wrong function, and that `prepareScanNodeFilterForTypeJoin` should be adjusted instead.

Please have a play around with that before merge

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -1185,6 +1185,74 @@ func TestQueryWithIndex_WithScalarAndRelationFilterAtTopLevel_ShouldApplyBothAsA
 	testUtils.ExecuteTestCase(t, test)
 }
 
+// TestQueryWithIndex_ExhaustiveOrderByRelationWithScalarFilter_ShouldFilterOrphans verifies
+// that @exhaustive with relation ordering still respects scalar filters on the primary type,
```
AndrewSisley, line 1189: question: What about non-scalar filters on the same object, such as JSON and inline array fields?

This might just be a documentation error, including in the production code.

AndrewSisley (approved):
LGTM, but please have a quick check to see if you are fixing the wrong function.

tests/integration/index/query_with_relation_filter_test.go
```diff
@@ -1185,6 +1185,74 @@ func TestQueryWithIndex_WithScalarAndRelationFilterAtTopLevel_ShouldApplyBothAsA
...
+// including for orphan documents (documents without a related record).
+func TestQueryWithIndex_ExhaustiveOrderByRelationWithScalarFilter_ShouldFilterOrphans(t *testing.T) {
```
islamaliev, line 1191: The test only exercises the ASC orphan-first path. Worth adding a DESC variant (orphans last) to confirm the fix holds in both orderings.

islamaliev (approved):
LGTM! Just one test case suggestion

internal/planner/orphan.go
fredcarle: Played around with this as suggested — keeping the merge-back approach rather than adjusting `prepareScanNodeFilterForTypeJoin`.

`prepareScanNodeFilterForTypeJoin` only splits the relation/FK fields onto `parent.filter` (via `getFieldsToSplitForTypeJoin`) and deliberately leaves the non-relation conditions on `scan.filter`. That's required on the normal (non-inverted) join path, where the parent scan *is* iterated directly: keeping those conditions on the scan lets the datastore scan and `selectIndex` use them (index optimization). Moving them onto `parent.filter` would shift that filtering to in-memory post-join evaluation and drop them from index selection on the hot path.

The merge-backs in `orphan.go` (`initStandaloneScan`) and `retrievePrimaryDocs` only matter on the two paths that build a *fresh* scan and ignore the original `scan.filter`. Both feed the merged filter into `selectIndex`, so the index benefit is retained there too.

(Also reworded the related "scalar filter" comments to "non-relation filter conditions" — JSON/inline-array parent filters follow the same non-complex path, which covers your other comment.)

PR #4675, opened 2026-03-29
fix: Ensure list p2p replicator nac gate works

### LINKED ISSUE #4109: Test ListReplicators/GetAllReplicators Operation with NAC
Test ListReplicators/GetAllReplicators Operation with NAC

While doing NAC gates for p2p operations I noticed that there was an un unused action for this, that isn't tested.

So add the test for normal and NAC case.

shahzadlone:
## Relevant issue(s)
Resolves #4109 

## Description
- Test list p2p replicator with NAC

### DIFF
--- a/tests/integration/p2p_replicator.go
+++ b/tests/integration/p2p_replicator.go
@@ -74,29 +74,6 @@ type DeleteReplicator struct {
 	ExpectedError string
 }
 
-// ListReplicators gets the configured replicators for the given node and compares them against the
-// expected results.
-// TODO: Test ListReplicators with and without NAC.
-// https://github.com/sourcenetwork/defradb/issues/4109
-type ListReplicators struct {
-	// NodeID is the node ID (index) of the node in which to get the replicators for.
-	NodeID int
-
-	// The identity of this request. Optional.
-	//
-	// If node acp is enabled, identity will be used to check if this operation can be performed.
-	Identity immutable.Option[state.Identity]
-
-	// ExpectedCollectionIDs are the collection IDs (indexes) of the collections expected.
-	ExpectedTargetNodeIDs []int
-
-	// Any error expected from the action. Optional.
-	//
-	// String can be a partial, and the test will pass if an error is returned that
-	// contains this string.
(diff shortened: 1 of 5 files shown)


tests/integration/p2p_replicator.go
```diff
@@ -74,29 +74,6 @@ type DeleteReplicator struct {
 	ExpectedError string
 }
 
-// ListReplicators gets the configured replicators for the given node and compares them against the
-// expected results.
-// TODO: Test ListReplicators with and without NAC.
-// https://github.com/sourcenetwork/defradb/issues/4109
-type ListReplicators struct {
```
AndrewSisley, line 81: todo: It looks like you have removed the old action, but not the old code that handled it.  Please remove that code too, as significant parts of it should now be dead.

AndrewSisley (changes requested):
It looks like the CLI client is failing the new tests.

tests/integration/p2p_replicator.go
shahzadlone: What old code that handled it (I thought this was dead code)?

The only other place I see references of `ListReplicators` (other than stuff introduced in this PR) is in the clients:
- `tests/clients/js/wrapper.go`
- `tests/clients/http/wrapper.go`
- `tests/clients/cli/wrapper.go`

shahzadlone:
> It looks like the CLI client is failing the new tests.

There was a bug found for cli client as they identity wasn't being passed to it. Should now have been fixed :)

tests/integration/p2p_replicator.go
AndrewSisley: Oh my bad, sorry - I thought you migrated an existing action to the `action` package :)

AndrewSisley (approved):
LGTM, thanks Shahzad.

suggestion: Consider changing the PR title to `fix:...` as this now fixes a bug

shahzadlone:
> suggestion: Consider changing the PR title to `fix:...` as this now fixes a bug

Good suggestion, will do!

PR #4680, opened 2026-03-31
fix: Resolve the identity panic when no keyring

### LINKED ISSUE #4580: no keyring and identity panic
# 1. Build defradb
  `go build -o /tmp/defradb-bin ./cmd/defradb`

  # 2. Start with local ACP and no keyring
 ` /tmp/defradb-bin start --no-keyring --rootdir /tmp/defradb-repro --document-acp-type local`

  # 3. In another terminal, add a policy
 ` /tmp/defradb-bin client --no-keyring --rootdir /tmp/defradb-repro \
    acp document policy add -f examples/policy/dac_policy.yml \
    -i e3b722906ee4e56368f581cd8b18ab0f48af1ea53e635e3f7b8acd076676f6ac`

  # 4. Add a schema linked to the policy (use the PolicyID from step 3)
  `/tmp/defradb-bin client --no-keyring --rootdir /tmp/defradb-repro \
    schema add 'type Users @policy(id: "<PolicyID from step 3>", resource: "users") { name: String }'`

  # 5. Add a document WITH identity - this triggers the panic
  `/tmp/defradb-bin client --no-keyring --rootdir /tmp/defradb-repro \
    collection add --name Users '{"name": "Secret"}' \
    -i e3b722906ee4e56368f581cd8b18ab0f48af1ea53e635e3f7b8acd076676f6ac`

  Step 5 returns Error: 500: and the server logs show:

 ```
panic: runtime error: invalid memory address or nil pointer dereference

  -> github.com/sourcenetwork/defradb/internal/core/block.signBlock
  ->   internal/core/block/signing.go:73
```

  The panic is at ident.PrivateKey().Type() - when --no-keyring is used, the identity has no private key, so
  PrivateKey() returns nil. Adding a document without --identity (step 4, schema add) works fine because no
  signing is attempted. The crash only happens when both --no-keyring and --identity are used together on a
  write operation.

shahzadlone:
## Relevant issue(s)
Resolves #4580 

## Description
- Handle the panic
- Write unit tests

### DIFF
--- a/internal/core/block/signing.go
+++ b/internal/core/block/signing.go
@@ -44,7 +44,7 @@ func extractFullIdentity(ctx context.Context) immutable.Option[identity.FullIden
 	}
 
 	fullIdent, ok := ident.Value().(identity.FullIdentity)
-	if !ok {
+	if !ok || fullIdent.PrivateKey() == nil {
 		return immutable.None[identity.FullIdentity]()
 	}
 
--- a/internal/core/block/signing_test.go
+++ b/internal/core/block/signing_test.go
@@ -1,0 +1,80 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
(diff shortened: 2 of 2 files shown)


nasdf (approved):
LGTM

PR #4674, opened 2026-03-29
chore: Remove docs website

SharanRP:
## Relevant issue(s)

Resolves #4655

## Description

This change removes the in-repository website documentation by deleting the `docs/website` directory. To keep the repository consistent after that removal, it also updates documentation related automation so it no longer depends on files under `docs/website`.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

I tested the updated docs targets by running `make docs:cli` and `make docs:http` on  Linux. The CLI docs were generated under `build/docs/references/cli`, and the HTTP OpenAPI output was generated under `build/docs/references/http/openapi.json`. The changes verify that docs generation and CI checks no longer rely on `docs/website`.

PR #4681, opened 2026-03-31
fix: Fix cli commands markdown output

### LINKED ISSUE #4655: Remove docs from repo
The docs will no longer be hosted within the repo. `docs/website` should be removed.

### LINKED ISSUE #4757: Broken markdown upon cli commands docs generation
There's [an issue](https://github.com/spf13/cobra/issues/2375) in cobra with non-escaped angle brackets (ex. `<output>`) that results in broken markdown, unless the string is placed within inline/block code.

ribiza:
## Relevant issue(s)

Resolves #4757

## Description

There's [an issue](https://github.com/spf13/cobra/issues/2375) in cobra with non-escaped angle brackets (ex. `<output>`) that results in broken markdown, unless the string is placed within inline/block code. 
For the time being we need to alter the source so that the output markdown will be error-free and can be published.

## How has this been tested?

`npm start` from the docs repo, with new cli docs files.

- Debian Linux


docs/website/getting-started.md
```diff
@@ -61,41 +107,59 @@ defradb client collection add '
...
   mutation {
-      add_User(input: {age: 31, verified: true, points: 90, name: "Bob"}) {
+      user1: add_User(input: {age: 31, verified: true, points: 90, name: "Bob"}) {
+          _docID
+      }
+      user2: add_User(input: {age: 28, verified: false, points: 15, name: "Alice"}) {
+          _docID
+      }
+      user3: add_User(input: {age: 35, verified: true, points: 100, name: "Charlie"}) {
           _docID
```
shahzadlone, line 114: question: have you manually tried this command? I would strongly suggest every user facing command/example is somehow ran and caught if it breaks as we don't want to mislead users.

docs/website/getting-started.md
```diff
@@ -289,41 +403,46 @@ defradb client collection add '
...
 
 Here we *do not* specify `--peers` as we will manually define a replicator after startup via the `rpc` client command.
 
-In another terminal, add the same collection to *nodeB*:
+In another terminal, add the same schema to *nodeB*:
 
-```shell
-defradb client collection add --url localhost:9182 '
+```bash
+defradb client schema add --url localhost:9182 '
```
shahzadlone: question: why did you change this?

shahzadlone (commented):
Just did a quick pass through and highlighted some stuff, will do another soon.

question: How much of this is AI assisted?

question: What caused you to change the `collection` terminology to `schema` terminology?

ribiza:
> question: How much of this is AI assisted?

None of it. Any slop, it's mine.

> question: What caused you to change the collection terminology to schema terminology?

I experimented with different workflows and options and moved files around many times, I think I just ended up committing the wrong version of it, and I didn't yet expect anybody to review this :hugs: Let me check again the getting started.

docs/website/getting-started.md
ribiza: Yes I have. I agree that we need to have some automated testing on commands we show. That's an infrastructure we should build as a next priority - as soon as docs and workflows for v1 are settled, I'd say.

ribiza:
> question: How much of this is AI assisted?

To elaborate on this, as it is perhaps not obvious: the initial version was just a copy-paste from the docs repo into this. With this pass I've taken a more thorough look. At this stage I'm only interested in the content being accurate, and bringing the two docs repos in sync again. I'll polish and refine content afterward.

shahzadlone:
Closing this PR as the internal docs team you likely take care of this

shahzadlone:
Ops my bad, sorry @ribiza I wanted to close this one: https://github.com/sourcenetwork/defradb/pull/4674

cli/keyring.go
```diff
@@ -28,13 +28,13 @@ The following keys are loaded from the keyring on start:
 	encryption-key: AES-128, AES-192, or AES-256 key (optional)
 
 To randomly generate the required keys, run the following command:
-	defradb keyring new
+	` + "`defradb keyring new`" + `
```
shahzadlone, line 31: question: Why do the `+`?

ribiza: On this and line 37 I did it for consistency, but it's not needed. Only line 34 needs the example to be backticked (and thus quoted with `"` and string-concatenated with `+`). 
Should I revert lines 31 and 37? I thought it would make it easier to avoid mistakes for future examples which would likely copy-paste, but it does look ugly.

docs/website/references/cli/defradb_keyring.md
shahzadlone, line 32: suggestion: Since you are already touching this file, wonder if we should fix the tab indented lines as it's under back ticks, also CommonMark may treat it like code block (perhaps docusaurus works well if disable indented code block on our theme?).

docs/website/getting-started.md
```diff
@@ -234,49 +332,55 @@ Start *nodeA* with a default configuration:
...
-- `--tcpaddr` is the multiaddress for the gRPC server to listen on
+- `--p2paddr` is a comma-separated list of multiaddresses to listen on for p2p networking
 - `--peers` is a comma-separated list of peer multiaddresses
 
 This starts two nodes and connects them via pubsub networking.
 
-### Collection subscription example
-
-It is possible to subscribe to updates on a given collection by using its ID as the pubsub topic. The ID of a collection is found as the field `collectionVersionID` in one of its documents. Here we use the collection ID of the `User` type we created above. After setting up 2 nodes as shown in the [Pubsub example](#pubsub-example) section, we can subscribe to collections updates on *nodeA* from *nodeB* by using the `rpc p2pcollection` command:
+To subscribe to updates on a collection, provide its ID as the pubsub topic. The ID of a collection is found as the field `collectionID` in one of its documents. To subscribe to collections updates on *nodeA* from *nodeB*, use [defradb client p2p collection add](https://docs.source.network/defradb/0.20.0/references/cli/defradb_client_p2p_collection_add):
```
shahzadlone, line 358: todo: We should not hardcode version in the URL IMO. Also while this says `0.20.0`, the docker example on line 38 uses `1.0.0-rc1` (that is also inconsistent even if you want to maintain a hard coded one). I suggest using a version-agnostic relative link.

docs/website/getting-started.md
```diff
@@ -183,48 +257,72 @@ It returns a structure similar to the following, which contains the update paylo
...
-Pubsub peering *passively* synchronizes data between nodes by broadcasting *Document Commit* updates to the topic of the commit's document key. Nodes need to be listening on the pubsub channel to receive updates. This is for when two nodes *already* have share a document and want to keep them in sync.
+```shell
+defradb client p2p info
+```
 
-Replicator peering *actively* pushes changes from a specific collection *to* a target peer.
+Two types of peer-to-peer relationships are supported: **pubsub** and **replicator**.
 
-### Pubsub example
+- Pubsub peering -- *Passively* synchronizes data between nodes by broadcasting document commit updates to the topic of the collection ID. Nodes need to be listening on the pubsub channel to receive updates. This is for when two nodes have already a shared collection and want to keep their documents in sync.
```
shahzadlone, line 319: nitpick: Reads a bit odd perhaps can improve grammer in `"This is for when two nodes have already a shared collection...``

docs/website/getting-started.md
```diff
@@ -306,50 +414,96 @@ defradb client collection add --url localhost:9182 '
...
 ```shell
 defradb start --tls
 ```
 
 The keys can be generated with your generator of choice or with `make tls-certs`.
 
-Since the keys should be stored within the DefraDB data and configuration directory, the recommended key generation command is `make tls-certs path="~/.defradb/certs"`.
+Since the keys should be stored within the DefraDB root directory, generate keys with `make tls-certs path="~/.defradb/certs"`.
 
 If not saved under `~/.defradb/certs` then the public (`pubkeypath`) and private (`privkeypaths`) key paths need to be explicitly defined in addition to the `--tls` flag or `tls` set to `true` in the config.
```
shahzadlone, line 453: todo: The flag is `privkeypath` I believe (singular). As used correctly in the example below.
```suggestion
If not saved under `~/.defradb/certs` then the public (`pubkeypath`) and private (`privkeypath`) key paths need to be explicitly defined in addition to the `--tls` flag or `tls` set to `true` in the config.
```

cli/keyring.go
shahzadlone: So you did this because there's no escape for a backtick inside a raw string if i understand correctly. Then it's fine, resolving :)

docs/website/getting-started.md
```diff
@@ -306,50 +414,96 @@ defradb client collection add --url localhost:9182 '
...
 
-By default, DefraDB will expose its HTTP API at `http://localhost:9181/api`. It's also possible to configure the API to use TLS with self-signed certificates or Let's Encrypt.
+By default, DefraDB exposes its HTTP API at `http://localhost:9181/api/v1`. You can also configure the API to use TLS with self-signed certificates or Let's Encrypt.
+
+To start defradb with self-signed certificates placed under `~/.defradb/certs/` with `server.crt` being the public key and `server.key` the private key, execute:
 
-To start defradb with self-signed certificates placed under `~/.defradb/certs/` with `server.key`
-being the public key and `server.crt` being the private key, just do:
 ```shell
 defradb start --tls
```
shahzadlone, line 446: question: Is there a `--tls` flag on start (cli/start.go)? Also IIRC TLS is activated implicitly when both --pubkeypath and --privkeypath are set (http/server.go:98-102)

docs/website/getting-started.md
```diff
@@ -306,50 +414,96 @@ defradb client collection add --url localhost:9182 '
...
 ```shell
-defradb client rpc addreplicator "Article" /ip4/0.0.0.0/tcp/9172/p2p/
-defradb client rpc replicator set -c "Article" /ip4/0.0.0.0/tcp/9172/p2p/<peerID_of_nodeB>
+defradb client p2p info --url localhost:9182
+```
+
+Set *nodeA* to actively replicate the Article collection to *nodeB*:
 
+```shell
+defradb client p2p replicator add -c Article <nodeB_peer_info_json>
```
shahzadlone, line 426: question: Can you confirm that info can be taken by JSON blob? I have not used that before. Usually in-repo examples even I see us pass multiaddrs directly: `defradb client p2p replicator add -c Article /ip4/0.0.0.0/tcp/9171/p2p/12D3Ko....`

docs/website/getting-started.md
```diff
@@ -234,49 +332,55 @@ Start *nodeA* with a default configuration:
...
 This starts two nodes and connects them via pubsub networking.
 
-### Collection subscription example
-
-It is possible to subscribe to updates on a given collection by using its ID as the pubsub topic. The ID of a collection is found as the field `collectionVersionID` in one of its documents. Here we use the collection ID of the `User` type we created above. After setting up 2 nodes as shown in the [Pubsub example](#pubsub-example) section, we can subscribe to collections updates on *nodeA* from *nodeB* by using the `rpc p2pcollection` command:
+To subscribe to updates on a collection, provide its ID as the pubsub topic. The ID of a collection is found as the field `collectionID` in one of its documents. To subscribe to collections updates on *nodeA* from *nodeB*, use [defradb client p2p collection add](https://docs.source.network/defradb/0.20.0/references/cli/defradb_client_p2p_collection_add):
 
 ```shell
-defradb client rpc p2pcollection add --url localhost:9182 bafkreibpnvkvjqvg4skzlijka5xe63zeu74ivcjwd76q7yi65jdhwqhske
+defradb client p2p collection add --url localhost:9182 bafkreibpnvkvjqvg4skzlijka5xe63zeu74ivcjwd76q7yi65jdhwqhske
```
shahzadlone, line 361: question: Maybe I am missing something, should this expect ID or Name instead (`internal/db/p2p/collection.go:44` does `SetCollectionName(col)` and returns `ErrCollectionNotFoundForName` on miss)?  We have some examples I believe like `defradb client p2p collection add User,Address`

shahzadlone, line 361: todo: This doesn't seem correct to me: "The ID of a collection is found as the field collectionID in one of its documents". 

The collection ID lives on the collection metadata, I think we can obtain it like `defradb client collection describe --name User` (client/collection_description.go:38).

docs/website/getting-started.md
```diff
@@ -4,51 +4,85 @@ title: Getting Started
...
+DefraDB has a built-in keyring for storing private keys securely. Keys will be randomly generated on the initial start of the node if they are not found. The following keys are loaded from the keyring on start:
 
-The GraphQL endpoint can be used with a GraphQL client (e.g., Altair) to conveniently perform requests (`query`, `mutation`) and obtain schema introspection.
+- `peer-key` Ed25519 private key (required)
+- `encryption-key` AES-128, AES-192, or AES-256 key (optional)
+- `node-identity-key` Secp256k1 private key (optional). This key establishes the node's identity, and is used to exchange encryption keys with other nodes.
 
-## Add a collection
+You can manually generate keys via the [`keyring new` command](/references/cli/defradb_keyring_new.md). By default, keys are stored in the `defradb` keyring under the password `secret`. You can initialize the keyring with a different password via the `DEFRA_KEYRING_SECRET` environment variable, via a `.env` file located in the working directory, or via a file at a path defined by the `--secret-file` flag.
```
shahzadlone, line 55: suggestion: I dislike the use of the word `password` here and in other places, IMO `secret` suffices. To me use of keys are different from passwords.

docs/website/getting-started.md
```diff
@@ -4,51 +4,85 @@ title: Getting Started
...
+To experiment with queries, you can use either the playground at `http://localhost:9181`, or a native GraphQL client ([Altair](https://altairgraphql.dev/#download) is a popular option). DefraDB's GraphQL endpoint is at `http://localhost:9181/api/v1/graphql` (the versionless endpoint `http://localhost:9181/api/graphql` always points to the latest version).
 
-Verify the local connection to the node works by executing `defradb client ping` in another terminal.
+The [`client` CLI command](/references/cli/defradb_client.md) interacts with the locally running node.
 
-## Configuration
+DefraDB's root directory is located at `~/.defradb/`. Among other things, it contains the [configuration file](/references/config.md).
 
-In this document, we use the default configuration, which has the following behavior:
+## Key Management - Initial Setup
```
shahzadlone, line 47: question: This section seems to suggest that all 3 keys are generated. Note: `cli/keyring_new.go:24-87` only creates peer-key and encryption-key. The node-identity-key is lazily generated at `defradb start (cli/start.go:414-456)`. Users who run `keyring new` expecting identity to be provisioned will be surprised.

docs/website/getting-started.md
```diff
@@ -4,51 +4,85 @@ title: Getting Started
...
 
-- `~/.defradb/` is DefraDB's configuration and data directory
-- `client` command interacts with the locally running node
-- The GraphQL endpoint is provided at <http://localhost:9181/api/graphql> and a versioned API endpoint is provided at <http://localhost:9181/api/v1/graphql>
+DefraDB has a built-in keyring for storing private keys securely. Keys will be randomly generated on the initial start of the node if they are not found. The following keys are loaded from the keyring on start:
 
-The GraphQL endpoint can be used with a GraphQL client (e.g., Altair) to conveniently perform requests (`query`, `mutation`) and obtain schema introspection.
+- `peer-key` Ed25519 private key (required)
+- `encryption-key` AES-128, AES-192, or AES-256 key (optional)
+- `node-identity-key` Secp256k1 private key (optional). This key establishes the node's identity, and is used to exchange encryption keys with other nodes.
```
shahzadlone, line 53: todo: The type is driven by `datastore.defaultkeytype` (default secp256k1, but can be ed25519 aswell)

shahzadlone (approved):
LGTM just left some comments, happy for you to merge after they are resolved

docs/website/getting-started.md
ribiza: You are right, it was removed in 2024 https://github.com/sourcenetwork/defradb/commit/ae6cebe49a562ad99952346d8ebd6df77da22566 (and perhaps made obsolete already by earlier commits).

> TLS is activated implicitly when both --pubkeypath and --privkeypath are set

Thanks for checking.

docs/website/getting-started.md
ribiza: very good point.

docs/website/references/cli/defradb_keyring.md
ribiza: You are right. The md is auto-generated, so I fixed it by using `EmbedCLIExample` for examples in the source `keyring.go`.

docs/website/getting-started.md
ribiza: I think you are right: it takes collections names. I have not tried out the p2p set up yet, I had only updated the renamed commands and... didn't check whether they worked or not 🥇 I'll update this command but defer to later the proper care of this section.

docs/website/getting-started.md
ribiza: Not at all. I have no idea what source that example had, but it doesn't work. I've put the multiaddr explicitly. This section will really need to be revisited.

PR #4656, opened 2026-03-25
fix: Treat multiple filters as AND when join is inverted

islamaliev:
## Relevant issue(s)

Resolves #4314

## Description

When a query filter combines scalar conditions with a relation condition at the top level without explicit `_and`, the scalar conditions were silently dropped if the join got inverted for index optimization. For example, `Book(filter: {rating: {_gt: 4.5}, author: {name: {_eq: "John Grisham"}}})` would ignore the rating condition and return all books by John Grisham. Using explicit `_and` worked around this because it triggered a different code path that kept all conditions together.

The problem was in `retrievePrimaryDocs`. When the join is inverted, the parent scan becomes the second side and its filter gets overwritten with just the relation ID lookup. Any scalar conditions that were sitting on the parent scan were lost. The fix merges the parent scan's original filter into the retrieval filter before overwriting, and properly saves/restores it afterward.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

Added integration tests covering mixed scalar + relation filters with and without explicit `_and`, verifying they produce the same results.

Specify the platform(s) on which this was tested:
- MacOS

PR #4684, opened 2026-04-01
fix: Sec. index edge cases in subscriptions and inv. joins

### LINKED ISSUE #4353: Sec. index: Fix all failing cases that are excluded for sec. index multiplier
We added test multiplier for sec. indexes that makes sure all regular (non-index) tests when injected with indexes produce the same result.

Currently some of such tests fail.

Look for `TODO: https://github.com/sourcenetwork/defradb/issues/4353`

islamaliev:
## Relevant issue(s)

Resolves #4353

## Description

Two fixes for cases where the secondary-index test multiplier exposed bugs in the query planner.

**Subscriptions with CID-based queries** were timing out because the planner selected a secondary index even though the `VersionedFetcher` operates on a temporary in-memory store that doesn't contain index data. The index lookup returned empty results, causing the subscription to never fire. The fix skips index selection when CIDs are present, since `VersionedFetcher` only supports the primary index.

**One-to-many inverted joins** returned fewer results because the outer iteration loop and `retrievePrimaryDocs` shared the same plan tree. When `retrievePrimaryDocs` reinitialized the plan, the `orderNode` consumed all fetcher results on its first pass, leaving nothing for subsequent iterations. The fix splits `retrievePrimaryDocs` into two paths: inverted joins use an independent cloned `scanNode` wrapped in an `orderNode`, while non-inverted joins continue using the original plan tree with save/restore. This also eliminates the manual sort that was previously needed — the `orderNode` handles both cases (index-ordered pass-through and in-memory sort).

The upstream fix for scalar filter merging in inverted joins (#4656) is incorporated into the new split-path structure.

Three `MultiplierExcludes` removed.

## How has this been tested?

Existing integration tests pass with and without the secondary-index multiplier. Added deterministic ordering to the one-to-many test to ensure consistent results across index/no-index modes.

Tested on macOS.


tests/integration/txn/patch_collection_test.go
```diff
@@ -14,16 +14,21 @@ package txn_testing
...
-	"github.com/sourcenetwork/immutable"
 )
 
 // This test runs PatchCollection inside of a transaction, and illustrates that commiting the transaction
 // results in the patch being applied.
 func TestTxn_PatchCollection_WithCommit_Succeeds(t *testing.T) {
 	test := testUtils.TestCase{
+		// The secondary-index multiplier adds @index directives which shifts field indices,
+		// causing the hardcoded "Fields/2" path to target the wrong field.
+		MultiplierExcludes: []string{multiplier.SecondaryIndex},
```
shahzadlone: question: I don't fully understand this, so PatchCollection fails with secondary indexes now?

internal/planner/type_join.go
```diff
@@ -694,21 +706,44 @@ func (r *primaryObjectsRetriever) retrievePrimaryDocs() ([]core.Doc, error) {
 				relFieldName, _ := r.primaryScan.documentMapping.TryToFindNameFromIndex(relFieldIndex)
 				relIDFieldName := request.ToFieldID(relFieldName)
 				relIDFieldMapIndex := r.primaryScan.documentMapping.FirstIndexOfName(relIDFieldName)
-				parentFilter := addFilterOnField(r.filter, r.primarySide.relIDFieldMapIndex.Value(),
-					r.targetSecondaryDoc.GetID())
-				orphan.setSubQueryContext(parentFilter, relIDFieldName, relIDFieldMapIndex)
+				orphan.setSubQueryContext(docFilter, relIDFieldName, relIDFieldMapIndex)
```
shahzadlone, line 714: question: No test hits this line

internal/planner/type_join.go
```diff
@@ -635,16 +630,55 @@ func (r *primaryObjectsRetriever) collectDocs() ([]core.Doc, error) {
...
+) ([]core.Doc, error) {
+	clone := r.primaryScan.cloneWithFilter(docFilter, index, ordering)
+
+	var source planNode = clone
+	if len(ordering) > 0 {
+		source = &orderNode{
+			p:         r.primaryScan.p,
+			plan:      clone,
+			ordering:  ordering,
+			needSort:  true,
```
fredcarle, line 664: question: Shouldn't this be false if the index can satisfy the order? Otherwise there might be an extra in-memory sort that isn't needed.

fredcarle (approved):
LGTM. Just one question to resolve before merge.

tests/integration/txn/patch_collection_test.go
islamaliev: Yeah, not PatchCollection itself, but the test uses a hardcoded JSON pointer path Fields/2 to remove a field. When the secondary-index multiplier adds @index to the SDL, it changes the structure of the Fields array, so index 2 no longer points to the field the test expects. The patch then targets the wrong element and the test breaks. Anything that hardcodes field indices into a patch path is fragile to schema structure changes. Removing the exclude would require rewriting the test to patch by field name instead of index position.

internal/planner/type_join.go
islamaliev: You're right, no test was hitting it in standard coverage runs. It was covered under DEFRA_MULTIPLIERS=secondary-index (via the SUM tests in one_to_many_to_one) but that's not captured by codecov. Added `TestQueryWithNestedOrderByRelationField_ExhaustiveWithPrimaryParentASC_ShouldIncludeOrphans` to cover it explicitly.

tests/integration/txn/patch_collection_test.go
AndrewSisley: > When the secondary-index multiplier adds @index to the SDL, it changes the structure of the Fields array, so index 2 no longer points to the field the test expects.

Can you expand on this a little bit please?  How do indexes change the field indexes, and when?  If you add an index after the collection has been created, does it move existing fields?

internal/planner/type_join.go
islamaliev: You're right about the asymmetry. When the index can satisfy ordering, the non-inverted path skips the in-memory sort entirely, but the inverted path was always wrapping in an orderNode regardless.

Your suggested fix of setting `needSort: false` in the constructor wouldn't work though. `Init()` unconditionally resets it to true before iteration starts, so the value gets thrown away before the first `Next()`.

The fix is to skip wrapping in orderNode altogether when canSatisfyOrder is true, mirroring what the non-inverted path does. I've implemented this:                                            

```go
if len(ordering) > 0 && !canSatisfyOrder {
    // only wrap in orderNode when the index can't satisfy ordering
}
```

Thanks for bringing this my attention into it. I had to do some more investigation.

I've also made the ephemeral orderNode iterations visible in the explain output (they were previously invisible since the clone is created at runtime, not in the plan tree). The existing test `TestQueryWithOrderOnOneToMany_WithParentFilterOnRelationAndSubOrder_ShouldOrderChildren` now asserts iterations: 5. Without the optimization it would be 8 (the extra 3 being the redundant orderNode buffering + re-sorting).

tests/integration/txn/patch_collection_test.go
islamaliev: hey, guys. When debugging I hastily made incorrect conclusion that the index caused the fields to shift. The more precise reason is that the patching just returns error "field does not exist. Name: name". The reason is that the field is indexes and thus can't be remove. I create an issue to improve error message to be more precise https://github.com/sourcenetwork/defradb/issues/4722 and updated the comments in code

PR #4688, opened 2026-04-07
feat: Add syntax sugar DeleteCollection cmd

### LINKED ISSUE #4678: Collection Delete syntax sugar command
Was discussed by the team to have the `collection delete` syntax sugar command

shahzadlone:
## Relevant issue(s)
Resolves #4678 

## Description
- Implement `collection delete` sugar syntax command
- Tie to the same permission as the Patch Collection permission

### DIFF
--- a/cbindings/delete_collection.go
+++ b/cbindings/delete_collection.go
@@ -1,0 +1,48 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cbindings
+
+/*
+#include <stdlib.h>
+#include "defra_structs.h"
+*/
+import "C"
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/defradb/client/options"
+	iIdentity "github.com/sourcenetwork/defradb/internal/identity"
+)
+
+//export DeleteCollection
+func DeleteCollection(nodePtr C.uintptr_t, name *C.char, identityPtr C.uintptr_t) C.Result {
+	ctx := context.Background()
+
+	ctx, err := contextWithIdentity(ctx, identityPtr)
+	if err != nil {
+		return returnC(returnGoC(1, err.Error(), ""))
+	}
+
+	store, err := getStoreFromPointer(nodePtr)
+	if err != nil {
+		return returnC(returnGoC(1, err.Error(), ""))
+	}
+
+	ctx = attachTxnFromPointer(nodePtr, ctx)
+
+	opt := options.WithIdentity(options.DeleteCollection(), iIdentity.FromContext(ctx))
+	err = store.DeleteCollection(ctx, C.GoString(name), opt)
+	if err != nil {
+		return returnC(returnGoC(1, err.Error(), ""))
+	}
+	return returnC(returnGoC(0, "", ""))
+}
--- a/cbindings/delete_collection.go
+++ b/cbindings/delete_collection.go
@@ -18,13 +18,14 @@ import "C"
 
 import (
 	"context"
+	"strings"
 
 	"github.com/sourcenetwork/defradb/client/options"
 	iIdentity "github.com/sourcenetwork/defradb/internal/identity"
 )
 
 //export DeleteCollection
-func DeleteCollection(nodePtr C.uintptr_t, name *C.char, identityPtr C.uintptr_t) C.Result {
+func DeleteCollection(nodePtr C.uintptr_t, names *C.char, identityPtr C.uintptr_t) C.Result {
 	ctx := context.Background()
 
 	ctx, err := contextWithIdentity(ctx, identityPtr)
@@ -40,7 +41,13 @@ func DeleteCollection(nodePtr C.uintptr_t, name *C.char, identityPtr C.uintptr_t
 	ctx = attachTxnFromPointer(nodePtr, ctx)
 
 	opt := options.WithIdentity(options.DeleteCollection(), iIdentity.FromContext(ctx))
-	err = store.DeleteCollection(ctx, C.GoString(name), opt)
+
+	var nameList []string
+	if joined := C.GoString(names); joined != "" {
+		nameList = strings.Split(joined, ",")
+	}
+
+	err = store.DeleteCollection(ctx, nameList, opt)
 	if err != nil {
 		return returnC(returnGoC(1, err.Error(), ""))
 	}
--- a/cbindings/delete_collection.go
+++ b/cbindings/delete_collection.go
@@ -25,7 +25,12 @@ import (
 )
 
 //export DeleteCollection
-func DeleteCollection(nodePtr C.uintptr_t, names *C.char, identityPtr C.uintptr_t) C.Result {
+func DeleteCollection(
+	nodePtr C.uintptr_t,
+	names *C.char,
+	activeOnly C.int,
+	identityPtr C.uintptr_t,
+) C.Result {
 	ctx := context.Background()
 
 	ctx, err := contextWithIdentity(ctx, identityPtr)
@@ -41,6 +46,7 @@ func DeleteCollection(nodePtr C.uintptr_t, names *C.char, identityPtr C.uintptr_
 	ctx = attachTxnFromPointer(nodePtr, ctx)
 
 	opt := options.WithIdentity(options.DeleteCollection(), iIdentity.FromContext(ctx))
+	opt.SetActiveOnly(activeOnly != 0)
 
 	var nameList []string
 	if joined := C.GoString(names); joined != "" {
--- a/cli/collection_delete.go
+++ b/cli/collection_delete.go
@@ -1,0 +1,52 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package cli
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client/options"
+	"github.com/sourcenetwork/defradb/internal/identity"
+)
+
+func MakeCollectionDeleteCommand(ctx context.Context) *cobra.Command {
+	var name string
+	var cmd = &cobra.Command{
+		Use:   "delete --name <name>",
+		Short: "Delete the active collection version",
+		Long: `Delete the active version of a collection by name.
+
+Only the latest (head) version is deleted per call. If the collection has multiple
+versions, earlier versions must be deleted separately after each head is removed.
+
+The collection must not contain any documents. Delete all documents first before
+deleting the collection.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if name == "" {
+				return NewErrRequiredFlagEmpty("name", "n")
+			}
+
+			cliClient := mustGetContextCLIClient(cmd)
+
+			opt := options.WithIdentity(options.DeleteCollection(), identity.FromContext(cmd.Context()))
+			return cliClient.DeleteCollection(cmd.Context(), name, opt)
+		},
+	}
+
+	EmbedCLIExample(ctx, cmd, "delete a collection by name",
+		`defradb client collection delete --name Users`)
+
+	cmd.Flags().StringVarP(&name, "name", "n", "", "Collection name to delete")
+
+	return cmd
+}
--- a/cli/collection_delete.go
+++ b/cli/collection_delete.go
@@ -12,6 +12,7 @@ package cli
 
 import (
 	"context"
+	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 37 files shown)


internal/db/store.go
```diff
@@ -245,6 +245,45 @@ func (db *DB) PatchCollection(
...
+	}
+
+	ctx, txn, err := ensureContextTxn(ctx, db, false)
+	if err != nil {
+		return err
+	}
+
+	defer txn.Discard()
+
+	col, err := db.getCollectionByName(ctx, name)
```
AndrewSisley, line 269: todo: This seems out of place, you are including business logic in the boilerplate layer.  This hides it, and makes it harder to call internally.

Please pull lines 269-282 out to a `deleteCollection` function

internal/db/store.go
```diff
@@ -245,6 +245,45 @@ func (db *DB) PatchCollection(
...
+	if err != nil {
+		return err
+	}
+
```
AndrewSisley, line 273: todo: You do not appear to be performing any validation on the changes, please make sure `validateCollectionChanges` is called.

Please also make sure you have tests covering stuff that the validation might flag, such as deleting collections that are referenced by other collections.

tests/integration/collection/delete/delete_collection_test.go
```diff
@@ -0,0 +1,400 @@
...
+					type Users {
+						name: String
+					}
+					type Books {
+						title: String
+					}
+				`,
+			},
+			&action.DeleteCollection{
+				Name: "Users",
```
AndrewSisley, line 163: todo: Related to the validation todo - this action looks like it will result in an invalid GQL schema, you should include a test that verifies that the resultant GQL schema is actually valid, `GetCollections` is not sufficient for this.

AndrewSisley (changes requested):
I think this needs a few tweaks before it can be merged.

thought: The description is incorrect, `Implement collection delete sugar syntax command` - the current implementation in the PR is not syntax sugar, it happens to call the same internal function, but it is not syntax sugar - this is apparent in the skipping of the validation, and any other logic current and future, that PatchCollection performs.  If we want it to be syntax sugar, then it should call `PatchCollection`.

internal/db/store.go
shahzadlone: lmk if happy with the changes.

internal/db/store.go
```diff
@@ -245,6 +247,59 @@ func (db *DB) PatchCollection(
...
+	err = db.deleteCollection(ctx, names)
+	if err != nil {
+		return err
+	}
+
+	return txn.Commit()
+}
+
+func (db *DB) deleteCollection(ctx context.Context, names []string) error {
```
AndrewSisley: suggestion:  A few collection operations take `GetCollectionsOptions` as a parameter, instead of just name(s) - it might be better if this did that too and the amount of extra code should be minimal.

cli/collection_delete.go
```diff
@@ -0,0 +1,66 @@
...
+		Use:   "delete [collectionNames]",
+		Short: "Delete the active collection versions",
+		Long: `Delete the active version of one or more collections by name.
+
+A single name, or a comma-separated list of names, may be provided. All named
+collections are removed atomically in a single operation. This can be used to
+delete collections that reference each other via relations, since deleting them
+one at a time would leave a dangling reference and be rolled back.
+
+Only the latest (head) version is deleted per call. If a collection has multiple
```
AndrewSisley: thought: I think we did talk about this in a standup sometime, but this does feel a bit like this function is struggling to find its identity, and that this is preserving a significant amount of the inconvenience that the lower level `PatchCollection` does.

cli/collection_delete.go
```diff
@@ -0,0 +1,66 @@
...
+versions, earlier versions must be deleted separately after each head is removed.
+
+The named collections must not contain any documents. Delete all documents first
+before deleting the collection.`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var names []string
+			for name := range strings.SplitSeq(args[0], ",") {
```
AndrewSisley, line 44: question: Do we do this the csv approach for other commands?  It feels strange to me.  It feels like it should really be space deliminated.

cbindings/delete_collection.go
```diff
@@ -0,0 +1,55 @@
...
+		return returnC(returnGoC(1, err.Error(), ""))
+	}
+
+	ctx = attachTxnFromPointer(nodePtr, ctx)
+
+	opt := options.WithIdentity(options.DeleteCollection(), iIdentity.FromContext(ctx))
+
+	var nameList []string
+	if joined := C.GoString(names); joined != "" {
+		nameList = strings.Split(joined, ",")
```
AndrewSisley, line 53: todo: This feels very odd - there are arrays in C.  Why are you requiring a csv string instead?

http/handler_store.go
```diff
@@ -142,6 +142,35 @@ func (h *storeHandler) PatchCollection(rw http.ResponseWriter, req *http.Request
...
 }
 
+func (h *storeHandler) DeleteCollection(rw http.ResponseWriter, req *http.Request) {
+	db := mustGetContextClientDB(req)
+	ctx := req.Context()
+
+	txn, hadTxn := datastore.CtxTryGetClientTxn(ctx)
+
+	names := req.URL.Query()["name"]
+	if len(names) == 0 {
```
AndrewSisley: question: This is the responsibility of the core code, not the http client - why are you doubling up the validation here?

http/handler_store.go
```diff
@@ -801,6 +830,17 @@ func (h *storeHandler) bindRoutes(router *Router) {
 	patchCollection.Responses.Set("200", successResponse)
 	patchCollection.Responses.Set("400", errorResponse)
 
+	deleteCollection := openapi3.NewOperation()
+	deleteCollection.OperationID = "delete_collection"
+	deleteCollection.Description = "Delete the active version of a collection by name"
```
AndrewSisley: todo: Please update this to note that it takes a csv set of names

internal/db/collection_define.go
```diff
@@ -378,7 +380,25 @@ existingVersionLoop:
...
 	}
 
-	return db.loadCollectionDefinitions(ctx)
+	if err := db.loadCollectionDefinitions(ctx); err != nil {
+		return wrapTypeNotFoundForRemoved(err, removedCollectionVersions)
+	}
+	return nil
+}
+
+// wrapTypeNotFoundForRemoved wraps a schema.ErrTypeNotFound emerging from a patch that
```
AndrewSisley: todo: This is a bug, if you do not wish to fix it in this PR, please open a new issue and link to it via a todo-comment so that anyone reading this knows this is tech-debt that we do not wish to keep.

suggestion: If you are sure this is actually happening here, I suggest just fixing the validation - it should take a similar amount of code.

AndrewSisley (approved):
Approving, with todos to be resolved before merge

cli/collection_delete.go
AndrewSisley: Fred spoke to me about this, named params cant have space deliminated values, ignore this question :)

cli/collection_delete.go
fredcarle: This should probably delete the entirety of the collection (all versions).

shahzadlone: > This should probably delete the entirety of the collection (all versions).

@fredcarle  I thought about this, how about a flag `--all` or `--allVersions` that does that? and by default it would delete only the latest version like patch (what we have now)?

fredcarle: I would do the opposite. Default deletes everything and we can have a flag to delete a singe version.

internal/db/store.go
shahzadlone: You mean to end up with something like this? 

```go
func (db *DB) DeleteCollection(
	ctx context.Context,
	names []string,
	opts ...options.Enumerable[options.DeleteCollectionOptions],
) error {
...
	seen := make(map[string]struct{}, len(names))
	ops := make([]string, 0, len(names))
	for _, name := range names {
		if _, dup := seen[name]; dup {
			continue
		}
		seen[name] = struct{}{}

		nameOps, err := db.deleteCollection(ctx, &options.GetCollectionsOptions{
			CollectionName: immutable.Some(name),
		})
		if err != nil {
			return err
		}
		ops = append(ops, nameOps...)
	}

	patch := "[" + strings.Join(ops, ",") + "]"
	if err := db.patchCollection(ctx, patch, immutable.None[model.Lens]()); err != nil {
		return err
	}

	return txn.Commit()
}

// deleteCollection returns the JSON-Patch ops needed to remove the collection
// version(s) matched by the given options.
func (db *DB) deleteCollection(
	ctx context.Context,
	opts *options.GetCollectionsOptions,
) ([]string, error) {
	cols, err := db.getCollections(ctx, opts, true)
	if err != nil {
		return nil, err
	}
	if len(cols) == 0 {
		return nil, client.ErrCollectionNotFound
	}
	ops := make([]string, 0, len(cols))
	for _, col := range cols {
		ops = append(ops, fmt.Sprintf(`{"op": "remove", "path": "/%s"}`, col.Version().VersionID))
	}
	return ops, nil
}
```

shahzadlone: I dislike this because a single getCollections(opts) call only encodes one filter (GetCollectionsOptions has a single CollectionName slot). To preserve atomicity across multiple names, the public layer aggregates and patches once. Letting the internal apply the patch per call would have failed TestDeleteCollection_ReferencedByRelation_* tests where two collections reference each other and need to be removed together.

cbindings/delete_collection.go
shahzadlone: I don't understand the todo, this line is splitting the names into a list.

http/handler_store.go
shahzadlone: - [x] Good catch - done

http/handler_store.go
shahzadlone: - [x] Changed to "Delete the active version of one or more collections. The 'name' query parameter accepts a comma-separated (CSV) list of collection names."

internal/db/collection_define.go
shahzadlone: The "re-add missing collections" loop unconditionally added back any existing collection that's  missing from newCollections, including ones being explicitly deleted. i think that loops job was to keep the previous version alive when a collection mutates into a new version? So by the time validateCollectionChanges runs, the deleted collection is back in newCollections (hence is active) and validateRelationPointsToValidKind cannot see the dangling reference. Then I think validation passes, as deleteCollectionVersions actually removes from storage, and the dangling reference only surfaces later when loadCollectionDefinitions hands the post-delete state to the GraphQL schema.

- [x] I added `validateRemovedCollectionsNotReferenced(removed, newColsByID)` that runs right after `removedCollectionVersions` is computed and before the re-add masks the deletion. It walks `newColsByID` (the post-patch state) and errors if any object field's Kind is a *NamedKind or *CollectionKind that points at a removed collection's name or CollectionID. LMK if I am missing anything or any push back

cli/collection_delete.go
shahzadlone: Implemented but want to flag something to make sure we are on the same page, I assumed that when using `--active-only` (original PR implementation) even after the deletion we do not auto-promote the previous version, so it leaves the collection "headless" (the only remaining head version still stays `IsActive: false`).

- [x] Done in commit: PR(MAIN): Add default ability to delete all versions

internal/db/store.go
AndrewSisley: > You mean to end up with something like this?

No, I mean something like:
```
type DeleteCollectionOptions = GetCollectionsOptions

func (db *DB) DeleteCollection(
	ctx context.Context,
	opts ...options.Enumerable[options.DeleteCollectionOptions],
) error {
    cols, err := db.getCollections(ctx, opts, true)
	if err != nil {
		return nil, err
	}
	ops := make([]string, 0, len(cols))
	for _, col := range cols {
		ops = append(ops, fmt.Sprintf(`{"op": "remove", "path": "/%s"}`, col.Version().VersionID))
	}
	return ops, nil
}
```

cbindings/delete_collection.go
AndrewSisley:
```
//export DeleteCollection
func DeleteCollection(
	nodePtr C.uintptr_t,
	names C.uintptr_t, // pointer to an array of strings
	activeOnly C.int,
	identityPtr C.uintptr_t,
) C.Result {
```

internal/db/collection_define.go
AndrewSisley: In my opinion, where you have added the call to `validateRemovedCollectionsNotReferenced` is wrong from a maintenance/architectural perspective - the whole point of `validateCollectionChanges` is to provide a single unified location through which we can expect all validation to exist, scattering other bits of validation throughout the code base defeats the point of it, and allows other code paths to accidentally skip validation that the assume will be run against all collection changes.

As well as that, if it is true that deleted collection versions are being passed into the validation code, then that *is* the bug that needs fixing - here you are just patching over one of the symptoms, and there may be others that we dont yet know about, but which our users may discover.

cbindings/delete_collection.go
shahzadlone: `C.uintptr_t` is just an integer that can hold a pointer right? So, then to consume it on the Go side wouldn't we then need a separate `nameCount` parameter (no length info in the pointer alone) and we would then need to cast uintptr -> unsafe.Pointer -> **C.char and walk it with unsafe.Slice or pointer arithmetic? On the wrapper side, allocate []*C.char, C.CString each name, defer-free each one, take &cNames[0], cast to uintptr.

Also every other multi-value cbinding in this codebase uses CSV style already i believe, like these `AddP2PReplicator`, `AddP2PCollection`, `AddP2PDocument`, `DeleteP2PReplicator`, `SyncP2PDocuments`, `SyncP2PCollectionVersions`.

example:
`func AddP2PCollection(nodePtr C.uintptr_t, collections *C.char, identityPtr C.uintptr_t) C.Result {
`

internal/db/store.go
shahzadlone: Oh you mean on `DeleteCollection` rather than `deleteCollection`.

Regardless, I am not 100% convinced that is better, or if it even works. Here are some reasons:

1) If we use `db.getCollections(ctx, opts, true)` as you suggested, `db.getCollections` takes options for 1 collection at a time (we want to be able to delete many in an atomic way).

While you can recover atomicity by redefining what variadic options mean for this op (but will be different from your suggestion), we can treat each enumerable is a separate filter, accumulate ops across all of them, and then patch once , we would have to do something like this instead:
```go
ops := make([]string, 0, len(cols))
for _, opt := range opts {
    cols, _ := db.getCollections(ctx, opt, true)
    if err != nil {
      return nil, err
    }
    for _, col := range cols {
      ops = append(ops, fmt.Sprintf(`{"op": "remove", "path": "/%s"}`, col.Version().VersionID))
    }
}
db.patchCollection(...)
```
But it's not better nor is it consistent with how other operations use `GetCollectionsOptions`
 
 2) I also dislike how the default option we just introduced (delete all versions) is opposite of `GetInactive` defaulting to `false`, also not the mention each collection option will have it's own `GetInactive` field rather than 1 for all the collections. Note: `options.DeleteCollection().SetCollectionName("Users")` would only delete the active head. We'd either regress the default or introduce a "delete uses inverted semantics for GetInactive" rule that's confusing.
 
 3) Empty options would means delete-everything. for example, if `db.getCollections(ctx, &options.GetCollectionsOptions{}, true)` returns every collection in the DB. So DeleteCollection(ctx) with no opts could nuke the whole schema (i know that right now we have a guard like this: `len(names) == 0 -> ErrCollectionNameRequired`, but I still worry).

internal/db/collection_define.go
shahzadlone: Thanks for flagging! Got some brain juices flowing and was wondering what you think if i do this (kind of a 2 parter):

1) Stop the re-add when there's no replacement. A collection in `removedCollectionVersions` with no successor in `newCollections` (no `CollectionID` match in the new set) is a real deletion, skip the re-add. The loop only re-adds for the mutation case (collection being replaced by a new version with the same `CollectionID`).

2) Let `validateRelationPointsToValidKind` produce the friendly error. Once the deleted collection is genuinely missing from newState, the existing validator naturally fires `NewErrFieldKindNotFound` for `Books.author -> Users`. Can improve it to look up the kind in oldState too, and when it's present in old but missing in new (it means it is removed), return `NewErrRemoveReferencedCollectionFromField` instead. Otherwise return the existing `NewErrFieldKindNotFound`.

Can then remove `validateRemovedCollectionsNotReferenced` function completely.

cbindings/delete_collection.go
AndrewSisley: Ah my bad, I thought C arrays held the length as the first value for some reason...

If we do this everywhere else, happy as-is :)

internal/db/collection_define.go
AndrewSisley: yeah, that makes sense to me - probably just need to tweak the `isMissing` block to fix the bug (plus the cleanup you mentioned) - it looks like it wasn't written with hard-deletes in mind.

internal/db/store.go
AndrewSisley: I think you might have misunderstood me a bit, but this `Empty options would means delete-everything` is a good catch - we don't want to do that :)  And I don't want the behaviour to diverge from `getCollections` if we use its options - happy leaving as-is

internal/db/collection_define.go
shahzadlone: - [x] Done in commit: `PR(FIX): Resolve the re-add bug`

AndrewSisley: Looks good, thanks Shahzad

PR #4692, opened 2026-04-07
fix: Remove 'concurrent' transaction concept

### LINKED ISSUE #4690: Remove `NewConcurrentTxn`
All transaction types should not be able to corrupt the database, no matter how and by who they are used.  Remove `NewTxn` and rename `NewConcurrentTxn` to `NewTxn`.  

`NewConcurrentTxn` doesn't completely protect against corrupting the database, but it is better than nothing: https://github.com/sourcenetwork/defradb/issues/4626

AndrewSisley:
## Relevant issue(s)

Resolves #4690

## Description

**Semver breaking change.**

All transactions are now concurrent, at least once https://github.com/sourcenetwork/defradb/issues/4626 is resolved.  The current implementation at least guards against concurrent writes to the same store keys.

`ConcurrentTxn` is now the only transaction type that users will be using.  Non-concurrent transactions are still used by the time-travel code (see `Remove concurrent txn` commit message for more info).

### DIFF
--- a/cbindings/txn.go
+++ b/cbindings/txn.go
@@ -24,17 +24,11 @@ import (
 )
 
 //export CreateTransaction
-func CreateTransaction(nodePtr C.uintptr_t, isConcurrent C.int, isReadOnly C.int) C.NewTxnResult {
+func CreateTransaction(nodePtr C.uintptr_t, isReadOnly C.int) C.NewTxnResult {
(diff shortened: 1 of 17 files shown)


fredcarle (approved):
LGTM. Thanks for simplifying the API.

Issue #4722, opened 2026-04-15
Improve error message when removing a field that has an index


islamaliev:
Right now if you try to remove a field via PatchCollection and that field has a secondary index on it, you get a generic "the given field does not exist" error which is confusing and misleading — the field clearly exists, you just can't remove it because there's an index depending on it.

Should say something like "cannot remove field 'name': field has a dependent index, drop the index first" or similar. Postgres does something like this too — it tells you about dependent objects and suggests CASCADE.

Found this while testing PatchCollection with the secondary-index multiplier enabled.

This also affects several collection version tests that remove or replace fields:
- `tests/integration/collection_version/updates/replace/field/simple_test.go`
- `tests/integration/collection_version/updates/remove/fields/simple_test.go` (2 tests)
- `tests/integration/collection_version/updates/remove/col_source_transform_test.go`
- `tests/integration/txn/patch_collection_test.go` (2 tests)

All excluded from the secondary-index multiplier for the same reason.

PR #4723, opened 2026-04-16
chore: Collection version templates in tests

islamaliev:
## Relevant issue(s)

Resolves part of #4353
Related: #4722

## Description

Adds template replacement support for collection version IDs in `ConfigureMigration`, so tests can use `{{.CollectionVersionID0}}` instead of hardcoded CID hashes. Writing multi-version migration tests without this means hand-computing CIDs for every step, which is tedious and makes those tests harder to maintain. Existing hardcoded tests keep working.

Also cleaned up the `MultiplierExcludes` comments across several test files:
- Tests that fail because of indexed-field removal now reference #4722 with a clear explanation instead of a bare TODO
- Tests that fail because lens migrations return nil with indexes present still reference #4353 with a descriptive comment

## Tasks

## How has this been tested?

Ran the migration and collection version test suites both with and without `DEFRA_MULTIPLIERS=secondary-index`.

- MacOS


AndrewSisley:
>  (e.g. when the secondary-index multiplier adds @index to all fields

This should have no impact on the collection version id at all, are you sure you have seen this?  If so, can you please provide a test/example of how this happens?

AndrewSisley (approved):
LGTM, and it is a good change, but I think it should not have been relevant to secondary indexes, and if they were causing the collection id to change, then there might be another bug there.

islamaliev:
You're right — I went back and tested this. Reverting one of the templated tests to its hardcoded hashes and running it under `DEFRA_MULTIPLIERS=secondary-index` passes fine. The `@index` directive doesn't feed into the collection version ID at all, so the premise I put in the description is wrong.

The actual reason I pulled in templating was that some of the new multi-version tests I was adding had several versions in flight and computing the hashes by hand was painful — not because existing hashes broke. I confused "I didn't want to hardcode them" with "they broke".

So the templating is still useful (easier to write multi-version migration tests without hand-hashing CIDs), but the justification in the PR description is misleading. I'll update it.

AndrewSisley:
Glad to hear, thanks for looking into it Islam :)

>  computing the hashes by hand was painful 

Just in case you've not yet tried and end up finding it easier, when I need them, I write the test with junk ids and then pull the real ids out from the logs

PR #4726, opened 2026-04-17
fix(i): Route CollectionSetDelta out of GetDelta

### LINKED ISSUE #4720: CRDT: `CRDT.GetDelta()` silently drops `CollectionSetDelta` due to duplicate case in switch
## Summary

`internal/core/crdt/ipld_union.go:62-80` contains a copy-paste bug in `CRDT.GetDelta()`: the switch has two identical `case c.CollectionDelta != nil` arms. The second is unreachable. Whenever a `CollectionSetDelta` is present, `GetDelta()` falls through every case and returns `nil` at line 79, silently dropping the delta. Any downstream code that applies `GetDelta()` on a CRDT block containing a CollectionSetDelta will observe a nil pointer and skip or mis-handle the merge.

Found during a parity review of the Rust reimplementation (`defradb.rs`), which uses exhaustive Rust pattern matching and correctly dispatches to `CollectionSetDelta`. The divergence surfaced the latent Go bug.

## Details

Current code (`internal/core/crdt/ipld_union.go:61-80`):

```go
// GetDelta returns the delta that is stored in the CRDT.
func (c CRDT) GetDelta() Delta {
    switch {
    case c.LWWDelta != nil:
        return c.LWWDelta
    case c.DocCompositeDelta != nil:
        return c.DocCompositeDelta
    case c.CounterDelta != nil:
        return c.CounterDelta
    case c.CollectionDelta != nil:           // line 70
        return c.CollectionDelta
    case c.CollectionDelta != nil:           // line 72 — DUPLICATE, should be CollectionSetDelta
        return c.CollectionSetDelta
    case c.CollectionDefinitionDelta != nil:
        return c.CollectionDefinitionDelta
    case c.FieldDefinitionDelta != nil:
        return c.FieldDefinitionDelta
    }
    return nil
}
```

The IPLD schema explicitly includes `CollectionSetDelta` (`ipld_union.go:54`):

```go
type CRDT union {
    | LWWDelta "lww"
    | DocCompositeDelta "composite"
    | CounterDelta "counter"
    | CollectionDelta "collection"
    | CollectionSetDelta "collectionSet"
    | CollectionDefinitionDelta "collectionDefinition"
    | FieldDefinitionDelta "fieldDefinition"
} representation keyed
```

And the sibling methods handle `CollectionSetDelta` correctly — e.g. `GetPriority()` at `ipld_union.go:93-94`:

```go
case c.CollectionSetDelta != nil:
    return c.CollectionSetDelta.GetPriority()
```

So the bug is isolated to `GetDelta()`.

## Impact

Any caller of `CRDT.GetDelta()` on a block that carries a `CollectionSetDelta` will receive `nil`. Based on a grep of the Go tree, `GetDelta()` is the canonical way to recover a `Delta` from a deserialized `CRDT` block. Silent nil results risk:

- Skipped merges for CollectionSetDelta blocks.
- Nil-pointer dereferences in callers that assume a non-nil result.
- CRDT convergence divergence between nodes that use `GetDelta()` and nodes that access the struct fields directly.

Rust's equivalent uses exhaustive pattern matching on an enum, so the bug does not manifest there — but the divergence itself means a Go replica and a Rust replica observing the same `CollectionSetDelta` block reach different merged states.

## Expected Behavior

`GetDelta()` returns `c.CollectionSetDelta` when that field is non-nil. The second case's condition is changed from `c.CollectionDelta != nil` to `c.CollectionSetDelta != nil`.

## Fix

Two-character change at `ipld_union.go:72`:

```diff
-    case c.CollectionDelta != nil:
+    case c.CollectionSetDelta != nil:
         return c.CollectionSetDelta
```

## Acceptance Criteria

- [ ] `CRDT.GetDelta()` returns `c.CollectionSetDelta` when that field is non-nil
- [ ] Unit test added: construct a `CRDT{CollectionSetDelta: ...}` and assert `GetDelta()` returns the `*CollectionSetDelta`, not nil
- [ ] Unit tests for the other six variants continue to pass (regression guard — ensure the fix doesn't shift any ordering)
- [ ] Any regression tests for collection-set replication flows are added or updated

## Files

- `internal/core/crdt/ipld_union.go:72` — one-line fix
- `internal/core/crdt/ipld_union_test.go` (or new test file) — regression test for each variant's `GetDelta()`

## References

- Found during parity review of `defradb.rs`. The Rust implementation in `crates/defra-core/src/block_delta.rs` uses a Rust enum with exhaustive matching — the Go issue was detected because Rust's behaviour diverged from what the Go code appeared to promise.
- Related `go-vet` or `staticcheck` may have flagged this historically under rule `SA4009` (duplicate case) or `SA4006`. Consider enabling the lint in CI to catch future duplicates.

SAY-5:
### Problem

`internal/core/crdt/ipld_union.go` declares a keyed IPLD union whose
arms are `LWWDelta`, `DocCompositeDelta`, `CounterDelta`, `CollectionDelta`,
**`CollectionSetDelta`**, `CollectionDefinitionDelta`, `FieldDefinitionDelta`.

`CRDT.GetDelta()` dispatches on whichever field is non-nil. Lines 70 and 72
both check the same field:

```go
case c.CollectionDelta != nil:
    return c.CollectionDelta
case c.CollectionDelta != nil:        // copy-paste: should be CollectionSetDelta
    return c.CollectionSetDelta
```

The second arm is unreachable, so a `CRDT` carrying only `CollectionSetDelta`
falls through every case and returns `nil`. Any downstream code that calls
`GetDelta()` on such a block observes a nil `Delta` and silently skips or
mis-handles the merge.

The sibling `GetPriority()` at lines 93-94 already dispatches on
`c.CollectionSetDelta != nil`, so the IPLD schema and the priority path
are consistent, only `GetDelta` is broken.

### Fix

Check `CollectionSetDelta` in the second arm, matching the schema and
`GetPriority`.

```diff
-    case c.CollectionDelta != nil:
+    case c.CollectionSetDelta != nil:
         return c.CollectionSetDelta
```

Fixes #4720


AndrewSisley (approved):
LGTM, thanks for cleaning up my mess @SAY-5 :)

PR #4727, opened 2026-04-18
fix(i): Clear Signature before marshalling signing bytes

### LINKED ISSUE #4719: P2P: `signAndSetMetaData` does not clear `Signature` before marshaling for signing
## Summary

`internal/db/p2p/message/message.go:272-301` (`signAndSetMetaData`) relies on the zero-value nil of `Message.GetSignature()` to produce correct signing bytes. It does not explicitly call `m.SetSignature(nil)` before `cbor.Marshal(m)`. Today this is harmless because messages are constructed fresh on every send and `Signature` is always nil at sign time. However, any future code path that reuses a message struct — either by re-signing an already-signed message or by populating fields defensively before the send path — will silently fold a stale `Signature` into the signed bytes and produce a signature that will never verify on the receiver.

`verifyMessage` already does the defensive clear (`message.go:255-256`). `signAndSetMetaData` should do the same for symmetry.

Caught during a parity review of the Rust reimplementation (`defradb.rs`); the Rust signing path in `crates/p2p/src/signing.rs:72-94` does `msg.set_signature(None)` explicitly before marshaling. Go does not.

## Details

Current implementation (`internal/db/p2p/message/message.go:272-301`):

```go
func signAndSetMetaData(h client.Host, m Message) error {
    if m.GetMessageID() == \"\" {
        messageID, err := uuid.NewV4()
        if err != nil {
            return err
        }
        m.SetMessageID(messageID.String())
    }

    nodePubKey, err := h.Pubkey()
    if err != nil {
        return err
    }
    m.SetVersion()
    m.SetPubkey(nodePubKey)
    m.SetSenderID(h.ID())

    forSigning, err := cbor.Marshal(m)   // <-- Signature field is whatever was on the struct
    if err != nil {
        return err
    }

    signature, err := h.Sign(forSigning)
    if err != nil {
        return err
    }
    m.SetSignature(signature)
    return nil
}
```

The `MetaData.Signature` field has `cbor:\",omitempty\"` (`internal/db/p2p/message/metadata.go:19`), so a nil slice is correctly omitted from the CBOR map. But any non-nil `Signature` — even an empty but non-nil `[]byte{}`, or a stale signature from a prior sign — will be included in the signing bytes, changing what gets hashed. On verify, the receiver clears the signature and re-marshals, so the bytes will disagree and verification fails.

### When this can bite

- Retry logic that clones-and-re-signs an outgoing message (none today, but plausible for future retry/redelivery work).
- Test fixtures that construct a `PushLogRequest` literal with `Signature: []byte{}` for readability.
- Future code that routes a message through a middleware chain and optionally re-signs.

Rust reference — does the defensive clear:

```rust
// crates/p2p/src/signing.rs:72-94
msg.set_signature(None);                      // <-- explicit clear
let bytes = serde_cbor::to_vec(&msg)...?;
let signature = keypair.sign(&bytes)?;
msg.set_signature(Some(signature));
```

`verifyMessage` in this same file already follows the defensive pattern (`message.go:254-256`):

```go
signature := m.GetSignature()
m.SetSignature(nil)
b, err := cbor.Marshal(m)
```

This issue is about making the sign path symmetric with the verify path.

## Expected Behavior

`signAndSetMetaData` explicitly sets `Signature` to nil before `cbor.Marshal`, regardless of what the caller did or didn't do with the struct. The behavior is the same in the happy path (fresh message); the change eliminates the latent bug class for any future reuse.

## Fix

Add one line before the marshal:

```go
m.SetVersion()
m.SetPubkey(nodePubKey)
m.SetSenderID(h.ID())
m.SetSignature(nil)                           // <-- add this

forSigning, err := cbor.Marshal(m)
```

## Acceptance Criteria

- [ ] `m.SetSignature(nil)` called immediately before `cbor.Marshal(m)` in `signAndSetMetaData`
- [ ] Unit test: a pre-populated `Signature` field on the input message does not affect the final signature. The test should construct two messages with identical payload — one with `Signature: nil`, one with `Signature: []byte{0xDE, 0xAD}` — run both through `signAndSetMetaData`, and assert that the resulting `Signature` bytes are identical
- [ ] Regression test: a fresh message with no `Signature` field set continues to sign correctly and verify
- [ ] No changes to wire format or observable behavior under current usage

## Files

- `internal/db/p2p/message/message.go:272-301` — add the explicit clear
- `internal/db/p2p/message/message_test.go` — add regression test for the stale-signature case

## References

- Found during parity review of `defradb.rs` (Rust reimplementation). Rust's equivalent explicitly clears: `crates/p2p/src/signing.rs:72-94` in the `defradb.rs` repo.
- Related defensive pattern already in use: `verifyMessage` at `internal/db/p2p/message/message.go:254-256`.
- This is a defensive-coding hardening, not an exploitable vulnerability in the current code — hence `priority/low`. Filing so the improvement isn't lost.

SAY-5:
### Problem

`signAndSetMetaData` relies on the zero-value `nil` of
`Message.GetSignature()` to produce correct signing bytes. It never
explicitly clears the signature before `cbor.Marshal(m)`:

```go
m.SetVersion()
m.SetPubkey(nodePubKey)
m.SetSenderID(h.ID())

forSigning, err := cbor.Marshal(m)   // signature is whatever was on the struct
...
```

Today every call site (`internal/db/p2p/message/message.go:122/140/186`)
constructs a fresh `Message` where `Signature` is already nil, so this
is harmless. Any future code path that reuses a `Message` struct across
sends, or populates fields defensively before the send path, would fold
a stale `Signature` into the signing bytes and produce a signature that
never verifies on the receiver.

`verifyMessage` already does the defensive clear
(`m.SetSignature(nil)` at line 256), and the Rust reimplementation
(`defradb.rs`) does the same before signing. #4719.

### Fix

Mirror the existing pattern in `verifyMessage`: clear the signature
before marshaling. No behaviour change for the current fresh-message
call sites; future reuse stays correct.

```diff
  m.SetVersion()
  m.SetPubkey(nodePubKey)
  m.SetSenderID(h.ID())
+ // Explicitly clear the signature before marshaling the signing bytes.
+ m.SetSignature(nil)

  forSigning, err := cbor.Marshal(m)
```

Fixes #4719


AndrewSisley (approved):
LGTM

PR #4729, opened 2026-04-18
fix: Clarify keyring CLI help text

### LINKED ISSUE #4704: CLI command `keyring new` docs inaccurate
The `keyring new` docs look inaccurate/incomplete in some places. 

> The DEFRA_KEYRING_SECRET environment variable must be set to unlock the keyring.

This should probably say "... to unlock an existing keyring", because the command can be run without that env var being set and it works fine (and till set the env var itself). 

We should also bring up that by default, the keyring will be initialized with `secret` as a password.

```shell
# ensure env var is not set
root@0e5df81defca:/# env -u DEFRA_KEYRING_SECRET
HOSTNAME=0e5df81defca
PWD=/
HOME=/root
TERM=xterm
SHLVL=1
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env

# generate keys
root@0e5df81defca:/# ./defradb keyring new --force
Apr 11 03:42:38.505 INF cli generated new encryption key
Apr 11 03:42:38.518 INF cli generated new peer key
```

> WARNING: This will overwrite existing keys in the keyring.

It won't, unless `--force` is used:

```shell
root@0e5df81defca:/# ./defradb keyring new       
Error: key encryption-key already exists, use --force to overwrite
root@0e5df81defca:/# ./defradb keyring new --force
Apr 11 03:41:48.799 INF cli generated new encryption key
Apr 11 03:41:48.810 INF cli generated new peer key
```

> --keyring-path string         Path to store encrypted keys when using the file backend (default "keys")

The default should be `~/.defradb/keys/` - or perhaps, at a closer look, it should say "Path (relative to the `.defradb` directory) to store...`. I don't know if we call that dir something specific.

SAY-5:
## What

Help-text fixes for the keyring CLI, per #4704:

1. `keyring new --help` warned that the command "will overwrite existing keys in the keyring". It won't unless `--force` is passed; a second run without `--force` returns `key ... already exists, use --force to overwrite`.
2. `--keyring-path` help didn't mention that the default `keys` is resolved against `--rootdir` by `LoadConfig`, so users on a first run guess wrong about where `keys/` lands.

No behaviour change, help strings only. The CLI reference docs are regenerated from the same metadata.

An earlier revision also reworded the `DEFRA_KEYRING_SECRET` paragraph based on the report in #4704; that was inaccurate (`openKeyring` errors when no secret is set) and has been reverted.


SAY-5:
Rebuilt the branch on top of current `develop` and force-pushed `d32968a`. The original help-text change is now replayed cleanly on the latest base, and the CLI reference docs were regenerated again on top of that so the docs check has current outputs.

PR #4730, opened 2026-04-18
fix: Return ExecRequest errors correctly over http

### LINKED ISSUE #4266: Wrong GQL response error formatting
There are some error paths in the http handler for `ExecRequest` that don't correctly marshal to a GraphQL error message format.

Anytime we call `responseJSON(rw, http.StatusBadRequest, errorResponse{err})` will result in the wrong error format that GQL clients are expecting.

Currently it returns:
```json
{
  "error": "some error message"
}
```

But it needs to return
```json
{
  "errors": [
    { "message": "some error message" }
  ]
}
```

This is causing the http client to incorrectly return errors.

SAY-5:
## What

`execHTTPRequest` and `execSSESubscription` in `http/handler_store.go` reused the generic `errorResponse{err}` helper on their pre-resolver error paths (bad JSON body, missing query field, subscription/transport mismatch, streaming unsupported). That helper marshals to:

```json
{"error": "some error message"}
```

But every GraphQL client in the wild, apollo, urql, graphql-request, `graphql.tada`, the defradb HTTP client's own `Run` method, only inspects the top-level `errors` array per the [GraphQL-over-HTTP spec](https://graphql.github.io/graphql-over-http/draft/):

```json
{"errors": [{"message": "some error message"}]}
```

So when the server rejects a request before the resolver runs, the client sees an empty `errors` slice on the decoded body and falls through to "unknown transport error" handling. Users get no indication of *why* their request failed. Per #4266.

## Fix

Add `gqlErrorResponse` alongside `errorResponse` in `http/errors.go`, marshalling to the standard GraphQL envelope:

```go
type gqlErrorResponse struct {
    Error error `json:"-"`
}

func (e gqlErrorResponse) MarshalJSON() ([]byte, error) {
    return json.Marshal(map[string]any{
        "errors": []map[string]any{{"message": e.Error.Error()}},
    })
}
```

Swap the four call sites on the ExecRequest paths:

- `execHTTPRequest`: bad request body (`extractGraphQLRequest` error) + subscription-on-HTTP-transport error.
- `execSSESubscription`: bad request body + streaming-unsupported error.

All four now ship `{"errors":[{"message":"..."}]}`, which `result.GQL` (the success path) already matches. The response shape is now consistent across both pre-resolver and post-resolver errors on the ExecRequest endpoint.

## Out of scope

Other handlers (`BasicImport` / `BasicExport` / `CCIP` / ACP / txn commit/discard / p2p / block / document) keep their existing `{"error":"..."}` shape, their consumers are REST-style and do not expect GraphQL framing. Only the two GraphQL-transport handlers are swapped here.

`errorResponse` itself is retained (including its `UnmarshalJSON` used by the HTTP client for non-GQL routes).

Fixes #4266


AndrewSisley (approved):
LGTM, thanks @SAY-5

Issue #4736, opened 2026-04-20
Lens doc-version stamp does not invalidate when migration chain changes


islamaliev:
Stumbled onto this while working on #4353. That fix landed the common "register in chain order" case, but the underlying cache-invalidation behavior is broken in a more general way that shows up under any secondary index.

The doc-version stamp that `updateDataStore` writes after a lens migration serves two purposes at once: it tells the fetcher whether to skip the pipeline next time (cache key), and it records the position the doc's stored field values correspond to. Those two roles don't line up cleanly. A `ConfigureMigration` that registers a new transform on a previously-nil edge does not change the stamp semantics at all — so any doc already stamped at the active version won't re-migrate, even though the chain definition is now different.

Concretely: with no index, nothing triggers a cache fill until the user actually queries, so most orders work out fine. With an index, every `ConfigureMigration` kicks off a reindex, docs get pulled through the current (partial) pipeline and stamped at the active version. The next migration registration never re-enters them.

This shows up in several shapes:

- **Incremental in-order chain**: register v0→v1, then v1→v2. First reindex stamps the doc at v2 even though email wasn't yet set. Second registration skips it. Field stays nil. This is what #4353 was about.
- **Out-of-order registration** (original content of this issue): register v0→v1 and v2→v3 first, close with v1→v2 last. First two reindexes stamp past the v1→v2 gap. Final registration doesn't re-enter. Field stays nil.
- **Restart between incremental registrations**: same as the first case, but with a restart in between — the stale stamp is persisted to disk and survives.

Andy's suggested direction (via DM) is to derive the stamp from something that actually changes when the chain definition changes, e.g. a hash composed from the lens CIDs along the path plus the schema version id. Since lenses now have CIDs, this is tractable. Any `ConfigureMigration` updates the hash for anything downstream, and the fetcher's `doc.CollectionVersionID() == targetVersionID` check naturally becomes a chain-equality check rather than a position check. Nil edges hash the same way any other edge does, so replacing a nil with a real transform is the same event as replacing transform A with transform B — no special-case logic needed.

The partial-chain-stamp workaround attempted in #4737 was rolled back after review. It patched `updateDataStore` to stamp at "the last version where a real transform ran," which handles the in-order case but conflates "nil because pending" with "nil because schema-only patch" and can't be extended to the out-of-order case without tracking that distinction in the data model. The hash-based approach sidesteps that entirely.

[...]

PR #4737, opened 2026-04-20
test: Document incremental lens migration bug under indexes

islamaliev:
Was originally going to land a fetcher-side workaround (stamp the doc at the last version where a real transform ran) for the chained-migration bug seen under the secondary-index multiplier. Andy pushed back: that model conflates "nil because pending migration" with "nil because schema-only patch", and the proper fix is at a different layer entirely — make the doc-version stamp derive from something that actually changes when the migration chain definition changes (e.g. a hash of the lens CIDs along the path). Reverted the fetcher change.

What's left: a test file documenting the bug surface (in-order chain, three-migration chain, out-of-order intermediate, stamp-survives-restart). Cases that require the proper fix are excluded under the `secondary-index` multiplier with a pointer to #4736 — the existing issue, which I rewrote to cover the general "stamp doesn't invalidate on chain change" problem rather than just the out-of-order variant.


internal/lens/fetcher.go
```diff
@@ -293,12 +298,100 @@ func (f *lensedFetcher) lensDocToEncodedDoc(docAsMap LensDoc) (fetcher.EncodedDo
 	}, nil
 }
 
+// effectiveStampVersion returns the version to stamp a doc with after migration.
+//
+// When migrations are incomplete (some links have no transform yet), the lens pipeline
+// passes docs through no-op links without actually transforming them. Stamping such a
+// doc with the target version would falsely mark it as fully migrated — future fetches
+// would skip the migration even after the missing transforms are registered.
```
AndrewSisley: todo: This sounds like behaviour that I do not think we want, and it does not look like this has been tested.

It might be that this documentation needs tweaking, and it is suggesting that it is doing something that it is not.

If a migration has not been set between two versions at the time of query, that, nil, is the correct migration, and the document should be tagged with the target version - the entire transform chain has been evaluated, as it was defined at the time of query.

The presence of `nil` migration should be treated the same as any other migration.  If it is queried/evaluated and the doc cached, and then the `nil` is updated to another migration, it should behave exactly the same as if they had set a non-nil migration, queried, and then updated the migration to equal something else.

AndrewSisley (changes requested):
I do not agree with this change, unless I am misunderstanding it.

I suggest discussion with the team if you disagree with my reasons and still want this change.

internal/lens/fetcher.go
islamaliev: Agreed after our DM exchange — the stamp-at-last-real-transform logic is patching the wrong layer. Reverted the fetcher changes in this PR. Kept the new tests as documentation of the bug surface; the cases that need the proper fix are excluded under the secondary-index multiplier with a pointer to #4736 (rewrote that issue to cover the general stamp-invalidation-on-chain-change problem). The hash-based stamp direction you suggested is noted there as the proposed fix.

fredcarle (approved):
LGTM

AndrewSisley (approved):
LGTM

shahzadlone (approved):
LGTM

PR #4740, opened 2026-04-21
fix: Exit CLI on API server errors

### LINKED ISSUE #4735: Node startup becomes unresponsive and unkillable if TLS error occurs
If a path to a key is not found during TLS setup, the command errors and subsequently hangs. CTRL-C doesn't yield anything. The only way to exit is `kill -s KILL <pid>`.
```
stef@carbonino:~/git/defradb$ DEFRA_KEYRING_SECRET=secret defradb start --pubkeypath ~/.defradb/certs/pubkey.crt --privkeypath ~/.defradb/certs/privkey.key
Apr 20 16:31:48.185 INF cli Starting DefraDB
Apr 20 16:31:48.206 INF p2p Created LibP2P host PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Address=[/ip4/127.0.0.1/tcp/9171]
Apr 20 16:31:48.207 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=doc-sync
Apr 20 16:31:48.207 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=sync-branchable
Apr 20 16:31:48.207 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=encryption
Apr 20 16:31:48.207 INF db Starting without nac (not configured/setup)
Apr 20 16:31:48.208 INF node Providing HTTP API at http://127.0.0.1:9181 PlaygroundEnabled=false
Apr 20 16:31:48.208 INF node Providing GraphQL endpoint at http://127.0.0.1:9181/api/graphql
Apr 20 16:31:48.208 ERR node HTTP server stopped $err="open /home/stef/.defradb/certs/pubkey.crt: no such file or directory"
^C^C^C
```

On Linux Mint, nightly build compiled today.

SAY-5:
## Summary

Fixes #4735.

When `defradb start` is given a bad `--pubkeypath` / `--privkeypath`, the API server goroutine inside `startAPI` exits immediately with a "no such file or directory" error. Today that error is only logged - the CLI's main `select` waits on `signalCh`, `purgeSub.Message()`, and `cmd.Context().Done()`, none of which fire, so the process hangs half-alive and Ctrl-C is ineffective:

```
INF node Providing HTTP API at http://127.0.0.1:9181 ...
ERR node HTTP server stopped $err="open /home/stef/.defradb/certs/pubkey.crt: no such file or directory"
^C^C^C              <-- hang; only kill -9 recovers
```

The surviving P2P + DB + event sub-systems keep the foreground loop running until the user `SIGKILL`s the pid.

## Fix

1. Give `Node` a buffered `apiErrCh chan error` (capacity 1) and a public `APIError()` accessor.
2. The Serve-loop goroutine in `startAPI` now non-blockingly sends the error onto `apiErrCh` on any non-`ErrServerClosed` exit.
3. The CLI `SELECT:` loop in `cli/start.go` adds a case for `<-n.APIError()` that logs the error and falls through into the existing `return n.Close(cmd.Context())` path - the same cleanup that SIGINT already triggers.

## Test plan

- [x] `go build ./...`
- [x] Reproduced #4735 locally against develop; with this patch, passing a bogus `--pubkeypath` now prints `ERR ... API server exited unexpectedly; shutting down` and exits within milliseconds instead of hanging
- [x] `go test ./node/` surfaces a pre-existing `bytedance/sonic` build failure on Go 1.26 unrelated to this change, called out separately


SAY-5:
Thanks @coderabbitai — you were right, the original doc was wrong. The channel is never closed; it only delivers at most one error, then stays open. Updated the doc comment to:

- State explicitly that it is buffered and never-closed.
- Note that it fires at most once over the Node's lifetime.
- Show the non-blocking select pattern callers should use, so nobody tries the `err, ok := <-ch` idiom and mistakes a healthy Node for a failed startup.

Verified locally: `gofmt -s` clean on the touched file. I couldn't run `go test ./node/...` end-to-end on my machine because the project pins go 1.25.8 while my local toolchain is 1.26.2, and bytedance/sonic v1.14.2 uses a runtime symbol (`GoMapIterator`) that was removed in 1.26; that is orthogonal to this change (doc-comment-only diff) and CI on the project's go 1.25.8 image will exercise the full suite.

SAY-5:
Good call — applied the suggestion. The APIError case now wraps the error (and any Close error) so the command returns non-zero, which is what systemd / any process supervisor needs to see to trigger a restart.

SAY-5:
Addressed the CodeRabbit review (https://github.com/sourcenetwork/defradb/pull/4740#pullrequestreview-4145797944):

`defradb start` now returns the API server error as the command result so the process exits non-zero on unexpected server death. `n.Close` is still invoked for cleanup, and if both fail, both errors are wrapped so neither is lost. This matches how SIGINT / context cancellation already behave and restores the systemd/supervisor restart signal that #4735 broke.

Used `errors.Wrap` from `github.com/sourcenetwork/defradb/errors` to stay consistent with the rest of the CLI package.

SAY-5:
Good call — switched to `errors.Join(apiErr, closeErr)` so both errors survive unwrap (`errors.As` / `errors.Is` can reach either). defradb's own `errors` package already exposes `Join`, so no stdlib import was needed. Force-pushed.

AndrewSisley (approved):
LGTM, thanks @SAY-5

Issue #4745, opened 2026-04-23
Apply template placeholders to Request.Results values


islamaliev:
### Problem

`action.Request.Request` strings are run through `replace` (see
`tests/integration/replace.go:72`), so a test can write
`cid: "{{.CID0_0_0}}"` in the query and have the placeholder resolved at
run time. However, `action.Request.Results` values are NOT run through
`replace`. As a result, a test cannot assert
`"cid": "{{.CID0_0_0}}"` in the expected results map — the literal
template string is compared against the actual CID, which fails.

Today the workaround is to use a loose `testUtils.NewUniqueValue()` /
`testUtils.ValidCID()` matcher, which asserts only "some valid CID was
returned" and loses the real intent ("the returned CID equals the one we
queried for").

This comes up when migrating commit tests away from hardcoded CIDs for
the `signed-docs` multiplier rollout (#4453).

### Proposal

Apply `replaceMap` over string values inside
`action.Request.Results` (recursively for nested maps / slices) before
comparing against actual results, mirroring what already happens for
`Request`.

A sketch:

- In `assertRequestResults` (or `areResultsEqual`), walk the expected
  `map[string]any` and, when a value is a `string` containing `{{`,
  resolve it via `replace(s, nodeID, v)` before comparison.
- Apply it to list elements and nested maps as well.

### Benefit

Tests can then write:

```go
&action.Request{
    Request: `query { _commits(cid: "{{.CID0_0_0}}") { cid } }`,
    Results: map[string]any{
        "_commits": []map[string]any{
            {"cid": "{{.CID0_0_0}}"},
        },
    },
}
```

and get a strict literal-equality assertion without hardcoding the CID.
This would unblock strict assertions in several commit tests currently
using `uniqueCid` / `ValidCID` as a workaround.

Issue #4744, opened 2026-04-23
Template placeholder for collection-level commit CIDs


islamaliev:
### Problem

Tests that query the collection-level commit on `@branchable` collections
(e.g. `tests/integration/query/commits/branchables/cid_test.go`) have to
hardcode the exact CID string in both the `_commits(cid: ...)` argument and
the result assertion. This over-specifies the test and breaks the moment
anything changes block bytes (e.g. enabling the `signed-docs` test
multiplier — see #4453).

Document composite CIDs are already resolvable via the `{{.CIDN_M_K}}`
template placeholder (see `tests/integration/replace.go:27` +
`state.NodeState.Composites`). No equivalent exists for collection-level
commit CIDs.

### Proposal

Track collection-level commit CIDs per node and expose them via a new
template placeholder, e.g. `{{.CollectionCommitCID{col}_{idx}}}`.

Sketch:
- Add `CollectionCommits map[int][]cid.Cid` + lock to `state.NodeState`
  (`tests/state/state.go`), indexed by collection index.
- Populate it in the same place `Composites` is populated, filtered to
  branchable collection commits.
- Register a new generator in `templateDataGenerators`
  (`tests/integration/replace.go`).

Once available, the branchable commit tests can drop their hardcoded CIDs
and their `MultiplierExcludes: []string{multiplier.SignedDocs}` opt-outs.

### Affected tests

All tests under `tests/integration/query/commits/branchables/` that
hardcode collection commit CIDs.

Issue #4742, opened 2026-04-22
Replicator double-applies initial counter delta when signing is enabled


islamaliev:
Found this while wiring up the signed-docs test multiplier (#4453).

Setup is a three-node test: node 0 creates a doc with points=10 on a pcounter/pncounter field, subscribes node 1 to the doc, sets node 2 as a replicator of node 0, then updates points by +10 on node 0 and waits for sync. Expected final points is 20 on every node. Without signing this passes. With EnableSigning: true, node 0 reads 20 but nodes 1 and 2 both read 30.

It only breaks when there's an update after the initial create. The WithAdd variants of these tests (no update, just replicate the create) pass under signing. Same thing on the local side — all the non-P2P counter update tests in mutation/update/crdt pass. So it's specifically signed counters going through replication, and specifically when there's a priority >= 2 block that arrives alongside a signed priority-1 field block.

My guess is the initial delta (10) is being applied twice on the replicated peers somewhere in the merge path. Worth looking at how the signed priority-1 field block interacts with a later unsigned delta when the whole chain is pulled over via the replicator — signing only signs composites and priority-1 field blocks, so there's a mixed-signature chain on the wire and I wonder if a verify/skip branch is double-processing the initial block.

Reproduction is literally the existing tests, just run with the multiplier active:

    DEFRA_MULTIPLIERS=signed-docs go test ./tests/integration/net/simple/peer_replicator/crdt -run 'TestP2PPeerReplicatorWithUpdate_P(N)?Counter_NoError' -count=1

Affects both pcounter and pncounter equally. Blocking the enabling of the signed-docs CI job until this is fixed.

islamaliev:
Broader than replicator-only. The same double-apply shows up under pubsub peer sync (ConnectPeers), not just AddReplicator:

- tests/integration/net/simple/peer/crdt/pcounter_test.go — TestP2PUpdate_WithPCounter_NoError fails with points=30 (expected 20)
- tests/integration/net/simple/peer/crdt/pncounter_test.go — TestP2PUpdate_WithPNCounter_NoError fails with points=30 (expected 20); TestP2PUpdate_WithPNCounterThreeNodeSimultaneousUpdate_NoError fails with points=330 (expected 130)

[...]

PR #4746, opened 2026-04-23
test: Add signed-docs test multiplier

### LINKED ISSUE #4453: Add test multiplier for signed docs
Signing documents/collections is a complexity multiplier, and should be tested as such.

islamaliev:
## Relevant issue(s)

Resolves #4453
Surfaced bug: #4742

## Description

Adds a `signed-docs` test multiplier that flips `TestCase.EnableSigning = true` across the integration suite, so every feature is exercised with DAG blocks that carry a `Signature` link. Mirrors the `secondary-index` multiplier in shape.

The multiplier itself is an `Apply`-noop: signing is a TestCase-level flag rather than an action, so the effect is applied via a small hook in `applyMultipliers` (`tests/integration/utils.go`). The only auto-skip is for `*action.VerifyBlockSignature` actions — everything else uses explicit `MultiplierExcludes: []string{multiplier.SignedDocs}` so incompatibilities are grep-able and documented at the test case level.

Enabled via `DEFRA_MULTIPLIERS=signed-docs`. Added a matching `test-coverage-signed-docs` job to the coverage workflow.

Also introduces a `testUtils.ValidCID()` matcher for places where a commit test just wants to assert "some valid CID", so those tests stay strong under the multiplier instead of hardcoding values.

Rolling out the multiplier surfaced a replicator bug — signed counter deltas are double-applied across peers (#4742) — so the pcounter/pncounter P2P tests are excluded pending a fix. Two test-harness follow-ups also came out of the audit and are tracked separately:
- #4744 — template placeholder for collection-level commit CIDs (blocks the `branchables/cid_test.go` exclusion)
- #4745 — apply template placeholders to `Request.Results` (blocks strict result-cid assertions in a few commit tests)

Once the job passes on `develop`, it will need to be added to the required status checks in branch protection to actually gate merges, the same as the `secondary-index` job.

## Tasks

## How has this been tested?

Ran the full integration suite with and without `DEFRA_MULTIPLIERS=signed-docs`, and the combined `DEFRA_MULTIPLIERS=secondary-index,signed-docs`.

- MacOS


docs/data_format_changes/i4453-signed-docs-multiplier.md
```diff
@@ -0,0 +1,8 @@
+# Signed-docs test multiplier
+
+Adds a `signed-docs` test multiplier that flips `TestCase.EnableSigning = true`
+across the integration suite. Signing attaches a `Signature` link to composite
+blocks and priority-1 field blocks in the Merkle DAG, which changes block bytes
+and therefore CIDs. This is intentionally incompatible at the block level with
+databases created without signing, so the change detector cannot verify
+compatibility across the two modes.
```
fredcarle: question: Did you mean `two nodes`? Or signing vs non-signing modes?

tests/integration/query/commits/with_doc_id_count_test.go
```diff
@@ -38,18 +40,9 @@ func TestQueryCommitsWithDocIDAndLinkCount(t *testing.T) {
...
-							"cid":   "bafyreigonvri5vfdosfgp4qxtq46snjxm7cnjlzizrod2wy3l53jbxiysm",
-							"COUNT": 0,
-						},
-						{
-							"cid":   "bafyreiejjfevlp5wrfl5o7bxbdtjj4th36lbdjov5gdkmy5n5jzs6dcmpu",
-							"COUNT": 2,
-						},
+						{"cid": uniqueCid, "COUNT": 0},
+						{"cid": uniqueCid, "COUNT": 0},
+						{"cid": uniqueCid, "COUNT": 2},
```
fredcarle, line 45: todo: This isn't acheiving exactly what you want. It works because the result set is in the correct order. But if you change the order, the test fails. For it to work, you need a `uniqueCID` for each of the results. Not the same one for all of them.

fredcarle (approved):
Approving with one todo before merge.

tests/integration/query/commits/with_doc_id_count_test.go
islamaliev: I think the single shared uniqueCid is actually doing what we want here. NewUniqueValue returns a matcher whose contract is that every occurrence of that same matcher across the result set must observe a distinct value, so reusing one uniqueCid across the three slots asserts the three cids are all different from each other (which matches the original hardcoded test). That holds regardless of order, since NonOrderedResults just permutes the slots. If we switched to three separate NewUniqueValue calls, each one would only see a single occurrence, so the uniqueness check trivially passes and we lose the assertion that the three cids are distinct, which is actually weaker than what we have now. So I think leaving it as is keeps the original intent. Happy to be wrong though if I am misreading the matcher semantics, let me know.

docs/data_format_changes/i4453-signed-docs-multiplier.md
islamaliev: Meant signing vs non signing modes, not two nodes. The change detector compares one db build against another, so the incompatibility here is between the signing enabled and signing disabled persistence formats. Reworded the line to spell that out.

tests/integration/query/commits/with_doc_id_count_test.go
fredcarle: I see what you mean. We have to remove the `NonOrderedResults` in that case because a change in order will cause these tests to fail.

PR #4750, opened 2026-04-27
fix: Seed CollectionVersions from DB

islamaliev:
## Relevant issue(s)

Follow-up to #4723.

## Description

#4723 introduced `{{.CollectionVersionID<N>}}` templates, but the only test using them, `TestCollectionMigrationQueryMigratesAcrossMultipleVersions`, fails under the change detector and is currently blocking PRs. Under the change detector, setup actions (`AddCollection`, `AddDoc`) run on the source branch in one process while the rest run on the target branch in another, and `s.CollectionVersions` doesn't survive that boundary — so `{{.CollectionVersionID0}}` resolved to v1 (the first patch result) instead of v0.

The data-format-changes doc added in #4723 made the change detector pass for that PR (a new doc file causes it to skip), but once merged the doc was no longer "new" and every subsequent PR started failing the same test.

This seeds `s.CollectionVersions` at the start of the test from whatever collection versions already exist in the DB, walking the `PreviousVersion` chain so the indices match creation order. With this in place the templates resolve correctly in both phases of the change detector.

## Tasks

## How has this been tested?

Reproduced the failing change detector run locally by splitting `TestCollectionMigrationQueryMigratesAcrossMultipleVersions` into setup-only and target phases via the `DEFRA_CHANGE_DETECTOR_*` env vars; confirmed it fails before the fix and passes after. Also ran the full `tests/integration/collection_version/...` suite under both phases.

- MacOS

### DIFF
--- a/tests/integration/utils.go
+++ b/tests/integration/utils.go
@@ -303,6 +303,7 @@ func executeTestCase(
 	// by the change detector so we should fetch them here at the start too (if they exist).
 	// collections are by node (index), as they are specific to nodes.
 	refreshCollections(s, immutable.None[int](), immutable.None[state.Identity]())
+	seedCollectionVersionsFromState(s)
 	refreshDocuments(s, testCase, startActionIndex)
 
 	for i := startActionIndex; i <= endActionIndex; i++ {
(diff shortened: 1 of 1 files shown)


tests/integration/utils.go
```diff
@@ -953,6 +954,71 @@ func refreshTokens(
 	}
 }
 
+// seedCollectionVersionsFromState populates s.CollectionVersions with the version IDs
+// of any collection versions already present in the database. This ensures that
```
AndrewSisley, line 958: todo: This change means that the change detector is not detecting changes here.  The template `{{.CollectionVersionID0}}` can have two different values, one on either side of the setup-assert change detector split.  

We are using the new Defra version's state to assert the new version's state, so in a lot of cases, it is not really testing anything (it can still assert that it is consistent within the post-setup test actions though).

I think if this issue was not blocking the build, I would ask you to fix this in this PR, but I am happy for you to instead create a new issue and link to it from a code comment here instead please.

AndrewSisley (approved):
Approving to unblock the CI, but I don't think the solution is the one we want long term.

tests/integration/utils.go
islamaliev: Good point. Filed #4752 to track the gap and added a comment in `seedCollectionVersionsFromState` linking to it.

PR #4755, opened 2026-04-27
fix: Persist test state for change detector

### LINKED ISSUE #4752: Change detector blind spot for CollectionVersionID templates
Just leaving a note about a gap I noticed while unblocking the change detector in #4750.

The fix there seeds the CollectionVersions list (used to resolve {{.CollectionVersionID0}} and friends in tests) by reading the collection history out of the database at the start of the test. That works fine for getting the templates to point at the right thing, but it means the values come from the post-upgrade Defra in both the setup phase and the assert phase of the change detector. So if a future change to Defra altered how a collection version ID is computed or serialized, a test that asserts using one of those templates would silently round-trip and the change detector wouldn't catch it.

For the test that motivated this fix it's mostly fine in practice — only the original collection version is shared across the source/target split, and reading it back from disk gives the same bytes the older Defra wrote. The real concern is more general: as soon as somebody writes a setup-phase test that produces multiple collection versions and then asserts on them with the templates, the assertions stop being protected by the change detector for those values.

The proper fix is probably to capture the resolved template values during the setup phase, persist them next to the test data dir, and read them back in the assert phase instead of recomputing from the post-upgrade state. Not urgent, but worth tracking so we don't forget the coverage hole.

islamaliev:
## Relevant issue(s)

Resolves #4752.

## Description

Follow-up to #4750. That PR unblocked the change detector by re-deriving `s.CollectionVersions` from the on-disk DB at the start of the assert phase, but it pulled the values from the post-upgrade Defra on both sides of the split — so any assertion using a `{{.CollectionVersionID<N>}}` template was effectively comparing target-against-target rather than source-against-target. The change detector wasn't actually detecting changes for those values.

This wires up a small JSON sidecar in the per-test change-detector data directory. The source phase writes `s.CollectionVersions` (and `s.LensIDs`, reserved for future use) to the file at the end of the action loop; the assert phase reads it back at startup and uses those exact strings to resolve templates. So if the new Defra version changes how a collection version ID is derived, a test that asserts on one of those templates will now flip from passing to failing — which is what the change detector is for.

`seedCollectionVersionsFromState` stays in place as the fallback, used when no sidecar is present (older source branch, or a test with no setup phase). I also confirmed the negative case manually by corrupting the sidecar between phases — the assert phase fails loudly, proving the coverage is real.

## Tasks

## How has this been tested?

Unit tests for the new `WriteTestState` / `ReadTestState` helpers (round-trip, missing file, corrupt file, no leftover .tmp). Locally simulated the change detector source/assert split for the originally-failing migration test and the full `tests/integration/collection_version/...` suite — both phases green. Also exercised the fallback path by deleting the sidecar between phases and the corruption path by tampering with it.

- MacOS


tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+// WriteTestState marshals state as JSON and writes it atomically into the
+// per-test change-detector data directory. The write goes to a sibling
+// `*.tmp` file and is then renamed into place, so a partially-written file
+// is never visible to a concurrent reader (or to the assert-phase process).
+//
+// It is a programming error to call this when changeDetector is not in
+// SetupOnly mode; callers must guard accordingly.
+func WriteTestState(t testing.TB, state TestState) error {
+	dir := DatabaseDir(t)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
```
AndrewSisley: todo: Please document these flags and why you have chosen them (also ln 68)

tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+
+// stateFilePath returns the absolute path of the sidecar file for the given test.
+func stateFilePath(t testing.TB) string {
+	return filepath.Join(DatabaseDir(t), stateFileName)
+}
+
+// WriteTestState marshals state as JSON and writes it atomically into the
+// per-test change-detector data directory. The write goes to a sibling
+// `*.tmp` file and is then renamed into place, so a partially-written file
+// is never visible to a concurrent reader (or to the assert-phase process).
```
AndrewSisley: question: Why are you bothering with this concurrency semi-protection?  The database file paths are deterministic, so running multiple change detectors concurrently will be horribly broken anyway.

tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := os.Rename(tmpPath, stateFilePath(t)); err != nil {
+		// best-effort cleanup of the tmp file on failure
+		_ = os.Remove(tmpPath)
+		return err
+	}
+	return nil
+}
+
+// ReadTestState reads the sidecar file written by WriteTestState in the
```
AndrewSisley: suggestion: This reads like an LLM wrote it, and I think it is causing more confusion to this reader than it could ever save.  Consider reworking it.

tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+	data, err := os.ReadFile(stateFilePath(t))
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return TestState{}, false, nil
+		}
+		return TestState{}, false, err
+	}
+
+	var state TestState
+	if err := json.Unmarshal(data, &state); err != nil {
```
AndrewSisley: todo: `return json.Unmarshal(data, &state)`

tests/integration/utils.go
```diff
@@ -951,18 +964,40 @@ func refreshTokens(
...
+// here so templates resolve to the SAME bytes the source side saw — that's
+// what gives the change detector its coverage of those values.
+//
+// If the file is missing (older source branch, or no setup phase), we fall
+// back to seedCollectionVersionsFromState which re-derives version IDs from
+// whatever is already in the on-disk DB.
+//
+// Outside the change detector we just call the seeder, which is a no-op when
+// the database is fresh.
+func seedCollectionVersions(s *state.State) {
```
AndrewSisley: suggestion: This is a very inefficient way of structuring the code - it suggests that each property will re-read the entire json file.  Please rename this function, or rework it.

tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+// backward compatible way because encoding/json silently ignores unknown
+// fields on read and zero-values missing fields.
+type TestState struct {
+	// CollectionVersions is the ordered list of collection version IDs created
+	// during the source phase, used to resolve {{.CollectionVersionIDN}}
+	// templates on the assert side against the values the source side produced.
+	CollectionVersions []string `json:"collectionVersions"`
+	// LensIDs is the ordered list of lens IDs registered during the source
+	// phase. Reserved for future use; not currently consumed by any test.
+	LensIDs []string `json:"lensIDs,omitempty"`
```
AndrewSisley: todo: Please either remove the dead code, or make it not dead

tests/change_detector/state.go
```diff
@@ -0,0 +1,102 @@
...
+// "file missing" (acceptable: source branch may pre-date the sidecar, or
+// the test wrote nothing) from "file corrupt" (a real error).
+//
+//   - (state, true,  nil) — file existed and parsed cleanly
+//   - (zero,  false, nil) — file did not exist
+//   - (_,     false, err) — file existed but could not be read or parsed
+func ReadTestState(t testing.TB) (TestState, bool, error) {
+	data, err := os.ReadFile(stateFilePath(t))
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
```
AndrewSisley: todo: This is dead code as the file is always written, whether it is empty or not, and the resultant boolean accommodating it complicates this function, and the function calling it an unreasonable amount at the moment, please remove it.

AndrewSisley (changes requested):
I'm trying to control my biases here, but it feels like I spent longer reviewing this than you did after claude generated it.

I hope that I am wrong, and I just happen to be in a harsh mood or something this morning, but it feels like your usual PRs are of a higher quality than this.

tests/change_detector/state.go
islamaliev: Will add a short note on each.

tests/change_detector/state.go
islamaliev: Fair, the two phases run sequentially so there's no concurrent reader to protect against. Will drop the temp+rename and write directly.

tests/change_detector/state.go
islamaliev: will do

tests/change_detector/state.go
islamaliev: will do

tests/integration/utils.go
islamaliev: hm... I can rename to `loadCollectionVersions`. I think it matches better

tests/change_detector/state.go
islamaliev: will remove

tests/change_detector/state.go
islamaliev: The bool is there because the source branch is usually whatever's currently on develop, and the day this merges every in-flight PR will hit a source phase that didn't write the file. So the missing case is real for a release cycle or two, but you're right that the tri-state return is ugly. I'll change the signature to return an error and let the caller check for the missing-file case in the usual way. Same fallback behaviour, cleaner shape.

islamaliev:
> I'm trying to control my biases here, but it feels like I spent longer reviewing this than you did after claude generated it.

You're right, this one was mostly LLM-driven. My intent was to use it for the back-and-forth on what the options were, then once the design felt solid have it implement the chosen one and ask you for a quick sanity check rather than a full review. I should have opened it as a draft and said that up front so you knew what kind of pass to do. That's on me, sorry for the wasted time, Andy. Will fold in your comments, and next time I'll be clearer about what the PR is and isn't.

tests/change_detector/state.go
AndrewSisley: Ah I see what you mean, thanks for the explanation, and thanks for tweaking it for long term consumption instead of the short-term need.

AndrewSisley:
>  You're right, this one was mostly LLM-driven.  My intent was to use it for the back-and-forth on what the options were, then once the design felt solid have it implement the chosen one and ask you for a quick sanity check rather than a full review.

Ah fair enough, thanks for coming clean - I like the overall approach (json temp file), I should have made that clearer in my original review :)

There might be more efficient ways to serialise the state if we want in the future, but that is easy enough to change later, and json is probably the easiest maintenance-wise.

AndrewSisley (approved):
LGTM, thanks Islam

PR #4763, opened 2026-04-30
fix: Prevent txn-actions from executing concurrently

### LINKED ISSUE #4288: It appears possible to commit/discard manually managed transactions during action execution

### LINKED ISSUE #4626: ConcurrentTransaction is a lie, and will break many things if concurrent calls are made to it

### LINKED ISSUE #4717: Providing an unknown txn id to the CLI results the command not using a txn
E.g. when there is no transaction with `id==100`, the below command and others, will ignore the param:
```
defradb client query --tx=100 '
  mutation {
      add_User(input: {age: 31, verified: true, points: 90, name: "John"}) {
          _docID
      }
  }
'
``` 

There is no warning to to user, the command will instead auto-commit, using an implicit transaction.

This is probably not just limited to the CLI.

This might be handled by https://github.com/sourcenetwork/defradb/issues/4626

### LINKED ISSUE #4756: Badger panics when using a discarded/committed txn
It should be easy enough for us to return an error instead, probably from `./db/txn.go` as it can make use of the threading protection there.

AndrewSisley:
## Relevant issue(s)

Resolves #4626
Resolves #4756
Resolves #4288
Resolves #4717

## Description

Prevents the execution of multiple actions on the same Defra transaction from executing concurrently.

Previously this would allow the database to be corrupted (concurrent writes), or could take down the badger instance (concurrent write-discard/commit).

It also prevents a panic which would occur if attempting to use a transaction that had already been discarded/committed - I fixed this here, as it was getting in the way of testing #4626, and the fix for the transaction reuse panic was easier easier by the concurrency-prevention fix.

Also removes the old ConcurrentTxn type, as individual transactions do not attempt to execute their actions concurrently anymore - leaving it without purpose.

It also returns an error when an unknown transaction id is provided to the http client #4717  - this is fixed here, as the fix was required by the tests for #4626 and that fix was required for the test added for #4717.

It might be easier to remove this one commit by commit, as a bunch of stuff needed to be changed/fixed in order to get the tests to pass, some of the commit messages try and explain what/why is being done and might be helpful.


internal/db/txn.go
```diff
@@ -31,41 +32,77 @@ import (
...
+	//
+	// To track this we use the `txnCtxInProgressKey` context key - because setting this creates
+	// a new child context, we never need to worry about leakage or cleanup beyond the top-level
+	// public txn call.  It also means read/writing it in this way is inherently thread safe - if
+	// two threads concurrently try to set it up, they will have two different contexts and the
+	// inProgressLock will make them run serially (as they should).
+	thisContextOwnsTxnLock := ctx.Value(txnCtxInProgressKey{}) == nil
+	if thisContextOwnsTxnLock {
+		ctx = context.WithValue(ctx, txnCtxInProgressKey{}, struct{}{})
+		txn.inProgressLock.Lock()
```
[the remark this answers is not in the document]
AndrewSisley: This is fine, but I need to document why

- [x] Expand documentation

http/middleware.go
```diff
@@ -85,10 +86,17 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 // CollectionMiddleware sets the collection context for the current request.
 func CollectionMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		db := mustGetContextClientDB(req)
+		//db := mustGetContextClientDB(req)
```
ChrisBQu: nitpick: It looks like this comment was meant to be removed

tests/action/utils.go
```diff
@@ -94,11 +94,11 @@ func RefreshCollections(
 // If one is not provided, it will default to running the GetCollections function on the node itself.
 // Importantly, this will use the same ordering as would be found in the node.Collections slice that
 // is refreshed by the RefreshCollections function.
-func GetCanonicallyOrderedCollections(
+func getCanonicallyOrderedCollectionsE(
```
ChrisBQu: suggestion: I don't love this function name. It's not a blocker for me, but I would prefer the E is removed, or that the name be longer, and abbreviations not used.

praise: Good catch on it not needing to be an exported.

internal/db/txn.go
```diff
@@ -31,41 +32,81 @@ import (
...
+		return nil, nil, NewErrUnsupportedTxnType(ctxTxn)
+	}
 
-		// If the txn has already been set on the context but it hasn't already been set as explicit,
-		// we create a copy of the txn and mark it as an explicit txn.
-		explicitTxn := &Txn{
-			txn.BasicTxn,
-			txn.db,
-			true,
+	if !txn.explicit && existsOnCtx {
```
ChrisBQu, line 53: suggestion: I liked the comment that was removed here. I don't think it is necessary, but I think it draws attention to the fact that we check !txn.explicit, and then set explicit to true on the new txn. This is purposeful, and correct. But at a glance, I'm not sure it's obvious why it is correct, and the comment makes the reader slow down.

This is purely opinion, and for your consideration. Disregard if you like. :)

ChrisBQu (approved):
LGTM. I have left a few thoughts, but I have no blockers, and no push back if you disagree with any or all of my suggestions.

http/middleware.go
AndrewSisley: That's a bad miss, thanks Chris.

- [x] remove commented out code

tests/action/utils.go
AndrewSisley: This is a new overload, the diff just makes it look like a rename. `GetCanonicallyOrderedCollections` still exists and still called by a couple of places in the old framework.

I think I'll rename `GetCanonicallyOrderedCollections` to `MustGetCanonicallyOrderedCollections`, and `getCanonicallyOrderedCollectionsE` to `getCanonicallyOrderedCollections` - I was pretty lazy naming this.

- [x] Rename `GetCanonicallyOrderedCollections` etc

internal/db/txn.go
AndrewSisley: Not a problem, I'll add it back.

- [x] Re-add old comment.

PR #4775, opened 2026-05-04
refactor: Use JS txn directly instead of context

### LINKED ISSUE #4774: Rework JS collection to use the txn on the collection
Instead of the old context path.

Broken out of, and a requirement for https://github.com/sourcenetwork/defradb/pull/4763

AndrewSisley:
## Relevant issue(s)

Resolves #4774
Broken out of, and a requirement for #4763

## Description

Uses the JS transaction directly instead of passing via context.

The previous solution meant that the transaction on the db collection object was not actually being used from JS, bypassing `db.Txn` completely.  It also acted as another source of friction when trying to de-tangle the mess of a situation we have with transactions and contexts.

### DIFF
--- a/js/client_collection.go
+++ b/js/client_collection.go
@@ -13,7 +13,6 @@
 package js
 
 import (
-	"sync"
 	"syscall/js"
 
 	"github.com/sourcenetwork/defradb/client"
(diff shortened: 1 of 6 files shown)


nasdf (approved):
Thanks for fixing this. It got out of sync at some point when we refactored the transaction API.

PR #4778, opened 2026-05-05
fix: KMS lookup under NAC for encrypted DAG sync

### LINKED ISSUE #4777: KMS encryption-key serving fails under NAC: internal collection lookup runs without node identity
## Problem

When document encryption is active and NAC is enabled, encrypted-block P2P sync via `SyncBranchableCollection` fails. The serving peer can't authorise its own internal collection lookup while answering an encryption-key request.

User-visible error on the requesting peer:

```
failed to retrieve encryption key during DAG sync: EOF
```

## Root cause

`internal/kms/pubsub.go:doesIdentityHaveDocPermission` calls `s.colRetriever.RetrieveCollectionFromDocID(ctx, docID)` with **no identity**. That call ends up in `internal/db/store.go:GetCollections` which gates on:

```go
db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeGetCollectionPerm)
```

With NAC enabled and no identity supplied, NAC denies (`not authorized to perform operation. Permission: get-collection`). `RetrieveCollectionFromDocID` then sees `len(cols) == 0` and returns `NewErrCollectionNotFoundForCollectionVersion(...)` — which the KMS handler propagates back as an empty wire reply — which the requester decodes as `EOF`.

The KMS service already holds a node-scoped DID (`s.nodeDID` in `internal/kms/pubsub.go`, set from `nodeIdentity.Value().DID()` in `internal/db/p2p/p2p.go:235`). It just isn't propagated into the collection lookup. NAC's `checkNodeAccess` already short-circuits when the caller IS the node identity, so the existing path would succeed if the identity were threaded through.

## Secondary issue (in scope for the same fix)

`internal/db/collection_retriever.go:73` swallows the underlying error:

```go
cols, _ := txn.GetCollections(ctx, opt)
if len(cols) == 0 {
    return nil, client.NewErrCollectionNotFoundForCollectionVersion(...)
}
```

This rewrote the real `"not authorized to perform operation"` error as `"collection not found. CollectionVersionID: …"`, which made the bug significantly harder to diagnose. Same pattern would mask transient errors and other classes of failures. Fix should surface the underlying error rather than collapsing all failure modes into "not found".

## Why this didn't surface before

- `tests/integration/encryption/**` tests don't enable NAC, so `checkNodeAccess` is a no-op.
- Tests that enable NAC don't use encryption today, so KMS isn't engaged.
- Surfaced by the encrypted-docs test multiplier (#4452) which combines encryption with the rest of the suite.

## Affected tests

- `tests/integration/acp/nac/sync_p2p_branchable_collection_test.go::TestNAC_GatesSyncP2PBranchableCollection_AuthorizedIdentity_AllowAccess`
- `tests/integration/acp/nac/relation_admin/sync_p2p_branchable_collection_test.go::TestNAC_AdminRelation_CanSyncP2PBranchableCollection`

Likely (not yet verified): other `acp/nac/**` tests that combine NAC with encrypted P2P sync.

## Reproduction (without the multiplier)

Take either affected test, add `IsDocEncrypted: true` to its `AddDoc`, and add `KMS: testUtils.KMS{Activated: true}`. Reproduces deterministically.

## Fix direction

Two viable shapes:

1. **Use the node identity for KMS-internal lookups.** Construct `colRetriever` (at KMS init) with `WithIdentity(<nodeIdentity>)`, or pass the identity through `RetrieveCollectionFromDocID`. Mirrors how the merge processor handles legitimate node-scoped operations.
2. **Bypass NAC for KMS-internal calls.** Less appealing — adds a separate "skip ACP" channel that's easy to misuse.

## Related

- #4452 — Encrypted-docs test multiplier (where this surfaced).
- #4762 — Querying encrypted docs by `cid:` fails (different bug, same multiplier-rollout).
- Multi-peer KMS first-responder bug (separate issue, also surfaced by #4452, not yet filed).

islamaliev:
## Relevant issue(s)

Resolves #4777

## Description

When NAC is enabled and a doc is encrypted, cross-peer sync of the encrypted blocks fails on the requester with `failed to retrieve encryption key during DAG sync: EOF`. The serving peer's KMS pubsub service was looking up the doc's collection without any identity attached, NAC denied the implicit `get-collection` call, the local key fetch errored out, and the empty pubsub reply was decoded as EOF on the requester.

The earlier revision of this PR fixed it by baking the node's identity into the `collectionRetriever` at construction time. Review feedback rightly flagged that as too broad — any future caller of that retriever instance would silently inherit node-level NAC privileges. The retriever is now identity-per-call: the caller hands in whichever identity it wants the lookup to run as, and the public retriever no longer carries any identity in its struct.

Inside KMS this surfaces a deliberate split between two identities. The serving peer's own collection lookup runs as the serving peer itself (a node-internal bookkeeping operation that NAC must authorise as the node, not as the requester). The doc-level DAC check runs as the requester (the principal asking for the key). These were previously conflated, which either over-restricted legitimate flows or under-restricted malicious ones depending on the test configuration.

The KMS wire request now carries the caller's user identity instead of the requester node's DID. `prepareFetchEncryptionKeyRequest` reads the identity from the request context (set by `SyncBranchableCollection` at the entry point) and falls back to the node identity only when no user identity is present (e.g. gossip-triggered fetches from inside `processPushlogRequest`). This means an authorised user's NAC/DAC grants on the serving peer actually apply to the key fetch, rather than the wire identity always being the requester's own node DID.

Collection-scoped encryption blocks (e.g. those produced for a `@branchable` collection's own head, where there is no DocID to authorise against) now route through a node-level NAC gate using the actor identity from the wire request, closing a leak where any peer could fetch the collection-head encryption key without any authorisation. The doc-permission helper itself has been tightened to require a non-empty docID, with the empty-DocID branch handled explicitly at the caller.

`pubSubService.nodeACP` is now a getter (`func() acpDB.NACInfo`) rather than a value captured at construction. NAC is initialised by `db.Start` after `p2p.New` runs, so the captured snapshot would always read as `NACNotConfigured` and silently disable all NAC checks in KMS. The getter reads the live state at check time.

`trySelfHasAccess` on the pushlog path was similarly looking up the local collection without an identity, which denied legitimate gossip-driven syncs under NAC. It now passes the serving node's identity to its `GetCollections` call.

The retriever no longer swallows the underlying `GetCollections` error and rewrites it as a generic "collection not found" — the original "not authorized" message now propagates, which made the diagnosis above possible in the first place.

## Tasks

## How has this been tested?

New integration tests live in `tests/integration/encryption/peer_nac_test.go` and exercise the branchable-collection sync path with KMS activated, NAC enabled, and `IsDocEncrypted: true`:

- An authorised identity completes the sync end-to-end.
- An unauthorised identity is denied at the upstream NAC gate.
- A request with no identity is denied.
- An identity granted admin via `AddNACActorRelationship` then succeeds.
- That same identity, once revoked via `DeleteNACActorRelationship`, is denied on the next sync.
- A separate test exercises the gossip-driven path (subscription + push) with the subscriber's node identity granted admin, confirming the gossip-triggered KMS fetch and `trySelfHasAccess` both authorise correctly under NAC.

Both pre-existing NAC sync tests (`tests/integration/acp/nac/sync_p2p_branchable_collection_test.go` and `tests/integration/acp/nac/relation_admin/sync_p2p_branchable_collection_test.go`) pass unmodified, and the full encryption and NAC suites stay green under both \`DEFRA_CLIENT_TYPE=go\` and \`DEFRA_CLIENT_TYPE=http\`.

- MacOS


tests/integration/encryption/peer_nac_test.go
```diff
@@ -0,0 +1,77 @@
...
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+// TestDocEncryptionNAC_SyncBranchableCollection_AuthorizedIdentity_AllowAccess reproduces a bug
+// where the KMS pubsub service on the serving peer cannot authorize its own internal collection
+// lookup when NAC is enabled and the doc is encrypted: doesIdentityHaveDocPermission calls
+// RetrieveCollectionFromDocID with no identity, NAC denies the implicit GetCollections call
+// (NodeGetCollectionPerm), the local key fetch errors out, and the requesting peer surfaces
+// "failed to retrieve encryption key during DAG sync: EOF".
```
shahzadlone: suggestion: Reword to describe current behavior and edge case we are testing rather than previous buggy behavior?

shahzadlone (changes requested):

internal/kms/pubsub.go
```diff
@@ -342,6 +342,12 @@ func (s *pubSubService) doesIdentityHaveDocPermission(
 	}
 
 	docID := string(entBlock.DocID)
+	if docID == "" {
+		// Collection-scoped encryption blocks (e.g. those produced for a `@branchable`
+		// collection's own head) carry no DocID and therefore have no doc-level permission
+		// to enforce here.
+		return true, nil
```
shahzadlone: question: All non-identity/empty-identity cases are now going to by-pass DAC? but go through DAC if identity is specific, isn't this a possible issue where to gain malicious access docID can be emptied?

question: The PR title suggested to me this is a NAC specific issue, isn't DAC already going to get bypassed anyways if we pass in an identity that has NAC access (so instead why not use an NAC granted identity)?

islamaliev: > question: All non-identity/empty-identity cases are now going to by-pass DAC? but go through DAC if identity is specific, isn't this a possible issue where to gain malicious access docID can be emptied?

this isn't bypassing DAC for the requester, it's recognizing that some encryption blocks aren't doc-scoped and therefore have no doc-level ACL to consult.
entBlock.DocID comes from the locally-stored encryption block, a malicious requester can't "empty the docID" to gain access

> question: The PR title suggested to me this is a NAC specific issue, isn't DAC already going to get bypassed anyways if we pass in an identity that has NAC access (so instead why not use an NAC granted identity)?

I'm not sure I totally understand
NAC gated txn.GetCollections. KMS used a retriever with no identity, so NAC saw "no identity" and denied the collection lookup before DAC was ever consulted. So I attached db.nodeIdentity to the retriever.
DAC is checked separately on the encryption block's own DocID once the collection lookup has succeeded.

shahzadlone: > this isn't bypassing DAC for the requester, it's recognizing that some encryption blocks aren't doc-scoped and therefore have no doc-level ACL to consult. entBlock.DocID comes from the locally-stored encryption block, a malicious requester can't "empty the docID" to gain access

suggestion: In that case how about we either document the private function to reflect the encryption block specific case and rename the function, I think the `doesIdentityHaveDocPermission` is a general name that makes me feel like what if it is accidentally called else where and we leak access. Or I would also be okay instead if you just want to do this if check in `getEncryptionKeysLocally` and pass perhaps the docID into this helper.

> I'm not sure I totally understand NAC gated txn.GetCollections. KMS used a retriever with no identity, so NAC saw "no identity" and denied the collection lookup before DAC was ever consulted. So I attached db.nodeIdentity to the retriever. DAC is checked separately on the encryption block's own DocID once the collection lookup has succeeded.

That was going to be my suggestion to attach the `nodeIdentity` (which would pass NAC and IIRC also bypass DAC), perhaps i missed that, where does that happen?

internal/kms/pubsub.go
```diff
@@ -62,7 +65,11 @@ type pubSubService struct {
 	nodeACP      acpDB.NACInfo
 	documentACP  immutable.Option[dac.DocumentACP]
 	colRetriever CollectionRetriever
-	nodeDID      string
+	// nodeIdentity is this node's own identity. Used to authorize node-internal
+	// operations (e.g. NAC-gated collection lookups while answering KMS key
+	// requests). It is NOT the requester's identity — that travels on the wire
```
shahzadlone: nitpick: I dislike these em-dashes (—) AI agents insert, prefer hyphens (`-`) over them.

shahzadlone (approved):
LGTM thanks for doing the changes

PR #4784, opened 2026-05-06
WIP

### LINKED ISSUE #4510: No protection for concurrent updates to collections during a version delete
Everything besides version deletion is either append only, or protected by transaction conflicts (local stuff such as indexes), and so should not be a problem.

Deleting versions that are not the last version should not be problematic, as the model is designed to allow for gaps in the version history.

When deleting the last version, and the short-id, there will be no documents in the store, but if a concurrent patch tries to add a new version, it will try and use the existing short-id, only for it's sid index to be deleted by the deleting thread - causing problems when documents are later added.

If the thread-txn adding a new version first creates a document, it will hold the collection read lock until it is committed, so this is not an issue in that scenario, as the delete will happen after the update.

This is probably dependent on https://github.com/sourcenetwork/defradb/issues/4511

AndrewSisley:
## Relevant issue(s)

Resolves #4510

## Description

### DIFF
--- a/tests/action/patch_collection.go
+++ b/tests/action/patch_collection.go
@@ -13,6 +13,7 @@ package action
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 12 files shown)


AndrewSisley:
Too much effort to get the test(s) to be reliable.  No bug found atm after several CI runs.

PR #4790, opened 2026-05-07
fix: Decrypt blocks during versioned-fetcher replay

### LINKED ISSUE #4787: bug(p2p+enc): GraphQL subscription on subscriber peer silently drops update for encrypted incoming doc
## Summary

When peer A creates an encrypted document and peer B has both a P2P collection subscription and an active GraphQL subscription on the same collection, peer B's GraphQL subscription never delivers a result. From the test harness's perspective the subscription times out (\"timeout occurred while waiting for data stream\"). The underlying cause is **silent**: the merge happens, the `event.Update` fires, the subscription handler receives it — but when the planner-driven selection runs it hits a CBOR decode error on encrypted field bytes and the handler drops the event without surfacing the error.

This is two bugs at once: a primary bug in the selection path that doesn't decrypt encrypted fields when used from a subscription, and an amplifier in `internal/db/subscriptions.go` that swallows the resulting error.

## Reproduction

```
DEFRA_CLIENT_TYPE=go DEFRA_MULTIPLIERS=encrypted-docs go test \
  ./tests/integration/net/simple/peer \
  -run TestP2PAdd_WithP2PCollectionAndSubscription_ShouldSucceed
```

Test fails with:
```
Error: timeout occurred while waiting for data stream
```

## Trace

Filtered trace from a failing run (`[sub]` = `internal/db/subscriptions.go`, `[p2p.processPushlogRequest]` = `internal/db/p2p/p2p.go`, `[kms]` = `internal/kms/pubsub.go`):

```
[sub] subscribed to event.UpdateName, waiting...
[kms] handleRequestFromPeer me=B from=A topic=encryption
[kms] handleRequestFromPeer ok nBlocks=1
[kms] response usable, nItems=1 — done
[kms] handleRequestFromPeer me=B from=A topic=encryption
[kms] handleRequestFromPeer ok nBlocks=1
[kms] response usable, nItems=1 — done
[kms] handleRequestFromPeer me=B from=A topic=encryption
[kms] handleRequestFromPeer ok nBlocks=1
[kms] response usable, nItems=1 — done
[p2p.processPushlogRequest] calling Merge DocID=\"bae-3d8258b9-…\" Cid=bafyrei…w6hu
[p2p.processPushlogRequest] Merge OK, publishing event.Update DocID=\"bae-3d8258b9-…\" Cid=bafyrei…w6hu
[sub] received event.Update DocID=\"bae-3d8258b9-…\"
[sub] RunSelection done DocID=\"bae-3d8258b9-…\" resultLen=0 err=cbor: 24 bytes of extraneous data starting at index 5
[sub] all ops produced empty slices, dropping event DocID=\"bae-3d8258b9-…\"
[sub] event channel closed
```

KMS works end-to-end. Merge runs and publishes. The subscription handler **does** receive the event. `RunSelection` returns empty result + a CBOR error. The handler drops the event silently and the subscriber times out.

### Primary: encrypted field bytes hit a CBOR decoder that doesn't know about encryption

`internal/db/fetcher/encoded_doc.go:58` — `encProperty.Decode()`:

```go
func (e encProperty) Decode() (any, error) {
    var val any
    err := cbor.Unmarshal(e.Raw, &val)
    if err != nil {
        return nil, err
    }
    return core.NormalizeFieldValue(e.Desc, val)
}
```

`e.Raw` is the field-value bytes pulled by the version fetcher used inside the planner. When the field's CRDT delta is encrypted, those bytes are an encryption envelope (CBOR header + ciphertext), and `cbor.Unmarshal` fails with `cbor: <N> bytes of extraneous data starting at index <K>`.

The merge-side decryption pipeline (KMS → encryption-block → key) successfully retrieves the per-doc key and stores it. The issue is that the version-fetcher path used by `planner.RunSelection` from `db.handleSubscription` does not consult that path; it reads the stored bytes directly and feeds them to the CBOR decoder.

### Secondary: subscription handler swallows the error

`internal/db/subscriptions.go` (around line 94 onward):

```go
result, err := p.RunSelection(ctx, s)
if err == nil && len(result) == 0 {
    txn.Discard()
    continue // Don't send anything back to the client if the request yields an empty dataset.
}

res := client.GQLResult{}

for op, data := range result {
    resultSlice, ok := data.([]map[string]any)
    if !ok {
        res.Errors = append(res.Errors, ErrBadDocsResultType)
    }
    if len(resultSlice) == 0 {
        delete(result, op)
    }
}

// now that weve filtered empty result sets, lets recheck
if len(result) == 0 {
    txn.Discard()
    continue
}
```

The first guard requires `err == nil` to drop on empty. But the second guard at the bottom drops on `len(result) == 0` **regardless of err**. When `RunSelection` returns `(empty, err)` — exactly what the encrypted case produces — the second check fires and the error never reaches the subscriber.

If the error were forwarded to `res.Errors` and delivered, the test would have failed loudly with the CBOR error instead of timing out.

## Suggested fix direction

1. **Primary** — make the version-fetcher path decryption-aware. The planner-side read needs to go through the same decryption layer the merge processor uses (or through a shared helper that resolves the per-doc key from the encryption-block lookup before CBOR-decoding `e.Raw`). Likely involves wiring an `encryption.Service`-equivalent into the fetcher used by `planner.RunSelection`.
2. **Secondary** — in `subscriptions.go`, treat `(empty, err)` distinctly: forward the error to the subscriber as `res.Errors` and deliver, instead of dropping. This won't make the subscription succeed but will make the failure observable.

## Affected scope

Any subscriber peer that uses GraphQL subscriptions on a collection where incoming docs are encrypted. The encrypted-docs multiplier surfaces this on every test that combines `&action.SubscriptionRequest{...}` with multi-peer P2P sync.

Workaround for now: tests that combine subscriptions + encrypted P2P sync opt out via `MultiplierExcludes: []string{multiplier.EncryptedDocs}`.

## Related

- #4779 (multi-peer KMS first-responder bug — separate issue, separate code path)
- PR #4778 (KMS-NAC collection lookup — unrelated; landed)

islamaliev:
## Relevant issue(s)

Resolves #4787

## Description

A subscriber peer running a GraphQL subscription on a collection that received an encrypted document was timing out instead of delivering the update. The merge happened correctly and the update event fired, but the planner selection that runs in response came back empty with a CBOR decode error, and the subscription handler dropped the event silently.

The root cause is in the version-replay path used by subscriptions. When a subscription receives an event, the planner switches to a CID-filtered selection backed by VersionedFetcher, which walks the DAG up to the target CID and replays each block's delta into a transient datastore. The live merge processor decrypts blocks before replaying them; the version-replay loop didn't, so the transient store ended up holding ciphertext bytes and the field decoder downstream choked. The fix lifts the load+decrypt orchestration into a single shared helper in `internal/core/block` and routes both replay paths through it, so the merge path and the version-replay path stay in lockstep by construction.

The handler dropping the error was a separate amplifier — the second empty-result guard didn't distinguish a no-rows result from an error result. Fixed there as well, so any future error from selection on an update event reaches the subscriber instead of being swallowed.

## How has this been tested?

A new integration test exercises the bug directly: two peers, KMS active, encrypted doc add, P2P collection subscription plus an active GraphQL subscription on the subscriber. Fails on develop with the documented data-stream timeout; passes with the fix. Verified under both `DEFRA_CLIENT_TYPE=go` and `DEFRA_CLIENT_TYPE=http`. Full `tests/integration/encryption/...` and `tests/integration/net/...` suites stay green.

- macOS


fredcarle (approved):
LGTM

PR #4793, opened 2026-05-08
feat: JS client p2p

### LINKED ISSUE #3838: Browser P2P implementation
Implement a browser compatible network stack.

nasdf:
## Relevant issue(s)

Resolves #3838

## Description

This PR enables p2p for the JS client.

Other notable Changes:
- JS client methods now parse all client options.
- HTTP API is disabled by default for JS builds.
- Replaced go-libp2p with sourcenetwork fork that supports wasm builds

## Tasks

## How has this been tested?

Integration tests and manual testing in the browser.

P2P integration tests will be enabled once we find a solution for relaying.

- MacOS


js/client_tx.go
```diff
@@ -40,7 +41,6 @@ func newTransaction(txn client.Txn, txns *sync.Map) js.Value {
 		"discard":                    goji.Async(wrapper.discard),
 		"addCollection":              goji.Async(wrapper.addCollection),
 		"patchCollection":            goji.Async(wrapper.patchCollection),
-		"deleteCollection":           goji.Async(wrapper.deleteCollection),
 		"setActiveCollectionVersion": goji.Async(wrapper.setActiveCollectionVersion),
```
fredcarle, line 45: question: Why was this removed?

js/js.go
```diff
@@ -30,28 +31,20 @@ func SetGlobal() {
 }
 
 // open creates a new DB client and returns it wrapped in a JS object.
-// acpType is optional and can be:
-// - "sourcehub" to use SourceHub ACP
-// - anything else (including undefined/null) to use Local ACP
+//
+// The first argument is an optional object containing a field for each node.Option.
 func open(this js.Value, args []js.Value) (js.Value, error) {
```
fredcarle, line 36: question: It seems that it is no longer possible to select ACP type. Is that indeed the case?

js/utils.go
```diff
@@ -13,7 +13,6 @@
 package js
```
AndrewSisley, line 13: praise: Thanks for cleaning up this file Keenan - it and the calling code looks a lot nicer.

js/utils.go
```diff
@@ -101,7 +137,7 @@ func contextIdentityArg(value js.Value) (immutable.Option[acpIdentity.Identity],
 		}
 		return immutable.Some[acpIdentity.Identity](identity), nil
 	}
-	ident := value.Get("identity")
+	ident := opts.Get(identityProp)
```
fredcarle: question: Don't we need to also handle `nodeIdentity` and `fullNodeIdentity`?

js/js.go
nasdf: This function now respects all of the node options. So you'd pass ACP type the same way the embedded client does.

fredcarle: ~~Right. But following the code path, it seems like it will only handle `identity` and `fullIdentity` and not the node identity types.~~

fredcarle: Sorry I mixed up the comments.

fredcarle: So it will be taken from the `goji.UnmarshalJS` call in `parseNodeOptions`?

js/client_tx.go
nasdf: good catch! I've added it back

js/js.go
nasdf: So this did turn out to be an issue. There's now parsing logic to handle identity or fullIdentity for the node options.

js/utils.go
nasdf: done

fredcarle (approved):
LGTM

PR #4794, opened 2026-05-08
feat: Support querying with multiple cids

### LINKED ISSUE #4303: feature request: for commit query, support cid list constraint
```gql
query { _commit(cid: [ "..", ".." ]) { .. } }
```

Right now, cidID: ".." only is supported. Requesting to be able to pass in multiple cid values.

This is basically like `filter: { cid: { _in: [ "..", ".." ] } }`. Similar rational as #4302. If I want to pull out some commit info from a pre-determined list of commits, by their cid, the top-level constraint is a nice-to-have sugar for the `filter` form.

### LINKED ISSUE #4304: feature request: for regular query, support cid list constraint
```gql
query { CredentialLastUsage(cid: [ "..", ".." ]) { .. } }
```

Similar idea to #4303, but *this one* is for the regular query on any collection (not just on _commits). Same use-case/rationale as #4302. If I've pre-determined all the composite commits that I want to pull my document fields from, would like this in one query result set instead of N separate queries.

AndrewSisley:
## Relevant issue(s)

Resolves #4304
Resolves #4303

## Description

Support querying with multiple cids, both for versioned collection queries, and via the commits api.

Also tests/documents existing behaviour of the `docID` collection query param.

### DIFF
--- a/internal/db/fetcher/versioned_multi.go
+++ b/internal/db/fetcher/versioned_multi.go
@@ -1,0 +1,172 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package fetcher
+
+import (
+	"context"
+
+	"github.com/sourcenetwork/defradb/acp/dac"
+	acpIdentity "github.com/sourcenetwork/defradb/acp/identity"
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/core"
+	"github.com/sourcenetwork/defradb/internal/datastore"
+	acpDB "github.com/sourcenetwork/defradb/internal/db/acp"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	"github.com/sourcenetwork/defradb/internal/planner/mapper"
+	"github.com/sourcenetwork/immutable"
+)
+
+// MultiVersioned is a short term solution that allows multiple prefixes to be
+// fetched via the VersionedFetcher(s).
+//
+// It and the VersionedFetcher need  a rework at some point, migrating them to the
+// `fetcher` interface, and resolving a few other their inefficiencies (such as the
+// one that this type represents).
+type MultiVersioned struct {
+	children     []*VersionedFetcher
+	currentChild int
+
+	ctx         context.Context
+	identity    immutable.Option[acpIdentity.Identity]
+	txn         datastore.Txn
+	nodeACP     acpDB.NACInfo
+	documentACP immutable.Option[dac.DocumentACP]
+	index       immutable.Option[client.IndexDescription]
+	col         client.Collection
+	fields      []client.CollectionFieldDescription
+	filter      *mapper.Filter
+	ordering    []mapper.OrderCondition
+	docmapper   *core.DocumentMapping
+	showDeleted bool
+}
+
+var _ Fetcher = (*MultiVersioned)(nil)
+
+func (f *MultiVersioned) Init(
+	ctx context.Context,
+	identity immutable.Option[acpIdentity.Identity],
+	txn datastore.Txn,
+	nodeACP acpDB.NACInfo,
+	documentACP immutable.Option[dac.DocumentACP],
+	index immutable.Option[client.IndexDescription],
+	col client.Collection,
+	fields []client.CollectionFieldDescription,
+	filter *mapper.Filter,
+	ordering []mapper.OrderCondition,
+	docmapper *core.DocumentMapping,
+	showDeleted bool,
+) error {
+	f.ctx = ctx
+	f.identity = identity
+	f.txn = txn
+	f.nodeACP = nodeACP
+	f.documentACP = documentACP
+	f.index = index
+	f.col = col
+	f.fields = fields
+	f.filter = filter
+	f.ordering = ordering
+	f.docmapper = docmapper
+	f.showDeleted = showDeleted
+
+	return nil
+}
+
+func (f *MultiVersioned) Start(ctx context.Context, prefixes ...keys.Walkable) error {
+	// Deduplicate the prefixes so that the child fetchers do not return duplicated results.
+	// This is in keeping with the behaviour of other query params, such as docIDs.
+	uniquePrefixStrings := map[string]struct{}{}
+	uniquePrefixes := []keys.Walkable{}
+	for _, prefix := range prefixes {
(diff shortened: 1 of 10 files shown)


getify:
does this also address #4303?

AndrewSisley:
> does this also address #4303?

Yes, sorry I didn't spot that they were two different tickets - thanks :)

tests/integration/query/commits/with_cid_test.go
```diff
@@ -196,7 +196,130 @@ func TestQueryCommitsWithUnknownCid(t *testing.T) {
...
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// This test documents undesirable behaviour and should be changed as part of:
+// https://github.com/sourcenetwork/defradb/issues/2469
+func TestQueryCommits_MultipleCidsDifferentDocs_NoAccessToSecondCid(t *testing.T) {
```
shahzadlone, line 242: nitpick: The name implies the second cid is hidden, but the assertion returns both cids, should now be fixed by my work that just merged.

tests/integration/query/simple/with_cid_test.go
```diff
@@ -282,19 +282,134 @@ func TestQuerySimple_MultipleCIDs(t *testing.T) {
...
+					"name": "Johnnn"
 				}`,
 			},
+			&action.AddDoc{
+				DocMap: map[string]any{
+					"name": "Fred",
+				},
+			},
 			&action.Request{
+				// Query with the cid for the iniitial version of `John`, and the updated version.
```
shahzadlone: typo: 
```suggestion
				// Query with the cid for the initial version of `John`, and the updated version.
```

tests/integration/query/commits/with_cid_test.go
AndrewSisley: Ah sweet - thanks for the reminder (and the fix).  I'll make sure I update the comments when I rebase - they might have been missed otherwise.

- [x] Make sure test docs updated once rebased

internal/planner/commit.go
```diff
@@ -201,17 +202,14 @@ func (n *dagScanNode) Next() (bool, error) {
...
-
-		cid, err := cid.Parse(n.commitSelect.Cids.Value()[0])
+	} else if n.commitSelect.Cids.HasValue() && n.currentSelectCidIndex < len(n.commitSelect.Cids.Value()) {
+		cid, err := cid.Parse(n.commitSelect.Cids.Value()[n.currentSelectCidIndex])
 		if err != nil {
 			return false, err
 		}
 
 		currentCid = &cid
+		n.currentSelectCidIndex++
```
shahzadlone, line 214: question: `n.depthVisited` is reset to 0 between heads at line 221 (for the HeadFetcher path), but the new index-based cid loop does not reset it between cids, so wouldn't with the depth gate at lines 283-284, a query like `_commits(cid: [a, b], depth: 2)` burn its depth budget on cid `a`'s sub-DAG and may starve cid `b` of any traversal? Maybe test this case (I didn't spot this test case coverage lmk incase I missed)?

internal/db/fetcher/versioned_multi.go
```diff
@@ -0,0 +1,172 @@
...
+	acpDB "github.com/sourcenetwork/defradb/internal/db/acp"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	"github.com/sourcenetwork/defradb/internal/planner/mapper"
+	"github.com/sourcenetwork/immutable"
+)
+
+// MultiVersioned is a short term solution that allows multiple prefixes to be
+// fetched via the VersionedFetcher(s).
+//
+// It and the VersionedFetcher need  a rework at some point, migrating them to the
```
shahzadlone, line 31: typo:

```suggestion
// It and the VersionedFetcher needs a rework at some point, migrating them to the
```

todo: Please mark with a todo linked with an issue as it's mentioned that a rework is needed, so essentially is a todo.

internal/db/fetcher/versioned_multi.go
```diff
@@ -0,0 +1,172 @@
...
+
+	doc, execInfo, err := f.children[f.currentChild].FetchNext(ctx)
+	if err != nil {
+		return nil, ExecInfo{}, err
+	}
+
+	if doc == nil {
+		f.currentChild++
+		return f.FetchNext(ctx)
+	}
```
shahzadlone, line 150: nitpick: recursing into `f.FetchNext` on an exhausted child grows the stack by one per empty child. Bounded by user-supplied cid count so not a practical risk, but a for `f.currentChild < len(f.children) { ... }` loop is (perhaps subjectively) easier to read and saferish?

internal/db/fetcher/versioned_multi.go
```diff
@@ -0,0 +1,172 @@
...
+	f.fields = fields
+	f.filter = filter
+	f.ordering = ordering
+	f.docmapper = docmapper
+	f.showDeleted = showDeleted
+
+	return nil
+}
+
+func (f *MultiVersioned) Start(ctx context.Context, prefixes ...keys.Walkable) error {
```
shahzadlone, line 84: nitpick: Init only stores/sets its inputs into fields; the actual child setup happens in Start (line 84) once prefixes are known. This is fine but unusual relative to the rest of the fetcher family, no? perhaps a  one-line doc comment on Init saying "actual fetcher construction happens in Start once prefixes are known" would save a reader the trip through Start to figure it out.

internal/planner/commit.go
AndrewSisley: That's a nice catch - thanks Shahzad.  Will add a test and and adjust prod code if needed (probably needed).

- [x] Add depth test

internal/db/fetcher/versioned_multi.go
AndrewSisley: not a problem

- [x] make a todo

internal/db/fetcher/versioned_multi.go
AndrewSisley: Going to leave this one as-is

internal/db/fetcher/versioned_multi.go
AndrewSisley: The versioned fetcher is unusual - it is the only fetcher that has Start/Init and IMO its a horrible pattern that needs to die.

Going to leave this as-is, uncommented, as it should be very clear to anyone reading the internals of these functions, and irrelevant to any callers of these functions.

internal/planner/commit.go
AndrewSisley: Leaving as-is, depth is not relevant when querying by cid, as children of found cids are not returned (is where depth kicks in).

internal/db/fetcher/versioned_multi.go
AndrewSisley: I have removed the comment, as I felt that a refactor/performance ticket lingering indefinitely in the backlog would not be useful, and the comment itself was not very useful.  This also resolves the typo.

shahzadlone (approved):
LGTM

PR #4796, opened 2026-05-08
test(i): Exclude LevelDB from multi-txn test

### LINKED ISSUE #4795: TestColVersionUpdateRemoveCollection_CollectionMayBeRedeclaredAndUsedByTxn doesn't work with level db
As it has two open txns at the same time.  The test needs to exclude leveldb.

AndrewSisley:
## Relevant issue(s)

Resolves #4795

## Description

Excludes LevelDB from some newer tests that open multiple transactions at the same time.


nasdf (approved):
Thanks for fixing this!

PR #4800, opened 2026-05-09
fix: Reject nested list SDL with clear error

### LINKED ISSUE #4788: Creation of collection with nested list fails with 500 and no error
Attempt at creating a collection having a property with nested list fails with error `500` but no error message. I don't know whether this should be supported or not, but it would be good to fail emitting an error message.
```
$ defradb client collection add '
  type User {
        name: [[String]]
        age: Int
        verified: Boolean
        points: Float
  }
'
Error: failed to add collection: 500:
```

SAY-5:
Nested list SDL currently panics the collection parser via an unchecked list-to-named type assertion, which bubbles out as a bare 500 from `client collection add`.

Reject nested list field types explicitly during SDL parsing, return a normal validation error instead, and cover it with schema and CLI regression tests. Fixes #4788.


jsimnz (approved):
LGTM. Thanks

PR #4801, opened 2026-05-09
test: Encrypt docs in replicator test

### LINKED ISSUE #4783: Searchable encryption test missing encrypted documents
In the SE replicator test https://github.com/sourcenetwork/defradb/blob/develop/tests/integration/searchable_encryption/replicator_test.go#L54 , documents are not encrypted when created, thus making the use of SE pointless. 

Compare it for example with the peer_query test https://github.com/sourcenetwork/defradb/blob/develop/tests/integration/searchable_encryption/peer_query_test.go#L50 , which uses `IsDocEncrypted: true` when creating documents.

SAY-5:
## Relevant issue(s)

Resolves #4783

## Description

The searchable-encryption replicator regression test currently adds plain documents, so it never actually exercises encrypted document creation on the source side.

Set `IsDocEncrypted: true` on the two inserted docs so the test covers the intended SE replication path.


AndrewSisley (approved):
LGTM, thanks for this!

PR #4805, opened 2026-05-10
fix: Clean schema heads on collection deletion/patch

### LINKED ISSUE #4803: Delete collection and Patch collection (with remove op) leaving missing heads
Delete collection and Patch collection (with remove op) leaving missing heads that causes `AddCollection` with same name to error out

shahzadlone:
## Relevant issue(s)
- Resolves #4803

## Description
Fix `DeleteCollection` (and `PatchCollection`'s op: remove, which goes through the same code path) where the schema heads were left dangling after the underlying blocks had been removed.

Note: I made sure that the cleanup only happens when the final version is deleted (or if deleting collection where you want all versions to be removed. i.e. without the active version only flag / default case)

### DIFF
--- a/internal/db/collection_define.go
+++ b/internal/db/collection_define.go
@@ -21,6 +21,7 @@ import (
 	jsonpatch "github.com/evanphx/json-patch/v5"
 	"github.com/ipfs/go-cid"
 
+	"github.com/sourcenetwork/corekv"
 	"github.com/sourcenetwork/immutable"
 	"github.com/sourcenetwork/lens/host-go/config/model"
 
@@ -28,9 +29,11 @@ import (
 
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/client/request"
+	"github.com/sourcenetwork/defradb/errors"
 	"github.com/sourcenetwork/defradb/internal/core"
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/db/description"
+	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
 func (db *DB) addCollections(
@@ -681,7 +684,7 @@ func (db *DB) deleteCollectionVersion(
 		return err
 	}
 
-	return nil
+	return deleteCollectionDefinitionHeads(ctx, db.collectionRepository, version)
 }
 
 func (db *DB) validateCollectionDoesNotHaveHigherVersion(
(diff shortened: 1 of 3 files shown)


internal/db/collection_define.go
```diff
@@ -742,6 +745,103 @@ func deleteCollectionBlocks(
 	return nil
 }
 
+// deleteCollectionSchemaHeads removes the schema-level headstore entries for
```
AndrewSisley: todo: Please don't use the word `schema` anywhere here - is is very confusing, and there should be nothing GQL-schema related at this lvel

internal/db/collection_define.go
```diff
@@ -742,6 +745,103 @@ func deleteCollectionBlocks(
...
+		// Other versions of this collection still exist; their heads must stay.
+		return nil
+	}
+
+	txn := datastore.CtxMustGetTxn(ctx)
+	headstore := txn.Headstore()
+
+	// A trailing '/' is appended so that prefix iteration matches only this
+	// collection name and not other names that happen to start with it (e.g.
+	// "User" must not pick up "Users").
```
AndrewSisley: praise: Thank you for this, and the comment - easy to miss

internal/db/collection_define.go
```diff
@@ -742,6 +745,103 @@ func deleteCollectionBlocks(
...
+	fieldPrefix := []byte(
+		keys.HeadstoreFieldDefinition{CollectionName: version.Name}.ToString() + "/",
```
AndrewSisley: todo: There is a conflict here, and with the `HeadstoreCollectionDefinition` deletes below.

The function `deleteCollectionSchemaHeads` is deleting by versionID, this is fine, there can only be one.

However, you are then fetching collections by `CollectionID`, and then deleting by `Collection.Name`.

Technically, at this key-level, there is nothing that prevents a single name being shared by multiple CollectionIDs - if that ever becomes true, then we'll have very annoying to find bug here.

More importantly, there can be multiple heads for a single collection.  The function `deleteCollectionSchemaHeads` is deleting by versionID, but it will delete all heads for that collection-name.  This is sort of protected by the current internal implementation of `description.GetCollectionsByCollectionID`, but not by its function signature.

It's also wasteful in terms of CPU and ambiguity - we are deleting a single collection version here - the entire key set is already known to us.  We have no reason to query it.

Please change this function so that is constructs the keys from the information held on the Collection/field version object passed in as a parameter, and then deletes them via simple `txn.Headstore.Delete(key)` calls, without (re)fetching data.

AndrewSisley (changes requested):
Thanks for picking this up!

Please simplify the code a little bit before mere, a lot of what is being done appears to be unnecessary.

internal/db/collection_define.go
shahzadlone: - [x] Done

internal/db/collection_define.go
shahzadlone: - [x] Done in commit : `PR(REVIEW): Delete collection-definition heads by exact key` 

Thanks @AndrewSisley! Your approach was genuinely better. Three real wins, lemme think/summarize out loud:

  1. Structural invariant vs implementation accident. Your direct-key approach makes "don't touch other versions' heads" a structural property, the key we build includes this version's CID, so if the head points at another version's CID the `Has` check naturally returns false. The invariant is in the code, not in another package.
  
  2. Name-collision robustness. If two CollectionIDs ever shared a name (nothing at the key level prevents this), my prefix-by-name delete would have silently wiped both. The new keys include the CID component, so cross-CollectionID collisions are impossible.
    
  3. No wasteful query/scan. We already have the version on hand, querying for siblings and then scanning the headstore was doing work we didn't need.

Inaddition to that, I didn't realize until I worked through the multi-version scenarios, under the old gate-based approach, `ActiveOnly` delete of the leaf version of a multi-version collection would leave a dangling head pointing at a deleted block (the gate skipped cleanup because siblings remained, but the head still referenced the now-gone leaf). The new approach removes that head precisely because the key matches. 

one thing I added was the `Has`-gate around Delete, I did remember seeing for `corekv` that the Delete on a missing key was undefined? I am not much well versed with the `corekv` stuff just wanted to make sure our delete legitimately is a no-op (if head points at a different version's CID).

AndrewSisley (approved):
LGTM, thanks Shahzad :)

Issue #4452, opened 2026-01-30
Add test multiplier for encrypted docs


AndrewSisley:
Encrypting documents/collections is a complexity multiplier, and should be tested as such.

Issue #4393, opened 2026-01-22
Migrate MutationType multiplier into testo multiplier


AndrewSisley:
This issue is referenced by a code-comment.

PR #4809, opened 2026-05-11
feat: Gate commit queries with DAC

### LINKED ISSUE #2469: DAC ACP <> DAG / Commit Operations
Parent Issue #1738 

This issue is to make sure we properly test access control (ACP) in combination to ensure there aren't any bugs, there might be some missing implementation that might need to be done or it might just work.

shahzadlone:
Relevant issue(s):
- Resolves #2469 

## Description
The `_commits` query was bypassing the document access control (DAC) layer, so any client (signed or anonymous) could read the commit history of documents they had no read access to. The deltas inside those commits expose the same field values DAC is meant to gate.

Plugged the gap with a single chokepoint inside `dagScanNode.Next()`: after each block is decoded, resolve its collection by `CollectionVersionID` and check `DocumentReadPerm` against the caller's identity. Denied commits are marked visited and skipped, so the same cid is not re-checked when it resurfaces via a `links` or `heads` sub-scan. Publc collections, public
docs, and bypass-DAC identities continue to fall through the existing short-circuits in the access check.

## For reviewers:
- The first commit writes tests we want to ensure work after the feature, I commented out the tests in the first commit that were failing due to DAC not gating commit queries, they are then re-enabled after the implementation, to test they pass now.
- Commit by commit review should be clean.

## How has this been tested?
- Added tests under `tests/integration/acp/dac/commits/` folder.

### DIFF
--- a/internal/planner/commit.go
+++ b/internal/planner/commit.go
@@ -49,12 +49,6 @@ type dagScanNode struct {
 	linksScanNodes []*dagScanNode
 	headsScanNodes []*dagScanNode
 
-	// cachedCols memoizes the per-versionID collection lookups used by both
-	// the DAC access check and dagBlockToNodeDoc. A `_commits` query may
-	// span docs from different collection versions, so the cache is keyed
-	// by collection version ID.
-	cachedCols map[string]client.Collection
-
 	execInfo dagScanExecInfo
 }
 
@@ -101,32 +95,6 @@ func (p *Planner) CommitSelect(commitSelect *mapper.CommitSelect) (planNode, err
 
 func (n *dagScanNode) Kind() string {
 	return "dagScanNode"
-}
-
-// getCollectionByVersionID resolves and caches a collection by its
-// CollectionVersionID. The lookup result is shared between the DAC access
-// check and dagBlockToNodeDoc.
-func (n *dagScanNode) getCollectionByVersionID(versionID string) (client.Collection, error) {
-	if c, ok := n.cachedCols[versionID]; ok {
-		return c, nil
-	}
-
-	cols, err := n.planner.db.GetCollections(
-		n.planner.ctx,
-		options.GetCollections().SetGetInactive(true).SetVersionID(versionID),
-	)
-	if err != nil {
-		return nil, err
-	}
-	if len(cols) == 0 {
-		return nil, client.NewErrCollectionNotFoundForCollectionVersion(versionID)
-	}
-
-	if n.cachedCols == nil {
-		n.cachedCols = make(map[string]client.Collection)
-	}
-	n.cachedCols[versionID] = cols[0]
-	return cols[0], nil
 }
 
 func (n *dagScanNode) Init() error {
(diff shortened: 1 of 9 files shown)


internal/planner/commit.go
```diff
@@ -47,6 +49,12 @@ type dagScanNode struct {
 	linksScanNodes []*dagScanNode
 	headsScanNodes []*dagScanNode
 
+	// cachedCols memoizes the per-versionID collection lookups used by both
+	// the DAC access check and dagBlockToNodeDoc. A `_commits` query may
+	// span docs from different collection versions, so the cache is keyed
+	// by collection version ID.
+	cachedCols map[string]client.Collection
```
AndrewSisley: suggestion: Please don't bother with this.  `GetCollections` and its friends already caches the collections and is designed so that we don't need to do this everywhere else.

Given the hopefully very small absolute savings to be had here, and the reasonably low importance of this code, I'd really much prefer if we optimise for simplicity and ease of maintenance instead of adding in unmeasured optimizations.

tests/integration/acp/dac/commits/basic_test.go
```diff
@@ -0,0 +1,465 @@
+// Copyright 2026 Democratized Data Foundation
```
AndrewSisley, line 1: thought: This is a very hard to read and maintain way of writing test, and is very inefficient IMO.

In my opinion, adding 1600 lines of test code for the sake of 30 lines of production code is unreasonable and quite a bad way of doing things.

We really need to be using the multiplier system for this (and other DAC stuff).

If we weren't really push for time due to v1, this comment would probably be a `todo`.

internal/planner/commit.go
shahzadlone: Thanks for flagging! I am happy to remove this micro optimization.

AndrewSisley (approved):
I have quite a strong preference for the collection caching to be removed before merge please.

tests/integration/acp/dac/commits/basic_test.go
shahzadlone: I think I am somewhat on the same page as you. Is there a tracking issue already? I can link there that the DAC and NAC specific tests can benefit from the multiplier system.

internal/planner/commit.go
shahzadlone: - [x] Remove cache optimization, done in commit: `PR(REVIEW): Remove per-versionID collection cache in commits planner`

tests/integration/acp/dac/commits/basic_test.go
AndrewSisley: Nothing comes to mind, but I haven't looked

shahzadlone: I see these:
- https://github.com/sourcenetwork/defradb/issues/4452
- https://github.com/sourcenetwork/defradb/issues/4393

Not sure if it is the same concern that they are addressing?

PR #4812, opened 2026-05-11
fix: Multi-peer KMS encryption-key fetch race

### LINKED ISSUE #4779: bug(p2p/kms): multi-peer encryption-key fetch loses to first non-holding responder
## Summary

When more than one peer is subscribed to the `encryption` pubsub topic and a node requests an encryption key for an encrypted block, the *first* peer to respond wins — even if that peer does not actually hold the key. Subsequent responses (including the one from the peer that *does* hold the key) are discarded. The requester then decodes the empty wire reply as `EOF` and the DAG sync fails with:

```
failed to retrieve encryption key during DAG sync: EOF
```

This makes encrypted DAG sync unreliable on any topology with three or more peers where the key is not held by every peer.

## Reproduction

Run the multi-peer branchable-collection sync tests under the encrypted-docs multiplier:

```
DEFRA_CLIENT_TYPE=go DEFRA_MULTIPLIERS=encrypted-docs \
  go test ./tests/integration/net/sync/branchable_collection/ \
  -run TestBranchableCollectionSync_WithMultipleDocsInComplexLinkedNetwork_ShouldSyncAll
```

Topology: 5 peers in a tree (Node 0 ── 1 ── 2 / Node 0 ── 3 ── 4). Documents are created on each node. After connecting, `SyncBranchableCollection` is invoked; the encrypted-docs multiplier flips `IsDocEncrypted = true` on every `AddDoc`, so every collection-level commit block carries an encrypted payload.

The test fails with `failed to retrieve encryption key during DAG sync: EOF`.

The same shape repeats across every test in `tests/integration/net/sync/branchable_collection/` that has 3+ peers when run with the multiplier active.

## Trace

Filtered `[kms]` trace from a failing run, requesting key for link `bafyrei…25ny` (peer IDs abbreviated):

```
[kms] GetKeys called peerID=A nCids=1
[kms] publishing to topic=encryption nCids=1
[kms] waiting on respChan...
    [kms] handleRequestFromPeer me=B from=A
        [kms] encStore.get err: ipld: could not find bafyrei…25ny     ← B does not have the key
    [kms] handleRequestFromPeer me=C from=A
        [kms] encStore.get err: ipld: could not find bafyrei…25ny     ← C does not have the key
    [kms] handleRequestFromPeer me=D from=A
        [kms] fetched encBlock DocID=\"bae-897c…\" keyLen=32           ← D has the key
        [kms] handleRequestFromPeer ok nBlocks=1
    [kms] handleRequestFromPeer me=E from=A
        [kms] encStore.get err: ipld: could not find bafyrei…25ny     ← E does not have the key
[kms] respChan recv: ok=true dataLen=0 from=B                          ← first responder (B) wins; D's reply ignored
ERR kms Failed to unmarshal encryption key response err=EOF
```

`dataLen=0` is the failure source: `s.encStore.get` returns an error for not-found, `getEncryptionKeysLocally` short-circuits with that error, `handleRequestFromPeer` returns `(nil, err)`, and pubsub propagates that as a zero-byte reply.

## Root cause

Two contributing issues, both in `internal/kms/pubsub.go`:

1. **First-responder wins.** `s.pubsub.PublishToTopic(ctx, pubsubTopic, data, false)` is called with `withMultiResponse = false`, so only one response is ever consumed from `respChan`. With multiple subscribers, the timing-winner is essentially random — and is statistically likely to be a peer that doesn't hold the key.

2. **Not-found is conflated with error.** When a peer doesn't have the requested key in its `encStore`, `s.encStore.get(ctx, link)` returns an IPLD \"could not find\" error rather than `(nil, nil)`. `getEncryptionKeysLocally` short-circuits the loop on err, so `tryGenEncryptionKeyLocally` returns `(nil, err)` and the wire reply is empty. Even if multi-response were enabled, the not-having peers would still be sending empty (error) replies instead of well-formed `fetchEncryptionKeyReply{Blocks: nil}` records that the requester could distinguish from \"corrupted\" replies.

## Suggested fix direction

1. In `internal/kms/enc_store.go::ipldEncStorage.get`, treat \"not found\" as `(nil, nil)` (the caller already has a `if encBlock == nil { continue }` branch — the design clearly intends not-found to be non-fatal).
2. In `internal/kms/pubsub.go::requestEncryptionKeyFromPeers`, switch to `withMultiResponse = true`, then in `handleFetchEncryptionKeyResponse` keep consuming responses until one carries `len(Blocks) > 0` (or a deadline fires). Need to be careful to `defer close(result)` correctly under the new loop, and to pick a sane timeout for \"no peer has the key\".

## Affected scope

Any encrypted DAG sync where the requester has 2+ peers subscribed to the pubsub `encryption` topic and the key is not held by every one of them. With `IsDocEncrypted` enabled by default (e.g. the encrypted-docs multiplier), this is essentially any 3+-peer encrypted test.

## Out of scope here

This issue is distinct from the KMS-NAC collection lookup bug fixed in #4778 / commit `fdbb5b4`. That fix unblocks the *2-peer* encrypted+NAC sync path; the multi-peer first-responder issue described here is what remains and was deliberately scoped out of that PR.

islamaliev:
## Relevant issue(s)

Resolves #4779

## Description

When more than one peer is subscribed to the KMS \`encryption\` pubsub topic, the requester used to consume only the first reply that arrived. The fastest reply often came from a peer that didn't actually hold the key, so the requester decoded the empty/error wire reply as \`EOF\` and DAG sync aborted with \`failed to retrieve encryption key during DAG sync: EOF\`.

Two fixes. Peers without the key now reply empty rather than error: \`ipldEncStorage.get\` treats \`ipld.ErrNotFound\` as \`(nil, nil)\` so the existing \`if encBlock == nil { continue }\` branch in \`getEncryptionKeysLocally\` is reachable. And the requester now publishes with \`withMultiResponse=true\` and consumes replies in a loop until one carries blocks or a 5s deadline fires (matching the \`sync_branchable_col\` / \`sync_doc\` timeout pattern). When the deadline fires with no holder ever responding, a typed \`ErrNoPeerSuppliedEncryptionKey\` is published on the result channel so \`loadBlockLinks\` has a deterministic error to propagate via \`NewErrRetrieveEncKey\` instead of the historical decoder-EOF.

The split between the two pieces is deliberate: without the enc-store change, a peer that doesn't have the key still replies with an error wire payload, so the requester can't tell that apart from a real fault and discards valid empty replies even with multi-response on. Without the multi-response change, the requester only ever sees one reply, so even well-formed empty replies still let a non-holder win the race.

## Tasks

## How has this been tested?

New integration test \`tests/integration/encryption/peer_multipeer_test.go::TestDocEncryptionPeer_MultiplePeersOnEncryptionTopic_ShouldFetchKeyFromHolder\` reproduces the race deterministically in CI by setting up a topology where the requester connects to two KMS-active "bystander" peers (subscribed to the encryption topic but not to the collection, so they receive fetch-key requests but never hold a key) ahead of the actual holder. Pre-fix the test failed ~9/10 runs with a ~30s timeout; with the fix it passes ~10/10 in ~0.2s.

Full encryption suite stays green under both \`DEFRA_CLIENT_TYPE=go\` and \`DEFRA_CLIENT_TYPE=http\`. The previously-merged dependency PR's reproducer (\`TestDocEncryptionNAC_SyncBranchableCollection_AuthorizedIdentity_AllowAccess\`) and the rest of the NAC encryption suite also remain green.

- MacOS


internal/kms/pubsub.go
```diff
@@ -212,20 +215,67 @@ type fetchEncryptionKeyReply struct {
...
+			}
+			result <- encryption.Result{Items: items}
+			return
+
+		case <-waitCtx.Done():
+			result <- encryption.Result{Error: ErrNoPeerSuppliedEncryptionKey}
+			return
+		}
+	}
+}
```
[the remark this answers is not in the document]
fredcarle: go-p2p closed the channel on `ctx.Done()`.

internal/kms/pubsub.go
```diff
@@ -212,20 +215,67 @@ type fetchEncryptionKeyReply struct {
...
+			if !ok {
+				result <- encryption.Result{Error: ErrNoPeerSuppliedEncryptionKey}
+				return
+			}
+			items, ok := s.tryHandleFetchEncryptionKeyResponse(resp, req, privateKey)
+			if !ok {
+				continue
+			}
+			result <- encryption.Result{Items: items}
+			return
```
fredcarle, line 254: todo: There is a goroutine leak situation happening here if we are expecting results from more than one peer. The `respChan` is an unbuffered channel so if the first response is valid and we return, the `go-p2p` side will block waiting for the channel to be consumed. I suggest you pass a cancellable context to `PublishToTopic` and `defer(cancel)` at the beginning of this function.

fredcarle (approved):
Approving with one todo to resolve before merge.

internal/kms/pubsub.go
islamaliev: nice catch. Fixed

PR #4827, opened 2026-05-21
fix: Subscription handler iterator leak on join collections

### LINKED ISSUE #4826: Subscription panic on `Unclosed iterator at time of Txn.Discard.` is back
Ran into this against the last fix (the one that added the recover so the process stays up — that part is great, the node doesn't die anymore). Problem is the underlying panic is still firing on every inbound event that lands on a live subscription. The recover catches it, but the goroutine returns right after, so the SSE stream closes silently with zero bytes and the client never sees the doc.

Took a while to narrow but it's consistent:

- Happens on any subscription whose top-level collection has more than one relation field (a "join" type — User+conversation on one row in our case). Single-relation collections never panic.
- Doesn't need P2P. A local mutation on the same node fires it too, as long as the subscription is open at the time.
- The data lands cleanly. A regular query for the doc right after the panic returns everything including the nested relations. So this is purely a subscription re-eval problem, not a storage one.
- With the recover scoped at the goroutine level, the symptom is: SSE request returns 200 with LengthBytes=0, then the client reconnects and the next event panics the same way, so the stream never re-establishes.

The panic itself points at the post-filter `Discard` in the subscription handler — RunSelection comes back with a non-empty result, the filter loop empties it, and then Discard trips on an iterator that selection left open. So somewhere on the planner side, the partial-error path is leaking an iterator on the parent badger txn instead of letting it get cleaned up with the rest of the plan.

A thing that fell out of digging: when one side of a join's Init or Start fails, the other side isn't getting Close called on it. And when sibling plans under a parallel node Close, the first one that errors short-circuits the rest. Either of those is enough on its own to strand an iterator; together they're sneaky because the node-level cleanup *looks* like it ran.

Two fixes that together close it:

- Close every sibling under a parallel node even if one's Close errors (accumulate, don't short-circuit).
- On a join, if one side's Init/Start fails, Close the other side before propagating; on Close, do both sides regardless.

With both of those in place, the panic stops firing for us in both the unit harness and the e2e repro. SSE events deliver normally again.

Happy to send a PR — most of the work's already on a branch on our side, just want to sanity-check the approach before opening it. Main thing I'd want a second pair of eyes on is whether there are other planner nodes with the same short-circuit-on-error shape that should get the same treatment. ParallelNode and the type-join were the two the stacks pointed at, but I didn't audit the whole planner tree.

Also: while we're in there, the recover in the subscription handler should probably move inside the per-event loop so one bad event doesn't end the whole stream. That's a separate change but it matters a lot for client behaviour — without it, even if the leak gets fixed, any future panic on the eval path silently kills subscriptions.

Reproduces in a single-node unit test if you (a) have a join-shape collection (two relations), (b) open a subscription on it with the default selection, (c) do a local mutation while the sub is open. About 50 lines.

islamaliev:
## Relevant issue(s)
- Resolves https://github.com/sourcenetwork/defradb/issues/4826

## Description

When a subscription evaluates an event whose top-level collection has more than one relation field, the planner builds a join tree where sibling plan nodes share the parent badger txn. Two close-propagation gaps in that tree strand iterators on the txn when one sibling errors mid-Init/Start: the parallel-node Close short-circuits on the first child error, and the type-join skips the other side's Close on a partial failure. Either alone is enough to trip badger's `Unclosed iterator at time of Txn.Discard.` check on cleanup.

Fix is to accumulate Close errors across every child of a parallel node, and to close both sides of a type-join unconditionally on Close and on partial Init/Start failure.

Per the 2026-05-21 dev sync, the subscription handler's per-event recover is not part of this PR — that conversation is tracked separately as a post-v1 item.

## How has this been tested?

New unit tests under `internal/planner/close_propagation_test.go` pin the contract directly with sentinel plan nodes: every child of a parallel node is closed even when an earlier child's Close errors, both sides of a type-join are closed even when one errors, and `Init`/`Start` failures on one side close the other before propagating.

Verified end-to-end against the reproducer that originally surfaced this (panic was firing 2–3× per round; now 0).


internal/db/subscriptions.go
```diff
@@ -27,6 +30,25 @@ type subscriptionSelector interface {
 	CheckDocIDFilter(docID string) bool
 }
 
+// subscriptionSelectionFn evaluates a single subscription event's
+// selection. Injected into processEvent so tests can substitute a
+// stub without touching package-level state.
+type subscriptionSelectionFn func(ctx context.Context, db *DB, sel request.Selection) (map[string]any, error)
```
AndrewSisley: todo: Please don't do this.  This is not a path I want us to go down.

This is introducing test only code into the production code space, to test something that is very much testable via public entry points.

Additionally, it is relying on context specific unit test instead of our standardized integration tests.  There appears to be very litte reason for this given how easy it is to reproduce this issue via public code paths, and the downsides of doing this are severe, numerous and well known.

If you disagree with this, please bring it up with the team and we can talk about it, quite a lot has been written on this topic in the past and I'm surprised to see this here.

internal/db/subscriptions.go
```diff
@@ -65,79 +87,98 @@ func (db *DB) handleSubscription(ctx context.Context, r *request.Request) (<-cha
...
+func processEvent(
+	ctx context.Context,
+	db *DB,
+	subRequest subscriptionSelector,
+	evt event.Update,
+	resCh chan<- client.GQLResult,
+	selectFn subscriptionSelectionFn,
+) {
+	defer func() {
+		if r := recover(); r != nil {
```
AndrewSisley: todo: This does not appear to be fully recovering, and things can be left in a somewhat broken state.  The rabbit gives an example of this RE txn closure, it seems likely that there are others (e.g. open iterators, that will prevent the transaction from being closed).

This is also a significant departure from our existing no-recover policy, it should probably be raised with the team.

internal/db/subscriptions.go
```diff
@@ -65,79 +87,98 @@ func (db *DB) handleSubscription(ctx context.Context, r *request.Request) (<-cha
...
+			stack := make([]byte, 64*1024)
```
AndrewSisley: todo: Please document this magic number.

AndrewSisley (changes requested):
I think this needs some changes before it can be merged.

AndrewSisley:
> Follow-up to #e9fd5cefb.

thought: Your PR descriptions have been a little unhelpful lately, and I've held back on calling them out.  This one contains something that is particularly unhelpful, and unlikely to to have been reviewed by a human.  They can also be quite waffley - I'm not looking for a story here, but an information-dense description of what has been changed and why.

internal/db/subscriptions.go
islamaliev: I hate myself adding stuff to production code just for testing. But I didn't see any other reasonable way of testing this case. Please let me know. Maybe I'm missing some testing framework capabilities.

AndrewSisley: > Maybe I'm missing some testing framework capabilities.

We can add any capabilities we want to it.

The underlying issue that leads to the multi-close bug that this PR fixes is fairly easy to hit - we should be able test that without doing any fancy error injection/simulation magic, and very probably via existing test actions.

This entire PR is working on the management of errors that should never occur in a well behaving node though, so if you don't want to rely on testing via other potential bugs, you will need to artificially trigger an error via a test action.  We have talked for a long time about introducing an action or something that could corrupt store state - that could probably trigger a failure where you want it.

---------------

Or... As I was writing that out I thought - this entire PR is really focused on internal behaviour.  It *is* testing internal stuff that should *never* be exposed to users.  So testing is very arguably an (optional) internal unit testing problem.

And all this PR really is concerned with (ignoring the `recover` change which is separate, and arguably unwanted), is whether a planner node closes all of its children.

Not only is this internal to Defra, but it is internal to each planner node.

I think if you really want to test this (and I see this as low value, highly disposable unit test code), the best solution might just be to write unit tests for each plan node type, and test the internal behaviour that we care about - that calling `node.Close` calls `child.Close` for all of its children no matter if they error or not.

e.g. `multi_test.go`:
```
func CloseClosesChildrenTest(t) {
   n := &parallelNode{
     children: twoShimChildrenWhereTheFirstErrors,
  }
   
   e := n.Close()
   
   assertBothChildrenHadCloseCalled()
   assertChildErrorReturned(e)
}
```

internal/planner/type_join.go
```diff
@@ -540,22 +540,31 @@ func (join *invertibleTypeJoin) Init() error {
 	if err := join.childSide.plan.Init(); err != nil {
 		return err
 	}
-	return join.parentSide.plan.Init()
+	if err := join.parentSide.plan.Init(); err != nil {
+		// Roll back childSide: its Init may have opened resources
+		// (e.g. iterators on the parent txn) that the outer Close
+		// would otherwise miss.
+		return errors.Join(err, join.childSide.plan.Close())
```
AndrewSisley, line 547: question: I do not understand why this, and the similar `childSide` close in `Start` is required.

If the `parentSide` `Init`/`Start` errors, `join` is invalid and should be closed by its owner, at which point `childSide` should be closed no?

At the moment this looks like an unnecessary code complication which provides readers with avoidable questions and stress.

AndrewSisley (approved):
LGTM, thanks Islam

AndrewSisley:
> thought: Your PR descriptions...

I want to expand a little bit on this, as I feel like we might place differing levels of importance on PR descriptions.

In my opinion, the description, and how it is written matters quite a lot.  It is the first thing reviewers read when opening a PR, and first impressions matter.  If the description appears low quality and low effort you are essentially priming the reviewer into thinking that the code changes following it are low quality and low effort.

IMO this is worsened when AI is added to the mix.  If the description reads like it was auto-generated and not reviewed by the author, then you are priming readers to expect the code to be in a similar state, and as well as the quality issue, this is also a trust and ownership issue, and one that is much less likely to generate empathetic comments from human reviewers than similar quality code produced by a human - something that further reduces the PR experience for both reviewers and author.

internal/db/fetcher/versioned_multi.go
```diff
@@ -121,12 +121,21 @@ func (f *MultiVersioned) Start(ctx context.Context, prefixes ...keys.Walkable) e
...
+		// Track the child immediately so that any resources it opened
+		// during Init (e.g. the in-memory `vf.root` store) and any
+		// resources it opens during Start (which may partially succeed
+		// before erroring) get released by Close(). Without this, a
+		// Start failure leaves an orphaned VersionedFetcher whose
+		// internal state — including badger iterators opened on the
+		// parent txn via the planner/fetcher path — cannot be
+		// reclaimed, which causes badger's Txn.Discard to panic with
+		// "Unclosed iterator at time of Txn.Discard".
+		f.children[i] = child
```
fredcarle: thought: `VersionedFetcher.Init` allocates `vf.root` before opening/copying the system store, so an `Init` failure returns before `f.children[i]` is assigned and `f.Close` cannot release that transient store. Should probably move this above `child.Init`.

PR #2602, opened 2024-05-08
feat: Enable sec. indexes with ACP

islamaliev:
## Relevant issue(s)

Resolves #2365

## Description

Enable secondary indexes with ACP.

Issue #2365, opened 2024-03-04
ACP - Secondary Indexes


shahzadlone:
Support Secondary Indexes with ACP

shahzadlone:
Had an intro call with @islamaliev he will be working on this task so we can start kicking off the items of the task list under: https://github.com/sourcenetwork/defradb/issues/1738

PR #4835, opened 2026-05-26
fix: P2P-merged ACP doc with @index no corrupt index

### LINKED ISSUE #4834: bug(p2p+acp+index): secondary-index Save silently skipped on inbound merge. ErrCorruptedIndex on next update
Found this while building group-chat feature. P2P-replicated doc under document ACP with `@index` on a mutable field — non-creator on the receiving node tries to update the indexed field and gets:

```
failed to update index: corrupted index. Please delete and recreate the index. Name: Users_name_ASC. IndexName: Users_name_ASC
```

The single-node DAC equivalent works fine (same policy, same `@index`, same non-creator update). The bug only fires when the doc arrived on the writing node via P2P replication.

## What's actually going wrong

The doc data and the secondary-index entry are written by **different code paths** on the receiving node, and only one of them goes through document ACP.

Doc data flows in via the low-level merge pipeline:
- `processPushlogRequest` runs `trySelfHasAccess` up front using the node's own DID — that's the P2P-level access gate, and it talks to SourceHub directly.
- Gate opens, `executeMerge` runs, `mergeComposites` writes CRDT deltas straight to `txn.Datastore()` and blocks to `txn.Blockstore()`. None of that consults document ACP — it's below that layer.

So the doc lands. `txn.Datastore().Get(primaryKey)` returns the value, the data is on disk.

The index is reconciled at the tail of `executeMerge`:

```go
for docID, oldDoc := range mp.docIDs {
    err = syncIndexedDoc(ctx, docID, mp.col, oldDoc)
    ...
}
```

`syncIndexedDoc` (`internal/db/merge.go:571`) needs the post-merge doc as a `client.Document` to compute the index entry. It does:

```go
newDoc, err := col.GetDocument(ctx, docID)
```

`GetDocument` is the user-facing read API. It goes through `permissionedFetcher`, which checks document-ACP against whatever identity is on `ctx`. The merge ctx on the receiving node carries no caller identity — the P2P transport doesn't stamp one on, the access check at the gate was done by passing the node DID inline to `CheckDocAccessWithIdentityFunc` and never written back to ctx.

So the fetcher sees an empty identity, fails the ACP check, returns no rows. `GetDocument` returns `nil, ErrDocumentNotFoundOrNotAuthorized`. `syncIndexedDoc` hits its both-nil early return:

```go
if oldDoc == nil && newDoc == nil {
    return nil
}
```

No log, no warning. The index Save never runs. The doc data is on disk; the index entry isn't.

Then later when someone updates the indexed field, `updateIndexedDoc` reads the current doc (succeeds — data is there), goes to `index.Update(oldDoc, newDoc)` which goes to `index.Delete(oldDoc)` which goes to `deleteIndexKey`:

```go
exists, err := txn.Datastore().Has(ctx, key)
if !exists {
    return NewErrCorruptedIndex(indexDesc.Name)
}
```

The deleter doesn't know "this entry was never written" vs "this entry was lost mid-write." Both look like corruption.

## Why this is index-only in symptoms

`mergeComposites` doesn't consult document-ACP per-block — gating happens at the P2P boundary, not at the datastore boundary. So data writes always succeed once the gate opens. The index is the only thing in the merge tail that fetches the doc back through the user-facing read path (because index reconciliation is implemented as "look up the doc, then compute its index entry").

Nothing about indexes themselves is broken. They're collateral damage from `GetDocument` being the wrong tool for the merge processor to fetch a doc it itself just wrote.

## Reproduction

```
DEFRA_DOCUMENT_ACP_TYPE=source-hub go test \
  ./tests/integration/acp/dac/p2p/ \
  -run TestACP_P2P_IndexedFieldUpdatedOnPeerNode -v -count=1
```

(test landing in the fix PR; setup is: two networked nodes, policy with `read = reader + updater + deleter`, schema with `@index` on a mutable field, doc on node 0, grants of `updater` to both a non-creator identity and node 1, then a non-creator update on node 1.)

## Fix direction

Two options I see, either works:

1. **Stamp the node identity onto ctx at the P2P boundary** before calling `p.db.Merge(...)`. Mirrors the existing fallback in `internal/db/document.go:491` ("if no caller identity, fall back to the node identity"). The fetcher's ACP check then asks SourceHub the same question the P2P gate already asked, gets the same yes, and the doc comes back non-nil. Pro: smaller diff, no new bypass. Con: redundant ACP round-trip per inbound doc, and silently couples merge correctness to the receiving node holding a document-ACP grant — which is enforced today by DGC-012 but is a separate invariant from "the P2P gate said yes."

2. **Bypass document-ACP for the internal read.** Add a private `getInternal` helper that initializes the fetcher with `documentACP = None` and call it from a small `getDocForMerge` wrapper in merge.go. The merge processor already trusts the P2P gate; asking ACP a second time inside the merge tells us nothing new. Pro: no redundant round-trip, and consistent with the fact that the data side of merge is already below ACP. Con: introduces a new escape hatch that future callers might misuse.

I have a working fix on `fix/update-indexed-acp-doc-by-non-owner` using option 2; PR coming once I've decided which approach to defend. Either way the silent both-nil early return in `syncIndexedDoc` should at minimum log — that's the bit that made this hard to track down.

## Related

- #2379 (closed) — older "secondary indexes don't work with P2P" bug, fixed before ACP entered the picture.
- #4405 (closed) — "Fix P2P encryption with index" — same family (index Save dropped on receiver), different cause (encryption).
- #4572 (open) — adjacent P2P+ACP bug (PushLog Creator field), but about sender-side enforcement, not the receiver-side index drop.

islamaliev:
## Relevant issue(s)
- Resolves #4834

## Description
When a doc under document ACP is replicated to a peer node and the schema declares `@index` on a mutable field, the secondary-index entry was never written on the receiving node — `syncIndexedDoc` fetched the post-merge doc via `col.GetDocument`, the merge ctx had no caller identity, the ACP read filter returned nil, and the index Save was silently skipped. The next update on the indexed field then hit `ErrCorruptedIndex` because `deleteIndexKey` could not find the prior-value entry.

The fix routes both merge-time doc fetches through a new `getDocForMerge` helper backed by `collection.getInternal`, which initializes the fetcher with `documentACP = None` and bypasses the read filter. Access is already gated at the P2P boundary by `trySelfHasAccess` using the node DID, so the merge processor doesn't need to re-check — and the data side of merge already runs below ACP, so the index side now matches.

The alternative considered was to stamp the node identity onto ctx at the P2P boundary before `db.Merge`, so the existing ACP path validates the read with the node DID. That mirrors the fallback at `document.go:491` and avoids adding a bypass, but costs a redundant SourceHub round-trip per inbound merge and couples merge correctness to the receiving node continuing to hold a document-ACP grant (today enforced by DGC-012). **Worth a quick team discussion before this lands** — the swap is small if the team prefers the identity-stamping shape.

Also added an info log on the both-nil early return in `syncIndexedDoc` so future silent skips are visible, and dropped a stale TODO referencing #2365 (resolved by #2602 in 2024).

## How has this been tested?
New integration tests under `tests/integration/acp/dac/`:
- `p2p/update_by_non_creator_on_peer_test.go` — full P2P + ACP + @index regression guard plus a non-indexed control that closes a broader coverage gap (no existing P2P DAC test had a non-creator `UpdateDoc` on the receiving node).
- `index/update_by_non_creator_test.go` — single-node sibling.

Full `./tests/integration/acp/...` suite (sourcehub backend) is green.


internal/db/collection_index.go
```diff
@@ -85,8 +85,6 @@ func (c *collection) updateIndexedDoc(
 		return err
 	}
 
-	// TODO-ACP: https://github.com/sourcenetwork/defradb/issues/2365 - ACP <> Indexing, possibly also check
```
shahzadlone, line 88: suggestion: Update line 839 in `acp/README.md` as well please.

It looks like this:

```
- [Adding Secondary Indexes](https://github.com/sourcenetwork/defradb/issues/2365)
```

internal/db/merge.go
```diff
@@ -485,14 +487,30 @@ func (mp *mergeProcessor) trackMergedDocument(ctx context.Context, docID client.
 	if exists {
 		return nil
 	}
-	doc, err := mp.col.GetDocument(ctx, docID)
+	doc, err := getDocForMerge(ctx, mp.col, docID)
 	if err != nil && !errors.Is(err, client.ErrDocumentNotFoundOrNotAuthorized) {
 		return nil
```
shahzadlone: question: do we still want to return `nil` for all non `client.ErrDocumentNotFoundOrNotAuthorized` errors?

suggestion: Perhaps it can be logged?

tests/integration/acp/dac/p2p/update_by_non_creator_on_peer_test.go
```diff
@@ -0,0 +1,274 @@
...
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+// Control for the test above: same scenario without @index. Confirms
+// the "non-creator updates on peer node" path itself was fine; the
+// indexed-vs-not difference is what the indexed test is exercising.
+// Also closes a coverage gap — no existing P2P DAC test had a
```
shahzadlone: suggestion: I really dislike this how claude uses EM dashes (—) instead of normal hyphen (-). please fix. 

```suggestion
// Also closes a coverage gap - no existing P2P DAC test had a
```

internal/db/collection_index.go
islamaliev: done

internal/db/merge.go
islamaliev: I changed it to return the error

shahzadlone (approved):
LGTM

tests/integration/acp/dac/p2p/update_by_non_creator_on_peer_test.go
fredcarle: This is the wrong use of an hyphen though and the correct use of an EM dash. LLMs have given them a bad rap but that they are using them correctly.

PR #4836, opened 2026-05-26
fix: Parse schemas with multi-byte chars in SDL comments

islamaliev:
## Description
A schema whose comment contained any non-ASCII character — an em-dash, a curly quote, an accented letter, an emoji — failed to load with a misleading syntax error pointing at a line that looked fine. The same schema with an ASCII-only comment worked. The cause was in the GraphQL lexer: while skipping comments and whitespace it miscounted multi-byte characters, shifting its position so the rest of the schema was read from the wrong place. This pulls in the upstream lexer fix and adds regression coverage for adding a collection whose schema carries such a comment.


fredcarle (approved):
LGTM

PR #4850, opened 2026-05-28
feat: Create a .deb file when building C bindings for Linux

### LINKED ISSUE #4848: Add .deb artifact for cbindings
From build week feedback: https://discord.com/channels/1374835078781468722/1508850738648973372

ChrisBQu:
## Relevant issue(s)

Resolves #4848

## Description

This introduces a new option for the Linux build script, which will generate a .deb file. It can be invoked with the following syntax:

`make build-c-shared-linux:deb`

In addition to this new build option, this PR introduces a new CI workflow which tests it in a Debian environment. It uses pre-existing C code, the same source code use for the shared object smoke test, to test a program using the generated .deb file.

## Tasks

## How has this been tested?

- WSL
- Debian by way of GitHub CI


Makefile
```diff
@@ -467,7 +467,7 @@ build-c-static-windows:
 	
 .PHONY build-c-shared-linux:
 build-c-shared-linux:
-	@tools/scripts/build-c-shared-linux.sh $(BUILD_FLAGS)
+	@MAKE_DEB=$(MAKE_DEB) tools/scripts/build-c-shared-linux.sh $(BUILD_FLAGS)
```
AndrewSisley: question: ` make foo MAKE_DEB=1` is a little suprising to me, but I don't have too much experience here - why did you decide to go with this instead of a new make command, such as `make foo:bar`?

AndrewSisley (approved):
Thank you very much Chris, and thanks for including the headers too :)

.github/workflows/test-deb-package.yml
```diff
@@ -0,0 +1,53 @@
+# Copyright 2026 Democratized Data Foundation
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+name: Build and Test Debian Package (Linux)
```
shahzadlone: suggestion: Either change name of job and file to be similar or just do this:
```suggestion
name: Test Debian Package (Linux)
```

fredcarle (approved):
LGTM. Please answer/clear-up Andy's question and Shahzad's suggestion before merge.

Makefile
ChrisBQu: The honest answer, is that I also lack experience in this area. In my mind, the ideal was having it as a flag, but I was convinced that this breaks the Make syntax, and would be inappropriate. I think I got too attached to the idea of it being passed as a flag/variable, of some kind, and I forced the square into the circular hole.

I have no push back here though, and I'm happy to make a new command. With one caveat; we cannot use a colon because that character has special meaning in Makefiles. We can use a hyphen though. Let's do that, unless you have another specific suggestion?

Edit: There was a bit of misunderstanding here. Sorted out with a quick Discord DM. We will use the colon, and an escape slash.

.github/workflows/test-deb-package.yml
ChrisBQu: Will do, good call.

Makefile
AndrewSisley: I have no preference at all, I was just curious - no need to change if you don't want to.

> With one caveat; we cannot use a colon because that character has special meaning in Makefiles.

With a colon it doesn't need to be a param, it would just be an overload of sorts:

`make build-c-shared-linux` and `make build-c-shared-linux:deb` (hardcoded, to make a .deb, not a variable)

PR #4852, opened 2026-05-28
test: Cover GraphQL subscriptions on ACP-gated collections

### LINKED ISSUE #4851: Add test coverage for GraphQL subscriptions on ACP-gated collections
We don't have any integration test that exercises a GraphQL subscription against a policy-gated collection with an authenticated subscriber. The subscribe path does honor identity end-to-end (HTTP middleware -> subscription goroutine -> per-event planner/fetcher ACP check) but nothing in the test suite proves it stays that way.

This came up while debugging a downstream integration where an SSE client subscribing without an Authorization header silently received no events on a gated collection. Turned out to be the correct ACP behavior, but it took some digging because there was no test to point at.

Worth adding a small test that covers the owner case, the granted-reader case, and the anonymous case so a future change can't regress the identity threading through the subscription goroutine without something failing.

Also worth noting: the SubscriptionRequest action in the test harness currently has no Identity field, so writing such a test needs a tiny harness extension first.

islamaliev:
## Relevant issue(s)

- Resolves https://github.com/sourcenetwork/defradb/issues/4851

## Description

We had no integration test that exercised a GraphQL subscription against a policy-gated collection with an authenticated subscriber. The subscribe path does honor identity end-to-end, but nothing in the suite proved it stayed that way.

Adds three small tests: an authenticated owner receives create events, a granted reader receives update events, and an anonymous subscriber receives nothing. The test harness's subscription action did not previously have an identity field, so this PR also adds one and wires it through, mirroring how the regular request action already handles identity.

While writing the anonymous case it also became clear that the harness's assertion logic was weaker than it looked. The "expect no events" path returned without ever listening on the channel, so any test using that pattern passed whether events were delivered or not. The exact-count path had the symmetric gap: once N events had been collected the goroutine exited, so an over-delivery of N+1 events was silently dropped and the test still passed. The collection logic is now extracted into a small helper with unit tests; it listens through the relevant window in both branches and a length check turns over-delivery and under-delivery into real failures. Existing subscription tests are unchanged in behavior because none of them were actually in a scenario where extra or unexpected events were being produced.

## How has this been tested?

The new ACP subscription tests pass on both the local document ACP backend and SourceHub. The new harness unit tests cover the no-care, expect-empty, and over-delivery cases and demonstrably catch the original silent-pass behavior. The pre-existing subscription suite is unchanged.


fredcarle (approved):
LGTM.

PR #4856, opened 2026-06-01
fix: Normalize trailing slashes on HTTP API routes

### LINKED ISSUE #4855: Trailing slash in path 404 http error
Discussed here: https://discord.com/channels/1374835078781468722/1511007169795981525/1511007174128828496

Basically this:

```
DELETE /collections?name=Book   -> 200  (DeleteCollection)

DELETE /collections/?name=Book  -> 404  body="404 page not found\n"   <-- the trailing slash

DELETE /collections/Book        -> 200  (DeleteDocumentsWithFilter)
```

shahzadlone:
## Relevant issue(s)
Resolves #4855

## Description
HTTP requests with a trailing slash before the query string (e.g. `DELETE /api/v1/collections/?name=Book`) returned a bare `404 page not found`.

chi matches paths exactly, so `/collections/` matched neither `/collections` nor `/collections/{name}` and fell through to the router's not-found handler.

The CLI and Go/JS clients were unaffected because they build slashless URLs, so this only surfaced for hand-crafted requests (curl/Postman).

Registers chi's `StripSlashes` middleware once in `NewHandler`, before any routes, so `/x` and `/x/` resolve to the same handler across every HTTP route.

`StripSlashes` (rather than `RedirectSlashes`) rewrites the routing path in place with no redirect, preserving the request method

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sourcenetwork/defradb/event"
 
 	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
 )
 
 // IsDevMode is a global variable for the development mode flag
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
LGTM. Sorry again for the flip flopping.

shahzadlone:
> LGTM. Sorry again for the flip flopping.

All G, the process was important for us to have a deeper understanding of why and what we want to pick!

PR #4862, opened 2026-06-01
fix: Make DAC enabled by default

### LINKED ISSUE #4824: Investigate why DAC was changed to `off` by default and make it enabled by default
Investigate why DAC was changed to `off` by default and make it enabled by default

shahzadlone:
## Relevant issue(s)
Resolves #4824 

## Description
Change the CLI default for acp.document.type from none to local so defradb start runs with document access control enabled by default. The node library default (DefaultNodeOptions) was already local; the CLI config was the only thing overriding it to none.
Collections without a `@policy` remain public; collections with a `@policy` are now enforced by default.


fredcarle (approved):
LGTM

PR #4804, opened 2026-05-10
test(i): Add edge cases for collection delete command

shahzadlone:
## Relevant issue(s)
- Resolves #4802

## Description
Adds breadth coverage for `DeleteCollection` edge cases that I was worried about.

PR #4868, opened 2026-06-02
test(i): Detect txns opened by any action in change detector

### LINKED ISSUE #4867: Fix the change detector wip improvement
I was trying to improve the change detector to track transactions opened by any action (e.g. also `DeleteCollection`), not just the three mutation types we had before. But the implementation swapped boundary-scoped tracking for whole-test global tracking, which is the regression.

shahzadlone:
## Relevant issue(s)
Resolves #4867 

## Description
- This reverts the accidental commit d9ff32a (from #4804) that got merged
- and then re-does the changes in the commit after that to ensure a clean review of the last commit.
- It now should detect transactions opened by any action (not just the previous existing mutations, which would miss new mutation actions that would be introduced, like the newly introduced one: `DeleteCollection`)
- More context on why this was needed: https://discord.com/channels/1374835078781468722/1390650826317627502/1511391395015950526

## For Reviewers:
- All the changes that happening in the accidental merge commit and improvements on it are in this one commit (ignore the first commit, it just reverts the previous changes): `PR(FIX): Skip change-detector tests whose transactions span the split`

### DIFF
--- a/tests/integration/utils.go
+++ b/tests/integration/utils.go
@@ -794,35 +794,55 @@ func getActionRange(t testing.TB, testCase TestCase) (int, int) {
 		return startIndex, endIndex
 	}
 
-	// A test that leaves a transaction open cannot be run by the change detector:
-	// uncommitted writes are never persisted, so the setup and assert phases would
-	// operate on different data. Any action can run inside a transaction - not just
-	// the mutation actions that delimit the setup phase below - so this is checked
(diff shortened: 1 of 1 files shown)


fredcarle (approved):
Thanks for correcting this.

ChrisBQu (approved):
LGTM

PR #4876, opened 2026-06-02
feat: Add GetNodeOptions functionality to C bindings/CLI

### LINKED ISSUE #4870: Add way to check node initialization options
When we create a node, we can configure it by passing in a variety of node options. However, there is currently not a way to investigate a node and check which options it was configured with.

### LINKED ISSUE #4894: There is no CLI flag to enable/disable pubsub
This came up while working on functionality and tests to get node initialization options. But there is a config default called `net.pubSubEnabled`, but no accompanying CLI flag. This small task can be tacked onto the existing PR for node option testing, I think.

ChrisBQu:
## Relevant issue(s)

Resolves #4870 
Resolves #4894 

## Description

A recent PR added functionality to the CLI and C bindings to configure the node with all possible options. This PR adds functionality for retrieving those options from a node, after it has been created. In the case of the C bindings, this means a a new function, `GetNodeOptions`, and in the case of the CLI, this means a new command, `node-options`.

In both cases, the list of options will be returned as a JSON string. Sensitive information, such as private key will be either nil, or the string, "<redacted>" for security reasons.

Beyond thee core functionality, this PR introduces tests for both clients. The tests are constructed such that we set all possible flags in two different ways, call them, Set A, and Set B. Then, in both cases, we check the return value of the new GetNodeOptions function, and compare them to what we expect them to be. We do two test sets to ensure that in no case does a test pass because the default value happens to be what we explicitly set it to (in other words: we are future proofing the tests, in case the default value should ever change.)

## Tasks

## How has this been tested?

- WSL


cli/start.go
```diff
@@ -93,7 +93,8 @@ func MakeStartCommand(ctx context.Context) *cobra.Command {
 				SetDisableP2P(cfg.GetBool("net.p2pDisabled"))
 			opts.Store().
 				SetPath(cfg.GetString("datastore.badger.path")).
-				SetBadgerInMemory(inMem)
+				SetBadgerInMemory(inMem).
+				SetBadgerFileSize(int64(cfg.GetInt("datastore.badger.valuelogfilesize")))
```
nasdf, line 97: praise: nice catch!

http/handler.go
```diff
@@ -88,12 +89,18 @@ type Handler struct {
 	txs *sync.Map
 }
 
-func NewHandler(db DB) (*Handler, error) {
+func NewHandler(db DB, nodeOpts ...*options.NodeOptions) (*Handler, error) {
```
nasdf: question: Is there a downside to making this non-variadic?

nasdf (approved):
I left one comment that might be worth looking into but otherwise looks good!

http/handler.go
ChrisBQu: None come to mind. I didn't do it because it would mean touching more lines of code to make them fit. But it's purely a stylistic difference. Happy to go ahead and change it now before merge.

PR #4877, opened 2026-06-02
fix: Handle truncation of more than 11MB

### LINKED ISSUE #4369: Truncate will probably fail when truncating more than 11MB of docs on badger
I should of tested this a while ago.

> The maximum size that a transaction can be looks to be (15 * badgeropt.MemTableSize) / 100 - default is 64 << 20 (67108864), which gives a max transaction size of 10066329 bytes.
>
> There is also a (default) limit of 104857 keys in a badger transaction, this is the size limit / 96.  Defra writes (roughly) (3*NumberOfFields)+1+NumberOfIndexedFields keys per document (add 2 for docs in branchable collections).

Testing this will be expensive due to the amount of data required for it to fail, I recommend not picking this up until the nightly CI has been setup, so that the tests can run there.

AndrewSisley:
## Relevant issue(s)

Resolves #4369

## Description

Handles the truncation of more than 11MB worth of key-values (a badger transaction limit), by only using the transaction in order to scope the collection lock - the deletion is now handled by direct writes to the store.

The order of operations has changed somewhat so that child records are deleted before the parents that reference them are deleted.  This means that if an error occurs part way through execution, there should be no orphaned key-values, allowing the truncate to be reattempted and eventually complete.

Additional testing of this fix is blocked behind the nightly build issue, as I don't want to slow down the normal test runs by creating and truncating large volumes of data.  There are existing tests for truncate that test using small amounts of data.   

Is probably worth reviewing this commit by commit, the messages and context-reduction might make it easier to understand what and why stuff was done.

### DIFF
--- a/client/collection.go
+++ b/client/collection.go
@@ -162,6 +162,12 @@ type Collection interface {
 	//
 	// This call will lock the collection, and no other read or write document operations on this collection
 	// will progress whilst this is executing.
+	//
+	// Truncate is not transactional. When called with a transaction context, the transaction is used only to
+	// scope the collection lock - the truncate effects are applied immediately.
+	//
+	// If an error is returned, the truncate may have been partially applied - because of this it is strongly
+	// suggested to call `Truncate` until it succeeds.
 	Truncate(ctx context.Context, opts ...options.Enumerable[options.TruncateCollectionOptions]) error
 }
 
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -415,76 +415,138 @@ func (c *collection) hardDeleteCollectionBlocks(
 // a block with this cid is found.
 //
 // If the block is not found, it will not error.
-func deleteBlocks(ctx context.Context, head cid.Cid) error {
+func deleteBlocks(ctx context.Context, currentCid cid.Cid) error {
 	txn := datastore.CtxMustGetTxn(ctx)
 	blockstore := txn.Blockstore()
 
-	toDelete := map[cid.Cid]struct{}{
-		head: {},
+	type block struct {
+		id    cid.Cid
+		block *coreblock.Block
 	}
-	for len(toDelete) != 0 {
-		var currentBlockCid cid.Cid
-		for v := range toDelete {
-			// Pop the first key off of the `toDelete` set.
-			currentBlockCid = v
-			delete(toDelete, currentBlockCid)
-			break
+
+	coreBlock, isFound, err := getBlock(ctx, blockstore, currentCid)
+	if err != nil {
+		return err
+	}
+	if !isFound {
+		return nil
+	}
+
+	toDelete := []*block{
+		{
+			id:    currentCid,
+			block: coreBlock,
+		},
+	}
+
+	i := -1
+	isReversed := false
+	increment := func() bool {
+		if isReversed {
+			i--
+		} else {
+			i++
 		}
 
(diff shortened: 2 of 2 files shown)


internal/db/collection_truncate.go
```diff
@@ -405,73 +418,137 @@ func (c *collection) hardDeleteCollectionBlocks(
...
+			// we only need to iterate through twice, once in either direction, once we have finished
+			// iterating in reverse, all blocks should have been deleted.
+			return false
 		}
 
-		decodedBlock, err := coreblock.GetFromBytes(currentBlock.RawData())
-		if err != nil {
-			return err
+		return true
+	}
```
fredcarle, line 468: thought: The cognitive overhead of this `increment` function and its usage bellow makes the `deleteBlocks` methode feel more complex than it needs to be. 

question: Why not use a `container/list` like the version fetcher does?

fredcarle (approved):
LGTM. Just a thought bellow as I don't want to block the PR for this but I'm a little concerned for future us for maintenance purposes..

client/collection.go
```diff
@@ -162,6 +162,9 @@ type Collection interface {
 	//
 	// This call will lock the collection, and no other read or write document operations on this collection
 	// will progress whilst this is executing.
+	//
```
fredcarle, line 165: todo: Please add that "`Truncate` is not transactional. When called with a transaction context, the transaction is used only to hold the collection lock, and truncate effects are applied immediately".

internal/db/collection_truncate.go
```diff
@@ -72,6 +72,15 @@ func (c *collection) truncate(
 	txn := datastore.CtxMustGetTxn(ctx)
 	c.db.lockSet.CollectionLock(txn, shortID)
```
fredcarle, line 74: todo: Please add documentation to say something like:
```
// The txn above is used only as the owner of the collection lock. The truncate
// mutations below intentionally bypass txn stores and write through db.rootstore,
// so they are not committed or rolled back with this txn.
```

fredcarle (commented):
Sorry I thought of a couple things after the approval. I'm adding a couple documentation todos.

internal/db/collection_truncate.go
AndrewSisley: > Why not use a container/list like the version fetcher does?

`++/--` is quite a lot simpler to to me, it is transaparent, and the couple of lines required to make it work is a lot simpler than reasoning with another type (especially an external one).

client/collection.go
AndrewSisley: Not a problem, will do.

- [x] expand docs

internal/db/collection_truncate.go
fredcarle: It is a stdlib type though. 

It's simpler to you because you wrote it. Not so much for others :)

internal/db/collection_truncate.go
AndrewSisley: I'm going to leave this as-is.

The second sentence is covered by the line immediately following this one.

The spirit of the first sentence is covered by the `client` documentation, and isn't relevant, or always true in its more literal sense.  This function and it's children do not own the transaction. It may be used for writes, both by us and users.  It can be, and is discarded/committed by the owning context.

internal/db/collection_truncate.go
AndrewSisley: > It is a stdlib type though.

That doesn't make it simpler, or used frequently enough for readers know it as readily as they know and understand `++/--`.

PR #4886, opened 2026-06-04
fix: Display string representation of field kind on collection describe

### LINKED ISSUE #4816: Display string representation of field kind on collection describe
```json
      {
        "Name": "plot",
        "Kind": 11,
        "Typ": 1,
        ...
      },
      {
        "Name": "title",
        "Kind": 12,
        "Typ": 1,
        ...
      }
```

SharanRP:
## Relevant issue(s)

Resolves #4816 

## Description

The collection describe command output showed field Kind as numeric values (e.g., 11, 12) instead of human-readable string representations (e.g., "String", "[String!]"), making it difficult to understand field types without referencing source code.
Added MarshalJSON method to CollectionFieldDescription that uses the existing Kind.String() method to serialize field kinds as human-readable strings.

## Tasks

## How has this been tested?

- Built and ran defradb client collection describe - output now shows "Kind": "String" instead of "Kind": 11
- Verified JSON round-trip (marshal -> unmarshal -> marshal) produces identical output

- Debian Linux

### DIFF
--- a/client/collection_field_description.go
+++ b/client/collection_field_description.go
@@ -99,32 +99,3 @@ func (f *CollectionFieldDescription) UnmarshalJSON(bytes []byte) error {
 
 	return nil
 }
-
-// MarshalJSON serializes the [CollectionFieldDescription] to JSON, using the string
-// representation of the [FieldKind] instead of its numeric value.
-func (f CollectionFieldDescription) MarshalJSON() ([]byte, error) {
-	var kindStr string
-	if f.Kind != nil {
-		kindStr = f.Kind.String()
-	}
-
-	return json.Marshal(&struct {
-		FieldID      string
-		Name         string
-		Kind         string
-		Typ          CType
-		RelationName immutable.Option[string]
-		IsPrimary    bool
-		DefaultValue any
-		Size         int
-	}{
-		FieldID:      f.FieldID,
-		Name:         f.Name,
-		Kind:         kindStr,
-		Typ:          f.Typ,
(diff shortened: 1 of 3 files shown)


client/collection_field_description.go
```diff
@@ -99,3 +99,32 @@ func (f *CollectionFieldDescription) UnmarshalJSON(bytes []byte) error {
 
 	return nil
 }
+
+// MarshalJSON serializes the [CollectionFieldDescription] to JSON, using the string
+// representation of the [FieldKind] instead of its numeric value.
+func (f CollectionFieldDescription) MarshalJSON() ([]byte, error) {
```
AndrewSisley: todo: Is it not possible to define`MarshalJSON` on the `FieldKind` interface implementations instead of `CollectionFieldDescription`?  This will save us from having to deal write a custom marshaller for stuff we don't need to care about, and makes a little more sense architecturally.

I think it would be bests not enforce this on `FieldKind` though, as it doesn't really need to be forced, just add the func to the concrete types (`ScalarKind` etc).

AndrewSisley (changes requested):
Thanks for picking this up Sharan!  Please make sure this has integration tests though, this one needs to be tested properly as it is fairly user-visible.

Probably using the `action.AddCollection` and `action.GetCollections` test actions (message on discord if you get stuck and want help :)).

ribiza:
Thanks Sharan for having take time to look at this. Another PR #4945 was merged with the fix.

PR #4889, opened 2026-06-04
fix: Handle the refresh of views larger than 11MB

### LINKED ISSUE #4386: Materialized views on badger probably cant refresh more than 11MB of docs at a time
Similar to https://github.com/sourcenetwork/defradb/issues/4369

We might get away with using collection locks and no transactions, similar to truncate.  I've not thought a whole load about this yet though.

Like truncate, it is probably sensible to wait for the nightly CI in order to properly test this.

AndrewSisley:
## Relevant issue(s)

Resolves #4386

## Description

Handles the refreshing of views larger than 11MB by using collection-locks instead of transactions, due to store-level transaction size limits.

Additional testing of this fix is blocked behind the nightly build issue, as I don't want to slow down the normal test runs by creating and refreshing large volumes of data. There are existing tests for view refresh that test using small amounts of data.

Some additional context/rational is provided by the commit messages.

### DIFF
--- a/client/db.go
+++ b/client/db.go
@@ -284,7 +284,7 @@ type Store interface {
 	// when making this call.
 	//
 	// This function will lock the selected views until it completes.  Its writes are not protected by transactions,
-	// so it errors, the database may be left in a state where the view has been partially refreshed - in this case,
+	// so if it errors, the database may be left in a state where the view has been partially refreshed - in this case,
 	// it is recommeded to retry the refresh.
 	RefreshViews(ctx context.Context, opts ...options.Enumerable[options.RefreshViewsOptions]) error
 
--- a/client/db.go
+++ b/client/db.go
@@ -282,6 +282,10 @@ type Store interface {
 	// The cached result is dependent on the ACP settings of the source data and the permissions of the user making
 	// the call.  At the moment only one cache can be active at a time, so please pay attention to access rights
 	// when making this call.
+	//
+	// This function will lock the selected views until it completes.  Its writes are not protected by transactions,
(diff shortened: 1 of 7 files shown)


client/db.go
```diff
@@ -282,6 +282,10 @@ type Store interface {
 	// The cached result is dependent on the ACP settings of the source data and the permissions of the user making
 	// the call.  At the moment only one cache can be active at a time, so please pay attention to access rights
 	// when making this call.
+	//
+	// This function will lock the selected views until it completes.  Its writes are not protected by transactions,
+	// so it errors, the database may be left in a state where the view has been partially refreshed - in this case,
```
fredcarle: nitpick: `so if it errors`

AndrewSisley: Cheers, will fix.

- [x] Fix documentation

shahzadlone (commented):
I read this in PR description:
> This PR will only actually resolve the problem once https://github.com/sourcenetwork/defradb/pull/4877 and https://github.com/sourcenetwork/defradb/issues/4891 has been merged, although it is not directly dependent upon the commits within it.

As both #4877 or #4891 are now merged, is there a way for us to add tests to check if this fix works?

AndrewSisley:
> As both https://github.com/sourcenetwork/defradb/pull/4877 or https://github.com/sourcenetwork/defradb/issues/4891 are now merged, is there a way for us to add tests to check if this fix works?

It is still dependent on the nightly stuff in order to automate testing.  I've removed that paragraph from the description, now that those are resolved, it made it a little confusing.

fredcarle (approved):
LGTM

PR #4892, opened 2026-06-05
test: Cover HTTP auth middleware rejection paths

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/4893

## Description

The HTTP auth middleware was almost entirely untested at the request level: only the no-token pass-through path was exercised, and every branch that produces the 403 had zero coverage. This adds four cases: token whose audience disagrees with the request host, no audience claim, malformed token, and the matching-audience counter-case.


nasdf (approved):
LGTM. Thanks for improving the coverage!

fredcarle (approved):
LGTM

PR #4895, opened 2026-06-05
fix: Support field kind for non-nillable array scalars

### LINKED ISSUE #4087: No non-array field kind for not nillable array scalars
e.g. there is no `Float64!` to go with `[Float64!]`.  This will break an unknown number of code places.

One of these places is the field definition blocks, as they deconstruct the kind using `SubTyp()`.  This means it is impossible to sync a array of not-nillable field definition correctly over P2P.

ChrisBQu:
## Relevant issue(s)

Resolves #4087 

## Description

This adds the following field kinds:

Boolean!
Int!
Float64!
Float32!
String!
DateTime!
Blob!
JSON!

It then adjusts the `IsObject` function inside `client/field_kind.go` to return the correct type in these cases of non-nillable arrays, and adjusts the `IsNillable` function to return false in the case of these new types.

The tests inside `collection_version/updates/add/field/kind/invalid_test.go` are updated to account for the new ScalarKind values associated with new types being in use.

Finally, a new set of integration tests has been included that illustrates these new types in action. This includes simple tests, as well as ones that sync between peers.

## Tasks

## How has this been tested?

- WSL


client/field_kind.go
```diff
@@ -95,16 +95,26 @@ func (k ScalarKind) String() string {
...
+	case FieldKind_BOOL:
+		return "Boolean!"
 	case FieldKind_NILLABLE_BOOL:
 		return "Boolean"
+	case FieldKind_INT:
+		return "Int!"
 	case FieldKind_NILLABLE_INT:
 		return "Int"
 	case FieldKind_NILLABLE_DATETIME:
 		return "DateTime"
```
fredcarle, line 109: todo: Can you please include `DateTime`, `Blob` and `JSON`.

tests/integration/query/inline_array/with_nonnillable_scalars_test.go
```diff
@@ -0,0 +1,425 @@
...
+}
+
+func TestQueryNonNillableDateTime_Null(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddCollection{
+				SDL: `type Users { name: String, createdAt: DateTime! }`,
+			},
+			&action.AddDoc{
+				Doc:           `{"name": "John", "createdAt": null}`,
```
AndrewSisley, line 323: question: This is odd... If the new types permit writing an explicit null (as well as permitting nulls via omission - another concern of mine that I haven't yet finished thinking about), what exactly is the difference between them and the nillable types?

Are you sure this is not a test artifact, and that the framework is actually passing the explicit null into Defra?

AndrewSisley: Or am I misreading the test, and you are matching on an error message containing a literal `null`?

In which case:

todo: Please don't assert on `null` being present in the string, please assert on the message body, if for no other reason than making the test make a lot more sense to people reading it.

AndrewSisley, line 323: todo: Please add tests for what happens when a non-nill field is completely omitted from the create statement and is nil implicitly.

client/field_kind.go
```diff
@@ -95,20 +95,36 @@ func (k ScalarKind) String() string {
 	switch k {
 	case FieldKind_DocID:
 		return "ID"
+	case FieldKind_BOOL:
+		return "Boolean!"
 	case FieldKind_NILLABLE_BOOL:
 		return "Boolean"
+	case FieldKind_INT:
+		return "Int!"
```
AndrewSisley, line 103: question: I can't remember if aggregates can ever return nil, if not, is it worth bothering to update those return types so they now use the nillable types now that we have them?

client/field_kind.go
```diff
@@ -296,14 +318,21 @@ const (
...
-	_                                ScalarKind      = 15 // safe to repurpose (was never used)
+	FieldKind_BOOL                   ScalarKind      = 15 // Repurposed 2026-06-05, was never used
 	_                                ScalarKind      = 16 // Deprecated 2024-03-15, was FieldKind_FOREIGN_OBJECT
 	_                                ScalarKind      = 17 // Deprecated 2024-03-15, was FieldKind_FOREIGN_OBJECT_ARRAY
 	FieldKind_NILLABLE_BOOL_ARRAY    ScalarArrayKind = 18
 	FieldKind_NILLABLE_INT_ARRAY     ScalarArrayKind = 19
 	FieldKind_NILLABLE_FLOAT64_ARRAY ScalarArrayKind = 20
 	FieldKind_NILLABLE_STRING_ARRAY  ScalarArrayKind = 21
 	FieldKind_NILLABLE_FLOAT32_ARRAY ScalarArrayKind = 22
+	FieldKind_INT                    ScalarKind      = 23
```
AndrewSisley, line 329: suggestion: Doesn't need to be done in this PR, and can probably be pushed to post v1, but the new numeric types should probably support the `PN_COUNTER` and `P_COUNTER` CRDTs.

tests/integration/query/inline_array/with_nonnillable_scalars_test.go
```diff
@@ -0,0 +1,425 @@
+// Copyright 2026 Democratized Data Foundation
```
AndrewSisley, line 1: todo: Quickly scanning the code I think it should work with the production code as-is (including gql stuff), but please add a couple of tests covering the sum and average of the new numeric types.

internal/request/graphql/schema/collection.go
```diff
@@ -759,7 +759,30 @@ func astTypeToKind(
 		}
 
 	case *ast.NonNull:
-		return client.FieldKind_None, ErrNonNullNotSupported
```
AndrewSisley, line 762: todo: It looks like you might have forgotten to enable filtering for the new types.  Please make sure this is done and tested.

suggestion: Might be worth adding an introspection test or two for this.

EDIT:
>  It looks like you might have forgotten 

I'm not sure reading the code, it might work now, but please add a quick test or two for this.  If it doesn't work, I suggest looking at `genLeafFilterArgInput` in `generate.go`

AndrewSisley (changes requested):
Requesting changes, it is hard to tell if some of the stuff missing from tests is deliberately omitted from this PR and is planned to be added or not.

I'm probably fine with a lot of these omissions to be done outside of this PR, or even post v1, but it needs to be a deliberate choice not an accidental omission.

tests/integration/query/inline_array/with_nonnillable_scalars_test.go
ChrisBQu: It's the latter, yes. The fact that it caused confusion indicates it was not self-evident, and is not clear enough. Will update this.

client/field_kind.go
ChrisBQu: I think for completeness, probably, it is? Not for all aggregates. For example, max and min can return nil if the array is empty. But for count, sum, and average? I think it's the right way to go. 

I'll have this bundled in with the other new tests you pointed out that we should have. That should be pushed in tomorrow.

ChrisBQu: Update: Upon thinking it through, this would be a breaking change. It would, specifically, break existing introspection tests. As such -- and I ran this past Andy -- I will be separating it into its own issue. It will be tackled following this one, but it will be addressed.

tests/integration/query/inline_array/with_nonnillable_scalars_test.go
```diff
@@ -0,0 +1,601 @@
...
+				SDL: `type Users { tags: [String!] }`,
+			},
+			&action.AddDoc{
+				Doc: `{}`,
+			},
+			&action.Request{
+				Request: `query { Users { tags } }`,
+				Results: map[string]any{
+					"Users": []map[string]any{
+						{"tags": nil},
```
AndrewSisley, line 528: question: Are we sure that this is the behaviour that we want?

If it is the behaviour we want, does the non-nilability really belong on in the collection block?  There is no guarantee that the field will not actually be nil. 

Does this break GQL?  The field type says it is not nil, yet the value can be nil.

client/field_kind.go
```diff
@@ -95,20 +95,36 @@ func (k ScalarKind) String() string {
 	switch k {
 	case FieldKind_DocID:
 		return "ID"
+	case FieldKind_BOOL:
```
AndrewSisley, line 98: suggestion: Doesnt need to be done in this PR if you don't think that is appropriate, but we still have a gap here and do not support `[Int]!` and similar.  Is there a good argument for requiring `Int!` for v1, but not `[Int]!`? 

Note the difference between `[Int!]`, `[Int!]!` and `[Int]!` - all should really be supported in order to meet the GQL spec.

AndrewSisley (changes requested):
Blocking behind the question `Are we sure that this is the behaviour that we want?` - otherwise looks good to me :)

internal/db/definition_validation.go
```diff
@@ -898,6 +899,36 @@ func validateFieldNotMutated(
...
 
+func validateNonNillableFieldNotAdded(
+	ctx context.Context,
+	db *DB,
+	newState *definitionState,
+	oldState *definitionState,
+) error {
+	var errs []error
+	for _, newCol := range newState.activeCollectionsByName {
+		oldCol, ok := oldState.activeCollectionsByName[newCol.Name]
```
AndrewSisley: suggestion: Matching by CollectionID should be more robust, when collection names become mutable, this check will misbehave

internal/db/definition_validation.go
```diff
@@ -898,6 +899,36 @@ func validateFieldNotMutated(
...
+		if !ok {
+			continue
+		}
+
+		oldFieldNames := map[string]struct{}{}
```
AndrewSisley: suggestion: Similar to above comment, matching on the FieldID should be more robust, as long term it should be possible to rename fields.

AndrewSisley (approved):
LGTM, approving now as I'm going to be off for the first half of next week and don't want to leave this blocked when we've agreed on the important bits etc.

PR #4910, opened 2026-06-09
fix: Clear stale errors in multiScanNode per accessor cycle

### LINKED ISSUE #4897: multiScanNode stores first error in n.err forever, returning it from every subsequent Init/Start/Next/Close
Split out from #4814 / #4827.

In `internal/planner/scan.go:400-414`:

```go
func (n *multiScanNode) countAndCall(count *int, f func() error) {
    if *count == 0 {
        err := f()
        if err != nil {
            n.err = err   // written on error, never cleared
        }
    }
    *count++
    if *count == n.numReaders {
        *count = 0
    }
}
```

`n.err` is set when `f()` returns an error, but it is never cleared on a subsequent successful call, and it is never overwritten with `nil` when the underlying `f()` succeeds. Every accessor (`Init`, `Start`, `Next`, `Close`) returns `n.err` directly (`scan.go:385, 392, 424, 447`), so once any call fails, every later call appears to fail with the same stale error.

This was the upstream cause of the iterator leak that #4827 fixed by patching the consumers (`invertibleTypeJoin`, `parallelNode`). The leak is contained, but the underlying surprising behavior in `multiScanNode` is still present and is a latent bug factory for any other consumer relying on `Close()`'s return value.

## Proposal

- Capture the return value of `f()` separately from any prior error.
- Either clear `n.err` on a successful call, or store one error per phase (`initErr`, `startErr`, `closeErr`) so `Close()` reports its own outcome rather than a stuck `Init` error.
- Add unit tests pinning the contract: a successful `Close()` after a failed `Init()` should return `nil`, not the `Init` error.

## Refs

- #4814
- #4827
- [Original analysis comment by @edjroz](https://github.com/sourcenetwork/defradb/issues/4814#issuecomment-4626730683)

edjroz:
## Relevant issue(s)

Resolves #4897

## Description
Remove the upstream cause: `multiScanNode` returning a stale error from every accessor after any one of them failed.

### The fix

  - Drop the single shared `n.err` field.
  - Add per-accessor cached error/result fields (`initErr`, `startErr`,`nextErr`, `closeErr`, plus the existing `nextResult`).
  - Remove `countAndCall` — it conflated cycle bookkeeping with error stickiness. Inline each accessor so the cached value is written cleanly overwrites a prior error.

### Note
This is a contract tightening on `multiScanNode`. Pre-fix, `Close() != nil` could mean "Close itself failed" OR "some earlier
phase set `n.err` and it's still stuck." 

Post-fix, `Close() != nil` means Close actually failed. No consumer in `internal/` branches on `multiScanNode.Close() == nil` — all use `errors.Join` / `errors.Is`. Safe contract change.

## Tasks

## How has this been tested?

Three new tests appended to `internal/planner/close_propagation_test.go` (the file PR #4827 introduced), plus a small `scriptedPlanNode` stub extension for scripted multi-call `Next()` sequences:

- `TestMultiScanNode_Close_ReturnsNilAfterFailedInit` — pins that a successful `Close()` after a failed `Init()` returns nil, not the stuck Init error.
- `TestMultiScanNode_Next_ClearsPriorErrorOnNextCycle` — pins that a stuck error from cycle N does not leak into cycle N+1.
- `TestMultiScanNode_ReadersShareSameResultWithinCycle` — regression net: pins the existing buffered-fan-out behaviour so the refactor cannot accidentally break it (already passes on develop; included to prevent regression during the refactor).
  
- MacOS


fredcarle (approved):
LGTM

PR #4911, opened 2026-06-09
fix: TopLevelNode.Close accumulates child errors

### LINKED ISSUE #4898: topLevelNode.Close() short-circuits on first child error, mirroring the bug fixed in #4827
Spotted by @edjroz in [his comment on #4814](https://github.com/sourcenetwork/defradb/issues/4814#issuecomment-4626730683), flagged as out-of-scope of that fix.

`internal/planner/top.go:96-113`:

```go
func (n *topLevelNode) Close() error {
    // ...
    for _, child := range n.children {
        err := child.Close()
        if err != nil {
            return err   // remaining children not closed
        }
    }
    return nil
}
```

This is the same shape as the `parallelNode.Close()` bug that #4827 patched. If the first child's `Close()` returns an error, the remaining children are skipped — any open iterators on those children stay open on the shared transaction, and `txn.Discard()` panics.

In practice `topLevelNode` usually has a single child (`selectTopNode`), so the path is not exercised today. It becomes a real risk if/when the plan tree is constructed with multiple top-level children — which is exactly the kind of change a future feature could introduce silently.

## Proposal

- Iterate every child unconditionally, accumulating errors (same pattern as #4827's fix in `parallelNode`).
- Adopt the test edjroz wrote in his comment (`TestTopLevelNode_Close_CallsAllChildrenEvenIfOneErrors`) under `internal/planner/close_propagation_test.go` alongside the existing #4827 tests.

## Refs

- #4814
- #4827
- [Original analysis comment by @edjroz](https://github.com/sourcenetwork/defradb/issues/4814#issuecomment-4626730683)

edjroz:
## Relevant issue(s)

Resolves #4898

## Description
`topLevelNode.Close` (`internal/planner/top.go:96-113`) iterates children but `return err`s on the first failure. If the first child's `Close()` fails, the loop exits and the remaining children's `Close()` is never called. Any open iterators on those children stay open on the shared transaction, and `txn.Discard()` panics. Same shape as the `parallelNode.Close()` bug PR #4827 patched.

###Notes
Practical exposure today is likely Zero. In current planner construction `topLevelNode` is built with exactly one child (`selectTopNode`), and with one child `return err` after closing it is equivalent to falling out of the loop. The bug becomes live the moment a future change introduces a second top-level child.

Therefore this fix is forward-looking, not reactive. With one child today, the current observable behaviour is unchanged.

Furthermore, this widens `topLevelNode.Close`'s return contract: it can now return an `errors.Join` aggregate (if multiple children fail). Callers that depend on the returned error's concrete type, identity (==), or .Error() string format would require a behavioural change, consumers using errors.Is, errors.As, or if err != nil are unaffected.

## Tasks

## How has this been tested?

  - `TestTopLevelNode_Close_CallsAllChildrenEvenIfOneErrors` — builds  a `topLevelNode` with three children, scripts the first to fail `Close()`, asserts (a) `Close()` returns the propagated error, and (b) all three children's `closeCalled` flag is true. Mirrors
- `TestParallelNode_Close_CallsAllChildrenEvenIfOneErrors` at the top of the same file.

- MacOS


fredcarle (commented):
LGTM

fredcarle (approved):

Issue #4789, opened 2026-05-07
bug(kms+@branchable): KMS authorization can't find collection-version that's still in flight during DAG sync


islamaliev:
## Summary

When two peers sync a `@branchable` collection whose schema has diverged on each side (each ran a `PatchCollection` independently), encrypted blocks created against the *peer-local* collection version can't be authorized on the *other* peer during the very sync that would deliver that collection version. Both peers fail `doesIdentityHaveDocPermission` because `RetrieveCollectionFromDocID` walks to the doc's `CollectionVersionID`, looks it up in the local store, and finds nothing — that version is part of the DAG bytes still being synced.

User-visible error:
```
failed to retrieve encryption key during DAG sync: no peer supplied the encryption key
```

## Reproduction

```
DEFRA_CLIENT_TYPE=go DEFRA_MULTIPLIERS=encrypted-docs go test \
  ./tests/integration/net/sync/branchable_collection \
  -run TestBranchableCollectionSync_WithBranchedVersionsAndDocs_ShouldSync
```

Topology: 2 peers. Each runs an independent `PatchCollection` on a `@branchable` collection, adds docs, then connects and `SyncBranchableCollection`s.

## Trace

Filtered `[kms]` / `[colRetriever]` / `[store.GetCollections]` excerpt for the failing doc `bae-fe627cdb-…`:

```
[kms] fetched encBlock DocID=\"bae-fe627cdb-…\" keyLen=32
> kms.doesIdentityHaveDocPermission
  [kms] doesIdentityHaveDocPermission: documentACP.HasValue=true entBlock.DocID=\"bae-fe627cdb-…\"
  > colRetriever.RetrieveCollectionFromDocID
    [colRetriever] r.ident.HasValue=true docID=bae-fe627cdb-… versionID=bafyreibqvtb7gtuijcpobuld2hhtqmeltnfou6hclkc6gixeggywymk6ou
      [store.GetCollections] identity.HasValue=true versionID={true bafyreibqvtb7gtuijcpobuld2hhtqmeltnfou6hclkc6gixeggywymk6ou}
      [store.GetCollections] getCollections returned len=0 err=failed to get collection by version ID: collection not found. VersionID: bafyreibqvtb…ymk6ou
    [colRetriever] GetCollections returned len=0 err=...collection not found...
  [kms] doesIdentityHaveDocPermission err: ...collection not found...
```

[...]

PR #4914, opened 2026-06-10
test: Add encrypted-docs test multiplier

### LINKED ISSUE #4452: Add test multiplier for encrypted docs

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4452

## Description

This adds an `encrypted-docs` multiplier that mirrors the existing `signed-docs` and `secondary-index` ones: when enabled, every `AddDoc` action of a test that doesn't already configure encryption is flipped to encrypted, so the whole integration suite doubles as encryption coverage.

Tests that already set up encryption themselves are skipped so their intent is preserved. Tests whose assertions depend on byte-level CID identity opt out via `MultiplierExcludes` with an inline reason or tracking issue. One test pinned to https://github.com/sourcenetwork/defradb/issues/4789 is added as a `t.Skip`-gated reproducer that will start passing once that bug is fixed.

Activate locally with `DEFRA_MULTIPLIERS=encrypted-docs`.

## How has this been tested?

Unit tests for the multiplier's `Apply` and `ShouldSkip`. Full integration suites in `tests/integration/{subscription,encryption,query/simple,query/commits,query/one_to_many,net/simple/peer,net/sync/branchable_collection}` run cleanly both with and without `DEFRA_MULTIPLIERS=encrypted-docs`.

- MacOS


AndrewSisley (approved):
LGTM - is nice to see more cids being abstracted away, thanks Islam :)

Issue #4907, opened 2026-06-08
Cannot add an index larger than ~11MB


AndrewSisley:
Size limit applies when defining the index - an index can grow to larger than this over time.

It is also impossible to delete or rebuild an index larger than this value.

This is due to badger transaction size limits.

Resolving this might take some re-imagining of the way indexes are constructed, as if we cannot use the same txn that is used to mutate the collection definition to build the index, then constructing the index probably needs to happen in another user-call, otherwise the index might be mutated without the collection definition and other oddities.

Adding/removing an index is different to patching a collection, in that the definition is/can-be added via the txn *before* the txn free changes, so for implicit transactions, it would be possible.  However for the building of explicit (user-provided) transactions, which cannot be committed mid-operation, it is not.

It might be reasonable to forbid explicit transactions for index definition, but this itself is an oddity as even the patch system supports this (and that system allows patching indexes, providing an additional dimension to the inconsistency - patching indexes would probably need a different flow or restrictions to the high level index ops). We could disable managing indexes via patch in the short term before switching to another approach long term, but that is a severe inconsistency and loss IMO.

It *would* be possible to sort of handle this in the definition transaction scope, after https://github.com/sourcenetwork/defradb/issues/4873 (should that system also accommodate secondary index building), and a change to `txn.OnSuccess` to allow a return of an error:
1. The txn locks immediately upon beginning the operation, before definition changes 
2. The txn performs the definition changes
3a. On txn commit error (outside of the Add/remove index operation if explicit txn), rebuild the index using the old definition (txn rolled back) without txn protection (due to limit)
3b. On commit success (outside of the Add/remove index operation if explicit txn), rebuild the index using the new definition without txn protection (due to limit)
4a. If either 3a or 3b fails, the error is returned to the user (on commit) and the collection lock is maintained until it succeeds https://github.com/sourcenetwork/defradb/issues/4873 - ideally, this will need to permit users to amend/remove the index in case it gets stuck, due to corruption/whatever, to that they can regain access to the collection, minus the index if they really need - this may get complicated due to the lock held
4b.  If 3a or 3b succeeds, release the lock 
- Synchronicity of 3a/b should ideally be controlled by the user, it doesnt *need* to be synchronous - see also https://github.com/sourcenetwork/defradb/issues/4888 and https://github.com/sourcenetwork/defradb/issues/4874
- This flow could be ported to materialized views to allow them to auto-truncate when dematerialized via patch, an improvement, and a consistency fix.
[...]

Issue #4888, opened 2026-06-04
Add means to cancel long running operations


AndrewSisley:
Such as truncate, index construction, subscriptions and materialized view refreshes.

This might already be somewhat handled, if the context is cancelled the store operations *might* error.  It should be tested though, and maybe handled more gracefully than whatever the current behaviour is.

Consider making this possible without direct access to the context, via https://github.com/sourcenetwork/defradb/issues/4872

And be careful with the security implications of this (NAC might be sufficient for now?).

suggestion: Consider creating events for the start and completion of these actions (if nothing else it might make testing better).

Issue #4874, opened 2026-06-02
Resume ongoing truncates and view refreshes on restart


AndrewSisley:
Ongoing truncates/view-refreshes include truncates/refreshes that errored out previously.

If an auto-resumed truncate/refresh fails, it should not cause the restart to fail. The truncates/refreshes do not need to block the restart.

Depends on https://github.com/sourcenetwork/defradb/issues/4873

Issue #4873, opened 2026-06-02
Maintain lock on collection if action fails


AndrewSisley:
Maintain the write lock on collection if truncate or view refresh fails, until it succeeds.  This lock needs to survive a database restart (for file based stores).

Keep https://github.com/sourcenetwork/defradb/issues/4872 in mind, it might be sensible to designed any keys needed for this ticket to also be useful for that ticket.

PR #4922, opened 2026-06-12
feat: Add Action system

### LINKED ISSUE #4872: Add function to return list of ongoing long running actions
Add function to return list of ongoing truncates.  A truncate that previously errored counts as ongoing, although a status should indicate this.

This can probably also accommodate other long running lock holding functions such as materialized views https://github.com/sourcenetwork/defradb/issues/4386

AndrewSisley:
## Relevant issue(s)

Resolves #4872

## Description

Adds a very basic Action system, allowing us and users to track long running actions.

Adds the system to truncate and view refresh.  Does not add it to the adding/deletion of indexes, as those functions have not yet been reworked to allow for large data volumes.

~~I am on holiday for the first 3 days of next week, and this is needed to get secondary indexes working, so I've kept it very basic for now.  We probably want to be creating events for the start and complete of these actions at some point, it will make it easy for users and our tests to listen for status changes.~~ Now a todo for this PR, timing the CI is proving too unreliable, lots of variation.

## Related issues requiring this

- https://github.com/sourcenetwork/defradb/issues/4888
- https://github.com/sourcenetwork/defradb/issues/4873
- https://github.com/sourcenetwork/defradb/issues/4874
- https://github.com/sourcenetwork/defradb/issues/4907

## Todo

- [x] Set error status for truncate/refresh on err
- [x] Use events for start/finish of run, and listen in tests


internal/keys/systemstore_action_status.go
```diff
@@ -0,0 +1,78 @@
...
+// ActionStatusKey points to the current status of an action execution.
+//
+// It is stored in the following format:
+// [CollectionID]/[Action] => [ActionStatus]
+type ActionStatusKey struct {
+	CollectionID string
+	Action       client.Action
+}
+
+var _ Key = (*CollectionKey)(nil)
```
islamaliev: todo: var _ Key = (*ActionStatusKey)(nil)

internal/db/action/action.go
```diff
@@ -0,0 +1,121 @@
...
+	binary.PutUvarint(val, uint64(status))
+
+	err := multistore.Systemstore().Set(
+		ctx,
+		keys.NewActionStatusKey(collectionID, action).Bytes(),
+		val,
+	)
+
+	return err
+}
```
islamaliev: todo: this basically limits us to only 1 action type per collection. You might have unlimited number of indexes running in parallel on a single collection

client/action.go
```diff
@@ -0,0 +1,42 @@
...
+	RefreshDatastoreAction Action = 2
+)
+
+// ActionStatus describes the current status of an action execution.
+type ActionStatus uint16
+
+const (
+	NoneActionStatus       ActionStatus = 0
+	InProgressActionStatus ActionStatus = 1
+	ErroredActionStatus    ActionStatus = 2
```
islamaliev, line 29: todo: I think it's important to store failure reason along with the err marker otherwise troubleshooting is going to be hell.

islamaliev (dismissed):

client/action.go
AndrewSisley: This can be added later without affecting design.  I want this merged as soon as possible as I'm off for the first half of next week and dont want to block anyone from picking up the index issue.

internal/keys/systemstore_action_status.go
AndrewSisley: Cheers :)

- [x] Fix type assert

internal/db/action/action.go
AndrewSisley: Is a really good catch, cheers Islam.

Thinking this through a bit, I think my preference is to add a `Context immutable.Option[string]` property to the key, so stuff like secondary indexes can handle multiple, whilst still keeping the keys deterministic (sec. indexes could set it to the index name/id perhaps).

However, as this is not needed for truncate and refresh, and I want to avoid adding it to secondary indexes until the PR that handles large indexes, I don't have a good enough way of testing that until then.

What are your thoughts/preferences?  Are you happy with that solution? Are you happy with deferring the implementation?  Or do you want it implemented now, and not properly test it until secondary indexes gets sorted?

client/action.go
```diff
@@ -0,0 +1,42 @@
...
+type ActionExecution struct {
+	// The ID of the collection that this action is executing on.
+	CollectionID string
+
+	// The current action being executed.
+	Action Action
+
+	// The current status of this action execution.
+	Status ActionStatus
+}
```
islamaliev, line 43: todo: we need a way to store additional action-specific payload. Indexes for examples need what I currently call watermark which a sort of cursor to continue backfilling from when a batch fails.

Something like this (just a idea, did't think through yet):
```go
type actionRecord struct {
    Status client.ActionStatus
    // Reason is set when Status is Errored.
    Reason string
    // Payload is action-specific state, opaque to the action system.
    Payload json.RawMessage
}
```

  with typed helpers so callers never touch raw JSON:

```go
func Set[T any](ctx, ms, collectionID, action, status ActionStatus, payload T) error
func GetPayload[T any](ctx, ms, collectionID, action) (T, ActionStatus, error)
```

AndrewSisley: ~~Transferring this to the other thread, I want to keep this conversation in one place~~ I misread, one sec :)

AndrewSisley: I don't think this should be stored in the same key-type.  I'm going to leave this out of this PR, as unless there is agreement to store all action-execution related data as a json value in one key, then it is not required within this PR.

Note the existing key is prefixed with `a/s`, the `s` (and key type name) indicating that this key is for the status only, and is not to be used for additional metadata.

AndrewSisley: Do you feel like the Key documentation needs to make this clearer?

AndrewSisley: My literacy skills are really lacking today.  You commented on the `client` result object, not the key...

I have no problem with this object being extended in the future, but it has no reason to do so now, in this PR.    Error message is a requested feature, and anything that indexes needs can also be added in some form.

As this object is not coupled to the storage format, I don't think this needs to be decided upon right now, and it can easily be extended as and when needed (i.e. in the future secondary index PR).

tests/integration/collection/refresh/list_actions_test.go
```diff
@@ -0,0 +1,111 @@
...
+					// Create enough users to allow the `ListActions` action to complete before the RefreshViews.
+					gen.WithTypeDemand("Users", 500),
+				},
+			},
+			&action.Async{
+				Child: &action.RefreshViews{},
+			},
+			testUtils.Wait{
+				// Delay the start of the `ListActions` action until the view refresh has registered the action.
+				Duration: time.Millisecond * 8,
```
islamaliev: suggestion: magic number. Would be nice to have something more descriptive.

I actually if we can come up with something more deterministic instead of guessing when the action starts

tests/integration/collection/truncate/simple_test.go
```diff
@@ -36,3 +36,21 @@ func TestTruncateCollection(t *testing.T) {
...
+					}
+				`,
+			},
+			&action.Truncate{},
+			&action.Truncate{},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
```
islamaliev, line 56: question: seems to be unrelated to the change. Why is it here?

internal/db/view.go
```diff
@@ -130,6 +131,16 @@ func (db *DB) refreshViews(ctx context.Context, opts *options.GetCollectionsOpti
 		}
 		db.lockSet.CollectionLock(txn, shortID)
 
+		multistore := datastore.NewMultistore(db.rootstore, db.lockSet, db.blockStoreChunkSize)
+
+		// Registerd using the background context, as stores rely on the closing of other contexts in order
+		// to time the flushing of writes - meaning if ctx was used to write, the status would not be readable
+		// by other threads until after the refresh has completed.
+		err = action.Register(context.Background(), multistore, col.CollectionID, client.RefreshDatastoreAction)
```
islamaliev: question: just wondering what happens if Truncate is called twice. Don't we forbid it?

internal/db/action/action.go
islamaliev: It's fine if you keep it simple for now. I can extend if for sec. index needs. But additional `immutable.Option[string]` should help

client/action.go
islamaliev: no problem. Keep this PR simple.

tests/integration/collection/refresh/list_actions_test.go
AndrewSisley: This is getting ripped out in this PR (see todo section in PR), it is too unreliable in the CI.

tests/integration/collection/truncate/simple_test.go
AndrewSisley: It makes sure that the truncate can run twice - because now we are writing unique keys, and erroring if one already exists, it becomes important to check that this still works.

internal/db/view.go
AndrewSisley: Truncate can be called twice serially.

If it is called twice concurrently, the collection lock will prevent the second Register call from beginning until the first lock is released, at which point the first action status will have have been deleted (on success), or updated to cancel/error - in which case the second register will deliberately overwrite/reset it to `InProgress`.

AndrewSisley:
I have run out of time today, the code is how I think I want it, but the documentation is lacking a bit around the event system and new test actions - it would be good to improve this before merge.

Happy for someone to take this and finish it off whilst I'm off if needed instead of blocking other dependent tasks (or just base off of this and wait for me to cleanup/merge on Wed/Thurs) - sorry for the bother if that is required.

cli/action.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+func MakeActionCommand(ctx context.Context) *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "action",
+		Short: "Manage DefraDB actions",
+		Long: `Manage DefraDB actions.
+Manage long running actions such as truncate, RefreshView, and the (re)building of indexes.
```
fredcarle, line 24: thought: Maybe we should highlight that this command isn't used to `create` those actions. Because `Manage actions` may lead to thinking that it does.

client/action.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// ActionStatus describes the current status of an action execution.
+type ActionStatus uint16
+
+const (
+	NoneActionStatus       ActionStatus = 0
+	InProgressActionStatus ActionStatus = 1
+	ErroredActionStatus    ActionStatus = 2
+	CompletedActionStatus  ActionStatus = 2
```
fredcarle: todo: I'm assuming that this is a mistake so this todo is based on that assumption. Please make sure that `ErroredActionStatus` and `CompletedActionStatus` are not the same value.

internal/db/collection_truncate.go
```diff
@@ -53,11 +55,33 @@ func (c *collection) Truncate(
...
+		return err
+	}
+
+	c.db.lockSet.CollectionLock(txn, shortID)
+
+	multistore := datastore.NewMultistore(c.db.rootstore, c.db.lockSet, c.db.blockStoreChunkSize)
+
+	// Registerd using the background context, as stores rely on the closing of other contexts in order
+	// to time the flushing of writes - meaning if ctx was used to write, the status would not be readable
+	// by other threads until after the truncate has completed.
```
fredcarle: thought: That comment is innacurate. The closing of the context doesn't affect flushing of writes. The transaction does. I think we might have a bug where we think that the txn isn't passed to `corekv` because we use the `rootstore` when defining the `multistore` but `corekv` will still extract the transaction if available on the context.

todo: Please update the comment here and in `view.go` to be representative of what is happening.

internal/db/action/action.go
```diff
@@ -0,0 +1,163 @@
...
+	val := make([]byte, binary.MaxVarintLen16)
+	binary.PutUvarint(val, uint64(status))
+
+	err := multistore.Systemstore().Set(
+		ctx,
+		keys.NewActionStatusKey(collectionID, action).Bytes(),
+		val,
+	)
+
+	events.Publish(event.NewMessage(event.ActionExecutionName, event.ActionExecution{
```
fredcarle, line 64: thought: This probably shouldn't publish if `Set` errors.

internal/db/action/action.go
```diff
@@ -0,0 +1,163 @@
...
+	events event.Bus,
+	collectionID string,
+	action client.Action,
+) error {
+	err := multistore.Systemstore().Delete(
+		ctx,
+		keys.NewActionStatusKey(collectionID, action).Bytes(),
+	)
+
+	events.Publish(event.NewMessage(event.ActionExecutionName, event.ActionExecution{
```
fredcarle, line 89: thought: Same as above. This probably shouldn't publish if `Delete` errors.

tests/clients/cli/wrapper.go
```diff
@@ -393,6 +393,24 @@ func (w *Wrapper) BasicExport(
 	return err
 }
 
+func (w *Wrapper) ListActions(
+	ctx context.Context,
+	opts ...options.Enumerable[options.ListActionsOptions],
+) ([]client.ActionExecution, error) {
+	args := []string{"client", "action", "list"}
+
+	data, err := w.cmd.execute(ctx, args)
```
fredcarle, line 405: todo: Please add the following before `execute`:
```go
opt := utils.NewOptions(opts...)
args = appendIdentityArg(args, opt.GetIdentity())
```

tests/action/wait.go
```diff
@@ -0,0 +1,63 @@
...
+var _ Action = (*Wait)(nil)
+var _ Stateful = (*Wait)(nil)
+
+func (a *Wait) Execute() {
+	for {
+		select {
+		case <-time.After(orDefault(a.Duration, time.Hour)):
+			if a.Duration.HasValue() {
+				return
+			}
```
fredcarle: question: Why not always return?

event/event.go
```diff
@@ -183,3 +184,14 @@ type TopicPeerEvent struct {
 	// EventType is the type of event: "JOINED" or "LEFT".
 	EventType string
 }
+
+type ActionExecution struct {
```
fredcarle, line 188: thought: Since this is exactly the same at `client.ActionExecution`, it might make sense to just embed it here instead of redefining the same fields and risking oversight bugs in the future.

fredcarle (approved):
Approving assuming all comments will be resolved before merge.

internal/db/collection_truncate.go
AndrewSisley: Ah yes.... I've hit that before and I think there is a ticket open for it, I'll link to it if I find it.  Thanks Fred!

- [x] Fix docs

client/action.go
AndrewSisley: Fix enum values

 - [x] Fix enum values

cli/action.go
AndrewSisley: I think it can be used for that in the future, at least to re-run failed actions.  Leaving docs as-is unless the conversation leads me elsewhere :)

internal/db/action/action.go
AndrewSisley: Covered by the todo in the PR description, resolving this conversation.

internal/db/action/action.go
AndrewSisley: Covered by the todo in the PR description, resolving this conversation.

tests/clients/cli/wrapper.go
AndrewSisley: Might have a test bug somewhere else, I copy-pasted this in.  I'll see if I can spot anything before fixing.

- [x] Fix cli wrapper opts

tests/action/wait.go
AndrewSisley: True, will simplify.

- [x] Simplify timeout

event/event.go
AndrewSisley: I defined it as such originally, and then deliberately flattened it.  I don't think they should be coupled in such a way, especially not for public items (decoupling them, for example in order to add a field to `client.ActionExecution` but not to the event becomes a breaking change), and it also comes with the baggage of *looking* like they should be coupled together for important reasons, when in fact they just happen to have the same properties right now.

cli/action.go
fredcarle: Curious to know what @ribiza thinks about this :)

ribiza: I'm late to the party - I don't have a strong opinion anyway. `Manage` implies that a _bunch_ of actions should be possible, which at this stage is not the case as only `list` is implemented. All adults are born children though and that's okay :) I don't expect to be able to _create_ if I'm told I can _manage_ something.

PR #4923, opened 2026-06-12
chore: Update aggregate return types for non-nillable fields

### LINKED ISSUE #4920: Update aggregate return types
The following PR (not yet merged, at the time of my writing this) introduces field kinds for non-nillable scalars: https://github.com/sourcenetwork/defradb/pull/4895/

This issue is a follow-up for that PR. We should update the return types of the aggregates that are returned, but only where appropriate.

 `_max` and `_min` should be nillable
`_count`, `_avg` and `_sum` should not be nillable.

To resolve this issue, the changes must be made to utilize the appropriate return types. However, the existing introspection tests must also be updated, because this will be a breaking change.

This is marked as a v1 issue because of its breaking-change nature.

ChrisBQu:
## Relevant issue(s)

Resolves #4920 

## Description

The core change can be seen inside `generate.go` and are minimal. Instead of `gql.Int` and `gql.Float`, we will return `gql.NewNonNull(gql.Int)` and `gql.NewNonNull(gql.Float)`. This, unfortunately, has widespread breaking changes across many of our tests. The reeason for this, is that before we expected a result in the form:

`{kind: SCALAR, name: "Int"}`

But with these changes, it will now be wrapped as: 

`kind: NON_NULL, name: null, ofType: {kind: SCALAR, name: "Int"}}`

All instances of tests expecting the old shape need to be updated.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/internal/request/graphql/schema/generate.go
+++ b/internal/request/graphql/schema/generate.go
@@ -225,7 +225,7 @@ func (g *Generator) generate(ctx context.Context, collections []client.Collectio
 			if err := g.expandInputArgument(obj.OfType.(*gql.Object)); err != nil {
 				return nil, err
 			}
-		case *gql.Scalar:
+		case *gql.Scalar, *gql.NonNull:
 			if _, isAggregate := request.Aggregates[def.Name]; isAggregate {
 				for name, aggregateTarget := range def.Args {
 					expandedField := &gql.InputObjectFieldConfig{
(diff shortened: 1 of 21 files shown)


internal/request/graphql/schema/generate.go
```diff
@@ -344,15 +344,12 @@ func (g *Generator) expandInputArgument(obj *gql.Object) error {
 				}
 				obj.AddFieldConfig(f, expandedField)
 			}
-		case *gql.Scalar:
+		case *gql.Scalar, *gql.NonNull:
```
fredcarle, line 349: todo: Please add a comment for future readers describing why having `NonNull` here is safe.

fredcarle (approved):
LGTM. Just one minor doc todo before merge.

PR #4926, opened 2026-06-15
fix: Fix syncing of collections with no inactive version

### LINKED ISSUE #4141: Impossible to test subscription to collection with no active versions
E.g. a collection that has synced over P2P but has not yet been activated.

Test code panics on line 118 of `tests/integration/p2p_collection.go`.  I dont know if the prod code works.

ChrisBQu:
## Relevant issue(s)

Resolves #4141 

## Description

This PR adjusts `internal/db/p2p/collection.go` by altering the `AddP2PCollections` and `DeleteP2PCollections` functions. It adjusts them such that the GetInactive flag is true, on the collections that are obtained. Then, when the stored collections are examined, we deduplicate by CollectionID before proceeding. This is necessary because querying by name with GetInactive set to true returns all versions of a collection, and all versions share the same CollectionID. Without this de-duplication step DeleteP2PCollections, for example, would attempt to delete the same key twice. 

This PR also adjusts `tests/integration/p2p_collection.go` by altering the `addCollectionSubscription`, `deleteCollectionSubscription`, and `listP2PColleections` functions. It does so by making it such that the collection names use `col.Name()` only if the collections are not nil. If they are nil, it falls back to the name that was passed during creation of the collection.

Several new integration tests are included to demonstrate this functionality behaves as expected.

## Tasks

## How has this been tested?

- WSL


fredcarle (approved):
LGTM

Issue #4893, opened 2026-06-05
bug(http+auth): all token-verify failures collapse to an opaque 403 "forbidden"


islamaliev:
**Describe the problem**

The HTTP auth middleware writes the same bare `forbidden` body and 403 status for every distinct token-verify failure: audience does not match the request host, no audience claim on the token, the token is malformed, or the signature is invalid. The underlying error from the identity package is discarded before it reaches the client, so callers have no way to tell a config mistake (wrong audience, missing claim) from a token-integrity problem.

In practice this was misdiagnosed externally as a node-startup race: a provisioning tool POSTing to `/api/v1/collections` and `/api/v1/acp/document/policy` saw the 403 immediately after node start, retried with backoff, and eventually one attempt happened to use a host matching the token's audience. Verification is deterministic per (token, host); there is no warm-up window.

**To Reproduce**

Mint a bearer token with audience `localhost:9181`. POST any auth-gated endpoint with `Host: 127.0.0.1:9181`. The middleware returns `403` with body `forbidden`. Same token against `Host: localhost:9181` passes the middleware. The same opaque body also appears for a token with no `aud` claim and for a non-JWT bearer value. Coverage tests for each failure mode are added in https://github.com/sourcenetwork/defradb/pull/4892.

**Expected behavior**

The 403 body distinguishes between the public causes the operator can act on — audience/host mismatch, missing audience, invalid token — without leaking signature internals. A stable, documented reason string would let `provision_acp`-style tooling stop retrying and stop guessing.

**Platform**

DefraDB `develop` (see linked PR for the exact commit range covered by the tests). Reproduced locally on macOS arm64; not platform-specific.

**Additional context**

The bare `forbidden` body comes from the auth middleware before any handler or ACP check runs, so this is unrelated to ACP/grant-cache readiness. The JSON `{"error": "..."}` body produced by ACP permission errors elsewhere in the API uses the existing error-to-status mapping and is not affected.

Tests pinning the current behaviour: https://github.com/sourcenetwork/defradb/pull/4892

ribiza:
@islamaliev something else to be done here, or has #4934 addressed this?

islamaliev:
this one id done

PR #4929, opened 2026-06-16
fix: Derive CLI auth token audience from the API URL

### LINKED ISSUE #4846: CLI JWT audience does not match URL
From build week feedback: https://discord.com/channels/1374835078781468722/1506975461043863593

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4846
Refs https://github.com/sourcenetwork/defradb/issues/4893

## Description

When the CLI minted an auth token, it set the audience to the raw configured API address — scheme and casing included — but the server validates a token's audience against the lowercased request host it actually receives. Any address that carried a scheme or uppercase letters, or simply differed between how the server was started and how the client dialed it, produced a mismatch, so authenticated commands were silently rejected as forbidden. The audience is now derived from the same address the client dials so the two always agree, with a new `--audience` flag to override it when the dialed host differs from the one the server expects.

This is the client side of the audience/host mismatch behind https://github.com/sourcenetwork/defradb/issues/4893; that issue also asks for the server's opaque 403 to spell out the cause, which is left for separate work.

While here, the host parsing that both the audience and the dialed client share no longer misreads a scheme-less host that happens to begin with "http", and an address that yields no host now fails with a clear error instead of minting a token that would be rejected.

## Tasks

## How has this been tested?

Unit tests covering audience derivation across scheme, casing, default-port, IPv6, and scheme-less hosts, plus the error paths, and a test pinning the derived audience to the host the client actually dials. CLI integration tests cover the full round-trip: an authenticated command with the auto-derived audience is accepted, and one with a deliberately mismatched audience is rejected by the server.

- MacOS


nasdf:
@islamaliev  I think we may want to use the `allowed-origins` argument as an allow list for the audience values. In most deployments these values would overlap, and this should fix the linked issue from the server's perspective instead.

nasdf (approved):
Can you add a note to the `--allowed-origins` flag about its use within the audience check? Otherwise code changes look great!

cli/utils.go
```diff
@@ -149,14 +149,24 @@ func setContextIdentity(cmd *cobra.Command, privateKeyHex string) error {
...
+
 	privKey := secp256k1.PrivKeyFromBytes(data)
 	ident, err := acpIdentity.FromPrivateKey(crypto.NewPrivateKey(privKey))
 	if err != nil {
 		return err
 	}
 	err = ident.UpdateToken(
 		authTokenExpiration,
-		immutable.Some(cfg.GetString("api.address")),
+		immutable.Some(audience),
```
fredcarle, line 169: thought: The `audience` should probably be forced to lowercase to match what we do with the `Host` address on comparison. Either here or at the comparison site.

http/auth.go
```diff
@@ -57,29 +57,55 @@ const (
...
+	}
+}
+
+// audiencesForOrigins normalizes CORS origins into the audience (host) form an
+// auth token would carry, so they can be compared against a token's audience
+// claim. Origins that do not yield a host are skipped.
+func audiencesForOrigins(origins []string) []string {
+	audiences := make([]string, 0, len(origins))
+	for _, origin := range origins {
+		audience, err := AuthAudienceForURL(origin)
```
fredcarle, line 102: thought: `origin` would be `*` which is not really valid for the audience. We should make sure we skip `*`.

fredcarle (commented):
Late but a couple comments that might require further work.

PR #4932, opened 2026-06-16
docs: Clarify `keyring new` command behavior

### LINKED ISSUE #4704: CLI command `keyring new` docs inaccurate

ribiza:
## Relevant issue(s)

Resolves #4704.

### DIFF
--- a/cli/keyring_new.go
+++ b/cli/keyring_new.go
@@ -32,11 +32,9 @@ func MakeKeyringNewCommand(ctx context.Context) *cobra.Command {
 Randomly generate and store private keys in the keyring.
 By default peer and encryption keys will be generated.
 
-The DEFRA_KEYRING_SECRET environment variable must be set to unlock the keyring.
+The DEFRA_KEYRING_SECRET environment variable must be set to initialize the keyring.
 This can also be done with a .env file in the working directory or at a path
-defined with the --secret-file flag.
-
-WARNING: This will overwrite existing keys in the keyring.`,
+defined with the --secret-file flag.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			k, err := openKeyring(cmd)
 			if err != nil {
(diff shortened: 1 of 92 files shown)


cli/keyring_new.go
```diff
@@ -32,11 +32,9 @@ func MakeKeyringNewCommand(ctx context.Context) *cobra.Command {
 Randomly generate and store private keys in the keyring.
 By default peer and encryption keys will be generated.
 
-The DEFRA_KEYRING_SECRET environment variable must be set to unlock the keyring.
+The DEFRA_KEYRING_SECRET environment variable must be set to initialize the keyring.
 This can also be done with a .env file in the working directory or at a path
-defined with the --secret-file flag.
-
-WARNING: This will overwrite existing keys in the keyring.`,
```
shahzadlone, line 39: Question: shouldn't the warning still be shown for the force flag?

shahzadlone (approved):
LGTM, just one comment

cli/keyring_new.go
ribiza: We could, but that would imply that users expect a confirmation on a force flag. But `rm -rf` doesn't ask for confirmation, nor does `git push -f`. Adding the force flag requires an extra action and is thus an intentional decision: I find any further confirmation being just a chore, because at that point users must already be sure that they want to do it.

PR #4933, opened 2026-06-16
chore: Prefer docID doc update/delete in c bindings

### LINKED ISSUE #4829: C binding document update/delete functions behave differently
In the case of both a filter and a docID being present, the C binding functions `UpdateDocument` and `DeleteDocument` will behave differently. The `UpdateDocument` function first checks for the filter, and if present, does the update. However, `DeleteDocument` first checks for the DocID.

The same is true in the CLI versions of these functions.

The behavior should be made the same, either with a canonical order of precedence.

ChrisBQu:
## Relevant issue(s)

Resolves #4829

## Description

Previously, the `DeleteDocument` and `UpdateDocument` functions in the C bindings behaved a little bit differently from one another. In the case of both a filter and docID being provided, one would rely on the docID to perform the operation, while the other would rely on the filter. This makes them consistent, in that they will both prioritize the docID over a filter, as was discussed in our team's standup meeting.

This fix is simple, and is expected to affect no other parts of the code base.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/document_delete.go
+++ b/cbindings/document_delete.go
@@ -58,6 +58,8 @@ func DeleteDocument(nodePtr C.uintptr_t,
 	docID := C.GoString(docIDStr)
 	filter := C.GoString(filterStr)
 	switch {
+
+	// If docID is provided, delete the document by docID
 	case docID != "":
 		ID, err := client.NewDocIDFromString(docID)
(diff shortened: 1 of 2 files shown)


fredcarle (approved):
LGTM

PR #4934, opened 2026-06-16
fix: Make auth token rejection 403s describe their cause

### LINKED ISSUE #4924: Helpful error messages when invalid HTTP `Authorization` header provided
Right now any wrongdoing in crafting the `Authorization: Bearer <jwtToken>` results in a 403 HTTP status with body `forbidden`. It would be massively helpful for DX if the message wasn't just `forbidden` for any attempt.

We should for example check that:
- the token length looks right, 64 char long I think (i.e. people are not trying to use a public key rather than a jwt token)
- the header (`alg` and `typ`) holds supported values
- the payload contains all expected fields (as far as I can tell, `sub`, `aud`, `exp`, `nbf`) and that they are of the right type and contain sensible values (a hostname for `aud`; correct expirations for `exp` and `nbf`; a DID key for `sub`)
- the signing key matches the payload public key

It would be fantastic if each of these fails returned a different message as HTTP content alongside the 403 status. It would help users correct mistakes rather than crafting a payload and hoping they've done everything right (and cry if not). No security is compromised by error messages clarifying what part of the public API specs was not fulfilled.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4924

## Description

Every way an auth token could fail verification produced the same blank "forbidden" response, so an operator had no way to tell a fixable mistake from a broken token. This was even misread once as a startup race, because a provisioning tool kept retrying the opaque rejection until a request happened to land on a host that matched the token. The rejection now names the cause for the mistakes an operator can actually act on: an audience that doesn't match the request host, a missing audience, an expired token, one that isn't valid yet, a subject that isn't a usable public key, and a missing or malformed key type. Signature failures stay deliberately vague so nothing about key validity leaks, and a token that simply can't be parsed reports as invalid. The reason text is treated as a stable contract that automation can branch on instead of guessing.

To make those causes distinguishable, token construction now only reads the identity it carries, and all of the validity checks happen in one place during verification, so each failure can be reported with its own reason.

## Tasks

## How has this been tested?

Unit tests assert that verification reports the correct distinct cause for each failure — audience mismatch, missing audience, expired, not yet valid, invalid subject, missing key type, and bad signature. Middleware tests confirm each cause produces its documented 403 response, that signature failures and unparseable tokens both fall back to the generic reason, and that a valid token on a matching host still passes.

- MacOS


nasdf (approved):
Thanks for improving the error messages!

PR #4936, opened 2026-06-16
fix(merge): only process composite blocks as merge heads (counter double-apply, #4935)

### LINKED ISSUE #4935: Counter multiply-applied / runaway on merge when field blocks are pushed as PushLog heads (ProcessBlock not idempotent)
## Summary

A counter (PN/P-counter) can be **multiply applied** during merge — sometimes running away to hundreds/thousands of times the correct value — when a peer delivers individual **field blocks as PushLog heads** (i.e. each DAG block pushed as its own `pushLogRequest`, as `push_existing_docs`/`retry`-style full-DAG replication does). The materialized counter value diverges permanently even though the commit DAG is identical across nodes.

Root cause: `coreblock.ProcessBlock` applies the delta (`crdt.Merge`) **unconditionally**, with no per-block "already merged" guard inside the apply. Counter idempotency relies entirely on the merge processor never *processing* a block twice — but when `db.Merge` is invoked with a **field-block** CID as the head, `mergeProcessor.loadComposites` (which only walks `Heads`) collects that field block and its predecessor field chain into the composite-processing list, and `mergeComposites`→`processBlock` then applies each field delta **directly** — in addition to the application that happens via the field block's owning composite's link recursion. Net: each affected counter field delta is applied twice. With the post-merge relay (`SendUpdate` after `db.Merge`), the extra applications can compound into a runaway.

## Where

- `internal/core/block/store.go::ProcessBlock` — `crdt.Merge(...)` with no `IsMerged` guard; `internal/core/crdt/counter.go::incrementValue` is a pure RMW (no nonce/CID dedup).
- `internal/db/merge.go::executeMerge`→`loadComposites`/`mergeComposites`/`processBlock` — no `IsMerged` check before applying; `loadComposites` follows `Heads` only, so a field-block head pulls the whole field chain into the processing list.

## Reproduction

Concurrent same-document counter increments across a 3-node mesh where at least one peer pushes the document DAG **block-by-block** (every block as its own `pushLogRequest`, including field blocks) — as the replicator backfill/retry path does. Observed counter values of 310, 4534, 16185 from 24 increments; also small over-counts (+1..+19) and occasional under-counts. Intermittent: gated by whether the field-block head is processed before its composite marks the field merged (the `IsMerged(headCID)` check in `processPushlogRequest` races the composite-merge path).

(We hit this reliably when a non-Go peer in the mesh pushes per-block heads; pure Go↔Go is far rarer because of delivery timing, but the same code path applies — Go's own `push_existing_docs`/retry sends the full DAG block-by-block, so a field block can become a merge head.)

## Suggested fixes

1. **Make application idempotent (primary).** Guard `ProcessBlock` on `txn.Blockstore().IsMerged(blockLink.Cid)` before `crdt.Merge`, and/or carry a visited-set through the merge walk. Closes the bug regardless of how a field block enters the merge.
2. **Don't push field blocks as heads (secondary).** Replication backfill/retry should push the document's **composite head(s)** only and let the receiver fetch linked blocks via DAG sync (matching `SendUpdate`), rather than sending every DAG block as its own `pushLogRequest`.

Happy to provide full instrumented traces (per-block apply counts, `loadComposites` collection of the field chain, and the field-head receive logs).

jackzampolin:
## What

Fixes #4935 — a counter (P/PN-counter) is **multiply applied** during merge when a field block arrives as a PushLog head, sometimes running away to hundreds/thousands of times the correct value, with permanent divergence despite an identical commit DAG.

## Why

`mergeProcessor.loadComposites` collects whatever block the merge head resolves to into the composite-processing list. When that head is a **field block** (e.g. a counter delta block delivered standalone — as happens when a peer pushes the document DAG block-by-block, every block as its own `pushLogRequest`), `processBlock` applies its delta **directly** — even though the same delta is already applied via the field block's owning composite's link recursion. The counter is double-counted, and the post-merge relay can compound it into a runaway.

A naive "skip if already merged" guard inside `ProcessBlock` does **not** work: `updateHeads` marks a composite's linked field CIDs as merged *before* the recursion applies them, so a field block already looks merged on its first (correct) application.

## Fix

Only **composite** blocks may enter the merge-processing list — they are the merge unit and carry their fields via links. A non-composite merge head becomes a no-op (the owning composite delivers the field). `loadComposites` only ever walks composite `Heads`, so normal merges are unaffected; this only filters a field-block *root* head.

## Tests

`TestMerge_CounterFieldBlockAsHead_IsNotReapplied` (added in the first commit, deterministic, single-node): merge a `+5` increment via its composite head (value → 5), then merge from the **field-block** head and assert the value is still 5. Fails on the unfixed processor (becomes 10); passes with the fix. Full `internal/db/...` suite passes.

The first commit adds the failing test; the second adds the fix.

## Context

Found via cross-implementation conformance testing (a Rust implementation interoperating over the same P2P protocol); see #4935 for the full investigation, instrumented traces, and the cross-impl repro. The interoperating implementation independently stopped emitting field-block heads from its backfill/retry path, which removed the runaway but **not** the residual — confirming the durable fix belongs here in the merge processor.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -247,7 +247,15 @@ func (mp *mergeProcessor) loadComposites(
 	// of the composite DAG. However, the new block and its children might have branched off from an older block.
 	// In this case, we also need to walk back the merge target's DAG until we reach a common block.
 	if block.Delta.GetPriority() >= mt.headHeight {
-		mp.composites.PushFront(block)
+		// Only composite blocks may enter the merge-processing list. A field
+		// (e.g. counter) block can reach here as the merge head when a peer pushes
+		// the DAG block-by-block (every block as its own pushLogRequest); processing
+		// it directly re-applies its delta — already applied via its owning
+		// composite's link recursion — double-counting counters (#4935). The
+		// composite is the merge unit and carries the field via its links.
+		if block.Delta.IsComposite() {
+			mp.composites.PushFront(block)
+		}
 		for _, head := range block.Heads {
 			err := mp.loadComposites(ctx, head.Cid, mt)
 			if err != nil {
--- a/internal/db/merge_counter_double_apply_test.go
+++ b/internal/db/merge_counter_double_apply_test.go
@@ -1,0 +1,127 @@
+// Copyright 2024 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package db
+
+import (
+	"context"
+	"testing"
+
+	"github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/corekv/blockstore"
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/event"
+	coreblock "github.com/sourcenetwork/defradb/internal/core/block"
+	"github.com/sourcenetwork/defradb/internal/core/crdt"
+	"github.com/sourcenetwork/defradb/internal/datastore"
+)
+
+const counterSchema = `
+type Tally {
+	hits: Int @crdt(type: pcounter)
+}
+`
+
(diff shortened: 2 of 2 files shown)


AndrewSisley:
Hey Jack, I'm going to close this for now, as the proposed fix is both small and incorrect, and this PR has sat idle for a couple of weeks.

The proposed fix will not just exclude field blocks, but also collection level blocks (both branchables, and col definitions), and anything else, preventing their sync.

If needed this can be re-opened once ready for review, or referenced from the 'closed' section of the PRs list.

PR #4937, opened 2026-06-16
chore: Standard use strings in CLI

### LINKED ISSUE #1651: Make the `Use` string standard across the CLI commands

ChrisBQu:
## Relevant issue(s)

Resolves #1651 

## Description

This PR seeks to standardize the Use strings for all commands across the CLI. The syntax rules are as follows:

- If the command name has no positional arguments, we just keep the command name. For example: `new`
- If there are required positional arguments, use angular brackets. For example: `get <docID>`
- If there are optional positional arguments, use square brackets. For example: `add [policy]`
- All flags are stripped from this string completely. Cobra will auto-generate those in the flag section.

This PR uses work done here as a base: https://github.com/sourcenetwork/defradb/pull/4478

## Tasks

## How has this been tested?

- WSL


fredcarle (approved):
LGTM. Code Rabbit flagged a valid point I think if you want to validate that before merge.

ChrisBQu:
> LGTM. Code Rabbit flagged a valid point I think if you want to validate that before merge.

I am going to make that change. The details are correct.

PR #4938, opened 2026-06-17
refactor: Default directive argument types

### LINKED ISSUE #4998: Clean up default directive verbose typing
From @ribiza 

> Right now, the @default directive requires the type to be specified:
active: Boolean @default(bool: true)
age: Int @default(int: 0)
status: String @default(string: "draft")
createdAt: DateTime @default(dateTime: UTC_NOW)
>
> Can we get rid of it, or is there good reasons to it? 
Reasons I see to get rid of it:
I don't think people are going to assume they can put a default true onto a string field (and if they do, the error will inform them they can't)
the difference in casing (Int vs int, Boolean vs bool, JSON vs json, etc) requires an extra element of care
it's (unnecessarily?) more verbose

jsimnz:
Resolves #4998
Refactors `@default` directive to use a specialize `Any` scalar as a catch all to resolve the awkward devex issue the current `@default` behavior of having typed arguments for each individual scalar type. While the old system provided stronger type gurantees, the new system will still enforce all types but at runtime and will be opaque to the introspection/intellisense/language servers.

Old:
```graphql
active: Boolean @default(bool: true)
age: Int @default(int: 0)
status: String @default(string: "draft")
createdAt: DateTime @default(dateTime: UTC_NOW)
```

New:
```graphql
active: Boolean @default(value: true)
age: Int @default(value: 0)
status: String @default(value: "draft")
createdAt: DateTime @default(value: UTC_NOW)
```

An obvious and easy additional change would be to shorted `value` literal for the argument name to `val`, open for debate if it reasonably adds anything to DX.

Staying in draft state till we decide if we want to A) move forward with this design B) merge it pre-v1 (nothing is left todo in this PR, only whatever PR comments arise)


nasdf (approved):
I'm happy with this change. Its a very big QoL improvement and will support any future types we add with minimal changes.

ribiza (approved):
Thanks for being so proactive on my feedback! 🤸

PR #4940, opened 2026-06-17
fix(i): P2P KMS async flakey behavior

### LINKED ISSUE #4939: P2P KMS timeout test flake
Currently only of the commonly failing tests is due to the P2P KMS causing a race condition in the tests.

Example run failure: https://github.com/sourcenetwork/defradb/actions/runs/27606507046/job/81619792939?pr=4800

jsimnz:
## Relevant issue(s)

Resolves #4939 

## Description
Fixes a flaky P2P KMS behavior  for encrypted document sync. Happens when multiple peers on the same topic are responding to KMS requests. There are circumstances where the intended node doesn't correctly recieve the KMS response for their request, which causes the tests to hang and eventually timeout.

The fix:
- KMS fetch requests now retry
- key requests now have unique requestIDs to disambiguate
- Including origin responding peer metadata within the ECIES AAD
- Updated response block/links for situations when responding peers dont have all the requested keys

## Tasks

## How has this been tested?

Expanded unit and integration tests. Re-ran integration tests 100 times to confirm flakeyness has been resolved (for this particular failure, there are obviously still other flakey tests/behavior out there)

- Linux NixOS

### DIFF
--- a/internal/kms/pubsub.go
+++ b/internal/kms/pubsub.go
@@ -258,7 +258,9 @@ func (s *pubSubService) handleFetchEncryptionKeyResponses(
 	privateKey *ecdh.PrivateKey,
 	result chan<- encryption.Result,
 ) {
-	// track latest `cancelPublish` func via the closure
+	// track latest `cancelPublish` func via the closure since the cancel func
+	// is mutated in this function and we need to cancel the latest value at defer
+	// time
 	defer func() {
 		cancelPublish()
 	}()
--- a/internal/kms/pubsub.go
+++ b/internal/kms/pubsub.go
@@ -14,6 +14,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/ecdh"
+	crand "crypto/rand"
 	"encoding/base64"
 	"time"
 
@@ -21,6 +22,7 @@ import (
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	grpcpeer "google.golang.org/grpc/peer"
 
+	"github.com/sourcenetwork/corelog"
 	"github.com/sourcenetwork/immutable"
 
 	"github.com/sourcenetwork/defradb/acp/dac"
@@ -40,6 +42,12 @@ const pubsubTopic = "encryption"
 // fetchEncryptionKeyResponseTimeout bounds the wait for a key-holding peer
 // when the request context has no deadline.
 const fetchEncryptionKeyResponseTimeout = 5 * time.Second
+
+// fetchEncryptionKeyRetryInterval bounds how long a single pubsub publish can
+// wait for a valid key response before the request is republished. Pubsub topic
+// membership can lag behind direct peer connections, especially in tests that
+// connect peers and immediately publish an encrypted block.
+const fetchEncryptionKeyRetryInterval = time.Second
 
 type PubSubServer interface {
 	AddPubSubTopic(
@@ -130,6 +138,7 @@ type fetchEncryptionKeyRequest struct {
 	Identity           []byte
 	Links              [][]byte
 	EphemeralPublicKey []byte
+	RequestID          []byte
 }
 
 // handleEncryptionMessage handles incoming FetchEncryptionKeyRequest messages from the pubsub network.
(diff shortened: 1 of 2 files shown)


internal/kms/pubsub.go
```diff
@@ -229,7 +258,10 @@ func (s *pubSubService) handleFetchEncryptionKeyResponses(
 	privateKey *ecdh.PrivateKey,
 	result chan<- encryption.Result,
 ) {
-	defer cancelPublish()
+	// track latest `cancelPublish` func via the closure
+	defer func() {
+		cancelPublish()
+	}()
```
fredcarle: question: Can you explain this? What do you mean by "track latest"? How is it really different from what it was before?

internal/kms/pubsub.go
```diff
@@ -239,12 +271,22 @@ func (s *pubSubService) handleFetchEncryptionKeyResponses(
...
+	retryTimer := time.NewTimer(fetchEncryptionKeyRetryInterval)
+	defer retryTimer.Stop()
+
 	for {
 		select {
 		case resp, ok := <-respChan:
 			if !ok {
-				result <- encryption.Result{Error: ErrNoPeerSuppliedEncryptionKey}
-				return
+				nextRespChan, nextCancel, err := s.republishFetchEncryptionKeyRequest(waitCtx, req, retryTimer)
```
fredcarle, line 283: suggestion: This approach opens the possibility of a high number of retries in a short amount of time. Instead, let the retryTimer handle republishing. This will give us a certain level of rate limiting.
```go
if !ok {
    respChan = nil
    continue
}

File: internal/kms/pubsub.go [hunk 3]
```diff
@@ -283,13 +372,27 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
 		// Peer didn't have the key; keep waiting for the one that does.
 		return nil, false
 	}
+	if len(keyResp.Links) != len(keyResp.Blocks) {
```
fredcarle, line 383: thought: Would be nice to have a test for when this is true.

fredcarle (approved):
LGTM. A few comments to resolve before merge.

internal/kms/pubsub.go
jsimnz: the `cancelPublish` is now mutated within this function on the retry branch. Before `defer cancelPublish` locked in the function at the time the `defer` line is executed, so updating `cancelPublish` later in the function had no effect. Now that its in a closure, it correctly tracks whatever the latest `cancelPublish` is, and correctly cleans/cancels the outstanding context.

A handy little trick but it looks odd without explanation, especially when compared to the old version, it kinda looks like theres no difference, but the details are hidden in the closure evaluation semantics.

I can expand the basic comment I added to give more context if you think its necessary

internal/kms/pubsub.go
```diff
@@ -194,29 +203,49 @@ func (s *pubSubService) requestEncryptionKeyFromPeers(
...
+	data []byte,
+) (<-chan client.PubsubResponse, context.CancelFunc, error) {
 	// Cancel signals the go-p2p relay goroutine to stop; without it, it would
 	// block forever on its unbuffered send once we return after the first valid reply.
 	pubCtx, cancel := context.WithCancel(ctx)
 	respChan, err := s.pubsub.PublishToTopic(pubCtx, pubsubTopic, data, true)
 	if err != nil {
 		cancel()
-		return errors.Wrap("failed publishing to encryption thread", err)
+		return nil, nil, err
```
islamaliev, line 239: suggestion: why do we ignore this error? Would be nice to have a comment if there is a good reason. Also maybe it makes sense at least to log it

internal/kms/pubsub.go
fredcarle: Maybe a `notice the mutation bellow` type comment would be nice to add :)

internal/kms/pubsub.go
jsimnz: the error isnt ignored, it still returned in this line, what changed was the additional error wrapping

PR #4942, opened 2026-06-17
feat: Support secondary index backfill

### LINKED ISSUE #296: Sec. Indexes: Allow backfill

### LINKED ISSUE #4907: Cannot add an index larger than ~11MB

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4907
Resolves https://github.com/sourcenetwork/defradb/issues/296

## Description

Secondary indexes can now be backfilled: creating an index over a collection that already
holds documents indexes all of them, in batches, so an index is usable immediately after
creation no matter how much data the collection holds. Deletion works the same way in
reverse, removing entries in batches. This also lifts the previous limit where a create or
delete that didn't fit in a single store transaction would fail.

An index's lifecycle (building, ready, failed, dropping) is tracked through the action-status
system introduced in #4922 — the same mechanism truncate and datastore refresh use. If a build
fails partway through, the database keeps working: the index is marked failed with a reason, and
it can be dropped and recreated. Index listings report this status, and in-flight index builds
appear in `ListActions` alongside other long-running operations. Writers keep maintaining an
index while it builds, the query planner only uses indexes that are fully ready and otherwise
falls back to a scan, and if the node restarts partway through, an interrupted build continues
from where it left off and an interrupted deletion finishes cleaning up. Indexes created before
this change have no record and are treated as ready, so existing databases are unaffected.

### Extensions to the action-status system from #4922 (for @AndrewSisley review)

This builds index lifecycle on the merged action-status system, which required extending it. The
extensions are backward compatible — truncate and datastore refresh keep their existing on-disk
keys and values:

- **Per-subject key segment.** The action key gains an optional trailing segment
  (`[collectionID]/[action]/[subject]`), so two index builds on one collection don't collide
  (subject = index ID). Collection-wide actions leave it empty and keep their two-segment key.
  `client.ActionExecution`/`event.ActionExecution` gain a `Subject` field.
- **Envelope value.** The bare-uvarint status value becomes an `{Status, Reason, Payload}`
  envelope so a failed index can carry its reason and a building index its watermark. Decoding
  falls back to the legacy bare-uvarint form, so records written before this change still read.
- **Transactional write path.** A `SetTxn`/`CompleteTxn` pair writes the action record through
  the caller's transaction (vs. the existing non-transactional `context.TODO()` path), so a
  building index's watermark commits atomically with the index entries of the same batch — which
  is what makes crash-resume correct. Per-batch watermark advances suppress the change event to
  avoid flooding the (blocking) event bus; only genuine status transitions publish.
- **New actions.** `BackfillIndexAction` (build, watermark in payload) and `DropIndexAction`
  (drop/GC, no watermark) so the two are distinguishable in `ListActions`.

## Tasks

- [x] Code commented where non-obvious.
- [x] Repository documentation updated (ai/docs/secondary-index.md lifecycle section).
- [x] PR title follows conventional commit style.
- [x] Limitations discussed (failed-index requires manual drop; async backfill deferred).

## How has this been tested?

Unit tests for the batched backfill and GC, the transaction-conflict retry path, the
unique-index build-time tolerance, and startup recovery (resume from watermark, interrupted
deletion, orphaned and failed records). Integration tests covering large-collection create
and delete, failed-index lifecycle (fail, persist, drop, recreate), explicit-transaction
create and delete, and status surfacing through listings. Run against the in-memory and file
backends, the secondary-index multiplier, and the http and cli client types.

Also validated end to end against a running node over a multi-MB collection: a point-lookup query dropped from a full scan at roughly 80ms to about
0.4ms once indexed, and a hard crash (kill -9) part way through both a build and a delete
recovered cleanly on restart, with the build finishing every document and the delete leaving
no leftover index entries.


client/index.go
```diff
@@ -62,6 +62,29 @@ type CollectionIndex interface {
 	Description() IndexDescription
 }
 
+// IndexStatus describes the lifecycle state of an index.
+type IndexStatus string
+
+const (
+	// IndexStatusBuilding indicates the index entries are being backfilled from existing documents.
+	IndexStatusBuilding IndexStatus = "building"
```
AndrewSisley: question: Why have you added a new set of statuses, instead of using the action statuses?

client/index.go
```diff
@@ -62,6 +62,29 @@ type CollectionIndex interface {
...
+	IndexStatusReady IndexStatus = "ready"
+	// IndexStatusFailed indicates the index backfill failed; the Reason field provides details.
+	IndexStatusFailed IndexStatus = "failed"
+	// IndexStatusDropping indicates the index entries are being garbage-collected.
+	IndexStatusDropping IndexStatus = "dropping"
+)
+
+// IndexDescriptionStatus combines the static index description with its mutable runtime state.
+type IndexDescriptionStatus struct {
+	IndexDescription
```
AndrewSisley: question: Why did you chose to embed this, instead of using a flatter object?

client/index.go
```diff
@@ -62,6 +62,29 @@ type CollectionIndex interface {
...
+	// Status is the current lifecycle state of the index.
+	Status IndexStatus
+	// Reason is set when Status is IndexStatusFailed, providing a description of the failure.
+	Reason string
```
AndrewSisley: question: Why did you chose to make this index specific, instead of adding this feature to the action system?

internal/db/db.go
```diff
@@ -219,6 +224,12 @@ func newDB(
 		return nil, err
 	}
 
+	db.recoveryWG.Go(func() {
+		if err := db.recoverIndexStates(db.ctx); err != nil {
+			log.ErrorE("index state recovery failed", err)
+		}
+	})
```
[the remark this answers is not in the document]
islamaliev: @coderabbitai it was added in 1.25 version https://pkg.go.dev/sync#WaitGroup.Go
Add to learnings never ever to flag compiler issues. We have CI for this

internal/db/action/action.go
```diff
@@ -11,17 +11,43 @@ package action
...
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/event"
 	"github.com/sourcenetwork/defradb/internal/datastore"
 	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
-// Register a new action for execution
+// Envelope is the stored value of an action record. Reason holds an errored action's cause
+// and Payload holds action-specific state such as an index build watermark.
+type Envelope struct {
```
AndrewSisley: question: Why did you chose to store everything against the same key as a json object instead of storing the new stuff against new keys?

client/index.go
islamaliev: When I reviewed your action PR we touched on this and my read was that the index-specific needs should be covered by index itself. Now that you mentioned it I need to think...

client/index.go
islamaliev: what do you mean flatter? To extend IndexDescription with a new field? IndexDescription is more or less static and the other one is dynamic so I didn't want to mix them. At the same time I wanter the index state to be visible to users that call `ListIndexes`

client/index.go
AndrewSisley: > When I reviewed your action PR we touched on this and my read was that the index-specific needs should be covered by index itself.

Yes, that was for index specific stuff, this looks like an error-string field, which should really have reason to exist for any action.

client/index.go
AndrewSisley: > what do you mean flatter? 
```
type IndexDescriptionStatus struct {
    Name string
	ID uint32
	Unique bool
    Status IndexStatus
    Reason string
}
```
It should be nicer to use, especially for json clients like http and cli, and it avoids unnesecary coupling.

Including the entire `IndexDescription` when users want a status feels like it is blending two different things together, whether embedded or flat.  

I think I see why you want a mix being returned from `ListIndexes`, maybe embedding is the wrong tool to do this? (Is this a personal preference of mine?)

suggestion:  For `ListIndexes`, would something similar to the following be more appropriate, allowing both desccription and status to be returned, but without blending them together in the same way embedding does?

```
 // I'm being very lazy with my naming suggestion here, no attachment to it at all
type ListIndexesResult struct {
  Description IndexDescription
  Status { // type/name tbd :) - looks like it could be the existing `ActionExecution` type, if `Reason` is added to it
    Status IndexStatus // or ActionStatus
    Reason string 
  }
}
```

internal/db/collection_index.go
```diff
@@ -230,37 +273,67 @@ func processNewIndexRequest(
...
 		c.def.Indexes = c.def.Indexes[:len(c.def.Indexes)-1]
-		return nil, err
+		return client.IndexDescription{}, nil, err
 	}
 
-	index, err := c.appendNewIndexAndIndexExistingDocs(ctx, desc)
+	// This writes the building state without an "already in progress" guard, which is safe
+	// because the index ID keying the record is sequence-allocated and therefore unique per
+	// build; a caller that reused an index ID would need to add one.
+	err = c.db.setIndexState(ctx, c.def.CollectionID, desc.ID, indexState{Status: client.IndexStatusBuilding})
```
AndrewSisley: If multiple transactions are open at the same time the store leveldb store will deadlock.  This looks like it will spawn multiple new transactions whilst the transaction used to set the descriptions etc in this parent function is still open. 

todo: Please make sure it is working with leveldb, the job is not required and is easy to accidentally miss.

internal/db/db.go
```diff
@@ -219,6 +224,12 @@ func newDB(
 		return nil, err
 	}
 
+	db.recoveryWG.Go(func() {
+		if err := db.recoverIndexStates(db.ctx); err != nil {
```
AndrewSisley, line 228: todo: This looks like it will break leveldb, multiple transactions cannot be opened at the same time otherwise it will deadlock.  Please make sure this is working.

client/index.go
islamaliev: I think I forgot to clean this up after your PR. Will take care of it

internal/db/db.go
```diff
@@ -219,6 +224,12 @@ func newDB(
 		return nil, err
 	}
 
+	db.recoveryWG.Go(func() {
```
AndrewSisley, line 227: thought: This is an index specific implementation of https://github.com/sourcenetwork/defradb/issues/4874 - I think we might want to rework this in the near future so that they all operate via the same system.

I'm 50-50 about whether I think it is fine to add this for secondary indexes right now, as it introduces an odd discrepancy, where indexes are resumed on restart, but the other actions are not.

internal/db/db.go
AndrewSisley: I spoke with Keenan - I misunderstood the leveldb limitation and introduced my own deadlock with the action system. We can avoid support for truncate/RefreshView/indexes with leveldb for v1 and fix it later.

Please make sure that this does not forever block subsequent database actions for leveldb post-restart though, and preferably make sure the routine isn't leaking (by deadlocking it).

internal/db/collection_index.go
AndrewSisley: I spoke with Keenan - I misunderstood the leveldb limitation and introduced my own deadlock with the action system. We can avoid support for truncate/RefreshView/indexes with leveldb for v1 and fix it later.

todo: Please make sure relevant secondary index tests are skipped when testing with leveldb - I will add this to the view/truncate tests.

AndrewSisley: I've created https://github.com/sourcenetwork/defradb/issues/4959 - please add a link to it to any tests that need to be skipped.

client/index.go
islamaliev: refactored

internal/db/action/action.go
islamaliev: after moving some stuff around decided to store 3 keys: status, reason, payload

internal/db/db.go
islamaliev: I added the link in a comment.

internal/db/collection_index.go
islamaliev: Checked the index suite on leveldb (DEFRA_LEVEL=true), it passes. The implicit create/delete paths commit the definition txn before backfilling, so they never hold two txns open at once. The only index tests that do are the explicit-txn ones in backfill_txn_test.go, and those already skip leveldb, I've added the #4959 link to that skip comment. So nothing else in the index suite needs skipping.

internal/db/db.go
islamaliev: Tested this on a real leveldb node: created an index over a large collection, killed process mid-build, restarted. The build resumes from its watermark, the node boots normally, post-restart queries/writes all return right away, and clean shutdown exits without hanging. No deadlock or leak. It's safe because recovery never holds two txns at once.

internal/db/action/action.go
```diff
@@ -40,7 +57,10 @@ func Register(
 	return Set(ctx, multistore, events, collectionID, action, client.InProgressActionStatus)
 }
 
-// Set the status for an existing action
+// Set the status for an existing action. Non-transactional, collection-wide.
+//
+// It passes context.TODO() to force a transaction-free write: corekv otherwise binds the
+// transaction on the context to the write (https://github.com/sourcenetwork/corekv/issues/107).
```
AndrewSisley, line 63: praise: This is a better place for this, thanks for moving it

internal/db/action/action.go
```diff
@@ -79,20 +153,87 @@ func Complete(
...
+	collectionID string,
+	action client.Action,
+	subject string,
+) error {
+	txn := datastore.CtxMustGetTxn(ctx)
+
+	for _, key := range [][]byte{
+		keys.NewActionStatusSubjectKey(collectionID, action, subject).Bytes(),
+		keys.NewActionReasonKey(collectionID, action, subject).Bytes(),
+		keys.NewActionPayloadKey(collectionID, action, subject).Bytes(),
```
AndrewSisley, line 182: praise: This is a nice way of handling this, thanks for showing me :)

tests/integration/index/backfill_large_test.go
```diff
@@ -0,0 +1,120 @@
...
+					}
+				`,
+			},
+			// Each doc's name is a unique ~50 KB value and its age equals the doc index,
+			// both derived from i so the query target is deterministic.
+			testUtils.GenerateDocs{
+				Options: []gen.Option{
+					gen.WithTypeDemand("User", largeDocCount),
+					gen.WithFieldGenerator("User", "name", func(i int, _ func() any) any {
+						return largeNameValue(i)
```
AndrewSisley, line 63: question: I avoided doing this for truncate/RefreshViews as I thought it would be too slow for me to want it in the main test suite - how long does this take in the CI/your-machine?  Are we sure we want this here?  I tend to run the tests frequently whilst deving, and the impact of some of our existing larger tests is notable.

AndrewSisley (approved):
LGTM, but I am a bit concerned about the large doc test's cost.

tests/integration/index/backfill_large_test.go
islamaliev: I measured it. On my machine the test itself takes about 0.25s normally, and around 2s under -race. Because the tests in the package run in parallel, the actual time it adds to a local make test run is small (well under half a second). Under race in CI it adds roughly 2s.

For context, it's the 3rd slowest test in the index package, but two of the existing query-performance tests are already in the same range, so it's not an outlier.

So locally the impact is minor. If the ~2s under race still feels like too much for the hot path, I'm happy to put it behind testing.Short() (or a build tag) so it runs in CI but gets skipped during normal local dev, or nightly or whatever else. I'd merge it like this for now and we see if it causes any trouble down the road.

AndrewSisley: Nice, thanks Islam!  2s is reasonable in the short term, we can pull it out when the nightly arrives :)

PR #4945, opened 2026-06-17
feat: Render field Kind and Typ as strings

### LINKED ISSUE #4816: Display string representation of field kind on collection describe

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4816

## Description

Describing a collection (via the CLI or the HTTP collections endpoint) printed each field's kind and CRDT type as bare numbers, which were only meaningful to someone reading the source. They now render as their human-readable strings, so a string field shows as `String` rather than `11` and a last-writer-wins field shows as `lww` rather than `1`.

This is a display-only change. The on-disk format and the collection patch operations still read and write the numeric form unchanged, so there is no data format change. The rendered output is reversible, so the HTTP client reconstructs the exact same collection from it, including relation fields.

## Tasks

- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, etc.

## How has this been tested?

Unit tests for the new string-to-CRDT-type parsing and the rendered output round-trip, plus an HTTP handler test asserting the response contains the string forms and not the numbers. The existing collection-describe integration suite runs across the go, http, and cli clients and is the round-trip guard; new cases cover a scalar field, a scalar-array field, and both self and cross-collection relation kinds.


client/ctype.go
```diff
@@ -68,3 +73,55 @@ func (t CType) String() string {
...
+// must keep emitting the numeric form. Only the describe display path emits the string.
+func (t *CType) UnmarshalJSON(b []byte) error {
+	if len(b) > 0 && b[0] == '"' {
+		var s string
+		if err := json.Unmarshal(b, &s); err != nil {
+			return err
+		}
+		ctype, ok := CTypeFromString(s)
+		if !ok {
+			return fmt.Errorf("unknown CRDT type: %q", s)
```
fredcarle: todo: Please make this into a proper error and make sure the parameter is passed via `errors.KV`

tests/integration/collection_version/describe_string_kind_test.go
fredcarle, line 1: thought: The 3 tests in here seem to test exactly the same thing. Maybe we can keep just one and improve the name of the test and the documentation that goes with it.

client/ctype.go
```diff
@@ -68,3 +73,55 @@ func (t CType) String() string {
 		return "unknown"
 	}
 }
+
+// CTypeFromString returns the [CType] matching the given string representation,
+// the inverse of [CType.String].
+//
+// The second return value is false if the string does not match a known CRDT type.
+func CTypeFromString(s string) (CType, bool) {
```
fredcarle, line 90: thought: I assume that this enables using the string representation of the type in patch operations? If so, it would me great to add integration tests for it.

client/ctype.go
islamaliev: done

tests/integration/collection_version/describe_string_kind_test.go
islamaliev: merged tests

client/ctype.go
islamaliev: done

internal/db/description/collection_render.go
```diff
@@ -0,0 +1,84 @@
...
+// "lww") rather than their numeric IDs. The rendered form round-trips back into a
+// [client.CollectionVersion].
+//
+// This is deliberately a separate type rather than a MarshalJSON on [client.CollectionVersion]:
+// that struct is marshalled as-is when persisted and when evaluating JSON-patch ops, both of
+// which must keep emitting numbers.
+//
+// [client.CollectionVersion] is embedded (so new fields carry through automatically) with only
+// Fields shadowed, encoding/json picks the shallower Fields and suppresses the embedded one.
+type RenderedCollectionVersion struct {
```
AndrewSisley: thought: This feels like an odd way of doing this, in an odd location.

todo: I don't think we should be coupling external layer clients to internal packages like this - if the http and higher level clients cannot be built using the embedded Go client, then the embedded Go client is not fit for purpose and needs to be fixed - if we cannot successfully consume it, what hope do our embedded Go users have?

suggestion: Please consider implementing a `CollectionVersion.Display()` or `CollectionVersion.PrettyPrint()` function on the CollectionVersion object in the `client` package so that users can call this when they like?

AndrewSisley (changes requested):
Please avoid making the high level clients dependent on internal code.

fredcarle (approved):
LGTM. Just make sure Andy's comments are resolve before merge.

internal/db/description/collection_render.go
islamaliev: I wasn't sure sure where to put the wrapper struct. Initially thought about adding it to `client`, but didn't want to extend API. Also didn't want to duplicate the struct in each client. So decided to go with this one. 
But adding a method that returns already ready result is a good option

islamaliev: done

AndrewSisley (approved):
LGTM, thank you for moving it!

PR #4961, opened 2026-06-18
fix: P2P KMS response key CID verification

### LINKED ISSUE #4948: P2P KMS response CID verification
Need additional verification of the responded blocks relative to the requested links

jsimnz:
## Relevant issue(s)

Resolves #4948 

## Description

Adds support to explicitly verify incoming encryption key response CIDs against the requested key IDs.

## Tasks

## How has this been tested?

Explicit unit tests

- Linux NixOS

### DIFF
--- a/internal/kms/enc_store.go
+++ b/internal/kms/enc_store.go
@@ -74,3 +74,45 @@ func (s *ipldEncStorage) put(ctx context.Context, blockBytes []byte) ([]byte, er
 
 	return []byte(link.String()), nil
 }
+
+func (s *ipldEncStorage) putBlock(ctx context.Context, block coreblock.Encryption) ([]byte, error) {
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.SetWriteStorage(blockstore.NewIPLDStore(s.encstore))
+
+	link, err := lsys.Store(linking.LinkContext{Ctx: ctx}, coreblock.GetLinkPrototype(), block.GenerateNode())
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(link.String()), nil
+}
+
+func (s *ipldEncStorage) computeLink(ctx context.Context, blockBytes []byte) ([]byte, error) {
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.SetWriteStorage(blockstore.NewIPLDStore(s.encstore))
+
+	var encBlock coreblock.Encryption
+	err := encBlock.Unmarshal(blockBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	link, err := lsys.ComputeLink(coreblock.GetLinkPrototype(), encBlock.GenerateNode())
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(link.String()), nil
+}
+
+func (s *ipldEncStorage) computeBlockLink(ctx context.Context, block coreblock.Encryption) ([]byte, error) {
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.SetWriteStorage(blockstore.NewIPLDStore(s.encstore))
+
+	link, err := lsys.ComputeLink(coreblock.GetLinkPrototype(), block.GenerateNode())
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(link.String()), nil
+}
--- a/internal/kms/enc_store.go
+++ b/internal/kms/enc_store.go
@@ -114,5 +114,5 @@ func (s *ipldEncStorage) computeBlockLink(ctx context.Context, block coreblock.E
 		return nil, err
 	}
 
-	return []byte(link.String()), nil
+	return []byte(link.Binary()), nil
 }
--- a/internal/kms/enc_store.go
+++ b/internal/kms/enc_store.go
@@ -57,47 +57,11 @@ func (s *ipldEncStorage) get(ctx context.Context, cidBytes []byte) (*coreblock.E
 	return coreblock.GetEncryptionBlockFromNode(nd)
 }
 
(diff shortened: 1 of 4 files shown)


internal/kms/pubsub.go
```diff
@@ -405,13 +401,29 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
...
+		var encBlock coreblock.Encryption
+		err = encBlock.Unmarshal(decryptedData)
+		if err != nil {
+			return nil, false, errors.Join(ErrDecodingEncryptionKey, err)
+		}
+
+		if !skipVerify {
+			link, err := s.encStore.computeBlockLink(s.ctx, encBlock)
+			if err != nil {
+				return nil, false, errors.New("key cid")
```
fredcarle: todo: Please return or wrap the `err` from `s.encStore.computeBlockLink`

internal/kms/pubsub.go
```diff
@@ -405,13 +401,29 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
...
 		}
 
-		if _, err := s.encStore.put(context.Background(), decryptedData); err != nil {
-			log.ErrorContextE(s.ctx, "Failed to store encryption key", err)
-			return nil, false
+		if _, err := s.encStore.putBlock(context.Background(), encBlock); err != nil {
+			return nil, false, errors.Join(ErrEncryptionKeyStore, err)
 		}
 
 		// todo: verify incoming response order block/CIDs match the request
```
fredcarle: todo: Please remove todo comment.

internal/kms/pubsub.go
```diff
@@ -405,13 +401,29 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
...
+		}
+
+		if !skipVerify {
+			link, err := s.encStore.computeBlockLink(s.ctx, encBlock)
+			if err != nil {
+				return nil, false, errors.New("key cid")
+			}
+
+			if !bytes.Equal(keyResp.Links[i], link) &&
+				!bytes.Equal(req.Links[i], link) {
```
fredcarle: thought: Shouldn't this be 
```go
if !(bytes.Equal(keyResp.Links[i], link) && bytes.Equal(req.Links[i], link))
```
We should return an error if one of them is false rather than if both are false.

internal/kms/pubsub.go
```diff
@@ -405,13 +401,29 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
...
+				return nil, false, errors.Join(ErrEncryptionKeyCIDMismatch, err)
```
fredcarle: todo: `err` is always `nil` here so either return `ErrEncryptionKeyCIDMismatch` or `errors.WithStack(ErrEncryptionKeyCIDMismatch)` if you want the added stack trace.

internal/kms/enc_store.go
```diff
@@ -57,20 +57,26 @@ func (s *ipldEncStorage) get(ctx context.Context, cidBytes []byte) (*coreblock.E
...
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.SetWriteStorage(blockstore.NewIPLDStore(s.encstore))
+
+	link, err := lsys.ComputeLink(coreblock.GetLinkPrototype(), block.GenerateNode())
 	if err != nil {
 		return nil, err
 	}
 
-	return []byte(link.String()), nil
+	return []byte(link.Binary()), nil
```
fredcarle, line 81: thought: It's a little odd that the output of `putBlock` and `computeBlockLink` return a different representation of the CID. It's fine at the moment because the output of `putBlock` is simply discarded but there is a chance for this to be a source of bug in the future. Maybe it's worth changing the output of `putBlock` to be the same as `computeBlockLink`

fredcarle (approved):
Approving with a few comments to resolve before merge.

internal/kms/pubsub.go
jsimnz: changed it to an `||` instead, so now either case failing evals to true

internal/kms/enc_store.go
jsimnz: I thought it was a little odd too, as you point out `putBlock` returned link isnt used anywhere so no call site ran into type mismatching. Originally I had it as string, but to do a quick easy compare with the `fetchEncryptionKeyRequest` and `fetchEncryptionKeyReply` which use binary version of the CIDs, so I opted to match that.

I'll leave the `putBlock` for now tho

internal/kms/pubsub.go
```diff
@@ -405,25 +401,38 @@ func (s *pubSubService) tryHandleFetchEncryptionKeyResponse(
...
+			return nil, false, errors.Join(ErrDecodingEncryptionKey, err)
+		}
+
+		if !skipVerify {
+			link, err := s.encStore.computeBlockLink(s.ctx, encBlock)
+			if err != nil {
+				return nil, false, errors.Join(ErrKeyCIDGeneration, err)
+			}
+
+			if !bytes.Equal(keyResp.Links[i], link) ||
```
islamaliev: todo: a peer can return only subset of requested blocks so we can't expect any order there and therefore we shouldn't use index i.

```go
// before the loop
reqSet := make(map[string]struct{}, len(req.Links))
for _, l := range req.Links {
    reqSet[string(l)] = struct{}{}
}
// in the loop
if !bytes.Equal(keyResp.Links[i], link) {
    return nil, false, ErrEncryptionKeyCIDMismatch
}
if _, ok := reqSet[string(link)]; !ok {
    return nil, false, ErrEncryptionKeyCIDMismatch
}
```

islamaliev (dismissed):
Overall looks good. Just one simple request

PR #4962, opened 2026-06-19
test(i): Skip Truncate and ViewRefresh leveldb tests

### LINKED ISSUE #4960: Skip Truncate and RefreshView tests with LevelDB
They deadlock.  The production behaviour is tolerated in the short term for v1, but we need to keep the CI happy in the meantime.

AndrewSisley:
## Relevant issue(s)

Resolves #4960

## Description

Skips Truncate and ViewRefresh leveldb tests.

Short term change only to fix the test run so that it doesn't hide other, newer errors until we fix this properly post v1.

Discussed with Keenan whether it was fine to leave leveldb not supporting Truncate and ViewRefresh until post v1 and he seemed okay with it.  Do push-back/discuss if uncomfortable about this.


nasdf (approved):
LGTM. Thanks for opening the tracking issue as well.

fredcarle (approved):
LGTM

Issue #4959, opened 2026-06-18
Truncate and RefreshView deadlocks with leveldb


AndrewSisley:
Exact reason to be discovered, I might have misunderstood its limitations - it is deadlocking when writing without a txn whilst a txn is open.

Some tests are skipped at the moment because of this, please see the comment linked in the codebase to this ticket.

This ticket is required for https://github.com/sourcenetwork/defradb/issues/4873 and https://github.com/sourcenetwork/defradb/issues/4874 - otherwise it would be possible to permanently lock a leveldb Defra instance (especially if/when lock resumes on database restart).

## Possible solution

- For implicit txns, scope the lock to a ghost-txn (not an actual store txn) - this can be done on txn create (instead of creating the current implicit txn), no need to impact the main body of code.
- RefreshViews queries Defra, the queries need to be executed within a txn, but that txn only needs to be read-only, and only needs to live for the duration of the query execution.
- AddView will to lock the collection (potentially to a ghost-txn if implicit) before the parent (creating) txn is committed.  The implicit creating txn will need too be committed before the refresh, but after the lock is acquired - the lock will ensure that any queries against the new view that are opened before the refresh has started do not execute until after the refresh has completed.
- The only way I can spot so far where leveldb works with explicit txns is for leveldb *only* to act upon the explicit txn instead of rootstore.  It might be possible to solve this at the corekv wrapper level - potentially via a new wrapper not coupled to leveldb (handy in case we ever need this for other store or use cases), this will remove this distraction from the DefraDB business logic and keep Truncate/refresh/etc focused on the action it is solving.  This element should probably be broken out to a new ticket.

edjroz:
This deadlock actually has 2 manifestations:

 **1. Self-locking (one goroutine deadlocking itself).**
When a truncate/refresh runs, it opens a transaction `T1`, and on leveldb that grabs the store's single write lock. Later in the same flow, `action.Register` does a **txn-free** write (it deliberately writes *outside* the transaction). But leveldb won't let a non-transactional write through while a transaction is open so that `DB.Put` waits for `T1` to close… but `T1` only closes once the flow finishes, which it can't, because it's stuck on the `DB.Put`. A cycle where `DB.Put` waits for `T1` <-> `T1` waits for `DB.Put`:

  ```mermaid
  sequenceDiagram
      participant F as Truncate / RefreshView
      participant DB as goleveldb (single write lock)
      F->>DB: open T1 (OpenTransaction) — takes the write lock
[...]

PR #4968, opened 2026-06-21
fix: Leveldb truncate refresh self deadlock

edjroz:
## Relevant issue(s)

Addresses #4959 First manifestation, the Truncate / RefreshView / AddView self-deadlock.

## Description

On the leveldb store, `Truncate`, `RefreshView`, and `AddView` would deadlock.

**Root cause.** These flows open a store transaction `T1` (which, on leveldb, takes the store's single write lock) and then issue **txn-free** writes within the same flow. The action execution markers, written via `context.TODO()` so they survive a rollback. leveldb blocks any  non-transactional write while a transaction is open, so the txn-free `db.Put` waits for `T1` to close, but `T1` only closes once the flow finishes — which it can't, because it's blocked on that write. A single goroutine deadlocks with itself. Badger and the in-memory store are MVCC, so they permit those concurrent writes and never hit this.

  **Fix.** Stop holding a store transaction open at the same time as a txn-free write:

  - A rootstore-backed **lock token** (`internal/db/lock_token.go`) holds the collection write lock —
  - `Truncate` (implicit) now runs fully txn-free under the token (this also removes a latent
    transaction-size blow-up: the bulk deletes were previously routed through the transaction contrary
    to the documented intent).
  - `RefreshView` writes its action markers with no transaction open, and runs the truncate-and-rebuild
    inside its own per-view transaction (the rebuild planner re-enters public APIs that need a real
    `*db.Txn`, and performs no txn-free writes).
  - `AddView` commits the collection metadata before running the txn-free refresh.
  - The leveldb test-skip helper is narrowed to only the still-unsupported cases (explicit-transaction
    and concurrent/parallel variants — those are the second manifestation).
    
### Before: the self-deadlock

  ```mermaid
  sequenceDiagram
      autonumber
      participant G as Truncate goroutine
      participant L as goleveldb DB<br/>(one exclusive write lock)
      participant S as systemstore<br/>(same DB, diff. prefix)

      G->>L: NewTxn() → OpenTransaction()  (collection_truncate.go:51)
      L-->>G: T1 — exclusive write lock acquired
      Note over L: write lock now held by T1

      G->>S: action.Register → Systemstore().Set(InProgress)<br/>txn-free, via context.TODO() (action.go:55)
      S->>L: db.Put(marker) — needs the write lock
      Note over L,S: ⛔ Put blocks: lock held by T1

      rect rgba(255,0,0,0.08)
      Note over G: would reach txn.Commit() (collection_truncate.go:94)<br/>to release T1 … but is blocked on the Put above
      end
      Note over G,L: cycle: Put waits for T1 to close ⇄ T1 closes only after the Put returns
  ```

  ### After: txn-free under a lock token

  ```mermaid
  sequenceDiagram
      autonumber
      participant G as Truncate / RefreshView goroutine
      participant K as LockSet<br/>(collection write lock)
      participant L as goleveldb DB<br/>(one exclusive write lock)

      G->>K: newLockToken() + CollectionLock(token)
      Note over K: write lock held by the token<br/>(this is what isolates readers)
      Note over L: NO OpenTransaction ⇒ goleveldb<br/>write lock stays free

      G->>L: action.Register → Systemstore().Set (txn-free)
      L-->>G: ok (no open txn to block on)
      G->>L: bulk deletes (txn-free, direct to rootstore)
      L-->>G: ok
      G->>L: action.Complete (txn-free)
      L-->>G: ok

      G->>K: token.release()
      Note over K: write lock freed
      rect rgba(0,150,0,0.08)
      Note over G,L: a transaction is never open while a txn-free write runs ⇒ no self-deadlock
      end
  ```
  

### Limitations
- **Explicit-transaction variants remain unsupported on leveldb.** Running these ops inside a user-supplied transaction still issues a txn-free write while that transaction is open, so it would still deadlock; those cases remain skipped on leveldb (second manifestation).
- **Reduced atomicity of the implicit path.** Truncate and the view-cache rebuild's clear step are now non-transactional, so a crash mid-operation can leave partial state rather than rolling back. Recovery relies on the action markers + the resume-safe delete ordering, not a single atomic commit.
  - **The lock token is a `datastore.Txn`, not a `*db.Txn`.** Any future read/write path that type-asserts the context transaction to the concrete `*db.Txn` will panic under the token; only the verified truncate paths run under it (refresh deliberately uses a real transaction for this reason).
  - **AddView is no longer atomic across create + refresh**, and `RefreshViews` rebuilds each view in its own transaction (a multi-view refresh is no longer atomic as a group).

## Tasks

## How has this been tested?

- Unit: `go test ./internal/db/ -run TestLockToken` and `go test ./node/ -run TestStoreLevel`
    (new characterization tests pinning leveldb's transaction model).
- leveldb integration (previously skipped, now passing):
    - `DEFRA_LEVEL=true go test -race ./tests/integration/collection/truncate/...`
    - `DEFRA_LEVEL=true DEFRA_VIEW_TYPE=materialized go test ./tests/integration/collection/refresh/... ./tests/integration/view/simple/...`
- Regression on the default (badger) backend across the changed truncate/view/refresh suites, plus `go test ./internal/... ./node/`.
- Note: a few view tests that require compiled Rust/wasm lenses fail locally for lack of those build artifacts — this is environmental and reproduces identically on badger (not caused by this change).

- MacOS

### DIFF
--- a/internal/db/collection_truncate.go
+++ b/internal/db/collection_truncate.go
@@ -37,7 +37,7 @@ const hardDeleteChunkSize int = 10000
 func (c *collection) Truncate(
 	ctx context.Context, opts ...options.Enumerable[options.TruncateCollectionOptions],
 ) error {
-	ctx, _, _ = getTxnAndSetCtxForCollection(ctx, c)
+	ctx, _, hadTxn := getTxnAndSetCtxForCollection(ctx, c)
 
 	ctx, span := tracer.Start(ctx)
 	defer span.End()
@@ -48,18 +48,58 @@ func (c *collection) Truncate(
 		return err
 	}
 
+	// Truncate performs txn-free writes: it deletes more key-values than a single store transaction
+	// may safely hold, and it records its action-execution markers outside any transaction (so they
+	// survive a rollback). leveldb blocks non-transactional writes while a transaction is open, so
+	// when no explicit transaction is in play we hold the collection write lock with a lock token
+	// rather than an open store transaction, and write everything directly to the rootstore.
+	//
+	// When the caller supplied an explicit transaction we must honour it. That path is unsupported on
+	// leveldb (a txn-free write would block on the open transaction - see issue #4959).
+	if hadTxn {
+		return c.truncateInTxn(ctx)
+	}
+	return c.truncateTxnFree(ctx)
+}
+
+// truncateTxnFree truncates the collection without opening a store transaction. The collection write
+// lock - which is what provides reader isolation - is held by a lock token for the duration, and
+// every read and write goes directly to the rootstore. This is the path that supports leveldb.
+func (c *collection) truncateTxnFree(ctx context.Context) error {
+	token := c.db.newLockToken()
+	defer token.release()
+
+	// The token serves reads (it satisfies datastore.Txn), and clearing the corekv transaction
+	// ensures every write goes straight to the rootstore with no transaction open. The short-id
+	// caches are normally initialised by InitContext; we initialise them here as we bypass it.
+	ctx = datastore.CtxSetTxn(ctx, token)
+	ctx = corekv.SetCtxTxn(ctx, nil)
+	ctx = id.InitCollectionShortIDCache(ctx)
+	ctx = id.InitFieldShortIDCache(ctx)
+
+	shortID, err := id.GetShortCollectionID(ctx, c.def.CollectionID)
+	if err != nil {
+		return err
+	}
+	c.db.lockSet.CollectionLock(token, shortID)
+
+	multistore := datastore.NewMultistore(c.db.rootstore, c.db.lockSet, c.db.blockStoreChunkSize)
+	return c.runTruncate(ctx, ctx, multistore)
+}
+
+// truncateInTxn truncates the collection through the explicit transaction already on the context.
+// The action-execution markers are still written txn-free (corekv issue #107).
+func (c *collection) truncateInTxn(ctx context.Context) error {
 	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
 	if err != nil {
 		return err
 	}
-
 	defer txn.Discard()
 
 	shortID, err := id.GetShortCollectionID(ctx, c.def.CollectionID)
 	if err != nil {
 		return err
 	}
-
 	c.db.lockSet.CollectionLock(txn, shortID)
 
 	multistore := datastore.NewMultistore(c.db.rootstore, c.db.lockSet, c.db.blockStoreChunkSize)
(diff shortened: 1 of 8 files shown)


internal/db/collection_truncate.go
```diff
@@ -48,18 +48,50 @@ func (c *collection) Truncate(
...
-	ctx, txn, err := ensureContextTxn(ctx, c.db, false)
+	// Truncate performs txn-free writes: it deletes more key-values than a single store transaction
+	// may safely hold, and it records its action-execution markers outside any transaction (so they
+	// survive a rollback). leveldb blocks non-transactional writes while a transaction is open, so
+	// when no explicit transaction is in play we hold the collection write lock with a lock token
+	// rather than an open store transaction, and write everything directly to the rootstore.
+	//
+	// When the caller supplied an explicit transaction we must honour it. That path is unsupported on
+	// leveldb (a txn-free write would block on the open transaction - see issue #4959).
+	if hadTxn {
```
AndrewSisley, line 59: todo: I don't think we want this complication, it will only work when there's only a small amount of data in the collection and makes the function behaviour harder for both us and users to understand.

It also prevents the lock/truncate being successfully scoped to an explicit transaction, which, leveldb problems aside, it can already do.

internal/db/store.go
```diff
@@ -490,24 +513,23 @@ func (db *DB) RefreshViews(ctx context.Context, opts ...options.Enumerable[optio
...
+		ctx, txn, err := ensureContextTxn(ctx, db, false)
+		if err != nil {
+			return err
+		}
+		defer txn.Discard()
 
-	err = txn.Commit()
-	if err != nil {
-		return err
+		if err := db.refreshViewsInTxn(ctx, opt); err != nil {
```
AndrewSisley, line 524: todo: Same as comment on truncate.  I don't think we want this.

The loss of lock is particularly impactful in this function (scoping it to the re-definition of collections is important).

internal/db/store.go
```diff
@@ -459,22 +459,45 @@ func (db *DB) AddView(
...
 	err = txn.Commit()
 	if err != nil {
 		return nil, err
 	}
 
+	// The committed transaction is still referenced by ctx; clear it so the refresh opens its own
+	// fresh transactions rather than reusing the closed one.
+	ctx = datastore.CtxSetTxn(ctx, nil)
+
+	if err := db.refreshNewMaterializedViews(ctx, defs, true); err != nil {
```
AndrewSisley, line 497: todo: I don't think this should be handled outside the lock scope.

This introduces a race condition - the new view will become available, and can be queried (returning nothing) before it is refreshed.

AndrewSisley (changes requested):
Hey Edwardo,

I disagree with the proposed behaviour changes within this PR.  I think it degrades the common behaviour (not-leveldb) for the sake of a niche use case (leveldb/wasm).

I think there is a way avoid this, I'm exploring it now.

edjroz:
> Hey Edwardo,
> 
> I disagree with the proposed behaviour changes within this PR. I think it degrades the common behaviour (not-leveldb) for the sake of a niche use case (leveldb/wasm).
> 
> I think there is a way avoid this, I'm exploring it now.


I think you are onto something and I may have expanded the blast radius undesirably here 🤔 
I had considered as an alternative to remove the `context.TODO()` but I was concerned with oversized transactions and the consequences of removing it meant half-truncations.

You can close it if you feel you have a better approach, on one of my suggestions for the 2nd manifestation is the adding MVCC to corekv/leveldb I guess technically  that would work and preserve the entire functionality as is but as I shared I was cagey on the decisions that it entails.

AndrewSisley:
> You can close it

Okay, will do - our PR backlog is very large atm, and we can re-open this if we need to.  Sorry about the bother, we are in a bit of a rush atm, and I think this is needed for a couple of the last items that need sorting for the v1 release.

PR #4973, opened 2026-06-22
fix: Make startup succeed if TLS is enabled for HTTP API

### LINKED ISSUE #4780: Startup fails when TLS is enabled for HTTP API
Generate TLS keys:
```shell
make tls-certs path="~/.defradb/certs"
```
Start defra providing paths to private and public key
```
stef@carbonino:~/git/defradb$ DEFRA_KEYRING_SECRET=secret defradb start --pubkeypath ~/.defradb/certs/server.crt --privkeypath ~/.defradb/certs/server.key 
Apr 20 15:27:25.501 INF cli Starting DefraDB
Apr 20 15:27:25.520 INF p2p Created LibP2P host PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Address=[/ip4/127.0.0.1/tcp/9171]
Apr 20 15:27:25.521 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=doc-sync
Apr 20 15:27:25.521 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=sync-branchable
Apr 20 15:27:25.521 INF p2p Adding pubsub topic PeerID=12D3KooWFn25iWpsLtJJT4F8o2tUvYzRtrAo7XEQA8QgSmqkF55d Topic=encryption
Apr 20 15:27:25.521 INF db Starting without nac (not configured/setup)
Apr 20 15:27:25.522 INF node Providing HTTP API at http://127.0.0.1:9181 PlaygroundEnabled=false
Apr 20 15:27:25.522 INF node Providing GraphQL endpoint at http://127.0.0.1:9181/api/graphql
```
Observe startup fail with error
```
2026/04/20 15:27:25 http: TLS handshake error from 127.0.0.1:52132: client sent an HTTP request to an HTTPS server
Error: 400: Client sent an HTTP request to an HTTPS server.
```

ChrisBQu:
## Relevant issue(s)

Resolves #4780 

## Description

The bug described in the issue was caused by a race condition inside `node/node_api.go`. The `Server.Serve()` function was called in a goroutine that was responsible for setting a boolean flag, `isTLS`, on the server struct. Immediately afterwards, on the main goroutine, `Address()` was called before thee `isTLS` flag had a chance to be updated. As a result, it would basically always be false, and the function would return a basic "http" URL even when TLS was configured. This caused a mismatch between the request sent, which was a plain HTTP request, and the HTTPS server.

To fix this, this PR removes the `isTLS` flag entirely, and instead changes the `Address()` function so that it gets the schema from the server's own options -- guaranteed to be accurate at the time of resolution. 

Additionally, it came up that even with the correct URL, the health check would still fail. This was because the `http.DefaultClient` rejects self-signed certificates. To combat this, a new client, `NewInsecureClient`, is introduced. It is exclusively used for this loopback self-check, and so is (I believe) safe.

This PR introduces two new unit tests:

`TestServerAddressWithTLSBeforeServe`
`TestStartAPIWithTLS`

The first test illustrates that the race condition that was described has been addressed and fixed by asserting that an "https" URL is obtained from the `Address()` function. The second test explicitly tests what the issue originally pointed out: that a node can be created, configured to have TLS, and that there will be no error. It then, also, confirms that the address contains "https".
 
## Tasks

## How has this been tested?

In addition to the integration testing included, this was manually tested:

`./build/defradb start --pubkeypath ~/.defradb/certs/server.crt --privkeypath ~/.defradb/certs/server.key --no-keyring`

Thee bug was recreatable on previous versions of Defra, but is resolved with these changes, and the 400 error is no longer returned.

- WSL

### DIFF
--- a/node/node_api.go
+++ b/node/node_api.go
@@ -63,7 +63,7 @@ func (n *Node) startAPI(ctx context.Context) error {
 	// not trusted by the system CA pool, so we skip certificate verification for
 	// this loopback health check only.
 	var c *http.Client
-	if n.opts.HTTP.TLSCertPath != "" {
+	if n.opts.HTTP.TLSCertPath != "" && n.opts.HTTP.TLSKeyPath != "" {
 		c, err = http.NewInsecureClient(n.APIURL)
 	} else {
 		c, err = http.NewClient(n.APIURL)
--- a/node/node_api.go
+++ b/node/node_api.go
@@ -58,7 +58,16 @@ func (n *Node) startAPI(ctx context.Context) error {
 	n.APIURL = n.server.Address()
 	// Check that the server is ready before returning. We do this to ensure that
 	// subsequent operation will behave as expected.
-	c, err := http.NewClient(n.APIURL)
+	//
+	// When TLS is configured the server uses a self-signed certificate that is
+	// not trusted by the system CA pool, so we skip certificate verification for
+	// this loopback health check only.
+	var c *http.Client
+	if n.opts.HTTP.TLSCertPath != "" {
+		c, err = http.NewInsecureClient(n.APIURL)
+	} else {
+		c, err = http.NewClient(n.APIURL)
+	}
 	if err != nil {
 		return err
 	}
--- a/http/client.go
+++ b/http/client.go
@@ -54,6 +54,16 @@ func NewClient(rawURL string) (*Client, error) {
 	return &Client{httpClient}, nil
 }
 
+// NewInsecureClient returns a Client that skips TLS certificate verification.
+// Only use for loopback health checks against a server with a self-signed cert.
+func NewInsecureClient(rawURL string) (*Client, error) {
+	httpClient, err := newInsecureHttpClient(rawURL)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{httpClient}, nil
+}
+
 func (c *Client) NewTxn(readOnly bool) (client.Txn, error) {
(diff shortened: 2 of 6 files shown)


node/node_api.go
```diff
@@ -58,7 +58,16 @@ func (n *Node) startAPI(ctx context.Context) error {
 	n.APIURL = n.server.Address()
 	// Check that the server is ready before returning. We do this to ensure that
 	// subsequent operation will behave as expected.
-	c, err := http.NewClient(n.APIURL)
+	//
+	// When TLS is configured the server uses a self-signed certificate that is
+	// not trusted by the system CA pool, so we skip certificate verification for
+	// this loopback health check only.
+	var c *http.Client
+	if n.opts.HTTP.TLSCertPath != "" {
```
nasdf: suggestion: Keep consistency with other checks `n.opts.HTTP.TLSCertPath != "" && n.opts.HTTP.TLSKeyPath != ""`

nasdf (approved):
One small suggestion, but otherwise looks great!

fredcarle (approved):
LGTM

PR #4975, opened 2026-06-22
fix: Add SkipChangeDetector mechanism for tests

### LINKED ISSUE #4974: Fix UTC_Now Change Detector Break
The exepected result of `TestQueryInlineArrayWithDateTime_UTC_NOW` cannot be cached meaningfully, because it needs to resolve to the current timestamp. As such, it doesn't play well with the change detector workflow.

ChrisBQu:
## Relevant issue(s)

Resolves #4974 

## Description

It was previously the case that the `TestQueryInlineArrayWithDateTime_UTC_NOW` test would always cause the change detector to fail. This is because the current time cannot be cached meaningfully by the change detector, as it must be resolved at the time of the test running.

The solution to this problem is to introduce a mechanism for tests to opt out of the change detector. This is done with the `SkipChangeDetector` field. To make it work, we simply create a new boolean for whether or not it should run, and modify the `PreTestChecks` to check for this boolean, bailing early if it is set to true.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/tests/change_detector/utils.go
+++ b/tests/change_detector/utils.go
@@ -77,9 +77,13 @@ func DatabaseDir(t testing.TB) string {
 }
 
 // PreTestChecks skips any test that can't be run by the change detector.
-func PreTestChecks(t testing.TB, collectionNames []string) {
+func PreTestChecks(t testing.TB, collectionNames []string, skipChangeDetector bool) {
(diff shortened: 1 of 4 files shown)


shahzadlone (approved):
LGTM

PR #4976, opened 2026-06-22
perf: Batch index rebuild on collection version switch

### LINKED ISSUE #4949: Sec. Indexes: Batch index rebuild on collection version switch (reindexNewActiveVersion)
Follow-up to https://github.com/sourcenetwork/defradb/issues/296

## Context

https://github.com/sourcenetwork/defradb/issues/296 (backfill) decoupled index creation and deletion from the definition transaction: existing documents are now indexed (and index entries removed) in batched transactions, so an index can be created or dropped on a collection of any size without exceeding the storage engine's transaction size limit.

One index-rebuild path was intentionally left out of that work: `reindexNewActiveVersion`. It runs when a collection's active version switches (lens migration, or a version switch via patch) and rebuilds every index by doing `RemoveAll` followed by a full re-index of all documents, all inside a single transaction. This is the same failure mode #296 fixed for create/delete, just reached through a different trigger: on a large collection it will exceed the transaction size limit and fail.

This matters because migrations and version switches are a common, repeatable operation (switching the active version back and forth is easy), so the limit is reachable in normal use, not just an edge case.

## Proposal

Rework `reindexNewActiveVersion` to reuse the batched primitives introduced by #296 (batched delete + batched backfill) instead of single-transaction `RemoveAll` + full re-index. The batched build/GC functions, the per-index lifecycle state, and the startup-recovery resume already exist, so this is mostly wiring:

- Replace the per-index `RemoveAll` + full re-index with the batched garbage collection followed by the batched backfill.
- The callers run inside the migration/patch transaction, so the rebuild must be deferred to after that transaction commits (the same implicit/explicit-transaction handling create and delete already use), rather than running inline in the open transaction.
- Plugging into the existing lifecycle state means an interrupted version-switch rebuild is resumed by startup recovery for free.

Once this lands, the only remaining caller of the single-transaction `RemoveAll` is the new-collection path, which only ever runs over a collection with no documents, so that path stays trivially safe.

## Acceptance criteria

- Switching the active version of a large collection (enough data to exceed a single transaction) rebuilds its indexes successfully.
- The rebuild runs after the version-switch transaction commits, and an interrupted rebuild is recovered on the next startup.
- Queries against the collection use the indexes correctly after the switch.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4949

## Description

Switching a collection's active version (a lens migration, or a version switch via patch)
rebuilds every secondary index. Previously that rebuild ran inline in the version-switch
transaction as `RemoveAll` + a full re-index of all documents — the same single-transaction
failure mode that #296 fixed for index create/delete, just reached through a different trigger.
On a large enough collection it exceeded the storage engine's transaction size limit and failed.

This reworks the rebuild to be batched, deferred, and crash-safe, by namespacing index entries
with an **epoch**:

- Each index's entries carry an epoch component in their key (`/collection/index/<epoch>/...`).
  The active epoch is the current value of a per-index epoch sequence — it is resolved from the
  sequence wherever the index is read or written, not stored on the index description.
- A rebuild allocates a fresh epoch and backfills it in batched transactions (reusing the #296
  primitives), while live writes and the to-be-discarded old epoch are left untouched. The build
  alone gates queryability: a building index is excluded from query planning, so queries fall back
  to a scan until the new epoch is complete, and never read a half-built index.
- The rebuild is staged inside the version-switch transaction and the batched work runs after that
  transaction commits, matching how create/delete already defer their batched work.
- Once the new epoch is built, every superseded epoch below it is garbage-collected in one batched
  range delete. This needs no per-rebuild record: the entries left below the live epoch are
  self-describing, so an interrupted collection is simply repeated. Startup recovery sweeps stale
  epochs across all active indexes, so an interrupted version-switch rebuild converges on the next
  start with no leaked entries.

A whole-index drop (`DeleteIndex`) is unchanged in spirit — it removes the definition and
collects every epoch.

## Tasks

- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project: https://github.com/sourcenetwork/defradb/blob/develop/CONTRIBUTING.md#commit-messages).

## How has this been tested?

- Unit tests for the epoch key encode/decode (byte layout, round-trip), the rebuild path
  (fills new epoch, collects old, stays queryable), and recovery of an interrupted stale-epoch
  collection.
- The migration + secondary-index integration cases that were previously excluded now pass.
- Full `internal/keys`, `internal/db`, and the index / migration integration suites.

- macOS (Apple Silicon)


tests/integration/collection_version/migrations/query/with_index_rebuild_test.go
```diff
@@ -0,0 +1,271 @@
+// Copyright 2026 Democratized Data Foundation
```
AndrewSisley: question: These all test existing behaviour, and all pass when run against develop.  We already have a set of migration tests for secondary indexes, and they look very similar to these.

What differences are there between the existing set of tests, and these new ones?  Are they duplicated? 

todo: If duplicated, please deduplicate them.

internal/keys/datastore_index.go
```diff
@@ -142,6 +153,26 @@ func DecodeIndexDataStoreKey(
 		return key, nil
 	}
 
+	// A non-zero epoch is a component between the index ID and the fields; consume it only
+	// when scanning a non-zero epoch keyspace.
+	if epoch != 0 {
```
AndrewSisley, line 160: question: The documentation on `IndexDataStoreKey.Epoch` says that a key will never be written with a zeroed epoch.  Is this check dead code?  Will it always be true?

todo: Please document the reason for this check or delete it.

internal/keys/systemstore_seq_index_epoch.go
```diff
@@ -0,0 +1,47 @@
...
+
+// IndexEpochSequenceKey keys the sequence that allocates index entry epochs.
+//
+// The sequence is specific to a single index. Each rebuild draws the next value as its
+// building epoch, so a building epoch never reuses the active epoch or one left behind by
+// an interrupted rebuild whose entries have not yet been collected.
+//
+// CollectionID is the stable collection ID, so the sequence survives version switches.
+type IndexEpochSequenceKey struct {
+	CollectionID string
```
AndrewSisley: suggestion: This can use the short collection id if you like, it looks like you already have it when reading this key so it shouldn't complicate the code I think.

internal/db/store.go
```diff
@@ -367,12 +371,18 @@ func (db *DB) SetActiveCollectionVersion(
...
 	}
 
-	return txn.Commit()
+	if err := txn.Commit(); err != nil {
+		return err
+	}
+
+	// The rebuild drives its own batched transactions, so it runs after the commit that
+	// recorded the builds.
+	return runRebuild(ctx)
```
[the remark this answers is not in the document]
AndrewSisley: @islamaliev I think the rabbit is correct here, please consider this a `todo` before merge.

`SetMigration` also has this problem.

suggestion: Might be easiest to solve this by executing `runRebuild` via the `OnSuccess` func of the txns.

AndrewSisley (changes requested):
LGTM, but I think a few things need tweaking before merge.

internal/keys/systemstore_seq_index_epoch.go
islamaliev: will do

internal/keys/datastore_index.go
islamaliev: updated the comment

tests/integration/collection_version/migrations/query/with_index_rebuild_test.go
islamaliev: you are right. removed

AndrewSisley (approved):
LGTM, thanks Islam!

PR #4982, opened 2026-06-22
fix(i): Make introspection schema deterministic

### LINKED ISSUE #4981: WriteSDL output is non-deterministic, letting generated SDL fixtures go stale
## Summary

`SchemaManager.WriteSDL` (`internal/request/graphql/schema/manager.go`) serializes an introspection query whose types and fields are ordered by Go map iteration, so the emitted SDL is **non-deterministic** — the byte output differs run to run.

The committed golden fixtures (`internal/request/graphql/schema/testfixtures/schema.{simple,relatedone,relatedmany}.gen.graphql`) are compared against fresh output by the npx tests (`manager_npx_test.go`, `cli/sdl_generate_npx_test.go`) using `@graphql-inspector/cli diff`, which is **order-insensitive**. That tooling choice is the only reason the tests pass despite the non-determinism — but it also masks the problem: there was no stable way to regenerate the fixtures, so they silently drifted out of date.

## Impact

PR #4915 ("Finalize DateTime inline array support") added five DateTime list/filter types to the generator:

- `DateTimeFilterArg`
- `DateTimeListOperatorBlock`
- `NotNullDateTimeFilterArg`
- `NotNullDateTimeListOperatorBlock`
- `NotNullDateTimeOperatorBlock`

…but never regenerated the golden fixtures. Because the npx tests run in a separate CI lane (`make test:npx`), the staleness wasn't caught until those tests were run, failing with `Detected 5 breaking changes` across all `TestSDLGenerate_*` and `TestWriteSDL_*` tests.

## Root cause

1. Non-deterministic introspection ordering in `WriteSDL` → fixtures can't be cleanly regenerated (every regen reshuffles the whole file).
2. No make target / documented path to regenerate the fixtures, so #4915 had no obvious step to update them.

## Fix

- Sort the introspection result by name before printing in `WriteSDL`, making SDL output deterministic.
- Regenerate the three golden fixtures (now sorted + include the DateTime types).
- Add a `DEFRA_UPDATE_SDL_FIXTURES`-gated regeneration path in the npx test (exits before the npx diff, so node is not required), exposed as `make sdl-fixtures` and wired into `make fix` so the fixtures stay current going forward.

## Notes

- Side effect (improvement): `defradb client schema sdl generate` output is now alphabetically sorted and stable.
- Follow-up worth considering: the npx tests only run in a dedicated lane; consider surfacing fixture staleness in the default test path so it's caught earlier.

fredcarle:
## Relevant issue(s)

Resolves #4981 

## Description

This PR makes the introspection schema generation deterministic (open to pushback on that one) and regenerates the test fixtures used by the NPX/JS CI workflow.

## How has this been tested?

Ran the NPX/JS tests

- MacOS


internal/request/graphql/schema/manager_npx_test.go
```diff
@@ -95,6 +96,14 @@ func runWriteSDLTest(t *testing.T, sdl string, fixtureName string) {
 
 	testFixturePath := getFullFixturePath(fixtureName)
 
+	// Regenerate the committed SDL fixtures from the current generator output.
+	// Run via `make sdl-fixtures` rather than setting this directly. This path
+	// exits before the npx diff, so node is not required.
+	if os.Getenv("DEFRA_UPDATE_SDL_FIXTURES") != "" {
```
jsimnz: suggestion: fixture generation should stay as an independent step, rather than implicitly done during a test run

jsimnz (approved):
Approving, up to you if we want to handle the suggestion

PR #4985, opened 2026-06-22
fix(i): Add truncate support to LevelDB

### LINKED ISSUE #4984: Add implicit-txn truncate support to leveldb
Broken out of https://github.com/sourcenetwork/defradb/issues/4959

AndrewSisley:
## Relevant issue(s)

Resolves #4984

## Description

Adds implicit-txn truncate support to LevelDB.

RefreshViews and AddView will also make use of the new shim, as detailed in https://github.com/sourcenetwork/defradb/issues/4959

Explicit transaction support can be added later for LevelDB, it is low priority (not v1) and a nice solution to that is probably a bit more involved.

### DIFF
--- a/internal/datastore/txn_shim.go
+++ b/internal/datastore/txn_shim.go
@@ -27,11 +27,9 @@ type TxnShim struct {
 	ts time.Time
 
 	successFns []func()
-	errorFns   []func()
 	discardFns []func()
 
 	successAsyncFns []func()
-	errorAsyncFns   []func()
 	discardAsyncFns []func()
 }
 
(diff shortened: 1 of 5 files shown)


internal/datastore/txn_shim.go
```diff
@@ -0,0 +1,126 @@
...
+// Calling any of the child store accessors will result in a panic.
+// The majority of Defra does not check for nil being returned from these
+// functions, so we might as well panic within this type where the stack
+// trace should be a little more obvious.
+type TxnShim struct {
+	id uint64
+	ts time.Time
+
+	successFns []func()
+	errorFns   []func()
```
fredcarle: question: This and `errorAsyncFns` are never used as the commit path never errors and thus never call the `OnError` methods. Why do we need these in the struct then?

fredcarle (approved):
LGTM. Just one question to answer before merge.

internal/datastore/txn_shim.go
AndrewSisley: I was torn on this and inertia won.  OnError is required to satisfy the interface, and I breifly was yielding errors via commit instead of panicing in the unimplemented funcs, they were also copy-pasted from the real Txn struct initially.

I'll remove them before merge since you are asking :)

- [x] remove dead props

PR #4987, opened 2026-06-23
fix: Make non-existent transactions return an error

### LINKED ISSUE #4977: Non-existant transactions do not return an error
When an invalid transaction ID is used in a CLI request it should return an error instead of silently ignoring it.

This bug may also exist in the HTTP client / server.

```
$ defradb client tx new
{"id":27}

$ defradb client query ' mutation { add_JJ(input:
{ title: "lala" }){title}}' --tx 27
------ Request Results ------
{
  "data": {
    "add_JJ": [
      {
        "title": "lala"
      }
    ]
  }
}

$ defradb client tx commit 27
```

ChrisBQu:
## Relevant issue(s)

Resolves #4977

## Description

`http/client.go` has been updated such that, in the case of a non-200 response code (with 200 meaning "okay"), a GQL error would be returned. Previously, this was just being ignored, and so a null object was quietly returned to the data field.

`http/middleware.go` was also updated in two places. In the case of an error being present, we now return the error explicitly. Before, the error was silently passing through. 

Later, in that file, we call `Load()` on the transaction ID, and check whether this was successful. In the case of it being unsuccessful, we return an error indicating that the transaction was not found. This is a minor, but significant change. Previously, an error of type `ErrTxnDiscarded` was being returned here, but there is actually no reason we can assume that a transaction was discarded here. It could also be that such a transaction was never valid to begin with. This change was discussed on the dev ex discord channel, where it was agreed to be appropriate. As such, the return type of the error was changed throughout `internal/db/txn.go` and `/internal/db/txn_collection.go` where appropriate.

This PR includes two new unit tests for this issue, one of which tests non-numeric transaction IDs, and the other which tests a numeric, but non-existent, transaction ID:

`TestTransactionMiddleware_WithInvalidTxID_ReturnsBadRequest`
`TestTransactionMiddleware_WithNonExistentTxID_ReturnsNotFound`

Lastly, existing tests that checked for the previously mentioned discarded transaction error, now check for the "transaction not found" error insteead.

## Tasks

## How has this been tested?

In addition to the included tests, this was tested manually on both the CLI, and through the HTTP endpoint.

- WSL

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -256,8 +256,6 @@ func ReviveError(message string) error {
 	switch message {
 	case ErrTransactionNotFound.Error():
 		return ErrTransactionNotFound
-	case ErrTransactionNotFound.Error():
-		return ErrTransactionNotFound
 	case ErrDocumentNotFoundOrNotAuthorized.Error():
 		return ErrDocumentNotFoundOrNotAuthorized
 	case corekv.ErrTxnConflict.Error():
--- a/client/errors.go
+++ b/client/errors.go
@@ -57,6 +57,7 @@ const (
 	errSetDocFieldValue                      string = "failed to set document field value"
 	errNullValueForNonNillableField          string = "null value provided for non-nillable field"
 	errMissingRequiredField                  string = "value not provided for non-nillable field"
+	errTransactionNotFound                   string = "transaction not found"
 )
 
 // Errors returnable from this package.
@@ -105,6 +106,7 @@ var (
 	ErrDocumentJSONParseFailed               = errors.New(errDocumentJSONParseFailed)
 	ErrNullValueForNonNillableField          = errors.New(errNullValueForNonNillableField)
 	ErrMissingRequiredField                  = errors.New(errMissingRequiredField)
+	ErrTransactionNotFound                   = errors.New(errTransactionNotFound)
 )
 
 // NewErrFieldNotExist returns an error indicating that the given field does not exist.
--- a/client/errors.go
+++ b/client/errors.go
@@ -254,6 +254,10 @@ func NewErrFailedToParseKind(kind []byte) error {
 // is wrapped in a new anonymous error type.
 func ReviveError(message string) error {
 	switch message {
+	case ErrTransactionNotFound.Error():
+		return ErrTransactionNotFound
+	case ErrTransactionNotFound.Error():
+		return ErrTransactionNotFound
 	case ErrDocumentNotFoundOrNotAuthorized.Error():
 		return ErrDocumentNotFoundOrNotAuthorized
 	case corekv.ErrTxnConflict.Error():
--- a/http/client.go
+++ b/http/client.go
@@ -14,6 +14,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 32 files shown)


client/errors.go
```diff
@@ -252,6 +254,10 @@ func NewErrFailedToParseKind(kind []byte) error {
 // is wrapped in a new anonymous error type.
 func ReviveError(message string) error {
 	switch message {
+	case ErrTransactionNotFound.Error():
+		return ErrTransactionNotFound
+	case ErrTransactionNotFound.Error():
```
nasdf, line 257: todo: Remove duplicate case

http/client.go
```diff
@@ -558,6 +558,19 @@ func (c *Client) ExecRequest(
...
+	// Convert those so the error surfaces to the caller instead of being silently
+	// swallowed as {"data": null}.
+	if res.StatusCode != http.StatusOK {
+		var raw map[string]any
+		if jsonErr := json.Unmarshal(data, &raw); jsonErr == nil {
+			if errMsg, ok := raw["error"].(string); ok {
+				result.GQL.Errors = append(result.GQL.Errors, client.ReviveError(errMsg))
+				return result
+			}
+		}
```
nasdf, line 573: suggestion: We should always return an error when the status is not 200. I think wrapping the raw body string in an error string would work.

nasdf (approved):
I left a minor change to consider before merging. Thanks for fixing this!

internal/db/txn.go
```diff
@@ -206,7 +206,7 @@ func (txn *Txn) PrintDump(ctx context.Context) error {
 	defer unlock()
 
 	if txn.isClosed {
-		return ErrTxnDiscarded
```
AndrewSisley, line 209: suggestion: I haven't checked properly, but if `ErrTxnDiscarded` is now dead, please make sure it is removed.

http/errors.go
```diff
@@ -50,7 +50,7 @@ var (
 	ErrMissingIdentity              = errors.New("required identity is missing")
 	ErrInvalidSubscriptionTransport = errors.New("invalid subscription transport")
 	ErrInvalidGraphQLRequest        = errors.New("invalid graphql request")
-	ErrTransactionNotFound          = errors.New("transaction not found")
+	ErrTransactionNotFound          = client.ErrTransactionNotFound
```
AndrewSisley: todo: I can't spot a reason to keep this if we have defined the same error in the `client` package.  Please delete this and update the refs to use the one in `client`.

AndrewSisley (approved):
LGTM, and thanks for cleaning up after me :) Please resolve the comments before merge

internal/db/txn.go
ChrisBQu: Good catch. Done.

http/errors.go
ChrisBQu: So, there was a reason for it. This ties to the http package to the client package in just one place. The `error.go` file imports it because it has to. But then the rest of the http package can use the error that will already be defined in the scope of that package. That was the reasoning.

But... it's actually not used in a lot of spots. I think this saves _one_ import statement, and has more lines of code overall. So, very happy to make this change. Good catch. Just wanted to explain the reasoning in the case of it seeming really odd.

AndrewSisley: The `client` package is required by pretty much everything outside of some very low level stuff, unless you guys want to decouple http/cli from client in the future I don't think `client` imports are something that we should worry about here.

PR #4838, opened 2026-05-27
feat: DocID based on genesis CID

iverc:
## Relevant issue(s)

Resolves #4912

## Description

This PR changes document identity so the public DocID is derived from the CID of the first materialized document composite block.

That creates a small chicken-and-egg problem during document creation: the public DocID cannot be known until after the first composite block is written, but the datastore still needs a stable document key while building the CRDT deltas. To solve that, DefraDB now uses a node-local short document ID internally for datastore/headstore keys, then persists a bidirectional mapping between the short ID and the public CID-derived DocID once the genesis composite CID is known.

For genesis document blocks, the DocID is intentionally omitted from the encoded delta payload. This keeps unsigned creates deterministic across nodes: if two nodes create the same document content without signatures, they materialize the same genesis composite CID and therefore the same public DocID. Follow-up document composites can carry the public DocID normally because it is known by then.

The change also updates the read, merge, query, indexing, P2P, CAR/KV import-export, purge/truncate, verification, and C bindings paths so external APIs continue to work with public DocIDs while storage uses short IDs internally. This includes preserving _docID index/query behavior via a reserved internal DocID index and remapping short IDs during cross-node import.

## Tasks

- [x] I made sure the code is well commented, particularly hard-to-understand areas.
- [x] I made sure the repository-held documentation is changed accordingly.
- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project can be found in [tools/configs/chglog/config.yml](tools/configs/chglog/config.yml)).
- [x] I made sure to discuss its limitations such as threats to validity, vulnerability to mistake and misuse, robustness to invalidation of assumptions, resource requirements, ...

## How has this been tested?

(*replace*) Describe the tests performed to verify the changes. Provide instructions to reproduce them.

Specify the platform(s) on which this was tested:
- *(modify the list accordingly*)
- Arch Linux
- Debian Linux
- MacOS
- Windows

PR #4990, opened 2026-06-24
feat: Branchable collection ACP

### LINKED ISSUE #3883: Branchable collection ACP
Currently, individual documents can be registered on the ACP system to enforce access rights. We should add this functionality at the branchable collection level as well.

fredcarle:
## Relevant issue(s)

Resolves #3883

## Description

This PR adds ACP enforcement to the branchable collections. Both from client requests and P2P sync. It uses the collectionID as the ACP object ID.

Note that some of the changes will need to be edited once the DocID generation PR #4838 merges (it should be merged first) but the logic will stay the same.

I took the opportunity to remove the redundant `WithACP` from a lot of the ACP functions so many line changes are related to that.

## Tasks

## How has this been tested?

Added branchable collection specific tests.

- MacOS


internal/db/acp/register.go
```diff
@@ -49,3 +49,38 @@ func RegisterDocOnCollectionWithDocumentACP(
...
+// (3) the request is permissioned (has an identity, which becomes the object owner).
+//
+// Otherwise, nothing is registered with document acp (the collection-level commits remain public).
+func RegisterCollectionObject(
+	ctx context.Context,
+	identity immutable.Option[acpIdentity.Identity],
+	documentACP dac.DocumentACP,
+	collection client.Collection,
+) error {
+	if !collection.Version().IsBranchable {
```
AndrewSisley: question: Is this check useful?

todo: If it is, please document why.

suggestion: If not, please consider deleting this function and renaming the existing `RegisterDocOnCollectionWithDocumentACP` to something like `RegisterCollectionObject` and its last param from `docID` to `id`.

internal/db/p2p/p2p.go
```diff
@@ -459,6 +451,47 @@ func (p *P2P) hasAccess(ctx context.Context, pid string, c cid.Cid) bool {
...
+	ctx context.Context,
+	identityFunc func() immutable.Option[identity.Identity],
+	col client.Collection,
+	block *coreblock.Block,
+) (bool, error) {
+	objectIDs := make([]string, 0, 2)
+	if docID := string(block.Delta.GetDocID()); docID != "" {
+		objectIDs = append(objectIDs, docID)
+	}
+	if col.Version().IsBranchable {
```
AndrewSisley: suggestion:  This feels like the wrong check to be making, and on the wrong blocks.

I *feel* like document blocks should not be blocked from syncing due to permissions pertaining to their parent collection-block.

I *feel* like we should really be checking to see if a **block** is a branchable collection block, and if then, check to see if we need to check it's ID.

If you disagree, please do let me know your thoughts, as I'm currious as to why this is done this way!

todo: If keeping the current check, please document why, it feels counter intuitive to me.

internal/db/collection_define.go
```diff
@@ -98,6 +98,12 @@ func (db *DB) addCollections(
 			return nil, err
 		}
 
+		// Register branchable, permissioned collections as an acp object so that access to their
+		// collection-level commit DAG can be gated. This is a no-op for all other collections.
```
AndrewSisley: suggestion:  If a collection isn't branchable, then we shouldn't try to register it.  It might look a bit noisier here, but it is clearer.

tests/integration/acp/dac/branchable/peer_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"testing"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+const commitsQuery = `
```
AndrewSisley, line 24: thought: Doing this and defining shared docs and sdl in fixtures.go makes the tests harder to read and couples them together.

suggestion: declare them in  the test so readers of the test can easily see what is going on, instead of constantly flicking between files/lines.

tests/integration/acp_dac.go
```diff
@@ -280,6 +280,101 @@ type DeleteDACActorRelationship struct {
 	ExpectedError string
 }
 
+// AddDACCollectionActorRelationship will attempt to add a new relationship between a branchable
+// collection's commit DAG (the collection object) and an actor.
+//
+// This is the collection-level analogue of [AddDACActorRelationship]: instead of sharing access to
+// a single document, it shares access to a branchable collection's collection-level commit DAG.
+type AddDACCollectionActorRelationship struct {
```
AndrewSisley: todo: Please define this in `tests/action` as implementing the testo interfaces - declaring new actions here creates immediate technical debt that we need to clean up later.

AndrewSisley (approved):
LGTM, but please resolve the comments before merge.

tests/integration/acp_dac.go
fredcarle: I forgot about that. No ACP actions are there so I didn't even think about it.

internal/db/acp/register.go
fredcarle: Function deleted :)

internal/db/collection_define.go
fredcarle: Done

tests/integration/acp/dac/branchable/peer_test.go
fredcarle: I knew you were going to say this :)
I also know that Islam prefers this approach but it's true that it's easier to read the tests when they are inline.

internal/db/p2p/p2p.go
fredcarle: There was a little mistake in the access check logic. Here is what it should be:

Access is granted if EITHER:
 - the actor was explicitly granted read access to the document — an actor with whom a specific
    document was shared may read that document even without access to the rest of a branchable
    collection; OR
  - the actor can read every object the document relates to: the document itself (when present)
    and, for a branchable collection, the collection object (gated using the collection id).
    Requiring collection access in the non-explicit case keeps a private branchable collection
    gating its public (unregistered) documents too.

internal/db/acp/check.go
```diff
@@ -86,6 +86,36 @@ func CheckDocAccessWithIdentityFunc(
...
+// distinguishes access that was explicitly granted via an acp registration from unrestricted
+// ("public") access.
+type docAccess struct {
+	// hasAccess is the final verdict: whether the actor may access the object.
+	hasAccess bool
+
+	// explicit is true when the verdict was decided by an explicit acp registration of the object
+	// (the object is registered, so the actor either was or was not granted a relationship on it).
+	// It is false for unrestricted access, i.e. a nac bypass, an unpermissioned collection, or a
+	// public (unregistered) object.
```
shahzadlone: todo: What is a `nac bypass`? I am guessing you mean DAC Bypass through NAC? Even in that case `explicit` should be `true`, not `false`. You do so correctly on line 138 of this file, but this comment contradicts.

internal/db/acp/check.go
```diff
@@ -102,16 +132,17 @@ func CheckDocAccessWithIdentityFunc(
 		identityValue = identity.Value().DID()
 	}
 
-	// Check if can bypass DAC, if not then continue DAC.
+	// Check if can bypass DAC, if not then continue DAC. A nac bypass is an explicit privilege
```
shahzadlone: nitpick: 

```suggestion
	// Check if can bypass DAC, if not then continue DAC. A nac dacbypass is an explicit privilege
```

internal/planner/commit.go
```diff
@@ -248,14 +247,16 @@ func (n *dagScanNode) Next() (bool, error) {
...
-	if n.planner.documentACP.HasValue() && len(docIDBytes) > 0 {
+	// doc-mapping cost. We only check when document acp is configured.
+	//
+	// A commit is gated by the read access of the document it belongs to: an explicit grant on the
+	// document is sufficient on its own, otherwise — for a branchable collection — the caller also
+	// needs access to the collection object, so a private branchable collection gates its entire
+	// commit DAG (collection-level commits, which carry no docID, are gated on the collection object
+	// only). [CheckDocReadAccess] short-circuits when the collection has no policy or the object is
+	// public (unregistered), so a no-docID delta that is not a branchable collection commit (e.g.
+	// schema-definition changes) is not gated.
```
shahzadlone: nitpick:  
```suggestion
	// schema definition changes) is not gated.
```

internal/planner/commit.go
shahzadlone: question: Should we add .SetIdentity(n.planner.identity) to the GetCollections opts?

Like what we have in verify.go:87-90 ish: 
```
		// Pass the requester's identity so the collection lookup is authorised as them (rather than
		// anonymously) when node acp gates the get-collection operation.
		getColOpts := options.GetCollections().SetGetInactive(true).SetVersionID(versionID)
		if opt.Identity.HasValue() {
			getColOpts = getColOpts.SetIdentity(opt.Identity.Value())
		}
```

tests/integration/acp/dac/branchable/collection_commits_test.go
shahzadlone, line 1: suggestion: Add tests to check stranger update/delete-denied, non-owner AddDACCollectionActorRelationship (use its ExpectedError), encrypted branchable key-denial

internal/db/acp/check.go
```diff
@@ -149,10 +180,89 @@ func CheckDocAccessWithIdentityFunc(
...
+//   - the actor was explicitly granted read access to the document — an actor with whom a specific
+//     document was shared may read that document even without access to the rest of a branchable
+//     collection; OR
+//   - the actor can read every object the document relates to: the document itself (when present)
+//     and, for a branchable collection, the collection object (gated using the collection id).
+//     Requiring collection access in the non-explicit case keeps a private branchable collection
+//     gating its public (unregistered) documents too.
+//
+// For a non-branchable collection this reduces to the document's own read access. An explicit denial
+// on the document always denies, regardless of collection access (the document's own acp wins).
```
shahzadlone, line 206: suggestion: It would be nice to state explicitly that the gate covers the commit DAG/history, not just current doc field data

tests/action/add_dac_collection_actor_relationship.go
```diff
@@ -0,0 +1,133 @@
...
+// getIdentityDID returns the DID of the given identity, or the empty string if no identity is set.
+func getIdentityDID(s *state.State, identity immutable.Option[state.Identity]) string {
+	if identity.HasValue() {
+		if identity.Value().Selector == "*" {
+			return identity.Value().Selector
+		}
+		return state.GetIdentity(s, identity).DID()
+	}
+	return ""
+}
```
shahzadlone: todo: Move to `tests/state/identity.go` beside `GetIdentity` to have one shared `getIdentityDID`

internal/db/verify.go
```diff
@@ -75,26 +76,37 @@ func (db *DB) VerifySignature(
 	}
 
 	if db.documentACP.HasValue() {
-		docID := string(block.Delta.GetDocID())
-		collection, err := NewCollectionRetriever(db).RetrieveCollectionFromDocID(ctx, docID, opt.Identity)
+		// Verifying a signature requires read access to the block's document (and therefore the
+		// block). This mirrors the commit gating in the planner (internal/planner/commit.go): an
+		// explicit grant on the document suffices, otherwise a branchable collection also gates on
+		// the collection object. We resolve by version id, not docID, so collection-level commits
+		// (which have no docID) are handled too.
```
shahzadlone: nitpick:  Perhaps repoint to `acpDB.CheckDocReadAccess` in `acp/check.go` as canonical? is "mirrors planner" stale here?

internal/kms/pubsub.go
```diff
@@ -569,13 +569,16 @@ func (s *pubSubService) doesIdentityHaveDocPermission(
 		return false, err
 	}
 
-	return acpDB.CheckAccessOfDocOnCollectionWithACP(
+	// Read access to the document gates access to its encryption key. An explicit grant on the
+	// document suffices; otherwise a branchable collection also gates on the collection object, so
+	// a private branchable collection gates its whole DAG. Mirrors the planner
+	// (internal/planner/commit.go).
```
shahzadlone: nitpick:  Perhaps repoint to `acpDB.CheckDocReadAccess` in `acp/check.go` as canonical? is "mirrors planner" stale here?

internal/db/p2p/p2p.go
```diff
@@ -459,6 +450,29 @@ func (p *P2P) hasAccess(ctx context.Context, pid string, c cid.Cid) bool {
 	return peerHasAccess
 }
 
+// checkBlockAccess reports whether the actor resolved by identityFunc may read block.
+//
+// A block is gated by the read access of the document it belongs to (see
+// [acpDB.CheckDocReadAccessWithIdentityFunc]): an explicit grant on the document is sufficient on
+// its own, otherwise — for a branchable collection — the actor additionally needs access to the
+// collection object, so a private branchable collection gates its entire commit DAG. This mirrors
+// the gating the planner applies to commit queries (internal/planner/commit.go).
```
shahzadlone: nitpick:  Perhaps repoint to `acpDB.CheckDocReadAccess` in `acp/check.go` as canonical? is "mirrors planner" stale here?

internal/db/p2p/p2p.go
shahzadlone: question: shouldn't we here derive collection/CollectionID from `block.Delta.GetCollectionVersionID()` to protect against spoofing `req.CollectionID`, possibly verify `req.CID == block.GenerateLink()`

shahzadlone, line 670: question: shouldn't we here derive collection/CollectionID from `block.Delta.GetCollectionVersionID()` to protect against spoofing `req.CollectionID`, possibly verify `req.CID == block.GenerateLink()`

internal/kms/pubsub.go
shahzadlone: thought: Possibly out of scope of this PR but was wondering if we should carry collection/version id so the collection-object DAC check can run

shahzadlone (approved):
Looks good overall. Just some questions and comments to resolve before merge.

internal/db/p2p/p2p.go
AndrewSisley: That makes sense to me, and I can't spot any holes in it - thanks for the explanation :)

internal/db/verify.go
fredcarle: Yeah this is stale. Initially the implementation was duplicated throughout. That has been rectified.

internal/kms/pubsub.go
fredcarle: This is going to change a bit as soon as the DocID generation PR gets merged in. I'll see what it looks like afterwards.

internal/db/acp/check.go
fredcarle: Done

internal/db/acp/check.go
fredcarle: done

internal/planner/commit.go
fredcarle: done

internal/planner/commit.go
fredcarle: yes good catch. Done

internal/db/acp/check.go
fredcarle: done

internal/db/p2p/p2p.go
fredcarle: This works as long as we have the version ID locally which is not guaranteed.

tests/action/add_dac_collection_actor_relationship.go
fredcarle: done

tests/integration/acp/dac/branchable/collection_commits_test.go
fredcarle: done

internal/planner/commit.go
shahzadlone: todo: Also pass the requester's identity here as well like your other changes.

fredcarle: done. Thanks for finding this.

internal/db/collection_acp.go
```diff
@@ -46,7 +46,46 @@ func (c *collection) registerDocWithACP(
...
+
+	if policyID, resourceName, hasPolicy := acpDB.IsPermissioned(c); hasPolicy && ident.HasValue() {
+		isRegistered, err := documentACP.IsDocRegistered(ctx, policyID, resourceName, c.Version().CollectionID)
+		if err != nil {
+			return err
+		}
+		if isRegistered {
+			return nil
+		}
+	}
```
shahzadlone: question: RegisterDocObject in `acp/dac/bridge.go` already returns nil for a same-owner re-register and errors for a different owner. The new owner blind pre-check converts that different-owner error into a silent skip i believe?
can remove the redundant pre-check and rely on the bridge's existing idempotency, that will do same thing i believe (no-op on same owner, error on different owner). If you want to keep a pre-check, can gate it on owner equality.

fredcarle: Done

PR #4991, opened 2026-06-24
fix: Fix index and ACP bugs related to doc deletes

### LINKED ISSUE #3554: Possible to bypass ACP permissions using externally managed transactions
Externally managed transactions do not automatically discard when an error is returned, the responsibility is on the user to do so.

ACP functions by returning an error if an actor lacks permission to do something.  

This means that if the ACP check is performed after a write action, the write action may still be committed by the user.

This is done in at least one location:
- Collection.Delete deletes the document's index before checking whether the user has permission to delete the document, meaning an externally managed transaction can delete the indexes to a document that it does not have delete permission for.  This will mean that the document is not returned from future queries that use the index to discover documents, a kind of local soft-delete.

Please review our codebase looking for similar issues when resolving this.  

Externally managed transactions should not be auto-discarded on ACP check failure, as this will prevent users from trying (and failing) to do stuff without rolling back their entire (previously valid) transaction (this is the current behaviour, just noting not to do this in a fancy attempt at preventing this from happening in the future).  We could perhaps come up with a way to automatically cover this with integration tests, although that should probably be done in another ticket.

### LINKED ISSUE #4988: Filter-based document deletion does not clean up index entries
This was discovered while investigating: https://github.com/sourcenetwork/defradb/issues/3554

A fix for that issue, depending on how it is done, may also solve this one.

Here is what is happening:

Inside the `DeleteDocument` function of `internal/db/document.go`, we can see how a document should be correctly deleted. First a call is made to `deleteIndexedDocWithID`, then a call is made to `applyDelete`.

But compare to what we have inside `deleteWithFilter` of `internal/db/document_delete.go`. There is a loop over the docIDs, and we call: `applyDelete`, however there is no accompanying call to `deleteIndexedDocWithID`.

ChrisBQu:
## Relevant issue(s)

Resolves #3554 
Resolves #4988 

## Description

The call to `deleteIndexedDocWithID` was moved from `DeleteDocument` to inside `applyDelete`, after the ACP permission check. With this change, index cleanup only happens if the caller is authorized to delete in the first place. This addresses and resolves the ACP permissions issues. However, it also resolves another issue: that filter-based deletes were never cleaning up indexes at all.

This PR includes two new tests to confirm this behavior:

`TestACP_DAC_DeleteDocByID_WithReaderOnlyAccess_DoesNotOrphanIndex`
`TestDocDeleteWithFilter_WithUniqueIndex_CleansUpIndexEntries`

## Tasks

## How has this been tested?

- WSL


AndrewSisley (approved):
LGTM, and nice spot RE the index issue :)

tests/integration/acp/dac/delete_with_index_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"testing"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+// TestACP_DAC_DeleteDocByID_WithReaderOnlyAccess_DoesNotOrphanIndex verifies that when an
+// identity with read-but-not-delete access attempts to delete a document by ID, the request
+// is denied and the document's index entries remain intact.
+func TestACP_DAC_DeleteDocByID_WithReaderOnlyAccess_DoesNotOrphanIndex(t *testing.T) {
```
shahzadlone, line 26: question: Shouldn't this test #3554 is fixed aswell? as atm i think we are using implicit txns. I would have expected `TransactionID: immutable.Some(0)` to the denied DeleteDoc and a trailing `action.CommitTransaction{TransactionID: 0}` before the final owner index query.

tests/integration/index/delete_test.go
```diff
@@ -111,6 +111,58 @@ func TestIndexDelete_ShouldRemoveIndexFromCollection(t *testing.T) {
...
+						{"name": "John", "age": int64(31)},
+					},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
```
shahzadlone, line 165: suggestion: Perhaps another new multi-match test, non-unique city @index matching 4 docs, each with a unique email @index(unique: true), delete-by-filter, then re-insert all 4 emails.

tests/integration/acp/dac/delete_with_index_test.go
ChrisBQu: I see no reason that it shouldn't. I will make the change.

shahzadlone (approved):
LGTM, please resolve the comments before merging

PR #4994, opened 2026-06-24
feat: Transaction TTL Cache for HTTP API

### LINKED ISSUE #4222: Need a TTL timeout for transaction IDs
Problem: client requests a transaction ID from the HTTP endpoint... something happens in the transmission of this from defra to the client. the transaction just sits there forever, and such lost transactions just leak memory (until a server restart).

Proposed solution (per @jsimnz): transactions need a TTL timeout -- possibly a minute? -- when unused after first creation, to clean up any of these orphaned transactions.

jsimnz:
## Relevant issue(s)

This PR supercedes #4271 which had drastically drifted from develop, so this is a clean reimplementation on top of the the latest develop

Resolves https://github.com/sourcenetwork/defradb/issues/4222 

Description
This adds a TTLCache to the HTTP API Transaction middleware so that HTTP transactions are gated by a timeout and will automatically discard after the timeout has elapsed. 

~~Currently this only refreshes the TTL timeout at the *start* of a transaction, so long transactions (eg large queries) have the (likely) possibility of expiring and therefore discarding during a long lived request. I am implementing this in a follow up PR before v1 (small quick change, but wanted to do it in parallel to the review of this PR).~~ This has been implemented in this PR in the 6f8b0a73224cf1c9bf234186d596746a06a0a765 commit

## How has this been tested?

Manual, unit tests, integration tests

- Linux NixOS

### DIFF
--- a/cli/tx_new.go
+++ b/cli/tx_new.go
@@ -32,12 +32,14 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 		Use:   "new",
 		Short: "Create a new DefraDB transaction.",
 		Long:  `Create a new DefraDB transaction.`,
-		RunE: func(cmd *cobra.Command, args []string) (err error) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			cliClient := mustGetContextCLIClient(cmd)
 
 			var tx client.Txn
+			var err error
+			var ttl time.Duration
 			if txnTTL != "" {
-				ttl, err := parseTxnTTLFlag(txnTTL)
+				ttl, err = parseTxnTTLFlag(txnTTL)
 				if err != nil {
 					return err
 				}
@@ -56,7 +58,8 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Transaction is read only")
-	cmd.Flags().StringVar(&txnTTL, "ttl", "", "Transaction idle TTL as a Go duration string, or seconds if no unit is provided")
+	cmd.Flags().StringVar(&txnTTL, "ttl", "",
+		"Transaction idle TTL as a Go duration string, or seconds if no unit is provided")
 	return cmd
 }
 
--- a/cli/tx_new.go
+++ b/cli/tx_new.go
@@ -13,6 +13,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -26,7 +27,7 @@ type txnTTLClient interface {
 
 func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 	var readOnly bool
-	var txnTTL time.Duration
+	var txnTTL string
 	var cmd = &cobra.Command{
 		Use:   "new",
 		Short: "Create a new DefraDB transaction.",
@@ -35,12 +36,16 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 			cliClient := mustGetContextCLIClient(cmd)
 
 			var tx client.Txn
-			if txnTTL != 0 {
+			if txnTTL != "" {
+				ttl, err := parseTxnTTLFlag(txnTTL)
+				if err != nil {
+					return err
+				}
 				ttlClient, ok := cliClient.(txnTTLClient)
 				if !ok {
 					return fmt.Errorf("transaction ttl is not supported by this client")
 				}
-				tx, err = ttlClient.NewTxnWithTTL(readOnly, txnTTL)
+				tx, err = ttlClient.NewTxnWithTTL(readOnly, ttl)
 			} else {
 				tx, err = cliClient.NewTxn(readOnly)
 			}
@@ -51,6 +56,18 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Transaction is read only")
-	cmd.Flags().DurationVar(&txnTTL, "ttl", 0, "Transaction idle TTL")
+	cmd.Flags().StringVar(&txnTTL, "ttl", "", "Transaction idle TTL as a Go duration string, or seconds if no unit is provided")
 	return cmd
+}
+
+func parseTxnTTLFlag(raw string) (time.Duration, error) {
+	txnTTL, err := time.ParseDuration(raw)
+	if err == nil {
+		return txnTTL, nil
+	}
+	seconds, err := strconv.ParseInt(raw, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return time.Duration(seconds) * time.Second, nil
 }
--- a/cli/tx_new.go
+++ b/cli/tx_new.go
@@ -12,12 +12,21 @@ package cli
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/client"
 )
+
+type txnTTLClient interface {
+	NewTxnWithTTL(bool, time.Duration) (client.Txn, error)
+}
 
 func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 	var readOnly bool
+	var txnTTL time.Duration
 	var cmd = &cobra.Command{
 		Use:   "new",
 		Short: "Create a new DefraDB transaction.",
@@ -25,7 +34,16 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			cliClient := mustGetContextCLIClient(cmd)
 
-			tx, err := cliClient.NewTxn(readOnly)
+			var tx client.Txn
+			if txnTTL != 0 {
+				ttlClient, ok := cliClient.(txnTTLClient)
+				if !ok {
+					return fmt.Errorf("transaction ttl is not supported by this client")
+				}
+				tx, err = ttlClient.NewTxnWithTTL(readOnly, txnTTL)
+			} else {
+				tx, err = cliClient.NewTxn(readOnly)
+			}
 			if err != nil {
 				return err
 			}
@@ -33,5 +51,6 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Transaction is read only")
+	cmd.Flags().DurationVar(&txnTTL, "ttl", 0, "Transaction idle TTL")
 	return cmd
 }
--- a/cli/tx_new.go
+++ b/cli/tx_new.go
@@ -12,7 +12,6 @@ package cli
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 	"time"
 
@@ -43,9 +42,16 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 				if err != nil {
 					return err
 				}
+
+				// `NewTxnWithTTL` is a helper that is specific to the HTTP client
+				// and not part of the default client interface, so we need to
+				// type cast to get to the underlying http client implementation.
+				// While the CLI client is hardcoded to use the HTTP client, the
+				// type system doesn't enforce this, so we need to gate the
+				// accidental unhappy path
 				ttlClient, ok := cliClient.(txnTTLClient)
 				if !ok {
-					return fmt.Errorf("transaction ttl is not supported by this client")
+					return ErrMissingTTLTxn
 				}
 				tx, err = ttlClient.NewTxnWithTTL(readOnly, ttl)
 			} else {
@@ -59,7 +65,7 @@ func MakeTxNewCommand(ctx context.Context) *cobra.Command {
 	}
 	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Transaction is read only")
 	cmd.Flags().StringVar(&txnTTL, "ttl", "",
-		"Transaction idle TTL as a Go duration string, or seconds if no unit is provided")
+		"Transaction idle TTL as a duration string, or seconds if no unit is provided")
 	return cmd
 }
 
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -74,7 +74,7 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 		}
 		tx, ok := txs.Load(id)
 		if !ok {
-			err := ErrMissingOrExpiredTransaction
+			err := client.ErrTransactionNotFound
 			if strings.Contains(req.URL.Path, "/graphql") {
 				responseJSON(rw, httpStatusFromError(err), gqlErrorResponse{err})
 			} else {
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -15,7 +15,6 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/cors"
@@ -44,7 +43,7 @@ func CorsMiddleware(allowedOrigins []string) func(http.Handler) http.Handler {
 }
 
 // ApiMiddleware sets the required context values for all API requests.
-func ApiMiddleware(db client.TxnStore, txs *sync.Map, nodeOpts *options.NodeOptions) func(http.Handler) http.Handler {
+func ApiMiddleware(db client.TxnStore, txs *txnCache, nodeOpts *options.NodeOptions) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
 			ctx := req.Context()
@@ -61,7 +60,7 @@ func ApiMiddleware(db client.TxnStore, txs *sync.Map, nodeOpts *options.NodeOpti
 // TransactionMiddleware sets the transaction context for the current request.
 func TransactionMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		txs := mustGetContextSyncMap(req)
+		txs := mustGetContextTxnCache(req)
 
 		txValue := req.Header.Get(txHeaderName)
 		if txValue == "" {
@@ -75,13 +74,16 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 		}
 		tx, ok := txs.Load(id)
 		if !ok {
-			responseJSON(rw, httpStatusFromError(db.ErrTxnDiscarded), errorResponse{db.ErrTxnDiscarded})
+			err := ErrMissingOrExpiredTransaction
+			if strings.Contains(req.URL.Path, "/graphql") {
+				responseJSON(rw, httpStatusFromError(err), gqlErrorResponse{err})
+			} else {
+				responseJSON(rw, httpStatusFromError(err), errorResponse{err})
+			}
 			return
 		}
 		ctx := req.Context()
-		if val, ok := tx.(client.Txn); ok {
-			ctx = db.InitContext(ctx, val)
-		}
+		ctx = db.InitContext(ctx, tx)
 		next.ServeHTTP(rw, req.WithContext(ctx))
 	})
 }
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -76,9 +76,9 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 		if !ok {
 			err := client.ErrTransactionNotFound
 			if strings.Contains(req.URL.Path, "/graphql") {
-				responseJSON(rw, httpStatusFromError(err), gqlErrorResponse{err})
+				responseJSON(rw, http.StatusBadRequest, gqlErrorResponse{err})
 			} else {
-				responseJSON(rw, httpStatusFromError(err), errorResponse{err})
+				responseJSON(rw, http.StatusBadRequest, errorResponse{err})
 			}
 			return
 		}
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -72,7 +72,7 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 			responseJSON(rw, http.StatusBadRequest, errorResponse{ErrInvalidTransactionId})
 			return
 		}
-		tx, ok := txs.Load(id)
+		lease, ok := txs.Acquire(id)
 		if !ok {
 			err := client.ErrTransactionNotFound
 			if strings.Contains(req.URL.Path, "/graphql") {
@@ -82,8 +82,10 @@ func TransactionMiddleware(next http.Handler) http.Handler {
 			}
 			return
 		}
+		defer lease.Release()
+
 		ctx := req.Context()
-		ctx = db.InitContext(ctx, tx)
+		ctx = db.InitContext(ctx, lease.Value().txn)
 		next.ServeHTTP(rw, req.WithContext(ctx))
 	})
 }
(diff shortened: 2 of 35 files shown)


http/middleware.go
```diff
@@ -70,18 +69,21 @@ func TransactionMiddleware(next http.Handler) http.Handler {
...
 		}
 		tx, ok := txs.Load(id)
 		if !ok {
-			responseJSON(rw, http.StatusNotFound, errorResponse{client.ErrTransactionNotFound})
+			err := ErrMissingOrExpiredTransaction
+			if strings.Contains(req.URL.Path, "/graphql") {
+				responseJSON(rw, httpStatusFromError(err), gqlErrorResponse{err})
+			} else {
+				responseJSON(rw, httpStatusFromError(err), errorResponse{err})
+			}
```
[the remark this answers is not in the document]
jsimnz: I dont think this is correct. The GQL over HTTP spec requires that response data is of a certain format/structure. The requesting client doesnt know that the error in question happens before the GQL exeuction has started.

cli/tx_new.go
```diff
@@ -12,26 +12,65 @@ package cli
...
+			var err error
+			var ttl time.Duration
+			if txnTTL != "" {
+				ttl, err = parseTxnTTLFlag(txnTTL)
+				if err != nil {
+					return err
+				}
+				ttlClient, ok := cliClient.(txnTTLClient)
+				if !ok {
+					return fmt.Errorf("transaction ttl is not supported by this client")
```
AndrewSisley: todo: Please document why this is needed, or delete it.

internal/ttl/wheel.go
```diff
@@ -0,0 +1,235 @@
...
+		if w.index[e.key] == e {
+			delete(w.index, e.key)
+			expired = append(expired, e.key)
+		}
+	}
+	w.cur = (w.cur + 1) % w.slotCount
+	w.mu.Unlock()
+
+	for _, key := range expired {
+		if w.expireFunc != nil {
```
AndrewSisley: todo: This is never nil, please simplify the code.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
```
AndrewSisley, line 10: question: Do you have any use cases in mind that would want to use this besides transaction-expiry?

internal/ttl/wheel.go
```diff
@@ -0,0 +1,235 @@
...
+package ttl
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+// Wheel tracks comparable keys in a fixed-size timing wheel.
+type Wheel[K comparable] struct {
```
AndrewSisley, line 20: question: Why is this type public?

http/txn_cache.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if opts.TxnTTLBuckets != 0 {
+		cfg.buckets = opts.TxnTTLBuckets
+	}
+	return cfg
+}
+
+// Store caches txn until it has been idle for txnTTL. A txnTTL of zero uses the
+// cache's default TTL.
+func (c *txnCache) Store(txn client.Txn, txnTTL time.Duration) error {
```
AndrewSisley: todo: The fallback to default looks error prone, and something that should only ever be decided at compile time.

By toggling this at runtime you are forcing readers to consider more possiblities than they need to when maintaining related code.  Including the risk of an infinite loop where `c.LoadAndDelete` returns zero, only to feed it back into this function where the timer is reset.

Please split this function into two - one that stores the value provided, and one that stores using the default (no provided value).

e.g.
```
c.Store(txn)
c.StoreFor(txn, ttl)
```

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+package ttl
```
AndrewSisley, line 11: todo: By defining this internally, we have created a situation where our public embedded tooling is insufficient for users to create their own clients using it.

Please either make this packge internal to the http package (`http/internal/ttl`) - where it becomes an implementation detail of *our* http package, or make it public.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+
```
AndrewSisley, line 12: question:  What benefits do you see the proposed solution having over a simple wrapper around `client.Txn`? e.g. something very roughly like the following:
```
type Txn {
  inner client.Txn
  lifetimeRemaining time.Duration
  lastCall time.Time
}

func (t *Txn) AddCollection(...) ... {
  timeIdle := time.Now() - t.lastCall
  if  timeIdle < t.lifetimeRemaining {
    return NewErrExpired()
  }
  t.lifetimeRemaining = t.lifetimeRemaining - timeIdle
  t.inner.AddCollection(...)
  t.lastCall = time.Now()
}
... etc
```

internal/ttl/wheel.go
jsimnz: it certainly could be nil. What I think is a better todo is to put the nil check in the Constructor instead of the tick loop, but theres nothing really preventing this from being set (accidentally or otherwise) to nil. Itll panic if its nil when called, its a very cheap check at runtime, and its reasonably happy standard go to add a sanity nil check, even if being over cautious. 

We've had this exact convo on random PRs over time regarding nil checks specifically. We have a diff of opinion. The current implementation is safe and its easy read, I dont qualify this code as "complicated".

So unless you have an extreme pref here and I am disregarding this. lmk

http/txn_cache.go
jsimnz: `LoadAndStore` doesn't return the ttl, not quite sure im following that exact example regarding the infinite loop, but I'll update the API to make it clearer.

Would you prefer the API youve presented or make the argument optional so it can be used as `c.Store(txn)` and `c.Store(txn, ttl)`?

internal/ttl/cache.go
jsimnz: Im happy to make it public as a general utility. However, there is no *need* for external developer custom clients needs to implement this system. Its a saftey net for *our* HTTP API. Not having access to it doesn't limit them to implement their use case.

internal/ttl/cache.go
jsimnz: I would classify this approach as a read/write-repair. In that it will maintain the intended semantics (transaction timeout) at the next read or write (operation) path. 

If the DB could tolerate unbounded txn lifetimes, this would be fine since we could just wait an unbounded time till the next operation and handle the repair (expiry/discard/etc).

But the whole point is that we can't wait an unbounded amount as open transactions start to cripple the GC process in DBs or in DBs that only support a single transaction we need that guranteed safety net that will expire old transactions. 

The reason this is only implemented for the HTTP API is because networking being unreliable, there is an easy edge case to hit where both the client and the server did everything correctly, but the network transparently fails/drops connections at the wrong time, resulting in a transaction being in limbo.

So A) We need a system that gurantees an upperbound on transaction lifetimes B) Needs to not be a notable perf overhead of a naive loop over outstanding txns and just checking their remaining ttl.

ttl wheels are a very effecient and reasonably straightforward structure to support effecient timeout eviction, the Cache is just a nicer DX layer over top of it.

Happy to discuss this

internal/ttl/cache.go
jsimnz: Yeah, theres places in the p2p it can be useful for graceful timeout/expiry use cases off the top of my head

internal/ttl/wheel.go
jsimnz: I don't see a reason for it not to be, its the underlying mechanic that actually makes the cache work. The cache type is just candy coating.

jsimnz: the context of my answer based on some other comments about if this is useful in other areas of the code

internal/ttl/cache.go
AndrewSisley: Ah of course, it is more about the forced disposal of resources than restricting the transactions themselves.  Thanks for the explanation John :)

internal/ttl/wheel.go
AndrewSisley: > What I think is a better todo is to put the nil check in the Constructor instead

I'm happy with this, it handles the (mis)behaviour earlier, which will make it easier to understand.  It also means readers only have to worry about it once per instance, instead of once per tick (and the impossible mutation it suggests it is is protecting against there).

internal/ttl/cache.go
AndrewSisley: Would user driven actions such as `SyncDoc` not be better serviced by context timeouts (and us handling them)?

http/txn_cache.go
AndrewSisley: I'm talking about `LoadAndDelete` not `LoadAndStore` - `txHandler.Commit` passes the `txnTTL` returned from `LoadAndDelete` straight back into `Store` if the commit errors.  This feels odd for other reasons, and I'll add a comment about that, but it creates a loop where a ttl from one func is passed into another, where that second function can erroneously promote the ttl from zero to default in a situation where it really shouldn't being assigning defaults to existing txns at all.

 > Would you prefer the API youve presented or make the argument optional so it can be used as c.Store(txn) and c.Store(txn, ttl)?

I do prefer the overload route instead of optional args for this in Go, Golang optional args still exist at runtime and readers still need to read them when working with the function.  They also bring the potential complication of being a set instead of a single item.

internal/ttl/cache.go
AndrewSisley: > Its a saftey net for our HTTP API.

If that is how we wish to view it then it should really be in `http/internal/ttl` IMO.

internal/ttl/cache.go
jsimnz: I dont think this is an either/or situation between these two primitives

internal/ttl/wheel.go
AndrewSisley: > I don't see a reason for it not to be,  its the underlying mechanic that actually makes the cache work

This logic could be applied to everything that is internal in our codebase, with the outcome that everything is public-internal and the relationships between stuff become fairly spaghettified.

Making this private declares that it *is* an underlying mechanic, and is not something that needs to be, or should be, called from outside of the package.

internal/ttl/cache.go
AndrewSisley: What I am trying to understand is what we are looking to gain by preserving a lot of the existing complexities in the proposal.  For example, the generics - they only ever have one type, and at the moment represent a waste of complexity that readers need to deal with.

There are other examples too, such as `Cache.onExpire` only ever using the second of two input params.

http/txn_cache.go
jsimnz: apologies, I thought this comment was linked to the `ttl` package cache not the `http` package (also the `LoadAndStore` was a brain fart on my end when reading the `Load` and `Store` APIs, I did mean `LoadAndDelete` but its a moot point because I was thinking about the wrong context and therefore the wrong function signature).

To your actual point. I agree with the API change, but I guess Im still not seeing where the concrete issue is right now with the `txHandler.Commit` implementation and how the proposed API would make things better/safer. The `LoadAndDelete` returns the ttl for the item, theres no distinction in the return value if its the configured default or a custom ttl value, its just w.e the current timeout is. `LoadAndDelete` will never return `0`. So if `0` is provided to `Store` then someone must have intentionally set the argument to that (implying they want the default time). With the updated API, the `Store` call would become `StoreFor` and the returned ttl from `LoadAndDelete` would still be passed in verbatim.

I'm happy to clean this up, but im not necessary seeing the issue your describing right now.

AndrewSisley: > but im not necessary seeing the issue your describing right now.

My issue is with what it looks like can happen, and the risk when maintaining or adding consuming code.  I have not spotted a bug in the PR.

internal/ttl/cache.go
jsimnz: I don't see the value/benefit of it being scoped (structurally) to http. It is of course the only consumer of the package but that in itself is not a reason enough imo to only scope it to the http package. 

Given the two options, I'd rather just make the `ttl` package public at the root. I didnt necessarily want to commit the API to semver atm hence the `internal`

internal/ttl/wheel.go
jsimnz: > This logic could be applied to everything that is internal in our codebase

Fair, the point tho was about its downstream possible consumers (eg p2p as I hinted at) should have the freedom of  using the sugar layer (Cache) or the lower layer (Wheel). Its not a situation where we actively dont want / cant have users consuming the `Wheel` type AFAICT

internal/ttl/cache.go
jsimnz: Its a generic (design not type) utility. In these scenarios, I'd rather stub out the minimal but meaninful API such that it can be consumed whereever, without anyone needing to dive into the internals of a package they might not want to, if all they want todo is use the ttl system.

It also means that theres no larger refactors that needs to happen if theres a use case that requires a *slightly* different than the one we overly optimize for (http txn ttl). 

Making this generic (both design and type), and having an open API means it will suit a large number of use cases equally without refactors or leaking internals.

> that readers need to deal with

I think more accurately that reviewers have to deal with. The intent is that once its in, theres little maintenence needed (which wouldnt be the case if we scoped it soley to the http txn ttl). Its not an overly complicated use of generics (as far as generics go), and the API is easy to reason about. Short of bugs, its unlikely anyone will need to dive into the internals here.

cli/tx_new.go
```diff
@@ -12,26 +12,65 @@ package cli
...
-		RunE: func(cmd *cobra.Command, args []string) (err error) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			cliClient := mustGetContextCLIClient(cmd)
 
-			tx, err := cliClient.NewTxn(readOnly)
+			var tx client.Txn
+			var err error
+			var ttl time.Duration
+			if txnTTL != "" {
+				ttl, err = parseTxnTTLFlag(txnTTL)
```
fredcarle, line 41: question: Any chance we can test this branch?

cli/tx_new.go
```diff
@@ -12,26 +12,65 @@ package cli
...
+			}
 			if err != nil {
 				return err
 			}
 			return writeJSON(cmd, map[string]any{"id": tx.ID()})
 		},
 	}
 	cmd.Flags().BoolVar(&readOnly, "read-only", false, "Transaction is read only")
+	cmd.Flags().StringVar(&txnTTL, "ttl", "",
+		"Transaction idle TTL as a Go duration string, or seconds if no unit is provided")
```
fredcarle: suggestion: Might be more future proof to avoid mentioning `Go` in the flag description.

internal/ttl/wheel.go
```diff
@@ -0,0 +1,235 @@
...
+	cur        int64
+	slots      [][]*entry[K]
+	index      map[K]*entry[K]
+	maxTTL     int64
+	mu         sync.Mutex
+	running    bool
+	stop       chan struct{}
+	ctx        context.Context
+	expireFunc func(K)
+}
```
fredcarle, line 33: thought: The complexity of this type is non-negligable. It might work in it's current form but it will be difficult for future *us* to maintain. It seems like using `time.AfterFunc` could give us pretty much the same functionality and save about 300 prod lines of complexity.

fredcarle (approved):
I'll approve because the feature works and the change I would like to see can be done later without breaking anything I believe. That being said, I do think this is over-engineered and I fear that it will come back to bite us in the future.

internal/ttl/wheel.go
jsimnz: The complexity is very localized and results in basically free runtime impact.

Timing wheels are very well understood data structures designed for this exact use case. There's nothing novel about this implementation.

The naivety of a time.AfterFunc is frustrating in how much runtime cost in maintaining. A single is trivial, hundreds probably negligible, thousands or more constantly created and expiring... Less negligible 

Every manual txn gets one, every afterfunc is executed in its own go routine, that's a lot of churn/scheduling pressure if the DB is under load, which is the exact time you don't want an explosion of goroutines.

The time wheel has a single goroutine for everything.

I'm not suggesting this is a perfect solution, but it is one with few unknowns and strictly bounded perf impact. I'm usually not a fan of complex implementations of various features and components because usually they are bespoke to our use case, designed off the cuff. This is just a generic data structure.

jsimnz:
> That being said, I do think this is over-engineered and I fear that it will come back to bite us in the future.

Reposting from another comment:
> Timing wheels are very well understood data structures designed for this exact use case. There's nothing novel about this implementation.

That being said, theres some low hanging engineering that can be cleaned up, but not the core design imo

internal/ttl/cache.go
jsimnz: Based on the ongoing discussion on the core implementation, I am leaving the package as is within `internal/ttl` so we can change if needed/decided without breaking

http/txn_cache.go
jsimnz: Added the seperate functions `Store` and `StoreFor`. 0 value is no longer overloaded to mean "default", additionally 0 ttl is now a valid ttl (basically means expire immediately on next tick)

internal/ttl/wheel.go
jsimnz: After discussions, leaving the type public but the package internal, no harm for now.

cli/tx_new.go
jsimnz: documented

cli/tx_new.go
jsimnz: update the cli integration suite to test this branch

internal/ttl/wheel.go
AndrewSisley: > Timing wheels are very well understood data structures designed for this exact use case. There's nothing novel about this implementation.

This does not affect the maintainability of the code a whole lot.  It is still fiddly sensitive multi-threaded code with mutexes.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// Store stores val at key with an idle expiration of ttl, replacing any
+// existing value and discarding its lease state.
+func (c *Cache[K, V]) Store(key K, val V, ttl time.Duration) error {
+	if err := c.wheel.validTTL(ttl); err != nil {
+		return err
+	}
+
+	c.mu.Lock()
```
AndrewSisley, line 84: todo: Please document the purpose of this mutex.  At the moment there is a race condition here - the `Store` function will delete a key from the wheel outside of the lock, before adding it - a concurrent read may return nothing when it really should be depending on the timing.  This `Store` function is not atomic, although the mutexes on both types makes it look like it is supposed to be at a glance. 

Please also document this on the `Cache` type.  And the other functions that also contain this race condition (it appears to be inconsistent, with some relying on `defer`, and others racing).

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	c.wheel = wheel
+	c.wheel.Start()
+	return c, nil
+}
+
+// Store stores val at key with an idle expiration of ttl, replacing any
+// existing value and discarding its lease state.
+func (c *Cache[K, V]) Store(key K, val V, ttl time.Duration) error {
+	if err := c.wheel.validTTL(ttl); err != nil {
```
AndrewSisley: suggestion: Keeping `Wheel` public, so other imagined future consumers can consume it directly, whilst having the only current consumer of it rely on private functions does not make sense to me.  Please reconcile this, and make sure that if it is public (for now), that at least all the functions that we need to use to consume it are also public.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+
+// UpdateTTL resets the expiration time for key if key is still cached.
+// It returns false if key was no longer cached.
+func (c *Cache[K, V]) UpdateTTL(key K, ttl time.Duration) (bool, error) {
+	if err := c.wheel.validTTL(ttl); err != nil {
+		return false, err
+	}
+
+	c.mu.Lock()
+	item, ok := c.items[key]
```
AndrewSisley, line 182: todo: You do not appear to be updating the `ttl` on `c.items[key]` anywhere - the value is diverging from the value in the wheel.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+// Delete removes key from the cache without expiring its value.
+func (c *Cache[K, V]) Delete(key K) {
+	c.mu.Lock()
+	item, ok := c.items[key]
+	if ok {
+		delete(c.items, key)
+	}
+	c.mu.Unlock()
+
+	if ok {
```
AndrewSisley: todo: Because of the race conditions between deleting from `c.items` and the wheel, it is possible that `ok` will be false whilst the wheel still holds a key, leaking memory.

Please either remove this `if ok` check, or remove the race conditions.

internal/ttl/cache.go
```diff
@@ -0,0 +1,257 @@
...
+	err := c.wheel.Add(cacheWheelKey[K]{key: key, seq: item.seq}, item.ttl)
+	if err != nil {
+		delete(c.items, key)
+	}
+	value := item.value
+	c.mu.Unlock()
+
+	// Add only fails for a TTL the wheel cannot hold, which Store already
+	// validated, so this is unreachable in practice. Expire the value if it
+	// ever happens so the resource is not leaked while untracked.
```
AndrewSisley: suggestion: It would be more appropriate to document this on the creation of `err` or the `wheel.Add` call, `err` is consumed in another location further up, with more sensitive business logic in that dead case - that clause is undocumented and untested.

suggestion: `_ := c.wheel.Add`, then we don't need to worry about the untested `delete` in the dead code above.

internal/ttl/wheel.go
```diff
@@ -0,0 +1,236 @@
...
+	if onExpire == nil {
+		return nil, ErrInvalidExpireFunc
+	}
+	return &Wheel[K]{
+		tick:       tick,
+		slotCount:  slotCount,
+		maxTTL:     int64(tick) * slotCount,
+		slots:      make([][]*entry[K], slotCount),
+		index:      make(map[K]*entry[K]),
+		stop:       make(chan struct{}),
```
AndrewSisley: suggestion: A huge amount of complexity in this file stems from the pause-resume feature, which is completely unused.

Please delete this and either add a simple close function, or close via context.Cancel.

AndrewSisley (commented):
This PR introduces a lot of complex code that lacks purpose at the moment and I think there are decent odds of us ripping a lot of it out in the future.  It is however localised and so long as it works right now, the cost of this is limited.

Please resolve the remaining todos before merge.

internal/ttl/cache.go
jsimnz: it should be pretty evident, its providing protection on the `c.items` access/update as well as the sequence increment.

The wheel has its own locks, and the only thing we need to be worried about is the sequence number. The `Delete` and `Add` don't need to be under the lock.

internal/ttl/cache.go
jsimnz: ill update the docs, the `delete` should stay in as a fail safe

internal/ttl/cache.go
jsimnz: ill remove for now, theres some follow up work I'll do, but if the future is unknown it doesnt need to get into the weeds now

internal/ttl/wheel.go
jsimnz: Removed a bunch of the noted complexity of resumable with a relatively straight forward one time start/close

internal/ttl/cache.go
AndrewSisley: There is a race condition here, on what is supposed to be a thread-safe generic type.  I think it is sensible to document that, instead of expecting users to dig through the implementation to check for race conditions.

If you want to leave the race in.

internal/ttl/wheel.go
AndrewSisley: What is the reason for preserving the `Start` function as separate from `NewWheel` if it can only be called once?

internal/ttl/cache.go
AndrewSisley: I see my mistake - the race condition does really matter (I think), as the contents of the Wheel never really matters to consumers of the Cache.

suggestion: I think a line of documentation within this func would be handy, as it looks worse than it is to readers.

PR #5004, opened 2026-06-26
perf: Asynchronous secondary-index backfill

### LINKED ISSUE #4950: Sec. Indexes: Asynchronous backfill (non-blocking NewIndex with observable building status)
Follow-up to https://github.com/sourcenetwork/defradb/issues/296

## Context

https://github.com/sourcenetwork/defradb/issues/296 made index backfill (and deletion GC) run in batched transactions so an index can be created or dropped on a collection of any size. In that work the backfill is **synchronous**: `NewIndex` does not return until every existing document has been indexed (and for a caller-provided transaction it runs inside the caller's commit). The lifecycle state machinery (building/ready/failed/dropping records, watermarked batches, startup recovery) was deliberately built so this can later go asynchronous with minimal change.

## Proposal

Run the backfill in the background: `NewIndex` commits the definition plus the `building` state record and returns immediately, while a worker performs the batched backfill and flips the index to `ready` (or `failed`) when done. Deletion GC can run the same way.

This needs:

- A background worker that drains pending `building` (and `dropping`) records, reusing the existing batched backfill/GC and watermark-resume.
- The `building` status becoming externally observable and meaningful — `ListIndexes` already reports it, but with synchronous backfill no caller can ever see it.
- An error contract change: a failed background build can no longer be returned from `NewIndex`; it surfaces only as `failed` + reason on the state record (this path already exists for the caller-provided-transaction case).
- Progress reporting and cancellation of an in-flight build (relates to https://github.com/sourcenetwork/defradb/issues/4888).
- Concurrency control: do not run two builds for the same index; bound parallelism across indexes.

The startup-recovery worker introduced by #296 is essentially the same loop, so the asynchronous path and the crash-recovery path converge on one mechanism.

## Why it is worth doing

- `NewIndex` on a very large collection no longer blocks the caller for the full build.
- It unblocks several things that are hard to test today precisely because backfill is synchronous: the `building` window is never externally observable, a build cannot be interrupted at a chosen point from the test framework, and concurrent writes cannot be interleaved with a running build. With an asynchronous build these become deterministic, in-process integration tests (observe `building`, restart or panic mid-build, write concurrently during the build) instead of relying on unit tests that seed state or an out-of-process crash harness. Testability is a consequence, not the main driver.

## Acceptance criteria

- `NewIndex` returns promptly on a large collection, with the index reported as `building`, and the index becomes `ready` once the background build completes.
- A query during the build sees the index as not-yet-ready and falls back to a full scan (correct results, no use of the partial index).
- A background build failure surfaces as `failed` + reason via `ListIndexes`.
- An in-flight build can be cancelled, and an interrupted build is resumed on restart.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4950

## Description

Creating a secondary index on a large collection used to block the caller until every existing document had been indexed. That work now happens in the background: creating an index records it as building and returns right away, a background worker fills it in and flips it to ready, and the same applies to dropping an index. The single worker that does this is also the startup-recovery path, so a build interrupted by a crash resumes from where it left off on the next start. While an index is building it stays out of query planning, so queries fall back to a full scan and still return correct results, and live writes keep the building index current so the backfill and concurrent writes converge.

One behavioural change worth calling out: a build that fails can no longer be reported back from the create call, since that call has already returned. A failure now surfaces as a failed status with its reason through the index listing, matching how a failure already surfaced when an index was created inside a caller-provided transaction.

Building several indexes on the same collection at once needed care. Each backfill now opens only what it needs for its own index rather than reading every sibling index's state, so concurrent builds on a collection no longer collide on each other's bookkeeping; and a build interrupted by a transient conflict retries itself rather than being left part-built. Dropping a previously-failed index also clears its stale failure record now.

## How has this been tested?

Unit and integration tests cover the build, drop, migration-rebuild, and crash-recovery paths, including resuming a build from a saved point after a restart and rebuilding an index when a migration runs while the first build is still going.

On top of that I ran a randomized stress harness against a real on-disk node. It creates many collections, thousands of documents, and hundreds of index create/delete operations, runs migrations that force index rebuilds, and crashes the node at random points (both graceful close/reopen and hard kill -9) so builds are interrupted mid-flight. After every step it checks the stored data directly: each live document has exactly one index entry in the current epoch, no leftover entries remain from old epochs, no stale build records are left behind, and every indexed query returns the same result as a full scan. It held up with no failures across many runs. I confirmed these checks actually catch problems by temporarily breaking the epoch cleanup and seeing them fail.

I also ran an upgrade check to confirm existing data still works. I seeded a data directory with a node built from the target branch (multiple collections with a relation, documents, single-field, composite, and unique indexes, plus a couple of schema migrations), then opened that same directory with this branch. Every document was still present, every indexed query returned the same results, the index listing was unchanged, and the unique constraint was still enforced. Creating a new index backfilled the pre-existing documents, and dropping an index, adding, updating, and deleting documents all kept the indexes correct. No migration or reindex was triggered on open, and the state survived a restart.

Tested on macOS with the on-disk Badger store.

## Tasks

- [x] I made sure the pull request title adheres to the conventional commit style (the subset used in the project: https://github.com/sourcenetwork/defradb/blob/develop/CONTRIBUTING.md#commit-messages).


tests/action/wait_for_index.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	nodeIDs, _ := getNodesWithIDs(a.NodeID, a.s.Nodes)
+	for index := range nodeIDs {
+		node := a.s.Nodes[index]
+		if node.Closed {
+			continue
+		}
+		collections := MustGetCanonicallyOrderedCollections(a.s, node, immutable.None[client.Txn]())
+		collection := collections[a.CollectionID]
```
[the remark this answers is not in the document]
islamaliev: Good catch. getNodesWithIDs returns the node IDs, but the loop indexed a.s.Nodes by the slice position, so a selected NodeID would wait on the wrong node. Switched to ranging over the returned IDs and indexing by the actual node ID, matching the NewIndex action.

internal/db/index_worker.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+
+	// A rebuild leaves superseded epochs with no record, and may crash after its build finishes
+	// but before collecting them. Sweep every index; it is a no-op for an index with only its live
+	// epoch, and the delete range is bounded strictly below the live epoch so it never touches an
+	// in-flight build.
+	if err := w.db.recoverStaleEpochs(ctx); err != nil {
+		log.ErrorE("Failed to collect stale index epochs during drain", err)
+	}
```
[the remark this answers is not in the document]
islamaliev: Fixed. The stale-epoch sweep now re-drives on a transaction conflict via scheduleRetry, same as the build and drop paths. Without an action record nothing else would re-dispatch it, so a single conflict could otherwise leave stale epochs uncollected until an unrelated wake.

internal/db/index_worker_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Give any erroneously-dispatched second build a chance to enter the gate.
+	select {
+	case <-entered:
+		t.Fatal("a second build started for the same index despite the in-flight guard")
+	case <-time.After(50 * time.Millisecond):
+	}
+
+	close(release)
+	db.indexBuildWorker.builds.Wait()
```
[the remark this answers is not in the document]
islamaliev: Fixed. The first build is now awaited with a one-second timeout instead of an unbounded receive, and the release channel is closed via t.Cleanup (guarded by sync.Once) so a failing assertion can never strand the build goroutine in the gate.

internal/db/collection.go
```diff
@@ -99,6 +94,26 @@ func (db *DB) newCollection(
 	return col, nil
 }
 
+// newBareCollection builds a collection without loading index build-states or the per-index write
+// instances. For callers that maintain one index explicitly (the backfill batch) and use neither
```
AndrewSisley, line 117: question: Why would we want 'normal' collection instances to load all the index build states and per-index write instances?

internal/db/collection_index.go
```diff
@@ -167,14 +166,17 @@ func (c *collection) deleteIndexedDocWithID(
...
+// planning, so queries full-scan and still return correct results.
 //
-// If the backfill fails, the index definition remains in place with a failed status.
-// It is not maintained by subsequent writes. Use DeleteIndex to remove it before recreating.
+// A build failure is not returned here, since the build runs after this call returns. It surfaces
+// as a "failed" status and reason via ListIndexes; callers that need the outcome poll for it. A
+// failed index is not maintained by writes; use DeleteIndex to remove it before recreating.
+//
+// Live writes maintain the building index, so the backfill and concurrent writes converge on the
+// same final entries.
```
AndrewSisley: todo: It looks like this documentation should be on the public `client` interfaces where it is visible to users, not here, where it is only visible to us

internal/db/db.go
```diff
@@ -224,15 +227,30 @@ func newDB(
...
+		})
+	}
 
 	return db, nil
 }
 
+// suppressIndexWorkerRun makes newDB construct the worker but not start its run loop, so a test can
+// drive draining via drainSync without the background loop racing its hand-seeded records. Test-only;
+// the db package runs its tests serially, so toggling it around one newDB call is safe.
+var suppressIndexWorkerRun bool
```
AndrewSisley, line 253: question: Why do we want this to be test only, and not a configurable option?

internal/db/index_worker.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+}
+
+// drainSync runs a drain and blocks until every dispatched build or drop has finished. It is
+// test-only: production never blocks on a build (see run). drainMu is held across both the dispatch
+// and the wait so no concurrent drain increments builds meanwhile. The in-flight guard is shared
+// with the background loop, so an index a prior pass is already building is not started twice;
+// drainSync waits for that work too, since builds is shared.
+func (w *indexBuildWorker) drainSync(ctx context.Context) {
```
AndrewSisley: question: Why are we introducing a test-only function here?

internal/db/index_worker.go
```diff
@@ -0,0 +1,270 @@
...
+	key keys.IndexStateKey,
+	action client.Action,
+	work func(context.Context) error,
+) {
+	guardKey := inFlightKey(key)
+	if _, loaded := w.inFlight.LoadOrStore(guardKey, struct{}{}); loaded {
+		return // a build or drop for this index is already in flight
+	}
+
+	w.builds.Go(func() {
```
AndrewSisley: question:  It looks like we have:
1. A hard coded limit on the concurrency (4)
2. An unbounded number of Go routines spawning

Why are we not fixing the number of routines?

internal/db/index_worker.go
```diff
@@ -0,0 +1,270 @@
...
+	drainMu sync.Mutex
+
+	// builds counts dispatched build goroutines so a drain can be awaited. Incremented under drainMu.
+	builds sync.WaitGroup
+}
+
+// newIndexBuildWorker constructs a worker subscribed to action events. The subscription is closed
+// by db.events.Close.
+func (db *DB) newIndexBuildWorker() (*indexBuildWorker, error) {
+	sub, err := db.events.Subscribe(event.ActionExecutionName)
```
AndrewSisley, line 93: question: I do not understand this solution - why are we not simply spawning `indexBuildConcurrency` number of routines with each reading from this single `sub` chan?

internal/db/collection.go
islamaliev: normal collections use both: QueryableIndexes reads the build states to keep a building or failed index out of query planning, and the write path iterates c.indexes to maintain indexes on writes. The backfill is the exception, it maintains one index itself and uses neither, so newBareCollection skips the load, which also avoids the read-set conflict between two concurrent backfills on the same collection.

We could flip the default the other way, construct bare, and load index state only where a caller plans queries or maintains documents. But that means auditing each existing call site for which ones need it, so I'd rather keep that as a separate change. Here I kept load-by-default to preserve the existing behavior at those sites.

internal/db/collection_index.go
islamaliev: updated docs

internal/db/db.go
islamaliev: there is no reason to suppress it in production. Without this indexes just don't work. 

if you are asking about the var itself, I can make it an argument of newDB, but that'd still be test-only. Let me know what you preferences are.

Hm... maybe you are asking why we need it. It only exists because the background worker races the db-package unit tests, which seed index records by hand and drive drainSync deterministically.

internal/db/index_worker.go
islamaliev: answered in the comment above https://github.com/sourcenetwork/defradb/pull/5004#discussion_r3484066914

internal/db/collection.go
AndrewSisley: > and the write path iterates c.indexes to maintain indexes on writes.

question: We need to maintain all index versions on write atm?

If yes, I feel like we can improve that in the future, but it doesn't need to be done for v1.

Thanks for the explanation Islam :)

internal/db/index_worker.go
islamaliev: You're reading it right. The 4 only limits how many builds run at once. Each pending index still gets its own goroutine, and the extra ones just sit waiting on the semaphore. It's capped at one goroutine per index with pending work (the in-flight guard stops duplicates), so not truly unbounded but on a big recovery it can be a lot of waiting goroutines.

It's built this way because drain can't block: the run loop has to keep reading the event bus, and a stuck subscriber blocks every publisher. So drain hands off the work and returns instead of waiting for a free slot. 

I actually though about a fixed pool of N goroutines pulling from a queue, but that seemed to have a bit more complex setup: needs some care to keep the per-index dedup and other stuff. 

I can do that here or file it as a follow-up. Which do you prefer?

internal/db/collection.go
islamaliev: I don't know about "versions" but we need to maintain at least active indexes

internal/db/db.go
AndrewSisley: > Without this indexes just don't work.

Oh true, sorry - my brain bounced back to the older solution and thought this would just disable rebuild on restart.  Definitely prefer this as a private var then over an option :)

I'd like to remove this long term, but doubt it is worth re-working what ever would be needed to get the tests working without it for v1. Thanks for the explanation Islam.

internal/db/index_worker.go
AndrewSisley: Ah okay, I'll suppress my allergy for now then, I'm assuming it would be a lot of hassle to remove this based on your reply.

internal/db/index_worker.go
islamaliev: The event on sub is just a hint that something changed, not the actual work to do. When the worker wakes up it ignores what's in the event and instead scans the stored index records to figure out what needs building or dropping. That's why N goroutines reading sub directly wouldn't work: after a crash there are no events at all, so they'd never wake up even though there's plenty to recover; one wake can also mean many records to process, while many events for the same index need to collapse into a single build; and a rolled-back transaction still fires an event but leaves no record, so acting on the event itself would try to build something that isn't there.

So the events just decide when to look, and the record scan decides what to do. A fixed pool of indexBuildConcurrency goroutines is still doable. They'd read from an internal queue the scan fills, not from sub directly.

AndrewSisley: Ah I see what you mean.  I feel like it should still be possible to replace all the mutex/semaphore stuff etc stuff with a simple(r) channel feeding a pool of straighforward routines though.  I'll review it as-is then though (maybe Monday).

It might be that that is roughly what you've got (minus the pooling), and I'm just thrown by some of the namings plus the complexities that reality brings :)

Thanks for the explanation.

internal/db/index_worker.go
AndrewSisley: > a stuck subscriber blocks every publisher

This should not be true, is a long time ago now, but I think there was a fair bit of effort put into ensuring that this should never be the case. 

> I can do that here or file it as a follow-up. Which do you prefer?

Will get back to you in a bit on that, Go routines are reasonably large (2 KB initially) and come with other costs so they make me a bit nervous by default.

AndrewSisley: > This should not be true, is a long time ago now, but I think there was a fair bit of effort put into ensuring that this should never be the case.

Looks like it is still true (scanning the implementation). Drain blocking should not cause problems for other event subscribers/publishers.

internal/db/collection.go
AndrewSisley: Ah okay, I think I misunderstood:

> and the write path iterates c.indexes to maintain indexes on writes

This refers only to indexes that have finished building?  Not to indexes that have started building?

Meaning if `Users` has a single index on `name`, it was previously built, and a new build is in-progress, when writing to the collection only the built index will get the immediate update? 

question: If so, is there not a risk of the in-progress index missing new writes, causing it to be incorrect by the time building completes and it is swapped in for active use?

internal/db/index_worker.go
AndrewSisley: suggestion: I've reviewed again, and think this function should be moved to the tests.  It itself is private, and it does nothing that cant be very easily done in an identical function in the `_test.go` file.

internal/db/index_worker.go
```diff
@@ -0,0 +1,270 @@
...
+		case <-ctx.Done():
+			return
+		}
+		defer func() { <-w.sem }()
+
+		if ctx.Err() != nil {
+			return
+		}
+		err := work(ctx)
+		if err == nil {
```
AndrewSisley: suggestion: Inverting the check as you've done here may reduce indentation, and calculated complexity, but it is counter intuitive and inverts the more expected
```
if isAbnormal {
  doAbnormal()
}
doNormal()
```
flow.

internal/db/index_worker.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// wake is a size-one coalesced wake-up channel for the initial drain and non-bus callers.
+	wake chan struct{}
+
+	// inFlight holds the keys of indexes whose build or drop is running, so a drain does not start
+	// a second one for the same index. Keyed by inFlightKey.
+	inFlight sync.Map
+
+	// sem bounds concurrent builds across indexes.
+	sem chan struct{}
```
AndrewSisley: suggestion: This variable name, and the comment attached to it are not as descriptive as I feel that they could be, I suggest rewording them a little bit.

client/collection.go
```diff
@@ -116,20 +116,32 @@ type Collection interface {
 	// only contain letters, numbers, and underscores.
 	// If the name of the index is not provided, it will be generated.
 	// WARNING: This method can not make a new index for a collection that has a policy.
+	//
+	// The index is built in the background: this returns once the index is recorded, before existing
+	// documents are indexed. The index starts in a "building" state and becomes "ready" once the
+	// backfill completes, or "failed" if it cannot. While building, queries that would use it instead
+	// full-scan and still return correct results. A build failure is therefore not returned here;
+	// check the index's status via ListIndexes to learn the outcome. A failed index must be deleted
+	// before it can be recreated.
```
AndrewSisley, line 125: todo: This is a semver breaking change and my understanding is that we have not agreed on how we want to handle their merge/branching.  Please wait on this before merging the PR.

AndrewSisley (approved):
LGTM, thanks Islam.

It is very well documented, but I still *feel* like the internals of the worker are more complex than they need to be.  As mentioned, this might just be my simplified imagination colliding with the messy real world. It is also tested and isolated, so there isn't much to lose IMO vs any imagined 'perfect' solution.

Sorry about the delay in finishing my review.

internal/db/collection.go
islamaliev: > is there not a risk of the in-progress index missing new writes, causing it to be incorrect by the time building completes and it is swapped in for active use?

all doc creations/updates also write to in-progress index that is being built. So it's always up-to-date

internal/db/index_worker.go
islamaliev: done

internal/db/index_worker.go
islamaliev: done

internal/db/index_worker.go
islamaliev: this field is gone

internal/db/index_worker.go
islamaliev: The drain now only enqueues and returns; builds run in the pool, so drain never blocks the bus reader. Reworded the comment that overstated this.

PR #5006, opened 2026-06-26
refactor: Inconsistent cli flags

### LINKED ISSUE #5002: Uniform arguments for CLI `collection truncate` and `collection delete`
`truncate` takes the collection flags, while `delete` doesn't want them and instead wants one positional argument. We should uniform them, and also allow `delete` to take multiple collections at once, otherwise dangling relations to deleted docs might arise.
```
stef@carbonino:~$ defradb client collection truncate --name J
stef@carbonino:~$ defradb client collection delete J
stef@carbonino:~$ defradb client collection delete --collection-id bafyreiecfhw4xwmsua3zutvfsgx46ccfs4bs7g4cptapapkgezgsaf3i5u
Error: accepts 1 arg(s), received 0
```

### LINKED ISSUE #5005: Move redundant global CLI flags
There are a number of global CLI flags that should be moved to more specific subcommands so that they do not pollute the CLI documentation.

nasdf:
## Relevant issue(s)

Resolves #5002
Resolves #5005

## Description

This PR fixes an inconsistency with the collection delete flags. It also moves a number of global CLI flags onto more specific sub-commands so we do not pollute the documentation with unused flags.

## Tasks

## How has this been tested?

Existing unit and integration tests.

- Linux (Pop-OS)

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -14,6 +14,8 @@ import (
 	"context"
 
 	"github.com/spf13/cobra"
+
+	"github.com/sourcenetwork/defradb/cli/config"
 )
 
 func MakeClientCommand(ctx context.Context) *cobra.Command {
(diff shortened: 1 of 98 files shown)


AndrewSisley (approved):
LGTM, thanks Keenan!

AndrewSisley (changes requested):
Looks like my earlier review wasn't very good, as the CI has caught some issues - I'm requesting changes as I think the fix should be reviewed, but if someone else beats me to a re-review please merge without waiting on me (I'll try take a look and see what I missed post-merge in that case)

ChrisBQu (changes requested):
todo: I believe the reason so many tests are failing, are that you haven't yet updated the test wrapper for the CLI client(?) For example, one failing CI test is as follows:

```
            Error:          Received unexpected error:
                            unknown command "Users" for "defradb client collection delete"
            Test:           TestNAC_GatesDeleteCollection_AuthorizedIdentity_AllowAccess
```

But look at the implementation of this function inside `tests/clients/cli/wrapper.go`:

```
func (w *Wrapper) DeleteCollection(
	ctx context.Context,
	names []string,
	opts ...options.Enumerable[options.DeleteCollectionOptions],
) error {
	args := []string{"client", "collection", "delete"}

	opt := utils.NewOptions(opts...)
	if opt.ActiveOnly {
		args = append(args, "--active-only")
	}
	args = append(args, strings.Join(names, ","))
	args = appendIdentityArg(args, opt.GetIdentity())

	_, err := w.cmd.execute(ctx, args)
	return err
}
```

This is still appending the collection names as a positional argument, and so the tests are failing.

ChrisBQu:
suggestion: This one is not a blocker for me. But Coderabbit brought up an interesting-enough point that you might wish to address. And here is what I have in mind beyond that: we don't seem to have a test that illustrates behavior of passing in a blank collection name to the CLI delete function. I think if we expect that to deterministically fail though, we should have a test of that expected failure.

AndrewSisley (approved):
LGTM, thanks Keenan!

ChrisBQu:
LGTM. Thank you for fixing the tests, I'm changing to approve.

ChrisBQu (approved):

PR #5008, opened 2026-06-26
refactor: Remove badger namespace from CLI path config

### LINKED ISSUE #5007: Remove "badger" namespace from config datastore path
We dont need to include "badger" in the `datastore.badger.path` within the CLI config, its misleading at best. The options package is just `Path`, so we need to make it consistent

jsimnz:
## Relevant issue(s)

Resolves #5007 

## Description

Removed `badger` from the CLI config datastore path, so `datastore.badger.path` becomes `datastore.path`.

The added namespace barely made sense when we only supported badger, and while we cant use LevelDB (or other CoreKV backends) for CLI other than in-memory (also badger atm), theres no reason to require the `badger` namespace when any file persistence kv store will use the same API.

This is why the `options` package uses `Path` on the `NodeStoreOptions`.

## How has this been tested?

test suite, manual CLI build

- Linux NixOS

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -45,7 +45,7 @@ const (
 
 // ConfigPaths are config keys that will be made relative to the rootdir
 var ConfigPaths = []string{
-	"datastore.badger.path",
+	"datastore.path",
 	"api.pubkeypath",
 	"api.privkeypath",
 	"keyring.path",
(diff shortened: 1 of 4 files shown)


AndrewSisley (approved):
Nice, thanks for this John :)

PR #5010, opened 2026-06-29
docs: Better categorizing of HTTP ACP endpoints

### LINKED ISSUE #5009: Improve HTTP API docs groupings for ACP endpoints
The ACP endpoints in the HTTP API are split under various entries, many of which contain only one endpoint:

Image

We should group them under two categories: `document access control` and `node access control` to facilitate discovery and browsing.

PS. I looked into whether it's possible to nest categories, so that we could have
```
document access control
> policy
>> create
> relation
>> create
>> delete
```
The [`parent` field](https://learn.openapis.org/specification/tags.html#parent-field-for-hierarchical-organization) allows to achieve that, and is available in OpenAPI 3.2, but our dependency `kin-openapi` has still [not implemented full 3.1 support](https://github.com/getkin/kin-openapi/issues/230), let alone 3.2.

ribiza:
## Relevant issue(s)

Resolves #5009 

## Description

- Group all DAC endpoints into a DAC category; similar for NAC. We refer to those systems as "Document Access Control" and "Node Access Control" (and not "Access Control Policy for documents"), so it's beneficial to users if we conform to that.
- The IDs for the endpoints now have a common prefix (`document relationship add` vs the old `add document relationship`) so that visual parsing is aided: it's quicker to recognize a common prefix than a common suffix. This does NOT affect the endpoint locations; only the URLs in the docs and sidebar labels.

## How has this been tested?

Screenshot from 2026-06-29 08-31-54

### DIFF
--- a/docs/website/references/http/openapi.json
+++ b/docs/website/references/http/openapi.json
@@ -674,7 +674,7 @@
         "/acp/document/policy": {
             "post": {
                 "description": "Add a policy using document acp system",
-                "operationId": "add dac policy",
+                "operationId": "dac policy add",
                 "requestBody": {
                     "content": {
                         "text/plain": {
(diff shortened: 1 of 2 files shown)


AndrewSisley (approved):
LGTM, thanks Stefano!

Issue #4383, opened 2026-01-21
Granular NAC Permissions should be properly tested


ChrisBQu:
Assuming this PR (https://github.com/sourcenetwork/defradb/pull/4373) gets merged in, consider the following test:

```
func TestNAC_GatesCollectionPatch_NoIdentity_NotAuthorizedError(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			// Starting with NAC, so only authorized user(s) can perform operations from here on out.
			testUtils.Close{},
			testUtils.Start{
				Identity:  testUtils.ClientIdentity(1),
				EnableNAC: true,
			},
			// Note: Doing setup steps after starting with nac enabled, otherwise the in-memory tests
			// will lose setup state when the restart happens (i.e. the restart that started nac).
			&action.AddSchema{
				Identity: testUtils.ClientIdentity(1),
				Schema: `
					type Users {}
				`,
			},

			// We haven't authorized non-identities. So, this should error.
			testUtils.PatchCollection{
				Identity: testUtils.NoIdentity(),
				Patch: `
					[
						{ "op": "add", "path": "/Users/Fields/-", "value": {"Name": "name", "Kind": "String"} }
					]
				`,
				ExpectedError: testUtils.FormatExpectedErrorWithPermission("collection-patch"),
			},
		},
	}

	testUtils.ExecuteTestCase(t, test)
}
```

This will work as expected on the HTTP, C, and Go clients. However, it will fail on the CLI client. On that client, the error will actually return `not authorized to perform operation. Permission: collection-get`

Issue #4090, opened 2025-10-23
Handle Gating Node Identity Special Case


shahzadlone:
TLDR: Gating node identity has some edge / special cases to handle, why we want to give the option to gate it, by default we want to internally make a relationship for everyone to have access to it (even empty identity requests).

This should preferably be done after granular NAC permissions is in.

Partial work done on my fork in commit:  `281a8d54 'PR(MAIN): Gate & Test - Get Node Identity'`

Summary:
1) Problem restarting node (`defradb start`) without identity if NAC was previously enabled, because `Node.startP2P` is called that calls `n.DB.GetNodeIdentity(ctx)`, which would require an identity.
Solution: Team agrees that we can require the NAC access check at `start` to make sure only the users that have access to start the node can do so. `i.e.` If NAC is enabled, you need to provide identity upon restart (unless `*` public permission given of course).

2) Getting Node Identity is gated but not getting node identity token, should both these be gate?
Solution: Team prefers that while we should gate both these, we should make the gating of them opt-in-able (not gated by default), so internally when a user enables NAC the first time, an internal `*` relationship be created to by default allow all users to access both of these operations. However a user is free later then to remove/delete the `*` relation to gate it with more finer granularity.

Context: https://discord.com/channels/1374835078781468722/1403123516806529025/1403123522800455774

PR #3218, opened 2024-11-07
feat: Make DAC write permission more granular

shahzadlone:
## Relevant issue(s)

Resolves #2905 

## Description
Split `write` into `update` and `delete` permission.

## How has this been tested?
- [ ] todo: Update previous test once migration of policyID is done
- [ ] todo: Split previous write tests

PR #5017, opened 2026-06-30
fix: Prevent NAC re-enable from leaking permissions

### LINKED ISSUE #5003: CLI command `acp node re-enable` gives away NAC status even if identity has no permission for `acp node status`
The no-identity lacks permission to see NAC status:
```
$ defradb client acp node status
Error: not authorized to perform operation. Permission: get-nac-status
```
However, the error message of `acp node re-enable` gives the status away:
```
$ defradb client acp node re-enable  
Error: node acp is already enabled
```

ChrisBQu:
## Relevant issue(s)

Resolves #5003 

## Description

The following two checks are performed when attempting to re-enable NAC permissions:

```
	if err := db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeReEnableNACPerm); err != nil {
		return err
	}
```

```
if err := db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeReEnableNACPerm); err != nil {
   return err
}
```

However, they were happening in the wrong order. In the case of a user lacking permissions to check the status, they could still see the status if NAC was already enabled. This puts the checks in the right order, and updates the tests related to this.

## Tasks

## How has this been tested?

- WSL


shahzadlone:
This was discussed while implementing, the consensus was to treat the status command as a special command in order to be able to always have the `status` command give useful information.

ChrisBQu:
> This was discussed while implementing, the consensus was to treat the status command as a special command in order to be able to always have the `status` command give useful information.

With this in mind, can you elaborate a little bit on why the `get-nac-status` permission exists? It's a little unclear to me what treating it as a special command implies and entails.

If this design is intended, I think the issue can be closed. However, I think it's possible that it might come up again in the future. I would propose, if we are closing the issue, that I undo the work done in this PR, _but_ that comments are added to these functions explaining that it is explicitly intended.

AndrewSisley:
> if we are closing the issue, that I undo the work done in this PR, but that comments are added to these functions explaining that it is explicitly intended.

:+1: , and detailing why it is intended.

shahzadlone:
`get-nac-status` and other permissions are all tied to the `admin` relations, during gating we made permissions for all operations we have, they were left to be consolidated/decided on how we want to collapse/group them once #4383 is implemented.

Happy to have it documented and the detailing to why it is/was intended, from what I remember it seemed to be more of a consensus thing due to having the commend having the same nature as `git status`, and also the fact that if NAC is temp-off or not configured, then anyone can run the command and it will always show it's temp-off/unconfigured, if it's enabled only the admin can run the command and it will always show enabled and the users that don't have access still know the status of it is on or not by that unauthorized to access error anyway.

One other interesting difference the NAC status command has that comes to mind is the fact that it is the only NAC specific operation that can be performed even when NAC is not configured (all others need it setup), hence the following code we have:   
```go
func (db *DB) checkNodeAccess(
	ctx context.Context,
	ident immutable.Option[acpIdentity.Identity],
	permissionNeeded acpTypes.NodeResourcePermission,
) error {
	// For nac specific operations, the node acp setup must be configured.
	if permissionNeeded.IsForNACOperation() &&
		db.nodeACP.NodeACPDesc.Status == client.NACNotConfigured &&
		permissionNeeded != acpTypes.NodeGetNACStatusPerm {
		return ErrNACIsNotConfigured
	}
```

ribiza:
I'm not sure I follow.

> to treat the status command as a special command in order to be able to always have the status command give useful information.

The issue I raised doesn't question this (even if I don't think that the implementation reflect the quoted intention: if `status` should always return useful information, it should return it to anybody - so `status` should be excluded from any permission check).

The issue is that `status` will error if I'm not an admin, but `re-enable` will expose the same information that I would get through `status`, even if I'm not an admin. So I'm effectively bypassing permissions.

Obviously we could also fix it by removing the permission check on `status`, if that's our wish.

ribiza:
> if NAC is temp-off or not configured, then anyone can run the command and it will always show it's temp-off/unconfigured, if it's enabled ... the users that don't have access still know the status of it is on or not by that unauthorized to access error anyway.

This to me is supporting ground for removing the permission check on `status`. If the status can be inferred, why hide it through the command that is intended to show it.

fredcarle:
I agree with @ribiza here. Lets remove perm checks on this command.

shahzadlone:
> The issue is that `status` will error if I'm not an admin, but `re-enable` will expose the same information that I would get through `status`, even if I'm not an admin. So I'm effectively bypassing permissions.

Thanks for this clarification, I misunderstood the problem this PR was fixing, I think I got confused by the PR desc. wording "lacking permissions to check the status", I just re-reviewed the changes and this is a sensible fix to have.

That being said, I am not in favor of removing of any gates from any operations / i.e. removing any permissions at this stage. In the v0 implementation of NAC all permissions are somewhat redundant, they were left to be thoroughly planned out for when the #3218 is being implemented, one of the reason I don't want to delete it now is depending on how we solve certain cases we might even bring it back like it is, and i don't want to miss or remove gates from any ops that can be missed later. One such case is when we do https://github.com/sourcenetwork/defradb/issues/4090, where there are certain cases we want to by default give the * permission to access certain OPs. The get status can adopt a similar model where we give * permissions to that operation by default, the benefit of this is that just like the Node Identity special case, the user can have the option to delete that relationship and make those operations be gated, if they like.

shahzadlone (approved):
LGTM

tests/integration/acp/nac/re_enable_nac_test.go
```diff
@@ -163,7 +163,7 @@ func TestNAC_ReEnableWithWrongIdentityWhenAlreadyEnabled_Error(t *testing.T) {
 
 			testUtils.ReEnableNAC{ // NAC is already enabled before.
```
shahzadlone: nitpick: improve the comment something like:
```suggestion
			testUtils.ReEnableNAC{ // NAC is already enabled before, but not authorized to re-enable, so that gate hits first.
```

tests/integration/acp/nac/re_enable_nac_test.go
```diff
@@ -139,7 +139,7 @@ func TestNAC_ReEnableWithNoIdentityWhenAlreadyEnabled_Error(t *testing.T) {
 			},
 
 			testUtils.ReEnableNAC{ // NAC is already enabled before.
```
shahzadlone: nitpick: improve the comment something like:
```suggestion
			testUtils.ReEnableNAC{ // NAC is already enabled before, but not authorized to re-enable, so that gate hits first.
```

PR #5019, opened 2026-06-30
docs(i): Signed-docs exclusion reason on counter tests

### LINKED ISSUE #4742: Replicator double-applies initial counter delta when signing is enabled

islamaliev:
## Relevant issue(s)

Closes https://github.com/sourcenetwork/defradb/issues/4742

## Description

The peer counter tests excluded the signed-docs multiplier with a comment claiming signed counter deltas are double-applied across peers. That framing was misleading: with signing, each node's genesis block is signer-specific, so the same document created independently on every node produces a distinct DocID per node — the nodes never converge on a shared document, which is what these tests assume. Updated the exclusion comments on the counter and branchable commit tests to state that reason and dropped the stale issue reference.

## How has this been tested?

Comment-only change to existing tests; no behaviour change.


shahzadlone (approved):
LGTM

todo: Seems like a documentation only PR, why not instead use something like `docs(i): Clarify signed-docs exclusion reason on counter tests`

PR #5021, opened 2026-07-01
fix: Timeout on concurrent doc-sync for the same doc

### LINKED ISSUE #5020: bug(p2p): concurrent SyncDocuments for the same doc time out for all but one caller
If you call `SyncDocuments` more than once at the same time for the same doc ID, only one call succeeds. The others block until the timeout and return `timeout while syncing doc` — even though the document actually did sync.

## Repro

1. Two connected nodes; node A has a doc, node B doesn't.
2. On node B, fire several `SyncDocuments` calls for that same doc ID concurrently.
3. One returns OK; the rest return `timeout while syncing doc` after the deadline.

## Impact

This is easy to hit when a client fans out sync requests and the same doc ID ends up in more than one in-flight call. The sync itself works (the successful call pulls the doc), but the extra callers wrongly report a timeout, so from the caller's side it looks like a failure. It also wastes the full timeout window on each of those calls.

## Hints

- Concurrent requests for the same doc IDs appear to collide with each other — the requests carry nothing unique to tell them apart.
- The KMS sync path handles concurrent requests fine; the doc-sync path doesn't, which might be a useful comparison point.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5020

## Description

Calling SyncDocuments more than once at the same time for the same document made all but one of the calls block until their deadline and return "timeout while syncing doc", even though the document actually synced. The concurrent requests were indistinguishable from each other, so they shared a single in-flight response slot: one call received the peer's reply and the rest were left waiting. Each request now carries a random nonce so concurrent syncs for the same document stay independent, matching what the encryption-key sync path already does.

## How has this been tested?

Added an integration test that fires several concurrent SyncDocuments calls for the same document and asserts they all succeed (it fails without the fix). Existing P2P sync tests continue to pass.


internal/db/p2p/sync_doc.go
```diff
@@ -40,6 +41,11 @@ const docSyncTopic = "doc-sync"
 // docSyncRequest represents a request to synchronize specific documents.
 type docSyncRequest struct {
 	DocIDs []string `json:"docIDs"`
+	// RequestID is a random nonce that makes each request's marshalled bytes unique. The
+	// pubsub-rpc layer keys its response channel by a hash of the request bytes, so without this
+	// two concurrent syncs for the same docIDs would collide on that key and one would time out
+	// while the other succeeds. Mirrors the KMS pubsub fix (fetchEncryptionKeyRequest).
+	RequestID []byte `json:"requestID"`
```
AndrewSisley, line 48: question: Why split the requests like this, instead of combining them?

What is the benefit to users in keeping them seperate?

islamaliev: combining them sounds interesting. If two callers ask for the same documents at the same time, sending two identical requests is wasteful. We can do that with single-flight at the DB level: the first caller does the real request, and any concurrent caller asking for the same documents waits on that result.

And to be clear: the routing collision only ever happens for byte-identical requests, i.e. the same docIDs. Different or even partially-overlapping docID sets already have different keys and never collide. 

I can proceed with single-flight solution unless you had another solution in mind.

islamaliev: Another caveat that we should consider if we decide with combining them is that we also have to include an identity to the serialized key. So in total we will serialize key on collection name on doc ids and on identity. We need to do this because for different users the results should be different. That kind of adds additional overhead. To every request! What do you think?

AndrewSisley: >  sending two identical requests is wasteful

This is a very small amount of wastage though no?

> Another caveat that we should consider if we decide with combining them

Sorry, my original comment was ambiguous.  I wasn't really talking about combining the request messages, only their results.

So if two sync requests were requested by users, two would be sent, and their timeouts/results would merge - the second request to be handled by the client-node would essentially extend the timeout begun by the first request.  The first request would timeout if/when it's timeout its timeout is reached, the second if/when its timeout is reached.

The server-nodes would need to re-check upon receipt of the second request in case their are new heads, but there is no need to keep the requests or responses seperate via a nonce - either results are recieved by the client-node within the relevant timeout, or they are not - it does not matter which request triggered the response, they are all equally valid?

AndrewSisley (approved):
LGTM, sorry - I probably should have approved this when I asked the pending question - it is not a blocker - is just me exploring the solution and wondering if we can make additional improvements (especially as dropping the new prop is a breaking change).

PR #5028, opened 2026-07-07
fix: persist typed slices in nillable array fields

mennatnaga:
- Passing a typed Go slice ([]string, []time.Time) to NewDocFromMap for a nillable array field ([String], [DateTime]) silently dropped it to an empty array, because getNillableArray and getNillableDateTimeArray had no case for a raw typed slice (only *fastjson.Value, []any, and the already-option-wrapped slice). This adds that case to both parsers, wrapping each element in immutable.Some, matching the non-nillable getArray/getDateTimeArray that already handled it.
-  addDelta signed every block even when a batch collector was active, redundant work the single Merkle-root batch signature already covers. Now skips per-block signing when a collector is present; the collector still records CIDs, so the batch signature is unchanged.

### DIFF
--- a/client/document.go
+++ b/client/document.go
@@ -608,6 +608,12 @@ func getNillableArray[T any](
 		}
 
 		array = arr
+	case []T:
+		arr := make([]immutable.Option[T], len(val))
+		for i, item := range val {
+			arr[i] = immutable.Some(item)
+		}
+		array = arr
 	case []immutable.Option[T]:
 		array = val
 	}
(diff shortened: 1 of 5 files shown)


dvncan (approved):

fredcarle:
> Passing a typed Go slice ([]string, []time.Time) to NewDocFromMap for a nillable array field ([String], [DateTime]) silently dropped it to an empty array, because getNillableArray and getNillableDateTimeArray had no case for a raw typed slice (only *fastjson.Value, []any, and the already-option-wrapped slice). This adds that case to both parsers, wrapping each element in immutable.Some, matching the non-nillable getArray/getDateTimeArray that already handled it.

Nice catch @mennatnaga!

PR #5032, opened 2026-07-08
fix: commit PurgeByDocIDs per chunk to stay under the store transacti…

mennatnaga:
PurgeByDocIDs committed the whole document set in one transaction, so large prunes (up to 50k docIDs from the pruner) exceeded badger's per-transaction size limit and failed with "Txn is too big", leaving disk uncleaned. When no caller transaction is supplied it now commits in chunks of 100 docs; a caller-provided transaction still runs inline unchanged.


jsimnz:
100 docs per txn seems low for the limit?

PR #5033, opened 2026-07-08
fix: Handle typed Go slices in nillable array parsers

### LINKED ISSUE #5029: NewDocFromMap silently drops typed Go slices for nillable array fields
### Problem

Passing a typed Go slice (e.g. `[]string`, `[]time.Time`) to `NewDocFromMap` for a **nillable** array field (`[String]`, `[DateTime]`, etc.) silently drops it to an empty array.

`getNillableArray` and `getNillableDateTimeArray` in `client/document.go` only handle `*fastjson.Value`, `[]any`, and the already-option-wrapped slice (`[]immutable.Option[T]`). They have no case for a raw typed slice `[]T`, so the value falls through to the empty default.

The non-nillable equivalents (`getArray`, `getDateTimeArray`) already handle `[]T`, so only nillable array fields are affected.

### Impact

Affects every nillable array kind: `[String]`, `[Boolean]`, `[Int]`, `[Float]`/`[Float64]`, `[Float32]`, `[DateTime]`. Data is lost with no error.

### Fix

Add a `[]T` case to both parsers, wrapping each element in `immutable.Some`, matching the non-nillable getters.

fredcarle:
## Relevant issue(s)

Resolves #5029

## Description

`getNillableArray` and `getNillableDateTimeArray` had no case for a raw typed slice (`[]string`, `[]time.Time`, etc.), so passing one to `NewDocFromMap` for a nillable array field silently dropped it to an empty array. This adds a `[]T` case to both, wrapping each element in `immutable.Some`, matching the non-nillable `getArray`/`getDateTimeArray` which already handled it.

Also expands `tests/integration/mutation/add/field_kinds/` to cover the full field-kind matrix (nillable/non-nillable, scalar/array, typed slice, `[]any`, and nil variants).

## Tasks

- [x] I made sure to discuss its limitations such as new dependencies, performance concerns, etc.

## How has this been tested?

Added integration tests covering every affected field kind; the full `field_kinds` suite and `client` package pass locally.


shahzadlone (approved):
Thanks for adding all the text cases :) LGTM

client/document.go
```diff
@@ -610,6 +610,12 @@ func getNillableArray[T any](
 		array = arr
 	case []immutable.Option[T]:
 		array = val
+	case []T:
```
shahzadlone: question/thought: Correct me if wrong but as things stand,  now our `switch val := v.(type)` recognizes these cases:
  - `*fastjson.Value`
  - `[]any`
  - `[]immutable.Option[T]`
  - `[]T`

  For an [Int] field, T is int64, so []T means []int64.

  These values work:
```go
  []int64{1, 2}                         // matches []T
  []any{int64(1), int64(2)}             // matches []any
  []immutable.Option[int64]{...}        // matches option slice
```
  These will not match i think:
```go
  []int{1, 2}       // []int is not []int64
  [2]int64{1, 2}    // array is not a slice
  type IDs []int64
  IDs{1, 2}         // defined type is not exactly []int64
```

A []int also cannot match []any as slices are not covariant in Go.

Then I think the function initializes its result as an empty slice: `array := []immutable.Option[T]{}`

Because atm we dont have a default case, an unsupported input falls through the switch unchanged. For an unconstrained array field where `size == 0`, the function then returns that empty slice with no error. wouldn't that mean if used incorrectly we can have a silent "data loss":
```go
  input:  []int{1, 2}
  result: []              // no error
```

Ofc atm for a fixed-size field, it is not silent as the empty result fails the size check and catches that. Just thinking out loud. This is not in scope of this PR though nor blocking as this already existing from before.

client/document.go
shahzadlone, line 736: thought: Similar story as my other comment above, this accepts []time.Time, but not []string, so:

```go
  []time.Time{time.Now()}                       // works
  []any{"2026-07-09T12:00:00Z"}                 // works; each item is parsed
  []string{"2026-07-09T12:00:00Z"}              // no matching case; becomes []
```

and so each string is individually valid to the DateTime converter, but a typed []string never reaches that converter.

tests/integration/mutation/add/field_kinds/field_kind_float_test.go
```diff
@@ -0,0 +1,671 @@
...
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestMutationAddFieldKinds_WithNonNillableFloatArray_FromAnySlice(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddCollection{
+				SDL: `
+					type User {
+						points: [Float64!]
```
shahzadlone: nitpick: This uses `[Float64!]` rather than [Float!]. This duplicates the Float64 path and makes the test name misleading? perhaps change the schema to `[Float!]`, or rename/reposition the test as a `Float64` test.

client/document.go
fredcarle: Fixed. Let me know what you think.

client/document.go
fredcarle: Fixed. Let me know what you think.

tests/integration/mutation/add/field_kinds/field_kind_float_test.go
fredcarle: Fixed.

Issue #4837, opened 2026-05-26
bug(p2p+acp): inbound P2P sync of policy-gated docs stalls when sender ACP check outruns receiver block-fetch deadline


islamaliev:
On a peer node that subscribes to a policy-gated collection (`AddCollectionSubscription`, not the replicator push path), an inbound doc can fail to land with a generic "context deadline exceeded" error during DAG sync. Once you hit it, the doc stays missing on that node until you give up and re-sync explicitly.

Reported from a downstream group-chat feature: each chat has a Conversation + Message + UserConversation document, all under document ACP, and the receiving peer gets stuck for new chats while every grant lands cleanly on SourceHub. Their workaround today is a public side-channel collection that triggers an explicit `sync_documents` on the receiver. Works, but it leaks invite metadata.

I reproduced the failure in a small two-node integration test and instrumented both sides to capture the timeline. What's actually happening:

1. The receiver gets the pushlog. The receiver-side access check (`trySelfHasAccess`) asks SourceHub whether this node has read access to the doc. If grants haven't propagated yet, the answer is no — and the receiver silently drops the message. No log, no error, no retry. The receiver has to wait for another pubsub announce to try again.

2. Once grants are in place and a later announce gets through, the receiver enters `syncDAG` and starts fetching linked blocks. Each fetch has a per-block deadline (`P2PBlockSyncTimeout`, default 5s).

3. On the sender side, every served block goes through `hasAccess`, which also queries SourceHub — once per block. If the requesting peer is not in the sender's replicators list (which it isn't, in the subscription path), there is no short-circuit; every linked block is an extra SourceHub round-trip on the sender.

4. If the sender's per-block round-trip is slower than the receiver's per-block deadline, the receiver gives up first. The sender then finishes its check, gets a `peerHasAccess=true`, and tries to serve the block — but the receiver's context has already been cancelled. From the receiver's perspective it looks like a network timeout. From the sender's perspective it looks like a successful authorize-and-serve. Neither side logs anything wrong.

That last point is the reporter's "node1 logs ZERO activity in response to node2's pushlog attempts" complaint. The sender IS running its hasAccess code; it just answers after the receiver has stopped listening.

## Repro

[...]

PR #5035, opened 2026-07-10
feat: Configurable P2P block-sync timeout

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/4837

## Description

Fetching each block during a P2P document sync was bounded by a fixed five-second timeout that operators could not change. When a node talks to a real access-control chain, the per-block authorization check can take longer than that, so syncs would fail with no way to give them more room. The timeout is now configurable via a flag and environment variable, its default is raised to thirty seconds, and an individual sync request can override it for that call alone. A block fetch that runs out of time now reports a distinct timeout error instead of a generic load failure.

This is one half of the work on the linked issue; a separate change fixes the underlying access-control cost that made the timeout matter.

## Tasks

## How has this been tested?

Unit tests for the timeout resolution and the distinct error, and an integration test that drives a sync with a per-request timeout too small to complete (the document does not sync) and again with a generous one (it does), confirming the override reaches the per-block fetch. Existing P2P sync suites continue to pass.

Tested on:
- MacOS


cli/config/config.go
```diff
@@ -104,6 +105,7 @@ var ConfigDefaults = map[string]any{
 	"net.peers":                         []string{},
 	"net.pubSubEnabled":                 true,
 	"net.relay":                         false,
+	"net.p2pblocksynctimeout":           30,
```
nasdf, line 108: thought: It could be nicer to use a duration here (https://pkg.go.dev/time#ParseDuration).

docs/website/references/cli/defradb_start.md
```diff
@@ -33,6 +33,7 @@ defradb start [flags]
       --no-signing                        Disable signing of commits.
       --no-telemetry                      Disables telemetry reporting. Telemetry is only enabled in builds that use the telemetry flag.
       --node-acp-enable                   Enable the node access control system.
+      --p2p-block-sync-timeout int        Timeout in seconds for fetching each block during P2P DAG sync (default 30)
```
nasdf, line 36: suggestion: It would be nice if this was a duration like the other flags.

nasdf (approved):
Changes look good! I left two comments I think are worth looking into.

PR #5041, opened 2026-07-13
fix: Cache P2P access decisions to stop sync stalls

### LINKED ISSUE #4837: bug(p2p+acp): inbound P2P sync of policy-gated docs stalls when sender ACP check outruns receiver block-fetch deadline

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4837

## Description

When a node subscribes to a policy-gated collection, the node serving a document's blocks re-ran the access-control check once per block of that document's history. Each check is a round-trip to the access-control system, so syncing a document with any history meant many sequential round-trips; when those outran the receiver's per-block fetch deadline, the receiver gave up and the document silently failed to sync even though access had been granted.

The serving node now reuses a recent positive access decision for a given peer and document, so fetching a whole document costs one access-control round-trip rather than one per block. Only granted access is cached, and only briefly, so a revoked grant takes effect quickly and a not-yet-propagated grant is always re-checked. The cache is keyed per collection as well, so a grant on one branchable collection is never reused for another.

## Tasks

## How has this been tested?

Unit tests for the cache (freshness, expiry, per-peer/collection/document scoping, the collapse of repeated checks into one, and concurrent access under the race detector) and an integration test on the subscription path that syncs a multi-update document under access control and confirms a non-creator can read it once grants land. The full document-access-control P2P suite passes against a real SourceHub chain.

Tested on:
- MacOS


internal/db/p2p/access_cache.go
```diff
@@ -0,0 +1,122 @@
...
+// block.
+//
+// Only positive (allowed) decisions are stored. A denial is never cached: on the subscription
+// path a denial usually means the access grant has not propagated yet, and caching it would keep
+// a peer locked out until the entry expired even after the grant lands. Re-checking a denial is
+// cheap relative to the sync failure that caching one would cause.
+//
+// Entries expire after ttl. Because only grants are cached, the ttl is short, and a cache hit does
+// not extend expiry, the worst case after a revocation is that a peer retains read access for at
+// most one ttl window — a briefly stale grant, never a stale denial. Expiry is lazy (stale entries
```
shahzadlone: nitpick: Haha sorry for bringing this up again, i dislike the AI em-dashes 
```suggestion
// most one ttl window - a briefly stale grant, never a stale denial. Expiry is lazy (stale entries
```

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
```diff
@@ -0,0 +1,206 @@
...
+						}
+					}`,
+				Results: map[string]any{
+					"Users": []map[string]any{
+						{"name": "Alice", "age": int64(33)},
+					},
+				},
+			},
+
+			// The #4837 assertion: the non-creator with a reader grant sees the document on the
```
shahzadlone: todo: Remove mentioning of the issue `#` that you are fixing in this PR, seems like AI slop/over comment

internal/db/p2p/access_cache_test.go
```diff
@@ -0,0 +1,162 @@
...
+	cache, _ := newTestAccessCache(time.Second)
+	cache.storeAllowed("peerA", "colX", "doc1")
+
+	assert.True(t, cache.allowed("peerA", "colX", "doc1"))
+	assert.False(t, cache.allowed("peerB", "colX", "doc1"), "a grant for one peer must not leak to another")
+	assert.False(t, cache.allowed("peerA", "colX", "doc2"), "a grant for one doc must not leak to another")
+	assert.False(t, cache.allowed("peerA", "colY", "doc1"), "a grant for one collection must not leak to another")
+}
+
+// TestAccessCache_CollectionLevelBlocksDoNotCollide guards the #4837 access-control boundary:
```
shahzadlone: todo: Remove mentioning of the issue `#` that you are fixing in this PR, seems like AI slop/over comment

internal/db/p2p/access_cache_test.go
```diff
@@ -0,0 +1,162 @@
...
+		clock.advance(100 * time.Millisecond)
+		assert.True(t, cache.allowed("peer", "col", "doc"))
+	}
+
+	clock.advance(200 * time.Millisecond) // now past the original 1s expiry
+	assert.False(t, cache.allowed("peer", "col", "doc"),
+		"expiry must be measured from the store, not extended by hits")
+}
+
+// TestAccessCache_CollapsesRepeatedLookups is the mechanism at the heart of the #4837 fix: many
```
shahzadlone: todo: Remove mentioning of the issue `#` that you are fixing in this PR, seems like AI slop/over comment

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"testing"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+// This mirrors the downstream group-chat timeline that surfaced #4837: on the subscription
```
shahzadlone: todo: Remove mentioning of the issue `#` that you are fixing in this PR, seems like AI slop/over comment

internal/db/p2p/access_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	key := accessCacheKey{peerID: peerID, collectionID: collectionID, docID: docID}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// Reads keep their own key fresh (see allowed), but a key that is never read again would leak.
+	// Store is the only place the map grows, so bound it here with a bulk sweep rather than scanning
+	// the whole map on every (hot-path) read.
+	if len(c.entries) >= sweepThreshold {
```
shahzadlone: question: should we bound the cache here? the full map scan on every store past 1024 entries would serialize block serving behind the global mutex ( example during a fresh grant burst)

internal/db/p2p/access_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		c.sweepExpired()
+	}
+	c.entries[key] = c.now().Add(c.ttl)
```
shahzadlone: question: should this expire from when the check starts rather than store time, just wodnering because the documentation said "at most one TTL"

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
shahzadlone, line 1: suggestion: Only if easy to do so, add a deterministic hasAccess level test (counting/delaying ACP) that actually fails on the pre fix code, plus denied then granted and revocation after expiry perhaps.

internal/db/p2p/access_cache.go
islamaliev: You mean to To limit number of iterations in a single sweep? We can do this of course. You can also implement a priority queue that would always remove the oldest ones.

internal/db/p2p/access_cache.go
islamaliev: The check starts few nanoseconds or milliseconds before I'm not sure this is critical. And it doesn't have to be precise.

internal/db/p2p/access_cache.go
islamaliev: Done

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
islamaliev: Done

internal/db/p2p/access_cache_test.go
islamaliev: Done

internal/db/p2p/access_cache_test.go
islamaliev: Done

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
islamaliev: Done

internal/db/p2p/access_cache.go
islamaliev: Done. Replaced the full scan with a bounded front-eviction over an insertion-ordered queue (uniform TTL means insertion order = expiry order), capped per store.

tests/integration/acp/dac/p2p/subscribe_grant_race_test.go
islamaliev: Done at the cache level: added denied-then-granted and revocation-after-expiry tests.

shahzadlone (approved):
Thanks for doing all the changes :)

LGTM

PR #5054, opened 2026-07-14
fix: Consolidate client NAC gates within harness

### LINKED ISSUE #4446: NAC Permission errors differ across differing environments
A small number of our tests (six at the time of my writing this) produce different results when run in different environments. For example, look at the following test:

```
func TestNAC_GatesDocumentCreate_NoIdentity_NotAuthorizedError(t *testing.T) {
	test := testUtils.TestCase{
		Actions: []any{
			// Starting with NAC, so only authorized user(s) can perform operations from here on out.
			testUtils.Close{},
			testUtils.Start{
				Identity:  testUtils.ClientIdentity(1),
				EnableNAC: true,
			},
			// Note: Doing setup steps after starting with nac enabled, otherwise the in-memory tests
			// will lose setup state when the restart happens (i.e. the restart that started nac).
			&action.AddSchema{
				Identity: testUtils.ClientIdentity(1),
				Schema:   `type User { name: String }`,
			},

			// We haven't authorized non-identities. So, this should error.
			&action.CreateDoc{
				Identity:     testUtils.NoIdentity(),
				CollectionID: 0,
				Doc:          `{ "name": "Shahzad" }`,
				ExpectedError: "not authorized to perform operation",
			},
			&action.Request{ // Should not be created
				Identity: testUtils.ClientIdentity(1),
				Request:  `query{ User { name } }`,
				Results:  map[string]any{"User": []map[string]any{}},
			},
		},
	}

}
```

This will always pass, because the expected error is simply a string, "not authorized to perform operation" which does not check for a specific permission. However, we want to be able to check for a specific permission, and we can't do so, because that permission may differ across environment setups. Notably, when I run this locally, using the Go client, the permission wall it hits is for `document-update` while on the GitHub CI, the one it hits is `collection-get`.

The command I used to run locally, on WSL was:

`go test -v ./tests/integration/acp/nac/ -run TestNAC_GatesDocumentCreate_NoIdentity_NotAuthorizedError -count=1
=== RUN   TestNAC_GatesDocumentCreate_NoIdentity_NotAuthorizedError`

and the environment variable `DEFRA_CLIENT_GO` was set to true.

shahzadlone:
## Relevant issue(s)
Resolves #4446
Partially #4383

## Description
The same NAC authorization test reported different permission errors depending on which client ran it, for example: `document-update` locally via the Go client, but `collection-get` in CI via HTTP. So the tests could only assert the vague generic "not authorized to perform operation" string, not the actual permission.
### Root cause
It was a test-harness artifact, not a product bug. Every operation is gated twice: first `get-collection` (to resolve the collection), then the operation's own gate. In production all clients resolve-then-operate as the caller, so an unauthorized actor always hits `get-collection` first. 

But the test harness resolved collections using the node's own identity, which bypasses acp (the node-identity DID match short-circuits the check). That masked the get-collection gate for the embedded Go/JS client, so it fell through to the operation-specific gate. While name-addressed clients (HTTP/CLI/C) still hit `get-collection`. 

### Solution
Make the harness resolve collections as the acting identity instead of the node identity, so every client uniformly hits `get-collection` first when unauthorized.

Concretely:
  - Deleted the node-identity fallback (if !identity.HasValue() { identity = NodeIdentity(nodeID) }) that was masking the gate.
  - Collapsed helpers into `GetCollectionsCanonically` (resolves as the acting identity)
  - Applied it to document ops and the collection/index ops (truncate, list/new/delete index, encrypted-index).
  - Deleted all the per-client `_CLIandCandHTTPClient_` split-test variants, every gate test now asserts a single, uniform `get-collection`.

NOTE: This will make life easier for when we do granular permissions (#4383)

@ChrisBQu you might have some context for when you made this issue, will appreciate a look from you

### DIFF
--- a/tests/action/new_index.go
+++ b/tests/action/new_index.go
@@ -96,7 +96,7 @@ func (a *NewIndex) Execute() {
 		if err != nil {
 			expectedErrorRaised := assertError(a.s.T, err, a.ExpectedError)
 			assertExpectedErrorRaised(a.s.T, a.ExpectedError, expectedErrorRaised)
-			return
+			continue
 		}
 
 		collection := collections[a.CollectionID]
@@ -131,9 +131,7 @@ func (a *NewIndex) Execute() {
 		desc, err := collection.NewIndex(a.s.Ctx, indexDesc, opts)
 
 		expectedErrorRaised := assertError(a.s.T, err, a.ExpectedError)
-		if expectedErrorRaised {
-			return
-		}
+		assertExpectedErrorRaised(a.s.T, a.ExpectedError, expectedErrorRaised)
 
 		// Unless the test wants to observe the building window, wait for the build so a following
 		// query sees a built index. With an explicit transaction the record is not committed until
@@ -142,8 +140,6 @@ func (a *NewIndex) Execute() {
 			waitForIndexBuilt(a.s, collection, desc.ID, listIndexesOptions(a.s, node))
 		}
 	}
-
-	assertExpectedErrorRaised(a.s.T, a.ExpectedError, false)
 }
 
 // indexBuildTimeout bounds the wait for a background build or drop to finish. The poll returns as
--- a/tests/action/new_index.go
+++ b/tests/action/new_index.go
@@ -85,18 +85,17 @@ func (a *NewIndex) Execute() {
 		node := a.s.Nodes[nodeID]
 
 		// Check if a transaction is attached to this action. If so, we will be using it.
-		var err error
-		var txn client.Txn
-		var collections []client.Collection
+		txnOption := immutable.None[client.Txn]()
 		if a.TransactionID.HasValue() {
-			txn, err = a.s.GetTransaction(node, a.TransactionID)
+			txn, err := a.s.GetTransaction(node, a.TransactionID)
 			require.NoError(a.s.T, err)
-			collections, err = txn.GetCollections(a.s.Ctx, options.GetCollections())
-		} else {
-			collections, err = node.GetCollections(a.s.Ctx, options.GetCollections())
+			txnOption = immutable.Some(txn)
 		}
 
+		collections, err := GetCollectionsCanonically(a.s, node, txnOption, a.Identity)
 		if err != nil {
+			expectedErrorRaised := assertError(a.s.T, err, a.ExpectedError)
+			assertExpectedErrorRaised(a.s.T, a.ExpectedError, expectedErrorRaised)
 			return
 		}
 
--- a/tests/action/utils.go
+++ b/tests/action/utils.go
@@ -90,23 +90,17 @@ func RefreshCollections(
 	}
 }
 
-// getCanonicallyOrderedCollections gets the collections inside of a transaction, if one is provided.
-// If one is not provided, it will default to running the GetCollections function on the node itself.
-// Importantly, this will use the same ordering as would be found in the node.Collections slice that
-// is refreshed by the RefreshCollections function.
-func getCanonicallyOrderedCollections(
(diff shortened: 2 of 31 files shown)


tests/integration/utils.go
```diff
@@ -1923,7 +1945,13 @@ func deleteEncryptedIndex(
 		}
 
 		nodeID := nodeIDs[index]
-		collections := action.MustGetCanonicallyOrderedCollections(s, node, txnOption)
+
+		collections, err := action.GetCollectionsCanonically(s, node, txnOption, a.Identity)
+		if err != nil {
+			expectedErrorRaised := AssertError(s.T, err, a.ExpectedError)
+			assertExpectedErrorRaised(s.T, a.ExpectedError, expectedErrorRaised)
+			return
```
nasdf: question: Why does this one `return` instead of `continue` like the other utils that assert errors?

nasdf (approved):
The fixes look good to me. I left one comment that may or may not be an issue.

tests/action/utils.go
```diff
@@ -90,23 +90,17 @@ func RefreshCollections(
...
-func getCanonicallyOrderedCollections(
-	s *state.State,
-	node *state.NodeState,
-	txn immutable.Option[client.Txn],
-) ([]client.Collection, error) {
-	return getCanonicallyOrderedCollectionsWithIdentity(s, node, txn, immutable.None[state.Identity]())
-}
-
-func getCanonicallyOrderedCollectionsWithIdentity(
+// GetCollectionsCanonically resolves the collections as the given actor identity, using the same
```
ChrisBQu, line 94: suggestion: I like the new context provided by this comment. But I think it loses the context provided the original, which explained what would happen in the case of a transaction not being provided. For your consideration, but not a blocker or a specific request for change.

ChrisBQu (changes requested):
todo: Changing the name/behavior of the `getCanonicallyOrderedCollections` function looks good to me, but the change is incomplete, so your CI actions are failing. Check out the logs if there's something I might have missed, but at minimum, you have a couple of stale references to `MustGetCanonicallyOrderedCollections` inside `tests/action/wait_for_index.go`.

tests/action/utils.go
shahzadlone: - [x] Fixed in commit: `PR(DOCS): Restore txn fallback note on GetCollectionsCanonically doc`

tests/integration/utils.go
shahzadlone: Thanks for this, there were 2 problems with the return:
  1. Multi-node things would false green. The loop runs over every node. return exits on the first node's denial, even though for now they are one node tests but it's good to improve this IMO
  2. Double assertion on that path. It calls assertExpectedErrorRaised inline and again in the final
  assertExpectedErrorRaised(..., false) after the loop.
  
 - [x] ~~Fixed this in commit: ` PR(FIX): Accumulate expected errors across nodes in test helpers `~~
 - [x] Update (because code rabbit flagged something in previous fix that i improved on): Fixed in `PR(FIX): Improve the raised error asserts...` commit now

tests/action/new_index.go
```diff
@@ -80,24 +80,24 @@ var _ Action = (*NewIndex)(nil)
...
-		} else {
-			collections, err = node.GetCollections(a.s.Ctx, options.GetCollections())
+			txnOption = immutable.Some(txn)
 		}
 
+		collections, err := GetCollectionsCanonically(a.s, node, txnOption, a.Identity)
 		if err != nil {
-			return
+			expectedErrorRaised = expectedErrorRaised || assertError(a.s.T, err, a.ExpectedError)
+			continue
```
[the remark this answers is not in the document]
shahzadlone: - [x] Resolved

PR #5056, opened 2026-07-15
fix: delete secondary index entries when purging documents

mennatnaga:
Purging a document deleted its data but left its secondary-index entries behind.
Re-adding the same document later then failed with a unique-index violation against the stale entry.

`purgeOneDoc` now deletes the document's index entries before its datastore values.


iverc (approved):
LGTM 🚀

PR #5069, opened 2026-07-20
fix: Validate empty p2p cli address args

### LINKED ISSUE #4899: Check number of args on CLI Connect and Disconnect commands.
The CLI commands should return early with an descriptive error if the command was used without specifying one or more addresses.

SharanRP:
## Relevant issue(s)

Resolves #4899 

## Description

Adds argument validation to the `client p2p connect` and `client p2p disconnect` commands so they fail early when called without any peer addresses. Previously, both commands accepted zero arguments and passed an empty address list into the client layer. This meant `disconnect` could silently no-op, and `connect` could continue into lower-level runtime behavior instead of returning a clear CLI error.

## Tasks

## How has this been tested?
Ran the below tests
  - `go test ./cli/test/integration/p2p`
  - `go test ./... -race -shuffle=on -timeout 10m`

- Debian Linux


nasdf (approved):
LGTM. Thanks for fixing this!

shahzadlone (approved):
LGTM, Thanks :)

Issue #5067, opened 2026-07-20
Network wire-type drift detection (RFC)


islamaliev:
## Problem

Some types that live in `internal` marshal onto the wire (P2P messages, CRDT blocks). They leave one machine and arrive on another in exactly the shape they're defined, so they're effectively part of our public, cross-version contract even though they aren't in a public package. Changing one of these types can silently break compatibility between node versions, and nothing today flags that a change touched a wire-facing type.

## Scope

Detect, at the PR level, when a change alters the on-the-wire shape of a type that other node versions depend on, so the author and reviewers are aware it carries cross-version and cross-client impact.

This is the cheap, shape-level counterpart to the behavioral P2P compatibility testing (separate issue). Shape detection catches "this type changed"; the compat matrix catches "this behavior broke."

## Open questions (not decided)

- Which types are in scope, and how do we identify them reliably (they're spread across internal packages today)?
- Public package vs. keep internal? Making wire types a public package makes the contract explicit but pulls things like CRDT blocks into the public API, which we may not want.
- What does "detection" mean concretely, and how do we acknowledge an intentional change without it becoming noise?

## Some early thoughts (my ideas, not a proposal)

Just starting points, nothing decided:

- Keep the types internal but document them clearly, and gather the wire-facing ones under a dedicated package convention (a DTO-style package) so the boundary is explicit.
- A linter or golden-file/snapshot check that scans that package and fails CI when a wire type's shape changes unless the change is explicitly acknowledged, similar in spirit to the change detector's opaque file.

Needs discussion and a decision before implementation. Related to #5066 (P2P cross-version compatibility testing).

PR #5078, opened 2026-07-23
feat: Wire format drift detection

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/5067

## Description

Renaming or retyping a field on a type that nodes send to each other silently breaks communication with nodes running an older version, and nothing caught it. This adds a registry where those node-to-node types declare themselves and a golden snapshot of their shape. When a wire type changes shape, a test fails until the snapshot is regenerated and a short note is added under `docs/wire_format_changes` explaining whether old and new nodes can still talk, the same bar already used for on-disk data format changes.

This covers both node-to-node wire layers: the CBOR message types (rendered from their Go fields, since the encoder keys by field name) and the IPLD block types exchanged as CRDT changes (rendered from the IPLD schema each declares, including the CRDT union's discriminator strings, which are wire-critical). It does not yet catch a brand-new wire type that was never registered at all; that completeness check is tracked as a follow-up (https://github.com/sourcenetwork/defradb/issues/5079).

## Tasks

## How has this been tested?

New tests:
- `internal/wire/wire_test.go`: registry behavior (a registered type is found, a local-only type is not treated as wire).
- `internal/wire/snapshottest`: the snapshot matches the committed golden, and a change to the golden versus the base branch requires an accompanying note. Verified end-to-end in a throwaway worktree: the pass path, the fail-without-note path, and that regenerating the golden alone does not bypass the note requirement. Also confirmed by falsification that both a CBOR field change and an IPLD union discriminator rename trip the snapshot.

- MacOS


internal/wire/snapshottest/wire_snapshot.golden
shahzadlone, line 1: question: what is this file?

.github/workflows/check-wire-format-changes.yml
```diff
@@ -0,0 +1,54 @@
...
+  check-wire-format-drift:
+    name: Check wire format drift job
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code into the directory
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
islamaliev: Done. Added contents: read and persist-credentials: false.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,158 @@
...
+// writeType renders one named type: a struct as its ordered fields, an interface
+// as its method set, anything else as its kind.
+func writeType(b *strings.Builder, t reflect.Type) {
+	switch t.Kind() {
+	case reflect.Struct:
+		fmt.Fprintf(b, "%s struct\n", typePath(t))
+		for i := range t.NumField() {
+			// Only exported fields are encoded, so only they are the wire shape.
+			if f := t.Field(i); f.IsExported() {
+				fmt.Fprintf(b, "\t%s %s\n", f.Name, typeName(f.Type))
```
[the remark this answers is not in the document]
islamaliev: Done. The snapshot now records the encoded key from the cbor/json tag, plus options like omitempty, not the Go field name.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+		}
+	case reflect.Interface:
+		fmt.Fprintf(b, "%s interface\n", typePath(t))
+		for _, m := range sortedMethods(t) {
+			fmt.Fprintf(b, "\t%s\n", m)
+		}
+	default:
+		fmt.Fprintf(b, "%s %s\n", typePath(t), t.Kind())
```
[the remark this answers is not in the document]
islamaliev: Done. A named container now shows its underlying shape, so an element-type change is caught.

internal/wire/snapshot_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+// TestSnapshot_Containers renders slice, nested slice, map, and array field
+// types structurally so a change to any element type is visible.
+func TestSnapshot_Containers(t *testing.T) {
+	out := snap(reflect.TypeFor[containers]())
+	assert.Contains(t, out, "Bytes []uint8")
+	assert.Contains(t, out, "Rows [][]string")
+	assert.Contains(t, out, "FixedArr [3]uint8")
+	assert.Contains(t, out, "ByKey map[string]")
+}
```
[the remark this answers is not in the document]
islamaliev: Done. The test now asserts the full map type including the value.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,214 @@
...
+// Snapshot renders the shape of every registered type as deterministic text. A
+// change to any field name, field type, field order, or IPLD discriminator
+// changes the output, so a diff against the committed golden flags a wire-format
+// change that must be intentional.
+//
+// Two wire layers are covered. CBOR message types are rendered from their Go
+// fields, since the encoder keys fields by their Go name. IPLD block types are
+// rendered from the schema they declare (including the CRDT union's wire-critical
+// discriminator strings), since that schema, not the Go struct, is their wire
+// shape.
```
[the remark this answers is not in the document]
islamaliev: Done. The snapshot now records the encoded key and options from the tag, not the Go name.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,214 @@
...
+// ordered fields, an interface as its method set.
+func writeType(b *strings.Builder, t reflect.Type) {
+	if schema, ok := ipldSchema(t); ok {
+		fmt.Fprintf(b, "%s ipld\n", typePath(t))
+		for line := range strings.SplitSeq(strings.TrimSpace(schema), "\n") {
+			// Collapse the schema literal's own indentation to one tab so the
+			// snapshot reads and diffs cleanly regardless of source formatting.
+			fmt.Fprintf(b, "\t%s\n", strings.Join(strings.Fields(line), " "))
+		}
+		return
```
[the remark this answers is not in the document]
islamaliev: Done. Whitespace inside a quoted value is now kept, so two discriminators that differ only by spacing stay distinct.

internal/wire/snapshot_test.go
```diff
@@ -0,0 +1,148 @@
...
+		snap(reflect.TypeFor[*structFields]()))
+}
+
+// TestSnapshot_Cycles terminates on self-referential and mutually-referential
+// types rather than looping. A regression that recurses before marking a type
+// seen would hang here.
+func TestSnapshot_Cycles(t *testing.T) {
+	assert.NotEmpty(t, snap(reflect.TypeFor[selfCycle]()))
+	assert.NotEmpty(t, snap(reflect.TypeFor[cycleA]()))
+}
```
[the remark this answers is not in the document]
islamaliev: Done. The test now asserts the fields and the referenced type are rendered, not just that output is non-empty.

internal/wire/snapshottest/snapshot_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+// noteAddedSince reports whether a note (not the README) was added under the wire
+// format changes directory between base and HEAD.
+func noteAddedSince(t *testing.T, base string) bool {
+	dir := changesDocPackage + "/" + changesDir
+	out, err := runGit(t, "diff", "--name-only", "--diff-filter=A", base+"...HEAD", "--", dir)
+	require.NoError(t, err)
+	for f := range strings.SplitSeq(strings.TrimSpace(out), "\n") {
+		if strings.HasSuffix(f, ".md") && !strings.HasSuffix(f, "README.md") {
+			return true
```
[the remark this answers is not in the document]
islamaliev: Leaving as is. This follows the same rule as the existing data format change check: a new note file is the signal, its content is for humans to read. Keeping both checks consistent.

internal/wire/snapshottest/wire_snapshot.golden
islamaliev: it's kind of described in the PR description.

It's a snapshot of current DTO structs that are traveling across the wire. So whenever the structure of any of these structs changes we will generate a similar file and do it diff against these two. And if there is a difference then we flag it a note to be written explaining why this change happened similarly how we do it with Change Detector

internal/wire/snapshottest/snapshot_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/sourcenetwork/defradb/internal/wire"
+
+	// Blank imports run each package's init so its wire types register.
+	_ "github.com/sourcenetwork/defradb/internal/core/block"
+	_ "github.com/sourcenetwork/defradb/internal/db/p2p"
+	_ "github.com/sourcenetwork/defradb/internal/db/p2p/message"
+	_ "github.com/sourcenetwork/defradb/internal/db/p2p/protocol"
+	_ "github.com/sourcenetwork/defradb/internal/kms"
+	_ "github.com/sourcenetwork/defradb/internal/se"
```
shahzadlone, line 39: question: Is there a cleaner way to do this? perhaps without using `init` approach? I find the init stuff super hard to follow order of control flow.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,283 @@
...
+				continue
+			}
+			key, opts, skip := cborField(f)
+			if skip {
+				continue
+			}
+			// The encoded key, not the Go field name, is the wire contract: the
+			// encoder honors a cbor or json tag. Options (omitempty, toarray,
+			// keyasint) are recorded too since they change the encoding.
+			fmt.Fprintf(b, "\t%s %s%s\n", key, typeName(f.Type), opts)
```
shahzadlone: question: Wouldn't embedded fields change from flat to nested without detection? Seems like current DTOs embed `message.MetaData`, which CBOR flattens. The snapshot instead records a `MetaData` field. Adding `cbor:"MetaData"` to the existing embed changes the real payload from flat metadata keys to a nested map, but produces the identical snapshot line. Would be worth checking as i think the golden test would still pass.

internal/wire/snapshot.go
```diff
@@ -0,0 +1,283 @@
...
+		visit(t)
+	}
+	return out
+}
+
+// writeType renders one named type. A type that declares an IPLD schema is
+// rendered from that schema, which already describes its fields and, for the
+// union, its wire-critical discriminators. Otherwise a struct is rendered as its
+// ordered fields, an interface as its method set.
+func writeType(b *strings.Builder, t reflect.Type) {
```
shahzadlone, line 111: question: would `MarshalCBOR` or `MarshalBinary` override field-based encoding? (but the snapshot only reflects fields). Adding either method can replace the entire wire representation without changing the golden. Registered types implementing these interfaces should fail unless they provide an explicit wire schema.

shahzadlone: cbor checks these before encoding: https://github.com/fxamacker/cbor/blob/v2.9.0/encode.go#L28-L34

internal/wire/snapshot.go
```diff
@@ -0,0 +1,283 @@
...
+		// method is not reached here; today no wire interface method carries one.
+		if t.Kind() == reflect.Struct {
+			for i := range t.NumField() {
+				// Only exported fields are encoded, so only they are part of the
+				// wire shape. Skipping the rest also avoids descending into a
+				// stdlib type's unexported internals (e.g. time.Time).
+				if f := t.Field(i); f.IsExported() {
+					visit(f.Type)
+				}
+			}
```
shahzadlone, line 91: question: I think named containers hide changes to element structs here? For ex: `type Items []Item`, traversal stops at `Items`. The snapshot names Item but never records its fields, so changing `Item.Value` from string to int is missed. Why not traverse slice, array, map and pointer contents even when the container itself is named?

.github/workflows/check-wire-format-changes.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Setup defradb
+        uses: ./.github/composites/setup-defradb
+
+      # Fails if a wire type changed shape without a note under
+      # docs/wire_format_changes. The test checks for the note against the base
+      # branch, so full history is needed (fetch-depth: 0 above). On a PR, diff
+      # against the actual target branch; otherwise the test falls back to develop.
+      - name: Run wire format drift detection
+        env:
+          DEFRA_CHANGE_DETECTOR_SOURCE_BRANCH: ${{ github.base_ref || 'develop' }}
```
shahzadlone, line 57: thought: probably don't need to worry but if retargeting a PR in future it will not rerun the base-dependent check. Note: since we use `github.base_ref`, but the default pull_request triggers exclude `edited`. A successful check against develop can therefore remain after retargeting to master. 

nitpick: We can possibly add types: [opened, synchronize, reopened, edited]? not a 100% sure though if we want to. 

moreover: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

.github/workflows/check-wire-format-changes.yml
shahzadlone, line 1: suggestion: Please name something similar to the other change detector workflow like: `check-wire-format-changes.yml` (also goes nicely with your folder name: `docs/wire_format_changes`)

.github/workflows/check-wire-format-drift.yml
```diff
@@ -0,0 +1,56 @@
...
+    ${{ github.ref == github.ref_protected && github.run_id || github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+# The job only reads the repo and runs tests over PR code; grant no write scope
+# and do not persist the checkout token while that code runs.
+permissions:
+  contents: read
+
+jobs:
+  check-wire-format-drift:
```
shahzadlone: suggestion: Similar to above, change the `drift` terminology to change
```suggestion
  check-wire-format-changes:
```

.github/workflows/check-wire-format-drift.yml
```diff
@@ -0,0 +1,56 @@
...
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+name: Check Wire Format Drift Workflow
```
shahzadlone: suggestion: Similar to above, change the `drift` terminology to change

```suggestion
name: Check Wire Format Changes Workflow
```

.github/workflows/check-wire-format-drift.yml
```diff
@@ -0,0 +1,56 @@
...
+  cancel-in-progress: true
+
+# The job only reads the repo and runs tests over PR code; grant no write scope
+# and do not persist the checkout token while that code runs.
+permissions:
+  contents: read
+
+jobs:
+  check-wire-format-drift:
+    name: Check wire format drift job
```
shahzadlone: suggestion: Similar to above, change the `drift` terminology to change
```suggestion
    name: Check wire format changes job
```

docs/wire_format_changes/i5060-wire-snapshot-baseline.md
```diff
@@ -0,0 +1,5 @@
+# Wire format snapshot baseline
+
+Adds the wire format drift detection: a registry of node-to-node types and a
```
shahzadlone: suggestion: Similar to above, change the `drift` terminology to change.

```suggestion
Adds the wire format changes detection: a registry of node-to-node types and a
```

.github/workflows/check-wire-format-drift.yml
```diff
@@ -0,0 +1,56 @@
...
+          persist-credentials: false
+
+      - name: Setup defradb
+        uses: ./.github/composites/setup-defradb
+
+      # Fails if a wire type changed shape without a note under
+      # docs/wire_format_changes. The test checks for the note against the base
+      # branch, so full history is needed (fetch-depth: 0 above). On a PR, diff
+      # against the actual target branch; otherwise the test falls back to develop.
+      - name: Run wire format drift detection
```
shahzadlone: suggestion: Similar to above, change the `drift` terminology to change

```suggestion
      - name: Run wire format changes detection
```

internal/core/block/wire.go
```diff
@@ -0,0 +1,37 @@
...
+package coreblock
+
+import (
+	"github.com/sourcenetwork/defradb/internal/core/crdt"
+	"github.com/sourcenetwork/defradb/internal/wire"
+)
+
+// The IPLD blocks exchanged between nodes are wire types. Their shape is the IPLD
+// schema each one declares, not the Go struct, so the snapshot records that
+// schema. Registering them here puts them under the same drift check as the CBOR
```
shahzadlone:
```suggestion
// schema. Registering them here puts them under the same wire changes check as the CBOR
```

shahzadlone (changes requested):
Left some questions and suggestions :)

internal/wire/snapshot.go
```diff
@@ -0,0 +1,283 @@
...
+		return
+	}
+	switch t.Kind() {
+	case reflect.Struct:
+		fmt.Fprintf(b, "%s struct\n", typePath(t))
+		for i := range t.NumField() {
+			f := t.Field(i)
+			if !f.IsExported() {
+				continue
+			}
```
shahzadlone: question:  Could this export check cause the snapshot to miss toarray wire format changes? cbor  i think treats the following unexported `_` field as a struct level encoding directive:

_ struct{} `cbor:",toarray"`

Because _ is not exported, it looks like we skip it before cborField can inspect it. Would adding this marker therefore change cbor.Marshal from a map encoding to a positional array without changing the snapshot? Would it be safer to detect struct-level options before applying IsExported() and cover this case with a regression test?

.github/workflows/check-wire-format-changes.yml
islamaliev: Done.

.github/workflows/check-wire-format-drift.yml
islamaliev: Done.

.github/workflows/check-wire-format-drift.yml
islamaliev: Done.

.github/workflows/check-wire-format-drift.yml
islamaliev: Done.

docs/wire_format_changes/i5060-wire-snapshot-baseline.md
islamaliev: Done.

.github/workflows/check-wire-format-drift.yml
islamaliev: Done.

internal/core/block/wire.go
islamaliev: Done.

internal/wire/snapshot.go
islamaliev: Good catch. Fixed. Untagged embeds are now flattened inline, a tag-renamed embed stays nested, so a switch between them is caught.

internal/wire/snapshot.go
islamaliev: Good catch. Fixed. Named containers now traverse into their element type.

internal/wire/snapshot.go
islamaliev: Good catch. Fixed. A struct-level toarray marker is now recorded.

.github/workflows/check-wire-format-changes.yml
islamaliev: Done. Added edited to the trigger types.

internal/wire/snapshot.go
islamaliev: Yes it would, but no wire type has a custom marshaler today. If one is added it must render from that instead. Resolving for now.

internal/wire/snapshottest/snapshot_test.go
islamaliev: Each package registers its own wire types, so init keeps that next to the types with no central list to keep in sync. Open to another approach if you prefer. Leaving open for your call.

shahzadlone: Not sure I have "better" alternatives atm. Can leave as is :) thanks

.github/workflows/check-wire-format-drift.yml
shahzadlone: Marked as done but not done.

.github/workflows/check-wire-format-drift.yml
shahzadlone: Marked as done but not done.

.github/workflows/check-wire-format-drift.yml
shahzadlone: Marked as done but not done.

.github/workflows/check-wire-format-drift.yml
shahzadlone: Marked as done but not done.

.github/workflows/check-wire-format-changes.yml
shahzadlone: Marked as done but not done.

shahzadlone (changes requested):
LGTM but i am confused as to why a lot of comments say done, but they aren't done.

shahzadlone (approved):
Thanks for doing all the changes :)

LGTM

Issue #5070, opened 2026-07-22
[Epic] Vector search & indexing


islamaliev:
DefraDB can already generate and store vector embeddings and score a document against a query vector, but similarity search still walks every document in the collection. That is fine for a handful of documents and unusable past a few thousand. This epic covers giving vectors a real index so nearest-neighbour search stays fast as collections grow, while keeping the whole thing at home on edge hardware.

The plan is to land one algorithm well before adding the next. We start with HNSW because it needs no training step and handles a stream of writes gracefully, which matches how documents actually arrive in DefraDB. Inverted-file (IVFFlat) follows as a second option better suited to large, mostly-static collections. Both are exposed the same way, so choosing between them is a one-word change in the schema and never a change to how you query.

A few decisions shape everything under this epic:

## One primitive for search, separate from generation
Generating an embedding and indexing a vector are different jobs, and DefraDB already has a dedicated directive for the first. Indexing gets its own dedicated directive rather than being folded into embedding config or into the existing scalar index directive. That keeps the three concerns independent and, importantly, lets you index vectors that DefraDB never generated — ones produced by your own or a proprietary model and written in directly.

## Built to run on small machines
The index lives on disk and only the actively-used parts are held in memory, under a budget the operator sets to fit the host. A device with little RAM caps it low; a server raises it. Nothing about the schema or the data changes between those two worlds. Holding an entire graph permanently in memory, the way some dedicated vector databases do, is explicitly out of scope because it does not fit the edge target.

## Maximum reuse of the existing index machinery
A vector index slots into the same lifecycle every other index already uses: the same background build worker, the same build/ready/failed states, the same crash-resume, the same write path (so peer-to-peer sync maintains it too), the same listing and explain surfaces. Only the two pieces that are inherently about sorted keys — how entries are encoded and how they are read back — are genuinely new.

## No new way to ask the question, at first
The initial query surface is the one that already exists: score by similarity, order by it, limit to k. The planner learns to route that shape to the index instead of scanning. A cleaner dedicated argument can come later, but existing queries should simply get faster with no rewrite.

Building on this, the initial release also corrects the current similarity scoring, which computes an unnormalised dot product while presenting itself as cosine, so results are wrong for vectors that are not already unit length.

The tracked work lives in the sub-issues. Beyond what is filed, advanced follow-ups such as filter-aware graph construction and vector quantization will be added as they are scoped.

PR #5096, opened 2026-07-28
feat: HNSW vector index for embedding fields

### LINKED ISSUE #5042: Hierarchical Navigable Small World (HNSW) - Vector index
Part of https://github.com/sourcenetwork/defradb/issues/5070.

Give vector fields a real index so nearest-neighbour search stops scanning the whole collection. This is the first vector index and the one that establishes the shared foundation the follow-ups build on, so it deliberately does a little more than "add HNSW": it also lays down the storage layout, the schema surface, and the query routing that every later algorithm reuses.

## What lands here
A new schema directive marks a vector field for indexing and carries its parameters (algorithm, distance metric, dimensions, and the graph build settings). It is separate from the embedding directive and from the existing scalar index directive, so it works whether the vector was generated by DefraDB or supplied from outside. Dimensions are inferred when the field is a generated embedding and required otherwise, and vectors whose length does not match are rejected on write rather than silently corrupting the index.

The index itself is HNSW, built and maintained by the same background worker that already handles other indexes: it backfills existing documents in batches, survives restarts from where it left off, and is kept up to date by ordinary writes and by peer-to-peer merges. The graph is stored on disk as one entry per node under the same key space every other index uses, so idle indexes cost disk rather than memory, and only the hot part of the graph is cached. Graph mutations are serialised through a single writer and deletes are handled as tombstones, which avoids the concurrent-update hazards that have historically been the fragile part of online graph indexes.

Search reuses the query that already exists — score by similarity, order by it, limit to k — with the planner recognising that shape and driving the index instead of a full scan, and reporting the index use through the existing explain output. This release also fixes the current scoring, which computes an unnormalised dot product but calls it cosine; the indexed path uses proper cosine distance on normalised vectors.

## Scope boundaries
Distance metric is cosine only for now; other metrics are a follow-up but the stored index configuration already carries the metric so adding them later is additive. The descriptor and storage layout are algorithm-agnostic from day one — they carry an algorithm discriminator and leave room for a training step — specifically so IVFFlat (https://github.com/sourcenetwork/defradb/issues/5043) slots in without reworking any of this. Combining a metadata filter with nearest-neighbour search correctly, and a dedicated query argument, are intentionally left for follow-ups.

## Testing
Integration tests covering: creating the index via the directive on both generated and externally-supplied vectors; dimension mismatch rejection; correctness of nearest-neighbour results against a brute-force baseline; index maintenance across create, update and delete; the background build completing over a pre-populated collection and resuming after an interrupted build; peer-to-peer replicated writes maintaining the index; and explain output confirming the index is used instead of a scan.

islamaliev:
DefraDB could already generate and store vector embeddings and score a document against a query vector, but similarity search still walked every document in the collection, which stops being viable once a collection grows past a few thousand documents. This adds a real approximate-nearest-neighbour index for vector fields using HNSW, so that similarity search stays fast as data grows while still running within the memory limits of an edge device.

Indexing a vector is now its own thing, declared with a dedicated directive that is separate from embedding generation and from the normal scalar index. Keeping it separate means a vector produced by DefraDB and a vector computed by some other model and written in directly are both indexable the same way. The index is built and kept up to date by the same background machinery every other index already uses, so it backfills existing documents, survives restarts partway through a build, and is maintained by ordinary writes and by peer-to-peer merges. The graph itself lives on disk with only its hot part held in memory, so an idle index costs disk rather than memory and a small device can cap how much it keeps resident.

The initial release ships cosine distance only and, in doing so, fixes the existing similarity scoring, which was computing an unnormalised dot product while presenting itself as cosine and so returning wrong results for vectors that were not already unit length. Existing similarity queries do not need to change to benefit — the planner recognises the shape they already use and drives the index instead of scanning, and reports that it did so in the query explanation. The stored index configuration and storage layout are deliberately written to be agnostic to the algorithm, so the inverted-file index that follows can reuse all of it without rework.

Part of https://github.com/sourcenetwork/defradb/issues/5070. Resolves https://github.com/sourcenetwork/defradb/issues/5042.


internal/db/collection_index.go
```diff
@@ -489,12 +496,70 @@ func processNewIndexRequest(
...
+	// No dimensions were given. That is only allowed when the field is an @embedding, since the
+	// model then fixes the dimensions. The value itself is filled in later.
+	for _, embedding := range def.VectorEmbeddings {
+		if embedding.FieldName == fieldName {
+			return nil
+		}
+	}
+
+	return NewErrVectorIndexMissingDimensions(fieldName)
+}
```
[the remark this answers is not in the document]
islamaliev: Done. Both are now rejected at index creation.

internal/db/collection_index.go
```diff
@@ -489,12 +496,70 @@ func processNewIndexRequest(
...
+	if params.M > client.MaxHNSWM {
+		return NewErrVectorIndexParamOutOfRange("M", params.M, client.MaxHNSWM)
+	}
+	if params.EfConstruction > client.MaxHNSWEfConstruction {
+		return NewErrVectorIndexParamOutOfRange("efConstruction", params.EfConstruction, client.MaxHNSWEfConstruction)
+	}
+	if params.EfSearch > client.MaxHNSWEfSearch {
+		return NewErrVectorIndexParamOutOfRange("efSearch", params.EfSearch, client.MaxHNSWEfSearch)
+	}
+	return nil
```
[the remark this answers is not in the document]
islamaliev: Done. Fixed once in openHNSW: a zero value now keeps the default.

internal/db/index.go
```diff
@@ -423,6 +429,152 @@ func (index *collectionSimpleIndex) Delete(
...
+
+	// Vectors of different lengths in one graph make distances meaningless. Dimensions is set for a
+	// directly-written field, so check against it. It is 0 only for an @embedding field, where the
+	// model fixes the length, so the only guard left is against an empty vector.
+	if index.desc.Vector.Dimensions > 0 && len(vec) != int(index.desc.Vector.Dimensions) {
+		return NewErrVectorDimensionMismatch(int(index.desc.Vector.Dimensions), len(vec), doc.ID().String())
+	}
+	if len(vec) == 0 {
+		return NewErrVectorDimensionMismatch(int(index.desc.Vector.Dimensions), 0, doc.ID().String())
+	}
```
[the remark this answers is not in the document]
islamaliev: Done. Added a dedicated empty vector error.

internal/db/index.go
```diff
@@ -423,6 +429,152 @@ func (index *collectionSimpleIndex) Delete(
...
+	if !found {
+		if index.building {
+			return nil
+		}
+		return NewErrCorruptedVectorIndex(index.desc.Name, doc.ID().String())
+	}
+
+	return idx.Delete(nodeID)
+}
```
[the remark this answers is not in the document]
islamaliev: Done. Delete now resolves the short id directly and no longer re-decodes the vector.

internal/db/vectorindex/hnsw_index.go
```diff
@@ -0,0 +1,90 @@
...
+	params := hnsw.DefaultParams(int(desc.HNSW.M))
+	params.EfConstruction = int(desc.HNSW.EfConstruction)
+	params.EfSearch = int(desc.HNSW.EfSearch)
+
+	store := newNodeStore(ctx, collectionShortID, indexID, epoch)
+	return &hnswIndex{
+		graph:    hnsw.New(store, metric, params, int64(indexID)),
+		efSearch: params.EfSearch,
+	}, nil
+}
```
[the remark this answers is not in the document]
islamaliev: Done. openHNSW now only overrides a default when the value is above zero.

internal/index/hnsw/codec.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if len(b) < off+countWidth {
+		return Node{}, ErrInvalidNodeEncoding
+	}
+	numLayers := binary.LittleEndian.Uint32(b[off:])
+	off += countWidth
+
+	var layers [][]NodeID
+	if numLayers > 0 {
+		layers = make([][]NodeID, numLayers)
```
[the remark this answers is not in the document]
islamaliev: Done. Added a bounds check before the allocation.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,344 @@
...
+		for _, sel := range selected {
+			if err := g.addLink(sel.id, id, layer, mmax); err != nil {
+				return err
+			}
+		}
+
+		// Candidates for the next (lower) layer's entry points are the
+		// full candidate set found at this layer.
+		entryPoints = w
+	}
```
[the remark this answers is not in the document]
islamaliev: Done. A new node with no live neighbours now becomes the entry point, so it stays searchable. Added a test.

internal/index/hnsw/hnsw.go
```diff
@@ -0,0 +1,125 @@
...
+		m = 16
+	}
+	return Params{
+		M:              m,
+		Mmax0:          2 * m,
+		EfConstruction: 128,
+		EfSearch:       64,
+		ML:             1 / math.Log(float64(m)),
+	}
+}
```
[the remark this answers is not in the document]
islamaliev: Done. The guard now covers m <= 1.

internal/index/hnsw/integration_test.go
```diff
@@ -0,0 +1,335 @@
...
+	meta, err := src.GetMeta()
+	require.NoError(t, err)
+	mb, err := MarshalMeta(meta)
+	require.NoError(t, err)
+	decodedMeta, err := UnmarshalMeta(mb)
+	require.NoError(t, err)
+	require.NoError(t, dst.PutMeta(decodedMeta))
+
+	return dst
+}
```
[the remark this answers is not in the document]
islamaliev: Good point. This is covered now by the new delete-then-insert test that exercises the tombstone path. Left the helper as is for the insert-only case.

internal/planner/vector_search.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	origScan.Prefixes(prefixes)
+	origScan.vectorIndexed = true
+	// Empty prefixes would otherwise let the scan fall back to reading the whole collection.
+	if len(prefixes) == 0 {
+		origScan.noResults = true
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
islamaliev: Done. The graph is now asked for limit plus offset. Added a test.

client/index.go
```diff
@@ -22,16 +23,159 @@ type IndexedFieldDescription struct {
...
+// DistanceMetric identifies the distance metric used to compare vectors.
+type DistanceMetric uint8
+
+const (
+	// DistanceMetricCosine identifies the cosine distance metric.
+	DistanceMetricCosine DistanceMetric = iota
+)
+
+// SecondaryIndexDescription holds config specific to secondary (scalar/unique/JSON) indexes.
+type SecondaryIndexDescription struct {
```
AndrewSisley: question: Why was `Unique` moved in a level?  

question: Are `Fields`, `Name`, and `ID` not also config specific to secondary (scalar/unique/JSON) indexes that are not found on primary indexes?

question: What other properties to you think will also live here in the future?

todo: This is very definitely a major semver breaking change - please retarget this branch to point at a v2 branch, not develop. If you keep this change.

client/index.go
```diff
@@ -22,16 +23,159 @@ type IndexedFieldDescription struct {
...
+	MaxHNSWEfSearch       uint32 = 4096
+)
+
+// VectorIndexDescription holds config specific to vector (ANN) indexes.
+type VectorIndexDescription struct {
+	// Algorithm is the algorithm used to build/search the index.
+	Algorithm VectorAlgorithm
+	// Metric is the distance metric used to compare vectors.
+	Metric DistanceMetric
+	// Dimensions is the number of dimensions of the vectors being indexed.
```
AndrewSisley: todo: This appears to be lacking important documentation found in the internal code:

> and the dimensions must be set, or come from an @embedding on the same field.

Although, on *that* documentation, I am not sure what it is talking about RE the `@embedding` - we dont appear to have a `dimensions` property on that directive.

question: Some of the tests appear to be declaring `Dimensions` to be three, but the property being indexed looks like a simple 2 dimensional array - how does that work?

client/index.go
```diff
@@ -22,16 +23,159 @@ type IndexedFieldDescription struct {
...
 	ID uint32
 	// Fields contains the fields that are being indexed.
 	Fields []IndexedFieldDescription
-	// Unique indicates whether the index is unique.
-	Unique bool
+	// Secondary holds config specific to secondary (scalar/unique/JSON) indexes. Non-nil iff
+	// this is a secondary index.
+	Secondary *SecondaryIndexDescription
+	// Vector holds config specific to vector (ANN) indexes. Non-nil iff this is a vector index.
+	Vector *VectorIndexDescription
```
AndrewSisley: question: Why are `Secondary`, `Vector` and `Vector.HNSW` pointers?

client/index.go
```diff
@@ -22,16 +23,159 @@ type IndexedFieldDescription struct {
 	Descending bool
 }
 
+// IndexKind identifies the kind of an index.
+type IndexKind uint8
```
AndrewSisley, line 30: question: Why is this, and `IndexDescription.Kind()` public?

client/index.go
```diff
@@ -22,16 +23,159 @@ type IndexedFieldDescription struct {
...
+
+const (
+	// DistanceMetricCosine identifies the cosine distance metric.
+	DistanceMetricCosine DistanceMetric = iota
+)
+
+// SecondaryIndexDescription holds config specific to secondary (scalar/unique/JSON) indexes.
+type SecondaryIndexDescription struct {
+	// Unique indicates whether the index enforces uniqueness.
+	Unique bool
```
AndrewSisley: thought: Mentioning because you mentioned the PR size in our one-one, no changes required.

This refactor looks like it would have been a good candidate to remove from this PR.  It is a refactor, and a semver major breaking change.  The main set of changes is a new feature, and so far only looks like a semver minor version change.

client/index.go
islamaliev: > Why was Unique moved in a level?

Because it's secondary-index-specific

> Are Fields, Name, and ID not also config specific to secondary (scalar/unique/JSON) indexes that are not found on primary indexes?

They are common for both (potentially for all) indexes

> What other properties to you think will also live here in the future?

For now nothing is planned.

> This is very definitely a major semver breaking change

I'll check what can be done.

client/index.go
islamaliev: > embedding - we dont appear to have a dimensions property on that directive.

Correct catch: @embedding has no dimensions arg; the phrasing means the embedding model fixes the length. The comment is imprecise and I'd reword it.

client/index.go
islamaliev: Pointers give a nil/non-nil signal that drives Kind(). We will have more indexes in very near future (for examples text indexes like bm25). This is the way to store different index configs under the same index namespace

client/index.go
islamaliev: Agreed. I'll split the Unique refactor out so this PR stays a semver-minor feature

client/index.go
AndrewSisley: > They are common for both (potentially for all) indexes

The only indexes described by `IndexDescription` are secondary indexes though?  What non-secondary indexes are you planning on using `IndexDescription` for?

client/index.go
AndrewSisley: How is `bm25` not a secondary index?

AndrewSisley: How is any index that is not a primary index, not a secondary index?

client/index.go
islamaliev: the planner routes on kind, so it needs to be public. If that's not sufficient justification, it's a fair simplification to reconsider. I can, in theory, change it to a Boolean like `IsVector` But as I mentioned above, we'll have more index types. So that means we would have a Boolean method for all of them.

client/index.go
islamaliev: They are all secondary indexes, more or less. The only non-secretary index is primary index. But secondary indexes are different you cannot just have one category for all of them. Like the ones we had so far are drastically different from vector indexes, and the text based indexes are also different. That's basically why they are under index description, but determined by a pointer. With more detailed type safe configuration.

client/index.go
AndrewSisley: Planner can re-implement this in planner if it needs to.

We are inventing data types that do not need to exist in the public space, this increases the semver protected space and adds noise to the user interfaces, for the sake of internal code.

Please consider this a todo now (and of course please argue against removing this if you feel that way :)).

client/index.go
islamaliev: https://github.com/sourcenetwork/defradb/pull/5096#discussion_r3692170540 
Let's not have different threads discussing the same point :)

client/index.go
AndrewSisley: todo: That's okay, but then we need to rename `Secondary` and `SecondaryIndexDescription` to something that reflects what it is.

question: Why can bm25 and other future indexes not have a 'unique' constraint?

AndrewSisley: suggestion: If the `Secondary` and `Vector` properties cannot co-exist, please strongly consider amending the model to reflect this.  Consider collapsing them into a single interfaced property, similar to `FieldKind`.

islamaliev: > we need to rename Secondary and SecondaryIndexDescription to something that reflects what it is

I don't know any better name for this. This is actually like a well known terminology used for this kind of indexes. Like regular indexes. You can also use scalar index, But I doubt it's better reflecting the essence. There's also inverted indexes, but it's something else. If you have any better idea let me know.

> Why can bm25 and other future indexes not have a 'unique' constraint?

These properties don't exist on such indexes. Maybe there are some other index types that have this property, but I don't know about any of them planned for the nearest future.

jsimnz: Ive thought about this convo for a day or so now and I think I have a cohesive structuring we should move forward with.

All these and other indexing types fall under the umbrella of "Secondary Index" compared to the default "Primary Index" which is effectively just the materialized store in the datastore.

"Secondary Index" is described/represented via the `IndexDescription` type we currently have, extending this to support the current vector index and other types we are working on in parallel (eg bm25/full text, trigram, etc), as well as staying compatible with the current v1 secondary index (field) implementation.

AFAICT we absolutely should surface the `Kind` type publicly, infact it should be explicit on the description, not a computed method on the `IndexDescription`. I dont see the concern of having the new Kind types from the other (https://github.com/sourcenetwork/defradb/pull/5096#discussion_r3691211680) thread since they are pretty fundemental to any secondary index structure/implementation

The hierarchy of index kinds should be:
- Secondary Index `IndexDescription`
  - Ordered / Scalar Index (current field level index we support)
    - Inverted (technically what we currently support)
    - Bitmap (future ordered index I've wanted to implement which falls into this category, can be ignored for this conversation, just highlighting more examples)
  - Vector Index
    - HNSW algo
    - IVF algo
  - FTS (FullTextSearch) Index
    - BM25 algo
  - Trigram (this could be generalized to TextPattern with a Trigram algo but that seems a little pedantic at this point)
  
 As far as implementation is concerned, keeping in my backwards compat, the `IndexDescription` should look like
 
 ```go
type IndexDescription struct {
	// Name contains the name of the index.
	Name string
	// ID is the local identifier of this index.
	ID uint32
	// Unique indicates whether the index is unique.
	Unique bool // must stay here for semver compat, even tho it wont be used on all kinds
	// Kind enum targeting the class of index (as mentioned above: `OrderedIndexKind, VectorIndexKind, FullTextIndexKind, TrigramIndexKind, etc`)
	Kind IndexKind

	// Config contains the kind/algo specific config, for simplicity this really should be an untyped catchall, which the individual implementations should validate/generate. I dont like the enumerated typed pointers, generic types is a non-starter imo.
	Config IndexKindConfig // (a custom type `type IndexKindConfig map[string]any`
}

with the cooresponding `OrderIndexConfig, VectorIndexConfig, FTSIndexConfig, TrigramIndexConfig, etc` which can each validate their own configs/schema.

The `IndexDescription` can validate that the `Kind` matches the `Config` type.

I know theres likely to be some medium-ish pushback on the semi-untyped nature here, but it gives us the most flexibility while we figure out some of this longer term design that we can eventually rework properly in v2. But this gets us far while being pretty easy to reason about for now.

One important call out, the `Unique` field must stay in the current `IndexDescription` for compat reasons.

We can surface a typed config via the GQL SDL directives for DX

islamaliev: why do we need `Kind` field if we can derive it from `Config`?

AndrewSisley: > `type IndexKindConfig map[string]any`

I dislike the suggested map quite a lot, it is as John said a short term hack, and it differs from the existing types that serve this exact same purpose in the description model that `IndexDescription` is a part of.

----

> These properties don't exist on such indexes.

Why do you say thist?  All indexes that can be stored in a key value store can have unique keys no?  Is it more that you feel that the user-simplicity-gain to be had in keeping it in the shared space is outweighed by the complexity of having the option to set it for what you see as lower value use-cases?

> You can also use scalar index. If you have any better idea let me know.

I'm happy enough with this - the problem with naming it 'secondary' index is that it implies that all the other secondary indexes are not secondary indexes, which I see as much more problematic than the possible ambiguity of `ScalarIndex`.  Might be worth asking dex-ex what they think if you are uncomfortable.

islamaliev: > Why do you say thist? All indexes that can be stored in a key value store can have unique keys no? Is it more that you feel that the user-simplicity-gain to be had in keeping it in the shared space is outweighed by the complexity of having the option to set it for what you see as lower value use-cases?

Well, you probably can force indexes to be unique. But for some indexes this is pretty bad idea. Just to give you some idea how that works: For regular indexes that users create, there is a one to one mapping between the value they entered and the index record that's being stored. So if the value matches some existing value, you have the direct correlation error surface to the user. 

With vector index it's entirely different situation. You might have a field that is user-facing that users can set manually or whichever other way and then you have other fields that store the vector values like embeddings derived from the values on those fields. For these derived fields values we create indexes. So now imagine a situation. User creates two documents and gives this index field two different values. But somehow, because the model decided so, it created the embeddings for them or other vectors the same value. This is highly unlikely, but this can happen, especially if you have low dimension vectors. So, if we had this unique constraint, then the user would get the error saying, hey, the indexes collide, so please change the value. But from the user's perspective, this is absolutely stupid because he entered different values into the original field. 

Moreover The index itself is not just a single value that you can read and return. More often than not, an index search requires multiple I.O. Operations because it requires touching different records that are presented as graphs. Or some kind of cluster, or whatever else.

AndrewSisley: > User creates two documents and gives this index field two different values.

This was a great example, thanks Islam.

> But for [common use cases and the vector models we have looked at] this is pretty bad idea.

thought: To me, it seems likely that many, if not most, of the application developers using vector indexes are going to understand them better than we do, and it does not seem implausible to me that there may be legitimate reasons for wanting to do this (lots of users, use-cases, and models).  I also believe that it is quite hard to set this accidentally, and that setting it deliberately is the user's own fault for using models they don't understand well enough; and regardless of why it was made unique, it would be fairly quick to debug and resolve if problematic.

This is however quite a personal design preference though IMO, and I'm happy enough with your choice :)  Thanks for the explanation Islam.

internal/db/vectorindex/hnsw_index.go
```diff
@@ -0,0 +1,97 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package vectorindex
```
AndrewSisley, line 11: question: Why are there two internal hnsw packages?  This one and `index/hnsw`.

thought: We had something similar for CRDTs and it caused a fair bit of grief.

internal/index/hnsw/store.go
```diff
@@ -0,0 +1,136 @@
...
+type Meta struct {
+	EntryPoint NodeID
+	TopLayer   int
+	Empty      bool // true when no entry point has been set yet
+}
+
+// NodeStore is the persistence port. The engine never imports a concrete
+// store; implementations may back this interface with any storage medium
+// (in-memory, a KV store, etc).
+type NodeStore interface {
```
AndrewSisley, line 33: todo: This feels like a large amount of needless complexity.  It only has a single production code implementation.  It is hard to imagine why it would ever want another production code implementation. The production code implementation already supports alternative store implementations (such as in memory, and test shimming), and 'never imports a concrete store'.

Please rework this.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+import (
+	"container/heap"
+	"math"
+	"math/rand"
+	"sort"
+	"sync"
+)
+
+// Graph is a Hierarchical Navigable Small World approximate-nearest-
+// neighbour index. It is safe for concurrent use: mutations (Insert,
```
AndrewSisley: > Graph is a Hierarchical Navigable Small World approximate-nearest-neighbour index.

suggestion: HNSW is the term used throughout the rest of the codebase, and the opening line for `Graph` says that it is exactly that, please can we rename this object to `HNSW` or `HNSWIndex` and save us from having to deal with a second word for the same thing.

internal/index/hnsw/store.go
```diff
@@ -0,0 +1,136 @@
...
+	// IterateNodes iterates over all non-deleted node ids (used for
+	// brute-force baselines / rebuilds). Order is unspecified. Iteration
+	// stops early if fn returns an error, and that error is returned to
+	// the caller of IterateNodes.
+	IterateNodes(fn func(Node) error) error
+}
+
+// memStore is an in-memory NodeStore implementation, safe for concurrent
+// use. It is useful for tests and for standalone/in-RAM use of the graph.
+type memStore struct {
```
AndrewSisley: question: This appears to only be used by tests.  Why is it in the production code?

question: Why is it acceptable to use the memory core-kv store in the production code (via txn), but unacceptable to use memory core-kv for tests or 'standalone/in-RAM use of the graph'?

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+		store:  store,
+		params: params,
+		metric: metric,
+		rng:    rand.New(rand.NewSource(seed)),
+	}
+}
+
+// randomLevel draws a random top layer for a new node, following the
+// exponential-decay level distribution from the HNSW paper:
+// l = floor(-ln(unif(0,1)) * ML).
```
AndrewSisley: todo: Please document why we want to use this decay.

internal/index/hnsw/heap.go
```diff
@@ -0,0 +1,70 @@
...
+// by the neighbour-selection heuristic).
+type candidate struct {
+	id     NodeID
+	dist   float32
+	vector []float32
+}
+
+// minHeap is a min-heap of candidates ordered by ascending distance
+// (nearest first out). Used as the candidate frontier "C" in SEARCH-LAYER.
+type minHeap []candidate
```
AndrewSisley: > minHeap is a min-heap of candidates ordered by ascending distance

No it is not.  Its a slice with a `Less()` function, and that `Less()` function appears to be unused.

The push and pop funcs are also a bit silly - `minHeap` is a slice of the concrete type `candidate` - yet push and pop operate on `any`.

todo: Please delete `minHeap` and `maxHeap`.  `maxHeap` is identical to `minHeap` minus the dead `Less()` func

AndrewSisley (changes requested):
The implementation needs some work before I'm happy with it being merged.

Parts of `index/hnsw` looks unreviewed, and copy-pasted without credit.  If the implementation was taken from somewhere else, please credit the original authors and link to the repo.

internal/keys/datastore_vector_index.go
```diff
@@ -0,0 +1,150 @@
...
+	// Epoch namespaces the entries of a single index build; a rebuild fills a fresh epoch
+	// disjoint from the current one.
+	Epoch uint32
+	// IsMeta is true for the graph's meta singleton key, false for a node entry key.
+	IsMeta bool
+	// NodeID is the node this key addresses. Only meaningful when IsMeta is false.
+	NodeID uint64
+	// hasNodeID distinguishes a node key (NodeID set) from a node-prefix key (no NodeID; used to
+	// iterate over every node entry of the index).
+	hasNodeID bool
```
AndrewSisley: todo: Please document why we need this, and the `if` in `Bytes()`, instead of checking for `0` like all our other keys do.  Or change it so that we dont need this (might need to change NodeID to start at `1`).

internal/index/hnsw/store.go
```diff
@@ -0,0 +1,136 @@
...
+	Vector  []float32  // normalized
+	Layers  [][]NodeID // Layers[l] = neighbour ids at layer l; len(Layers) = node's top level + 1
+	Deleted bool       // tombstone
+}
+
+// Meta is the graph's global state.
+type Meta struct {
+	EntryPoint NodeID
+	TopLayer   int
+	Empty      bool // true when no entry point has been set yet
```
AndrewSisley: question: Why are we storing empty `Meta`s?  Skimming the code it does not appear to have purpose.

internal/db/index.go
```diff
@@ -423,6 +427,155 @@ func (index *collectionSimpleIndex) Delete(
...
+		return err
+	}
+	if !found {
+		if index.building {
+			return nil
+		}
+		return NewErrCorruptedVectorIndex(index.desc.Name, doc.ID().String())
+	}
+
+	return idx.Delete(nodeID)
```
AndrewSisley, line 591: todo: This does not look like it will actually remove the document from the index, due to the soft deleting logic within `index/hnsw`.  If you want this behaviour, please document why `*collectionVectorIndex.Delete(doc)` would only want to soft delete the doc, otherwise, please change the behaviour.

Please make sure there are integration tests for this.

islamaliev:
> Parts of `index/hnsw` looks unreviewed, and copy-pasted without credit. If the implementation was taken from somewhere else, please credit the original authors and link to the repo.

Just to clear it up: the code isn't copied. I wrote it from the [HNSW paper (Malkov & Yashunin),](https://arxiv.org/abs/1603.09320) which is already cited at the top of hnsw.go. I did read a couple of existing Go libraries while learning the algorithm, but I didn't take their code, and in one important spot I did the opposite of what coder/hnsw does (they pick neighbours in a simple way, I used the paper's smarter method, which matters a lot for result quality).

Happy to add a note at the top of the package pointing at the paper and saying it was inspired by those libraries, if that helps. And if some parts look off to you, point me at them and I'll go through them.

internal/db/vectorindex/hnsw_index.go
islamaliev: They're two different things, not a duplicate.

`internal/index/hnsw` is just the algorithm on its own. It doesn't import anything from DefraDB, only the standard library, so in principle it could be pulled out into its own repo and open-sourced. It talks to storage through a small interface and knows nothing about transactions or our key/value store.

`internal/db/vectorindex` is the glue between DefraDB and vector indexes in general. It holds the Index interface that the rest of DefraDB uses (insert, delete, search by id), and picks which implementation to use based on the index config. The files named `hnsw_*.go` in there are the bit that wires the pure algorithm to our datastore. That's why the engine is only imported in those two files.

The reason for keeping them apart: one answers "how does HNSW work" and stays portable, the other answers "how does DefraDB use a vector index, whichever kind." It's also the abstraction we'd use to add another engine later, for example a C++-based one behind a build tag, without touching the planner or storage. Only that glue file changes.

internal/index/hnsw/store.go
islamaliev: I hope my reasoning [here](https://github.com/sourcenetwork/defradb/pull/5096#discussion_r3711437622) answers your concern.

internal/index/hnsw/graph.go
islamaliev: Well, it is technically a graph. And it's under HNSW package. So i think i'd say the naming is fine. But I will rename it no problem. I don't have a strong preference here.

internal/index/hnsw/graph.go
islamaliev: will do

internal/index/hnsw/heap.go
islamaliev: Less isn't actually unused, it's called by Go's container/heap package under the hood. But you're right that the two heaps were near-duplicates, so I've merged them into one type with a flag

internal/index/hnsw/store.go
islamaliev: I will move it

internal/keys/datastore_vector_index.go
islamaliev: yeah, I checked again and found a way to live without it. So I will remove it

internal/index/hnsw/store.go
islamaliev: I removed it and added bool to the return param of `GetMeta` so now it's consistent with `GetNode`

internal/db/index.go
islamaliev: Yeah, this is a soft delete on purpose. The node gets marked deleted but stays in the graph. The reason it stays is that other nodes might route through it to reach their own neighbours, so if we ripped it out right away we'd have to go and repair all those links in the same write, and that repair-on-delete path is exactly where a lot of vector DBs have historically had bugs. So we mark it deleted instead and leave the links alone.

The important bit is that search skips deleted nodes, so once you delete a doc it stops showing up in similarity results straight away, in the same transaction. From a query point of view the doc is gone. The only downside of leaving the old node around is that recall can slowly drift as tombstones pile up, which is a quality thing, not a correctness thing.

Cleaning up the tombstones for real (unlinking them and rebuilding that part of the graph) is a background pass we've got planned but haven't built yet. I'll write it up as a proper follow-up issue and link it here so it's tracked rather than just living in a comment.

client/index.go
jsimnz: > All indexes that can be stored in a key value store can have unique keys no?

Just double clicking on this, theres many different kinds of indexes where "unique" just simply doesnt have any semantic meaning. The key/value store part is independent of the unique property on indexes.

My only statement about keeping unique on the top level `IndexDescription` is purely semver.

> I dislike the suggested map quite a lot, it is as John said a short term hack, and it differs from the existing types

Yea I debated about dozen times actually sending the message with that suggestion because I knew it would be met with pushback. We can do something original to what Islam is suggesting with the typed index config behind a pointer/optional type, I just feel like it would get A) busy B) put us into a semver corner at a time we *might* still experiment a bit with the index framework as a whole (but maybe its a good thing to just commit and lock in ;) )

> why do we need Kind field if we can derive it from Config?

I see little to be gained by implied type from other types. The index `kind` is the foundational piece of data to understand anything about the `IndexDescription` and its sub index type descriptions. Theres no way to enforce at the type level that a single struct field in a set of fields can be set at once 

Eg if we have: 
```go
Vector *VectorIndexDescription
Fts *FTSIndexDescription
Trigram *TrigramIndexDescription
etc...
```

Theres no way to say that *only* `Vector` or `FTS` or `Trigram` can be populated. So there is an easy-ish edge case to hit where either during serialization, or schema Patching, or anywhere we accidentally populate two sub type description fields. Theres some clever ways around this probably, but all that can be trivially resolved by the single `Kind` field that is explicit and obviously can only hold a single value, it is the sole authority, nothing computed or verified, or gated during read and write, just a field value. The entire set of `*Description` types in the `client` package are intended to be reasonably transparent and clear. Theyre public, many (most?) of them are exposed through the patch API, are serialized verbatim into the system store. I can imagine a few instances where some computed fields of lesser importance could absolutely exist as a read only method, but in the case of indexes the most important field is `Kind` (*given the current type system we have designed for the `Description` structs*)

> I'm happy enough with this - the problem with naming it 'secondary' index is that it implies that all the other secondary indexes are not secondary indexes, which I see as much more problematic than the possible ambiguity of ScalarIndex. Might be worth asking dex-ex what they think if you are uncomfortable.

Agreed, we can't scope the current field index naming to "Secondary" as it is all kinds of confusing. Either `ScalarIndex` or `OrderedIndex`. Theres precedent for both in the DB world, I think `Ordered` is used *slightly* more.

internal/db/vectorindex/hnsw_index.go
AndrewSisley: >  for example a C++-based one behind a build tag

Why would we ever want to do this?

internal/db/index.go
AndrewSisley: Thanks for the great explanation here Islam!

question: Do you think you will implement this tombstone cleanup this sprint?

question: To me it looks like deleting a document that is hnsw indexed will not free the storage consumed by the index (until a future implementation with cleanup) - is this true?

internal/index/hnsw/heap.go
AndrewSisley: Oof I dislike quite a lot of things about what I saw in that (container/heap) package...

todo: The documentation of `minHeap` and `maxHeap` is still incorrect. The types themselves are not ordered by anything beside the normal slice indexing.  If and when `heap.Push` is used to call `minHeap.Push`, `heap.Push` will insert items into the slice in the order described.

suggestion: I completely missed the `_ heap.Interface = (*minHeap)(nil)` at the bottom of this file - it is in a location I did not expect.  Please move these to a more prominent location, such as imediately following each type declaration.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+						heap.Pop(&results)
+					}
+				}
+			}
+		}
+	}
+
+	out := make([]candidate, len(results))
+	copy(out, results)
+	sort.Slice(out, func(i, j int) bool { return out[i].dist < out[j].dist })
```
AndrewSisley, line 343: todo: `container/heap` looks like it inserts items into `results` in the order that you are sorting.  Please either remove this sort, or document why it is needed. 

I think I saw the same sort somewhere else, please also apply this comment to there if it exists.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+//
+// Deleted (tombstoned) nodes are traversed (their neighbours are still
+// explored, keeping the graph connected) but are never added to the
+// result set W, per the package's tombstone-deletion strategy.
+//
+// The returned candidates are sorted nearest-first.
+func (g *Graph) searchLayerMulti(query []float32, entryPoints []candidate, ef, layer int) ([]candidate, error) {
+	visited := make(map[NodeID]struct{})
+	var candidatesHeap minHeap
+	var results maxHeap
```
AndrewSisley: todo: Please document why these heap variables need to be continuously ordered.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+func (g *Graph) addLink(from, to NodeID, layer, mmax int) error {
+	node, ok, err := g.store.GetNode(from)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return nil
+	}
+
+	for len(node.Layers) <= layer {
```
AndrewSisley: question: How can this happen?  And is it safe to add default-value `NodeID` inbetween the `from` and `to` layers like this?

todo: Please document the answers to these questions here if the code remains as-is.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+		node, ok, err := g.store.GetNode(nearest.id)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
+
+		var neighbours []NodeID
+		if layer < len(node.Layers) {
```
AndrewSisley: suggestion: This looks like it can be checked on line one of `searchLayerMulti` instead of here - including this in the loop here is making already very complicated code more complicated than it needs to be.
```go
if layer >= len(node.Layers) {
  return nil, nil
}
```

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+// searchLayerMulti runs SEARCH-LAYER (Algorithm 2), the ef-bounded greedy
+// search of a single layer, starting from a set of entry points. Distances
+// on the supplied entryPoints are recomputed against query as needed.
+//
+// Deleted (tombstoned) nodes are traversed (their neighbours are still
+// explored, keeping the graph connected) but are never added to the
+// result set W, per the package's tombstone-deletion strategy.
+//
+// The returned candidates are sorted nearest-first.
+func (g *Graph) searchLayerMulti(query []float32, entryPoints []candidate, ef, layer int) ([]candidate, error) {
```
AndrewSisley: question: `searchLayerMulti` only ever appears to be called with a single `entryPoint`, why is `entryPoints` a slice?

EDIT: It is not always a single item, `Insert` calls this with multiple, but that itself is confusing me as they are from different layers.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+		// Add bidirectional links and shrink neighbours that exceed mmax.
+		for _, sel := range selected {
+			if err := g.addLink(sel.id, id, layer, mmax); err != nil {
+				return err
+			}
+		}
+
+		// Candidates for the next (lower) layer's entry points are the
+		// full candidate set found at this layer.
+		entryPoints = w
```
AndrewSisley, line 149: question: Why are we preserving the `entryPoints` from previous layers here?  This is causing multiple layers to be fed into each `searchLayerMulti`, which it looks like is incorrect.

suggestion: Please consider expanding the documentation here.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
+
+		var neighbours []NodeID
+		if layer < len(node.Layers) {
+			neighbours = node.Layers[layer]
```
AndrewSisley: question: We appear to be checking everything at the specified layer anyway - so what is the benefit of this function being provided an `entryPoints` param, vs having this function simply start from the first node at this layer?

suggestion: Consider documenting this somewhere, maybe on the function signature.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+
+	for _, ep := range entryPoints {
+		n, ok, err := g.store.GetNode(ep.id)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
+		if _, seen := visited[ep.id]; seen {
```
AndrewSisley, line 264: todo: This looks like dead code - the only place I spotted that calls `searchLayerMulti` with multiple `entryPoints`, sources each entryPoint from a different layer - I assume a node cannot exist in multiple layers?

If keeping this check, please document why it is needed.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+		if err := g.store.PutNode(Node{ID: id, Vector: v, Layers: layers}); err != nil {
+			return err
+		}
+		return g.store.PutMeta(Meta{EntryPoint: id, TopLayer: topLevel, Empty: false})
+	}
+
+	// Descent: greedily walk down from the entry point at the top layer
+	// to layer topLevel+1, keeping only the single closest node found so
+	// far (ef=1 search at each layer).
+	entry, _, err := g.store.GetNode(meta.EntryPoint)
```
AndrewSisley: todo: `entry` will be `nil` if `_` is `false`, please either handle this, or document why it is safe to ignore it.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+// distance to q, returning them as candidates (unsorted).
+func (g *Graph) candidatesFromIDs(q []float32, ids []NodeID) ([]candidate, error) {
+	out := make([]candidate, 0, len(ids))
+	for _, id := range ids {
+		n, ok, err := g.store.GetNode(id)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
```
AndrewSisley, line 220: todo: Please document why we do not want to error in this case - it looks to me like this could only happen in case of a programming mistake.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+	var candidatesHeap minHeap
+	var results maxHeap
+
+	for _, ep := range entryPoints {
+		n, ok, err := g.store.GetNode(ep.id)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
```
AndrewSisley, line 260: todo: Please document why we do not want to error in this case - it looks to me like this could only happen in case of a programming mistake.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+			break
+		}
+		heap.Pop(&candidatesHeap)
+
+		node, ok, err := g.store.GetNode(nearest.id)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
```
AndrewSisley, line 299: todo: Please document why we do not want to error in this case - it looks to me like this could only happen in case of a programming mistake.

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+				continue
+			}
+			visited[nb] = struct{}{}
+
+			nbNode, ok, err := g.store.GetNode(nb)
+			if err != nil {
+				return nil, err
+			}
+			if !ok {
+				continue
```
AndrewSisley, line 319: todo: Please document why we do not want to error in this case - it looks to me like this could only happen in case of a programming mistake.

internal/index/hnsw/store.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// NodeStore is the persistence port. The engine never imports a concrete
+// store; implementations may back this interface with any storage medium
+// (in-memory, a KV store, etc).
+type NodeStore interface {
+	GetNode(id NodeID) (Node, bool, error)
```
AndrewSisley: suggestion: I have not spotted the benefit in having the second return param from this function - there are todo's asking for documentation in the places that handle this.  Please consider removing it if you are unable to provide good reasons in those locations.

internal/db/vectorindex/hnsw_index.go
islamaliev: because C++ is fast. But we will have to benchmark it. There are quite some good C++ battle-tested libraries out there

AndrewSisley: nitpick: C++ isn't really much faster than Go, it just offers more tools which can be used to further optimise code.

question: If we chose to use a faster C++ library instead of the Go implementation in this PR, why would we continue to maintain the Go implementation?

question: If there are better implementations (in any language) than the one in this PR, why are we bothering with the one in this PR at all?  Why does this PR not simply consume those better libraries?

internal/index/hnsw/graph.go
islamaliev: we need sort here as heap sort is only partially sorted

internal/index/hnsw/graph.go
islamaliev: done

internal/index/hnsw/graph.go
islamaliev: The layer for every note is chosen randomly. So the one we linking with might have different amount of layers. Having empty layers is not a problem. I added a comment

internal/index/hnsw/graph.go
islamaliev: added a check, covered with a unit test

internal/index/hnsw/graph.go
islamaliev: added a comment

internal/index/hnsw/graph.go
islamaliev: added a comment

internal/index/hnsw/graph.go
islamaliev: added a comment

internal/index/hnsw/graph.go
islamaliev: added a comment

internal/db/vectorindex/hnsw_index.go
islamaliev: we don't know yet if it's better or faster. It requires additional research and benchmarking. With the interface in place we can play around with different options.

internal/index/hnsw/heap.go
islamaliev: I merged minHeap and maxHeap into one.

moved interface check up.

internal/db/index.go
islamaliev: If a doc is deleted, tombstone remains. 
Not sure if this sprint I will have time. Will have to see first if there are other more pressing things for vector indexes to handle.

internal/index/hnsw/graph.go
islamaliev: node is different each time round the loop, and nodes have different heights, so this check has to stay per-node. Moving it to the top and returning nil would end the whole search early and drop results, when a short node just means "nothing to add at this layer, move on

internal/index/hnsw/graph.go
islamaliev: The slice is intentional. After searching a layer, we use the whole set of neighbours we found there as the starting points for the next layer down, so that layer gets multiple entry points.
They're not from different layers, a node exists on every layer up to its own height, so nodes found at layer L are valid entry points at L-1. The single-entry-point case is just the descent step.

internal/index/hnsw/graph.go
islamaliev: The slice is intentional. After searching a layer, we use the whole set of neighbours we found there as the starting points for the next layer down, so that layer gets multiple entry points. They're not from different layers, a node exists on every layer up to its own height, so nodes found at layer L are valid entry points at L-1. The single-entry-point case is just the descent step. I changed comments a bit

internal/index/hnsw/graph.go
islamaliev: It's not checking everything at the layer, it's a greedy walk: it only visits nodes reachable by following edges from the entry points, and stops early once the result set is full. So the entry points decide where the walk starts and which part of the graph it explores. Starting from an arbitrary first node would explore the wrong neighbourhood and lower recall. The whole point of HNSW is starting the search from good entry points handed down from the layer above.

internal/index/hnsw/graph.go
islamaliev: a node does exist on every layer up to its height, so the entry points here aren't from "different layers" in a way that matters. And you're right that with the current callers this dedup never fires,  they pass unique ids. But it's a cheap guard that keeps the function correct for any entry-point set, not just today's two callers, so I'd rather keep it than have it silently double-count if a future caller passes a duplicate. I'll add a comment.

internal/index/hnsw/store.go
islamaliev: Every caller uses it now, and they react differently: the entry-point load errors if the node is missing (torn graph), the traversal skips missing nodes (a reclaim pass can remove one a backlink still names), and search returns empty on a missing entry point. Without the bool they couldn't tell a missing node from a zero-value one, which is exactly what the tombstone-skip logic relies on. The one caller that ignored it before now uses it. So I'd keep it.

internal/index/hnsw/graph.go
AndrewSisley: > It's not checking everything at the layer,

todo: Please double check this, and document where this is happening - at the moment it very much looks like you are looping through every single `neighbours`, fetching it, and calculating distance for it.

question: Is your `if results.Len() < ef` in the wrong place, or missing a `break`?

internal/index/hnsw/graph.go
```diff
@@ -0,0 +1,356 @@
...
+
+		// Add bidirectional links and shrink neighbours that exceed mmax.
+		for _, sel := range selected {
+			if err := g.addLink(sel.id, id, layer, mmax); err != nil {
+				return err
+			}
+		}
+
+		// Candidates for the next (lower) layer's entry points are the
+		// full candidate set found at this layer.
```
AndrewSisley: suggestion: This comment provides a verbose description of what `entryPoints = w` does, but it does not explain why we want to do this, please consider reworking a bit.

internal/index/hnsw/graph.go
AndrewSisley: Ah I missed that - I think I'm too used to `node` being the (constant) machine Defra is running on.  I have no good suggestions, but if another word comes to mind it might be worth renaming to avoid this conceptual clash.

Thanks Islam

internal/index/hnsw/graph.go
AndrewSisley: question: Why are the heaps only partially sorted?  Why is a partial-sort useful?

todo: Please document that this is required *because* `results` is only partially sorted.  

suggestion: Please also consider documenting this partial sortedness on the heaps, and why it is useful.

internal/db/vectorindex/hnsw_index.go
AndrewSisley: And you think the cost of the complexity and indirection of the interface is lower than the cost*probability of swapping one internal concrete type for another?

internal/index/hnsw/store.go
AndrewSisley: todo: Please don't forget to move it :)

internal/index/hnsw/store.go
AndrewSisley: It does not - that conversation was centered around being able to swap the engine out.  This interface here is a very thin wrapper around core kv - which is already interfaced to death, and I can not spot any possible reason for ever wanting the option to swap core-kv out for a reimagined corekv implementation/interface, only for hnsw.

client/index_test.go
```diff
@@ -127,3 +129,70 @@ func TestCollectIndexesOnField(t *testing.T) {
 		})
 	}
 }
+
+func TestIndexDescription_UniqueIndex_RoundTrips(t *testing.T) {
```
AndrewSisley: suggestion: I do not think testing the deault json marshaller is useful, I suggest removing these new tests.

AndrewSisley (approved):
LGTM

internal/index/hnsw/graph.go
islamaliev: This is sort of known property of the heapsort. The only thing it guarantees is the first value to pop is either the smallest one or the greatest one. But I'll add a comment.

internal/db/vectorindex/hnsw_index.go
islamaliev: I believe so, yes. I also think the algorithm should be implemented in a way that allows to reuse it for different use cases. And making it dependant on defra-specific code I consider a bad practice

internal/index/hnsw/store.go
islamaliev: Thanks for reminding :)

internal/index/hnsw/graph.go
islamaliev: You're right that the inner loop looks at every neighbour of each node it pops, and fetches + measures each one, that part is neighbours × M, and it's meant to. What it doesn't do is scan the whole layer: it only ever reaches nodes connected to the frontier, so the cost is (nodes visited × M), not N. My earlier "never scans the layer" was true but glossed over the per-node neighbour work. I've documented it properly now.

AndrewSisley: I see, thanks Islam.

I think my confusion here is perhaps driven by two naming issues.  First is `node` and the baggage that comes with that (noted elsewhere, no good alternative suggested), the second is `node.Layers` - a layer to me transcends a single node and its children:
```
Layer 1:      A1          B1
            |   |        |   |
Layer 2:   C2  D2        E2  F2     
```
In this case, `A1.Layers[1]` would include C2...F2.

But the way you responded above, it sounds more like it would only be C2..D2?

suggestion: Please consider renaming `node.Layers` to `Children`, or `ChildLayers` or similar - so that it is clearer that this is scoped only to the node.

client/index_test.go
islamaliev: removed

internal/index/hnsw/graph.go
islamaliev: updated commets

internal/index/hnsw/graph.go
islamaliev: You're right that Layers reads like the global layer concept when it's actually just this node's own neighbour lists, one per layer it lives on. I'd go with `Neighbours` rather than `Children`: the edges are bidirectional so there's no parent/child relationship, and node.Neighbours[l] reads clearly as "this node's neighbours at layer l." I'll rename it and update the doc comment.

PR #5097, opened 2026-07-29
feat: Add filtered collection truncate

### LINKED ISSUE #5117: Support filtered collection truncate
## Problem

DefraDB supports replicated soft deletion of selected documents and local hard truncation of an entire collection, but it cannot permanently remove only the documents matching a filter. Retention workflows need to propagate tombstones and then reclaim local storage for the expired subset without introducing another delete operation.

## Proposed behavior

- Add an optional filter to the existing collection truncate operation.
- Preserve the existing collection-wide behavior when no filter is provided.
- Remove matching document values, heads, indexes, ID mappings, primary markers, signatures, encryption blocks, and searchable-encryption records.
- Optionally prune history blocks that are not shared by another document.
- Expose filtered truncate through the Go, HTTP, CLI, and C APIs and through `truncate_<Collection>` GraphQL mutations.
- Reuse the existing truncate authorization.

## Constraints

- Truncate is local and does not replicate.
- Callers should propagate soft-delete tombstones first when peers must not reintroduce the data.
- History pruning is rejected for branchable collections until collection-DAG pruning semantics are defined.
- Filtered truncate runs in bounded chunks and may be partially applied if a later chunk fails.
- Filtered truncate cannot run inside a caller-supplied transaction.
- A GraphQL truncate must be the only mutation field in the operation.

iverc:
## Relevant issue(s)

Resolves #5117

## Description

Adds an optional filter to the existing collection truncate operation so callers can permanently remove matching documents from the local node without introducing another delete command. Truncating without a filter keeps the existing collection-wide behavior.

Filtered truncate removes the selected document data and associated local state, including heads, indexes, ID mappings, signatures, encryption blocks, and searchable-encryption records. The optional `pruneHistory` setting also removes history blocks that are not shared by another document.

The filter is supported by the Go, HTTP, CLI, and C APIs, and GraphQL now exposes `truncate_<Collection>` mutations. No new dependencies are added.

Limitations:

- Truncate is local and does not replicate. Propagate soft-delete tombstones first when peers must not reintroduce the data.
- History pruning is rejected for branchable collections until collection-DAG pruning semantics are defined.
- Filtered truncate runs in bounded chunks and may be partially applied if a later chunk fails.
- Filtered truncate cannot run inside a caller-supplied transaction. GraphQL truncate must be a standalone mutation.

## Tasks

## How has this been tested?

- Unit coverage for filtering, storage cleanup, shared history, indexes, ID mappings, signatures, encryption, searchable encryption, transaction chunking, and LevelDB behavior.
- Integration coverage for collection and GraphQL truncate flows, DAC authorization, and searchable encryption.
- Race tests for the changed database and GraphQL paths.
- Repository lint and generated documentation checks.

- MacOS


iverc:
Final scale and correctness verification:

- The permanent 500-document regression first runs the same purge in one caller transaction and confirms Badger rejects it with "Txn is too big". The bounded implicit path then purges all 500 successfully.
- The permanent ownership test confirms the new presence check reads one owner key, while the previous full-list path reads all 500 owner mappings.
- An ad hoc 5x1s benchmark with 1,000 owners on an Apple M5 Max measured a median 1.47us, 1,792 B, and 16 allocs for the new first-owner check versus 52.9us, 52,976 B, and 1,027 allocs for the previous full scan: about 36x faster, 30x less memory, and 64x fewer allocations.
- Added coverage for two documents created in one caller transaction sharing a block, then purging one while retaining the shared block for the survivor.

Verified internal/db and internal/db/id package tests, the purge integration matrix, focused race tests, and golangci-lint with 0 new issues.

iverc:
CI note:

- Coverage passed 36/37 jobs. The remaining LevelDB failure matches the latest `develop` run exactly: `TestTxn_DeleteIndex_WithCommit_Succeeds` times out waiting for an index drop, then `TestColVersionUpdateReplaceIsActive_GetCollectionsWithTxn` hangs until the 10-minute package timeout.
- The macOS run failed the exact same six tests as the latest `develop` run: five system-keyring checks and `TestP2PAdd_WithP2PCollectionOnLastNodeInNodeChain_ShouldPropagateUpdate` timing out while waiting for a merge event.
- The vulnerability workflow is also failing on `develop`; its expected vulnerability count is stale relative to the current dependency scan.

Base coverage: https://github.com/sourcenetwork/defradb/actions/runs/30392652964/job/90391332793
Base macOS: https://github.com/sourcenetwork/defradb/actions/runs/30392652787/job/90387826797
Base vulnerability check: https://github.com/sourcenetwork/defradb/actions/runs/30392652739

cli/collection_purge_docs.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2025 Democratized Data Foundation
```
shahzadlone: todo: Fix copy-paste typo for year on license headers.

```suggestion
// Copyright 2026 Democratized Data Foundation
```

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2025 Democratized Data Foundation
```
shahzadlone: todo: Fix copy-paste typo for year on license headers.

```suggestion
// Copyright 2026 Democratized Data Foundation
```

cli/collection_purge_docs.go
iverc: Fixed in 5755ac68a.

internal/db/collection_purge.go
iverc: Fixed in 5755ac68a.

docs/website/references/cli/defradb_client_document_purge.md
```diff
@@ -0,0 +1,51 @@
...
+### Synopsis
+
+Permanently remove a set of documents by DocID from the local node, including all
+datastore values and headstore entries. When --prune-history is set, it also removes
+reachable blockstore blocks that are no longer owned by another document.
+
+Unlike the soft-delete performed by the delete command, this operation is irreversible and
+does not propagate to other nodes in the peer network.
+
+```
```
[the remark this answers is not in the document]
iverc: These CLI reference files are generated by cmd/genclidocs, and the other generated command docs use the same fences. Editing this file directly would be overwritten and fail the documentation generation check, so I am leaving it unchanged.

internal/db/collection_truncate.go
```diff
@@ -486,21 +484,64 @@ func (c *collection) deleteBlocks(
 	systemstore corekv.ReaderWriter,
 	docID string,
 	currentCid cid.Cid,
+	prunedOwners map[string]struct{},
 ) error {
 	blockstore := datastore.NewMultistore(c.db.rootstore, c.db.lockSet, c.db.blockStoreChunkSize).Blockstore()
 
+	// Keep immutable DAG reads out of the purge transaction's conflict set.
+	readTxn, err := c.db.NewTxn(true)
```
shahzadlone: question/suggestion: Should we avoid opening this independent read transaction while another transaction owns the backend? I wonder if on LevelDB we might get timeouts where `PurgeByDocIDs(..., true)` might block here trying to open `readTxn` while its outer write transaction is open, while the existing Truncate path opens this read transaction and then blocks its direct blockstore delete until the deferred read transaction closes.

perhaps we use a backend-capability-aware or transaction-free read path, and preserve the previous truncate behavior where necessary, and add LevelDB regression tests for both targeted pruning and Truncate.

internal/db/collection_index.go
```diff
@@ -361,23 +362,29 @@ func (c *collection) deleteIndexedDocWithID(
 	if err != nil {
 		return err
 	}
+	if includeDeleted {
+		_, isDeleted, err := c.exists(ctx, primaryKey)
```
shahzadlone: todo: I am not sure if we should use c.exists to decide whether purge index cleanup is needed. exists applies document-read authorization using the identity in ctx, while PurgeByDocIDs authorizes the destructive operation with the options identity.

A node administrator can legitimately have truncate permission without document-read permission.

Can be easy to reproduce with a private indexed document. after the owner soft-deletes it, the administrator’s purge would see false, false here, loads the deleted document, and reports a corrupted index because soft deletion already removed the index entry.

Maybe use an internal/raw deletion-status lookup after the node-level authorization and test both the document owner and a node administrator lacking document-read access.

Assuming I understand the use case correctly? Please feel free to push back if this is the desired behaviour.

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
...
+			DocShortID:        docShortID,
+		}
+		if err := c.hardDeleteDatastorePrefix(ctx, prefix); err != nil {
+			return err
+		}
+	}
+
+	systemstore := datastore.NewMultistore(c.db.rootstore, c.db.lockSet, c.db.blockStoreChunkSize).Systemstore()
+	if pruneHistory {
+		if err := c.hardDeleteDocumentBlocks(ctx, systemstore, docShortID, prunedOwners); err != nil {
```
shahzadlone: todo: Perhaps we should handle the collection-level DAG before pruning document blocks in a branchable collection. Branchable writes create a HeadstoreColKey root whose block links to the document composite (root in internal/db/p2p/sync_branchable_col.go:302-319).

as it stands you can possibly purge where the collection head and collection block remain, but the linked document block is removed. The retained and advertised DAG is therefore no longer walkable.

either reject history pruning for branchable collections until safe semantics exist, rewrite/tombstone the collection DAG, or retain every block reachable from its collection roots.

Also please consider adding a test that walks every retained collection head and exercises branchable synchronization.

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/sourcenetwork/defradb/client/options"
+	"github.com/sourcenetwork/defradb/errors"
+	"github.com/sourcenetwork/defradb/internal/datastore"
+	"github.com/sourcenetwork/defradb/internal/db/id"
+	"github.com/sourcenetwork/defradb/internal/keys"
+	"github.com/sourcenetwork/defradb/internal/utils"
+)
+
+// Keep implicit purge transactions below the store's size limit.
+const purgeChunkSize = 100
```
shahzadlone: question: Is this based on a document count? if so are we happy as this being the transaction-size bound. One document can have more history blocks and owner mappings than the store permits in one transaction i think. With a 2 MiB Badger memtable, one document with 600 updates makes the implicit `PurgeByDocIDs(..., true)` return Txn is too big. Because the loop cannot subdivide one document, retrying can never make progress.

Perhaps we can try to bound work by bytes/blocks and make history deletion resumable below the document level, or use another resumable design. The regression test should use one worst-case document, not only many small documents.

internal/db/collection_truncate.go
shahzadlone: question: Would purging one signed document deletes a surviving shared signature?

suggestion: Should we treat a signature link as part of its parent’s ownership/reachability. Signature CIDs never receive SetBlockDocIDMapping (correct me if i am wrong) so new writes and merges only map DAG and encryption CIDs.

Can possibly test this with two Ed25519-signed documents sharing an identical field block.

basically i am wondering here if purging one document correctly retains the field block because the other document owns it, but deletes the shared signature here because the signature appears ownerless, breaking verification of the surviving block.

perhaps can track metadata-child ownership or base deletion on retained-parent reachability, and add a signed shared-block regression test?

shahzadlone: todo: Encryption blocks are stored in txn.Encstore() (internal/core/block/store.go:165), but deleteBlock only deletes from the normal blockstore (from what i see so far).

Can test this by seeing if we can reproduce an encrypted document whose encryption CID is still present in Encstore after PurgeByDocIDs(..., true) succeeds, i think we can. I also noticed that Its owner mapping has meanwhile been removed, leaving an orphan that is harder to discover later.

Thinking if we should maybe be deleting unshared encryption CIDs from Encstore and add a regression test asserting both the normal DAG and encryption-key store are clean?

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Index entries are keyed by the document's field values, so they must be deleted before
+	// the datastore prefixes holding those values are removed.
+	if err := c.deleteIndexedDocWithID(ctx, docID, true); err != nil {
+		return err
+	}
+
+	// InstanceType sits between CollectionShortID and DocShortID in the encoded key, so we
+	// must include it in the prefix; a key without InstanceType matches nothing.
+	for _, itype := range []keys.InstanceType{keys.ValueKey, keys.PriorityKey, keys.DeletedKey} {
```
shahzadlone: question: Shouldnt the primary document marker and searchable-encryption artifacts also be removed? I feel like they will survive purge at the moment.

todo: For searchable enc `keys.DatastoreSE` records are stored separately as `/se/<collection>/<index>/<tag>/<DocID>` roughly and are not matched by the value, priority, or deleted prefixes here. A document-specific DatastoreSE record remains after a successful purge, allowing encrypted-index state to continue referencing the purged DocID (which we seem to promise to remove all datastore values in docs so should fix).

Also should remove all applicable SE artifacts for the canonical DocID and its aliases in the same transaction, with an end-to-end encrypted-index regression test.

todo: For primary doc can delete the exact PrimaryDataStoreKey{CollectionShortID: shortID, DocShortID: docShortID} as well. 

The /collection/pk/doc marker is outside all three CRDT prefixes and i think survives the purge (correct me if i am wrong). 

Can even do  test to see if you can reproduce both the remaining markers and its user-visible consequence, For example i think if you do a purging action on the collection’s only document, then DB.DeleteCollection still would likly fail with  something like "cannot delete collection that has documents" because `collectionHasDocumentsReader` scans this primary-key prefix.

Better to delete the marker before removing the ID mappings and test both marker absence and successful collection deletion.

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
...
+	opts ...options.Enumerable[options.TruncateCollectionOptions],
+) error {
+	ctx, _, hasCallerTxn := getTxnAndSetCtxForCollection(ctx, c)
+
+	ctx, span := tracer.Start(ctx)
+	defer span.End()
+
+	opt := utils.NewOptions(opts...)
+
+	if err := c.db.checkNodeAccess(ctx, opt.Identity, acpTypes.NodeTruncateCollectionPerm); err != nil {
```
shahzadlone: question: Why is Purge using the same permission type as Truncate? is this desired or was discussed already? 

todo: IMO should add proper new NAC permission and properly add all the tests like all other existing NAC tests (can help you with that :))

internal/db/collection_purge.go
```diff
@@ -0,0 +1,192 @@
...
+// PurgeByDocIDs permanently removes local state for the given documents.
+// When pruneHistory is true, it also removes reachable blocks with no other owner.
+//
+// Without a caller transaction, each chunk commits independently. With one, the caller owns
+// the commit and the full purge must fit in that transaction.
+func (c *collection) PurgeByDocIDs(
+	ctx context.Context,
+	docIDs []client.DocID,
+	pruneHistory bool,
+	opts ...options.Enumerable[options.TruncateCollectionOptions],
```
shahzadlone: todo: Define PurgeByDocIDsOptions instead of exposing TruncateCollectionOptions. Otherwise future truncate-only options automatically become part of this API even when they do not apply to targeted purging.

shahzadlone (changes requested):
Thanks for this PR, I have done the first round of review and left some comments. I think there is some work that needs to be done before we can merge this. 

In addition to the comments I noticed a pattern where you piggy-backed/used some truncate code for the purge code, is this purposeful? For example rather than defining a new NAC ACP permission using the truncate one or also the use of  `options.TruncateCollectionOptions` over a tailored option for  Purge. Please find and sort any and all such actions that I may not have caught where this same pattern might be repeated. 

One additional `nitpick`: The parent collection command synopsis in cli/collection.go still lists add, describe, patch, set-active, delete, and truncate, but omits purge-docs. The regenerated parent documentation preserves that omission. Please include the new command so it is discoverable from the parent help.

shahzadlone:
suggestion: I don't see a linked issue and the PR description is not using the DefraDB PR description template style, will appreciate if you can adhere to that please :). 

More info: https://github.com/sourcenetwork/defradb?tab=contributing-ov-file#-link-with-relevant-issues

internal/db/collection_purge.go
iverc: agreed, purge has its own options type now across all client surfaces

iverc:
Addressed the review in dba27a222 and 48ee8e9eb:

- Added dedicated purge permission and options instead of reusing truncate contracts.
- Audited and updated Go, HTTP, CLI, C, JS, mocks, and test wrappers.
- Added purge-docs to the parent collection synopsis and regenerated CLI docs.
- Added LevelDB, large-history, branchable, DAC/NAC, shared signature/encryption, marker, alias, and two-node searchable-encryption regressions.

Local validation is green across Go, HTTP, CLI, C, and WASM purge flows, plus both pinned lint configurations.

internal/db/collection_truncate.go
iverc: yep, removed the extra read transaction. added LevelDB tests for purge and truncate too

internal/db/collection_index.go
iverc: yep, purge uses the raw deletion state now, so a node admin doesnt need document read access. added coverage for that case

iverc:
Addressed the review round in `dba27a222` and `48ee8e9eb`:

- removed the remaining purge coupling to truncate options and permissions
- added a dedicated `purge-document` permission and `PurgeByDocIDsOptions`
- removed the nested LevelDB read transaction and added regressions for purge and truncate
- made implicit history pruning resumable for a single large document
- rejected branchable history pruning until its collection-DAG semantics are defined
- preserved shared blocks and signatures, and cleaned ownerless encryption blocks
- removed primary markers and canonical, alias, and unmatched searchable-encryption records
- added DAC, NAC, LevelDB, shared-history, signed-document, encrypted-document, 3,000-update, and end-to-end searchable-encryption coverage
- added `purge-docs` to the parent collection synopsis

I also added issue #5117 and updated the PR body to the repository template with the operation limitations. Focused tests, race tests, lint, generated CLI docs, and generated OpenAPI checks pass locally. Fresh CI is running now.

iverc:
Follow-up CI fix: removed the test-only alternate collection identity. Purge tests now follow the normal collection lookup on every client; unauthorized callers fail at `get-collection`, while node admins proceed through the dedicated `purge-document` check.

The targeted NAC tests pass through Go, HTTP, CLI, and C, and the full client matrix is green. The remaining vulnerability check is also failing on `develop` because its expected govulncheck count is stale.

iverc:
Addressed the review items and pushed the follow-up fixes. Purge now has its own options and NAC permission, the parent help/docs include purge-docs, and the SourceHub ACP and secondary-index matrices are green. The vulnerability job is failing the same way on develop; LevelDB and macOS are still queued. Ready for another pass.

iverc:
Stacked #5119 so CI now runs against the fixes for the three default-branch failures. The LevelDB hang was a nested-transaction regression from #5101: the exact failing test went from timing out after 45 seconds to passing in 1.65 seconds, and the purge flow now passes on LevelDB as well. All review threads are resolved and this is ready for another look.

iverc:
The first full rerun exposed one more baseline test-harness issue: transactional DeleteIndex was waiting for background cleanup before its transaction committed. #5119 now gives it the same transaction guard already used by NewIndex. The full LevelDB transaction package and 10 repeated runs of the failing test pass locally; the follow-up is stacked here for the new CI run.

iverc:
The remaining deterministic LevelDB failures had the same teardown cause: integration cases left explicit transactions open, then node shutdown waited on an index worker blocked acquiring LevelDB transaction access. The harness now discards those transactions before closing nodes. All three formerly hanging cases pass three consecutive runs; their full packages and the purge tests pass under the race detector. The Wazero failure was a separate P2P sync timeout, and the two affected sync cases passed 40 combined consecutive local executions, so no product change was made for it. Fresh CI is running.

iverc:
Stacked the verified peer-sync baseline from #5119. The latest `#5097` macOS retry had no purge failure; its only failure was the open Darwin race-runtime crash in [golang/go#80292](https://github.com/golang/go/issues/80292).

The shared update fixes the underlying peer-ID bookkeeping and network-readiness races, and adds a bounded macOS retry that still fails on data races, multiple failures, or a repeated failure. On the stacked branch, P2P internals, network info, document and branchable sync, and purge integration all pass together under shuffled race execution. Fresh CI is running.

iverc:
@coderabbitai review

tests/integration/acp/nac/purge_document_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Identity:     testUtils.ClientIdentity(1),
+			CollectionID: 0,
+			Doc:          `{"name":"alice"}`,
+		},
+		&action.PurgeDocs{
+			Identity:        testUtils.NoIdentity(),
+			CollectionIndex: 0,
+			DocIndexes:      []int{0},
+			ExpectedError:   testUtils.FormatExpectedErrorWithPermission(acpTypes.NodeGetCollectionPerm),
+		},
```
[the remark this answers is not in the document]
iverc: The preflight cases are now named explicitly, and `TestPurgeByDocIDsChecksDedicatedNACPermission` bypasses collection lookup by retaining the owner-resolved collection. It verifies that the second actor is denied specifically with `Permission: purge-document`, then grants `admin` and verifies success. The built-in NAC policy only defines the all-or-nothing `admin` relation, so a collection-get-only relationship cannot be represented without changing the production policy just for this test.

tests/integration/searchable_encryption/purge_test.go
```diff
@@ -0,0 +1,84 @@
...
+					encrypted_User(filter: {name: {_eq: "alice"}}) {
+						docIDs
+					}
+				}`,
+				Results: map[string]any{
+					"encrypted_User": []map[string]any{{
+						"docIDs": gomega.BeEmpty(),
+					}},
+				},
+			},
```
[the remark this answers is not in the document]
iverc: I verified this against the running integration test. Node 0 is the query coordinator and node 1 is its SE replicator, where the artifacts are stored. Node 0 forwards the encrypted query to node 1; node 1 itself has no replicator to query and returns empty regardless of purge state. The existing node-0 assertion is therefore the one that proves node 1 local cleanup, not replicated purge behavior. I kept it and added a short topology comment; the networked test passes under `-race`.

iverc:
Addressed the final review pass in 50ac61edf: the SE cleanup scan now resumes within the collection range, block cleanup explicitly ignores missing blocks during retries, LevelDB cleanup runs on timeout, and both Badger limit checks use the sentinel. The direct NAC test also verifies the exact `purge-document` permission before and after an admin grant. Focused unit and integration tests pass under `-race`; lint reports 0 issues.

@coderabbitai review

iverc:
Added the denied-state assertion in 13a59455f. The owner still finds the document before the admin grant; the focused test passes under `-race` and lint remains clean.

@coderabbitai review

.github/workflows/check-vulnerabilities.yml
```diff
@@ -47,10 +47,12 @@ jobs:
...
       - name: Check if only known vulnerabilities were found (there are new vulnerabilities if this fails)
-        run: cat govulncheck.txt | grep "Your code is affected by 2 vulnerabilities from 1 module."
+        run: >-
+          sed -n 's/^Vulnerability #[0-9][0-9]*: //p' govulncheck.txt |
+          sort |
+          diff -u .github/known-vulnerabilities.txt -
 
-    # Use the steps below once the x/crisis (crisis.init) bug is fixed or if the
-    # ability to silence is implemented: https://github.com/golang/go/issues/61211
+    # Remove allowlisted findings as compatible upstream fixes become available.
```
shahzadlone: question: Why are these changes in this PR?

.github/workflows/test-macos.yml
```diff
@@ -76,4 +76,10 @@ jobs:
         uses: ./.github/composites/setup-defradb
 
       - name: Run integration tests
+        # Work around https://github.com/golang/go/issues/80292 without suppressing data races.
+        env:
+          GOTESTSUM_FLAGS: >-
+            --rerun-fails=2
+            --rerun-fails-max-failures=1
+            --rerun-fails-abort-on-data-race
```
shahzadlone: question: Is this needed for the purge changes you did?

.github/known-vulnerabilities.txt
shahzadlone, line 1: question: Why are these changes in this PR?

acp/types/types.go
shahzadlone, line 1: praise: thanks for doing the todo :)

cli/wizard/unit_test.go
```diff
@@ -353,27 +355,25 @@ func Test_GenerateKeysInSystemKeyring_AllKeys(t *testing.T) {
 	// Assign a unique namespace for the test keyring so we can remove it afterwards
 	keyringNamespace := fmt.Sprintf("test-system-keyring-%d", time.Now().UnixNano())
 	setConfigValueForTest(t, ctx, "keyring.namespace", keyringNamespace)
+	openKeyring := keyring.OpenSystemKeyring(keyringNamespace)
+	t.Cleanup(func() {
+		_ = openKeyring.Delete("node-identity-key")
+		_ = openKeyring.Delete("peer-key")
+		_ = openKeyring.Delete("encryption-key")
+		_ = openKeyring.Delete("searchable-encryption-key")
+	})
```
shahzadlone: question: Is this related to the Purge feature?

cli/collection.go
```diff
@@ -25,7 +25,7 @@ func MakeCollectionCommand(ctx context.Context) *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "collection",
 		Short: "Interact with a collection.",
-		Long:  `Add, describe, patch, set-active, delete, and truncate collections.`,
+		Long:  `Add, describe, patch, set-active, delete, purge-docs, and truncate collections.`,
```
shahzadlone: question: Why purge-docs? 

question: Why on collection path rather than on document path?

cli/collection_purge_docs.go
```diff
@@ -0,0 +1,66 @@
...
+		Short: "Permanently remove documents by DocID from the local node",
+		Long: `Permanently remove a set of documents by DocID from the local node, including all
+datastore values and headstore entries. When --prune-history is set, it also removes
+reachable blockstore blocks that are no longer owned by another document. Shared blocks
+are retained until their final owning document is purged.
+
+History pruning is not supported for branchable collections.
+
+Unlike the soft-delete performed by the delete command, this operation is irreversible and
+does not propagate to other nodes in the peer network.`,
```
shahzadlone: thoght: I don't like the fact that you have to specify Docs after Purge

internal/db/collection_index.go
```diff
@@ -361,28 +364,51 @@ func (c *collection) deleteIndexedDocWithID(
...
-		ctx,
-		primaryKey,
-		c.collectIndexedFields(indexes),
-		false,
-	)
+	fields := c.collectIndexedFields(indexes)
+	var doc *client.Document
+	if includeDeleted {
+		// Purge is authorized at the node level and must not apply the document read filter.
+		doc, err = c.getInternal(ctx, primaryKey, fields, true)
```
shahzadlone: question: What does it mean that  Purge is authorized at the node level?

question: Why do we need a `getInternal`?

internal/db/collection_purge.go
shahzadlone: What do you mean by "The regression uses one document with 3,000 updates"? I see` purgeChunkSize = 100` where does the 100 comes from and what does that mean now?

internal/db/collection_truncate.go
```diff
@@ -299,22 +297,16 @@ func (c *collection) hardDeleteDatastorePrefix(
 		}
 		datastore, ok := ds.(unsafestore)
 		if !ok {
-			return NewErrTruncateDatastoreKey(errors.New("datastore does not expose unsafe writer"), prefix.ToString())
+			return NewErrTruncateDatastoreKey(
+				errors.New("datastore does not expose unsafe writer"),
```
shahzadlone: todo: put in errors.go file please

internal/db/collection_truncate_test.go
shahzadlone, line 1: question: What is this testing? 

question: Is this testing `truncate` behaviour?

internal/db/errors.go
```diff
@@ -106,6 +106,7 @@ const (
 	errSelfReferenceWithoutSelf            string = "must specify 'Self' kind for self referencing relations"
 	errColNotMaterialized                  string = "non-materialized collections are not supported"
 	errColMutatingIsBranchable             string = "mutating IsBranchable is not supported"
+	errCannotPruneBranchableCollection     string = "cannot prune document history in a branchable collection"
```
shahzadlone: question: Is this documented somewhere user facing and the reason for it?

tests/action/active_peers.go
```diff
@@ -20,6 +23,11 @@ import (
 	"github.com/sourcenetwork/defradb/tests/state"
 )
 
+const (
+	activePeersTimeout      = 5 * time.Second
+	activePeersPollInterval = 10 * time.Millisecond
```
shahzadlone: question: Is this purge feature related or an independant change?

tests/action/utils_test.go
shahzadlone, line 1: question: Is this general refactor or needed for purge feature?

tests/integration/acp/nac/purge_document_test.go
```diff
@@ -0,0 +1,91 @@
...
+			Identity:        testUtils.ClientIdentity(1),
+			CollectionIndex: 0,
+			DocIndexes:      []int{0},
+		},
+	}}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestNAC_GatesPurgeDocument_NoIdentity_CollectionLookupDenied(t *testing.T) {
```
shahzadlone: todo: AI slop test name haha? fix the `CollectionLookupDenied` that is not what we are testing here nor the result (it might be the first gate that is hit in which case you should also mark all the NAC tests like others with todo marked in comments to be fixed later).

tests/integration/acp/nac/purge_document_test.go
```diff
@@ -0,0 +1,91 @@
...
+			CollectionIndex: 0,
+			DocIndexes:      []int{0},
+			ExpectedError:   testUtils.FormatExpectedErrorWithPermission(acpTypes.NodeGetCollectionPerm),
+		},
+	}}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestNAC_GatesPurgeDocument_WrongIdentity_CollectionLookupDenied(t *testing.T) {
```
shahzadlone: todo: AI slop test name haha? fix the `CollectionLookupDenied` that is not what we are testing here nor the result (it might be the first gate that is hit in which case you should also mark all the NAC tests like others with todo marked in comments to be fixed later).

tests/integration/net/info/connect_peers_test.go
```diff
@@ -99,10 +98,6 @@ func TestNetInfoConnectMultiplePeers(t *testing.T) {
 					"{{.Peer2_Address0}}",
 				},
 			},
-			&action.Wait{
-				// Wait for the connections to propagate
-				Duration: immutable.Some(time.Millisecond * 50),
-			},
```
shahzadlone, line 105: question: Independent change or needed for purge stuff?

tests/integration/net/simple/peer/with_add_test.go
```diff
@@ -311,6 +313,30 @@ func TestP2PAdd_WithP2PCollectionOnLastNodeInNodeChain_ShouldPropagateUpdate(t *
...
+				ExpectedPeersByTopic: map[string][]int{
+					docSyncTopic: {3},
+				},
+			},
+			&action.WaitForPeersEvents{
+				NodeID: 3,
+				ExpectedPeersByTopic: map[string][]int{
+					docSyncTopic: {4},
+				},
+			},
```
shahzadlone: question: Can you explain what is happening here?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sourcenetwork/defradb
 
-go 1.25.9
```
shahzadlone: question: Why was this done in this PR?

Makefile
```diff
@@ -68,6 +68,7 @@ BUILD_FLAGS+=-tags $(BUILD_TAGS)
 endif
 
 TEST_FLAGS=-race -shuffle=on -timeout 10m
+GOTESTSUM_FLAGS?=
```
shahzadlone: question: What does this fix?

shahzadlone (dismissed):
Just finished my second review of this PR and I feel like there some work that needs to be done before I feel comfortable having this merged.

One of my main concerns with the current implementation is that I don't really think `Purge` should be a collection level operation, is there a reason for why you have went with that route? Pre V1 we went through some pain lol to strip out such operations off of collection path.

Secondly, there is a lot of unrelated changes to the purge work that can be separate PRs out of this already quite large 3200+ line PR. I have asked questions for confirmation in many places incase they are needed for purge or not to see if i should even bother context switching and reviewing those changes or not, please do confirm if they are related and i will happily review those.

I also feel like there is perhaps too much automation, or perhaps AI assisted management on this PR such that almost all my questions I asked in the last round of PR review I got no replies or answers to further my own learning and to perhaps discuss things before coming to a consensus on how we should proceed, but just two comment replies per each thread that repeat the same info of highlights (duplicates of each other) indicating that the suggestion was just done/fixed (which even i was not 100% sure if we wanted as it would be dependent on the question and answers to them). In addition to this I also notice some comments likely also AI-slop ones on the PR that indicate some metrics or stats or like ci statuses which can pollute the PR context IMO would prefer to tune those down perhaps if there is consensus from others too. For example like these: 
image

I definitely prefer the work done in a PR and all the important bits of information to be reflected on the PR description. 

The agentic stuff is definitely new territory lol and perhaps we can discuss some of these as a team and see how we can use such automation in a balance that don't take away that human element.

shahzadlone (commented):
I tried re-reviewing but I see a lot of (if not all) my previous questions and comments not answered or replied to. Some I understand were obvious and can be fixed or resolved, those you can do a quick reply to and make it easy for re-review on how you solved it or fixed, and the other ones would help my learn and understand the PR more

.github/workflows/check-vulnerabilities.yml
iverc: unrelated to purge, moved to #5119

.github/workflows/test-macos.yml
iverc: nope, unrelated CI change. moved to #5119

.github/known-vulnerabilities.txt
iverc: same here, unrelated. moved to #5119

cli/wizard/unit_test.go
iverc: not related, moved to #5119

cli/collection.go
iverc: yeah agreed. its store-level now, and the CLI is `defradb client document purge`; collection name is just an input

cli/collection_purge_docs.go
iverc: yep, changed it to just `document purge`

internal/db/collection_index.go
iverc: `purge-document` is a node permission because this is local maintenance, not a replicated doc mutation. `getInternal` only loads index fields after that check, so a node admin doesnt also need `read-document`

internal/db/collection_truncate.go
iverc: done, moved them to `internal/db/errors.go`

internal/db/collection_truncate_test.go
iverc: It tests `deleteBlockIfPresent`, the shared deletion helper used by both truncate and purge. Missing blocks must be treated as success because multiple heads may reach the same block and a resumed purge may revisit it; other deletion errors still propagate.

internal/db/errors.go
iverc: Yes. The Store GoDoc, CLI reference, and OpenAPI description now state that branchable history pruning is unsupported because its retained collection DAG still links to document blocks. The HTTP API returns 422 for that case.

tests/action/active_peers.go
iverc: It was independent and has been removed from this PR. The peer-readiness work is now isolated in #5119.

tests/action/utils_test.go
iverc: It was independent and has been removed from this PR. It is now isolated in #5119.

tests/integration/acp/nac/purge_document_test.go
iverc: Agreed. The names are now `NoIdentity_NotAuthorizedError` and `WrongIdentity_NotAuthorizedError`, and both assert the dedicated `purge-document` permission. The old `CollectionLookupDenied` names are gone.

tests/integration/net/info/connect_peers_test.go
iverc: Independent; it has been removed from this PR and is now isolated in #5119.

go.mod
iverc: unrelated Go baseline bump, moved to #5119

Makefile
iverc: baseline test change, not purge. moved to #5119

internal/db/collection_purge.go
iverc: The old `purgeChunkSize = 100` bounded documents per transaction, which did not solve the single-document history case you identified. That design was removed. Implicit purge now prunes history incrementally under the collection lock and commits each document logical cleanup separately; `hardDeleteChunkSize` only caps iterator batches in memory. The 3,000-update test proves one large document completes without a caller transaction, while a caller-owned transaction still fails if it exceeds the backend limit.

internal/db/collection_truncate_test.go
iverc: it tests `deleteBlockIfPresent`, shared by truncate and purge. missing blocks are ok because two heads can point to the same block; other errors still fail

internal/db/errors.go
iverc: yep. Go docs, CLI docs and OpenAPI all say branchable history pruning isnt supported and why; HTTP returns 422

tests/action/active_peers.go
iverc: unrelated, moved to #5119

tests/action/utils_test.go
iverc: same, unrelated and moved to #5119

tests/integration/acp/nac/purge_document_test.go
iverc: yeah that name was bad. now they are `NoIdentity_NotAuthorizedError` and `WrongIdentity_NotAuthorizedError`, checking `purge-document` directly

tests/integration/acp/nac/purge_document_test.go
iverc: same fix here, this case is now `WrongIdentity_NotAuthorizedError`

tests/integration/net/info/connect_peers_test.go
iverc: unrelated, moved to #5119

tests/integration/net/simple/peer/with_add_test.go
iverc: peer readiness stuff, unrelated to purge. moved to #5119

internal/db/collection_purge.go
iverc: the 100 was a doc-count batch and didnt solve the case you pointed out, so its gone. implicit purge now walks history outside that transaction limit and commits each doc's logical cleanup separately; the 3k update test covers that

iverc:
@shahzadlone  yeah fair, sorry about that. went back through the threads and answered the open questions. unrelated changes are out of this PR now too.

AndrewSisley (changes requested):
Hey Ivan, I'm not sure we want to expose this behaviour in this format - I've opened a discussion on discord, please do not merge this until the team has reached a decision.

tests/integration/collection/truncate/simple_test.go
```diff
@@ -54,3 +54,39 @@ func TestTruncateCollection_TruncateTwice(t *testing.T) {
...
+			&action.Truncate{
+				CollectionIndex: 0,
+				DocIndexes:      []int{0},
+				PruneHistory:    true,
+			},
+			&action.Truncate{
+				CollectionIndex: 0,
+				DocIndexes:      []int{0},
+				PruneHistory:    true,
+			},
```
AndrewSisley: todo: Please either document why you are truncating twice with the same params, or remove the second truncate

tests/integration/collection/truncate/simple_test.go
```diff
@@ -54,3 +54,39 @@ func TestTruncateCollection_TruncateTwice(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+func TestTruncateCollection_WithFilter(t *testing.T) {
```
AndrewSisley: suggestion:  Moving this test to a `filter_test.go` file in the same directory would be more in keeping with our test structure, and will let devs more easily find it.

tests/integration/mutation/truncate/simple_test.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2026 Democratized Data Foundation
```
AndrewSisley, line 1: suggestion: Moving these tests to a filter_test.go file in the same directory would be more in keeping with our test structure, and will let devs more easily find it.

client/options/collection.go
```diff
@@ -410,6 +410,10 @@ func (b *ExistsDocumentOptionsBuilder) SetIdentity(id identity.Identity) *Exists
 type TruncateCollectionOptions struct {
 	// Identity is the identity of the actor performing the operation.
 	Identity immutable.Option[identity.Identity]
+	// Filter limits the truncate to matching documents. A nil filter truncates the full collection.
+	Filter any
+	// PruneHistory removes unshared history for documents selected by Filter.
+	PruneHistory bool
```
AndrewSisley: todo: Please expand the documention for this property - it is very unclear as to why we or our users would want this, and I had to dig through the implementation in order to be sure as to what it meant.  Please be clear as to why you think users should be able to delete blocks being used by other documents (this looks like an old bug?).

internal/request/graphql/schema/generate.go
```diff
@@ -1288,7 +1288,20 @@ func (g *Generator) GenerateMutationInputForGQLType(obj *gql.Object) ([]*gql.Fie
 		},
 	}
 
-	return []*gql.Field{add, update, delete, upsert}, nil
+	truncate := &gql.Field{
+		Name:        "truncate_" + obj.Name(),
+		Description: "Permanently remove all or matching documents from this node.",
```
AndrewSisley: suggestion: Please consider rewording this a little bit - `all or` looks a little randomized :)

Maybe the below would be better, but I have no attachment to this particular phrasing:
```
Remove documents and their history from this node.
```
Please also consider using this space to document what the returned boolean represents.

client/collection.go
```diff
@@ -59,12 +59,10 @@ type Collection interface {
 	// will be created.
 	SaveDocument(ctx context.Context, doc *Document, opts ...options.Enumerable[options.SaveDocumentOptions]) error
 
-	// DeleteDocument will attempt to delete a document by DocID.
+	// DeleteDocument soft-deletes a document by DocID and publishes the deletion for replication.
```
AndrewSisley, line 62: praise: Thanks for clarifying this!

cbindings/collection_truncate.go
```diff
@@ -28,6 +29,37 @@ func TruncateCollection(
 	nodePtr C.uintptr_t,
 	opts C.CollectionOptions,
 	identityPtr C.uintptr_t,
+) C.Result {
+	return truncateCollection(nodePtr, opts, identityPtr, nil, false)
+}
+
+//export TruncateCollectionWithFilter
+func TruncateCollectionWithFilter(
```
AndrewSisley, line 42: question: Why did you add a new function here?

todo: If this was added to avoid a semver breaking change because the original C `TruncateCollection` doesn't take a dedicated options struct, please document this here, and add a `deprecated: ...` line to the documentation of `TruncateCollectionWithFilter` explaining that it will be removed in v2.0.0

cbindings/collection_truncate.go
```diff
@@ -28,6 +29,37 @@ func TruncateCollection(
...
+	pruneHistory C.int,
+) C.Result {
+	if filterJSON == nil {
+		return returnC(returnGoC(1, "filter is required", ""))
+	}
+	var filter any
+	if err := json.Unmarshal([]byte(C.GoString(filterJSON)), &filter); err != nil {
+		return returnC(returnGoC(1, err.Error(), ""))
+	}
+	if filter == nil {
```
AndrewSisley, line 56: suggestion: This looks like duplicate, dead and untested code, might be better to remove it. (suggestion not a todo as the entire func might be deprecated)

internal/planner/operation.go
```diff
@@ -94,7 +94,7 @@ func (n *operationNode) Next() (bool, error) {
 	n.currentValue = n.documentMapping.NewDoc()
 	for i, child := range n.children {
 		switch child.(type) {
-		case *topLevelNode:
+		case *topLevelNode, *truncateNode:
```
AndrewSisley, line 98: question: Are you sure this is required?

todo: Please document why it is required, or remove it.

internal/planner/explain.go
```diff
@@ -57,6 +57,7 @@ var (
 	_ explainablePlanNode = (*selectTopNode)(nil)
 	_ explainablePlanNode = (*sumNode)(nil)
 	_ explainablePlanNode = (*topLevelNode)(nil)
+	_ explainablePlanNode = (*truncateNode)(nil)
```
AndrewSisley: todo: If you are going to add explain support for truncate, please add integration tests for it.  If not adding tests, please remove this line.

suggestion: Please avoid adding explain support for truncate in this PR, it will make reviewing it more complicated than it needs to be compared to a seperate PR.

AndrewSisley (changes requested):
Is looking good, thank you very much Ivan!  Just a handful of smaller requests from me before merge.

iverc:
> Is looking good, thank you very much Ivan! Just a handful of smaller requests from me before merge.

Thank you so much, @AndrewSisley . Addressing the feedback!

client/options/collection.go
```diff
@@ -410,6 +410,11 @@ func (b *ExistsDocumentOptionsBuilder) SetIdentity(id identity.Identity) *Exists
 type TruncateCollectionOptions struct {
 	// Identity is the identity of the actor performing the operation.
 	Identity immutable.Option[identity.Identity]
+	// Filter limits the truncate to matching documents. A nil filter truncates the full collection.
```
AndrewSisley, line 413: todo: Sorry for the late todo, there is no indication on this `any` property as to what the user should be providing to it - please let them know, preferably with an example.

AndrewSisley (approved):
Looks great to me, thanks for all the extra effort on this one Ivan.

Approving now, but please make sure `PruneHistory` is removed before merge (for ref. discussed over discord), and please expand the documentation on the filter option.

iverc:
> Looks great to me, thanks for all the extra effort on this one Ivan.
> 
> Approving now, but please make sure `PruneHistory` is removed before merge (for ref. discussed over discord), and please expand the documentation on the filter option.

Thank you! Removed `PruneHistory` option: https://github.com/sourcenetwork/defradb/pull/5097/commits/85dca42888d6dd738b0f0a8d317bef94f31ea96a

islamaliev:
Bug bash results:
tested against real defradb running locally. Everything is working properly but I wonder if `filter: {}` should remove all the docs. This is how it works now.

PR #5101, opened 2026-07-29
fix: Isolate document short ID allocation

### LINKED ISSUE #5100: P2P may silently drop concurrent document merges
Concurrent P2P pushlog handlers can call `DB.Merge` at the same time for different documents. New documents still update the shared document short-ID sequence, so these merges can conflict even when their document IDs differ.

`DB.Merge` retries transaction conflicts, but currently returns `nil` after exhausting the retry budget. The caller then treats the merge as successful, while the document was never committed. Under a burst of 24 replicated documents this consistently leaves several documents missing without reporting an error.

The P2P merge/commit step should be serialized while leaving DAG synchronization concurrent, and exhausted merge retries must return an error.

iverc:
## Relevant issue(s)

Resolves #5100

## Description

Concurrent new-document transactions update the same node-wide short-ID sequence. That can exhaust `DB.Merge` retries even when the documents are unrelated; the old retry loop then returned `nil` and silently lost the merge.

Short IDs are now reserved under a DB-level mutex in a small independent transaction shared by local document creation and `DB.Merge`. Only sequence reservation is serialized. DAG loading, CRDT merging, index synchronization, and existing-document merges remain concurrent. A failed document transaction may leave an unused ID, which is safe because short IDs are internal and only need to be unique.

`DB.Merge` now makes at least one attempt and reports exhausted transaction conflicts. The collection short-ID cache also uses atomic access so concurrent operations on one collection handle are race-free.

## Tasks

## How has this been tested?

- `go test ./internal/db ./internal/db/p2p`
- 24-way concurrent local-create and direct-`DB.Merge` tests, 10 runs under `-race`
- golangci-lint against `origin/develop`: 0 new issues
- Trust API two-node P2P convergence test under `-race`: 10/10 runs, 24 concurrent audit writes per run

Platform: macOS

### DIFF
--- a/internal/db/collection.go
+++ b/internal/db/collection.go
@@ -12,6 +12,7 @@ package db
 
 import (
 	"context"
+	"sync/atomic"
 
 	"github.com/sourcenetwork/immutable"
 
@@ -40,7 +41,7 @@ type collection struct {
 	// shortID caches the collection's short ID, which is stable for the collection's life. Resolved
 	// once on first use so the write path does not re-read it every time. Zero means not yet resolved,
 	// since a valid short ID starts at 1.
-	shortID uint32
+	shortID atomic.Uint32
 	// indexBuildStates holds the build (backfill) state of the collection's indexes that have one,
 	// keyed by index ID: an index is present while it is building or failed, and absent once ready.
 	// Presence therefore means "not yet queryable". Populated at construction time from the index
--- a/internal/db/collection.go
+++ b/internal/db/collection.go
@@ -40,7 +40,7 @@ type collection struct {
 	indexesCounter uint64
 	// shortID caches the collection's short ID, which is stable for the collection's life. Resolved
 	// once on first use so the write path does not re-read it every time. Zero means not yet resolved,
-	// since a valid short ID starts at 1.
+	// since a valid short ID starts at 1. Document writes may access the cache concurrently.
 	shortID atomic.Uint32
 	// indexBuildStates holds the build (backfill) state of the collection's indexes that have one,
 	// keyed by index ID: an index is present while it is building or failed, and absent once ready.
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -265,7 +265,8 @@ func (db *DB) NewTxn(readonly bool) (client.Txn, error) {
 
 // reserveDocShortID commits the node-wide sequence separately from document writes so concurrent
 // document transactions do not contend on the sequence key. Failed document transactions may leave
-// gaps; short IDs only need to be unique.
+// gaps; short IDs only need to be unique. The lock spans transaction creation through commit so each
+// transaction sees the last committed reservation.
 func (db *DB) reserveDocShortID(ctx context.Context) (uint64, error) {
 	db.docShortIDMu.Lock()
 	defer db.docShortIDMu.Unlock()
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -116,6 +116,7 @@ type DB struct {
 
 	docMergeQueue *mergeQueue
 	colMergeQueue *mergeQueue
+	docShortIDMu  sync.Mutex
 
 	p2p *p2p.P2P
 	// Retry intervals when a replicator failure occurs.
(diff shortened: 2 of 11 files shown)


iverc:
Final verification against the real Trust API two-node flow (24 concurrent audit writes, same Trust API revision; only the DefraDB checkout changed):

- Before, using the purge-only DefraDB head without this PR: 0/5 runs converged. The nodes retained 17, 18, 21, 20, and 19 of 24 audit documents while aggregate stats incorrectly reported all 24. Every run hit the 15s convergence timeout.
- After, using the exact combined PR heads b0121527 + 36d92be4: 10/10 runs converged in 20.96s total test time, about 2.10s per run.

Also verified internal/db and internal/db/p2p package tests, focused race tests, and golangci-lint with 0 new issues. The follow-up routes document sync and branchable collection sync through the same serialized merge path as pushlog delivery.

iverc:
CI note:

- Coverage passed 36/37 jobs. The remaining LevelDB failure matches the latest `develop` run exactly: `TestTxn_DeleteIndex_WithCommit_Succeeds` times out waiting for an index drop, then `TestColVersionUpdateReplaceIsActive_GetCollectionsWithTxn` hangs until the 10-minute package timeout.
- macOS hit the same five system-keyring failures as `develop`, plus a one-second branchable-sync timeout. The exact affected sync test passes 100/100 locally on macOS with `-race` at `36d92be4`; the latest `develop` macOS run also fails on a P2P timeout in another test.
- The vulnerability workflow is also failing on `develop`; its expected vulnerability count is stale relative to the current dependency scan.

Base coverage: https://github.com/sourcenetwork/defradb/actions/runs/30392652964/job/90391332793
Base macOS: https://github.com/sourcenetwork/defradb/actions/runs/30392652787/job/90387826797
Base vulnerability check: https://github.com/sourcenetwork/defradb/actions/runs/30392652739

internal/db/p2p/p2p.go
```diff
@@ -715,6 +717,12 @@ func (p *P2P) processPushlogRequest(
 	return nil
 }
 
+func (p *P2P) merge(ctx context.Context, evt event.Merge) error {
+	p.mergeMu.Lock()
+	defer p.mergeMu.Unlock()
+	return p.db.Merge(ctx, evt)
```
islamaliev: todo: You are protecting a public critical section from the wrong place. This doesn't solve the concurrency problem because it can be called from any other place. Why not to make the mutex part of the DB itself?

todo: Another point is that the `Merge` function is quite heavy And ideally we would want to keep it concurrent-capable. I would say it makes sense to find a real critical section and protect it in a very narrow window.

islamaliev (changes requested):
I might be missing something here but this doesn't seem to be the right solution.

internal/db/p2p/p2p.go
iverc: You are right. I removed the P2P-level lock. Short IDs are now reserved under a DB-level mutex in a tiny independent transaction used by both AddDocument and DB.Merge, so only the sequence update is serialized and the rest of Merge stays concurrent. The separate transaction also avoids waiting merges keeping a stale sequence snapshot. The new 24-way create/merge tests pass 10x under -race, and the Trust API two-node flow passed 10/10.

internal/db/collection.go
```diff
@@ -40,7 +41,7 @@ type collection struct {
 	// shortID caches the collection's short ID, which is stable for the collection's life. Resolved
 	// once on first use so the write path does not re-read it every time. Zero means not yet resolved,
 	// since a valid short ID starts at 1.
-	shortID uint32
+	shortID atomic.Uint32
```
islamaliev, line 44: todo: please document why it needs to be atomic

internal/db/db.go
```diff
@@ -262,6 +263,29 @@ func (db *DB) NewTxn(readonly bool) (client.Txn, error) {
 	return wrapDatastoreTxn(txn, db), nil
 }
 
+// reserveDocShortID commits the node-wide sequence separately from document writes so concurrent
+// document transactions do not contend on the sequence key. Failed document transactions may leave
+// gaps; short IDs only need to be unique.
+func (db *DB) reserveDocShortID(ctx context.Context) (uint64, error) {
+	db.docShortIDMu.Lock()
+	defer db.docShortIDMu.Unlock()
```
islamaliev, line 272: suggestion: this can be done after the transaction is created. Also seems like you can release before the txn is committed.

islamaliev (approved):
Left 2 more requests.

Side note: I wonder if it makes sense to move shortIDs from global scope to collection scope.

internal/db/collection.go
iverc: Documented in 2917737d8. Concurrent document writes may initialize or read this cache.

internal/db/db.go
iverc: I kept the lock before transaction creation and through commit. NewTxn creates the underlying corekv transaction immediately, so creating it before the lock can give a waiting reservation a stale snapshot. The reservation is not durable until Commit; unlocking earlier would allow another transaction to read the same sequence value and reintroduce conflicts. I added that lock-scope rationale in 2917737d8.

Issue #5066, opened 2026-07-20
P2P cross-version compatibility testing (RFC)


islamaliev:
## Problem

Now that v1.0 is released, every future minor must stay compatible with nodes running older supported versions. A v1.1 node has to talk to a v1.0 node without breaking. We have no automated way to catch it when a change breaks that compatibility, so a wire-level break could ship unnoticed and only surface in the field (as it did during the Shinzo testnet release).

We want to catch these breaks early, ideally at the PR level.

## Scope

Cross-version *behavioral* compatibility of the networking layer: run our network/P2P tests with the latest node against each version we still support (per the LTS policy, the last 3 minors), and confirm behavior stays compatible.

## Open questions (not decided)

- Where does this run? PR-level for fast feedback, nightly for the fuller matrix, or a split?
- How many versions back, and how does it interact with the LTS window as it grows?
- How do we handle tests whose expected behavior legitimately changes across versions (a new feature a node adds), so they aren't treated as failures?

## Direction (clearer now, still open to change)

Since this was first written, a few things settled in the July syncs and while building the drift detector (#5079, https://github.com/sourcenetwork/defradb/pull/5078):

**Exec pre-compiled version binaries, not native embedding.** Importing multiple versions of the node into one test binary was explored and rejected — it won't build (dependency hell across versions; Go plugins don't help). Instead, each release is compiled as a build artifact and the test execs the right binary per node. Locally, build the last ~3 supported versions (a worktree + `make` per version).

**Both directions.** New→old and old→new are asymmetric and both matter: new→old tolerates unknown fields (CBOR ignores them), while old→new depends on the new node handling an absent field as a sensible zero value. The matrix should replay each scenario both ways.

**Delivered through the test multiplier.** Broaden the node-config so a test node can carry a version (git tag/commit), then let the multiplier expand a scenario across the version matrix. The current `ConfigureNode` abstraction is network-only, so it needs widening to a node-config that can express a version. The multiplier transforms action sets, so the version has to be reachable as part of a node-config action (or the multiplier interface extended).

**Coverage = the P2P tests replayed.** The matrix only proves compatibility for the scenarios actually run through it. Widening P2P test coverage directly widens the guarantee.

## How this relates to drift detection

Drift detection (#5079 / #5067) and this are complementary, not overlapping:
- The **snapshot** flags that a wire type *changed shape* (shipped in https://github.com/sourcenetwork/defradb/pull/5078).
[...]

PR #5103, opened 2026-07-31
test(i): P2P cross-version compatibility testing

islamaliev:
## Relevant issue(s)

Refs #5066

## Description

Older nodes need to keep syncing with newer ones once we ship minors, but nothing checked that automatically, so a wire break could ship and only surface in the field. This adds test infrastructure to run the current build against a real older release and confirm a document still syncs between them, in both directions.

The old node runs as a separate process from its published release binary, driven over the normal HTTP and P2P paths, because two versions of the database cannot live in one test binary. A node in the test suite can now declare which version it is; leaving it unset keeps the existing in-process behavior, so nothing else changes. The release binary is fetched over plain HTTP from the project's release page, so no extra tooling or credentials are needed.

This lands the foundation plus two smoke tests for the one pair that matters today, the latest release against the current build. A follow-up will generalize it with a multiplier so the existing suite can be replayed across versions.

## Tasks

## How has this been tested?

Two integration tests exercise real bidirectional P2P document sync between the current build and a downloaded v1.0.0 binary, plus unit tests for version resolution and the external-node wrapper. The tests skip cleanly on platforms without a matching release asset. The existing net suite passes unchanged, confirming the native path is unaffected.

- MacOS


tests/integration/version/version.go
```diff
@@ -0,0 +1,233 @@
...
+//
+// See tests/LICENSE for details.
+
+/*
+Package version resolves a DefraDB version string (e.g. "v1.0.0") to a path
+to an executable release binary for the current platform, downloading and
+caching it on demand.
+
+This exists so P2P cross-version compatibility tests can drive a pre-built
+release binary as a separate process alongside a native HEAD node.
```
AndrewSisley, line 18: suggestion: This looks way more complicated than we need it to be.  We already do pretty much all that this file does in the change detector.  By rebuilding locally we avoid needing to discover and handle a lot of the stuff within this file.

I suggest reworking this to clone and rebuild using tags instead of having to worry about all the stuff in here.  You can probably refactor the change detector code out and share it between these two without much trouble too.

tests/integration/net/simple/replicator/cross_version_test.go
```diff
@@ -0,0 +1,147 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// A doc created on the current build replicates to a v1.0.0 node.
+// Skips when v1.0.0 has no release asset for this platform.
+func TestP2PCrossVersion_HeadToV1_DocSyncs(t *testing.T) {
```
AndrewSisley, line 75: todo: This should really be a test multiplier, not an explicit suite of tests.  There is a large and complex surface area, and it will probably grow over time - I do not think it is sensible to rely on explicit tests long term for this.

I understand you feel a bit of a rush on this though in order to get *some* protection in asap.  I'm happy enough with this being merged as an emergancy solution and ticket being added to the backlog if you prefer.

AndrewSisley (approved):
I do not think this is the approach we want long term, but if you want it in as-is in the short term I am fine with that :)

tests/integration/version/version.go
islamaliev: Downloading the release binary was a deliberate choice to keep the first version simple: no local toolchain or build step, just fetch and run. Build-from-tag (and sharing the change-detector code) is a good direction and I noted it as a follow-up.

tests/integration/net/simple/replicator/cross_version_test.go
islamaliev: Agreed, a multiplier is the right long-term shape and these explicit tests don't scale. This is the emergency stopgap to get some protection in now; the multiplier is the planned next step and I'll add a backlog ticket for it.

PR #5107, opened 2026-08-03
fix(rustffi): reject empty disconnect lists

iverc:
## Summary

- reject empty address lists in the Rust FFI disconnect wrapper before per-address dispatch
- return the same `addresses cannot be empty` validation error expected from other clients
- complete the peer-info acceptance coverage for [sourcenetwork/defradb.rs#1261](https://github.com/sourcenetwork/defradb.rs/pull/1261) and [sourcenetwork/defradb.rs#1257](https://github.com/sourcenetwork/defradb.rs/issues/1257)

## Verification

- [x] `TestNetInfoDisconnectEmptyList`
- [x] `net/info`: 7 passed, 0 failed
- [x] `net/simple/replicator`: 29 passed, 0 failed
- [x] `gofmt`
- [x] `git diff --check`

### DIFF
--- a/tests/clients/rustffi/wrapper.go
+++ b/tests/clients/rustffi/wrapper.go
@@ -2051,6 +2051,9 @@ func (w *Wrapper) Disconnect(
 	addresses []string,
 	opts ...options.Enumerable[options.DisconnectOptions],
 ) error {
+	if len(addresses) == 0 {
+		return fmt.Errorf("addresses cannot be empty")
+	}
 	opt := utils.NewOptions(opts...)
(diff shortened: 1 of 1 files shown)


nasdf (commented):
This should be fixed on the latest develop branch. The error was moved into the DB package so all clients return the same error.

https://github.com/sourcenetwork/defradb/blob/develop/internal/db/p2p.go#L70

iverc:
That DB fix is already in this branch, but the Rust FFI test wrapper does not call the DB method; it loops over the single-address FFI method directly. With an empty slice the loop is skipped and the wrapper returns nil, so it still needs this check for client parity.

PR #5109, opened 2026-08-03
feat: reclaim orphaned blocks left by discarded merges

mennatnaga:

Stacked on #5146.

Blocks written outside a merge transaction, by `sync_dag.go` and `car.go`, are unreachable if that merge never completes. `hardDeleteDocumentBlocks` walks out from a document's heads, so history prune can only reach blocks an existing document owns. Nothing else reclaims this class.

The sweep expires the to-merge markers left behind and removes a block only when no committed document owns it. A marker on a block that did merge is dropped and the block kept. A merge that takes ownership clears the marker in the same transaction, so a sweep holding that marker conflicts rather than reclaiming the block.

A signature block is not in `AllLinks`, so neither merge path cleared its marker: `updateHeads` never sees it, and a field the collection does not define returns before reaching `updateHeads`. Both now clear it.

The marker records the first fetch and nothing renewed it, so a block re-received while still unmerged kept ageing toward the cutoff. `Put` and `PutMany` now refresh it, but only when a marker is still there, since an absent one means the block merged and must stay merged. The check and the write share a transaction, so a merge clearing the marker in between is not undone.

A chunked store appends a suffix to every key it writes, so `cid.Cast(marker[1:])` rejected the key and the sweep reclaimed nothing. The CID is read off the front of the marker instead, and a chunked value is deleted by prefix because it spans several keys.

It runs by default. Because it deletes blocks, it can be retuned or turned off on a running node without a rebuild: `DEFRA_ORPHAN_GC_DISABLED`, `DEFRA_ORPHAN_GC_INTERVAL` (default 5m) and `DEFRA_ORPHAN_BLOCK_TTL` (default 30m), which is how old a marker must be before its block is eligible.

Also moves the block-owner index reads into their own package. The sweep lives in `internal/datastore` and needs those reads, but `internal/db/id` imports `internal/datastore`, so calling into it directly is an import cycle.

### DIFF
--- a/internal/datastore/blockstore.go
+++ b/internal/datastore/blockstore.go
@@ -12,6 +12,8 @@ package datastore
 
 import (
 	"context"
+	"encoding/binary"
+	"time"
 
 	ipfsBlockstore "github.com/ipfs/boxo/blockstore"
 	blocks "github.com/ipfs/go-block-format"
@@ -49,8 +51,10 @@ type bstore struct {
 var _ Blockstore = (*bstore)(nil)
 
 const (
-	objectMarker       = byte(0xff)
 	toMergeIndexPrefix = byte('m')
+	// toMergeValueLen is the length of a current-format marker value: an 8-byte
+	// big-endian unix timestamp of when the block was fetched.
+	toMergeValueLen = 8
 )
 
 func newToMergeKey(cid []byte) []byte {
@@ -59,6 +63,24 @@ func newToMergeKey(cid []byte) []byte {
 	copy(key[1:], cid)
 	key[0] = toMergeIndexPrefix
 	return key
+}
+
+// newToMergeValue encodes the time a block was fetched. Only the orphan sweep reads
+// this value, to tell a fetch still in flight from an abandoned one; IsMerged checks
+// the marker's presence, not its contents.
+func newToMergeValue(t time.Time) []byte {
+	v := make([]byte, toMergeValueLen)
+	binary.BigEndian.PutUint64(v, uint64(t.Unix()))
+	return v
+}
+
+// toMergeTime decodes a marker value. A value that is not a full timestamp (an older
+// single-byte marker) decodes as (zero, false), and the sweep leaves those alone.
+func toMergeTime(v []byte) (time.Time, bool) {
+	if len(v) != toMergeValueLen {
+		return time.Time{}, false
+	}
+	return time.Unix(int64(binary.BigEndian.Uint64(v)), 0), true
 }
 
 // IsMerged reports whether the block is stored and carries no to-merge marker. Callers
@@ -108,7 +130,7 @@ func (bs *p2pBlockStore) Put(ctx context.Context, block blocks.Block) error {
 	if err == nil && exists {
 		return nil // already stored.
 	}
-	err = bs.store.Set(ctx, newToMergeKey(block.Cid().Bytes()), []byte{objectMarker})
+	err = bs.store.Set(ctx, newToMergeKey(block.Cid().Bytes()), newToMergeValue(time.Now()))
 	if err != nil {
 		return NewErrStoreBlock(err)
 	}
@@ -126,7 +148,7 @@ func (bs *p2pBlockStore) PutMany(ctx context.Context, blocks []blocks.Block) err
 		if err == nil && exists {
 			continue
 		}
-		err = bs.store.Set(ctx, newToMergeKey(b.Cid().Bytes()), []byte{objectMarker})
+		err = bs.store.Set(ctx, newToMergeKey(b.Cid().Bytes()), newToMergeValue(time.Now()))
 		if err != nil {
 			return NewErrStoreBlock(err)
 		}
--- a/internal/datastore/blockstore.go
+++ b/internal/datastore/blockstore.go
@@ -12,6 +12,8 @@ package datastore
 
 import (
 	"context"
+	"encoding/binary"
+	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	ipfsBlockstore "github.com/ipfs/boxo/blockstore"
@@ -65,8 +67,10 @@ type bstore struct {
 var _ Blockstore = (*bstore)(nil)
 
 const (
-	objectMarker       = byte(0xff)
 	toMergeIndexPrefix = byte('m')
+	// toMergeValueLen is the length of a current-format marker value: an 8-byte
+	// big-endian unix timestamp of when the block was fetched.
+	toMergeValueLen = 8
 )
 
 func newToMergeKey(cid []byte) []byte {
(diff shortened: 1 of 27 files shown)


internal/datastore/blockstore.go
```diff
@@ -108,7 +130,7 @@ func (bs *p2pBlockStore) Put(ctx context.Context, block blocks.Block) error {
 	if err == nil && exists {
 		return nil // already stored.
 	}
-	err = bs.store.Set(ctx, newToMergeKey(block.Cid().Bytes()), []byte{objectMarker})
+	err = bs.store.Set(ctx, newToMergeKey(block.Cid().Bytes()), newToMergeValue(time.Now()))
```
shahzadlone, line 140: todo: Could an existing unmerged block renew its marker before returning, while preserving merged-block markers?

Can likely i think do a production like sweep probe and see if the old timestamp is unchanged after an active re-receive and reclaimed the block. Could `Put` refresh it before returning?

What i mean is like `Put` returns as soon as the block bytes already exist, so a re-received block never refreshes its to-merge marker, does that make sense?

internal/datastore/blockstore_gc.go
```diff
@@ -0,0 +1,236 @@
...
+		if err != nil {
+			return err
+		}
+		if t, decoded := toMergeTime(value); !decoded || !t.Before(cutoff) {
+			// Untimestamped, or re-written since the scan.
+			continue
+		}
+
+		// A marker key is the to-merge prefix followed by the block's CID.
+		blockCID, err := cid.Cast(marker[1:])
```
shahzadlone: todo: Could the sweep operate through the configured chunk-aware blockstore or decode its suffix before ownership and deletion?

can likely do a manual test to see if `marker[1:]` is scanned with one expired chunked marker and zero blocks reclaimed (as i think parses marker[1:] as a CID, but chunked storage appends a suffix to that physical key)

internal/datastore/blockstore.go
mennatnaga: Yes, you're right, and PutMany has the same early return - I added a refresh that restamps the marker, guarded on one still being there so a merged block stays merged.

The check and the set share a transaction, otherwise a Put racing MarkAsMerged puts the marker back on a
block that just merged.

This works on the car path as importCAR puts every block in the car, but the bitswap path doesn't reach it though, blockservice checks Has before calling Put, so a block we already hold never gets there. Will fix that on a follow-up.

internal/datastore/blockstore_gc.go
mennatnaga: Confirmed, and I get the numbers you predicted: with a chunk size set the sweep scans 1 and reclaims 0, and without one it reclaims 1.

Going through the chunk-aware store doesn't work, its iterator cuts every key down to chunkKeyLen, which we set to the length of a CID written as base32, while the keys in the blockstore are raw CID bytes, so it panics - I decoded the suffix instead.

And it's not only the parse, a chunked store can split one value across several keys, so the delete has to remove all of them and not just the first.

internal/datastore/blockstore_gc.go
```diff
@@ -0,0 +1,236 @@
...
+// never reaches these blocks because it walks out from documents, so without this sweep
+// they accumulate for the life of the store.
+//
+// It resumes from startKey, nil beginning at the start of the marker index, and scans at
+// most scanLimit markers per call. The scan only nominates candidates: it collects them
+// without mutating the iterator, then reclaimBatch re-decides each one under the
+// transaction that deletes it.
+//
+// Only markers carrying a timestamp are eligible. A marker without one was written before
+// the index recorded fetch times, by a store that also predates the block-owner edges this
```
shahzadlone: question: Correct me if wrong, but this new comment implies that every "untimestamped" marker predates the owner index, but IIRC didn't the owner edges land before this PR while single-byte markers still existed?

I think it's easy to test if the unowned legacy marker and block stored give us a `reclaimed=0`.

suggestion: Could we stamp and age unowned legacy markers while retaining the ownership check?

node/store_badger.go
```diff
@@ -172,3 +238,57 @@ func (s *badgerStore) reclaimValueLog() {
...
+//
+// The first sweep is held back one TTL, so nothing can be reclaimed sooner than that after
+// the store opens.
+func (s *badgerStore) runOrphanBlockGC(ctx context.Context) {
+	defer s.wg.Done()
+
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(s.orphanBlockTTL):
```
shahzadlone, line 254: suggestion: Could the first `runOrphanBlockGC` pass start after one interval and rely on the `orphanBlockTTL` cutoff for safety? (in other words `runOrphanBlockGC` waits a full `orphanBlockTTL` before creating its interval ticker i believe).

nitpick: If you try a reopen probe/test manually it might show an orphan already older than TTL (can unswept through ten 20 ms intervals after open), so this way you can see restart heavy nodes can perform no reclamation before the next TTL elapses?

internal/datastore/blockstore_gc.go
shahzadlone: > And it's not only the parse, a chunked store can split one value across several keys, so the delete has to remove all of them and not just the first.

Sorry, I didn't 100% follow what are you suggesting?

internal/datastore/blockstore.go
shahzadlone: > Yes, you're right, and PutMany has the same early return - I added a refresh that restamps the marker, guarded on one still being there so a merged block stays merged.
> 
> The check and the set share a transaction, otherwise a Put racing MarkAsMerged puts the marker back on a block that just merged.
> 
> This works on the car path as importCAR puts every block in the car, but the bitswap path doesn't reach it though, blockservice checks Has before calling Put, so a block we already hold never gets there. Will fix that on a follow-up.

Makes sense to me if i am following properly! Thanks Menna for taking care of this.

internal/datastore/blockstore_gc.go
mennatnaga: Ohh you're right, my bad.

On stamping legacy markers, doesn't checking ownership straight away get the same answer with no new write path? as a single-byte marker can only have come from an older build, so it can't be a fetch that's still going.

node/store_badger.go
mennatnaga: afaik the cutoff doesn't cover it, as markers store wall-clock time and the cutoff is now minus the ttl, so neither knows the node was down, so after a long shutdown everything written before it is already expired and goes on the first tick.

But I think it's fine to remove it, as reclaiming early costs a re-fetch, while keeping the delay risks never reclaiming at all.

internal/datastore/blockstore_gc.go
mennatnaga: I meant to describe what I fixed in the latest commit I recently pushed, the "fix: sweep markers written through a chunked store" commit.

There were two problems on a chunked store, the parse was the first: cid.Cast(marker[1:]) fails because the physical key carries the chunk suffix. The delete was the second: a chunked store splits a large value across more than one key, so deleting the block's key on its own leaves the other pieces behind.

deleteKey now takes a chunked flag, on an unchunked store it deletes the single key; on a chunked one it scans the prefix and deletes every key under it. The test checks the prefix is empty afterwards rather than checking one key, which is what catches a partial delete. Please lmk if I missed something.

internal/datastore/blockstore_gc.go
shahzadlone: I am not a 100% but i think that makes sense

vertexclique:
We have achieved this at Rust side please check the implementation and proofs related to it.
We are not necessarily pushing the version gc to background and making it timer based, ratther than set is tracked across until threshold is passed. [(which is proven under every concurrency condition by the both TLA and Lean proofs)](https://github.com/sourcenetwork/defradb.rs/pull/1622/changes#diff-3fee09b2f701e11e15f17ef3aefe68c6a6b0fc95419bdc4f4fb84b1e1288e5f0)

* [HeadSet mechanism with inlined GC](https://github.com/sourcenetwork/defradb.rs/pull/1622/changes#diff-d85018ab42765abc2d5fc1fdc31ff6d2f58e5fa55504e4a5b0070fc24556293d)
* [HeadSet mechanism](https://github.com/sourcenetwork/defradb.rs/pull/1622/changes#diff-61c6185b94b50a452e115eab710f7a8c3d197919ec419fc2147cd4101b8f4340)

cc/ @mennatnaga @shahzadlone @iverc

internal/datastore/blockstore_gc.go
mennatnaga: Okay I looked more into this more and ownership on its own wouldn't cover it, signature blocks get a marker but never an owner edge, so a live one reads as unowned and we'd end up deleting it. Keeping the skip and fixed the comment, and opened #5218 to follow up.

dvncan (approved):
g2g

PR #5111, opened 2026-08-04
fix: Isolate document short ID allocation (#5101)

### LINKED ISSUE #5100: P2P may silently drop concurrent document merges

mennatnaga:

Backport of #5101. Resolves #5100 for shinzo.

Document short IDs were allocated inside the caller's transaction, so the sequence key joined that transaction's conflict set and concurrent merges contended on it. `reserveDocShortID` allocates in its own transaction under a mutex, so the sequence commits independently of the merge that needs the ID.

A retry budget below one meant zero attempts, so the merge never ran, and an exhausted budget returned nil instead of an error.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -113,6 +113,7 @@ type DB struct {
 
 	docMergeQueue *mergeQueue
 	colMergeQueue *mergeQueue
+	docShortIDMu  sync.Mutex
 
 	p2p *p2p.P2P
 	// Retry intervals when a replicator failure occurs.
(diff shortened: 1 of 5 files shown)


tzdybal (approved):
I can confirm this change is a backport of #5101, except the changes the `shortID` into `atomic.Uint32 (which doesn't exist on our shinzo branch)

tzdybal (approved):

PR #5112, opened 2026-08-04
test(i): Generalize per-node test config into a struct

islamaliev:
## Relevant issue(s)

Refs #5066

## Description

Setting up a test node took a function that returned networking options, so
networking was the only thing a test could say about a node. When nodes needed to
run from an older release, the version had nowhere to go and became its own
separate action next to the first one.

Both are now one struct holding the version and the networking config as plain
fields. The function that builds a random networking config returns that struct
instead, so the several hundred places that call it are unchanged. Only the two
tests that asked for an older node were edited by hand.

Plain fields also matter for the multiplier promised in #5103, which rewrites
actions to re-run tests under different settings. It can rewrite values but
cannot see inside a function.

## Tasks

## How has this been tested?

Nothing should change, so the existing suite is the check. Replicator tests under
both clients, node access control tests under the HTTP client, both cross version
tests against a real v1.0.0 binary, and the full query suite. Plus `go vet
./tests/...`, which type checks every call site, the wasm build, and lint.

Worth a look in review: the merged setup function now runs the platform skip and
the version assignment on both paths, not just the older node path. Both do
nothing on the normal path, but it is a deliberate widening.

- MacOS

### DIFF
--- a/tests/integration/db.go
+++ b/tests/integration/db.go
@@ -84,7 +84,7 @@ func defaultNodeOpts() *options.NodeOptionsBuilder {
 		// The test framework sets this up elsewhere when required so that it may be wrapped
 		// into a [client.TxnStore].
 		SetDisableAPI(true).
-		// The p2p is configured in the tests by [ConfigureNode] actions, we disable it here
+		// The p2p is configured in the tests by [NodeConfig] actions, we disable it here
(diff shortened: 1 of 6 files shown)


jsimnz (approved):
lgtm

PR #5114, opened 2026-08-04
fix: Make skipBackupTests behave with multiple clients

### LINKED ISSUE #5113: skipBackupTests doesn't behave correctly with multiple test clients
Coderabbit picked up on this one. In the `client_setup.go` and `client_setup_js.go` files, the C and JS clients set the `skipBackupTests` value to true. However, `skipBackupTests` is package-wide. The `ExecuteTestCase` function inside `tests/integration/utils.go` invokes `skipIfBackupTest` *before* iterating over the clients. So the problem is, if one client that skipped these tests were run alongside a client that didn't, they would be skipped for all clients.

This is not something that happens in our Git action workflows, so it doesn't affect our testing. But it is a legitimate bug. It could be fixed by moving the check out to a per-client execution path, or by another appropriate method.

ChrisBQu:
## Relevant issue(s)

Resolves #5113 

## Description

There was an issue with how skipping backup tests was being handled. There was a variable being set during initialization of a client. For example, initialization of the C or JS clients would set the value to true, and then during test execution, any backup-related tests would be skipped. This plays nicely with the way our CI jobs are isolated, but is actually incorrect. It is actually possible to run tests with multiple clients, some of which support backup tests, and some of which do not. And currently, they would be skipped for _all_ clients.

This fixes that by removing the variable, and instead filtering the list of clients to execute the test actions to the ones that can actually do so. This occurs inside the `utils.go` file.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/tests/integration/client_setup.go
+++ b/tests/integration/client_setup.go
@@ -32,7 +32,6 @@ func init() {
 	}
 	if cClient {
 		skipNetworkTests = false
-		skipBackupTests = true
 	}
 }
 
--- a/tests/integration/client_setup_js.go
+++ b/tests/integration/client_setup_js.go
@@ -26,8 +26,6 @@ func init() {
 	jsClient = true
 	// JavaScript networking stack is managed externally
 	skipNetworkTests = true
-	// Backup API is not suitable for browser environments
-	skipBackupTests = true
 }
 
 // setupClient returns the client implementation for the current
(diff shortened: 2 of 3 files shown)


nasdf (approved):
Thanks for fixing this!

shahzadlone (approved):
LGTM

PR #5118, opened 2026-08-04
test(i): Move node setup into the action package

islamaliev:
## Relevant issue(s)

Refs #5066

## Description

The harness drops anything that is not an action before handing the list to the
multiplier engine. Node config was a plain struct, so it never got there and no
multiplier could change how a node is set up. One that tried would register, run,
pass, and do nothing.

Node config is now an action and its setup code became the Execute method. What
it needed came with it: client setup, sourcehub setup, the node option defaults,
and the js versions of each. The test case type stayed put, since node setup only
reads two fields off it.

Sourcehub setup used to look through the action list for document ACP actions,
and those types are still in the integration package. Moving it as it was would
have needed an import going back the other way, so the caller works out whether
the test needs sourcehub and passes a bool instead.

Nodes behave the same, and a second copy of the identity helpers is gone.

## Tasks

## How has this been tested?

Nothing should change, so the existing suite is the check. All net and collection
version packages, encryption, index, searchable encryption and signature. Node
access control under the HTTP client, which guards native address discovery. Both
cross version tests against a real v1.0.0 binary. Plus `go vet ./tests/...`, the
wasm build, gofmt and lint.

Worth a look in review: the type switch case for node config is gone, since the
generic action case now handles it. That is the point of the change, but it fails
quietly if wrong, so there is a test for it.

- MacOS

### DIFF
--- a/tests/action/acp_dac_config.go
+++ b/tests/action/acp_dac_config.go
@@ -1,0 +1,45 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// This file is part of the DefraDB test suite.
+//
+// The DefraDB test suite is licensed under either:
+//
+//   (1) GNU Affero General Public License v3
+//   (2) Business Source License 1.1
+//
+// See tests/LICENSE for details.
+
+package action
+
+import (
+	"os"
+
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+const (
+	documentACPTypeEnvName = "DEFRA_DOCUMENT_ACP_TYPE"
+	sourcehubImageEnvName  = "DEFRA_SOURCEHUB_IMAGE"
+)
+
+var (
+	// DocumentACPType is the document ACP implementation under test.
+	//
+	// Node setup and the test harness both read this, so it is resolved once
+	// here rather than copied into each package.
+	DocumentACPType state.DocumentACPType
+
+	// sourcehubImage is the container image used to run SourceHub.
+	sourcehubImage string
+)
+
+func init() {
+	DocumentACPType = state.DocumentACPType(os.Getenv(documentACPTypeEnvName))
+	if DocumentACPType == "" {
+		DocumentACPType = state.LocalDocumentACPType
+	}
+	sourcehubImage = os.Getenv(sourcehubImageEnvName)
+	if sourcehubImage == "" {
+		sourcehubImage = "ghcr.io/sourcenetwork/sourcehub:dev"
+	}
+}
--- a/tests/action/node_config.go
+++ b/tests/action/node_config.go
@@ -1,0 +1,115 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// This file is part of the DefraDB test suite.
+//
+// The DefraDB test suite is licensed under either:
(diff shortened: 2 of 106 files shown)


tests/action/acp_dac_config.go
```diff
@@ -0,0 +1,45 @@
...
+// This file is part of the DefraDB test suite.
+//
+// The DefraDB test suite is licensed under either:
+//
+//   (1) GNU Affero General Public License v3
+//   (2) Business Source License 1.1
+//
+// See tests/LICENSE for details.
+
+package action
```
AndrewSisley: thought: The stuff in this file does not belong in the action package - these are environment variables specific to the execution environment, they should not be preserved and shared across execution environments in the same way that actions are.  They are also not actions, and so don't really belong in this package regardless of whether or how they are shared or not.

suggestion: Move these back into the `integration` package.

tests/action/node_config.go
```diff
@@ -0,0 +1,115 @@
...
+func (a *NodeConfig) WithVersion(version string) *NodeConfig {
+	clone := *a
+	clone.Version = version
+	return &clone
+}
+
+// Execute configures and starts a new Defra node.
+//
+// Any errors generated during configuration will result in a test failure.
+func (a *NodeConfig) Execute() {
```
AndrewSisley: suggestion: I think the name of this type is out of step with the other actions.  This is really a `NewNode` action, executing it creates a new node.  Please consider renaming it.

tests/action/node_config.go
```diff
@@ -0,0 +1,115 @@
...
+// config was supplied.
+func (a *NodeConfig) P2POptions() options.NodeP2POptions {
+	if a.Network == nil {
+		return options.NodeP2POptions{}
+	}
+	return a.Network()
+}
+
+// WithVersion returns a copy of the config that runs the node as an external process
+// from the given published release, e.g. "v1.0.0".
```
AndrewSisley: suggestion: This function only performs a shallow clone, there are inner elements that are not cloned - I suggest noting this in the description.

tests/action/node_options.go
```diff
@@ -0,0 +1,169 @@
...
+	"time"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client/options"
+	changeDetector "github.com/sourcenetwork/defradb/tests/change_detector"
+	"github.com/sourcenetwork/defradb/tests/state"
+)
+
+const (
```
AndrewSisley: suggestion:  Similar to my other comment, I quite strongly believe that these vars/consts/int-logic do not belong in this package, and I would much prefer that they are moved back in to the `integration` package.

AndrewSisley (approved):
I would much prefer that a lot of what is being moved, is not moved, but will not block the PR over it.

tests/action/node_config.go
shahzadlone: I don't really mind it, is a somewhat established naming pattern in the `action` package.

shahzadlone (approved):
LGTM

PR #5119, opened 2026-08-04
fix: Restore CI baseline

iverc:
## What this does

- update Go from 1.25.9 to 1.25.13, removing all currently reachable standard-library findings
- update Pion DTLS to 3.1.4, removing the reachable crafted-handshake panic
- replace the stale vulnerability-count assertion with an exact allowlist of the six known upstream findings
- regenerate the committed SDL fixtures after the HNSW schema additions
- isolate system-keyring tests with temporary config roots and clean up partial key writes
- keep LevelDB document IDs and index cleanup in the caller transaction
- discard explicit integration transactions before closing their nodes
- normalize active peer addresses to peer IDs before matching sync responses
- wait for observed pubsub readiness in branchable-sync, signed-document, simple-sync, and multi-hop peer tests
- give the branchable-sync test action the same five-second response window as the implementation
- retry transient transaction conflicts for implicit integration requests

CIRCL cannot yet be upgraded to its fixed release because ACP Core's current ssi-sdk dependency uses the removed Dilithium API. The remaining OpenPGP, libp2p DHT, and Cosmos findings also have no compatible fixed version today. Every reviewed ID is explicit; new, removed, or changed findings remain a required CI failure.

## Validation

- exact vulnerability allowlist and `govulncheck ./...`: six reviewed IDs, no standard-library or Pion DTLS findings
- NPX-tagged CLI and GraphQL schema packages
- repository-wide compile check and `go vet ./...`
- `go mod verify` and `go mod tidy -diff`
- full LevelDB transaction integration package
- concurrent document and merge short-ID tests under the race detector
- exact failed branchable-sync matrices: 20/20 HTTP and 10/10 CLI race-enabled runs
- multi-hop peer propagation: 100/100 focused race-enabled runs and five full package runs
- simple branchable sync: 100/100 basic and 50/50 secondary-index race-enabled runs
- migration update requests: 500/500 secondary-index race-enabled runs
- action helper package under the race detector
- signed-document peer package under the race detector
- changed-code lint: 0 issues

Before this change, the LevelDB transaction case deadlocked past 45 seconds; it now completes in about 1.7 seconds. The peer tests now wait on observable readiness rather than scheduler-dependent sleeps.


cli/wizard/unit_test.go
```diff
@@ -277,9 +279,12 @@ func Test_GenerateIndividualKeyrings_SystemKeyring(t *testing.T) {
...
 	openKeyring := keyring.OpenSystemKeyring(keyringNamespace)
-	if err != nil {
-		t.Fatalf("failed to reopen keyring: %v", err)
-	}
+	t.Cleanup(func() {
+		_ = openKeyring.Delete("node-identity-key")
+		_ = openKeyring.Delete("peer-key")
+		_ = openKeyring.Delete("encryption-key")
+		_ = openKeyring.Delete("searchable-encryption-key")
+	})
```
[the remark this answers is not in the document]
iverc: Fixed in 5e23695c5. Every system-keyring test now registers cleanup before its first callback, including the two existing fixtures that previously deleted keys only after all assertions passed.

iverc:
The macOS rerun exposed three test-readiness issues plus the open Go race-runtime crash:

- sync tracked pending peers by multiaddress while responses carry peer IDs, so successful requests waited for the five-second timeout
- active-peer assertions and document-sync requests could run before network/topic convergence
- one heavier CLI P2P case inherited a one-second whole-test deadline

`e95de3df3` fixes those contracts and keeps macOS race detection enabled. The macOS job may retry one failed test once for [golang/go#80292](https://github.com/golang/go/issues/80292); data races, multiple failures, and repeated failures still fail the job.

The formerly flaky single-node sync passed 20/20 race runs in 7.6s, down from 105.6s before. All document-sync cases, branchable sync, network-info, NAC P2P, CLI P2P, and the affected package aggregate pass under shuffled race runs.

iverc:
Addressed the remaining LevelDB sequence-coverage nit in `3b4789b81`: the regression now proves a committed reservation advances the sequence and a discarded reservation rolls it back. The exact test passed 50 consecutive race-enabled runs.

iverc:
@coderabbitai review

iverc:
Covered both `activePeerIDs` error paths in `133006116`: host errors are asserted with `ErrorIs`, malformed addresses are rejected, and the existing peer-ID deduplication case remains. The focused cases passed 50 consecutive race-enabled runs, the full P2P package is green under the race detector, and lint reports 0 issues.

iverc:
@coderabbitai review

.github/known-vulnerabilities.txt
```diff
@@ -0,0 +1,6 @@
+GO-2023-1821
```
AndrewSisley, line 1: todo: Please add documentation detailing how to look these up.

tests/action/request.go
```diff
@@ -133,6 +133,14 @@ nodeLoop:
 		var result *client.RequestResult
 		if hadTxn {
 			result = txn.ExecRequest(a.s.Ctx, request, reqOption)
+		} else if a.ExpectedError == "" {
+			_ = withRetryOnNode(node, func() error {
```
AndrewSisley, line 137: todo: Requests are not read-only actions, retrying them is not going to be sensible in a lot of cases.

Please either rework this, or add a code comment detailing why you think this is sensible/required.

cli/test/integration/p2p/replicator_list_test.go
```diff
@@ -131,6 +132,7 @@ func TestReplicatorGetAll_WithMultipleCollectionsnAndDeleteACollection_ShouldRet
 
 func TestReplicatorGetAll_WithMultipleCollectionsAndMultiplePeers_ShouldSucceed(t *testing.T) {
 	test := &integration.Test{
+		Timeout: 5 * time.Second,
```
AndrewSisley, line 135: todo: Please document why this test needs this timeout

AndrewSisley (changes requested):
Hey Ivan, thanks for the fixes.

By my counting, this PR includes 4 user-visible changes (3 in this codebase, and one via the `pion` version bump).  The PR title does not reflect this, and as such, neither will the change log.

We are post-v1 now, and IMO we should be fairly strict in letting users know the behaviour that we have changed.  If you think this is unnecessary, please open a discussion thread on discord and tag the db and dev-ex teams, otherwise, please split up this PR and name them accordingly - it also makes it much easier for reviewers to figure out what is going on as multiple changes are not tangled up together in the PR diff.

In the past some of the team has also been fairly strict about linking PRs to issues.

cli/test/integration/p2p/replicator_list_test.go
shahzadlone: question: Why do we even need this?

suggestion: Is there a way to avoid having this?

.github/known-vulnerabilities.txt
shahzadlone, line 1: suggestion: Not 100% sure if I like polluting the `.github` directory.

How about either place it with the workflow that is using this in the `.github/workflows` directory or in:  `tools` folder.

internal/db/db.go
```diff
@@ -263,14 +264,16 @@ func (db *DB) NewTxn(readonly bool) (client.Txn, error) {
...
-// document transactions do not contend on the sequence key. Failed document transactions may leave
-// gaps; short IDs only need to be unique. The lock spans transaction creation through commit so each
-// transaction sees the last committed reservation.
+// reserveDocShortID commits the node-wide sequence separately from document writes where the store
+// supports concurrent transactions. LevelDB can only hold one transaction, so it uses the caller's.
 func (db *DB) reserveDocShortID(ctx context.Context) (uint64, error) {
 	db.docShortIDMu.Lock()
 	defer db.docShortIDMu.Unlock()
 
+	if _, ok := db.rootstore.(*leveldb.Datastore); ok {
```
shahzadlone, line 273: question: Could the LevelDB reservation path use a capability rather than an exact `*leveldb.Datastore` concrete type?
for example right now this line would select the LevelDB caller-transaction path only when the dynamic root store is exactly `*leveldb.Datastore`

tests/integration/utils.go
shahzadlone, line 1: question: Could we make the integration retry helper use the same bounded-attempt and final-error semantics as the fixed action helper?

The retained overlay probe reproduced both assertions: tests/integration/utils.go:2018-2032 still swallows the final transaction conflict and skips the action when MaxTxnRetries is zero.

If i am not mistaken in tests/action/utils.go (around maybe these lines 154-172) I think it handles both cases.

Makefile
```diff
@@ -68,6 +68,7 @@ BUILD_FLAGS+=-tags $(BUILD_TAGS)
 endif
 
 TEST_FLAGS=-race -shuffle=on -timeout 10m
+GOTESTSUM_FLAGS?=
```
shahzadlone, line 71: question: What is this for?

.github/workflows/test-macos.yml
```diff
@@ -76,4 +76,10 @@ jobs:
         uses: ./.github/composites/setup-defradb
 
       - name: Run integration tests
+        # Work around https://github.com/golang/go/issues/80292 without suppressing data races.
+        env:
+          GOTESTSUM_FLAGS: >-
+            --rerun-fails=2
+            --rerun-fails-max-failures=1
+            --rerun-fails-abort-on-data-race
```
shahzadlone, line 84: question: What does this do? Please also document in addition to the link you added.

shahzadlone (commented):
question: Could we split this PR into independently reviewable changes?

Currently this PR has:
- changes to 2 different ci workflows (each should in my opinion have it's own `ci: xyz` titled PR)
- security/dependency updates
- LevelDB transaction semantics
- P2P response accounting
- test retry/readiness
- timeout changes
- keyring/teardown cleanup
- generated SDL fixtures

Some of these can be linked together and be one PR, but the makefile, ci and other stuff that is fundamentally different should definitely not be all in one in my opinion.

PR #5120, opened 2026-08-04
test: Fix system keyring fixtures

iverc:
## Summary

- isolate system-keyring tests with a temporary config root
- select the intended keyring backend and identity key types
- remove temporary keychain entries during cleanup

## Validation

- `go test ./cli/wizard -run SystemKeyring -count=1`
- `go test ./cli/wizard -race -shuffle=on -count=3`
- changed-code golangci-lint: 0 issues


iverc:
Folded into #5119 so the vulnerability and macOS fixes restore the CI baseline together.

PR #5131, opened 2026-08-11
feat: Add index type to C bindings NewEncryptedIndex

### LINKED ISSUE #5130: C bindings' NewEncryptedIndex silently drops type
In a separate PR, Shahzad discovered the following bug. I quote:

> Wouldn't NewEncryptedIndex forward only FieldName and silently drop EncryptedIndexDescription.Type?
> 
> thought: Then cbindings/encrypted_index_new.go:40-42 probably will be constructing a request with the zero type, which the Go implementation defaults to equality. basically then a non-empty requested type then silently changed instead of being preserved (or rejected as unsupported

ChrisBQu:
## Relevant issue(s)

Resolves #5130 

## Description

Previously, the C bindings' `NewEncryptedIndex` function had no way to pass in type. This was not _technically_ a gap in functionality yet, because we only support one type right now. However, the other clients have a way to pass in the type, so there was a discrepancy here. This PR addresses that by adding a new prrameter to the function signature in the C bindings package.

## Tasks

## How has this been tested?

The C client wrapper has been altered, and it does pass in the new parameter now. However, with only one supported type at the moment, there's not a whole lot more that can be done.


AndrewSisley (approved):
LGTM, but with the large caveat that this is a semver breaking change, and I'm not sure if we came to a conclusion as to how we want to handle this for the non-Go embedded clients.

shahzadlone (approved):
LGTM (the problem / methodology of how we deal with breaking change is one we need to tackle)

islamaliev (approved):

PR #5132, opened 2026-08-11
chore: conflict and ingest instrumentation

mennatnaga:
Stacked on the CAR import guard PR.

Adds the per-interval counters needed to see what an ingesting node is actually doing: queue depth and byte usage against their budgets, batch and document outcomes, wire-message duplication, CAR import and generation, DAG sync, and merge conflicts broken down by the class of key the losing transaction had read.

Three things this covers that were previously invisible:

- **Why a document was dropped.** `docsDropped` only ever counted merge-call failures; eleven earlier discard sites were counted nowhere and three of them logged nothing either. Each now names a reason, and documents deliberately not merged (already held, no access, filtered) are counted as `docsSkipped` rather than as loss, so merged plus dropped plus skipped accounts for every document in a request.
- **The inbound CAR path.** Every document this node stores arrives through `importCAR`, which had no counters at all, while all five CAR counters described the outbound side. An import that fails part-way leaves blocks in the store owned by no document, so it reports how many it had already written.
- **Whether relaying works.** `docsRelayed` was incremented on the same condition as the merge counter and before the send, so it restated the merge count and a total send failure would not have changed it. It now counts after the send, paired with `relayFailed`.

Reason sets are fixed rather than derived from error strings, so a remote peer cannot grow the maps. Repeated failures cost a counter increment; the first occurrence of each reason gets the log line.

### DIFF
--- a/internal/datastore/readkinds.go
+++ b/internal/datastore/readkinds.go
@@ -1,0 +1,145 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package datastore
+
+import (
+	"strings"
+	"sync/atomic"
+)
+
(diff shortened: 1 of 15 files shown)


mennatnaga:
Folded into #5125, which already carries the telemetry commit. Same concern, and the two overlapped on merge.go and p2p.go.

PR #5133, opened 2026-08-11
fix: shut down the P2P subsystem on DB.Close

mennatnaga:
Stacked on #5137.

`Close` closed the message queue to stop the workers. The pubsub dispatcher keeps calling the handler until libp2p tears the subscription down, which `Close` does not do, and a send on a closed channel is a ready case in a select rather than a fallthrough to the default arm, so a message arriving during shutdown panics the dispatcher. The queue is now left open and a separate signal ends the worker pool.

Not complete. `(*P2P).Close` has no production caller, so the panic is unreachable as things stand: `node.Peer` is `github.com/sourcenetwork/go-p2p`, so `n.peer.Close()` does not reach this one. Wiring `DB.Close` to shut the P2P subsystem down is what makes this matter, and that is not done here.

### DIFF
--- a/internal/db/p2p/p2p.go
+++ b/internal/db/p2p/p2p.go
@@ -233,6 +233,12 @@ type P2P struct {
 	msgQueueBytes atomic.Int64
 	// msgQueueMaxBytes caps msgQueueBytes. Zero or less disables the byte bound.
 	msgQueueMaxBytes int64
+
+	// stopAccepting ends the worker pool. The queue itself is never closed: the pubsub
+	// dispatcher keeps calling the handler until libp2p tears the subscription down, and a
+	// send on a closed channel is a ready case in a select, so closing it would panic the
+	// dispatcher rather than fall through to the default arm.
+	stopAccepting chan struct{}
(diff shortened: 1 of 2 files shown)


internal/db/p2p/p2p.go
```diff
@@ -677,6 +684,8 @@ func (p *P2P) pubSubMessageHandler(from string, topic string, msg []byte) ([]byt
 
 	select {
 	case p.msgQueue <- queuedMessage{req: req, size: size}:
+	case <-p.stopAccepting:
+		p.releaseQueueBytes(size)
 	case <-p.ctx.Done():
 		p.releaseQueueBytes(size)
```
jsimnz, line 734: question: Without this PR, shouldn't the `ctx.Done` case be able to handle this graceful shutdown if you cancel the context object referenced?

mennatnaga: you're right, DB.Close cancels p.ctx, so ctx.Done already covers this -  I will fix this once I finalise the other prs

PR #5136, opened 2026-08-12
fix: delete the document's primary key on purge

mennatnaga:

Stacked on #5145.

`purgeOneDoc` deletes a document's state by key prefix. The primary key is not a `DataStoreKey`, so those prefix deletes never reach it: `/pk` and `/p` diverge at the byte after `p`. What is left is a document marker with no document behind it.

Head backfill then walked the primary prefix and returned `ErrDocumentNotFoundOrNotAuthorized` on the first key it could not resolve, which stopped the backfill for every document sorting after it. One stale key was enough to skip the rest of the collection. It now skips such a key and reports how many it skipped.

### DIFF
--- a/internal/db/p2p/replicator.go
+++ b/internal/db/p2p/replicator.go
@@ -210,12 +210,17 @@ func (p *P2P) pushHeadsForAllDocs(ctx context.Context, col client.Collection, pe
 		}
 	}()
 
+	var stale int
 	for {
 		hasNext, err := iter.Next()
 		if err != nil {
 			return NewErrIterateReplicatorDocs(err)
 		}
 		if !hasNext {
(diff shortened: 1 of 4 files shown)


internal/db/p2p/replicator.go
```diff
@@ -210,12 +210,17 @@ func (p *P2P) pushHeadsForAllDocs(ctx context.Context, col client.Collection, pe
 		}
 	}()
 
+	var stale int
```
shahzadlone, line 213: nitpick: few words to explain the use of this variable would be nice.

shahzadlone (approved):
LGTM

iverc (approved):

PR #5137, opened 2026-08-12
fix: stop the document fetch from reading the next document's key

mennatnaga:

Stacked on #5147.

The document fetch scanned with a start/end range whose end was `prefix.PrefixEnd()`. corekv leaves that unbounded in the store and checks the end in Go, so the iterator reads the first key past the document before stopping. That key belongs to the next document, and reading it puts it in this transaction's read set.

A prefix iterator bounds the scan in the store, so the read set holds only the document being fetched.

`internal/db/fetcher/document.go` is identical on `develop`, so this applies there unchanged.

### DIFF
--- a/internal/db/fetcher_overread_test.go
+++ b/internal/db/fetcher_overread_test.go
@@ -1,0 +1,67 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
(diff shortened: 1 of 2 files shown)


internal/db/fetcher_overread_test.go
shahzadlone, line 1: praise: Thanks for the test :)

shahzadlone (approved):
LGTM, thanks for the test :)

iverc (approved):

PR #5139, opened 2026-08-12
fix: Make CollectionFieldDescription unmarshal JSON properly

### LINKED ISSUE #5138: CollectionFieldDescription UnmarshalJSON can misbehave for certain types
In a recent PR, Shahzad made the following observation:

> thought: Since the generic result decoder cannot preserve concrete values held in any. AddCollection (line 367), AddView (line 597), and GetCollections (line 743) round-trip CollectionVersion.Fields[].DefaultValue through JSON, which would maybe then change a DateTime default from time.Time to string and an Int default from int32 to float64 i think.
> 
> IMO what makes this more odd is that our test comparator as it stands seems to be more "relaxed" which masks these type changes, while callers that inspect or type-assert returned defaults observe values different from the direct Go client, and the corrupted version metadata is also retained by Collection.
> 
> suggestion: What if we decode collection defaults according to each field's Kind (or use a typed, lossless result codec) and add concrete type Java parity assertions for every supported default type through add, view, and get collection paths rather than relying on the relaxed result comparator?

I made the following response:

> I have looked into this. It is a real issue, but I think it runs deeper than the Java client. Inside client/collection_field_description.go is a function, on line 79, UnmarshalJSON. All clients besides the Go client should be affected by round-tripping through JSON here. I think that's really what needs to change, instead of trying to hack around it here.
> 
> In my head, the part we replace is line 88. Instead of doing this direct assignment, I envision using a switch statement, and on a field kind-basis, making sure the problematic types get unmarshalled correctly. I don't think it's difficult work, but I think it's out of scope for this PR (hitting this a lot, but I want to do this right.) Let me get a PR open for that, and try to stick it ahead of this PR. If you end up really hating the approach, we can talk it over(?)

ChrisBQu:
## Relevant issue(s)

Resolves #5138 

## Description

It was possible for this function to silently coerce one type into another during the unmarshalling process. This PR corrects that, and introduces some unit tests of the new functionality.

For more context on this issue, you can read the conversation available inside the linked issue.
## Tasks

## How has this been tested?

- WSL


shahzadlone (approved):
LGTM.

Praise: Thanks for the tests

PR #5141, opened 2026-08-12
fix: Properly handle blank timeout strings in C bindings

### LINKED ISSUE #5140: C client wrapper SyncBranchableCollection misbehaves with blank timeouts
This block of code, inside cbindings/wrapper.go is problematic:

`	deadline, hasDeadline := ctx.Deadline()
	timerStr := ""
	if hasDeadline {
		timerStr = time.Until(deadline).String()
	}
	cTimerStr := C.CString(timerStr)
	defer C.free(unsafe.Pointer(cTimerStr))

	cIdentity := optionToUintptr(opt.GetIdentity())
	defer C.FreeIdentity(cIdentity)

	res := ConvertAndFreeCResult(
		C.SyncP2PBranchableCollection(C.uintptr_t(w.handle), cCollectionID, cTimerStr, cIdentity))
`

In the case of a blank timeout string, a non-null value is passed to the exported C function. That is, instead of a null value, a valid but blank string is passed in. And under the hood, a nil check will fail, and the C function will try to parse it as a time, but fail to do so. This is a bug, but one we have gotten away with because no tests exist which would have caught it. The fix is simple, but proper resolution would also entail new test code.

The fix itself should, ideally, alter the C bindings rather than the wrapper. We would like a blank string to be treated as no timeout value.

ChrisBQu:
## Relevant issue(s)

Resolves #5140

## Description

There was an odd behavior in the C bindings. It was not technically a bug, but it was a foot gun. The function for syncing branchable collections could take in a timeout string, and could take in a null value to represent no deadline. However, a blank string is actually a valid, non-null value. This made it easy for users to trip up -- and indeed, our test wrapper messes up in this way. The only reason it didn't create a problem is because we didn't have any test that triggers this behavior.

This PR opts to adjust the C bindings themselves rather than the test wrapper. The argument is that a blank string _should_ be interpreted as no timeout value.

With regards to the gap in testing: this PR introduces two new tests, which specifically target the serializing of timeout strings in the C bindings. There is one which checks a finite deadline (behavior that always worked), and one which checks a blank deadline (which is now properly working.) To accommodate this, a `NoDeadline` flag had to be added to the test action, but the changes are hopefully relatively straightforward.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/p2p_collection_sync_branchable.go
+++ b/cbindings/p2p_collection_sync_branchable.go
@@ -34,13 +34,14 @@ func SyncP2PBranchableCollection(nodePtr C.uintptr_t,
 	if err != nil {
 		return returnC(returnGoC(1, err.Error(), ""))
 	}
-	if timeoutStr != nil {
-		timeout, err := time.ParseDuration(C.GoString(timeoutStr))
+	timeout := C.GoString(timeoutStr)
+	if timeout != "" {
+		timeoutDuration, err := time.ParseDuration(timeout)
 		if err != nil {
 			return returnC(returnGoC(1, err.Error(), ""))
 		}
 		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, timeout)
+		ctx, cancel = context.WithTimeout(ctx, timeoutDuration)
 		defer cancel()
 	}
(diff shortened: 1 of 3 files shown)


shahzadlone (approved):
LGTM

PR #5143, opened 2026-08-13
fix: merge in bounded chunks and relay only the documents that committed

mennatnaga:

Stacked on #5142.

A merge batch ran in one transaction, so a single unmergeable event discarded every other event in it. Merging now happens in chunks of 8, and a chunk that fails is re-run one event at a time, so a deterministic failure drops only the event that caused it.

`MergeBatchWithTxn` also returned no indication of which events committed, so the caller relayed and acknowledged events that were never stored. It now returns a slice parallel to the input, and the error names every dropped event.

A zero retry budget meant zero attempts, so no merge ran at all.

### DIFF
--- a/internal/db/merge_test.go
+++ b/internal/db/merge_test.go
@@ -14,9 +14,11 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
+	badgerds "github.com/dgraph-io/badger/v4"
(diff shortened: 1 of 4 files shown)


internal/db/merge_test.go
shahzadlone, line 1: praise: Thanks for these tests :)

shahzadlone (approved):
LGTM

PR #5144, opened 2026-08-13
fix: bound the pubsub message queue by bytes, not just slots

mennatnaga:

Stacked on #5143.

The inbound pubsub queue was bounded by slot count only, so its memory use was set by whatever the largest messages happened to be. It is now bounded by bytes as well, with the budget derived from the process memory limit.

Wire size alone does not bound what the queue holds: a decoded batch entry costs a fixed-size struct however little it carries. That per-entry cost is charged after the decode, and the decoder rejects a batch carrying more entries than the cap admits.

The budget comes from `GOMEMLIMIT`. The Go runtime does not read a container's memory ceiling, so a node without it set gets the fallback, which may sit above the ceiling it runs under.

### DIFF
--- a/internal/db/p2p/p2p.go
+++ b/internal/db/p2p/p2p.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"reflect"
 	"runtime/debug"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


internal/db/p2p/p2p.go
```diff
@@ -69,8 +72,36 @@ const (
...
+// accounting can be reversed once the message has been processed. The wire size stands
+// in for the retained size: the decoded payload is a copy of those same bytes.
+type queuedMessage struct {
+	req  *protocol.PushLogRequest
+	size int64
+}
+
+// queueByteBudget returns the byte budget for the incoming message queue, derived from
+// the process memory limit so a node with a smaller limit queues proportionally less.
+func queueByteBudget() int64 {
```
shahzadlone, line 124: question: Could `queueByteBudget` derive its fallback from an operator-configured limit or document that `GOMEMLIMIT` is required?

I wonder what a retained runtime probe shows for `math.MaxInt64` and a 1 GiB budget when no `GOMEMLIMIT` is set, a container below 1 GiB can maybe exceed its memory ceiling?

question: Could `queuedMessage` charge decoded `PushLogRequest` retained size, or use a conservative multiplier, instead of charging only `len(msg)`?

Seems to me that the quarter limit budget can retain roughly the full process limit after decoded headers and allocator overhead.

mennatnaga: you're right on both.

Re the fallback, I documented in the deployment guide rather than a config option.

Re the charge: yes, it's the decoded headers. A peer can send an empty cbor map, one byte on the wire and a whole struct once decoded, so the old handler charged far less than such a message holds, the decoder now caps how many entries a message can carry and charges each one its decoded cost, which brings the two into line.

PR #5145, opened 2026-08-13
fix: skip rewriting CAR blocks that are already stored

mennatnaga:

Stacked on #5144.

`importCAR` used a blind-write blockstore, on the premise that CAR blocks are new by definition. Field blocks are content-addressed and recur across documents sharing a field value, so a CAR routinely carries blocks the node already stores. Blind-writing them re-stamps a to-merge marker on a block that already merged, so it reads as an abandoned fetch afterwards. Those writes also commit individually, since `importCAR` holds no transaction, so a concurrent merge reading those block keys conflicts.

The blind-write blockstore is removed with it, since the change above was its last caller. The merged-CID cache in `internal/core/block` is removed too; it had no caller at all.


internal/db/p2p/car.go
```diff
@@ -170,7 +170,10 @@ func (p *P2P) importCAR(ctx context.Context, carData []byte) (*coreblock.Block,
 		return nil, ErrEmptyCARRoots
 	}
 
-	bstore := datastore.BlindWriteP2PBlockstoreFrom(p.db.Rootstore(), immutable.None[int]())
+	// Content-addressed field blocks recur across documents that share a field value, so a
+	// CAR routinely contains blocks already stored. The guarded store skips those instead of
+	// rewriting the block and re-stamping a to-merge marker on one that already merged.
+	bstore := datastore.P2PBlockstoreFrom(p.db.Rootstore(), immutable.None[int]())
```
shahzadlone, line 176: question: Correct me if i am wrong but looks like the current raw-layout check misses the block and makes it unmerged after the guarded re-put? 

suggestion: Could `importCAR` use `P2PBlockstoreFrom(..., immutable.None[int]())` with the configured chunk size for its existence check so a merged chunked block keeps its marker?

mennatnaga: You're right, importCAR passes None, so the check looks for the exact key, and on a chunked node the block sits under that key plus a suffix, the check misses and we write the block again with a fresh marker.

Fixing importCAR on its own wouldn't change anything there though. generateCAR and loadBlockLinks pass None too, and generateCAR is the one that decides whether a CAR gets sent: on a chunked node it can't find the blocks, errors out, and both callers fall back to dag sync. so a chunked node never sends a CAR and nothing reaches importCAR. All three need to move together, and importCAR needs a way to reach the chunk size that isn't exposed yet, so I'd rather do that as a follow-up than widen this one.

Deployments run disk badger, which is unchunked, and that's the path this pr fixes.

shahzadlone: >  I'd rather do that as a follow-up than widen this one

Happy with that :) perhaps make an issue for it and assign yourself so it doesn't slip through the cracks.

PR #5146, opened 2026-08-13
chore: merge, ingest and store telemetry

mennatnaga:
Related shinzonetwork/shinzo-host-client#365

Stacked on #5137.

Counters for the p2p ingest and merge paths, reported once per interval and reset on report, so each line carries a rate. Failures get their own lines, grouped by cause: dropped documents, CAR build and import failures, DAG walk failures.

Dropped inbound messages are counted per interval rather than logged per message, and reported at error level so they are visible on a node running above info.

Early returns left the batch and outcome counters untouched, so a rate could be built from a denominator that missed the batches which failed before reaching it. Every terminal path now updates them.

The value log GC line reports on-disk size on every pass, so store growth can be read from the log.

A unique-index violation named only the document that lost. It now also names the document already holding the value.

### DIFF
--- a/internal/db/index.go
+++ b/internal/db/index.go
@@ -438,6 +438,8 @@ func saveUniqueKey(
 	}
 
 	if len(val) != 0 {
+		// This read puts the unique key in the transaction's read set, so two transactions
+		// writing the same index value conflict at commit rather than at this check.
 		existing, err := txn.Datastore().Get(ctx, &key)
 		if err != nil && !errors.Is(err, corekv.ErrNotFound) {
 			return NewErrCheckUniqueIndexConstraint(err)
@@ -446,7 +448,7 @@ func saveUniqueKey(
 			if tolerateSameDoc && string(existing) == string(val) {
 				return nil
 			}
-			return newUniqueIndexError(doc, fieldsDescs)
+			return newUniqueIndexError(ctx, doc, fieldsDescs, existing)
 		}
 	}
 
@@ -456,8 +458,33 @@ func saveUniqueKey(
 	return nil
 }
 
-func newUniqueIndexError(doc *client.Document, fieldsDescs []client.CollectionFieldDescription) error {
-	kvs := make([]errors.KV, 0, len(fieldsDescs))
+// incumbentDocID resolves the document already holding a unique index entry, from the
+// short ID the entry stores. Returns an empty string if it cannot be resolved, since this
+// only decorates an error that is being returned either way.
+func incumbentDocID(ctx context.Context, encodedShortID []byte) string {
+	shortID, err := keys.DecodeDocShortID(encodedShortID)
+	if err != nil {
+		return ""
+	}
+	docID, found, err := id.GetDocID(ctx, shortID)
+	if err != nil || !found {
+		return ""
+	}
+	return docID
+}
+
+func newUniqueIndexError(
+	ctx context.Context,
+	doc *client.Document,
+	fieldsDescs []client.CollectionFieldDescription,
+	existing []byte,
+) error {
+	kvs := make([]errors.KV, 0, len(fieldsDescs)+1)
+	// Naming the document that already holds the value is what makes the two comparable;
+	// without it the error only says which one lost.
+	if incumbent := incumbentDocID(ctx, existing); incumbent != "" {
+		kvs = append(kvs, errors.NewKV("HeldBy", incumbent))
+	}
 	for iter := range fieldsDescs {
 		fieldVal, err := doc.TryGetValue(fieldsDescs[iter].Name)
 		var val any
--- a/internal/db/index.go
+++ b/internal/db/index.go
@@ -14,6 +14,7 @@ import (
 	"context"
 
 	"github.com/sourcenetwork/corekv"
+	"github.com/sourcenetwork/corelog"
 
 	"github.com/sourcenetwork/defradb/client"
 	"github.com/sourcenetwork/defradb/errors"
@@ -448,7 +449,17 @@ func saveUniqueKey(
 			if tolerateSameDoc && string(existing) == string(val) {
 				return nil
 			}
-			return newUniqueIndexError(ctx, doc, fieldsDescs, existing)
+			// The holder stays on this node: the error is rendered to whoever asked for the
+			// write, and on the merge path that is the peer that pushed it. An empty holder
+			// means the entry could not be resolved to a document.
+			var heldBy string
+			if shortID, decodeErr := keys.DecodeDocShortID(existing); decodeErr == nil {
+				heldBy, _, _ = id.GetDocID(ctx, shortID)
+			}
+			log.InfoContext(ctx, "unique index violation",
+				corelog.String("docID", doc.ID().String()),
+				corelog.String("heldBy", heldBy))
+			return newUniqueIndexError(doc, fieldsDescs)
 		}
 	}
 
@@ -458,33 +469,8 @@ func saveUniqueKey(
 	return nil
 }
 
-// incumbentDocID resolves the document already holding a unique index entry, from the
-// short ID the entry stores. Returns an empty string if it cannot be resolved, since this
-// only decorates an error that is being returned either way.
-func incumbentDocID(ctx context.Context, encodedShortID []byte) string {
-	shortID, err := keys.DecodeDocShortID(encodedShortID)
-	if err != nil {
-		return ""
-	}
-	docID, found, err := id.GetDocID(ctx, shortID)
-	if err != nil || !found {
-		return ""
-	}
-	return docID
-}
-
-func newUniqueIndexError(
-	ctx context.Context,
-	doc *client.Document,
-	fieldsDescs []client.CollectionFieldDescription,
-	existing []byte,
-) error {
-	kvs := make([]errors.KV, 0, len(fieldsDescs)+1)
-	// Naming the document that already holds the value is what makes the two comparable;
-	// without it the error only says which one lost.
-	if incumbent := incumbentDocID(ctx, existing); incumbent != "" {
-		kvs = append(kvs, errors.NewKV("HeldBy", incumbent))
-	}
+func newUniqueIndexError(doc *client.Document, fieldsDescs []client.CollectionFieldDescription) error {
+	kvs := make([]errors.KV, 0, len(fieldsDescs))
 	for iter := range fieldsDescs {
 		fieldVal, err := doc.TryGetValue(fieldsDescs[iter].Name)
 		var val any
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -72,8 +72,10 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 		}
 		return nil
 	}
+	// A single event has no smaller write set to retry over, so exhaustion is a drop rather than
+	// a fallback into per-event isolation.
 	db.stats.markDropped(dropRetryExhausted)
-	return client.NewErrMaxTxnRetries(err)
+	return NewErrMergeEventDropped(client.NewErrMaxTxnRetries(err), evt.DocID, evt.Cid.String())
 }
 
 // mergeChunkSize bounds how many events share a transaction. Badger re-sorts the
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -43,6 +43,7 @@ import (
 func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 	col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 	if err != nil {
+		db.stats.markDropped(dropCollection)
 		return err
 	}
 
@@ -62,13 +63,16 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 	for i := 0; i < db.txnAttempts(); i++ {
 		err = db.executeMerge(ctx, col, evt)
 		if errors.Is(err, corekv.ErrTxnConflict) {
+			db.stats.chunkConflicts.Add(1)
 			continue
 		}
 		if err != nil {
+			db.stats.markDropped(mergeDropReason(err))
 			return err
 		}
 		return nil
 	}
+	db.stats.markDropped(dropRetryExhausted)
 	return client.NewErrMaxTxnRetries(err)
 }
 
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -43,7 +43,7 @@ import (
 func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 	col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 	if err != nil {
-		db.stats.markDropped(dropCollection)
+		db.stats.markDropped(collectionDropReason(err))
 		return err
 	}
 
@@ -117,7 +117,7 @@ func (db *DB) MergeBatchWithTxn(ctx context.Context, merges []event.Merge) ([]bo
 		col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 		if err != nil {
 			errs = append(errs, NewErrMergeEventDropped(err, evt.DocID, evt.Cid.String()))
-			db.stats.markDropped(dropCollection)
+			db.stats.markDropped(collectionDropReason(err))
 			continue
 		}
 		entries = append(entries, mergeEntry{evt: evt, col: col, index: i})
--- a/internal/db/merge.go
+++ b/internal/db/merge.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 
 	"github.com/ipfs/go-cid"
@@ -76,6 +77,12 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 // several, so a bigger chunk sorts a bigger set more times.
 const mergeChunkSize = 8
 
+// Phases a merge chunk can fail in, reported on the retry-exhaustion log line.
+const (
+	phaseRead   = "read"
+	phaseCommit = "commit"
+)
+
 type mergeEntry struct {
 	evt event.Merge
 	col *collection
@@ -106,6 +113,7 @@ func (db *DB) MergeBatchWithTxn(ctx context.Context, merges []event.Merge) ([]bo
 		col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 		if err != nil {
 			errs = append(errs, NewErrMergeEventDropped(err, evt.DocID, evt.Cid.String()))
+			db.stats.markDropped(dropCollection)
 			continue
 		}
 		entries = append(entries, mergeEntry{evt: evt, col: col, index: i})
@@ -178,6 +186,7 @@ func (db *DB) MergeBatchWithTxn(ctx context.Context, merges []event.Merge) ([]bo
 		for i := range chunk {
 			if err := db.mergeChunk(ctx, chunk[i:i+1]); err != nil {
 				errs = append(errs, NewErrMergeEventDropped(err, chunk[i].evt.DocID, chunk[i].evt.Cid.String()))
+				db.stats.markDropped(mergeDropReason(err))
 				continue
 			}
 			db.publishMergeComplete(chunk[i : i+1])
@@ -197,11 +206,27 @@ func (db *DB) txnAttempts() int {
 	return 1
 }
 
+// namedDocs renders the documents a transaction touched, as collection/docID. Badger
+// reports conflicts without naming the contended key, so this is the only lead available
+// for working out which documents contend with each other.
+func namedDocs(entries []mergeEntry) string {
+	docIDs := make([]string, len(entries))
+	for i, e := range entries {
+		docIDs[i] = e.col.Name() + "/" + e.evt.DocID
+	}
+	return strings.Join(docIDs, ",")
+}
+
 // mergeChunk merges every event of the chunk inside one transaction, retrying the
 // whole chunk on transaction conflict. Isolating a failing event is the caller's job.
 func (db *DB) mergeChunk(ctx context.Context, entries []mergeEntry) error {
-	// Held so that exhausting the retry budget can report the conflict that caused it.
+	// Held so that exhausting the retry budget can report the conflict that caused it and
+	// where the last one was raised.
 	var conflictErr error
+	var phase string
+	// Whether each event created its document, kept until the transaction commits so a
+	// retried attempt does not count its events twice.
+	creates := make([]bool, 0, len(entries))
 	for i := 0; i < db.txnAttempts(); i++ {
 		txn, err := db.NewTxn(false)
 		if err != nil {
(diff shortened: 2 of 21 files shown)


internal/db/merge.go
```diff
@@ -178,6 +186,7 @@ func (db *DB) MergeBatchWithTxn(ctx context.Context, merges []event.Merge) ([]bo
 		for i := range chunk {
 			if err := db.mergeChunk(ctx, chunk[i:i+1]); err != nil {
 				errs = append(errs, NewErrMergeEventDropped(err, chunk[i].evt.DocID, chunk[i].evt.Cid.String()))
+				db.stats.markDropped(mergeDropReason(err))
```
shahzadlone, line 195: todo: Could every received batch and terminal document outcome update `docsDropped`, `batches`, and `batchesWithDrops` before each return?

if you do maybe like an invalid batch test/priobe of `docsDropped=1` while `batches=0` and `batchesWithDrops=0` the same early return and single document paths i think might omit the denominator and outcome counters?

mennatnaga: You're right, i get docsDropped=1, batches=0, batchesWithDrops=0 from that. it also happens when every document was skipped as already merged. and batchesWithDrops only goes up when the merge errors, so the drop paths before storage never set it.

Fixed it, the batch is counted when it arrives, batchesWithDrops comes from the batch's own drop count, and every exit on the single-document path records a drop or a skip. enqueue also returns without running the handler when the document is in flight or the queue is full.

The doc-sync path recorded nothing either, so it does now, and db.Merge counts its conflicts and drops.

Test asserts merged + dropped + skipped equals what arrived.

internal/db/stats.go
```diff
@@ -0,0 +1,170 @@
...
+	dropOther          = "other"
+)
+
+// mergeDropReason names why an event was dropped: the sender could not supply the DAG,
+// two documents claim one indexed value, or the write kept losing to a concurrent one.
+func mergeDropReason(err error) string {
+	switch {
+	case errors.Is(err, ipld.ErrNotFound{}):
+		return dropMissingBlock
+	case errors.Is(err, errors.New(errCanNotIndexNonUniqueFields)):
```
shahzadlone: todo: Could `mergeDropReason` use stable sentinels or typed matching without constructing stack traces per drop?

These probably do multiple allocations for one retry-exhaustion classification which might somewhat contradict the new stats comment that counting does not allocate until reporting?

internal/db/p2p/stats.go
```diff
@@ -0,0 +1,101 @@
...
+// dropDoc counts an inbound document that was meant to be stored and was not, naming why.
+// The reason set is fixed, so a peer cannot grow the map by varying its errors.
+func (p *P2P) dropDoc(reason string) {
+	p.statDroppedDocs.Add(1)
+	p.docDropReason.record(reason)
+}
+
+// skipDoc counts an inbound document deliberately not merged: already held, or excluded
+// by access or the replication filter. Kept apart from drops, which are losses.
+func (p *P2P) skipDoc(reason string) {
```
shahzadlone, line 177: suggestion: Could `skipDoc` use a separate map or a `document outcomes` report label instead of `reportFailureReasons("document drops", ...)`? seems like skipDoc` stores `skip:alreadyMerged` in the same map drained by `reportFailureReasons("document drops", ...)`.

internal/db/index.go
```diff
@@ -456,8 +458,33 @@ func saveUniqueKey(
...
+	ctx context.Context,
+	doc *client.Document,
+	fieldsDescs []client.CollectionFieldDescription,
+	existing []byte,
+) error {
+	kvs := make([]errors.KV, 0, len(fieldsDescs)+1)
+	// Naming the document that already holds the value is what makes the two comparable;
+	// without it the error only says which one lost.
+	if incumbent := incumbentDocID(ctx, existing); incumbent != "" {
+		kvs = append(kvs, errors.NewKV("HeldBy", incumbent))
```
shahzadlone: question: Could `incumbentDocID` and `HeldBy` be gated by read authorization or limited to internal merge diagnostics?

internal/db/merge.go
```diff
@@ -229,15 +263,29 @@ func (db *DB) mergeChunk(ctx context.Context, entries []mergeEntry) error {
...
+			db.stats.markCreateOrUpdate(isCreate)
+		}
 		return nil
 	}
 
-	// Nothing was committed, so callers must not treat the events as merged.
+	// The chunk used its whole retry budget without committing. The caller then re-runs it
+	// one event at a time, so this counts conflict pressure rather than loss. What was
+	// actually lost is named in the caller's error.
+	db.stats.markExhausted()
```
shahzadlone: suggestion: Could `mergeChunk` suppress duplicate `merge chunk exhausted its retries` logs and count the logical `chunkExhausted` outcome once across the initial chunk and one-event isolation retries?

internal/db/p2p/stats.go
```diff
@@ -0,0 +1,101 @@
...
+	p.statSkippedDocs.Add(1)
```
shahzadlone, line 178: suggestion: Could fixed reason counters avoid the allocation from `"skip:" + reason` and the shared `failureReasons` mutex on `skipDoc("alreadyMerged")`?

i think right now there is a one allocation per duplicate call, and all workers likely share the `failureReasons` mutex.

internal/db/merge.go
```diff
@@ -61,13 +63,16 @@ func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
...
+			db.stats.chunkConflicts.Add(1)
 			continue
 		}
 		if err != nil {
+			db.stats.markDropped(mergeDropReason(err))
 			return err
 		}
 		return nil
 	}
+	db.stats.markDropped(dropRetryExhausted)
```
shahzadlone, line 77: suggestion: Could direct `DB.Merge` and `MergeBatchWithTxn` share one exhaustion accounting path that preserves `chunkConflicts` and `markExhausted()` semantics?

I tried a manual direct-merge test it showed me a `chunkConflicts=5` but `chunkExhausted=0` after direct retries were exhausted, the incremental fix seemss to have corrected only the conflict counter, worth looking into imo.

internal/db/p2p/car.go
```diff
@@ -31,63 +32,104 @@ import (
...
 
 	blockCIDs := make(map[string]struct{})
-	if err := p.collectDAGBlocks(txnCtx, &linkSystem, rootLink.Cid, blockCIDs); err != nil {
-		return nil, err
+	var missingLinks int64
+	if err := p.collectDAGBlocks(txnCtx, &linkSystem, rootLink.Cid, blockCIDs, &missingLinks); err != nil {
+		return nil, p.carFailure(reasonWalk, err)
 	}
 
+	p.statCARMissing.Add(missingLinks)
```
shahzadlone: question: Correct me if wrong, but atm seems like `collectDAGBlocks` inserts a CID into `blockCIDs` before a failed load increments `missing`? then `buildCAR` reads that same CID and returns `carFailure(reasonBlockRead, err)`.

meaning the `carMissingLinks` counter therefore advances for a CAR that is never built, contrary to the short-CAR documentation.

suggestion: Could the skipped CID be excluded from the write set, or should this be reported as CAR failure?

internal/db/merge.go
```diff
@@ -42,6 +43,7 @@ import (
 func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 	col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 	if err != nil {
+		db.stats.markDropped(dropCollection)
```
shahzadlone: suggestion: looks like `getCollectionFromCollectionID` can fail while creating a transaction or reading collections, but this new branch records every error as `collectionNotFound`. Could only the typed not-found error use `collectionNotFound`, with storage, transaction, cancellation, and unknown failures classified separately?

internal/db/p2p/p2p.go
shahzadlone, line 1: question: Was going through this file and wondering does this mean that an in-flight duplicate returns `nil` immediately, and the branch maps the not-handled state to `skipDoc("inFlight")` before the active request's result is known. will the first request run later returning an error, so the same CID can be reported as one drop and one skip. Could duplicates inherit the active result, or should the counters be explicitly defined per unique CID?

internal/db/p2p/p2p.go
```diff
@@ -804,59 +911,82 @@ func (p *P2P) processPushlogRequest(
 
 	// Handle batched multi-document push using a single shared transaction.
 	if len(req.Documents) > 0 {
-		results, err := p.processBatchedDocuments(ctx, req, isReplicator)
+		// Counted before the documents are processed, so a batch still counts when none
+		// of them survives.
+		p.statBatches.Add(1)
+
+		results, dropped, err := p.processBatchedDocuments(ctx, req, isReplicator)
```
shahzadlone, line 931: question: batch branch bypasses the keyed processQueue and passes the untrusted req.Documents slice to MergeBatchWithTxn, whose contract requires independent document and collection keys. Repeated CIDs can therefore be reported merged and relayed more than once. Could the batch be deduplicated or rechecked under the same keyed merge gate?

internal/db/p2p/car.go
```diff
@@ -31,63 +32,104 @@ import (
...
+	if p.carFailureReason.record(reason) {
+		log.ErrorE("Failed to generate CAR", err, corelog.String("reason", reason))
+	}
+	return err
+}
+
+// carImportFailure records an abandoned import: how it failed and how many blocks it had
+// already written. Those blocks sit in the store owned by no document until a later merge
+// claims them.
+func (p *P2P) carImportFailure(reason string, err error) error {
```
shahzadlone: suggestion: Could `importCAR` pass the number of blocks written into `carImportFailure` and report it with the failure?atm telemetry exposes `carImports=1` and `carImportFailed=1` totals but no written-block count.

internal/db/stats.go
mennatnaga: Good catch. It built the errors it compared against, and constructing one captures a stack trace and renders it immediately. Fixed it so it matches package sentinels now, ErrCanNotIndexNonUniqueFields and a new client.ErrMaxTxnRetries.

It still allocates. defraError.Is compares Error() strings when the target is not a defraError, so the ipld.ErrNotFound case costs. The type comment now states what counting costs rather than claiming nothing allocates.

Cancellation and deadline were landing in other, they have their own contextDone reason now.

internal/db/p2p/stats.go
mennatnaga: Implemented this, skips now go into their own map and get printed on their own document skips line.

The running totals were already separate, docsDropped and docsSkipped are different fields on the stats line. The problem was only the per-reason breakdown. Skip reasons went into the drop map with a skip: prefix, and when the reporter drained that map it printed all of them under one heading, document drops. So anyone reading that line, or grouping on it, saw skip:alreadyMerged sitting next to real losses.

There are tests for the split, and for a report draining both maps.

internal/db/index.go
mennatnaga: I went with keeping it on the node rather than gating it.

As far as I understand gating on read authorization wouldn't not work here, because the merge path has no caller identity to check against. I believe that is the same reason getDocForMerge already reads without the ACP filter  (please correct me if I'm wrong). So now HeldBy is off the error entirely and newUniqueIndexError is back to the shape it had before this PR. The holder's docID now goes only to this node's log, next to the docID of the write that lost. incumbentDocID had a single caller, so its body is inlined into saveUniqueKey

internal/db/merge.go
mennatnaga: Implemented both, the counter and the log now fire only when the chunk holds more than one event, which is the case that has a smaller write set to fall back to. That covers the isolation re-runs, since the isolation loop always passes a single-event slice.

It also means a single-event chunk's own first exhaustion is no longer counted or logged, including the trailing chunk when a batch is not a multiple of the chunk size. When that event then fails it is recorded as a retryExhausted drop, so it is not lost, it just is not counted as a chunk.

The drop accounting on this path did not change. An event that exhausts on its own is still counted under retryExhausted.

internal/db/p2p/stats.go
mennatnaga: I removed the allocation. Skips have their own map now, so the skip: prefix is no longer needed, and with no prefix there is no string to build. skipDoc measures zero allocations per call now

I did not fix the mutex. Drops and skips no longer share one, but that is two locks rather than the per-reason counters afaik. Every skip still takes the skip map's lock, and the callers are not one pool: the pubsub workers, the sync queue workers, and a goroutine per replicator stream. please correct me if I'm wrong

internal/db/merge.go
mennatnaga: You are right, it was inconsistent: mergeChunk counted any exhausted chunk including a single-event one, DB.Merge counted none.

Fixed by narrowing mergeChunk rather than adding to DB.Merge. A chunk of several is retried one event at a time and may still land everything. A single event has nothing smaller to retry, so it is a drop. One counter for both would hide that difference.

Both paths do record retryExhausted, on the merge drops line rather than the stats line.

Renamed chunkConflicts to txnConflicts, exhausted to chunkExhausted, and markExhausted() is gone.

internal/db/p2p/car.go
mennatnaga: You are right about the trace, but afaik excluding the CID would ship a short CAR. A receiver imports a CAR rather than walking the DAG, so a block missing from it never gets fetched, and the merge then fails on that link and drops the document. Reporting it as a CAR failure is what already happens, since the write loop cannot read the block either.

I moved the Add below the write loop, and a link is counted only for a CAR that gets returned. Both directions are covered: an abandoned CAR counts no missing link and no build, and a CAR that ships with two unfollowable links counts one build and two links.

internal/db/merge.go
mennatnaga: You are right. Fixed it so that only a real not-found error gets collectionNotFound, and everything else falls through to the normal drop classifier. That applies at both call sites, the direct merge and the batch one.

It does not give you all four categories tho, cancellation has its own contextDone reason now, and that covers the transaction case too, since opening a transaction can only fail with the database context's error.

internal/db/p2p/p2p.go
mennatnaga: Yes, correct. The duplicate returns nil straight away, and the same head does end up as one drop and one skip when the running arrival fails. I ran that case to be sure.

The counters record arrivals rather than unique CIDs. Two deliveries are two arrivals: one did the work and lost the document, the other did nothing. Since drops and skips are separate totals, the duplicate does not inflate the loss count.

Inheriting means blocking on the active result, and the second arrival still did no work. The branch also picks skip or drop from the returned error, so inheriting a non-nil result would record a syncQueueFull drop. Keying per CID would hide how often duplicates arrive.

internal/db/p2p/p2p.go
mennatnaga: You are right. As far as I understand, req.Documents does not reach MergeBatchWithTxn directly, processBatchedDocuments filters it into a fresh slice first. That slice is never deduplicated tho, so the repeats survive.

Rechecking under the keyed gate would not help if I understand correctly, enqueue blocks until its handler returns, so two copies in one request take the key in turn rather than at once, and the second still sees the block as unmerged, since the to-merge marker is only cleared by the merge itself. What saves the single-document path is that db.Merge runs inside the handler.

Stored data is unaffected, the rootstore is byte-identical after one merge and after two. The cost is the report and the relay. docsMerged counts the copy, and if the duplicate is of a create the second lands in updates, so one document shows in both. Each copy runs all of SendUpdate, so two pushes per replicator and two batch entries.

I have not fixed this yet. Caller side is the smaller change and lets me count the collapsed copy. Callee side covers every future caller, but then we have to pick what counts as a duplicate: on docIDs and collectionIDs it would drop a second genuine head, on head CID it removes only exact repeats. I would do the callee on head CID, returning true for the collapsed copy. Any recommendation or thoughts about this?

internal/db/p2p/car.go
mennatnaga: Added this. It now reports as carImportOrphanBlocks, with blocksWritten on the first log line for each reason.

It counts the blocks the import added, not the ones it tried to store. A CAR usually carries blocks we already hold, and Put returns nil either way, so the count comes from a Has first.

internal/db/p2p/car.go
```diff
@@ -31,63 +33,108 @@ import (
...
-// collectDAGBlocks recursively collects all block CIDs in the DAG by following Links.
+// collectDAGBlocks recursively collects block CIDs by following Links.
+// missing is incremented for every link whose block could not be loaded. The CID stays in the
+// write set either way, so the block is still read when the CAR is written.
 func (p *P2P) collectDAGBlocks(
 	ctx context.Context,
 	linkSystem *linking.LinkSystem,
 	blockCID cid.Cid,
 	visited map[string]struct{},
+	missing *int64,
```
tzdybal: nit: I think returning `missing` instead of passing it as argument is more idiomatic / clean.

internal/db/p2p/stats.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// failureReasons counts occurrences by reason, drained once per report interval. Callers
+// that also log keep the set of reasons already logged, so a repeated occurrence costs a
+// counter increment rather than a line.
+type failureReasons struct {
+	mu      sync.Mutex
+	counts  map[string]int64
+	flagged map[string]struct{}
+}
```
tzdybal, line 104: Add `newFailureReasons() failureReasons` that returns object with initialized maps. There is no reason to check on each `record*` call, with mutex locked. It's also not consistent with how `P2P` is initialized - with all the maps allocated in constructor. This change would break the nil-safe usage, of `failureReasons`, especially for tests, but we just need small helpers to initialize `P2P` instead of using `&P2P{}` in tests.

internal/db/p2p/car.go
```diff
@@ -31,63 +33,108 @@ import (
...
+		p.statCARFailed.Add(1)
+		return nil, err
+	}
+	p.statCARBuilt.Add(1)
+	return data, nil
+}
+
+// carFailure records the reason a CAR could not be built and logs the first occurrence of
+// that reason, so a persistent failure shows up as a count rather than a line per call.
+func (p *P2P) carFailure(reason string, err error) error {
```
tzdybal: I completely don't know what to think about this method and how it's used in the code (same for `carImportFailure`). It works, it's smart, it's very compact on call, but feels very strange. Like a side effect to error.

I prefer how `syncDAGFailure` is implemented and used. It's more explicit, needs 1 more line of code for every call, but error handling feels more idiomatic.

internal/db/stats.go
```diff
@@ -0,0 +1,175 @@
...
+	return mergeDropReason(err)
+}
+
+// markDropped records an event that did not merge, under the given cause.
+func (s *mergeStats) markDropped(reason string) {
+	s.dropMu.Lock()
+	defer s.dropMu.Unlock()
+	if s.dropReasons == nil {
+		s.dropReasons = make(map[string]int64)
+	}
```
tzdybal: Again, this should be in constructor, not on every call to `markDropped`.

internal/db/p2p/sync_dag.go
```diff
@@ -47,18 +48,49 @@ func (p *P2P) syncDAG(ctx context.Context, block *coreblock.Block) error {
...
 // Uses an explicit stack to avoid goroutine stack overflow on deep DAGs (#2722).
-func (p *P2P) loadBlockLinks(ctx context.Context, linkSys *linking.LinkSystem, block *coreblock.Block) error {
+//
+// written is incremented for each link the walk loads. On error the returned reason names the
+// step that failed, for the caller's counters.
+func (p *P2P) loadBlockLinks(
+	ctx context.Context,
+	linkSys *linking.LinkSystem,
+	block *coreblock.Block,
+	written *int64,
```
tzdybal: Again, this could be returned among error.

internal/db/p2p/sync_dag.go
```diff
@@ -47,18 +48,49 @@ func (p *P2P) syncDAG(ctx context.Context, block *coreblock.Block) error {
...
+) (string, error) {
```
tzdybal: And for the returned `reason` I think we can just skip it, and have a helper function to get a reason based on the error type, and use it in `syncDAGFailure`.

internal/db/p2p/sync_doc.go
```diff
@@ -212,6 +212,7 @@ func (p *P2P) handleDocSyncItem(
 		if err != nil {
 			log.ErrorE("Failed to parse CID from bytes", err,
 				corelog.String("DocID", item.DocID))
+			p.dropDoc("invalidCID")
```
tzdybal: This begs for a constant.

internal/db/p2p/sync_doc.go
```diff
@@ -220,6 +221,7 @@ func (p *P2P) handleDocSyncItem(
 				results[item.DocID] = append(heads, docCid)
 			} else {
 				// we've seen this head already, just skip
+				p.skipDoc("duplicateHead")
```
tzdybal: Another constant.

internal/db/p2p/sync_doc.go
```diff
@@ -245,6 +247,7 @@ func (p *P2P) syncDocumentAndMerge(
 ) error {
 	err := p.syncDocumentDAG(ctx, head)
 	if err != nil {
+		p.dropDoc("syncDAG")
```
tzdybal: And another constant

internal/db/p2p/sync_doc.go
```diff
@@ -256,7 +259,12 @@ func (p *P2P) syncDocumentAndMerge(
 		CollectionID: collectionID,
 	}
 
-	return p.db.Merge(ctx, evt)
+	if err := p.db.Merge(ctx, evt); err != nil {
+		p.dropDoc("mergeFailed")
```
tzdybal: And the last constant from this file.

internal/db/merge.go
```diff
@@ -42,6 +43,7 @@ import (
 func (db *DB) Merge(ctx context.Context, evt event.Merge) error {
 	col, err := getCollectionFromCollectionID(ctx, db, evt.CollectionID)
 	if err != nil {
+		db.stats.markDropped(collectionDropReason(err))
```
tzdybal, line 46: Oh, here is the type helper function to get the reason string from error, just like I suggested for `loadBlockLinks`/`syncDAGFailure`.

internal/db/merge.go
```diff
@@ -197,29 +212,50 @@ func (db *DB) txnAttempts() int {
...
 		if err != nil {
 			return err
 		}
 		txnCtx := InitContext(ctx, txn)
 
+		creates = creates[:0]
 		var mergeErr error
 		for _, e := range entries {
-			if mergeErr = db.mergeInTxn(txnCtx, e.col, e.evt); mergeErr != nil {
+			isCreate, err := db.mergeInTxn(txnCtx, e.col, e.evt)
```
tzdybal: nit: `created` instead of `isCreate` (in all places).

tzdybal (changes requested):
There are inconsistencies between packages, missing constants, strange flow or values passed by reference & returned. But in general lot of good work.

dvncan (approved):

tzdybal (approved):

PR #5147, opened 2026-08-13
perf: purge in chunks of 8

mennatnaga:

Stacked on #5136.

`purgeChunkSize` was 100. Badger re-sorts the transaction's pending writes on every iterator open, and each document opens several, so a bigger chunk sorts a bigger set more times.

Ran two identical hosts for about 22 hours, one on chunk size 100 and one on 8. At 100, merges fail about 20x more often. Each failed merge throws away its batch, the peers send it again, and the node merges it a second time, so it does more update work and hits 1,843 transaction conflicts against 54.

`BenchmarkPurgeByDocIDsChunkSize` opened stock badger options and forced `pruneHistory` true, so it measured neither what ships nor the shipped default. It now opens the options a node runs, from `node/store_badger.go`, and sweeps both `pruneHistory` branches.

### DIFF
--- a/internal/db/collection_purge_bench_test.go
+++ b/internal/db/collection_purge_bench_test.go
@@ -1,0 +1,134 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+package db
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	badgerds "github.com/dgraph-io/badger/v4"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcenetwork/corekv/badger"
+
+	"github.com/sourcenetwork/defradb/client"
+	acpDB "github.com/sourcenetwork/defradb/internal/db/acp"
+)
+
+// newOnDiskDB opens a DB on a badger directory with stock options. It returns a close func
+// rather than registering cleanup so a caller can release it per iteration. An in-memory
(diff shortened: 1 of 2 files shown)


internal/db/collection_purge_bench_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return docIDs
+}
+
+// BenchmarkPurgeByDocIDsChunkSize measures how long it takes to purge a fixed set of
+// documents as the number of them sharing a transaction changes, so purgeChunkSize can be
+// chosen from a measurement rather than assumed.
+//
+// pruneHistory is on because that is the deployed setting and it adds the per-document DAG
```
shahzadlone: suggestion: Could the benchmark use the deployed Badger options and include both `pruneHistory` branches, or narrow its rationale?

Pinned source tracing shows this benchmark opens stock Badger options instead of the deployed 256-byte threshold, ZSTD settings, and value-log GC wrapper, and it forces `pruneHistory=true` (unless i am missing something haha, do correct me if i am wrong), but shipped CLI and HTTP defaults are false. The measured chunk optimum therefore likely does not cover the production options or both purge branches?

mennatnaga: You're right, I just pushed a commit that opens the options a node opens, sweeps both pruneHistory branches, and takes the sweep below 8.

I didn't include the gc wrapper, it lives in the node package and this test can't import it, please correct me if I'm wrong. ZSTDCompressionLevel was already badger's default. And re prune-history, that comment was describing our deployment on the hosts and generators to reclaim disk space rather than defradb's default, so I removed it.

One thing to note: 8 isn't the best number; it was just the lowest my old sweep tried. 1, 2 and 4 are all better. But I'm leaning towards leaving it at 8, as it gives better performance than the previous 100, and I tested how a chunk holds up while merges are landing on the two hosts, unlike the 100 chunk size. Please lmk if you have any recommendations or if I'm missing something.

iverc (approved):

PR #5153, opened 2026-08-14
fix: Make AddCollection discards txn on failure in C wrapper

### LINKED ISSUE #5152: Properly discard transaction in AddCollections function of C wrapper
The call to `defer txnDiscard()` is made only after the C function is called, and the unmarshalling is performed on its result. In the case of an error, the transaction will never be discarded, creating a leak. This should be addressed. Moreover, tests should be intorudced for each of these failure paths.

ChrisBQu:
## Relevant issue(s)

Resolves #5152 

## Description

The `AddCollection` function of the C client wrapper had a bug. If it was called without an explicit transaction, an implicit transaction would be created. If this happened, and the C function call returned an error, that transaction would never be properly discarded. This fixes that by simply moving the `defer discard` up a bit.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/wrapper.go
+++ b/cbindings/wrapper.go
@@ -436,12 +436,16 @@ func (w *CWrapper) AddCollection(
 	if hadTxn {
 		txn = gotTxn
 	} else {
-		clientTxn, _ := w.NewTxn(false)
+		clientTxn, err := w.NewTxn(false)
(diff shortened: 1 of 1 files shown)


nasdf (approved):
LGTM

PR #5155, opened 2026-08-14
fix: Fix lens config bug in C bindings' PatchCollection

### LINKED ISSUE #5154: Fix lens config bug in C bindings PatchCollection function
See that the `PatchCollection` function inside `cbindings/collection_patch.go` contains the following code:

```
		// Length being greater than 0 also means it is not nil, so no need to check
		if len(lensCfg.Lenses) > 0 {
			migration = immutable.Some(lensCfg)
		}
```

This treats having zero lens models the same as having no configuration sent, and collapses `Some(model.Lens{})` to `None`. 

This should be addressed, and a proper resolution should ideally contain tests that illustrate the correct behavior.

ChrisBQu:
## Relevant issue(s)

Resolves #5154

## Description

There was a bug where a present-but-empty lens configuration was treated the same as no lens being present at all. This meant the C bindings diverged from the Go behavior. This PR addresses that, and introduces two integration tests to illustrate the new behavior.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/collection_patch.go
+++ b/cbindings/collection_patch.go
@@ -50,10 +50,7 @@ func PatchCollection(nodePtr C.uintptr_t,
 			return returnC(returnGoC(1, err.Error(), ""))
 		}
 
-		// Length being greater than 0 also means it is not nil, so no need to check
-		if len(lensCfg.Lenses) > 0 {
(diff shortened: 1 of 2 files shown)


nasdf (approved):
LGTM

PR #5156, opened 2026-08-14
feat: Add HTTP client options

iverc:
## Summary
- allow callers to provide a default token identity for HTTP requests
- allow callers to provide a configured `net/http` client
- cover transaction authentication, TLS transport configuration, and Node ACP authorization

## Testing
- `go test ./http ./node`
- `go test -race ./http ./node`


nasdf (approved):
This is a nice quality of life change. Thanks!

PR #5160, opened 2026-08-17
fix: Make blank CID in AddView consistent across clients

### LINKED ISSUE #5159: C binding AddView should behave the same as Go client
Currently, the `AddView` function of the C bindings contains the following block of code:

```
	if transformCIDValue != "" {
		opts.SetTransformCID(transformCIDValue)
	}
```

This results in no transform being set in the case of a blank string. This is subtly different from the Go client, which will actually error out in the case of a blank CID being passed in. The behaviors should be identical, though.

ChrisBQu:
## Relevant issue(s)

Resolves #5159 

## Description

Previously, if you passed a blank CID string into the C bindings' `AddView` function, it would never configure a view. However, the Go path would try to decode the empty string, fail to do so, and return an error.  Incidentally, the CLI behaves the same way as the C bindings, and unlike the Go path. This difference in behavior is inconsistent, and this PR addresses that by adjusting the C function and C wrapper, and the CLI code, to behave the same way as the Go code.

This PR also introduces an integration test to illustrate the correct behavior.

As an aside, this PR removes the `stringFromImmutableOptionString` utility function from the C bindings. It was never used in other places, so we no longer need it.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/tests/integration/view/simple/transform_cid_presence_test.go
+++ b/tests/integration/view/simple/transform_cid_presence_test.go
@@ -1,0 +1,99 @@
+// Copyright 2026 Democratized Data Foundation
+//
+// This file is part of the DefraDB test suite.
+//
+// The DefraDB test suite is licensed under either:
+//
+//   (1) GNU Affero General Public License v3
+//   (2) Business Source License 1.1
+//
+// See tests/LICENSE for details.
+
+package simple
+
+import (
+	"testing"
+
+	"github.com/sourcenetwork/immutable"
+
+	"github.com/sourcenetwork/defradb/client"
+	"github.com/sourcenetwork/defradb/client/options"
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+// AddViewOptions.TransformCID being omitted (None) and being explicitly set to ""
+// (Some("")) must not be treated the same. An empty CID is invalid and should be
+// rejected, while an omitted one means "no transform". These illustrate the two
+// different cases.
+
+func TestAddView_NoTransformCID_CreatesViewSuccessfully(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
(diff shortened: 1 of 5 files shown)


tests/integration/view/simple/transform_cid_presence_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/sourcenetwork/defradb/tests/action"
+	testUtils "github.com/sourcenetwork/defradb/tests/integration"
+)
+
+// AddViewOptions.TransformCID being omitted (None) and being explicitly set to ""
+// (Some("")) must not be treated the same. An empty CID is invalid and should be
+// rejected, while an omitted one means "no transform". These illustrate the two
+// different cases.
+
+func TestAddView_NoTransformCID_CreatesViewSuccessfully(t *testing.T) {
```
AndrewSisley: suggestion: We should already have tests that cover this, I'm not sure duplicating it here adds enough value to justify the cost.  I suggest removal.

tests/integration/view/simple/transform_cid_presence_test.go
```diff
@@ -0,0 +1,90 @@
...
+					}
+				`,
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
+
+func TestAddView_BlankTransformCID_FailsAndCreatesNoView(t *testing.T) {
```
AndrewSisley, line 30: praise: Thanks for adding an int. test for this Chris, is a good gap to close :)

AndrewSisley (approved):
LGTM, thanks Chris!

tests/integration/view/simple/transform_cid_presence_test.go
ChrisBQu: It's hard to make a strong argument against this. The functionality is definitely tested elsewhere (implicitly.) I'll remove it, and just update the wording on the comment.

PR #5163, opened 2026-08-18
refactor: Detangle crdts

### LINKED ISSUE #5115: Self-contain CRDTs
At the moment the various elements that make up a CRDT are somewhat scattered through out the codebase - for example, the kinds they are applicable to, and the inputs/functions they have.  This makes introducing new CRDTs more painful and error prone than it needs to be.

Define an interface that allows a single concrete type to define everything that a CRDT needs to have in order to be exposed via Defra, and hook it up.

PRs split, follow-up branch is `5115-crdt-self-contain-gql-declaration`, `5115-crdt-self-contain-gql-declaration-2`

AndrewSisley:
## Relevant issue(s)

Partially resolves #5115

## Description

Detangles the crdts in preparation for more complex ones (e.g. multi-operation text crdts), and for allowing their declarations to define related elements (such as GQL interfaces) - making them complete declarations, where by introducing new ones wont necessitate also modifying a bunch of other code locations in order to expose them.

- The CRDT Go types have been made stateless, this clarifies the scope of the variables that were held in state and the functions that used them.
- CRDT operations (i.e. `Delta` funcs) have no access to the store, or anything to do with it (such as keys).  They no longer share a common interface.
- Priority is now immutable.
- The `Merge` interface is now optional, and defined based on CRDT *scope*, not type.  The no-op implementation of types that pretended to implement it has been removed, and so have the calls to those unimplemented functions. 
- `HeadstoreKey()` has been removed.

The CRDT constructors remain, even though they return empty objects - I will very likely be creating a static registry of available CRDTs in a soon-to-arrive follow-up PR that will resolve the linked issue and make the crdts 'complete' - removing the need for changes to other parts of the codebase when introducing new ones, and handling the dynamic selection of operation (`Delta(...)`) based on user selection.  This follow up PR should hopefully clean up some of the remaining `switch` statements dotted about.

Even though not many lines were changed in this PR, I expect the collective diff to be quite hard to follow when reviewing - it might be much easier to review commit-by-commit, they should be clean, and each one contains extra information in the message body that should hopefully help you guys understand why each change was made.


internal/core/block/store.go
```diff
@@ -127,8 +123,7 @@ func addDelta(
 		return cidlink.Link{}, nil, NewErrStoreBlock(err)
 	}
 
-	// merge the delta and update the state
-	err = ProcessBlock(ctx, crdtData, block, link)
+	err = UpdateHeads(ctx, headstorePrefix, block, link)
 	if err != nil {
 		return cidlink.Link{}, nil, NewErrProcessBlock(err)
```
nasdf, line 128: nitpick: rename error to match new function

internal/core/block/store.go
```diff
@@ -55,42 +56,36 @@ type AddDeltaOptions struct {
 // It checks the current heads, sets the delta priority, adds it to the blockstore, then runs ProcessBlock.
```
nasdf: nitpick: replace ProcessBlock with UpdateHeads

internal/db/fetcher/versioned.go
```diff
@@ -420,24 +420,27 @@ func (vf *VersionedFetcher) merge(c cid.Cid, docShortID uint64) error {
...
 		}
 
-		var mcrdt crdt.ReplicatedData
 		switch {
 		case block.Delta.IsCollection():
-			mcrdt = crdt.NewCollection(
-				vf.col.Version().VersionID,
-				keys.NewHeadstoreColKey(collectionShortID),
-			)
+			// no-op: collection value blocks are not merged
```
nasdf, line 425: question: Does this fix the double counter increment bug?

internal/db/merge.go
```diff
@@ -452,17 +454,18 @@ func (mp *mergeProcessor) setLinkedBlockDocIDMappings(
 	return nil
 }
 
-func (mp *mergeProcessor) initCRDTForType(
+func (mp *mergeProcessor) mergeBlock(
```
nasdf, line 457: question: The bool return value seems to always be false when an error is returned. Would it make more sense to check the error instead of the additional bool?

nasdf (approved):
This is a much better design! I left a few small notes that you may want to address but no blockers from me.

internal/core/block/store.go
AndrewSisley: Thanks, will change (or delete the wrap)

- [x] Update err

internal/core/block/store.go
AndrewSisley: - [x] Update store.go doc

internal/db/fetcher/versioned.go
AndrewSisley: I'll check, but I doubt it - from memory I thought that issue was a write problem, this is the version fetcher (only used when specifying a cid in a normal query).

- [x] Check double increment counter issue

internal/db/merge.go
AndrewSisley: Sounds like a nice simplification - will have a look - thanks Keenan

- [x] Check `mergeBlock` fail params

internal/core/block/store.go
AndrewSisley: I've updated it, but I might revert this and leave it as-is depending on what we decide on discord.

internal/db/fetcher/versioned.go
AndrewSisley: I don't think it should impact that issue - no behaviour in the merge processor (or anywhere) should have been changed in this PR.

internal/db/merge.go
AndrewSisley: Leaving as-is, `false, nil` is returned if the field is not known locally, this gets handled differently by the calling code - so there is still value in keeping the split.

internal/core/block/store.go
AndrewSisley: - [x] If team decided errors receive semver protection revert  https://github.com/sourcenetwork/defradb/pull/5163/changes/4547544d19310af656afc3e3d9827e0fdef52d17 https://github.com/sourcenetwork/defradb/pull/5163/changes/b2e5dddeba8ca6c6b3e11b7953daef327b2b6f50 and maybe tag them with `//@deprecate` or something

shahzadlone (approved):
LGTM

internal/core/block/store.go
AndrewSisley: I'm going to drop these and unblock the PR.  I feel like the discussion is leaning towards semver-ness in some form.

islamaliev:
Bug bash result: Ran different tests with p-counters and regular LWW with different nodes and replicators. Everything looks good

PR #5166, opened 2026-08-18
fix: Make ExecuteQuery preserve int64 precision

### LINKED ISSUE #5165: C/HTTP/CLI clients' ExecuteQuery do not preserve int64 precision
The `ExecuteQuery` function inside the C bindings decodes GraphQL request variables from a JSON string into `map[string]any` using the standard library's default number handling, which decodes every  JSON number as float64.  As a result, any int64 variable above 2^53 gets silently rounded before the request executes. The CLI and HTTP clients have similarly bugged behavior.

Proper resolution of this issue would ideally contain an integration test of this behavior.

ChrisBQu:
## Relevant issue(s)

Resolves #5165

## Description

This PR introduces several helper functions to  the `internal/utils.go` file: `UnmarshalJSON`, `DecodeJSONVariables`, and `normalizeJSONNumbers`. These functions are introduceed here because they are shared by the QueryExecute functions of the C bindings, CLI, and HTTP clients. The purpose is to allow decoding of integers in the query variables in such a way that precision is preserved. Previously, large integers above 2^53 would be rounded.

In addition to making these changes, this PR introduces an integration test of the new behavior.
## Tasks

## How has this been tested?

- WSL


internal/utils/util.go
```diff
@@ -35,6 +37,50 @@ func NewOptions[T any](opts ...enumerable.Enumerable[func(*T)]) *T {
...
+func DecodeJSONVariables(data []byte) (map[string]any, error) {
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.UseNumber()
+
+	var vars map[string]any
+	if err := dec.Decode(&vars); err != nil {
+		return nil, err
+	}
+	normalizeJSONNumbers(vars)
+	return vars, nil
```
nasdf, line 62: todo: I think coderabbit is correct here. If you call `dec.More()` you could return an error when true.

nasdf (approved):
I think the one todo I left is worth addressing, but otherwise looks good!

shahzadlone (approved):
LGTM

Issue #5072, opened 2026-07-22
Optional metric argument for similarity queries


islamaliev:
Part of https://github.com/sourcenetwork/defradb/issues/5070. Follows the additional distance metrics (https://github.com/sourcenetwork/defradb/issues/5073).

A similarity query names a field and a vector, and the metric comes from whatever index that field has. That works while a field carries at most one vector index, which is what we enforce today: a second index with a different metric is rejected, because the planner takes the first index it finds and the choice would silently decide the results rather than just the cost.

Lifting that restriction needs a way for the query to say which index it wants. The natural spelling is an optional metric alongside the vector, defaulting to the field's index when there is only one and required when there is more than one. A field could then carry a cosine index for topical search and a Euclidean one for positional search, with each query naming what it means.

This is deliberately narrower than what this issue originally proposed. The first version was a dedicated nearest-neighbour argument bundling the field, vector, k and a search-effort knob in one place. The team's read was that it did not enable anything the existing shape cannot express, and was not worth the extra API surface for the ergonomic gain alone, so the scope is now just the part that unblocks something real.

## Testing
Integration tests confirming a query with no metric still uses the field's only index, that naming a metric selects the matching index when a field has several, that a metric with no matching index is rejected, and that results and ordering match the equivalent single-index query.

islamaliev:
Rescoped from "Dedicated nearest-neighbour query argument" after discussing it with the team.

The original proposal added an argument naming the vector field, the query vector, k and a search-effort knob together, replacing the score-order-limit shape. The objection was that similarity and distance are the same number flipped, so the new surface enabled nothing the current shape cannot express, and the ergonomic gain alone did not justify the extra API. The one thing it did solve, the silent full scan when a query is ordered the wrong way, is better handled by warning about it, tracked separately in https://github.com/sourcenetwork/defradb/issues/5173.

What survives is the part that unblocks something real. https://github.com/sourcenetwork/defradb/pull/5169 makes a similarity query score by whichever metric its field's index was built with, which is what lets the new metrics be searched at all. It also has to reject a second index on a field when the metrics differ, since the planner takes the first index it finds and the choice would decide the results. An optional metric on the query is what removes that restriction, and it is now the whole of this issue.

PR #5169, opened 2026-08-19
feat: Vector distance metrics L2 and dot product

### LINKED ISSUE #5073: Additional vector distance metrics (L2, dot product)
Part of https://github.com/sourcenetwork/defradb/issues/5070. Follows the HNSW index (https://github.com/sourcenetwork/defradb/issues/5042).

The first release ships cosine distance only, which is the right default for text embeddings. Some vectors are better compared by straight-line (Euclidean) distance or by dot product, so we should let the metric be chosen when the index is created. The index configuration already records the metric, so this is mostly adding the encode, validation, and search paths for the additional metrics and rejecting a metric change on an existing index (that requires a rebuild).

## Testing
Integration tests for each metric: index creation with the metric set, correctness of nearest-neighbour results against a brute-force baseline for that metric, and rejection of an attempt to change the metric on an existing index.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5073

## Description

Vector indexes only supported cosine, which compares direction and ignores magnitude. This adds Euclidean and dot product, picked when the index is created.

Normalization was the interesting part. Every vector used to be scaled to unit length on both insert and search, because that is what cosine needs, but the other two must not have that: it collapses Euclidean distances onto the unit sphere and strips dot product of the magnitude that defines it. Normalization is now something each metric declares, so one added later has to answer the question rather than quietly inherit cosine's behaviour.

A similarity query now scores by whatever metric its field's index was built with instead of always computing cosine, which is what lets the new metrics actually be searched. Otherwise an index would rank documents one way while the query scored them another and the results would not be the ones asked for. Each metric is mapped onto the larger-is-nearer convention the similarity field already has, which changes direction only and leaves the metric's own ordering intact. A field with no vector index still scores as cosine, so existing queries are unaffected.

Creating a second index on a field that already has one with a different metric is rejected, since it would be arbitrary which of them answered a query. Dropping and recreating is the way to change metric, because a graph is not just searched by its metric, it is built by it.

## Tasks

## How has this been tested?

Recall against a brute-force baseline computed under the same metric, plus small hand-placed graphs where a vector's length decides the ranking, which fail if normalization ever creeps back. Integration tests cover the directive and descriptor, the build and maintenance path, the rejection, and drop-and-recreate.

The load-bearing one runs the same query over one indexed collection twice, once using the index and once forced to a full scan, over vectors the metrics deliberately rank differently, so a wrong metric cannot pass by coincidence.

- macOS, on the Go, HTTP, CLI and C clients


internal/index/hnsw/hnsw.go
```diff
@@ -125,3 +153,25 @@ func distance(metric Metric, a, b []float32) float32 {
...
 	return float32(1 - dot(a, b))
 }
+
+// squaredEuclideanDistance computes the squared straight-line distance between a and b. The square
+// root is skipped: it is monotonic, and distances are only ever compared against each other. If the
+// lengths differ, only the shared leading elements are used, matching dot.
+//
+// Squaring costs range: the sum is held in float64, but the result narrows to float32, so components
+// beyond roughly 1e19 saturate to +Inf. Such a vector still sorts as the farthest, so ordering holds.
+func squaredEuclideanDistance(a, b []float32) float32 {
```
nasdf, line 163: question: Why not use the functions defined in `utils/math/vector.go`?

tests/integration/results.go
```diff
@@ -249,17 +249,25 @@ func CosineSimilarity(source, vector []float64) *cosineSimilarity {
 	return &cosineSimilarity{source: source, vector: vector}
 }
 
+// SimilarityScore matches a _similarity result under the given metric. Use it when the field being
+// queried has a non-cosine vector index, since the score follows the index's metric.
+func SimilarityScore(metric client.DistanceMetric, source, vector []float64) *cosineSimilarity {
+	return &cosineSimilarity{source: source, vector: vector, metric: metric}
+}
+
 type cosineSimilarity struct {
```
nasdf: nitpick: This supports more than cosine now.

nasdf (changes requested):

internal/db/collection_index.go
```diff
@@ -583,6 +587,30 @@ func validateVectorIndexDescription(def client.CollectionVersion, desc client.Ne
...
 }
 
+// validateNoConflictingVectorIndexMetric rejects creating a vector index on a field that another
+// vector index already covers with a different metric.
+//
+// There is no operation that changes an existing index, so this is about two indexes coexisting. The
+// planner uses the first vector index it finds on a field, so which metric answered a query would be
+// arbitrary. A different metric also means a differently built graph, which is why the fix is to drop
+// and recreate rather than to keep both.
+func validateNoConflictingVectorIndexMetric(
```
nasdf, line 600: thought: I'm not sure if I like this limitation. Might be worth a discussion.

internal/index/hnsw/hnsw.go
islamaliev: This package imports only the standard library so it stays extractable as a standalone HNSW implementation, so it can't reach into internal/utils/math.

tests/integration/results.go
islamaliev: good catch. Updated

internal/db/collection_index.go
islamaliev: this limitation is temporary until we have a way to query with specific metric. I created added a comment referencing [the GH issue](https://github.com/sourcenetwork/defradb/issues/5072)

nasdf (approved):
Great work!

Issue #5170, opened 2026-08-19
Cross-version tests cannot restart an external node


islamaliev:
The cross-version multipliers cannot run tests that restart a node.

An external node runs as a separate process started from a downloaded release binary. The wrapper starts it with `--development --no-keyring`, and in that combination the node generates an ephemeral identity on every start. A restarted node therefore comes back with a different PeerID and cannot read the store it wrote before, so the test sees an empty node rather than the data it expects.

The store itself is not the problem. A branch experiment made the wrapper reuse its rootdir across a restart, and the data was still on disk and still reachable. The identity is what does not survive.

Things already tried, none of which fixed it:

- Reusing the rootdir instead of minting a fresh one per start.
- Clearing the cached auth tokens for the node before restart. The restarted process binds a new API port and tokens are bound to the host they were minted for, so the first failure was a 403. Clearing them is not enough on its own because the audience is read from the node's current client, which still points at the old address during setup.
- Enabling a file keyring with a non-interactive `DEFRA_KEYRING_SECRET`. The keyring was created but the PeerID still changed.

`cli/start.go` is identical in v1.0.0 and at HEAD, so this is a property of how the test wrapper starts the node rather than a difference between versions.

To fix this the wrapper needs a way to start the node with the same identity twice. Once that exists the affected tests can drop their exclusions.

Note that restart is only exercised when the store is file-backed (`DEFRA_BADGER_FILE=true`). On the default in-memory store `restartNodes` returns early, so these tests currently pass without restarting anything.

Affected tests, currently excluded via `MultiplierExcludes`:

- `TestP2POneToOneReplicatorWithRestart`
- `TestP2PPeerReplicatorWithUpdateAndRestart`
- `TestP2PWithSingleDocumentSingleUpdateFromChildAndRestart`

Refs #5066

PR #5174, opened 2026-08-20
test(i): Add cross-version P2P test multiplier

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/5066

## Description

We had no automated check that a change keeps working against nodes running an older release, only two hand-written smoke tests. This reruns the existing networked suite with one node on v1.0.0, so the tests we already have double as the compatibility check. Both directions run as separate CI jobs, because they fail differently: a new node sending to an old one relies on the old node ignoring fields it does not know, while the reverse relies on the new node reading a missing field as a zero value.

Most of the work turned out to be in the test harness rather than the multiplier. A node running in another process has no event bus the test can watch, and the harness learns what to expect by watching event buses, so waits returned immediately and assertions ran against data that had not arrived yet. The source node is now asked for the head it just wrote, which feeds the same bookkeeping the in-process path already uses.

Restarting a node in another process is excluded for now, since it comes back with a new identity and cannot read what it wrote. Tracked in https://github.com/sourcenetwork/defradb/issues/5170

## Tasks

## How has this been tested?

Unit tests for the rewriting and the skip logic, which download nothing. The whole networked suite three ways, natively and in both directions against a real v1.0.0 binary, all at zero failures.

To confirm it detects a break rather than passing by construction, I renamed the replicator communication channel on the current build: the existing suite caught two failures, the cross-version runs caught thirteen more.

- MacOS


tests/action/eventually.go
```diff
@@ -0,0 +1,163 @@
...
+//
+// Use it when the harness cannot tell that something has finished and the test
+// has to keep asking. The clearest case is a node running an older release: it
+// can hold a document written against a schema it has never seen, but it cannot
+// report the commit, so there is no signal to wait for.
+//
+// The wrapped action's assertions are captured rather than failing the test, so
+// a failed attempt is just a retry. The final attempt's failure is reported as
+// this action's failure.
+type Eventually struct {
```
AndrewSisley, line 37: praise: This is a fun action, I'm nervous about us potentially using it to cover up crimes, but it is very nice addition :)

tests/action/utils_events.go
```diff
@@ -50,7 +51,16 @@ func waitForUpdateEvents(
...
-			continue // external node: its event bus is in another process; the cross-version test polls a query to confirm sync
+			// A node in another process emits no events we can read, so there is
+			// nothing to wait for here. The write still has to reach the nodes this
+			// one replicates to, and normally those nodes learn what to expect from
+			// this event. Tell them the document ID directly instead, or they would
+			// wait for nothing and the test would read the data too early.
+			//
+			// A node with no replicators has no one to tell, so this does nothing
+			// when networking is not in use.
+			MarkDocsExpectedOnTargets(s, i, collectionIndex, docIDs, ident)
```
nasdf, line 63: thought: We could add the event bus to the client API. It could be useful for other use cases as well.

nasdf (approved):
No blockers from me. I'm curious on your thoughts about making the event bus public.

tests/action/utils_events.go
AndrewSisley: I thought it was public :grin:  The `event` package is public, so it might have been public at one point.  It is public to `DB`, and I would fully support it being public when the concrete `DB` becomes public.

It might have been made private due to the effort required to make it work via the remote clients.

AndrewSisley (dismissed):
I partially reviewed a few days ago but saw that the new multipliers do not seem to work in the CI.  Please make sure they do, and please make them `required`.

tests/action/utils_events.go
nasdf: @AndrewSisley Would you also support adding it to HTTP, CLI and other clients? That would allow the tests to work the same across embedded and external clients.

AndrewSisley: I would.  Originally I started writing in my first message that those clients have subscriptions for this, but there is a difference between the two.

Although embedded also have subscriptions, and IIRC some effort was made into making it work nicely for embedded clients for v1 - so there's a very good argument to be made that we already have the main problem that I am concerned about for the remote clients.

Which is that I would prefer to not to confuse users by offering them too many ways to do something, and I would also like to not have to bother maintaining both.

However, I think events is to subscriptions what the collection API is to the rest of GQL.  And I think my idealised Defra architecture would be collection API and events at the bottom, core stack, with planner, GQL and subscriptions sat on top of that.  And so structurally, it makes sense.  From a users perspective, it could be seen as odd not to expose it via the remote clients, but otherwise I have no good arguments for or against it atm I think.

Sorry if my reply is a bit confusing, I'm not entirely sure how I feel, and I think the current Defra architecture blends too many layers together making it difficult to provide a technically-definite answer.

tests/action/eventually.go
islamaliev: Yeah, this one might be used as an escape hatch for certain use cases. And the action name itself is common in other testing frameworks. Like Ginkgo, where it's just meant to be used for async checks, just when you need to wait for something to happen. And this is what is mentioned in the comment as well. So I think it's fine the way it is for now. We can just see how the usage of this action develops.

PR #5181, opened 2026-08-21
ci: Refresh vulnerability baseline

iverc:
## Summary

- update Go to 1.25.13 and Pion DTLS to 3.1.4
- reduce the active vulnerability scan from 20 findings to 6
- compare finding IDs against an explicit allowlist so a changed finding cannot pass on count alone

The remaining CIRCL findings require migrating the pinned SSI SDK because current Dilithium APIs do not compile with CIRCL 1.6.3. The other allowlisted findings have no compatible upstream fix.

## Testing

- `go mod tidy -diff`
- `govulncheck -C . -format text ./...`
- `go test ./internal/db/p2p ./node -count=1`
- `go test ./tests/integration/net/simple/peer -count=1`


.github/known-vulnerabilities.txt
```diff
@@ -0,0 +1,6 @@
+GO-2023-1821
```
AndrewSisley: todo: Please document how to find out what each line refers to - these are not currently very googlable on their own.

PR #5188, opened 2026-08-22
refactor: Unify index directives

### LINKED ISSUE #5168: Generalize index directive
The recent #5042 PR introduced a new `@vectorIndex` directive that scopes the vector index configuration on the schema SDL level, and lives along side previous existing `@index` directive which is now reserved for the Ordered index.

I am proposing that we unify them and only keep a single `@index` directive. I dont want to set a precedent that each index type we add uses its own independent directive.

We currently have two index types `ordered` and `vector`, so they each get their own namespace within the `@index` directive. 

We move all the existing `vectorIndex` arguments to the `vector` argument within `@index`. 
eg: `@index(vector: {dimensions: 128, HNSW: {m: 40, efConstruction: 100, ...}})`

Specifically for `vector` indexes, we have algorithm such as `hnsw` and future algs like `ivfflat`, `ssg`, etc. We can leave the current `hnsw` arguments, but if you want to use the configured defaults you would have `@index(vector: { ..., hnsw: {}})` which is a very awkward and misleading syntax. So, I also propose to add a new `alg` argument to the `vector` namespace, which is optional but can be used to accept defaults.

Eg: `@index(vector: {..., alg: hnsw})`

Validation rules: 
- Either `alg: hnsw` or `hnsw: {...}` can be used, they imply the same thing. 
- You cant mix algorithm types (eg if we have `ivfflat` then we cant do `vector: {..., alg: hnsw, ivfflat: {...}}`

Looking at all the other index types we want to add in the short and long term such as `fullText` and `trigram`, along with the existing `ordered` index, the full set of directive namespace is:
- `@index(vector: ...)`
- `@index(ordered: ...)`
- `@index(trigram: ...)`
- `@index(fullText: ...)`

This highlights one more important syntax design, some index types either dont have arguments, or are commonly used with default values. In such cases (like the vector `alg` example above) we shouldnt use an empty `{}` object to select an index type. Instead, we will have a `kind` argument in lieu of an empty `{}` object.

Eg: `@index(kind: trigram)` instead of `@index(trigram: {})`.

Lastly, the `ordered` index is the same as the existing top level `@index` arguments, and supports the exact same syntax as the current top level `@index` directive. 
Eg: `@index(unique: true)` is equivalent to `@index(ordered: {unique: true})`

--

tldr; unify under `@index` directive, each of the indexes gets its own namespace `vector`, `ordered`, `fullText`, `trigram`, etc. Within `vector` there is an optional `alg` to select an algorithm with default configs. Top level `@index` directive support `kind` argument to select a kind with default configs.

jsimnz:
## Relevant issue(s)

Resolves #5168 

## Description

This unifies the `@vectorIndex` and `@index` directive, while maintaining full backwards compatability, and clean cut outs for adding new index types in the future.

## Tasks

## How has this been tested?

Expanded unit and integration tests. Additional manual testing

- Linux (nixos)

### DIFF
--- a/internal/db/vector_index_test.go
+++ b/internal/db/vector_index_test.go
@@ -82,6 +82,21 @@ func vectorIndexSearch(
 	return docIDs
 }
 
+func TestValidateVectorIndexDescription_EmbeddingRequiresDimensions(t *testing.T) {
+	const fieldName = "embedding"
+	def := client.CollectionVersion{
+		Fields:           []client.CollectionFieldDescription{{Name: fieldName, Kind: client.FieldKind_FLOAT32_ARRAY}},
+		VectorEmbeddings: []client.VectorEmbeddingDescription{{FieldName: fieldName}},
+	}
+	desc := client.NewIndexRequest{
+		Fields: []client.IndexedFieldDescription{{Name: fieldName}},
+		Vector: &client.VectorIndexDescription{},
+	}
+
+	err := validateVectorIndexDescription(def, desc)
+	require.ErrorContains(t, err, "vector index dimensions must be greater than zero")
+}
+
 func TestCollectionVectorIndex_Save_InsertsIntoGraphAndIsSearchable(t *testing.T) {
 	ctx, db, col := newVectorIndexTestDB(t, 3)
 
--- a/internal/db/vector_index_test.go
+++ b/internal/db/vector_index_test.go
@@ -24,7 +24,7 @@ import (
 )
 
 // newVectorIndexTestDB opens an in-memory badger-backed DB with a collection carrying a
-// [Float32!] @vectorIndex field, ready for document writes.
+// [Float32!] @index(vector: {...}) field, ready for document writes.
 func newVectorIndexTestDB(t *testing.T, dimensions int) (context.Context, *DB, client.Collection) {
 	t.Helper()
 	ctx := context.Background()
@@ -35,7 +35,7 @@ func newVectorIndexTestDB(t *testing.T, dimensions int) (context.Context, *DB, c
 	_, err = db.AddCollection(ctx, `
 		type Users {
 			name: String
-			embedding: [Float32!] @vectorIndex(dimensions: `+strconv.Itoa(dimensions)+`, HNSW: {metric: COSINE})
+			embedding: [Float32!] @index(vector: {dimensions: `+strconv.Itoa(dimensions)+`, hnsw: {metric: COSINE}})
 		}
 	`)
 	require.NoError(t, err)
--- a/internal/request/graphql/schema/types/types.go
+++ b/internal/request/graphql/schema/types/types.go
@@ -49,20 +49,23 @@ const (
 	IndexDirectivePropUnique    = "unique"
 	IndexDirectivePropDirection = "direction"
 	IndexDirectivePropIncludes  = "includes"
+	IndexDirectivePropKind      = "kind"
+
+	OrderedIndexKind = "ordered"
 
 	EncryptedIndexDirectiveLabel    = "encryptedIndex"
 	EncryptedIndexDirectivePropType = "type"
 
-	VectorIndexDirectiveLabel          = "vectorIndex"
-	VectorIndexDirectivePropDimensions = "dimensions"
-	VectorIndexDirectivePropHNSW       = "HNSW"
+	VectorIndexKind             = "vector"
+	VectorIndexPropDimensions   = "dimensions"
+	VectorIndexPropAlgorithm    = "alg"
+	VectorIndexPropHNSW         = "hnsw"
+	VectorIndexAlgorithmHNSW    = "hnsw"
+	VectorIndexConfigPropMetric = "metric"
 
-	// Fields of the per-algorithm config objects. `metric` is shared in name but each config declares
-	// its own so a client only sees the knobs of the algorithm it is configuring.
-	VectorIndexConfigPropMetric         = "metric"
-	VectorIndexConfigPropM              = "M"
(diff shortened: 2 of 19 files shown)


internal/db/vector_index_test.go
```diff
@@ -35,7 +35,7 @@ func newVectorIndexTestDB(t *testing.T, dimensions int) (context.Context, *DB, c
 	_, err = db.AddCollection(ctx, `
 		type Users {
 			name: String
-			embedding: [Float32!] @vectorIndex(dimensions: `+strconv.Itoa(dimensions)+`, HNSW: {metric: COSINE})
+			embedding: [Float32!] @index(vector: {dimensions: `+strconv.Itoa(dimensions)+`, hnsw: {metric: COSINE}})
```
jsimnz, line 38: this was a slight nitpick of mine having the `HNSW` param name capital. I can leave it as is if theres preference

internal/request/graphql/schema/types/types.go
jsimnz, line 1: I went back and forth on this, I might revert again. Current I have changed it so the `includes` index param is scoped to the `ordered` index type, but there are some other index types that might support compound indexes or just prefer to be defined on the schema object itself with the explicit field listed, instead of at the field level.

internal/request/graphql/schema/vector_index_parse_test.go
```diff
@@ -67,6 +71,48 @@ func TestParseVectorIndex_OnField_ParsesArgsAndDefaults(t *testing.T) {
 				},
 			},
 		},
+		{
+			description: "vector kind uses vector and HNSW defaults",
+			sdl: `type user {
+				embedding: [Float32!] @index(kind: vector)
```
jsimnz: This is a bug I noticed before push but wanted to leave it in as a point of discussion.

This PR obviously (among many things) adds this `kind` parameter. Its added to make nicer/clearer syntax for using an index type with a default settings (otherwise you would need to do `vector: {}`). This is more obvious for the `ordered` index that easily supports default config options (ie: `@index(kind: ordered)`)

In this example, it actually *doesnt* make sense to only rely on default values for the `vector` index type, because we *need* a >0 `dimension` configured value.

There are other index types we will add, some support fully default config, some dont.

So, should we include the index types that *dont* support default values soley from the `kind` Enum. If yes, this would remove the `vector` option from the `kind` enum

islamaliev: I'm not sure `kind` is necessary at least at this point. I assume you can still specify an ordered index with just `@index` on a field. If you want to add details you do `@index(ordered: {unique: true})`. If you need a vector index you use `@index(vector: {dimensions: 3})`. We might need it later at some point to make intent more explicit, but for now I'd leave it out

internal/db/vector_index_test.go
islamaliev: I have no preference. Maybe slight inclination toward lower case

islamaliev (approved):
Based on your own comments it feels like you didn't quite finished yet the pr but so far it looks legit to me.

internal/request/graphql/schema/vector_index_parse_test.go
jsimnz: My hope is merge the `trigram` and `bm25` indexes pretty soon into the 1.2 release cycle, where this will certainly show up, since `trigram` doesnt really have params that need to be configured.

internal/request/graphql/schema/types/types.go
islamaliev: I'd keep it scoped to ordered and not revert. `includes` isn't really a generic field list as it carries per-field direction, which only means something for an ordered index. I think Trigram and BM25 both take a field list without directions.

Another point is that there are already legacy param on top level and if we add more it gets more complex for parser. A future kind that needs a field list should declare its own inside its own config object

internal/request/graphql/schema/types/types.go
```diff
@@ -106,7 +109,36 @@ func OrderingEnum() *gql.Enum {
 	})
 }
 
-// VectorDistanceMetricEnum is an enum for the `metric` field of a @vectorIndex algorithm config.
+func IndexKindEnum() *gql.Enum {
```
islamaliev, line 112: I think I mentioned this already somewhere, maybe on the stand up or our 1-on-1 or somewhere where we did discuss it, but I don't think `kind` Is strictly necessary at least at this stage. I think we can easily add it later if needed. But if later we decide to remove it, that would be a breaking change.

internal/request/graphql/schema/types/types.go
jsimnz: good point

internal/request/graphql/schema/types/types.go
jsimnz: You mentioned it on one of the other comments on this PR. My reason for having it is the pending index work that will be landing soon after this in the 1.2 cycle, which needs to take advantage of the `kind` param

islamaliev: I have nothing against having to specify index like `@index(trigram: {})` or `@index(bm25: {})`

jsimnz: It was discussed on the standup last monday and there was overwhelming support to not use this syntax, as it can be confusing what exactly `{}` is trying to say

AndrewSisley:
Bug bashing the following PRs together (wip):
- This one
- https://github.com/sourcenetwork/defradb/pull/5169
- https://github.com/sourcenetwork/defradb/pull/5096

Bugs found:
- https://github.com/sourcenetwork/defradb/issues/5219
- https://github.com/sourcenetwork/defradb/issues/5220

PR #5189, opened 2026-08-24
feat: Add extensions to GraphQL responses

### LINKED ISSUE #5175: Support the GraphQL extensions field in responses
GraphQL responses currently carry only `data` and `errors`. There is no way for the database to tell the caller anything about a request that is not either the result itself or a failure, and several pieces of planned work need exactly that.

The GraphQL spec reserves a top-level `extensions` object for this: an implementation-defined map that travels next to `data` and `errors` and is ignored by clients that do not care about it. Adding it is a small change to the response type, but it has to reach every client we ship, and the current serialization makes that less automatic than it looks.

## Why now

Three separate pieces of work are already waiting on this channel, none of which can be done well without it.

Moving the explain output out of `data` (https://github.com/sourcenetwork/defradb/issues/4063) needs somewhere else for that output to go. Today an explain query returns a differently shaped result under the field name of the collection being queried, which does not match the schema the caller asked against.

Warning that a similarity query fell back to a full scan (https://github.com/sourcenetwork/defradb/issues/5173) needs a way to say so without failing the request. The query is valid and the results are correct; it is just far more expensive than the author expected.

Filtered nearest-neighbour search (https://github.com/sourcenetwork/defradb/issues/5071) needs to report that it returned fewer documents than were asked for because the candidates it examined ran out, not because no more matching documents exist. Without somewhere to say that, a short result is indistinguishable from an exhaustive one, which reads as a correctness bug. This is the case that makes the channel load-bearing rather than convenient: it is what allows the feature to return an honest partial answer instead of silently guessing on the caller's behalf.

## Scope

Add an `extensions` field to the GQL response and carry it, intact, across every client.

The field is a map, populated only when there is something to report and omitted entirely otherwise, so responses for ordinary requests do not change shape.

The one thing worth deciding up front is what an entry looks like. Warnings want to be machine-readable: a caller should be able to detect "this result was truncated" and react to it without matching on English prose. That points at each entry carrying a stable identifier alongside any human-readable message and whatever structured detail is useful, rather than a bare list of strings. Retrofitting structure onto a string is painful once clients depend on the format, so it is cheaper to settle now. The identifiers themselves belong to the issues that introduce them; this issue only needs to fix the shape.

Populating the field is left to the consuming issues. This one delivers the channel and proves it survives the trip.

## What makes this less trivial than it looks

`client.GQLResult` does not use the default struct marshalling. It has hand-written `MarshalJSON` and `UnmarshalJSON` that copy `data` and `errors` field by field through a private `gqlResult` type, so that error values serialize to spec-compliant messages. A new field added to `GQLResult` alone is silently dropped in both directions: it would appear to work on the Go client, which passes the struct through in process, and vanish the moment the response crossed HTTP.

That shim is also the choke point that makes this tractable. We ship five clients, and the four that serialize all route through those two methods rather than enumerating fields themselves: HTTP unmarshals whole responses, the subscription reader does the same per event, the CLI serializes the result as-is, the C bindings marshal the whole struct, and the JS/WASM client delegates to the same encoder. So teaching the shim about the field should carry it everywhere at once, and the work is in confirming that rather than in per-client plumbing.

The decoder does not reject unknown fields, so an older client receiving a response with extensions ignores it rather than failing. Version skew is safe in that direction.

Two places do not inherit the field and need a deliberate decision rather than an assumption.

Transport-level failures in the HTTP layer are serialized by a separate error envelope that hand-builds its own JSON and never touches `GQLResult`. These are failures raised before or around execution, so there is likely nothing to report on them, but that should be a decision rather than an oversight.

Subscriptions return before the planner runs, and each event builds a fresh result. Anything discovered while parsing or planning would therefore need to be captured and re-attached per event, or it would appear only on the initial response, which for a subscription is empty. Whether extensions repeat on every event or appear once is a real question and worth answering explicitly.

## Testing

Integration tests confirming that a response carrying extensions arrives intact over each of the serializing clients, HTTP, CLI, C and JS, and over a subscription; that a response with nothing to report omits the field rather than sending an empty object; and that a client which ignores extensions is unaffected by their presence.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5175

## Description

This PR adds the spec's third top level field, `extensions`, to GraphQL response besides existing `data` and `errors` and a path for the execution layer to attach to it. Nothing fills it yet. It is left out of the response when there is nothing to report, so existing responses do not change.

Instead of developing some contrived integration tests specifically for this, with some ways to inject the warnings, I decided to make it only unit test tested. The next follow ups that I'm working on with vector indexes directly use it, and this will be proof on the integration test side that it's working.

## Tasks

## How has this been tested?

Unit tests.

- macOS


nasdf (approved):
LGTM

Issue #5176, opened 2026-08-20
JS client loses int64 precision crossing the Go/JS boundary via goji.MarshalJS


ChrisBQu:
The Go-side json.Marshal step is precision-safe, but the result is then parsed by JavaScript's own JSON.parse, which has no int64-aware parsing mode. Rvery JSON number, regardless of how it was produced, becomes a JS Number (IEEE 754 double) the instant it crosses this boundary. The result is that any integer above 2^53 is rounded to the nearest representable double at that point, silently. 

For context relating to this, see: https://github.com/sourcenetwork/defradb/pull/5162

Specifically, look at the test which must be disabled for the JS client: `TestCollectionVersionPatch_LargeIntegerLensArguments_AreNotDeduplicated`

PR #5192, opened 2026-08-24
fix: Preserve filter type/int precision in doc updates

### LINKED ISSUE #5191: Update/delete filters may be parsed incorrectly
56gShahzad made this observation in another PR:

> Consider this:
> 
> Some(Filter{Conditions: C}) becomes {"Conditions": C} after cbindings/document_update.go:90-98 decodes that into:
> 
> map[string]any, makeSelectionPlan treats Conditions as an actual document field instead of using C.
> 
> None becomes untyped nil
> 
> Which then i think will fail as an unsupported filter instead of selecting all documents.
> 
> Raw map[string]any filters should be accepted but every number is decoded as float64: int64(9007199254740993) rounds to 9007199254740992, and Connor's mixed numeric comparison can then match and update/delete the document containing the latter value instead of the requested former one. direct Go passes both option and map values unchanged.
> 
> possibly can improve this by carrying a tagged filter representation or otherwise reconstruct the exact supported types and number lexemes.

It was proposed that this exclusively affected the C bindings, but that may not be true. Proper resolution of this issue would entail investigating this, and then creating integration tests that illustrate the behavior.

ChrisBQu:
## Relevant issue(s)

Resolves #5191 

## Description

 Filters sent to `UpdateDocumentsWithFilter`/`DeleteDocumentsWithFilter` through any client that round-trips them as JSONcould silently target the wrong documents. This included the C bindings, as well as the HTTP client, CLI, and JS clients.

There were actually a number of issues going wrong here. `immutable.Some(request.Filter{...})` was marshaled to `{"Conditions": ...}`, which decoded back into a filter with a literal `Conditions` field instead of the intended conditions. Further, raw `map[string]any` filters lost integer precision on decode (encoding/json decodes numbers as float64), so a condition on an integer above 2^53 could match a different document than intended. This was an issue that appeared, and was addressed in, several other recent PRs.

This PR introduces new functions, `NormalizeFilterrForJSON` and `DecodeJSONFilter` used on the encoding and decoding sides, respectively, to properly wire the values into the clients code. 

To attain parity across the clients as much as possible, the JS client's filter parameter now also accepts an object as well as a filter string. **However, because JavaScript does not support large integers, we are limited on that particular part off the issue. See: https://github.com/sourcenetwork/defradb/issues/5176**

This PR contains new integration tests to illustrate the behavior. This includes two on which JS client support is marked as a to-do.

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/tests/integration/mutation/delete/with_filter_action_test.go
+++ b/tests/integration/mutation/delete/with_filter_action_test.go
@@ -69,15 +69,6 @@ func TestDeleteWithFilter_Succeeds(t *testing.T) {
 
 func TestDeleteWithMapFilter_Succeeds(t *testing.T) {
 	test := testUtils.TestCase{
-		SupportedClientTypes: immutable.Some(
-			[]state.ClientType{
-				state.GoClientType,
-				state.HTTPClientType,
-				state.CLIClientType,
-				state.CClientType,
-				state.JSClientType,
-			},
-		),
 		Actions: []any{
 			&action.AddCollection{
 				SDL: `
@@ -191,15 +182,6 @@ func TestDeleteWithMapFilter_LargeIntegerStraddling2To53_DeletesOnlyExactMatch(t
 
 func TestDeleteWithSomeOptionFilter_Succeeds(t *testing.T) {
 	test := testUtils.TestCase{
-		SupportedClientTypes: immutable.Some(
-			[]state.ClientType{
-				state.GoClientType,
-				state.HTTPClientType,
-				state.CLIClientType,
-				state.CClientType,
-				state.JSClientType,
-			},
-		),
 		Actions: []any{
 			&action.AddCollection{
 				SDL: `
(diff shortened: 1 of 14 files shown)


tests/integration/mutation/delete/with_filter_action_test.go
```diff
@@ -63,6 +67,238 @@ func TestDeleteWithFilter_Succeeds(t *testing.T) {
...
 
+func TestDeleteWithMapFilter_Succeeds(t *testing.T) {
+	test := testUtils.TestCase{
+		SupportedClientTypes: immutable.Some(
+			[]state.ClientType{
+				state.GoClientType,
+				state.HTTPClientType,
+				state.CLIClientType,
+				state.CClientType,
+				state.JSClientType,
```
AndrewSisley: question:  This looks like all of our client types are listed here, why have you done this?

todo: Please document why, or remove it.  Please apply this to the other similar tests below.

AndrewSisley (approved):
LGTM, thanks Chris :)

PR #5194, opened 2026-08-25
refactor: Move CRDT periphery logic into crdt declaration

### LINKED ISSUE #5115: Self-contain CRDTs

AndrewSisley:
## Relevant issue(s)

Partially resolves #5115

## Description

Moves most of the CRDT periphery logic into the CRDT declarations.

Areas remaining:
- Collection API - document.save - this will be a bit more involved and I wanted to get the easy bits in earlier, so that PR can be a bit more focused
- CType marshalling logic - I just didn't feel like playing with this right now - there is a CType.Display func used by the remote clients (easy), and the PatchCollection substitution logic (fiddly, main reason for procrastination here), both are low value IMO

The commits should be clean, and contain some addition documentation.  Sorry this PR is not so cohesive perhaps - please flag any commit you are not so comfy with in isolation, and would prefer to have the additional context that would come with the final document refactor.

### DIFF
--- a/client/ctype.go
+++ b/client/ctype.go
@@ -53,6 +53,8 @@ func (t CType) IsSupportedFieldCType() bool {
 }
 
 // IsCompatibleWith returns true if the CRDT is compatible with the field kind
+//
+// Deprecated: This function will be removed in Defra v2.0.0
 func (t CType) IsCompatibleWith(kind FieldKind) bool {
 	switch t {
 	case PN_COUNTER, P_COUNTER:
--- a/client/ctype.go
+++ b/client/ctype.go
@@ -68,6 +68,8 @@ func (t CType) IsCompatibleWith(kind FieldKind) bool {
 }
 
 // String returns the string representation of the CRDT.
+//
+// Deprecated: This function will be removed in Defra v2.0.0
 func (t CType) String() string {
 	switch t {
 	case NONE_CRDT:
--- a/client/ctype.go
+++ b/client/ctype.go
@@ -41,6 +41,8 @@ const (
 )
 
 // IsSupportedFieldCType returns true if the type is supported as a document field type.
+//
+// Deprecated: This function will be removed in Defra v2.0.0
 func (t CType) IsSupportedFieldCType() bool {
 	switch t {
 	case NONE_CRDT, LWW_REGISTER, PN_COUNTER, P_COUNTER:
--- a/internal/core/crdt/merklecrdt.go
+++ b/internal/core/crdt/merklecrdt.go
@@ -27,6 +27,10 @@ var FieldCRDTs = []FieldValueCRDT{
 	NewCounter(false),
 }
 
+// TryGetFieldCRDT returns the cached instance of the given crdt.
+//
+// A `NONE_CRDT` will return `nil`.  If nothing is found, `false` will
+// be returned, else `true`.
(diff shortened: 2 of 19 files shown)


client/ctype.go
```diff
@@ -41,6 +41,8 @@ const (
 )
 
 // IsSupportedFieldCType returns true if the type is supported as a document field type.
+//
+// Deprecated: This function will be removed in Defra v2.0.0
```
islamaliev, line 45: question: don't we leave a hit on what should be used instead or at least where to look for users?

internal/core/crdt/merklecrdt.go
```diff
@@ -21,7 +21,36 @@ import (
 	"github.com/sourcenetwork/defradb/internal/keys"
 )
 
+var FieldCRDTs = []FieldValueCRDT{
+	NewLWW(),
+	NewCounter(true),
+	NewCounter(false),
+}
+
+func TryGetFieldCRDT(ct client.CType) (FieldValueCRDT, bool) {
```
islamaliev, line 34: suggestion: Would be nice to get a comment on this function, especially documenting what the bool indicates.

internal/db/index.go
```diff
@@ -136,7 +137,8 @@ func buildIndexBase(
 			return collectionBaseIndex{}, NewErrUnsupportedIndexFieldType(field.Kind)
 		}
 		if field.Typ == client.PN_COUNTER || field.Typ == client.P_COUNTER {
-			return collectionBaseIndex{}, NewErrCannotIndexAccumulatedCRDTField(field.Name, field.Typ.String())
+			ct, _ := crdt.TryGetFieldCRDT(field.Typ)
```
islamaliev, line 140: todo: I know we're operating here on field, but still the function can return nil, and we do not check anything here. So, at least write a comment mentioning why it's safe here.

islamaliev (approved):
LGTM!

client/ctype.go
AndrewSisley: In this case, I thought the enum was plenty descriptive enough, and the function was really just internal validation posing as something useful to users.  I don't think users really need to be told to not try and introduce random integers to the public enum, but I can add something if you feel like they would benefit from such instructions.

internal/core/crdt/merklecrdt.go
AndrewSisley: Not a problem, will do.

- [x] Document `TryGetFieldCRDT`

internal/db/index.go
AndrewSisley: > I know we're operating here on field

Not just a field, but a field with known, constant, crdt, described on the line immediately above this one.  I think adding a line of documentation between these would make the code more confusing - the documentation provided by line 138 is clear and concise, and should not be distanced from the `TryGetFieldCRDT` call.

My understanding is also that this is quite temporary code here, and will not live too much longer.

PR #5197, opened 2026-08-27
feat: Warn when a similarity query does not use the vector index

### LINKED ISSUE #5173: Warn when a similarity query falls back to a full scan
Part of https://github.com/sourcenetwork/defradb/issues/5070. Follows the HNSW index (https://github.com/sourcenetwork/defradb/issues/5042).

A similarity query uses the vector index only when it takes a particular shape: ordered by the similarity descending, with a limit, no filter, and a single similarity field. Miss any of those and the query still returns correct results, it just reads the whole collection and scores every document on the way. There is no error and nothing in the response to say what happened, so a query that looks fine in development quietly becomes expensive as the collection grows.

The sort direction is the easiest to get wrong. Nearest means the largest similarity, so it has to be descending; ascending is a legitimate query asking for the farthest documents, which the index cannot serve. Someone thinking in terms of distance rather than similarity will reach for ascending and get a full scan.

Other databases have the same trap, so the aim is not to reject these queries but to say when one of them is about to do far more work than intended. The natural place is a warning in the response next to the data. Our GraphQL responses currently carry only data and errors, so a channel for this has to be added first; the standard spelling is a top-level extensions object.

## Testing
Integration tests confirming a query that uses the index reports no warning, that each of the shapes that fall back reports one, and that the warning travels intact over the HTTP, CLI and C clients.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5173

## Description

A similarity query uses the vector index only in one shape: a single similarity field, ordered by it descending, with a limit and no filter. Miss any of those and the query still returns the right answer, it just reads the whole collection on the way. Nothing in the response said so, so a query that looks fine in development gets slower as the collection grows.

It now reports a warning in `extensions` naming which part of the shape ruled the index out, so the author knows what to change. A field with no vector index does not warn, since there is nothing to fall back from.

This also fixes a separate bug it uncovered, worth reviewing on its own. A filtered similarity query was reaching the index when it should not have: the guard against it read the filter from a field that is already cleared by that point, so it never fired. The index returned the nearest documents, the filter then rejected some or all of them, and nothing refilled the gap, so the query could return fewer documents than match, sometimes none, while matching documents sat there unreturned. It now reads the filter from where it actually lives. This is on `develop` today.

Draft until https://github.com/sourcenetwork/defradb/pull/5189 merges, which this builds on.

## Tasks

## How has this been tested?

Integration tests for each shape that falls back, plus the shape the index serves as the control, so the tests fail if every query starts warning. Test support was added so a request asserts its warnings, with expecting none as the default, which caught existing tests that had started warning.

The fallback tests run on the HTTP, CLI and C clients. That is what proves the warning survives serialization, which the extensions work could not test on its own.

The data loss fix has a regression test that fails without it: a filter matching one document the graph would not have returned.

Full integration suite green.

- macOS


islamaliev:
is already resolved as part of https://github.com/sourcenetwork/defradb/pull/5189

PR #5221, opened 2026-09-01
fix: Prevent node from leaking during C binding CloseNode

### LINKED ISSUE #5206: Possible leak of the node `cgo.Handle` when `node.Close` fails in `CloseNode`
# Possible leak of the node `cgo.Handle` when `node.Close` fails in `CloseNode`

`CloseNode` deletes the handle only on the success path. If `node.Close` returns an
error the function returns first, and the handle is never deleted.

cbindings/node_close.go:24

```go
//export CloseNode
func CloseNode(nodePtr C.uintptr_t) C.Result {
	node, err := getNodeFromPointer(nodePtr)
	if err != nil {
		return returnC(returnGoC(1, err.Error(), ""))
	}
	err = node.Close(context.Background())
	if err != nil {
		return returnC(GoCResult{1, err.Error(), ""})
	}
	cgo.Handle(nodePtr).Delete()
	return returnC(GoCResult{0, "", ""})
}
```

A live `cgo.Handle` is a strong reference in the runtime's handle map, so the leaked
entry keeps the whole `*node.Node` reachable: its store, its P2P peer and everything they
reference stay alive for the process lifetime. The handle slot is not reusable either.
`CloseNode` is the only place that deletes a node handle, so once this path is taken
there is no second chance to release it.

The Go caller does not notice. cbindings/wrapper.go:1112:

```go
func (w *CWrapper) Close() {
	C.CloseNode(C.uintptr_t(w.handle))
}
```

The `C.Result` is discarded, so a failed close is silent and the wrapper is dropped with
its handle still registered.

The transaction bindings in the same package already use the pattern that survives an
early return. cbindings/transaction_commit.go:25:

```go
//export CommitTransaction
func CommitTransaction(txnPtr C.uintptr_t) C.Result {
	h := cgo.Handle(txnPtr)
	defer h.Delete()
	txn := h.Value().(client.Txn) //nolint:forcetypeassert

	err := txn.Commit()
	if err != nil {
		return returnC(returnGoC(1, err.Error(), ""))
	}

	return returnC(returnGoC(0, "", ""))
}
```

`CommitTransaction` releases the handle whether or not `Commit` fails; `CloseNode` does
not.

Reaching it needs `node.Close` to return an error, so this is an error-path leak rather
than one that shows up on a normal shutdown. `node.Close` propagates errors from the
underlying store and from the P2P shutdown, so a node that is closed while its datastore
is unhappy hits it.

Fix, matching what the transaction bindings do:

```go
//export CloseNode
func CloseNode(nodePtr C.uintptr_t) C.Result {
	node, err := getNodeFromPointer(nodePtr)
	if err != nil {
		return returnC(returnGoC(1, err.Error(), ""))
	}
	defer cgo.Handle(nodePtr).Delete()

	if err := node.Close(context.Background()); err != nil {
		return returnC(GoCResult{1, err.Error(), ""})
	}
	return returnC(GoCResult{0, "", ""})
}
```

`CWrapper.Close` could also surface the result instead of discarding it, so a caller can
tell that the node did not shut down cleanly.

If you could credit me as a reporter for my contributions to security advisory I will be thankful.

ChrisBQu:
## Relevant issue(s)

Resolves #5206 

## Description

It was the case that the C bindings' `CloseNode` function would leak the node if the `node.Close` call, for some reason, failed. This was because this failure path leads to the error being returned, and the CGO handle not being deleted. This PR fixes that by moving the delete above the `node.Close` call, and having it deferred. 

## Tasks

## How has this been tested?

- WSL

### DIFF
--- a/cbindings/node_close.go
+++ b/cbindings/node_close.go
@@ -27,10 +27,10 @@ func CloseNode(nodePtr C.uintptr_t) C.Result {
 	if err != nil {
 		return returnC(returnGoC(1, err.Error(), ""))
 	}
-	err = node.Close(context.Background())
-	if err != nil {
+	defer cgo.Handle(nodePtr).Delete()
+
+	if err := node.Close(context.Background()); err != nil {
 		return returnC(GoCResult{1, err.Error(), ""})
 	}
-	cgo.Handle(nodePtr).Delete()
 	return returnC(GoCResult{0, "", ""})
 }


AndrewSisley (approved):
LGTM, nice find Chris

PR #5224, opened 2026-09-02
fix: Keep vector distances in float64 to avoid overflow

### LINKED ISSUE #5220: Euclidean similarity appears to overflow
The order of results returned from a euclidean similarity indexed appears to be incorrect once values get close to Float32.MaxValue - they seem to flip to the other side of the result set, suggesting a silent overflow.

Setup:
```
defradb client collection add '
  type User {
    name: String
    age: Int
    verified: Boolean
    vector: [Float32!] @index(vector:{dimensions:1, hnsw:{metric: EUCLIDEAN}})
  }
'
defradb client query '
  mutation {
      add_User(input: {age: 31, verified: true, vector: [-0.90], name: "Alice"}) {
          _docID
      }
  }
'
defradb client query '
  mutation {
      add_User(input: {age: 31, verified: true, vector: [90], name: "Bob"}) {
          _docID
      }            
  }                                                                      
'
defradb client query '
  mutation {
      add_User(input: {age: 31, verified: true, vector: [0], name: "Fred"}) {
          _docID
      }            
  }                                                                      
'
// The greatest negative value that can be accurately held in a float32
defradb client query '
  mutation {
      add_User(input: {age: 31, verified: true, vector: [-340282346638528859811704183484516925440], name: "John"}) {
          _docID
      }
  }       
'
```
Querying with an identical vector correctly returns John first:
```
defradb client query '
  query {
    User(order:{_alias:{sim:DESC}}, limit: 100) {
      _docID
      age
      name
      vector
      sim: SIMILARITY(vector:{vector:[-340282346638528859811704183484516925440]}) }
  }
'
```
Removing the last digit from the query vector returns John last, even though it is the most similar value.  The value of `sim` is given as `-9.379158110139737e+76` (the `vector` value is stated to be ~`-3.4028235e+38`).
```
defradb client query '
  query {
    User(order:{_alias:{sim:DESC}}, limit: 100) {
      _docID
      age
      name
      vector
      sim: SIMILARITY(vector:{vector:[-34028234663852885981170418348451692544]}) }
  }
'
```
Also, documents with positive `vector` values are returned before negative ones, suggesting that they too either overflowed or lost precision (the top 3 all have the same printed `sim` value, so they might have been judged to be equal in terms of order) during the similarity calculation - as they are less similar.
```
------ Request Results ------
{
  "data": {
    "User": [
      {
        "_docID": "bae-88d27ca4-7803-5e51-a36e-79da74216e27",
        "age": 31,
        "name": "Bob",
        "sim": -1.1579207543382391e+75,
        "vector": [
          90
        ]
      },
      {
        "_docID": "bae-58a06828-7a4a-53de-82e6-fe676e1efe3d",
        "age": 31,
        "name": "Alice",
        "sim": -1.1579207543382391e+75,
        "vector": [
          -0.9
        ]
      },
      {
        "_docID": "bae-957a2f6f-180a-5dbc-a179-c84d048ad9cd",
        "age": 31,
        "name": "Fred",
        "sim": -1.1579207543382391e+75,
        "vector": [
          0
        ]
      },
      {
        "_docID": "bae-5c365f2e-f396-53bd-bd9b-095df7d468ee",
        "age": 31,
        "name": "Bob",
        "sim": -9.379158110139737e+76,
        "vector": [
          -3.4028235e+38
        ]
      }
    ]
  }
}
```
The least significant `sim` query parameter that returns the first 3 documents in the correct order is `-340282346638528859`, which is well below the limits of a float32.  John remains the last value returned when `sim` is anything but an exact match.

Whilst I have only tested this with Euclidean Similarity, it seems probable that this issue exists with some of the others too.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5220

## Description

Euclidean vector search now returns results in the right order when the vectors hold large values. Past a certain size the distance between two documents no longer fit in the number type used to compare them, so it came out as infinity: every far document looked equally far, and the order among them was whatever the search happened to produce. Comparisons are now done in a wider type, which covers the full range any stored vector can produce.

Only the comparison changed. Vectors are still stored as they were, and the similarity score in a response was already computed separately and was never affected, which is why the reported scores looked sensible while the ordering did not.

## Tasks

## How has this been tested?

Unit and integration tests, both using values from the issue. Confirmed they fail without the fix.

- macOS

### DIFF
--- a/tests/integration/index/vector_metrics_test.go
+++ b/tests/integration/index/vector_metrics_test.go
@@ -262,3 +262,37 @@ func TestVectorIndex_DropThenRecreateWithDifferentMetric_IsAllowed(t *testing.T)
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+// Squaring a large vector component overflows a float32, which used to leave every distance equal
+// and the results in an arbitrary order. The query matches "john" exactly, so it comes first and the
+// rest follow by how far they are.
+// https://github.com/sourcenetwork/defradb/issues/5220
+func TestVectorIndex_EuclideanOnLargeVectors_OrdersByDistance(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddCollection{
+				SDL: `type User {
+					name: String
+					vector: [Float32!] @index(vector: {dimensions: 1, hnsw: {metric: EUCLIDEAN}})
+				}`,
+			},
+			&action.AddDoc{DocMap: map[string]any{"name": "alice", "vector": []float32{-0.9}}},
+			&action.AddDoc{DocMap: map[string]any{"name": "bob", "vector": []float32{90}}},
+			&action.AddDoc{DocMap: map[string]any{"name": "john", "vector": []float32{-3.4028235e+38}}},
+			&action.WaitForIndexReady{CollectionID: 0},
+			&action.Request{
+				Request: `query {
+					User(order: {_alias: {sim: DESC}}, limit: 1){
+						name
+						sim: SIMILARITY(vector: {vector: [-3.4028235e+38]})
+					}
+				}`,
+				Results: map[string]any{
+					"User": []map[string]any{{"name": "john", "sim": float64(0)}},
+				},
+			},
+		},
+	}
+
+	testUtils.ExecuteTestCase(t, test)
+}
--- a/tests/integration/index/vector_metrics_test.go
+++ b/tests/integration/index/vector_metrics_test.go
@@ -298,3 +298,75 @@ func TestVectorIndex_EuclideanOnLargeVectors_OrdersByDistance(t *testing.T) {
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+// The dot product of two large vectors also exceeds a float32, so it needs the same float64 handling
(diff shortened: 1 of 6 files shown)


tests/integration/index/vector_metrics_test.go
```diff
@@ -262,3 +262,39 @@ func TestVectorIndex_DropThenRecreateWithDifferentMetric_IsAllowed(t *testing.T)
 
 	testUtils.ExecuteTestCase(t, test)
 }
+
+// Vectors whose squared distance exceeds a float32 are searchable and scored correctly. The graph's
+// own ordering is covered by the engine tests; this covers the query path end to end, where the
+// score is computed separately and was never affected.
+// https://github.com/sourcenetwork/defradb/issues/5220
+func TestVectorIndex_EuclideanOnLargeVectors_OrdersByDistance(t *testing.T) {
```
AndrewSisley, line 270: suggestion: Whilst they look like they work right now, it looks plenty possible to break cosine and doc metric'd indexes in the future with the same issue - I suggest also covering them with integration tests.

AndrewSisley (approved):
LGTM, thanks Islam!

PR #5227, opened 2026-09-02
fix: Prevent DB.Connect from panicking with P2P disabled

### LINKED ISSUE #5226: DB.Connect panics with nil pointer dereference when P2P is disabled
The `DB.Connect` function inside `internal/db/p2p.go` lacks a `db.p2p == nil` check that is present in its sibling `Disconnect` function. If you start a node with P2P disabled, `db.p2p` will remain nil. As a result, a call made on`db.p2p.Connect` will panic.

Over HTTP, this will degrade to a 500 error. However, on embedded clients, this will result in a crash.

### LINKED ISSUE #5230: Test harness should support nodes with P2P disabled
This is currently not an option that the configuration has, which makes testing certain error paths impossible. This should be added.

ChrisBQu:
## Relevant issue(s)

Resolves #5226
Resolves #5230

## Description

This PR adds a nil guard to the `DB.Connect` function. This change prevents a potential crash from occurring in the embedded clients. Prior, the crash could have been triggered by disabling P2P, but then calling `DB.Connect` with a non-nill addresss.

To be able to test this fix, this PR also extends the test harness. A new flag has been added to the `NewNode` struct: `DisableP2P`, and it is used in the `Execute` function to assign the value to the actual node object. Changes also include a new functions for creating such a configured node with P2P disabled, inside the `p2p_config.go` and `p2p_config_js.go` files: `NoNetworkingConfig()`. 

With these changes, integration testing becomes possible. See: `TestNetInfoConnectPeers_SourceP2PDisabled`

## Tasks

## How has this been tested?

- WSL


internal/db/p2p_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestDB_Connect_WithoutP2P_ReturnsError tests that an error will be returned if
+// Conneect is called with P2P disabled.
+func TestDB_Connect_WithoutP2P_ReturnsError(t *testing.T) {
```
AndrewSisley: suggestion: This should really be an integration test.  I'm leaving this as a suggestion as the existing `NewNode` action looks like it doesn't yet allow full configuration of the node - if you want to put that effort in it would be great, but no worries if not.

AndrewSisley (approved):
LGTM, thanks Chris

shahzadlone (approved):
LGTM

internal/db/p2p_test.go
ChrisBQu: You're spot on for why it wasn't done as an integration test. I don't mind putting in the work to allow for it, but it seemed inappropriate to do during a feature freeze bug bash. I'm happy to use a unit test for now, and come back to this with an integration test next cycle, if you are?

AndrewSisley: > inappropriate to do during a feature freeze bug bash

No production change should be needed, if `options.DBOptions.P2P` is `None`, `db.p2p` will be nil and you can test this - only the `NewNode` action should need to change (to expose the public config, which it should have been doing anyway)

PR #5231, opened 2026-09-03
fix: Render field Kind and Typ as strings on collection

### LINKED ISSUE #4816: Display string representation of field kind on collection describe

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/4816

## Description

Adding a collection or a view now reports each field's type and CRDT the way describing one already did, as readable names rather than bare numbers that only mean something to someone reading the source. Describing a collection was fixed in https://github.com/sourcenetwork/defradb/pull/4945, but the two add paths kept returning the numeric form, over both the CLI and HTTP.

The existing coverage could not have caught this. A field type decodes from either the numeric or the readable form, so a test that compares the decoded result passes either way; the new tests assert the response itself.

## Tasks

## How has this been tested?

Unit and integration tests, plus the reported command against a running node.

- macOS


http/handler_store.go
```diff
@@ -128,7 +128,13 @@ func (h *storeHandler) AddCollection(rw http.ResponseWriter, req *http.Request)
...
 	}
 
-	responseJSON(rw, http.StatusOK, cols)
+	display, err := client.DisplayCollectionVersions(cols)
+	if err != nil {
+		responseJSON(rw, httpStatusFromError(err), errorResponse{err})
+		return
+	}
+
+	responseJSON(rw, http.StatusOK, display)
```
nasdf, line 137: question: Does this change the OpenAPI response type as well?

nasdf (approved):
Fix looks good to me but I would double check that it doesn't need a change to the OpenAPI response types before merging.

PR #5233, opened 2026-09-03
fix: Reject a direction on a vector index

### LINKED ISSUE #5232: No tests for vector indexes with `direction` parameter
We have no tests for vector indexes with `direction` parameter, for example:
```
type user {
	embedding: [Float32!] @index(vector: {dimensions: 3, alg: hnsw}, direction: ASC)
}
```
Current behaviour is an `index with invalid argument` error.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5232

## Description

Asking for a vector index with a sort direction is now rejected everywhere. Through the schema it already was, but the Go, HTTP, CLI and C clients accepted it, saved the direction, and then ignored it, while still naming the index ascending. A vector index is searched by nearness rather than read in order, so there is nothing for a direction to do, and silently keeping one meant the stored index disagreed with its own name.

The schema error also said only that an argument was invalid. Since the argument you pass is what picks the kind of index, naming a vector one alongside an ordered one asks for two kinds at once, which is what the error says now.

The issue asked for tests covering this. Writing them is what turned up the clients accepting it.

## Tasks

## How has this been tested?

Integration tests. Three existing parser tests changed: they pinned the old wording for what is now the two-kinds error.

- macOS


tests/integration/collection_version/vector_index_test.go
```diff
@@ -185,3 +185,20 @@ func TestCollectionVersion_VectorIndexWithUnsupportedMetric_ShouldError(t *testi
...
+func TestCollectionVersion_VectorIndexWithDirection_ShouldError(t *testing.T) {
+	test := testUtils.TestCase{
+		Actions: []any{
+			&action.AddCollection{
+				SDL: `
+					type Users {
+						embedding: [Float32!] @index(vector: {dimensions: 3, hnsw: {metric: COSINE}}, direction: ASC)
+					}
+				`,
+				ExpectedError: "index arguments ask for two different kinds of index",
```
AndrewSisley: suggestion: To a user, I do not think this error message will help a lot - they have only provided a single index type.  Why not allow the db validation to return the error?

internal/request/graphql/schema/collection.go
```diff
@@ -280,9 +280,11 @@ type indexDirectiveConfig struct {
 	vector  ast.Value
 }
 
+// selectKind records which kind of index the arguments seen so far ask for. The argument chooses the
+// kind, so two arguments belonging to different kinds are a conflict, not an unknown argument.
 func (c *indexDirectiveConfig) selectKind(kind string) error {
 	if c.kind != "" && c.kind != kind {
-		return ErrIndexWithInvalidArg
+		return NewErrIndexKindConflict(c.kind, kind)
```
AndrewSisley: question: Why is this validation being performed here?  It does not appear to be GQL specific, and looks like it is preventing the db validation from returning a more helpful error message.

AndrewSisley (approved):
LGTM

Issue #5104, opened 2026-07-31
Refactor IndexDescription to an interface-typed config (FieldKind-style)


islamaliev:
Part of #5070.

### Context

The vector index feature (#5042) needs to hang algorithm-specific config (algorithm, metric, dimensions, HNSW params) off `client.IndexDescription`. To keep that PR a non-breaking semver-minor change, it only adds a nullable `Vector *VectorIndexDescription` field and leaves the rest of the descriptor as it was on `develop` (top-level `Unique bool`).

During review, Andy suggested a cleaner model for the descriptor: the different index kinds (normal/unique, vector, and future text/BM25) are mutually exclusive, so the type should reflect that rather than have several sibling nullable pointer fields whose "exactly one is set" invariant lives in a comment.

### Proposal

Collapse the kind-specific config into a single interface-typed field, following the existing `client.FieldKind` pattern (a `FieldKind` interface with concrete `ScalarKind`, `CollectionKind`, etc. that the codebase type-switches on).

Rough shape:

```go
type IndexDescription struct {
    Name   string
    ID     uint32
    Fields []IndexedFieldDescription
    Config IndexConfig // interface; the concrete type is the kind
}

type IndexConfig interface { /* small marker/shared method */ }

type NormalIndexConfig struct { Unique bool }
type VectorIndexConfig struct { Algorithm ...; Metric ...; Dimensions ...; HNSW ... }
// later: BM25IndexConfig, etc.
```

Consumers type-switch on `Config` instead of nil-checking sibling pointers. This makes illegal states (both set / neither set) unrepresentable and lets each config name itself, sidestepping the "what do we call the non-vector family" naming problem (scalar is too narrow given array/JSON support; secondary is vague).

### Why separate from the feature PR

This is a **breaking change** on two axes:
- Go API: removes/replaces public fields on `client.IndexDescription`, affecting every consumer and external callers.
- Persistence: an interface field needs custom JSON marshalling with a type discriminator (like `FieldKind`), which changes the on-disk shape and needs a migration for existing descriptors.

That is a semver-major change and should land on the appropriate next-major branch, migrating from a single known shape (the additive one shipped by #5042), rather than riding on the feature PR.

### Acceptance
- `IndexDescription` uses an interface-typed config field, `FieldKind`-style.
- Custom marshalling with a discriminator; existing persisted descriptors migrate cleanly.
- Consumers (planner, db, http, fetcher) type-switch on the config.
- No behaviour change for existing normal/unique indexes.

PR #5235, opened 2026-09-04
refactor: Make each index kind own its fields

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/5104

## Description

Each kind of index now describes its own fields, in the shape it can actually honour. An ordered index keeps a direction per field because it reads its entries in key order; a vector index takes names only, since a graph is searched by nearness and has nothing to sort. Asking a vector index for a direction is no longer something we accept and then reject, it is something that cannot be written down.

Until now every kind shared one field shape, the one ordered indexes need. That is why a vector index could be handed a direction at all, and why reading an index's fields meant asking which kind it was first. Both the old spelling and the old way of reading it still work and are kept in sync; they are deprecated and go away in v2.0.0.

## Tasks

## How has this been tested?

Unit and integration tests, including a descriptor built the old way in memory rather than read back from storage, which is the one path nothing upgrades on the way in.

- macOS

### DIFF
--- a/cbindings/wrapper_collection.go
+++ b/cbindings/wrapper_collection.go
@@ -93,8 +93,9 @@ func (c *Collection) NewIndex(
 	copts.getInactive = 0
 
 	//nolint:staticcheck // a request has no accessor; the deprecated field is the only source
-	orderedFields := make([]string, len(indexDesc.Fields))
-	for i, f := range indexDesc.Fields {
+	requested := indexDesc.Fields
+	orderedFields := make([]string, len(requested))
+	for i, f := range requested {
 		order := "ASC"
 		if f.Descending {
 			order = "DESC"
--- a/cbindings/wrapper_collection.go
+++ b/cbindings/wrapper_collection.go
@@ -92,6 +92,7 @@ func (c *Collection) NewIndex(
 	copts.name = cName
 	copts.getInactive = 0
 
+	//nolint:staticcheck // a request has no accessor; the deprecated field is the only source
 	orderedFields := make([]string, len(indexDesc.Fields))
 	for i, f := range indexDesc.Fields {
 		order := "ASC"
--- a/client/index.go
+++ b/client/index.go
@@ -44,6 +44,9 @@ const (
 type IndexKindDescription interface {
 	// isIndexKindDescription is unexported so only types in this package can implement this interface.
 	isIndexKindDescription()
+
+	// FieldNames returns the names of the indexed fields, in the order the kind stores them.
+	FieldNames() []string
 }
 
 // VectorAlgorithm identifies the algorithm used to build/search a vector index. It is a string so it
@@ -109,14 +112,31 @@ const (
 type OrderedIndexDescription struct {
 	// Unique indicates whether the index enforces uniqueness.
 	Unique bool
+
+	// Fields are the indexed fields, in key order. Each carries a direction, since an ordered index
+	// is read in key order.
+	Fields []IndexedFieldDescription
 }
 
 func (*OrderedIndexDescription) isIndexKindDescription() {}
+
+// FieldNames returns the indexed field names, dropping each field's direction.
+func (d *OrderedIndexDescription) FieldNames() []string {
+	names := make([]string, len(d.Fields))
+	for i, field := range d.Fields {
+		names[i] = field.Name
+	}
+	return names
+}
 
 var _ IndexKindDescription = (*OrderedIndexDescription)(nil)
 
 // VectorIndexDescription is the config for a vector (ANN) index.
 type VectorIndexDescription struct {
+	// Fields are the names of the indexed fields. Names only: a graph is searched by nearness, so
+	// there is no direction to carry.
+	Fields []string
+
 	// Algorithm is the algorithm used to build/search the index.
 	Algorithm VectorAlgorithm
 	// Metric is the distance metric used to compare vectors.
@@ -129,6 +149,11 @@ type VectorIndexDescription struct {
 
 func (*VectorIndexDescription) isIndexKindDescription() {}
 
+// FieldNames returns the indexed field names, which is all a vector index stores.
+func (d *VectorIndexDescription) FieldNames() []string {
+	return d.Fields
+}
+
 var _ IndexKindDescription = (*VectorIndexDescription)(nil)
 
 // IndexDescription describes an index.
@@ -138,6 +163,9 @@ type IndexDescription struct {
 	// ID is the local identifier of this index.
 	ID uint32
 	// Fields contains the fields that are being indexed.
+	//
+	// Deprecated: use the kind config's Fields. This field will be removed in Defra v2.0.0 and is
+	// kept in sync until then.
 	Fields []IndexedFieldDescription
 
 	// Unique indicates whether the index enforces uniqueness. It only applies to ordered indexes.
@@ -184,15 +212,66 @@ func (d IndexDescription) GetUnique() bool {
 	return d.Unique
 }
 
-// Normalize returns a copy with KindDescription and the compat Unique field consistent: a nil
-// ordered KindDescription is filled from Unique, and Unique is set from the resolved config. Use it
-// to compare two descriptors that may have been built in different styles (only Unique, or only
-// Kind/KindDescription).
-func (d IndexDescription) Normalize() IndexDescription {
+// fields returns the index's fields with their directions. Only an ordered index has them, so a
+// vector index's entries come back ascending. Only [IndexDescription.normalize] needs the
+// directions; fieldNames serves the rest.
+func (d IndexDescription) fields() []IndexedFieldDescription {
+	if ordered, ok := d.KindDescription.(*OrderedIndexDescription); ok && len(ordered.Fields) > 0 {
+		return ordered.Fields
+	}
+	names := d.fieldNames()
+	fields := make([]IndexedFieldDescription, len(names))
+	for i, name := range names {
+		fields[i] = IndexedFieldDescription{Name: name}
+	}
+	return fields
+}
+
+// fieldNames returns the names of the index's fields, falling back to the deprecated top-level ones
+// for a descriptor built in memory before the kind carried them.
+func (d IndexDescription) fieldNames() []string {
+	if d.KindDescription != nil {
+		if names := d.KindDescription.FieldNames(); len(names) > 0 {
+			return names
+		}
+	}
+	//nolint:staticcheck // the fallback this helper exists to hide
+	names := make([]string, len(d.Fields))
+	//nolint:staticcheck // the fallback this helper exists to hide
+	for i, field := range d.Fields {
+		names[i] = field.Name
+	}
+	return names
+}
+
+// normalize returns a copy with KindDescription and the compat fields consistent: a nil ordered
+// KindDescription is filled from the deprecated fields, and those are set back from the resolved
+// config. Both (un)marshalling paths run it, so a stored descriptor is always consistent.
+func (d IndexDescription) normalize() IndexDescription {
 	if d.Kind == IndexKindOrdered && d.KindDescription == nil {
-		d.KindDescription = &OrderedIndexDescription{Unique: d.Unique}
+		//nolint:staticcheck // reading it is how a legacy descriptor is upgraded
+		d.KindDescription = &OrderedIndexDescription{Unique: d.Unique, Fields: d.Fields}
+	}
+	// A caller that set only the old spelling still gets a complete config.
+	switch config := d.KindDescription.(type) {
+	case *OrderedIndexDescription:
+		if len(config.Fields) == 0 {
+			//nolint:staticcheck // upgrading a descriptor that predates the kind carrying fields
+			config.Fields = d.Fields
+		}
+	case *VectorIndexDescription:
+		if len(config.Fields) == 0 {
+			//nolint:staticcheck // upgrading a descriptor that predates the kind carrying fields
+			for _, field := range d.Fields {
+				config.Fields = append(config.Fields, field.Name)
+			}
+		}
 	}
 	d.Unique = d.GetUnique()
+	if fields := d.fields(); len(fields) > 0 {
+		//nolint:staticcheck // kept in sync so old readers still see it
+		d.Fields = fields
+	}
 	return d
 }
 
@@ -243,15 +322,15 @@ func (d *IndexDescription) UnmarshalJSON(bytes []byte) error {
 	default:
 		return NewErrUnknownIndexKind(uint8(mirror.Kind))
 	}
-	// Keep the compat Unique field in sync with the resolved config so old readers see it.
-	d.Unique = d.GetUnique()
+	// Upgrades a descriptor stored before the kind carried its own fields.
+	*d = d.normalize()
 	return nil
 }
 
 // MarshalJSON writes the Kind, its config, and the compat top-level Unique (so a reader that
 // predates [Kind] still gets the uniqueness). They are always emitted consistently.
 func (d IndexDescription) MarshalJSON() ([]byte, error) {
-	d = d.Normalize()
+	d = d.normalize()
 	// The concrete configs are plain structs; their Marshal cannot fail.
 	kindDescription, _ := json.Marshal(d.KindDescription)
 	return json.Marshal(indexDescription{
@@ -271,6 +350,9 @@ type NewIndexRequest struct {
 	// Name contains the name of the index.
 	Name string
 	// Fields contains the fields that are being indexed.
+	//
+	// Deprecated: set them on [NewIndexRequest.Ordered] or [NewIndexRequest.Vector]. This field will
+	// be removed in Defra v2.0.0. Setting both is an error unless they agree.
 	Fields []IndexedFieldDescription
 
 	// Unique indicates whether the index is unique.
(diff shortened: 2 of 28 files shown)


client/index.go
```diff
@@ -44,6 +44,9 @@ const (
 type IndexKindDescription interface {
 	// isIndexKindDescription is unexported so only types in this package can implement this interface.
 	isIndexKindDescription()
+
+	// FieldNames returns the names of the indexed fields, in the order the kind stores them.
+	FieldNames() []string
```
AndrewSisley, line 49: question: Why is this public?  Making it public exposes us to future semver problems.

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+	for i, field := range d.Fields {
+		names[i] = field.Name
+	}
+	return names
+}
+
+// normalize returns a copy with KindDescription and the compat fields consistent: a nil ordered
+// KindDescription is filled from the deprecated fields, and those are set back from the resolved
+// config. Both (un)marshalling paths run it, so a stored descriptor is always consistent.
+func (d IndexDescription) normalize() IndexDescription {
```
AndrewSisley, line 253: todo: Making `Normalize` private is a breaking change, please don't do this.

AndrewSisley (commented):
Part way through review and I need food, submitting comments so far for now.

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+	fields := make([]IndexedFieldDescription, len(names))
+	for i, name := range names {
+		fields[i] = IndexedFieldDescription{Name: name}
+	}
+	return fields
+}
+
+// fieldNames returns the names of the index's fields, falling back to the deprecated top-level ones
+// for a descriptor built in memory before the kind carried them.
+func (d IndexDescription) fieldNames() []string {
```
AndrewSisley, line 237: todo: Please either rewrite or remove this function.  It is badly written and wasteful - I assume by an LLM.

I only spot one call to it, from `IndexDescription.GetFields()`. So the following is in reference to that function:

- The nil check check on `KindDescription` comes *after* accessing the property in the parent function - the nil check is dead code, it would have panicked before reaching it
- The `FieldNames()` interface func call is made after the concrete type is already known at complie time thanks to the attempted cast to `*OrderedIndexDescription` and early return in the parent function (unless it was an empty slice, in which case the readers time is wasted).
- It pretends to copy the slice twice, but in reality the second copy is dead code, as if it contained any elements, this inner function would never have been called
- From the parent func, the slice might be copied, or a pointer to the underlying might be returned, depending.

EDIT: I found a second call, but still, please rework this, it is pretty bad.

client/index.go
AndrewSisley: I see why :) Resolving this thread

AndrewSisley: No I don't - reopening, sorry I thought this was something else.

internal/db/collection_index.go
```diff
@@ -493,6 +493,56 @@ func (c *collection) NewIndex(
...
+//
+// A descriptor read from the store always carries its kind config, but one built in memory by a
+// caller that predates the kind carrying fields does not, so fall back to the deprecated field.
+func indexFieldNames(desc client.IndexDescription) []string {
+	if desc.KindDescription != nil {
+		if names := desc.KindDescription.FieldNames(); len(names) > 0 {
+			return names
+		}
+	}
+	//nolint:staticcheck // the fallback this helper exists to hide
```
AndrewSisley, line 506: todo: It looks like you forgot to finish writing this comment - please tweak it a bit, I do not know what it is trying to communicate (same goes for the copied one 2 lines below).

internal/db/collection_index.go
```diff
@@ -493,6 +493,56 @@ func (c *collection) NewIndex(
...
+		return desc.Ordered.Fields
+	}
+	if desc.Vector != nil && len(desc.Vector.Fields) > 0 {
+		fields := make([]client.IndexedFieldDescription, len(desc.Vector.Fields))
+		for i, name := range desc.Vector.Fields {
+			fields[i] = client.IndexedFieldDescription{Name: name}
+		}
+		return fields
+	}
+	//nolint:staticcheck // the deprecated field is still supported until v2.0.0
```
AndrewSisley, line 542: todo: Please remove this comment, it adds no new information to the reader.

Please also document somewhere within this function why the if-else[implicit] on `desc.Vector` exists.

internal/db/collection_index.go
```diff
@@ -493,6 +493,56 @@ func (c *collection) NewIndex(
 	return indexDesc, nil
 }
 
+// indexFieldNames returns the names of the index's fields.
+//
+// A descriptor read from the store always carries its kind config, but one built in memory by a
+// caller that predates the kind carrying fields does not, so fall back to the deprecated field.
+func indexFieldNames(desc client.IndexDescription) []string {
```
AndrewSisley, line 500: todo: This looks like a function that exists in `client`, please delete this and use that, if that function survives.

internal/db/collection_index.go
```diff
@@ -543,16 +597,24 @@ func processNewIndexRequest(
...
 	kind := client.IndexKindOrdered
-	var kindDescription client.IndexKindDescription = &client.OrderedIndexDescription{Unique: unique}
+	var kindDescription client.IndexKindDescription = &client.OrderedIndexDescription{
+		Unique: unique,
+		Fields: fields,
+	}
 	if desc.Vector != nil {
 		kind = client.IndexKindVector
-		kindDescription = desc.Vector
+		vector := *desc.Vector
```
AndrewSisley, line 606: todo: Please document why you are doing this, and why you are doing this in this way

internal/planner/index_helpers.go
```diff
@@ -37,12 +37,30 @@ func queryableIndexes(col client.Collection) []client.IndexDescription {
 	return col.Version().Indexes
 }
 
+// firstIndexedField returns the name of the index's first field, or "" if it has none.
+//
+// A descriptor read from the store always carries its kind config, but one built in memory by a
+// caller that predates the kind carrying fields does not, so fall back to the deprecated field.
+func firstIndexedField(idx client.IndexDescription) string {
+	if idx.KindDescription != nil {
```
AndrewSisley, line 45: todo: Please document under what circumstances that this can be nil here.

tests/action/new_index.go
```diff
@@ -116,18 +116,22 @@ func (a *NewIndex) Execute() {
 		}
 
 		if a.FieldName != "" {
+			//nolint:staticcheck // the action exposes both spellings so tests can cover each
```
AndrewSisley, line 119: question: What do you mean by spelling here?

tests/action/results.go
```diff
@@ -30,13 +30,14 @@ import (
...
-// test expectation written in either style (only Unique, or only Kind) compares against the actual.
-func normalizeIndexes(indexes []client.IndexDescription) []client.IndexDescription {
-	out := make([]client.IndexDescription, len(indexes))
-	for i, idx := range indexes {
-		out[i] = idx.Normalize()
-	}
+// normalizeIndexes round-trips each index through JSON, which is what makes the deprecated and
+// current spellings consistent. A test expectation written in either style then compares equal to
+// the actual.
+func normalizeIndexes(t testing.TB, indexes []client.IndexDescription) []client.IndexDescription {
```
AndrewSisley, line 36: suggestion: Remove this function, it adds more confusion than clarity to the code.

AndrewSisley (changes requested):
I think this needs some tweaks before it can be merged, it feels a little rushed.

internal/db/collection_index.go
```diff
@@ -493,6 +493,56 @@ func (c *collection) NewIndex(
...
+	return names
+}
+
+// indexFields returns the index's fields with their directions. Only an ordered index has them, so a
+// vector index's entries come back ascending. Prefer indexFieldNames where only names are used.
+func indexFields(desc client.IndexDescription) []client.IndexedFieldDescription {
+	if ordered, ok := desc.KindDescription.(*client.OrderedIndexDescription); ok && len(ordered.Fields) > 0 {
+		return ordered.Fields
+	}
+	names := indexFieldNames(desc)
```
shahzadlone, line 521: question: Could the legacy fallback preserve each field's Descending flag? it looks atm that internal/db/collection_index.go:521-525 reduces a fields only descriptor to names and reconstructs ascending fields, while the index write path uses that direction. couldn't a descending descriptor therefore write ascending key, perhaps can centralize this through the complete descriptor resolver and add a legacy descending write/read test?

cbindings/wrapper_collection.go
```diff
@@ -92,8 +92,10 @@ func (c *Collection) NewIndex(
 	copts.name = cName
 	copts.getInactive = 0
 
-	orderedFields := make([]string, len(indexDesc.Fields))
-	for i, f := range indexDesc.Fields {
+	//nolint:staticcheck // a request has no accessor; the deprecated field is the only source
```
shahzadlone, line 95: question/suggestion: Could both(C and CLI) adapters resolve the canonical Ordered.Fields (and vector fields) before serialization? I was looking at cbindings/wrapper_collection.go:95-105 and tests/clients/cli/wrapper_collection.go:99-119 send only deprecated Fields, so a config-only ordered request crosses the boundary empty and loses direction. what if we add ascending/descending C and CLI adapter tests? (unless i missed it and there are some already)

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+// GetFields returns the index's fields with their directions, reading the kind's own config and
+// falling back to the deprecated [IndexDescription.Fields]. Only an ordered index has directions, so
+// a vector index's entries come back ascending.
+func (d IndexDescription) GetFields() []IndexedFieldDescription {
+	if ordered, ok := d.KindDescription.(*OrderedIndexDescription); ok && len(ordered.Fields) > 0 {
+		return ordered.Fields
+	}
+	// The deprecated field is the only one carrying directions for a descriptor built in memory, so
+	// it is returned whole rather than reduced to names.
+	if len(d.Fields) > 0 {
```
shahzadlone, line 224: question: Can field resolution dispatch by Kind and choose one authoritative vector representation?

ATM client/index.go:218-279 shows GetFields can prefer deprecated top-level fields while fieldNames/planner paths prefer Vector.Fields, leaving conflicting descriptors unsynchronized.

todo: If my hunch above is correct perhaps consolodate by a resolver and add a conflicting vector marshal/accessor test.

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+	// caller's descriptor.
+	switch config := d.KindDescription.(type) {
+	case *OrderedIndexDescription:
+		if len(config.Fields) == 0 {
+			upgraded := *config
+			upgraded.Fields = d.Fields
+			d.KindDescription = &upgraded
+		}
+	case *VectorIndexDescription:
+		if len(config.Fields) == 0 {
```
shahzadlone, line 268: suggestion: Could typed nil kind configurations be treated as absent instead of dereferenced? i think new resolver can panic on a public IndexDescription value?

internal/db/collection_define.go
```diff
@@ -929,7 +929,8 @@ func findIndexWithFirstField(
 	fieldName string,
 ) (isUnique bool, found bool) {
 	for _, index := range newIndexes {
-		if len(index.Fields) > 0 && index.Fields[0].Name == fieldName {
+		//nolint:staticcheck // a request has no accessor; the deprecated field is the only source
+		if fields := index.Fields; len(fields) > 0 && fields[0].Name == fieldName {
```
shahzadlone, line 933: I think I agree with the code rabbit comment if i understand correctly.

suggestion: Can this relation check resolve Ordered.Fields before deciding whether an index already exists? atm canonical ordered requests prolly appear absent to findIndexWithFirstField.

client/index.go
```diff
@@ -109,14 +112,31 @@ const (
 type OrderedIndexDescription struct {
 	// Unique indicates whether the index enforces uniqueness.
 	Unique bool
+
+	// Fields are the indexed fields, in key order. Each carries a direction, since an ordered index
+	// is read in key order.
+	Fields []IndexedFieldDescription
```
shahzadlone, line 118: todo/question: Can we avoid breaking existing unkeyed literals, new `Fields` here and at my other comment in this same file change positional "layout" for downstream code, no?.

If keyed literals are now required, we need to document properly according to our release breaking rules.

shahzadlone (changes requested):

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+	}
+	return fields
+}
+
+// fieldNames returns the names of the index's fields, falling back to the deprecated top-level ones
+// for a descriptor built in memory before the kind carried them.
+func (d IndexDescription) fieldNames() []string {
+	if d.KindDescription != nil {
+		if names := d.KindDescription.FieldNames(); len(names) > 0 {
+			return names
```
shahzadlone: nitpick: Please add a test to hit this if not tedious to do so.

client/index.go
```diff
@@ -184,15 +212,71 @@ func (d IndexDescription) GetUnique() bool {
...
+	}
+	// The deprecated field is the only one carrying directions for a descriptor built in memory, so
+	// it is returned whole rather than reduced to names.
+	if len(d.Fields) > 0 {
+		return d.Fields
+	}
+	names := d.fieldNames()
+	fields := make([]IndexedFieldDescription, len(names))
+	for i, name := range names {
+		fields[i] = IndexedFieldDescription{Name: name}
```
shahzadlone: nitpick: Please add a test to hit this if not tedious to do so.

client/index.go
islamaliev: This one encapsulated some logic that otherwise would need to land in the packages themselves that use it. I wasn't myself sure if it should go there. I know it's a new public surface. But yes, you're right. I think we should remove it.

client/index.go
islamaliev: Yeah, I will also deprecate Normalize. And it should be gone once as the other deprecated stuff is gone.

client/index.go
islamaliev: refactored

internal/db/collection_index.go
islamaliev: done

internal/db/collection_index.go
islamaliev: after refactor this function is gone

internal/db/collection_index.go
islamaliev: this function in `client` is private.

internal/db/collection_index.go
islamaliev: done

internal/planner/index_helpers.go
islamaliev: refactored. Now it's simpler

tests/action/new_index.go
islamaliev: rephrased

tests/action/results.go
islamaliev: removed

client/index.go
islamaliev: fixed

cbindings/wrapper_collection.go
islamaliev: fixed

internal/db/collection_define.go
islamaliev: The duplicate helpers are gone

client/index.go
islamaliev: the old code dropped the direction and rebuilt everything ascending. Now, it keeps the fields as they are now.
Test already exists

client/index.go
islamaliev: added a note to breaklog

client/index.go
```diff
@@ -184,15 +195,70 @@ func (d IndexDescription) GetUnique() bool {
...
+	case *VectorIndexDescription:
+		if config != nil && len(config.Fields) > 0 {
+			fields := make([]IndexedFieldDescription, len(config.Fields))
+			for i, name := range config.Fields {
+				fields[i] = IndexedFieldDescription{Name: name}
+			}
+			return fields
+		}
+	}
+	return d.Fields
```
AndrewSisley: todo: The comment in the C bindings states that `d.Fields` should be used over the new properties when it is set, yet this function appears to do the opposite and instead falls back to the legacy `d.Fields` only if `d.KindDescription` is either nil or invalid.

Something looks like it needs to change.

If you correct this function to match the rules stated in the C bindings you should be able to eliminate the code branching/logic from there completely and just call this (corrected) function.

You could also instead update the code-comment and code in the C bindings so that matches what the Go client appears to think is correct. 

suggestion:  The switch does not differentiate between `nil` and an invalid type (I think, maybe nil panics?) - I suggest erroring if it is an invalid type.

client/index.go
```diff
@@ -184,15 +195,70 @@ func (d IndexDescription) GetUnique() bool {
...
+		names[i] = field.Name
+	}
+	return names
+}
+
+// Normalize returns a copy with the kind config and the deprecated fields consistent, so a
+// descriptor is the same whichever one was set. Both (un)marshalling paths run it.
+//
+// Deprecated: this only reconciles the deprecated Fields and Unique, so it will be removed along
+// with them in Defra v2.0.0.
```
AndrewSisley: praise: This is very well documented, thank you very much!

client/index.go
```diff
@@ -271,6 +337,9 @@ type NewIndexRequest struct {
 	// Name contains the name of the index.
 	Name string
 	// Fields contains the fields that are being indexed.
+	//
+	// Deprecated: set them on [NewIndexRequest.Ordered] or [NewIndexRequest.Vector]. This field will
+	// be removed in Defra v2.0.0. Setting both is an error unless they agree.
```
AndrewSisley: thought: Related to an earlier comment - `Setting both is an error unless they agree.` appears to differ in a new way from both `client.IndexDescription.GetFields` and the logic and comment within the C bindings.

internal/db/collection_index.go
AndrewSisley: Cheers, and I completely agree with splitting it like you have, thanks Islam.

AndrewSisley (approved):
Looks great to me, thank you very much for the clean-up.  Please do resolve the new todo before merge though, things at the moment don't quite seem to pair up.

internal/db/collection_index.go
islamaliev: the old code dropped the direction and rebuilt everything ascending. Now, it keeps the fields as they are now.
Test already exists

PR #5243, opened 2026-09-05
fix: Use correct constant in parse kind error

### LINKED ISSUE #5238: NewErrFailedToParseKind uses wrong error message constant
## Problem

In `client/errors.go:259-264`:

```go
func NewErrFailedToParseKind(kind []byte) error {
    return errors.New(
        errCRDTKindMismatch,
        errors.NewKV("Kind", kind),
    )
}
```

It uses `errCRDTKindMismatch = "CRDT type %s can't be assigned to field kind %s"` instead of `errFailedToParseKind = "failed to parse kind"`.

Consequences (verified by reading `errors/errors.go:48` — `errors.New` does no formatting):
- The returned message contains a literal `%s` instead of formatted values.
- `errors.Is(err, ErrFailedToParseKind)` (defined at `client/errors.go:87` from `errFailedToParseKind`) will never match errors built by this constructor.

Compare with `NewErrCRDTKindMismatch` at line 251-253 which correctly does `fmt.Sprintf(errCRDTKindMismatch, ...)`.

## Expected

Constructor should use `errFailedToParseKind` so the message and sentinel match.

## Location

- `client/errors.go:38`, `:87`, `:259-264`

Verified on current `develop`. No existing issue covers `FailedToParseKind`.

Ayush7614:
## Relevant issue(s)

Resolves #5238

## Description

`NewErrFailedToParseKind` in `client/errors.go:259-264` used `errCRDTKindMismatch` (which contains `%s` verbs) instead of `errFailedToParseKind`. Since `errors.New` does no formatting, the message contained a literal `%s` and `errors.Is(err, ErrFailedToParseKind)` could never match.

Changes the constructor to use `errFailedToParseKind`, matching the sentinel at `client/errors.go:87`.

Verified with `go test ./client/...` passing.

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -258,7 +258,7 @@ func NewErrInvalidJSONPayload(payload any) error {
 
 func NewErrFailedToParseKind(kind []byte) error {
 	return errors.New(
-		errCRDTKindMismatch,
+		errFailedToParseKind,
 		errors.NewKV("Kind", kind),
 	)
 }


ChrisBQu (approved):
LGTM, thank you.

PR #5251, opened 2026-09-08
test(i): Let an external test node run under NAC

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/5170

## Description

A test node that runs as a separate process can now be asked to turn node access control on. The request used to be dropped without a word, so tests asserting that an unauthorized caller is refused kept passing against a node that was never enforcing anything.

Two things stopped it working: the node was never told to enable access control, and once it was, the tokens it was sent were minted for the address it had before it restarted, which it rejects.

No test is un-excluded yet. A node in another process still generates its own node identity, which the harness relies on elsewhere to get past access control, so the tests excluded for this need that solved first.

Two guards come with it, since a silently dropped setting is what made the failure invisible. Setup now confirms the node came up with access control on rather than assuming the request landed, and a test fails if a node setting is added without saying whether a separate process honours it or ignores it.

## Tasks

## How has this been tested?

Unit tests, plus the networked suite under both cross-version directions. Reverting each fix in turn was used to confirm the tests fail without it.

- macOS

### DIFF
--- a/tests/action/node_options.go
+++ b/tests/action/node_options.go
@@ -57,6 +57,10 @@ type NodeSetupConfig struct {
 	// instantiated. A zero pool size leaves the node default in place.
 	LensRuntime  options.NodeLensRuntimeType
 	LensPoolSize int
+	// EnableNAC turns on node access control, owned by NACOwner. While it is on the
+	// node lets through no one else.
+	EnableNAC bool
+	NACOwner  immutable.Option[state.Identity]
 }
 
 func applyHTTPOptions(opts *options.NodeOptionsBuilder, httpOpts options.NodeHTTPOptions) {
--- a/tests/action/node_options.go
+++ b/tests/action/node_options.go
@@ -57,10 +57,18 @@ type NodeSetupConfig struct {
 	// instantiated. A zero pool size leaves the node default in place.
 	LensRuntime  options.NodeLensRuntimeType
 	LensPoolSize int
-	// EnableNAC turns on node access control, owned by NACOwner. While it is on the
-	// node lets through no one else.
-	EnableNAC bool
-	NACOwner  immutable.Option[state.Identity]
+	// NodeACP overrides the node's access control settings when set. While access
+	// control is on the node lets through no one but NACOwner.
+	NodeACP immutable.Option[options.NodeACPOptions]
+	// NACOwner is the identity access control is set up under. The node takes it
+	// from the identity it is started with, so this is a reference to that one
+	// rather than a setting of its own.
+	NACOwner immutable.Option[state.Identity]
+}
+
+// nacEnabled reports whether the test asked for node access control.
+func nacEnabled(cfg NodeSetupConfig) bool {
+	return cfg.NodeACP.HasValue() && cfg.NodeACP.Value().IsEnabled
 }
 
 func applyHTTPOptions(opts *options.NodeOptionsBuilder, httpOpts options.NodeHTTPOptions) {
--- a/tests/action/node_options.go
+++ b/tests/action/node_options.go
@@ -57,13 +57,9 @@ type NodeSetupConfig struct {
 	// instantiated. A zero pool size leaves the node default in place.
 	LensRuntime  options.NodeLensRuntimeType
 	LensPoolSize int
-	// NodeACP overrides the node's access control settings when set. While access
-	// control is on the node lets through no one but NACOwner.
+	// NodeACP overrides the node's access control settings when set. While it is on
+	// the node lets through no one but the identity it was started with.
 	NodeACP immutable.Option[options.NodeACPOptions]
-	// NACOwner is the identity access control is set up under. The node takes it
-	// from the identity it is started with, so this is a reference to that one
-	// rather than a setting of its own.
-	NACOwner immutable.Option[state.Identity]
 }
 
 // nacEnabled reports whether the test asked for node access control.
--- a/tests/action/node_setup.go
+++ b/tests/action/node_setup.go
@@ -303,7 +303,7 @@ func setupExternalNode(
 	defer func() { s.CurrentSetupHost = "" }()
 
 	nacIdentity := immutable.None[state.Identity]()
-	if cfg.EnableNAC {
+	if nacEnabled(cfg) {
 		// The node ignores a flag it does not know, so a test could assert an access
 		// rule against a node that never enforced one.
 		require.Contains(s.T, w.StartupLog(), nacEnabledLog,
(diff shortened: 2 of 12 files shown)


tests/action/node_options.go
```diff
@@ -57,6 +57,10 @@ type NodeSetupConfig struct {
 	// instantiated. A zero pool size leaves the node default in place.
 	LensRuntime  options.NodeLensRuntimeType
 	LensPoolSize int
+	// EnableNAC turns on node access control, owned by NACOwner. While it is on the
+	// node lets through no one else.
+	EnableNAC bool
+	NACOwner  immutable.Option[state.Identity]
```
AndrewSisley: thought: I think adding in options one by one like this, with custom props that do not necessarily pair up with their production counterparts is a mistake that we need to correct, and by continuing to use this pattern you are adding technical debt here that we are going to have to rewrite.

AndrewSisley (approved):
LGTM

tests/action/node_options.go
islamaliev: Good point, Andy. I fixed both: the access control one now uses the same type production uses, and the owner one is gone, since the node already gets that identity passed in.

ChrisBQu (approved):
LGTM.

suggestion: I was looking into the issue raised by Coderabbit. In the case of native nodes, NAC status/owner is recovered from the disk (see: `initializeNodeACP` inside `internal/db/db_nac.go`). However, this would not be true of external nodes. However, as you'd know, because it's your issue... cross-version tests _can't_ restart an external node (see: https://github.com/sourcenetwork/defradb/issues/5170)

Should this be made future proof? Maybe. We could use the same pattern we use to cache `P2POpts` and `DisableP2P` to survive restart. I don't think there's a downside to this.

tests/action/node_setup.go
```diff
@@ -279,18 +297,50 @@ func setupExternalNode(s *state.State, cfg NodeSetupConfig, ver string) (*state.
...
+	// Setup calls this node before it is on Nodes, so its tokens can only be minted
+	// from here.
+	s.CurrentSetupHost = w.Host()
+	defer func() { s.CurrentSetupHost = "" }()
+
+	nacIdentity := immutable.None[state.Identity]()
+	if nacEnabled(cfg) {
+		// The node ignores a flag it does not know, so a test could assert an access
+		// rule against a node that never enforced one.
+		require.Contains(s.T, w.StartupLog(), nacEnabledLog,
```
shahzadlone: todo: Could startup confirmation wait for the log marker or an authenticated NAC status result before asserting?

atm i think `NewWrapper` does not synchronize `streamLogs`, and the `require.Contains` assertion can invoke `FailNow` before the wrapper is attached or closed. 

just thinking maybe `StartupLog()` read is therefore racy.

tests/action/node_setup_test.go
```diff
@@ -0,0 +1,113 @@
...
+// TestExternalNodeFlags_ReadsEveryFlaggedField fails if a field recorded as
+// flagged stops being read, which would drop it silently.
+func TestExternalNodeFlags_ReadsEveryFlaggedField(t *testing.T) {
+	source := readNodeSetupSource(t)
+
+	for name, handling := range setupFields {
+		if handling != fieldFlagged {
+			continue
+		}
+		assert.True(t, strings.Contains(source, "cfg."+name),
```
shahzadlone: suggestion: Could `TestExternalNodeFlags_ReadsEveryFlaggedField` assert each `cfg.<field>` configuration field and the `node_options.go` path's actual flags or unsupported result instead of source text? the way it is i think arbitrary comments or dead code can satisfy the current `strings.Contains` check. can probably satisfy it while flags are still dropped. perhaps can replace this with table driven flag/unsupported output assertions, including enabled/disabled/missing/non-secp256k1 NAC identities (if we dont have those tested also).

tests/integration/utils.go
```diff
@@ -1074,11 +1074,14 @@ func startNodes(s *state.State, testCase TestCase, start Start) {
 			opts.DB().SetNodeIdentity(state.GetIdentity(s, NodeIdentity(s.CurrentSetupNodeID)))
 			opts.P2P().SetAll(p2pOpts)
 			opts.SetDisableP2P(s.Nodes[nodeID].DisableP2P)
-			opts.NodeACP().SetEnabled(start.EnableNAC)
+			nacOpts := options.NodeACPOptions{IsEnabled: start.EnableNAC}
+			opts.NodeACP().SetAll(nacOpts)
+			setupConfig := testCase.nodeSetupConfig()
+			setupConfig.NodeACP = immutable.Some(nacOpts)
```
shahzadlone, line 1080: question: Just to make sure we want to not have `restartNodes` preserve the NAC enablement and owner identity between restarts and want to rebuild `NodeACPOptions` in `startNodes` and calling `startNodes` from empty `Start{}` and dropping `setupConfig.NodeACP`, `--node-acp-enable`, and `--identity`?

shahzadlone (approved):
Giving LGTM, just some minor todo(s)/suggestion(s)/question(s)

thought: I just wonder atm tests cover only audience parsing and source text while cross-version cases we are lacking in perhaps we can add later some coverage that invokes externalNodeFlags and proves owner success, unauthenticated denial, and the post-restart audience (basically `externalNodeFlags` + authenticated `PeerInfo` setup end to end)

tests/action/node_setup.go
islamaliev: Good catch, you're right. It now asks the node for its NAC status instead of reading the log, so there's nothing to race.

tests/action/node_setup_test.go
islamaliev: Swapped it for table driven tests on the actual flags

tests/integration/utils.go
islamaliev: Yes, intentional. Passing false doesn't turn NAC off, it just means the start command isn't asking to enable it. Once NAC is set up the node recovers it from disk and ignores the flag, otherwise anyone who can restart the node could switch access control off.

Let me know if you disagree.

islamaliev:
> LGTM.
> 
> suggestion: I was looking into the issue raised by Coderabbit. In the case of native nodes, NAC status/owner is recovered from the disk (see: `initializeNodeACP` inside `internal/db/db_nac.go`). However, this would not be true of external nodes. However, as you'd know, because it's your issue... cross-version tests _can't_ restart an external node (see: #5170)
> 
> Should this be made future proof? Maybe. We could use the same pattern we use to cache `P2POpts` and `DisableP2P` to survive restart. I don't think there's a downside to this.

Agreed external nodes have nothing to recover from, but they can't restart yet anyway, so I'd rather do the caching as part of #5170 where that actually gets built.

Issue #5252, opened 2026-09-08
Allow indexing wider numeric arrays with an explicit precision opt-in


islamaliev:
Part of https://github.com/sourcenetwork/defradb/issues/5070.

Similarity queries work on `[Float64!]`, `[Float32!]` and `[Int!]`, but a vector index only accepts `[Float32!]`. Anyone querying similarity on a field probably wants an index on it, so a type you can query but never index is a trap that shows up only as unexplained slowness.

Wider types are rejected rather than narrowed, because the HNSW graph stores `float32` and narrowing silently would lose precision the field had. The rejection now says so, which closes the confusing half of the problem, but it leaves `[Float64!]` unindexable.

## What to do

Let the user opt into the narrowing, so the loss of precision is their explicit choice rather than something DefraDB does behind their back.

```graphql
type Users {
    embedding: [Float64!] @index(vector: {dimensions: 3, precision: FLOAT32})
}
```

Omitting `precision` on a `[Float64!]` field keeps rejecting the index, so nothing changes for anyone who has not asked for it. This is the same shape as pgvector, which cannot index a `double precision[]` column directly either and requires an expression index with an explicit cast, documented as indexing "at a lower precision".

`precision` is also the natural home for the quantization work that would follow: float16 and int8 are the same trade, just further along.

## Int arrays

`[Int!]` should stay unindexable, and not for a reason `precision` fixes. The distance maths is fine on integers, but float32 holds integers exactly only up to ~16.7 million, so larger values would silently collide and rank wrong. An int array in practice holds IDs or counts rather than embeddings, so there is nothing to gain by widening this.

That leaves `[Int!]` queryable but not indexable, which is still a mismatch, but a deliberate one that can be explained rather than the current accidental one.

## Testing

Tests covering an index created on a `[Float64!]` field with the opt-in, that querying it returns the same ordering as the equivalent float32 field, that omitting the opt-in still errors, and that `[Int!]` still errors either way.

AndrewSisley:
Why are you planning on narrowing them, instead of storing more precise values?

PR #5253, opened 2026-09-08
fix: Explain what SIMILARITY and vector indexes accept

### LINKED ISSUE #5249: Introspection info for `SIMILARITY` is lacking
The docs surfaced via graphql clients are rich for most functions, but not so much for `SIMILARITY`. 

Image
Image

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5249
Refs https://github.com/sourcenetwork/defradb/issues/5252

## Description

Similarity search now explains itself in the three places it stayed quiet. Exploring the API in a GraphQL client showed the similarity function and its arguments with no documentation beside them, and what documentation there was still described the result as cosine similarity, from before the other metrics shipped. It now says what the function returns, that the metric follows whichever one the field was indexed with, and that higher is nearer for all of them.

Pointing similarity at a field that cannot hold a vector reported an unknown argument, which read as though the field itself did not exist. It now names the field and its type, and says one holding numbers is what is needed.

Asking for a vector index on a field that is not a 32 bit float array said only that the type was unsupported, leaving the reader to guess what is. Wider numbers stay rejected rather than quietly narrowed to what the index stores, since that would lose precision the field had; letting a user opt into that trade is https://github.com/sourcenetwork/defradb/issues/5252.

## Tasks

## How has this been tested?

Integration tests. Two existing tests changed: they pinned the old wording for errors this reword replaces.

- macOS


AndrewSisley (approved):
LGTM

PR #5260, opened 2026-09-09
fix: Format IndexKind as text string instead of numeric ID

### LINKED ISSUE #5229: Show index type as text rather than numeric ID
When creating or showing an index, the output for the `Kind` field is a numeric ID of the type. As we did for collection fields, it's a nice UX improvement to display the corresponding string value there.

```
stef@carbonino:~$ defradb client collection add '
type Book11 {
  title: String
  plot: String @index(kind:ordered)
  about_v: [Float32!] @index(vector: {dimensions: 3, hnsw: {metric: COSINE}})  @embedding(
    fields: ["title", "plot"], 
    provider: "ollama", 
    model: "embeddinggemma"
  )
}
'
[
  {
    "Name": "Book11",
...
    "Indexes": [
      {
        "Name": "Book11_plot_ASC",
        "ID": 1,
        "Fields": [
          {
            "Name": "plot",
            "Descending": false
          }
        ],
        "Kind": 0,
        "KindDescription": {
          "Unique": false
        },
        "Unique": false
      },
      {
        "Name": "Book11_about_v_ASC",
        "ID": 2,
        "Fields": [
          {
            "Name": "about_v",
            "Descending": false
          }
        ],
        "Kind": 1,
        "KindDescription": {
          "Algorithm": "HNSW",
          "Metric": "COSINE",
          "Dimensions": 3,
          "HNSW": {
            "M": 16,
            "EfConstruction": 128,
            "EfSearch": 64
          }
        },
        "Unique": false
      }
    ],
```

Tyagiquamar:
Fixes #5229

### DIFF
--- a/client/index.go
+++ b/client/index.go
@@ -50,7 +50,12 @@ func (k IndexKind) String() string {
 }
 
 func (k IndexKind) MarshalText() ([]byte, error) {
-	return []byte(k.String()), nil
+	switch k {
+	case IndexKindOrdered, IndexKindVector:
+		return []byte(k.String()), nil
+	default:
+		return nil, NewErrUnknownIndexKind(uint8(k))
+	}
 }
 
 func (k *IndexKind) UnmarshalText(text []byte) error {
--- a/client/index.go
+++ b/client/index.go
@@ -38,6 +38,34 @@ const (
 	IndexKindVector
 )
 
+func (k IndexKind) String() string {
+	switch k {
+	case IndexKindOrdered:
+		return "ordered"
+	case IndexKindVector:
+		return "vector"
+	default:
+		return "unknown"
+	}
+}
+
+func (k IndexKind) MarshalText() ([]byte, error) {
+	return []byte(k.String()), nil
+}
+
+func (k *IndexKind) UnmarshalText(text []byte) error {
+	switch string(text) {
+	case "ordered", "b-tree", "0":
+		*k = IndexKindOrdered
+		return nil
+	case "vector", "1":
+		*k = IndexKindVector
+		return nil
+	default:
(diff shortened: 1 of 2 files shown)


Tyagiquamar:
Hi, friendly ping for review on this PR when you have a moment. Happy to address any feedback. Thanks!

client/index.go
```diff
@@ -38,6 +38,39 @@ const (
...
+func (k *IndexKind) UnmarshalText(text []byte) error {
+	switch string(text) {
+	case "ordered", "b-tree", "0":
+		*k = IndexKindOrdered
+		return nil
+	case "vector", "1":
+		*k = IndexKindVector
+		return nil
+	default:
+		return NewErrUnknownIndexKind(0)
```
ChrisBQu, line 70: Thank you for this PR. It mostly looks good to me. This is the one change I am going to request though.

todo: This passes in a value of 0 to `NewErrUnknownIndexKind`, but 0 is not an appropriate value here. 0 is the valuee of `IndexKindOrdered` as per its definition earlier in this file. What would really be best, is if this error function took a text parameter instead. I understand that this function is called in only one other place, also in this file, so making the adjustment should only increase the scope of this PR a tiny bit. But it would improve it greatly.

ChrisBQu (changes requested):

client/index.go
Tyagiquamar: Updated NewErrUnknownIndexKind to accept a string parameter instead of uint8. Call sites now pass string(text) in UnmarshalText, and string representations in MarshalText and UnmarshalJSON. Also updated client/index_test.go to verify that the unknown kind text is included in error outputs. All tests pass cleanly.

PR #5262, opened 2026-09-10
test(i): Give an external test node a stable identity

### LINKED ISSUE #5170: Cross-version tests cannot restart an external node

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5170

## Description

Cross-version tests can now run against a node in a separate process while NAC is on.

A node in another process used to generate its own identity at startup, so it was a stranger to the test that started it and access control turned every request away. It is now started as the identity the test already addresses it by, the same one a node running in-process is given.

## Tasks

## How has this been tested?

Unit tests, plus the networked and access control suites under both cross-version directions. 

- macOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,91 @@
 
+<a name="v1.1.0"></a>
+## [v1.1.0](https://github.com/sourcenetwork/defradb/compare/v1.0.0...v1.1.0)
+
+> 2026-09-08
+
+### Features
+
+* Add extensions to GraphQL responses ([#5189](https://github.com/sourcenetwork/defradb/issues/5189))
+* Add index type to C bindings NewEncryptedIndex ([#5131](https://github.com/sourcenetwork/defradb/issues/5131))
+* Add filtered collection truncate ([#5097](https://github.com/sourcenetwork/defradb/issues/5097))
+* Vector distance metrics L2 and dot product ([#5169](https://github.com/sourcenetwork/defradb/issues/5169))
+* Add HTTP client options ([#5156](https://github.com/sourcenetwork/defradb/issues/5156))
+* HNSW vector index for embedding fields ([#5096](https://github.com/sourcenetwork/defradb/issues/5096))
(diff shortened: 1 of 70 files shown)


ChrisBQu (approved):
LGTM.

The diff is funny, because it's mostly just being able to remove excluded multipliers. That's awesome.

PR #5264, opened 2026-09-10
refactor: GraphQL package

### LINKED ISSUE #1195: `subscriptionType` of `__schema` is `null`
From the perspective of the schema `null` indicates subscription is not a supported feature.

To reproduce:

```
defradb client  query '
query IntrospectionQuery {
        __schema {
                queryType { name }
                mutationType { name }
                subscriptionType { name }
        }
}
'
```

Reference: https://spec.graphql.org/October2021/#sec-The-__Schema-Type

### LINKED ISSUE #1218: No GQL introspection tests for `commit` type
`_version` is noted in the default fields, but it is a shallow/list type and we currently have no tests that verifies that the `Commit` type actually contains the properties we want it to (e.g. `cid`, `dockey`, `height`, etc)

### LINKED ISSUE #1384: Order of schema fields in introspection is not deterministic
https://user-images.githubusercontent.com/363965/233669237-339ca4b9-caae-4102-a6fc-2dbe3b6cb7b1.mov

relates to https://github.com/sourcenetwork/graphql-go

### LINKED ISSUE #5263: Refactor GraphQL package
We should switch to a more recent and maintained GraphQL library so that there is less burden on the DB team to maintain a forked repo.

At the same time we should switch the schema generation from a code based solution to a templated one. I believe this will improve the DX for the internal DB team as we already design features from a GraphQL standpoint.

nasdf:
## Relevant issue(s)

Resolves #5263

The following are fixed and tested here: `internal/request/graphql/introspection_execution_test.go`

Fixes #1195  `TestExecuteIntrospectionReturnsSubscriptionType`
Fixes #1218 `TestExecuteIntrospectionCommitFields`
Fixes #1384 `TestExecuteIntrospectionFieldOrderIsDeterministic`

## Description

This PR refactors the GraphQL package with a more recent and maintained library. It also converts the schema generation from a code based approach to a template based approach.

## Tasks

## How has this been tested?

Existing integration and unit test coverage plus some new unit tests. 

- Ubuntu Linux


internal/request/graphql/schema/collection.go
```diff
@@ -942,13 +1008,16 @@ func constraintsFromAST(kind client.FieldKind, directive *ast.Directive) (constr
...
-func setCRDTType(field *ast.FieldDefinition, kind client.FieldKind) (client.CType, error) {
+func setCRDTType(field *collectionFieldDefinition, kind client.FieldKind) (client.CType, error) {
 	if directive, exists := findDirective(field, "crdt"); exists {
 		for _, arg := range directive.Arguments {
 			switch arg.Name.Value {
 			case "type":
+				if stringValue, ok := arg.Value.(*collectionStringValue); ok {
+					return 0, newErrInvalidArgument(arg.Name.Value, stringValue.Value)
+				}
 				cTypeString := arg.Value.GetValue().(string)
```
ChrisBQu: thought: I don't think this is safe. It assumes that the value is a string literal. It will only be a string literal if the query is clean though, right? It should return an error instead of panicking if the type is bad. For example, I think this would panic:

`@crdt(type: true)`

internal/request/graphql/schema/collection.go
```diff
@@ -892,29 +958,29 @@ func parseUint32ASTValue(value ast.Value) (uint32, error) {
...
+	fieldDef *collectionFieldDefinition,
 ) (client.VectorEmbeddingDescription, error) {
 	embedding := client.VectorEmbeddingDescription{
 		FieldName: fieldDef.Name.Value,
 	}
 	for _, arg := range directive.Arguments {
 		switch arg.Name.Value {
 		case types.VectorEmbeddingDirectivePropFields:
-			val := arg.Value.(*ast.ListValue)
+			val := arg.Value.(*collectionListValue)
```
ChrisBQu: thought: I think this type assertion is unsafe. A non-string element inside fields will panic instead of return an error, right?

internal/request/graphql/schema/manager.go
```diff
@@ -11,170 +11,73 @@
...
-		Body: []byte(sdl),
-	})
-	doc, err := gqlp.Parse(gqlp.ParseParams{
-		Source: src,
-	})
-	if err != nil {
-		return nil, err
+	document, report := astparser.ParseGraphqlDocumentString(sdl)
+	if report.HasErrors() {
+		return nil, report
```
ChrisBQu, line 87: thought: This function concerns me. If you look at the previous one, it validates thee document. This one doesn't. I think this will silently continue without error if a directive is not supported. For example:

`type User { name: String @notrealdirective }`

internal/request/graphql/parser/coercion.go
```diff
@@ -0,0 +1,176 @@
...
+				return utcNowValue{}
+			}
+			if parsed, err := time.Parse(time.RFC3339Nano, value); err == nil {
+				return parsed
+			}
+		}
+	case "JSON":
+		return coerceJSONNumbers(value)
+	}
+	return value
```
ChrisBQu, line 222: question: Is Blob meant to be here as well?

internal/request/graphql/parser/coercion.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return result
+}
+
+func coerceScalar(name string, value any) any {
+	switch name {
+	case "Int":
+		switch value := value.(type) {
+		case float64:
+			return int32(value)
```
ChrisBQu, line 199: thought: This seems like a bug. Surely we cannot call int32() on a value that is float64 without a range check.

internal/request/graphql/parser/coercion.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	case "DateTime":
+		if value, ok := value.(string); ok {
+			if value == "UTC_NOW" {
+				return utcNowValue{}
+			}
+			if parsed, err := time.Parse(time.RFC3339Nano, value); err == nil {
+				return parsed
+			}
+		}
```
ChrisBQu, line 218: thought: I think you can cause a panic here downstream. At minimum, it doesn't seem correct. If `time.Parse` fails, we will fall all the way through and return the original value. So consider the following:

`Users_Create(input: {createdAt: "not-a-datetime"})`

internal/request/graphql/schema/errors.go
```diff
@@ -11,10 +11,54 @@
...
+func (e wrappedSchemaValidationError) Error() string        { return e.message }
+func (e wrappedSchemaValidationError) Unwrap() error        { return e.cause }
+func (e wrappedSchemaValidationError) Is(target error) bool { return target == ErrGraphQLValidation }
+
+func newErrUnknownDirectiveArgument(name, directive string) error {
+	return schemaValidationError(fmt.Sprintf("Unknown argument %q on directive %q", name, directive))
+}
+
+func newErrInvalidArgument(name string, value any) error {
+	return schemaValidationError(fmt.Sprintf("Argument %q has invalid value %v", name, value))
```
ChrisBQu, line 42: todo: %v is incorrect here. We want %q instead, to preserve quotes. See the following test, which is currently broken, but will pass with this change:

`TestCollectionAdd_ContainsPNCounterWithInvalidType_Error`

internal/request/graphql/schema/manager.go
```diff
@@ -11,170 +11,73 @@
...
 }
 
-// NewSchemaManager returns a new instance of a SchemaManager
-// with a new default type map
+// NewSchemaManager returns a manager initialized with the canonical default
+// schema definition.
 func NewSchemaManager(isSearchableEncryptionEnabled bool) (*SchemaManager, error) {
-	schema, err := defaultSchema()
-	if err != nil {
+	sm := &SchemaManager{isSearchableEncryptionEnabled: isSearchableEncryptionEnabled}
```
ChrisBQu, line 39: thought: Trying to do some digging here. 

`NewSchemaManager` calls `setDefinition`

However, `Generate` inside `internal/request/graphql/schema/generate.go` also calls `setDefinition`. This function is called by `SetSchema` inside `internal/request/graphql/parser.go`. That function is called by `loadCollectionDefinitions` inside `internal/db/collection.go`. And finally, where is _that_ called from? It is called from... all sorts of places, such as the `AddSchema`, `PatchSchema`, and `CreateIndex` functions.

What I'm getting at, is that I think the parsing is happening more times than it needs to happen. Can we cache it in some way? Is the waste problematic touugh, or is this not worth it? Unclear.

PR #5266, opened 2026-09-14
fix: Close p2p handler streams and finish the retry sweep

### LINKED ISSUE #5265: Replicator loses documents permanently under load: unclosed streams, nack read as success, and a retry sweep that abandons the set
## Problem

Under sustained mixed-mesh load a Go node permanently loses documents it believes it delivered. Four
defects compound, and each one masks the next. Inbound handlers never close the streams they read
(`go-p2p host.go:220-224`, `comm_channel.go:98-131`), so a peer's reservations leak until the
per-protocol inbound cap is reached — 96 on the host measured — after which every further reply
stream is refused forever. The sender then sees a timeout rather than the receiver's nack.

On timeout, `handleReplicatorFailure` is handed the already-expired context and returns at
`internal/db/p2p/replicator.go:466-468` before writing anything, so no retry record exists and the
document never re-pushes. Even when a reply does arrive, `message.go:174-177` tests the *request's*
`ErrMessage` rather than the reply's, so a backpressure nack is read as success. And the retry sweep
abandons the whole remaining set on a single per-document error, then paces the survivors as one
unthrottled burst that the receiver's rate limiter rejects wholesale.

## Plan

- Close inbound handler streams after the handler returns, so reservations are released.
- Read the reply's `ErrMessage`, so a backpressure nack is recorded as a failure worth retrying.
- Continue the retry sweep past a per-document error instead of abandoning the set.
- Pace the sweep with an at-capacity wait and a first-rung return, so the post-load burst is not
  rate-limited away.
- Ship all four together and verify on a soak arm; each alone is harmful or insufficient.

edjroz:
## What

A replicator under load permanently loses documents it reports as delivered. Inbound handlers leak
stream reservations until a 96-stream cap refuses every further reply; the sender reads that as a
timeout and, with an expired context, writes no retry record at all.

Where a reply does arrive, the code checks the request's error field rather than the reply's, so a
backpressure nack is treated as success. The retry sweep then abandons its whole set on one
per-document error, and paces the remainder as a single burst the receiver rate-limits away.

## What Changed

- Inbound handlers close their stream after returning, releasing the reservation.
- `message.Receive` tests the reply's `ErrMessage` instead of the request's.
- The retry sweep continues past a per-document error rather than abandoning the remaining set.
- The sweep paces its pushes, waits on an at-capacity reply, and returns to the first rung on success.

## Verification

- [x] Go unit tests for stream close and reply-error handling
- [x] Five soak arms, 2r2g / 20 ops/s / 900 s load, each isolating one change
- [x] Documents missing on a Rust node **748 → 0** (final arm, 3600 s settle)
- [x] `final_mismatches` 5,993 → 3, `unresolved` 0, 18,001 ops executed in both
- [x] Go→Go `stream reset by remote, error code: 4098` 21,133 → 0
- [x] Expired-context handler failures 45,354 → 214
- [x] Retry failures from rate limiting ~84% → 0
- [x] Receiver `pending_dag_expired` 50 → 24; `pending_dags` ended at 0, never pinned at cap

## Notes

- **These must ship together.** Closing the streams alone *increased* loss (748 → 844): it removed an
  accidental retry path, because a stream reset left the context live and did write a retry record,
  whereas a cleanly delivered nack was silently swallowed. Pacing without the at-capacity wait and
  first-rung return also regressed (451 → 779).
- A Rust-only mitigation was built and soak-tested against stock Go, and **withdrawn**: it repaired
  reply delivery (Go timeouts 45,381 -> 2,613) yet loss rose 748 -> 1,214, because a delivered nack
  that the sender reads as success is worse than an undelivered one. That is the third independent
  demonstration that the nack change is load-bearing.
- The pacing interval is tied to the receiver's default 50/s limit rather than negotiated. That is a
  known ceiling; the durable fix is for the sender to read a retry hint from the limiter, which the
  nack text does not carry today.
- The zero is n=1 at one seed, and 1,624 sender retry markers remained outstanding at the final
  sample despite every document having converged. Worth understanding before release.

---

Closes #5265

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Qy6KdB84tUcS6DxiVRmnNo


internal/db/p2p/message/message.go
```diff
@@ -91,6 +91,13 @@ type proto interface {
 
 // Receive takes in a network stream and store the unmarshalled message in the provided [Message]
 func Receive(stream io.Reader, peerID string, proto proto, m Message) error {
+	// Inbound handlers never use the stream after this call. libp2p releases
+	// the stream's resource-manager reservation only on a local close, so an
+	// unclosed stream holds a per-(protocol, peer) inbound slot until the
+	// connection drops; the peer's replies are refused once the slots are gone.
+	if closer, ok := stream.(io.Closer); ok {
+		defer closer.Close()
```
fredcarle: question: Are we sure it's ok to ignore the `closer.Close` returned error? 

todo: If so, please document it and explicitly discard it. If not, please `join` the error on return.

internal/db/p2p/replicator.go
```diff
@@ -767,24 +770,66 @@ func (p *P2P) retryReplicator(ctx context.Context, peerID string) {
...
+		}
 		if err != nil {
 			log.ErrorContextE(ctx, "Failed to retry doc", err, corelog.String("DocID", key.DocID))
-			if err = p.handleCompletedReplicatorRetry(ctx, peerID, false); err != nil {
-				log.ErrorContextE(ctx, "Failed to handle completed replicator retry", err)
+			failed = true
+			// Still saturated after the wait budget: stop the sweep and let the
+			// next interval try again. Any other failure is specific to this
+			// document; keep its record and carry on with the rest.
+			if strings.Contains(err.Error(), "at capacity") {
```
fredcarle: suggestion: Use the same pattern as above to avoid a different behavior (more on that in the next one):

```go
if _, transient := backpressureWait(err); transient {
```

internal/db/p2p/replicator.go
```diff
@@ -767,24 +770,66 @@ func (p *P2P) retryReplicator(ctx context.Context, peerID string) {
...
 }
 
+// backpressureWait maps a receiver's transient rejection to how long to wait
+// before pushing the same document again.
+func backpressureWait(err error) (time.Duration, bool) {
+	msg := err.Error()
+	switch {
+	case strings.Contains(msg, "rate limited"):
+		return 100 * time.Millisecond, true
+	case strings.Contains(msg, "at capacity"):
```
fredcarle: todo: Please define these strings as constants and document where they come from so that we can try to keep them in sync (high risk of drift here).

PR #5269, opened 2026-09-16
Shinzo - carfile exchange

dvncan:
## Description

Replaces the push-based CAR exchange with request/response. An update announcement now carries only the CID and root block; the receiver asks for the CAR only once it has decided it needs the document.

Announce — pubsub update or replicator push carries CID + root block, no CAR.
Check — receiver runs its existing checks first: already merged, CID matches, access, replication filter. If any says no, the exchange stops and costs no CAR bytes.
Request — over /defradb/car_req/0.0.1, one request per inbound message. Asks the update's creator first, then the delivering peer for anything the creator didn't serve. A failing peer is skipped for 30s.
Serve — sender applies the same hasAccess check BitSwap uses, builds or takes from cache, replies. An empty entry means it wouldn't or couldn't serve that head.
Import — receiver verifies each CAR is rooted at the CID it asked for. Anything unserved falls back to syncDAG over BitSwap.

## Tasks


internal/db/p2p/car_cache.go
```diff
@@ -0,0 +1,124 @@
...
+		c.mu.Lock()
+		delete(c.building, key)
+		if b.err == nil && len(b.data) > 0 {
+			c.add(key, b.data)
+		}
+		c.mu.Unlock()
+		close(b.done)
+	}()
+	b.data, b.err = build()
+	return b.data, false, b.err
```
tzdybal, line 121: In case of panic, `b.err` will be `nil`. All waiting callers will get empty return with no error.

internal/db/p2p/car_cache.go
```diff
@@ -0,0 +1,124 @@
...
+		data, err = build()
+		return data, false, err
+	}
+
+	c.mu.Lock()
+	if el, ok := c.entries[key]; ok {
+		c.order.MoveToFront(el)
+		c.mu.Unlock()
+		// entries only ever holds elements whose Value is a *carEntry.
+		return el.Value.(*carEntry).data, true, nil //nolint:forcetypeassert
```
tzdybal: Shouldn't `data` be retrieved under the mutex locked?

tzdybal (commented):
I found two small issues. One is potential race and second is invalid error handling.

internal/db/p2p/car_cache.go
dvncan: done in commit 9e46689

internal/db/p2p/car_cache.go
dvncan: done in commit 9e46689

tzdybal (approved):
lgtm

Issue #5193, opened 2026-08-25
Cross-version external node wait cannot express a head that is withheld by ACP


islamaliev:
## Problem

When a node runs as a separate process (`NodeConfig.Version`, used for cross-version
testing) the harness cannot read its event bus, so `WaitForSync` falls back to polling
the node for the commits it expects (`waitForHeadsOnNode`,
`tests/integration/events.go:361`).

That poll waits for every pending head. It has no way to express "this head is never
going to arrive, and that is the correct behaviour".

Document ACP makes that case real. A test can create two documents that share a field
block and grant a peer access to only one of them. The sending node's access check then
correctly withholds the blocks of the document the peer was not granted, so one of the
expected heads never arrives. The poll keeps asking for 30 seconds and the test fails
with:

```
timeout waiting for commits to sync to external node
still missing: [bafyreihc2v653hbb7hydzf3mry3ngudpspczz4yi3475t32nih762euhea]
```

The node is behaving correctly. The wait is asking the wrong question.

## Affected tests

`TestACP_P2PSyncSharedFieldBlock_AccessToFirstDoc` and
`TestACP_P2PSyncSharedFieldBlock_AccessToSecondDoc`
(`tests/integration/acp/dac/p2p/sync_shared_field_test.go`).

Both fail only under `cross-version-new-source`. The direction decides which node runs
as the external process, and so which side's withheld block ends up being polled for
over HTTP.

## Not an auth problem

Worth stating, because it looks like one at first. Every poll returns HTTP 200 with a
consistent 169 byte body, so the queries are answered rather than rejected. Non-ACP
networked tests pass under the same multiplier. The block is withheld by design, not
hidden by a missing identity.

This is also distinct from #5123, which is about writes that do not name a node, and
from #5170, which is about restarting an external node.

## Possible directions

- Let the expectation carry the fact that a head is not expected to arrive, so the
  external wait can stop asking for it.
- Have the wait consult the same access decision the sender used, so it only waits for
  what the peer is allowed to receive.

Either way this needs a way to say "not arriving is the expected outcome", which the
external wait path currently cannot represent.

## Meanwhile

The two tests are excluded from both cross-version multipliers with a comment pointing
here.

Part of #5066.

PR #5270, opened 2026-09-16
test(i): Read a head as the collection creator

### LINKED ISSUE #5196: Cross-version head read-back is unauthenticated for documents written without an identity
## Problem

`MarkDocsExpectedOnTargets` (`tests/action/utils_events.go`) works out what a peer should
receive by asking the source node for the head of each document it wrote. A node running
as a separate process has no event bus the harness can read, so this query is the only
way to learn the head.

The query runs as the identity the write used. When a document is written without an
identity into a collection gated by document ACP, that identity is `None`, so the
read-back is unauthenticated and returns nothing. The head cannot be found and the run
fails at the `require.True` guarding it:

```
node 0 could not report the head of bae-fc882d4f-f054-5a63-8aa8-c1c7817c27af
    tests/action/utils_events.go:157
```

The document is present and the write succeeded. Only the read-back is blind.

## Affected tests

The three tests in `tests/integration/acp/dac/branchable/peer_test.go`. Each adds a
document with no identity to an ACP-gated collection:

```go
&action.AddDoc{
    NodeID:       immutable.Some(0),
    CollectionID: 0,
    Doc:          userDoc,
},
```

Confirmed by experiment: making the read-back use an identity that can read the
collection lets all three pass.

## Not the same as the neighbouring issues

- #5123 is about writes that do not name a node. These name `NodeID: 0`.
- #5193 fails later, in `waitForHeadsOnNode` (`tests/integration/events.go:415`), when a
  head that ACP deliberately withheld never arrives. This one fails earlier and for a
  different reason: the head is never looked up successfully in the first place.

## Possible direction

The identity used for the read-back does not have to be the identity that performed the
write. It has to be one that can read the document. Some way of resolving a reader for
the collection, rather than reusing the writer's identity, would close this.

## Meanwhile

The three tests are excluded from both cross-version multipliers with a comment pointing
here.

islamaliev:
## Relevant issue(s)

Resolves https://github.com/sourcenetwork/defradb/issues/5196

## Description

For asserting a certain document or head arrived to a certain node we need to know which CID to wait for and when there's an external node that wrote the document that is under ACP there was no way to request the head because it was gated by the collection ACP. So this PR adds the explicitly identity of the writer node in the request, so that we know which head to wait for.

Two of the three tests this unblocks still hit https://github.com/sourcenetwork/defradb/issues/5193 and stay excluded, now pointing there.

## Tasks

## How has this been tested?

Unit tests, plus the networked and access control suites under both cross-version directions. 

- macOS


nasdf (approved):
LGTM

fredcarle (approved):
LGTM

shahzadlone (approved):
LGTM

PR #5274, opened 2026-09-21
fix: Count a merged counter increment once

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/4935

## Description

A counter that several peers increment at the same time now ends on the right total. Before this, it could settle higher than the sum of every increment, and the peers could disagree with each other. The wrong number was saved, so restarting a node did not clear it.

Each increment is stored as its own block, and a node adds up the ones it has not seen yet. Deciding what counts as "not seen yet" is where it went wrong. When a peer sends changes that are older than what the node already has, the node looks back through its own history to find where the two sides last agreed. While doing that it was overwriting the very list it uses to recognise what it has already added up, so increments it had counted minutes ago looked new again and were added a second time. A normal field never showed this, because storing the same value twice is the same as storing it once. A counter adds it twice.

A node now checks whether an increment is already included in its total before adding it, rather than trusting that list to still be accurate.

There was a second problem in the same place. A document can have several latest versions at once, each with different height, and the node was keeping only one of those numbers, whichever it happened to read last. So the same merge could behave differently from one run to the next. It now uses the earliest of them, so the result no longer depends on iteration order.

## Tasks

## How has this been tested?

Unit and integration tests, plus the race detector on the networked suites.

- macOS


nasdf (approved):
Thanks for fixing!

internal/db/merge.go
```diff
@@ -425,6 +457,157 @@ func (mp *mergeProcessor) processBlock(
...
+// counted. The blockstore's IsMerged flag cannot stand in for this: it is keyed by block while
+// applying is per document, and UpdateHeads clears it for a block's links before they are
+// applied.
+func (mp *mergeProcessor) isAlreadyApplied(
+	ctx context.Context,
+	block *coreblock.Block,
+	blockLink cidlink.Link,
+) (bool, error) {
+	// No value to double-count, and walking them is what tells the field blocks below which
+	// document they belong to.
```
AndrewSisley: suggestion: I think the code is currently correct, given the current behaviour of these blocks, but it might not always be so, and I think the comment is incorrect.

I think the reason it is correct to return early for these types *right now* is that composite blocks are not CRDTs, and collection blocks *currently* self-de-duplicate as the internal data structures inherently prevent adding a document to a collection multiple times (due to store key format)

Whilst composite blocks are not CRDTs, and probably never will be, I'm not sure that the branchable collection CRDT will always function this way, and the code here *might* need to change in the future.

I doubt it is sensible to change the code to accommodate this future possibility, but I think the comment should be corrected - it is currently misleading, and that may cost us in the future should changes here ever be required.

internal/db/merge.go
```diff
@@ -425,6 +457,157 @@ func (mp *mergeProcessor) processBlock(
...
+	for len(stack) > 0 {
+		current := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+
+		if _, ok := seen[current]; ok {
+			continue
+		}
+		seen[current] = struct{}{}
+
+		nd, err := mp.blockLS.Load(
```
AndrewSisley: thought: I hate that this could in theory be fetching blocks from a remote peer if the constructor for `mp` is changed slightly.  It made me chase down to make sure that it wasn't already configured in such a way...

No action required, just venting :grin:

internal/db/merge.go
```diff
@@ -425,6 +457,157 @@ func (mp *mergeProcessor) processBlock(
...
+			linking.LinkContext{Ctx: ctx},
+			cidlink.Link{Cid: current},
+			coreblock.BlockSchemaPrototype,
+		)
+		if err != nil {
+			// Not held locally, so it proves nothing about what is below it. Skipping risks
+			// applying a block twice, never dropping one.
```
AndrewSisley: todo: There are other errors that can be returned here besides `key not found` - please either handle them, or adjust the documentation to make it clearer as to why it is also safe to ignore those (such as IO errors).

internal/db/merge.go
```diff
@@ -425,6 +457,157 @@ func (mp *mergeProcessor) processBlock(
...
+
+	set := &ancestorSet{seen: seen, floor: floor}
+	mp.ancestorCache[cacheKey] = set
+	return set, nil
+}
+
+// walkAncestors walks back from the given heads, stopping at floor.
+//
+// Links only point backwards, so a block below floor cannot lead back to one above it. Stopping
+// there keeps the walk off the whole history.
```
AndrewSisley: nitpick: I feel like there is a word missing from this last line, maybe the below?
```
// there keeps the walk off of the whole history.
```

internal/db/merge.go
```diff
@@ -425,6 +457,157 @@ func (mp *mergeProcessor) processBlock(
...
+func (mp *mergeProcessor) walkAncestors(
+	ctx context.Context,
+	heads []cid.Cid,
+	floor uint64,
+) (map[cid.Cid]struct{}, error) {
+	seen := make(map[cid.Cid]struct{}, len(heads))
+	stack := make([]cid.Cid, 0, len(heads))
+	stack = append(stack, heads...)
+
+	for len(stack) > 0 {
```
AndrewSisley: question: We have very ridged rules as to which head is returned from a query (i.e. cached in the datastore).  Why is the code here operating across all the heads when only one branch will ever be merged and returned to users at any given moment?

Why not just handle that head and its children?

AndrewSisley (approved):
Approving now, but please make sure that the todo is resolved, and the question is answered before merge.  The PR itself was great and easy to review - thanks for the effort put into it!

PR #5275, opened 2026-09-21
fix: Prevent leak during partial node creation

### LINKED ISSUE #5127: Leak is created during partial node creation
Look at lines 150 through 172 of `node/node.go`:

The `Start` function opens rootstore via NewStore, and folds them into n.DB when db.NewDB succeeds. Later, n.Close will tear them down, but that assumes that db.NewDB succeeded. However, if a function such as `n.startP2P` fails, the function will immediately return an error... but all of the previously created local variables (rootstore, docuumentACP, nodeACP, etc.) will be orphaned.

 It was discussed in the review of an open PR for the Java client that this would also create a situation where the node retains its badger lock on a directory.

**This should not be done until the following issue is merged:** https://github.com/sourcenetwork/defradb/pull/5099

ChrisBQu:
## Relevant issue(s)

Resolves #5127 

## Description

 If `Start `failed after opening the store (for example in `startP2P`), the store, ACPs and peer were never closed. The store kept its Badger directory lock. Cleanup only occurred inside `Close`, and that function assumes that everything succeeded. With this PR, `Start` now has a deferred cleanup that releases everything it created in the case of early return due to error.

As a minor adjustment to accomodate this, `NewDB`'s result is now assigned to n.DB only on success. Before, a failed call left a non-nil interface holding a nil *DB, which would make the call to `Close` panic.

This PR also adds a new unit test, `TestStartFailureReleasesStore`. It forces a P2P failure, which makes the call to `Start` fail, and then starts a second node on the same path to verify the Badger lock was released. It would fail without the changes introduced in this PR. 

## Tasks

## How has this been tested?

- WSL


nasdf (approved):
Nice work!

shahzadlone (approved):

PR #5276, opened 2026-09-22
fix: Two lost-update bugs in peer sync

islamaliev:
## Relevant issue(s)

Refs https://github.com/sourcenetwork/defradb/issues/5066

## Description

Two bugs that lost or delayed updates between peers, both found while chasing the branchable sync tests that fail intermittently in CI.

A write announces itself on the document topic and then on the collection topic. A failure on the first skipped the second, and a peer subscribed to a branchable collection hears about a document commit only through the collection topic, so one transient error dropped the update from the network for good. The topic is joined and closed around every publish, so two writes landing together race on the join and the single retry is not always enough.

Separately, a document sync asks its peers for heads and waits for each to answer. The waiting set was keyed by address while a reply carries only the peer id, so an answer never cleared its sender and every sync ran to its five second timeout. The heads still arrived, just always late, and under load that wait collided with the deadlines around it.

Also lets a test node in another process be started with p2p off, which was silently ignored, and gates the p2p-disabled test on v1.1.0, since v1.0.0 panics where it expects an error.

## Tasks

## How has this been tested?

Unit and integration tests. The publish bug only reproduces with several test processes competing for cpu, which is why it shows up in CI and not on a laptop: 14 failures in 450 runs before, none after. The sync wait shows up as a timing change, with those tests dropping from five seconds to under a fifth of one.

- macOS


tests/action/node_options.go
```diff
@@ -60,6 +60,8 @@ type NodeSetupConfig struct {
 	// NodeACP overrides the node's access control settings when set. While it is on
 	// the node lets through no one but the identity it was started with.
 	NodeACP immutable.Option[options.NodeACPOptions]
+	// DisableP2P starts the node with no p2p system at all.
+	DisableP2P bool
```
AndrewSisley: suggestion:  A similar thing was flagged on another PR (I think one of @ChrisBQu 's) - adding flattened config options one by one to this type scales horribly, and allows us to do other horrible things that obfuscate what the real production API looks like from the tests. 

I have a feeling Chris had this with this exact property, so he might be able to point you to something that already exists, but either way please avoid continuing this pattern.

Perhaps `NewNode.Network` should become `immutable.Option[ConfigureNode]`, so that we can differentiate between a default value, and an explicit nil within `NewNode.P2POptions`.

AndrewSisley (approved):
LGTM, thanks Islam.

suggestion: The PR title isn't going to be as helpful as it probably should be in the change log.  I don't know if I would consider it worth the bother to split the PR if I was the author and had already (for whatever reason) bundled them together, and I don't have any quick suggestions, but if you are able to improve it it would be nice :)

tests/action/node_options.go
islamaliev: done

